PR #17486, opened 2017-08-07
*: un-singletonize the cluster settings

tbg:
Apologies to drop such a huge refactoring, but it was a bit of a coup de force
required (or at least strongly suggested) by my inflight PR on version
migration (#17411).

We were previously using our singleton-based `settings` package that originally
replaced the use of environment variables. Its use, however, has since
diversified, the cluster version setting being the most elaborate addition.

In #17411 it has turned out that with these singletons in place, a lot of
effort and hacks are required to even be able to run tests at the same version.
However, it would be nice to test nonidentical versions in cluster tests.

Consequently, this PR

- introduces a struct `cluster.Settings` which now holds all the settings
  previously scattered around the code base; and
- plumbs this struct everywhere it's needed.

There are some open FIXMEs but I'd like a reviewer to review the bulk that's
here in the initial commit (which should pass all tests). I will then address
the FIXMEs (which require more focused attention) in follow-up commits, before
merging this PR.

PR #18001, opened 2017-08-29
setting: move back to a global registry

### LINKED ISSUE #17512: settings: move back to a global (but indexing-only) registry
#17486 removed the singletons from the cluster settings package in a large refactor. Unfortunately, that refactor went a bit further than reviewers had hoped: it centralizes the registration of all settings in the `settings/cluster` package.

@RaduBerinde pointed out that we could keep a global registry which packages can call into to register a setting (which allows setting registration to remain close to settings usage).

This new "registry" wouldn't be much like the old. It would only store a mapping from (say) cluster setting key name to an index (which is incremented on each registration call into the registry).

Additionally, there is a matching `Settings` object that is passed around (today's `cluster.Settings`) which is then used to access the previously-defined settings.

I think it's worth setting this up in a toy example and perhaps write up in a small RFC. For example, it's not clear to me what the `Settings` object will look like. I think the intention in the suggestion above is that one does not have to make any changes to the `Settings` object when registering a new setting, but that means that `Settings` will be opaque, with only some methods to access each setting by id. This means that settings need to implement a common interface, something they don't today, and there are other issues to bikeshed such as how callbacks are registered.

@RaduBerinde, could you share your thoughts on this?

RaduBerinde:
This change moves back the registration of cluster setting to global statements
in the appropriate packages while retaining the benefits achieved by #17486.

We separate the immutable aspects of a setting (like description, default value)
from the mutable aspect (current value). The Setting now represents the
immutable aspect; the values are stored in a separate container
`settings.Values`. Each setting has a corresponding "slot" in the container;
this mapping is static. Each server has its own values container (part of
`cluster.Settings`). Of course, we now have to pass a reference to this
container wherever we get or set a setting.

Fixes #17512.

### DIFF
--- a/docs/RFCS/20170608_decommission.md
+++ b/docs/RFCS/20170608_decommission.md
@@ -134,7 +134,7 @@ states:
 
 Regardless of which state the dead node has, it cannot set its `Draining` flag
 because it is dead. Instead, its leases will expire and will be taken by other
-nodes. After `COCKROACH_TIME_UNTIL_STORE_DEAD` elapses, the replicas of its
+nodes. After `server.time_until_store_dead` elapses, the replicas of its
 ranges will actively be rebalanced. Wait for the replica count to reach 0 as for
 live nodes.
 
@@ -147,13 +147,13 @@ decommissioning process.
 No changes required. The existing CockroachDB process, described below, is
 sufficient.
 
-During a temporary removal, `COCKROACH_TIME_UNTIL_STORE_DEAD` could be updated
+During a temporary removal, `server.time_until_store_dead` could be updated
 to to the length of the downtime to avoid unnecessary movement of ranges.
 However, it is difficult to predict the length of downtime. This is an
 optimization which could be implemented later.
 
 After a node is detected as unavailable for more than
-`COCKROACH_TIME_UNTIL_STORE_DEAD` (an env variable with default: 5 minutes), the
+`server.time_until_store_dead` (an env variable with default: 5 minutes), the
 node is marked as incommunicado and the cluster decides that the node may not be
 coming back and moves the data elsewhere. Its leases would have been transferred
 to other nodes as soon as they expired without being renewed. Ranges are
@@ -211,7 +211,7 @@ be nice to proactively detect this but the effort required is disproportionate
 compared to the gain.
 
 Recently dead nodes may cause decommissioning to hang until
-`COCKROACH_TIME_UNTIL_STORE_DEAD` elapses.
+`server.time_until_store_dead` elapses.
 
 Recommissioning, i.e. undoing a decommission, requires node restart. We could
 avoid this: if the node is live, the coordinating process can directly tell it
--- a/docs/RFCS/20170815_version_migration.md
+++ b/docs/RFCS/20170815_version_migration.md
@@ -466,31 +466,11 @@ func (ecv *ExposedClusterVersion) BootstrapVersion() ClusterVersion
 func (ecv *ExposedClusterVersion) IsActive(v roachpb.Version) bool
 ```
 
-The remaining complexity lies in how `version *settings.StateMachineSetting` is initialized:
-
-```go
-s.version = r.RegisterStateMachineSetting("version",
-		"set the active cluster version in the format '<major>.<minor>'.", // hide optional `-<unstable>`
-		versionTransformer(minVersion, serverVersion, func() ClusterVersion {
-			return *s.baseVersion.Load().(*ClusterVersion)
-		}),
-	)
-```
-
-where `versionTransformer` contains all of the update logic (mod reading from
-the table: we've updated the settings framework to use the table for all
-`StateMachineSettings`, of which this is the only instance at the time of
-writing):
-
-```go
-func versionTransformer(
-	minSupportedVersion, serverVersion roachpb.Version, defaultVersion func() ClusterVersion,
-) settings.TransformerFn {
-  // ...
-}
-```
-
-The returned `settings.TransformerFn` takes
+The remaining complexity lies in the transformer function for
+`version *settings.StateMachineSetting`. It contains all of the update logic
+(mod reading from the table: we've updated the settings framework to use the
+table for all `StateMachineSettings`, of which this is the only instance at the
+time of writing); `versionTransformer` takes
 
 - the previous encoded value (i.e. a marshalled `ClusterVersion`)
 - the desired transition, if any (for example "1.2").
--- a/pkg/ccl/sqlccl/backup.go
+++ b/pkg/ccl/sqlccl/backup.go
@@ -600,7 +600,7 @@ func backupPlanHook(
 	}
 
 	if err := utilccl.CheckEnterpriseEnabled(
-		p.ExecCfg().Settings, p.ExecCfg().ClusterID(), p.ExecCfg().Organization.Get(), "BACKUP",
+		p.ExecCfg().Settings, p.ExecCfg().ClusterID(), p.ExecCfg().Organization(), "BACKUP",
 	); err != nil {
 		return nil, nil, err
 	}
@@ -804,7 +804,7 @@ func showBackupPlanHook(
 	}
 
 	if err := utilccl.CheckEnterpriseEnabled(
-		p.ExecCfg().Settings, p.ExecCfg().ClusterID(), p.ExecCfg().Organization.Get(), "SHOW BACKUP",
+		p.ExecCfg().Settings, p.ExecCfg().ClusterID(), p.ExecCfg().Organization(), "SHOW BACKUP",
 	); err != nil {
 		return nil, nil, err
 	}
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -436,10 +436,10 @@ func writeRocksDB(
 	cache := engine.NewRocksDBCache(0)
 	defer cache.Release()
 	r, err := engine.NewRocksDB(engine.RocksDBConfig{
-		RocksDBSettings: cluster.MakeTestingClusterSettings().RocksDBSettings,
-		Dir:             rocksdbDir,
-		MaxSizeBytes:    0,
-		MaxOpenFiles:    1024,
+		Settings:     cluster.MakeTestingClusterSettings(),
+		Dir:          rocksdbDir,
+		MaxSizeBytes: 0,
+		MaxOpenFiles: 1024,
 	}, cache)
 	if err != nil {
 		return 0, errors.Wrap(err, "create rocksdb instance")
--- a/pkg/ccl/sqlccl/restore.go
+++ b/pkg/ccl/sqlccl/restore.go
@@ -857,7 +857,7 @@ func restorePlanHook(
 		return nil, nil, nil
 	}
 	if err := utilccl.CheckEnterpriseEnabled(
-		p.ExecCfg().Settings, p.ExecCfg().ClusterID(), p.ExecCfg().Organization.Get(), "RESTORE",
+		p.ExecCfg().Settings, p.ExecCfg().ClusterID(), p.ExecCfg().Organization(), "RESTORE",
 	); err != nil {
 		return nil, nil, err
 	}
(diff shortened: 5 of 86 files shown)


petermattis:
:lgtm: though I mainly focused on the changes in the `settings` package.

---

Review status: 0 of 86 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/settings/setting.go, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQcsM2y1ZrT0uKN2J:-KsjQcsM2y1ZrT0uKN2K:bcqriqy) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L22)):*
> ```Go
> )
> 
> const maxSettings = 128
> ```

Can you determine the number of settings from `Registry` in `Values.Init`? 

---

*[pkg/settings/setting.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQNOXG7Rocv-hPyYe:-KsjQNOXG7Rocv-hPyYf:b-prm8pw) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L30)):*
> ```Go
> type Values struct {
> 	intVals     [maxSettings]int64
> 	genericVals [maxSettings]atomic.Value
> ```

The alternative to special casing `int` vs `generic` would be to have `[maxSettings]unsafe.Pointer` and require the `Setting` to initialize a pointer of the correct type. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-:-KsjRKg_5AKK9hxhCzIM:b-7onga0)*

petermattis:
Review status: 0 of 86 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/settings/setting.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQNOXG7Rocv-hPyYe:-KsjRiWuWE7ExGUXcN2n:b-3r6ts6) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L30)):*

The alternative to special casing `int` vs `generic` would be to have `[maxSettings]unsafe.Pointer` and require the `Setting` to initialize a pointer of the correct type. 
</blockquote></details>

I should mention that the current approach is fine too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001)*

RaduBerinde:
Review status: 0 of 86 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/settings/setting.go, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQcsM2y1ZrT0uKN2J:-KsjXImBV7BpxD7gJj0I:b-g9ulu7) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L22)):*

Can you determine the number of settings from `Registry` in `Values.Init`? 
</blockquote></details>

TFTR! I aimed for keeping them statically allocated to avoid indirection. There are only 40 or so settings so this should be plenty. I can switch to dynamically allocating them if you feel strongly.

---

*[pkg/settings/setting.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQNOXG7Rocv-hPyYe:-KsjXDYZDxQTcpAXjp-m:bfxs1na) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L30)):*

I should mention that the current approach is fine too.
</blockquote></details>

I believe we have some`Get` calls on critical paths, it's likely better to avoid extra indirection.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001)*

dt (approved):
Glanced over the settings parts and a few of the usages (assuming the rest are pretty mechanical). Really happy to see the central registry gone before 1.1 is cut!

petermattis:
Review status: 0 of 86 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/settings/setting.go, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQcsM2y1ZrT0uKN2J:-KsjY8XgzSucOEXGM1mg:b-elubex) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L22)):*

TFTR! I aimed for keeping them statically allocated to avoid indirection. There are only 40 or so settings so this should be plenty. I can switch to dynamically allocating them if you feel strongly.
</blockquote></details>

You think there is a perf difference between indexing into `[maxSettings]int64` and `[]int64`? Surprised if there is anything measurable there. Fine the way it is, though.

---

*[pkg/settings/setting.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQNOXG7Rocv-hPyYe:-KsjiCnkxHPlWZ3frFp1:bbccsxk) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L30)):*

I believe we have some`Get` calls on critical paths, it's likely better to avoid extra indirection.
</blockquote></details>

I'd be surprised if there is a performance difference between `[maxSettings]int64` and `[maxSettings]unsafe.Pointer` where the `unsafe.Pointer` is an `*int64`. Yes, there would be an extra indirection. Regardless, this is fine as-is. Let's get this merged.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001)*

RaduBerinde:
Review status: 0 of 86 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/settings/setting.go, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KsjQcsM2y1ZrT0uKN2J:-KsjkuZsYvnEZLz1I2Ic:b4b63mz) ([raw file](https://github.com/cockroachdb/cockroach/blob/03e9c44a3f2cdfbc9e2b0deb27b0bab379663490/pkg/settings/setting.go#L22)):*

You think there is a perf difference between indexing into `[maxSettings]int64` and `[]int64`? Surprised if there is anything measurable there. Fine the way it is, though.
</blockquote></details>

You're probably right.. I'll leave things as they are for now though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001)*

tschottdorf:
:lgtm:, let's get this merged! Thanks for this major cleanup, @raduberinde.

---

Reviewed 75 of 86 files at r1, 11 of 11 files at r2.

---

*[pkg/ccl/utilccl/license_check.go, line 21 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslHmVnr6C5BI1M0KE8:-KslHmVnr6C5BI1M0KE9:b-wppc3w) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/ccl/utilccl/license_check.go#L21)):*
> ```Go
> )
> 
> // cluster.ValidateEnterpriseLicense,
> ```

Comment needs loving (or removal).

---

*[pkg/server/updates.go, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslITmr_EpjlJybvERc:-KslITmr_EpjlJybvERd:b-vmmom0) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/server/updates.go#L75)):*
> ```Go
> )
> 
> // FIXME(tschottdorf): should be NonNegative?
> ```

I think so.

---

*[pkg/settings/float.go, line 36 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslIp4ET4LXer35x8pP:-KslIp4ET4LXer35x8pQ:bovl059) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/settings/float.go#L36)):*
> ```Go
> // Get retrieves the float value in the setting.
> func (f *FloatSetting) Get(sv *Values) float64 {
> 	return math.Float64frombits(uint64(sv.getInt64(f.slotIdx)))
> ```

Heh, was going to complain about the inconsistent camel casing, but.. definitely not your code :)

---

*[pkg/settings/setting.go, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslJRQLy-U90Cq52KCm:-KslJRQLy-U90Cq52KCn:b-g4aojj) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/settings/setting.go#L90)):*
> ```Go
> // Cannot be called concurrently with setGeneric/setInt64.
> func (sv *Values) setOnChange(slotIdx int, fn func()) {
> 	sv.onChange[slotIdx] = fn
> ```

may wanna panic here if a callback is overwritten (as I don't think we intentionally do this).

---

*[pkg/settings/cluster/settings.go, line 239 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslK4P-qtlGKr11jSeh:-KslK4P-qtlGKr11jSei:b-807afv) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/settings/cluster/settings.go#L239)):*
> ```Go
> }
> 
> // versionTrasnformer is the transformer function for the version StateMachine.
> ```

Typo.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-:-KslL5GOJui_wpGlPBvu:bmugn5t)*

RaduBerinde:
Review status: 82 of 86 files reviewed at latest revision, 7 unresolved discussions.

---

*[pkg/ccl/utilccl/license_check.go, line 21 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslHmVnr6C5BI1M0KE8:-KsnHrXQY4AL-WSErupi:ba3dfob) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/ccl/utilccl/license_check.go#L21)):*

Comment needs loving (or removal).
</blockquote></details>

Removed

---

*[pkg/server/updates.go, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslITmr_EpjlJybvERc:-KsnI-UTJP9CzQmfjA0V:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/server/updates.go#L75)):*

I think so.
</blockquote></details>

Done.

---

*[pkg/settings/setting.go, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslJRQLy-U90Cq52KCm:-KsnIAw-42okwkInzja6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/settings/setting.go#L90)):*

may wanna panic here if a callback is overwritten (as I don't think we intentionally do this).
</blockquote></details>

Done.

---

*[pkg/settings/cluster/settings.go, line 239 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001#-KslK4P-qtlGKr11jSeh:-KsnIGXQ835Bg3EFkNiD:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/669cf1f5c49b648a66095b69cb0217ac0f15d34e/pkg/settings/cluster/settings.go#L239)):*

Typo.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18001)*

PR #18003, opened 2017-08-29
sqlccl: display zero start times as null

dt:


danhhz:
:lgtm:

---

---

*[pkg/ccl/sqlccl/backup_test.go, line 1896 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjZGapTy2a-_28XrmK:-KsjZGapTy2a-_28XrmL:b-rcdp9p) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup_test.go#L1896)):*
> ```Go
> 		&unused, &unused, &start, &end, &dataSize,
> 	)
> 	if start != nil {
> ```

should probably add to this test and check that this _is_ set for incremental backups

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-:-KsjZMuQYpz-JehLb605:bnfp4nl)*

dt:

---

*[pkg/ccl/sqlccl/backup_test.go, line 1896 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjZGapTy2a-_28XrmK:-KsjZW7hBKegPW8EPBlY:bqd3k5n) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup_test.go#L1896)):*

should probably add to this test and check that this _is_ set for incremental backups
</blockquote></details>

i was hoping you wouldn't notice that :/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

mjibson:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-KsjdxscOPXLH2leCX8S:bxhkcjc) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*
> ```Go
> 		start := parser.DNull
> 		if desc.StartTime.WallTime != 0 {
> 			start = parser.MakeDTimestamp(time.Unix(0, desc.StartTime.WallTime), time.Nanosecond)
> ```

I think it's better to use Microseconds here because then users can correctly use `SHOW BACKUP` in square brackets as input to other stuff.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

dt:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-KsjeVIfrz7NPA8cDVe-:bukegwp) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*

I think it's better to use Microseconds here because then users can correctly use `SHOW BACKUP` in square brackets as input to other stuff.
</blockquote></details>

@danhhz went back and forth on this one -- IIRC he went with nanos since it has to match exactly when used (e.g. in `incremental from`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

jseldess (approved):
LGTM. Thanks for the quick update, @dt!

mjibson:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-Ksl8dPHmbAeFX_LbXW_:bgod5wt) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*

@danhhz went back and forth on this one -- IIRC he went with nanos since it has to match exactly when used (e.g. in `incremental from`).
</blockquote></details>

Ok. I'd almost prefer this to just be a string because of that. But I'm not opposed to leaving as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

danhhz:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-Ksn9ZV8j-tsi1j7C4Wp:bbd29u6) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*

Ok. I'd almost prefer this to just be a string because of that. But I'm not opposed to leaving as is.
</blockquote></details>

String makes sense, especially if we ever want to make this an hlc timestamp. Not sure why I didn't think of that. Let's do that before we're locked in to backward compatibility

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

mjibson:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-Kso7c4LOg59mx7TlqNY:b-dubg0z) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*

String makes sense, especially if we ever want to make this an hlc timestamp. Not sure why I didn't think of that. Let's do that before we're locked in to backward compatibility
</blockquote></details>

Do we allow backups with an HLC with a non-zero logical part? If so do we need to expose that somehow? Should we have both a human-readable string timestamp here and the full decimal HLC, as separate columns?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

dt:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-KsoSZZeCgcVXnPCd_uT:bn0z4f5) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*

Do we allow backups with an HLC with a non-zero logical part? If so do we need to expose that somehow? Should we have both a human-readable string timestamp here and the full decimal HLC, as separate columns?
</blockquote></details>

@mjibson We're currently using `SELECT * FROM [SHOW BACKUP $1]` and scanning the results into `time.Time`s and comparing them to other times in our tests, which seems not entirely an unexpected use-case, and switching to strings would make it a bit more cumbersome. That said, we truncate to micros in pgwire even if we don't here so i'm not sure how much it matters?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

mjibson:

---

*[pkg/ccl/sqlccl/backup.go, line 863 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjdxscOPXLH2leCX8R:-KsoXBTd9TLL0uPm_HNI:bhyhbgk) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup.go#L863)):*

@mjibson We're currently using `SELECT * FROM [SHOW BACKUP $1]` and scanning the results into `time.Time`s and comparing them to other times in our tests, which seems not entirely an unexpected use-case, and switching to strings would make it a bit more cumbersome. That said, we truncate to micros in pgwire even if we don't here so i'm not sure how much it matters?
</blockquote></details>

I'm totally ok leaving it as a timestamp with nanos, just as long as we understand the problems, like if you do `SELECT * FROM [SHOW BACKUP $1]` then try to restore from one of those exact times it may not exist because of the nano -> micro change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

dt:
okay, I'm gonna go ahead and merge this one then, since it is just about 0 vs null. We can change the type to string or nanos vs micros too, but no need to tie that to this change.

dt:

---

*[pkg/ccl/sqlccl/backup_test.go, line 1896 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003#-KsjZGapTy2a-_28XrmK:-Ksoh1Qo86WlY3-i99DF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d04124f0f0540589eda4b0c6a38dd4101b03352d/pkg/ccl/sqlccl/backup_test.go#L1896)):*

i was hoping you wouldn't notice that :/
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18003)*

PR #17789, opened 2017-08-21
*: Include a hypothetical allocator run in range debug output

a-robinson:
This is somewhat of a proof-of-concept since my typescript is hot
garbage and there may be differing opinions on whether/where/how to
expose this, but the gist is that this will run a replica through the
replicate queue, taking care not to actually make any changes, and
spit back out all the verbose logging messages accumulated throughout
the process.

This is the second of the two pieces needed to make debugging the
allocator easier without needing to enable verbose logging. Step 1 was
including info in the range log (so that we can understand decisions
that have been made), while this helps us understand decisions that
weren't made (i.e. helps us understand why rebalancing isn't happening).

Suggestions very welcome on how to improve everything, particularly the UI aspect 🐶

PR #18004, opened 2017-08-29
storage: Add a new unsatisfiable allocation request metric

### LINKED ISSUE #14174: storage: add a metric for unsatisfiable allocation request
Adding this metric would help diagnose under-replicated ranges in the event that we can't allocate a replica on a node / store with the required zone config constraints.

BramGruneir:
This metric tracks when a range has been trying to upreplicate but cannot.

Closes #14174.

screen shot 2017-08-29 at 15 52 56


BramGruneir:
@a-robinson I wasn't sure if this was the best approach to capture all failed/denied/unable to upreplicate, let me know what you think.

BramGruneir:
*[pkg/storage/storagebase/state.proto, line 105 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-KsojiFos6W0BrGbJ8VW:-KsojiFpXLBVui_-0lw4:blth2zp) ([raw file](https://github.com/cockroachdb/cockroach/blob/3d003e891191c320a5b1255c8fd989f25d594317/pkg/storage/storagebase/state.proto#L105)):*
> ```protobuf
>   // The number of failed uprelication attempts for this range. Resets upon a
>   // successful upreplication.
>   int32 unsatisfiable_allocation_requests = 8;
> ```

Not sure if this should go in RangeInfo or ReplicaState.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004)*

a-robinson:
I have to say this is not what I expected, although that's not necessarily a bad thing. I wasn't expecting it to be per-replica state or to be added to a storagebase proto, which it doesn't seem important enough to be part of.

I'm curious about specifically how this will be used? Do the hypothetical allocator runs from #17789 serve the original purpose of "help[ing] diagnose under-replicated ranges in the event that we can't allocate a replica on a node / store with the required zone config constraints" well enough?

---

Reviewed 9 of 10 files at r1.
Review status: 9 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/storage/replicate_queue.go, line 345 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-KsoqUOi_3t_hzWlcTCM:-KsoqUOj1tJSNz5TYFin:bsi4eh1) ([raw file](https://github.com/cockroachdb/cockroach/blob/3d003e891191c320a5b1255c8fd989f25d594317/pkg/storage/replicate_queue.go#L345)):*
> ```Go
> 			details,
> 		); err != nil {
> 			repl.incUnsatisfiableAllocationRequestCount()
> ```

Do we really want to increment this here? The allocation request was satisfiable, it just failed for some reason. The most likely reason simply being that the destination store's snapshot queue was already in the middle of handling another snapshot.

There may be value in monitoring failed snapshots, but I don't think we want it conflated with the allocation request not being satisfiable. There could easily be other valid stores in the cluster that would accept this replica even if this one didn't on this attempt.

---

*[pkg/storage/replicate_queue.go, line 348 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-Ksos63m1gDih1sgyGNV:-Ksos63m1gDih1sgyGNW:b-dtvmzh) ([raw file](https://github.com/cockroachdb/cockroach/blob/3d003e891191c320a5b1255c8fd989f25d594317/pkg/storage/replicate_queue.go#L348)):*
> ```Go
> 			return false, err
> 		}
> 		repl.resetUnsatisfiableAllocationRequestCount()
> ```

It seems like we'd also have to clear this upon losing (or at least gaining) the lease for the range. Otherwise non-leaseholder replicas could get stuck with high counts forever, well past the point at which things weren't working.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-:-KsosYpNE_8TJX18iJIR:b-ylcgig)*

BramGruneir:
> I have to say this is not what I expected, although that's not necessarily a bad thing. I wasn't expecting it to be per-replica state or to be added to a storagebase proto, which it doesn't seem important enough to be part of.

I agree entirely.  I'm going to move this into serverpb.rangeinfo.  And populate it using metrics, which it looks like I started to do with this PR than skipped for some reason.  So that should make this a lot less invasive.

> I'm curious about specifically how this will be used? Do the hypothetical allocator runs from #17789 serve the original purpose of "help[ing] diagnose under-replicated ranges in the event that we can't allocate a replica on a node / store with the required zone config constraints" well enough?

This is just to help quickly diagnose when we have a range that is not up-replicating for some reason.  What that reason is, would require more investigations and specifically looking at the simulation would help, but not tell us when up-replication is failing.

But here's the real question.  Would this be helpful at all.  If not, I'll just drop it.

---

Review status: 9 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/storage/storagebase/state.proto, line 105 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-KsojiFos6W0BrGbJ8VW:-KsouZROsa1MKNGjtewR:b4feu56) ([raw file](https://github.com/cockroachdb/cockroach/blob/3d003e891191c320a5b1255c8fd989f25d594317/pkg/storage/storagebase/state.proto#L105)):*

Not sure if this should go in RangeInfo or ReplicaState.
</blockquote></details>

I'm going to move this into rangeinfo under serverpb.rangeinfo, instead of here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-:-KsouhZUAnTrAlCUg4xN:b-970ug6)*

a-robinson:
Doesn't the problem ranges page show the ranges that are under-replicated? As far as I can imagine, the three pieces of information we want are:

1. Are any ranges under-replicated?
2. Which ranges?
3. Why?

The problem ranges page seems like a more direct and precise way of getting the first two pieces of information, and the hypothetical allocator runs seem like a more informative source of the third.

Am I missing something? If not, I'm leaning towards saying this isn't needed.

---

Reviewed 1 of 10 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-:-KsowwTyRGTMMla-PVKa:b5egt50)*

BramGruneir:
The "problems" section on the range page even shows underreplicated ranges too.  

Now with the simulator, I think you're right, we can just drop this entirely.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-:-Ksp05kwm_mMiszYu7-L:b9ho9ic)*

mrtracy:
The front end portions currently done are LGTM, although it appears some of this is in flux from a design perspective.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18004#-:-KspEHJYEE2KBlKsfYuf:b-dugn1f)*

BramGruneir:
Closing this as it is not needed with the incoming simulator.

Issue #17524, opened 2017-08-08
user cluster has three wedged timeseries ranges


tschottdorf:
Forked from https://github.com/cockroachdb/cockroach/issues/17478#issuecomment-321046079.

We have access to this cluster via a Cisco WebEx session (you'll need a chrome plugin) at a meeting link that can be activated by Gitter user @HeikoOnnebrink (he'll supervise while you are connected).

It's a 9node cluster running inside of Docker on CoreOS. I've only looked more closely at r104 (see Archive.zip below), which stopped working on 7/31 and then saw some more activity on 8/3: The problematic member here is node1. In grepping the logs, I saw that node1 briefly got the lease on 7/31. The next activity is on 8/3, when it receives 3-4 snapshots including almost no log entries.

There are two other ranges that are perhaps not comparable. In particular, one of the two has a 120mb raft log.

I have no bandwidth to investigate this further. It's fairly tedious due to the remote connection and the fact that this is a 9node cluster. Still, we should follow through and gather what we can. I tried to enable lower-level raft logging via access to /debug/vmodule/raft=8 but somehow it didn't work. That plus grep for `r104/` should turn up *something*.

Inlined my initial investigation comment below:
> 
> > I should also mention that the reason node 5 is unwilling to become raft leader is because it doesn't have an entry for itself in its own progress map (prs map[uint64]*Progress), which make it un-"promotable". This seems pretty surprising, but I'm not familiar with the expectations for it.
>
> I thought the `Progress` is only populated *on* the Raft leader. Which piece of code are you talking about?
>
> I'm currently looking at a user's cluster who also lost his timeseries ranges. Data attached.
>
> There's definitely a problem with the quota pool on that cluster. @irfansharif, any thoughts on the below? It might be an artifact of the ranges being horked, which the attached range pages can hopefully illustrate.
[Archive.zip](https://github.com/cockroachdb/cockroach/files/1209223/Archive.zip)

```
goroutine 35976949 [chan receive, 6999 minutes]:
github.com/cockroachdb/cockroach/pkg/util/timeutil.(*Timer).Reset(0xc422777740, 0xdf8475800)
	/go/src/github.com/cockroachdb/cockroach/pkg/util/timeutil/timer.go:89 +0x9f
github.com/cockroachdb/cockroach/pkg/storage.(*quotaPool).acquire(0xc420e88c30, 0x7fbd9cbebff8, 0xc420e5c3f0, 0x23b, 0x0, 0x0)
[...]

PR #18005, opened 2017-08-29
storage: add timeout to NodeLiveness heartbeat loop

nvb:
While investigating #17524, we found that a NodeLiveness request was stuck, which
blocked the NodeLiveness semaphore. This prevented any liveness updates from
being sent by the source of the stuck liveness update. While introducing a
timeout here on the liveness loop won't fix the issue we saw, it should prevent
stuck liveness updates in the future from preventing a node from ever updating
its liveness record, which makes everything harder to deal with.

### DIFF
--- a/pkg/storage/node_liveness.go
+++ b/pkg/storage/node_liveness.go
@@ -323,25 +323,30 @@ func (nl *NodeLiveness) StartHeartbeat(
 		incrementEpoch := true
 		for {
 			if !nl.pauseHeartbeat.Load().(bool) {
-				ctx, sp := ambient.AnnotateCtxWithSpan(context.Background(), "heartbeat")
-				// Retry heartbeat in the event the conditional put fails.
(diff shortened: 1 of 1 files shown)


bdarnell:
:lgtm: but I don't follow how this would "bring down the entire liveness range". Wouldn't the impact be limited to this node being unable to update its liveness? 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18005#-:-KsngrtZzsTZtYb7iD6a:b-nomdzu)*

tschottdorf:
LGTM. I think the fact that this hit us so hard in our debugging session which motivated this was since there was more going wrong elsewhere. @nvanbenschoten, do you remember the details?

nvanbenschoten:
Yeah, there was definitely more going wrong so this alone "bringing down the entire liveness range" is probably an inaccurate hyperbole. This in concert with something else was having that effect. I changed the wording to be more specific.

PR #18007, opened 2017-08-29
acceptance: allow cluster restarts in localcluster

tbg:
Use that nodes (now) write their gossip address early in the startup
sequence to allow for cluster-wide restarts.

Extend the mixed version harness test slightly to demonstrate that this
works even in a cluster that contains a v1.0 binary (that does *not* write
the gossip address early).

Use `RunLocal` for `TestGossipRestart`, which was previously not possible.

### DIFF
--- a/pkg/acceptance/gossip_peerings_test.go
+++ b/pkg/acceptance/gossip_peerings_test.go
@@ -88,10 +88,18 @@ func TestGossipRestart(t *testing.T) {
 	s := log.Scope(t)
 	defer s.Close(t)
 
+	ctx := context.Background()
+	cfg := readConfigFromFlags()
 	RunLocal(t, func(t *testing.T) {
-		// TODO(bram): #4559 Limit this test to only the relevant cases. No chaos
-		// agents should be required.
-		runTestWithCluster(t, testGossipRestartInner)
+		// TODO(tschottdorf): https://github.com/cockroachdb/cockroach/issues/18027.
+		// When that is addressed, use the standard runner again.
+		if len(cfg.Nodes) > 3 {
+			cfg.Nodes = cfg.Nodes[:3]
+		}
+		c := StartCluster(ctx, t, cfg)
+		defer c.AssertAndStop(ctx, t)
+
+		testGossipRestartInner(ctx, t, c, cfg)
 	})
 }
 
--- a/pkg/acceptance/gossip_peerings_test.go
+++ b/pkg/acceptance/gossip_peerings_test.go
@@ -22,6 +22,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/acceptance/cluster"
+	"github.com/cockroachdb/cockroach/pkg/acceptance/localcluster"
 	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
@@ -87,14 +88,7 @@ func TestGossipRestart(t *testing.T) {
 	s := log.Scope(t)
 	defer s.Close(t)
 
-	// TODO(tschottdorf): at the time of writing, this can't run under RunLocal
-	// simply because that mode uses ephemeral ports, and so restarting a whole
-	// cluster leads to no node knowing where to reach any other node. To change
-	// this, we could introduce an RPC we can send to the cluster that gives it
-	// "hints" for its --join list after it has started. CI servers may cycle
-	// through ports rapidly, so we can't rely that our previous port is available
-	// when a node restarts.
-	RunDocker(t, func(t *testing.T) {
+	RunLocal(t, func(t *testing.T) {
 		// TODO(bram): #4559 Limit this test to only the relevant cases. No chaos
 		// agents should be required.
 		runTestWithCluster(t, testGossipRestartInner)
@@ -139,24 +133,15 @@ func testGossipRestartInner(
 		}
 
 		log.Infof(ctx, "restarting all nodes")
-		if _, ok := c.(*cluster.DockerCluster); ok {
-			// It is not safe to call Restart in parallel when using
-			// cluster.DockerCluster because expected container shutdown events
-			// may be reordered.
-			for i := 0; i < num; i++ {
-				if err := c.Restart(ctx, i); err != nil {
-					t.Errorf("error restarting node %d: %s", i, err)
-				}
-			}
-		} else {
-			ch := make(chan error)
-			for i := 0; i < num; i++ {
-				go func(i int) { ch <- c.Restart(ctx, i) }(i)
-			}
-			for i := 0; i < num; i++ {
-				if err := <-ch; err != nil {
-					t.Errorf("error restarting node %d: %s", i, err)
-				}
+		var chs []<-chan error
+		for i := 0; i < num; i++ {
+			// We need to restart asynchronously because the local cluster nodes
+			// like to wait until they are ready to serve.
+			chs = append(chs, c.(*localcluster.LocalCluster).RestartAsync(ctx, i))
+		}
+		for i, ch := range chs {
+			if err := <-ch; err != nil {
+				t.Errorf("error restarting node %d: %s", i, err)
 			}
 		}
 		if t.Failed() {
--- a/pkg/acceptance/localcluster/cluster.go
+++ b/pkg/acceptance/localcluster/cluster.go
@@ -27,7 +27,9 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
+	"sync/atomic"
 	"text/tabwriter"
 	"time"
 
@@ -117,9 +119,16 @@ func MakePerNodeFixedPortsCfg(numNodes int) map[int]NodeConfig {
 // operations, access to the underlying nodes and per-node KV and SQL clients.
 type Cluster struct {
 	cfg     ClusterConfig
+	seq     *seqGen
 	Nodes   []*Node
 	stopper *stop.Stopper
 	started time.Time
+}
+
+type seqGen int32
+
+func (s *seqGen) Next() int32 {
+	return atomic.AddInt32((*int32)(s), 1)
 }
 
 // New creates a Cluster with the given configuration.
@@ -129,6 +138,7 @@ func New(cfg ClusterConfig) *Cluster {
 	}
 	return &Cluster{
 		cfg:     cfg,
+		seq:     new(seqGen),
 		stopper: stop.NewStopper(),
 	}
 }
@@ -205,14 +215,41 @@ func (c *Cluster) Close() {
 }
 
 func (c *Cluster) joins() []string {
-	var joins []string
+	type addrAndSeq struct {
+		addr string
+		seq  int32
+	}
+
+	var joins []addrAndSeq
 	for _, node := range c.Nodes {
 		advertAddr := node.advertiseAddr()
 		if advertAddr != "" {
-			joins = append(joins, advertAddr)
+			joins = append(joins, addrAndSeq{
+				addr: advertAddr,
+				seq:  atomic.LoadInt32(&node.startSeq),
+			})
 		}
 	}
-	return joins
+	sort.Slice(joins, func(i, j int) bool {
+		return joins[i].seq < joins[j].seq
+	})
+
+	if len(joins) == 0 {
+		return nil
+	}
+
+	// Return the node with the smallest startSeq, i.e. the node that was
+	// started first. This is the node that might have no --join flag set, and
+	// we must point the other nodes at it, and *only* at it (or the other nodes
+	// may connect sufficiently and never bother to talk to this node).
+	//
+	// See https://github.com/cockroachdb/cockroach/issues/18027 and note that this
+	// code is just an unsubstantiated claim. We have observed similar problems with
+	// this code. Likely what's described above is only a theoretical problem, but
+	// it's good to avoid those problems too.
+	//
+	// TODO(tschottdorf): revisit after #18027 above closes.
+	return []string{joins[0].addr}
 }
 
 // IPAddr returns the IP address of the specified node.
@@ -241,6 +278,7 @@ func (c *Cluster) makeNode(ctx context.Context, nodeIdx int, cfg NodeConfig) (*N
 	node := &Node{
 		cfg:    cfg,
 		rpcCtx: rpcCtx,
+		seq:    c.seq,
 	}
 
 	args := []string{
@@ -405,6 +443,9 @@ func (c *Cluster) RandNode(f func(int) int) int {
 type Node struct {
 	cfg    NodeConfig
 	rpcCtx *rpc.Context
+	seq    *seqGen
+
+	startSeq int32 // updated atomically on start
 
 	syncutil.Mutex
 	cmd                      *exec.Cmd
@@ -514,6 +555,7 @@ func (n *Node) setNotRunningLocked() {
 	n.cmd = nil
 	n.rpcPort = ""
 	n.httpPort = ""
+	atomic.StoreInt32(&n.startSeq, 0)
 }
 
 func (n *Node) startAsyncInnerLocked(ctx context.Context, joins ...string) error {
@@ -526,6 +568,8 @@ func (n *Node) startAsyncInnerLocked(ctx context.Context, joins ...string) error
 	n.cmd = exec.Command(n.cfg.ExtraArgs[0], args...)
 	n.cmd.Env = os.Environ()
 	n.cmd.Env = append(n.cmd.Env, n.cfg.ExtraEnv...)
+
+	atomic.StoreInt32(&n.startSeq, n.seq.Next())
 
 	_ = os.MkdirAll(n.logDir(), 0755)
 
--- a/pkg/acceptance/localcluster/cluster.go
+++ b/pkg/acceptance/localcluster/cluster.go
@@ -117,10 +117,7 @@ func MakePerNodeFixedPortsCfg(numNodes int) map[int]NodeConfig {
 // operations, access to the underlying nodes and per-node KV and SQL clients.
 type Cluster struct {
 	cfg     ClusterConfig
-	rpcCtx  *rpc.Context
 	Nodes   []*Node
-	Clients []*client.DB
-	Status  []serverpb.StatusClient
 	stopper *stop.Stopper
 	started time.Time
 }
@@ -132,9 +129,6 @@ func New(cfg ClusterConfig) *Cluster {
 	}
 	return &Cluster{
 		cfg:     cfg,
-		Nodes:   make([]*Node, cfg.NumNodes),
-		Clients: make([]*client.DB, cfg.NumNodes),
-		Status:  make([]serverpb.StatusClient, cfg.NumNodes),
 		stopper: stop.NewStopper(),
 	}
 }
@@ -144,19 +138,11 @@ func New(cfg ClusterConfig) *Cluster {
 // can be used to pass extra arguments to every node. The perNodeArgs parameter
 // can be used to pass extra arguments to an individual node. If not nil, its
 // size must equal the number of nodes.
-func (c *Cluster) Start() {
-	ctx := context.Background()
+func (c *Cluster) Start(ctx context.Context) {
 	c.started = timeutil.Now()
 
-	baseCtx := &base.Config{
-		User:     security.NodeUser,
-		Insecure: true,
-	}
-	c.rpcCtx = rpc.NewContext(log.AmbientContext{Tracer: tracing.NewTracer()}, baseCtx,
-		hlc.NewClock(hlc.UnixNano, 0), c.stopper)
-
-	chs := make([]<-chan error, len(c.Nodes))
-	for i := range c.Nodes {
+	chs := make([]<-chan error, c.cfg.NumNodes)
+	for i := 0; i < c.cfg.NumNodes; i++ {
 		cfg := c.cfg.PerNodeCfg[i] // zero value is ok
 		if cfg.Binary == "" {
 			cfg.Binary = c.cfg.Binary
@@ -177,7 +163,9 @@ func (c *Cluster) Start() {
 			cfg.NumWorkers = c.cfg.NumWorkers
 		}
 		cfg.ExtraArgs = append(append([]string(nil), c.cfg.AllNodeArgs...), cfg.ExtraArgs...)
-		c.Nodes[i], chs[i] = c.makeNode(i, cfg)
+		var node *Node
+		node, chs[i] = c.makeNode(ctx, i, cfg)
+		c.Nodes = append(c.Nodes, node)
 		if i == 0 && cfg.RPCPort == 0 {
 			// The first node must know its RPCPort or we can't possibly tell
 			// the other nodes the correct one to go to.
@@ -197,13 +185,8 @@ func (c *Cluster) Start() {
 
 	for i := range chs {
 		if err := <-chs[i]; err != nil {
-			log.Fatalf(ctx, "node %d: %s", i, err)
+			log.Fatalf(ctx, "node %d: %s", i+1, err)
 		}
-	}
-
-	for i := range c.Nodes {
-		c.Clients[i] = c.makeClient(i)
-		c.Status[i] = c.makeStatus(i)
 	}
 
 	log.Infof(context.Background(), "started %.3fs", timeutil.Since(c.started).Seconds())
@@ -221,14 +204,20 @@ func (c *Cluster) Close() {
 	}
 }
 
+func (c *Cluster) joins() []string {
+	var joins []string
+	for _, node := range c.Nodes {
+		advertAddr := node.advertiseAddr()
+		if advertAddr != "" {
+			joins = append(joins, advertAddr)
+		}
+	}
+	return joins
+}
+
 // IPAddr returns the IP address of the specified node.
 func (c *Cluster) IPAddr(nodeIdx int) string {
 	return c.Nodes[nodeIdx].IPAddr()
-}
-
-// RPCAddr returns the RPC address of the specified node.
-func (c *Cluster) RPCAddr(nodeIdx int) string {
-	return net.JoinHostPort(c.IPAddr(nodeIdx), c.RPCPort(nodeIdx))
 }
 
 // RPCPort returns the RPC port of the specified node. Returns zero if unknown.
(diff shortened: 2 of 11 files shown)


tamird:
Reviewed 1 of 1 files at r1, 6 of 6 files at r2, 7 of 7 files at r3, 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[pkg/acceptance/mixed_version_test.go, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjq910BnPb68i2CG3A:-Ksjq911fl5nA5YiN17a:bfgn9dn) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L101)):*
> ```Go
> 	var chs []<-chan error
> 	// Restart the nodes asynchronously and in a way that doesn't start the
> 	// first node (at v1.0.5 first) (this is due to technical limitations in
> ```

huh.

what limitations? seems likely that some future test might hit this as well, so extra verbosity seems warranted.

---

*[pkg/acceptance/mixed_version_test.go, line 117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqQmTUpi76wTB-lAV:-KsjqQmTUpi76wTB-lAW:b-68tsgm) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L117)):*
> ```Go
> 		t.Fatal(err)
> 	}
> 	var cnt int
> ```

nit: count?

---

*[pkg/acceptance/mixed_version_test.go, line 118 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqOMgLDtKNBKyEF2Z:-KsjqOMgLDtKNBKyEF2_:b9ujub6) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L118)):*
> ```Go
> 	}
> 	var cnt int
> 	if err := db.QueryRow("SELECT COUNT(*) FROM system.settings WHERE name = 'version';").Scan(&cnt); err != nil {
> ```

nit: `COUNT DISTINCT`?

---

*[pkg/acceptance/mixed_version_test.go, line 125 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqWKt7Yn5abiWVEui:-KsjqWKt7Yn5abiWVEuj:b9svni4) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L125)):*
> ```Go
> 	// the version setting should have run.
> 	//
> 	// NB: we could do this check before the restart as well. The new nodes
> ```

why not do that, then?

---

*[pkg/acceptance/localcluster/cluster.go, line 168 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjqfl6AQq7yqq4hKCA:-Ksjqfl6AQq7yqq4hKCB:babevcg) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L168)):*
> ```Go
> 		cfg.ExtraArgs = append(append([]string(nil), c.cfg.AllNodeArgs...), cfg.ExtraArgs...)
> 		var node *Node
> 		node, chs[i] = c.makeNode(i, cfg)
> ```

nit: preallocate `c.Nodes = make([]*Node, c.cfg.NumNodes)` next to where you create the slice of channels? then you could `c.Nodes[i], ch[i] = ...` here.

---

*[pkg/acceptance/localcluster/cluster.go, line 235 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjquoeQlil0FO_cFcL:-KsjquoeQlil0FO_cFcM:b-pgmd8k) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L235)):*
> ```Go
> 
> func (c *Cluster) makeNode(nodeIdx int, cfg NodeConfig) (*Node, <-chan error) {
> ```

random newline

---

*[pkg/acceptance/localcluster/cluster.go, line 240 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqyQ-fFk1O44RZ7Nr:-KsjqyQ-fFk1O44RZ7Ns:b-ta9ckk) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L240)):*
> ```Go
> 		Insecure: true,
> 	}
> 	rpcCtx := rpc.NewContext(log.AmbientContext{Tracer: tracing.NewTracer()}, baseCtx,
> ```

hm, is there a reason it's better to have a separate rpc context for each node?

---

*[pkg/acceptance/localcluster/cluster.go, line 504 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjrOJNficqAHh0pqLi:-KsjrOJNficqAHh0pqLj:b-mkc9vg) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L504)):*
> ```Go
> 
> // Start starts a node.
> func (n *Node) Start(joins ...string) {
> ```

nit: i believe callers of this and StartAsync already have contexts to hand in, might be nice to plumb them while you're here.

---

*[pkg/acceptance/localcluster/cluster.go, line 537 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjreoQjwo55i-2jGHU:-KsjreoQjwo55i-2jGHV:br9arp1) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L537)):*
> ```Go
> 	n.setNotRunningLocked()
> 
> 	args := append([]string(nil), n.cfg.ExtraArgs[1:]...)
> ```

might be better to write `args := n.cfg.ExtraArgs[1:len(n.cfg.ExtraArgs):len(n.cfg.ExtraArgs)]` so that appends don't mutate the original

---

*[pkg/acceptance/localcluster/cluster.go, line 637 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjsJ5LE25VMbsClIzj:-KsjsJ5LE25VMbsClIzk:bz31aem) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L637)):*
> ```Go
> 			_, port, err := net.SplitHostPort(gossipAddr)
> 			if err != nil {
> 				continue
> ```

log, or something? but actually, if gossipAddr is non-empty and this error is non-nil, i think something bad has happened.

---

*[pkg/acceptance/localcluster/cluster.go, line 645 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjsUidxOaUBYF-kDk7:-KsjsUidxOaUBYF-kDk8:b5bdre7) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L645)):*
> ```Go
> 		select {
> 		case <-isServing:
> 			n.Lock()
> ```

what's going on with the locking, here? why not defer the Lock?

---

*[pkg/server/server.go, line 742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjow5uWC3tFhofHSm6:-Ksjow5uWC3tFhofHSm7:br0j7y0) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L742)):*
> ```Go
> 	log.Event(ctx, "started gossip")
> 
> 	statusFiles := [][2]string{
> ```

should this be a map? seems like uniqueness here is a thing

---

*[pkg/server/server.go, line 743 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjpGokY3BrM37anYhA:-KsjpGokY3BrM37anYhB:br5um2v) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L743)):*
> ```Go
> 
> 	statusFiles := [][2]string{
> 		{".advertise-addr", unresolvedAdvertAddr.String()},
> ```

these addresses are all unresolved, which seems like a bit of a sharp edge, which at the very least deserves some tests that demonstrate the behaviour.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007)*

tschottdorf:

---

*[pkg/acceptance/mixed_version_test.go, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjq910BnPb68i2CG3A:-KsjtZbDKxj2AQ0HQ7BS:bobrs96) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L101)):*

huh.

what limitations? seems likely that some future test might hit this as well, so extra verbosity seems warranted.
</blockquote></details>

Added a pointer to `StartAsync`.

---

*[pkg/acceptance/mixed_version_test.go, line 117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqQmTUpi76wTB-lAV:-Ksjto_BbV3o0MAKguG7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L117)):*

nit: count?
</blockquote></details>

Done.

---

*[pkg/acceptance/mixed_version_test.go, line 118 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqOMgLDtKNBKyEF2Z:-Ksjtg6TKowYxMy1iqhf:btx6nkc) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L118)):*

nit: `COUNT DISTINCT`?
</blockquote></details>

There's only one entry, or we're really screwed.

---

*[pkg/acceptance/mixed_version_test.go, line 125 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqWKt7Yn5abiWVEui:-KsjtorDzmmHE_ztFL10:b-wpgsl1) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/mixed_version_test.go#L125)):*

why not do that, then?
</blockquote></details>

Expanded comment (want to check the cluster works after restart).

---

*[pkg/acceptance/localcluster/cluster.go, line 168 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjqfl6AQq7yqq4hKCA:-KsjtrWYkrCVTezTKBee:b-yh0t7i) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L168)):*

nit: preallocate `c.Nodes = make([]*Node, c.cfg.NumNodes)` next to where you create the slice of channels? then you could `c.Nodes[i], ch[i] = ...` here.
</blockquote></details>

No, we can't have a trivial/nil `*Node` in `c.Nodes` because `makeNode` calls various things that looks at all existing nodes.

---

*[pkg/acceptance/localcluster/cluster.go, line 235 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjquoeQlil0FO_cFcL:-Ksju2taV4j5jGV1N6Mo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L235)):*

random newline
</blockquote></details>

Done.

---

*[pkg/acceptance/localcluster/cluster.go, line 240 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqyQ-fFk1O44RZ7Nr:-Ksju3PtuyaK8CImS8iA:bc4f8ga) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L240)):*

hm, is there a reason it's better to have a separate rpc context for each node?
</blockquote></details>

It's what you expect in an acceptance test. Besides, otherwise we'll have to pass it down from the cluster and so not much is won.

---

*[pkg/acceptance/localcluster/cluster.go, line 504 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjrOJNficqAHh0pqLi:-KsjuvphKtQeYj5qAihE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L504)):*

nit: i believe callers of this and StartAsync already have contexts to hand in, might be nice to plumb them while you're here.
</blockquote></details>

Done.

---

*[pkg/acceptance/localcluster/cluster.go, line 537 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjreoQjwo55i-2jGHU:-Ksjv3JSdAb3hfC03tw9:bqffmj1) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L537)):*
Am I looking a the wrong thing? I'm copying it on this line already.

> args := append([]string(nil), n.cfg.ExtraArgs[1:]...)

---

*[pkg/acceptance/localcluster/cluster.go, line 637 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjsJ5LE25VMbsClIzj:-KsjvVM1h7AMkIkkWXcE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L637)):*

log, or something? but actually, if gossipAddr is non-empty and this error is non-nil, i think something bad has happened.
</blockquote></details>

Done.

---

*[pkg/acceptance/localcluster/cluster.go, line 645 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjsUidxOaUBYF-kDk7:-KsjvXTwXLCkdjpTEd8n:b-5m4wf0) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L645)):*

what's going on with the locking, here? why not defer the Lock?
</blockquote></details>

I agree that this pattern is unusual, but defer'ing a `Lock()` just looks like someone messed something up. I should just break up the giant `defer n.Unlock()` at the very top.

---

*[pkg/server/server.go, line 742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjow5uWC3tFhofHSm6:-KsjvuHfoalqDwCse5fb:bnrtbu1) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L742)):*

should this be a map? seems like uniqueness here is a thing
</blockquote></details>

I don't want to deal with reorderings.

---

*[pkg/server/server.go, line 743 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjpGokY3BrM37anYhA:-KsjvwoqH47fLuZqm5_O:b-ib1bbu) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L743)):*

these addresses are all unresolved, which seems like a bit of a sharp edge, which at the very least deserves some tests that demonstrate the behaviour.
</blockquote></details>

Which sharp edge? They should be exactly what we print to stdout in the node message below. Agreed that a test wouldn't hurt.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007)*

tamird:
Reviewed 4 of 4 files at r5.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/acceptance/localcluster/cluster.go, line 168 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjqfl6AQq7yqq4hKCA:-Ksjx07ACL97ys4cCzGa:btdsfzr) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L168)):*

No, we can't have a trivial/nil `*Node` in `c.Nodes` because `makeNode` calls various things that looks at all existing nodes.
</blockquote></details>

I see.

---

*[pkg/acceptance/localcluster/cluster.go, line 240 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjqyQ-fFk1O44RZ7Nr:-Ksjx40yhHEflbAmMWoI:bsiai3k) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L240)):*

It's what you expect in an acceptance test. Besides, otherwise we'll have to pass it down from the cluster and so not much is won.
</blockquote></details>

Heh, okay.

---

*[pkg/acceptance/localcluster/cluster.go, line 537 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjreoQjwo55i-2jGHU:-Ksjx8yUjnUM_uWiKzed:b-85imwh) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L537)):*

Am I looking a the wrong thing? I'm copying it on this line already.

> args := append([]string(nil), n.cfg.ExtraArgs[1:]...)
</blockquote></details>

You're looking at the right thing - I'm just saying that copying is not strictly necessary. It's a nit, so ignore it if you don't eat more CI.

---

*[pkg/acceptance/localcluster/cluster.go, line 645 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjsUidxOaUBYF-kDk7:-KsjxcT2Ublf0FT4tqFn:b2nhktp) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L645)):*

I agree that this pattern is unusual, but defer'ing a `Lock()` just looks like someone messed something up. I should just break up the giant `defer n.Unlock()` at the very top.
</blockquote></details>

OK.

---

*[pkg/server/server.go, line 742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjow5uWC3tFhofHSm6:-KsjxgEo8N4ubNEDR-bf:b9ye4s5) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L742)):*

I don't want to deal with reorderings.
</blockquote></details>

Wah? Order matters here?

---

*[pkg/server/server.go, line 743 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjpGokY3BrM37anYhA:-KsjxjeCGgUqYV5wNpzt:b-513os1) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L743)):*

Which sharp edge? They should be exactly what we print to stdout in the node message below. Agreed that a test wouldn't hurt.
</blockquote></details>

Are you sure? The unresolved address might have a port of zero, for instance, which I believe is not what goes to stdout. (I can't see the entire file, so if you're certain, OK.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007)*

tschottdorf:
Still pending a test for the first commit.

---

Review status: 7 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/acceptance/localcluster/cluster.go, line 537 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjreoQjwo55i-2jGHU:-KsjzlTaGsXijFJxSZGS:b-ppyemi) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L537)):*

You're looking at the right thing - I'm just saying that copying is not strictly necessary. It's a nit, so ignore it if you don't eat more CI.
</blockquote></details>

Ah, I see. I'll keep this simple. The copy does not matter.

---

*[pkg/acceptance/localcluster/cluster.go, line 645 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjsUidxOaUBYF-kDk7:-KskBtuxQhlrHm83_DEI:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8b3dc8c2971ed688f409b86e4f035a46caed8569/pkg/acceptance/localcluster/cluster.go#L645)):*

OK.
</blockquote></details>

Done.

---

*[pkg/server/server.go, line 742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksjow5uWC3tFhofHSm6:-Ksjyeo1UGaTNPb2b6pS:b-y308me) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L742)):*

Wah? Order matters here?
</blockquote></details>

It shouldn't, but I still prefer things happening in a fixed order. If we want to avoid duplicates, a map doesn't help - `./foo` is equal to `foo` etc. I think it's a non-concern (admittedly like my ordering concerns).

---

*[pkg/server/server.go, line 743 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsjpGokY3BrM37anYhA:-Ksjyx1nrsClXcxgzcZ5:bhaxzhl) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e0458b0c21eed44075f823a89920fa1a17b034a/pkg/server/server.go#L743)):*

Are you sure? The unresolved address might have a port of zero, for instance, which I believe is not what goes to stdout. (I can't see the entire file, so if you're certain, OK.)
</blockquote></details>

Relatively sure (also it wouldn't work otherwise):

```
	unresolvedListenAddr, err := officialAddr(ctx, s.cfg.Addr, ln.Addr(), os.Hostname)
	if err != nil {
		return err
	}
```

We'll know for sure when I add the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-:-KskC_L0uPSlK3Wi8Wx4:bkhm3so)*

tamird:
:lgtm:

---

Reviewed 3 of 3 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-:-KslQ6WOrkBXJrxI1Ild:bnfp4nl)*

tamird:
Yikes! Smells like a bug that deserves investigation. LGTM.

---

Reviewed 11 of 11 files at r7, 7 of 7 files at r8, 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-:-Ksnf2HJSMGkVDldz5PI:b1mw1lz)*

tschottdorf:
Absolutely, I filed https://github.com/cockroachdb/cockroach/issues/18027.

bdarnell:
:lgtm:

The first commit is fine with me, but it could use some commentary about the difference between these address files (always written early) and `--listening-url-file` (optionally written to signal server readiness)

---

Reviewed 4 of 11 files at r7, 6 of 7 files at r8, 1 of 1 files at r9.

---

*[pkg/acceptance/mixed_version_test.go, line 44 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksnv9NNO8Zb6DFJNa35:-Ksnv9NNO8Zb6DFJNa36:bq778be) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0d4ca9d2c5249910773e3f381f97ba99b68fab5/pkg/acceptance/mixed_version_test.go#L44)):*
> ```Go
> 	// I'm not 100% sure why, but the node started first may sometimes not be
> 	// able to connect to Gossip. That node is special in that it gets started
> 	// without join flags, but even though it is passed as a peer to all of the
> ```

What if you use the init command instead of omitting the join flag?

---

*[pkg/server/server.go, line 525 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsnwIviPh9o1dK92IiP:-KsnwIviPh9o1dK92IiQ:bj092s4) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0d4ca9d2c5249910773e3f381f97ba99b68fab5/pkg/server/server.go#L525)):*
> ```Go
> func (li listenerInfo) Iter() map[string]string {
> 	return map[string]string{
> 		".advertise-addr": li.advertise,
> ```

Prefix these names with "cockroach" instead of hiding them (If it's important enough to be in a dotfile, it's important enough not to be in a dotfile)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-:-Ksnx1vD-QHPdF9jhV5T:bdctf0r)*

tschottdorf:
TFTRs! Added commentary.

---

Review status: 2 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/acceptance/mixed_version_test.go, line 44 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-Ksnv9NNO8Zb6DFJNa35:-KsoDePPqFNatMdmdx9U:b-2g62s9) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0d4ca9d2c5249910773e3f381f97ba99b68fab5/pkg/acceptance/mixed_version_test.go#L44)):*

What if you use the init command instead of omitting the join flag?
</blockquote></details>

The problem occurs after restarting the whole cluster, so I don't think changing the init mode can help. If I'm missing something, let's discuss further in #18027.

---

*[pkg/server/server.go, line 525 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-KsnwIviPh9o1dK92IiP:-KsoDpWep8Sj4vIExFwy:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0d4ca9d2c5249910773e3f381f97ba99b68fab5/pkg/server/server.go#L525)):*

Prefix these names with "cockroach" instead of hiding them (If it's important enough to be in a dotfile, it's important enough not to be in a dotfile)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-:-KsoDqHyuLc-gLBd0d32:bidg4eg)*

tschottdorf:
Some more stuff to review, unfortunately.

---

Review status: 2 of 11 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007#-:-KsokxXoUkHOHFnCM9Tt:b-z9j8nn)*

tamird:
Reviewed 9 of 9 files at r10, 7 of 7 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 1 of 1 files at r14.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18007)*

PR #18019, opened 2017-08-30
sql: improve comments in TestPushedTxnDetectionInFirstBatch

andreimatei:
The comments around the test are inaccurate by suggesting that the
FirstBatch state directly has a role in whether serializability pushes
are detected or not. This was never quite true; it's whether or not
we're in an auto-retriable execution, which is not limited to state
FirstBatch (we can also be in Open and be under automatic retries).

### DIFF
--- a/pkg/sql/txn_restart_test.go
+++ b/pkg/sql/txn_restart_test.go
@@ -1658,11 +1658,12 @@ func TestRollbackToSavepointFromUnusualStates(t *testing.T) {
 }
 
 // Test that a serializable txn that has pushed its timestamp is detected and
-// retried automatically while the SQL txn is in state FirstBatch. It also tests
-// that, once the SQL txn has moved out of FirstBatch, the pushed timestamp is
-// no longer detected - the txn is left alone to run until completion and lay
-// down its intents, and only finds out that it can't commit at COMMIT time.
-func TestPushedTxnDetectionInFirstBatch(t *testing.T) {
+// retried automatically while the SQL txn can still be retried automatically.
+// It also tests that, once the SQL txn can no longer be retried automatically,
+// the pushed timestamp is no longer detected - the txn is left alone to run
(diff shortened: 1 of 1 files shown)


knz (approved):

nvanbenschoten:
:lgtm:

---

---

*[pkg/sql/txn_restart_test.go, line 1660 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18019#-KsnlZK_nEItP8StY_XB:-KsnlZKauZC5CXjXALfw:b-cx65xe) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0b0591869e2c62b2e89bb20c8073ea85b468ca0/pkg/sql/txn_restart_test.go#L1660)):*
> ```Go
> }
> 
> // Test that a serializable txn that has pushed its timestamp is detected and
> ```

`s/Test/TestPushedTxnDetection tests/`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18019#-:-KsnmGSHwWMLNmrkdxhS:bnfp4nl)*

andreimatei:

---

*[pkg/sql/txn_restart_test.go, line 1660 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18019#-KsnlZK_nEItP8StY_XB:-KsoBTb1lMX0SPbA1t1M:b-3jno8y) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0b0591869e2c62b2e89bb20c8073ea85b468ca0/pkg/sql/txn_restart_test.go#L1660)):*

`s/Test/TestPushedTxnDetection tests/`
</blockquote></details>

neah, I don't think that's necessary for tests. The test method names are frequently contrived and synthetic, so repeating the name takes space for no reason.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18019)*

PR #18020, opened 2017-08-30
sql: scope ResultsGroup's as SQL transactions

andreimatei:
The ResultsGroup interface is meant to represent a transaction's
results. Its main affordance, besides allowing statement results to be
streamed, is letting the client Reset() it, discarding all accumulated
results. This is meant to support automatic txn retries, when previous
results need to be dropped.
Instances of ResultsGroup were supposed to correspond to transaction but
that's not what the code was doing. A ResultsGroup was created for each
transaction, but one was also created at the beginning of each batch
regardless of whether we were already inside a txn or not.

This patch properly ties ResultsGroups to txns, like the interface
intended (otherwise it's hard to say what one of these "groups" is). The
group can now be stored in the txnState, where it belongs, and
initialized/closed in resetStateForNewSQLTxni()/finishSQLTxn(), like its
other txn-scoped brethren. This also means that ResultsGroups are now
always closed - before you could have dangling ones which is a violation
of the interface contract.

This proper scoping triggered the need for an extension to the
interface: the point of the interface is letting you discard past
results, but not all results should be discarded. Namely, results of
special statements that leave a txn in the FirstBatch state (BEGIN,
SAVEPOINT, etc) should not be discarded, because they will not be
retried. More exactly, before this patch, whether or not these
statements were automatically retried depended on whether they happened
to be part of the same batch as the later statement that caught a
retriable error - if so, they were retried, if not they weren't. When
they were retried, their previous results were _not_ Reset(), and
vice-versa. This is where the fact that we created new groups for new
batches was useful: it helped with not resetting results from a previous
batch.  The new solution to this problem is a Flush() method, which
provides a reference point for future Reset()s - they'll only discard up
to the flush point, going backwards.

The introduction of Flush() in turn triggered a modest restructuring of
Executor internals; the problem is that the natural time to Flush() is
when transitioning from the FirstBatch state to Open. But, alas, before
this patch, automatic retries were scoped like invocations of
execStmtsInCurrentTxn(), which dealt with all the statements in a
transaction (or at least a prefix of a transaction, if a statements
batch finished in the middle of a txn). The FirstBatch->Open transition
happened within a call to execStmtsInCurrentTxn() and so, if we were to
Flush() at this time and then retry, statements that we've flushed would
be re-executed and produce duplicate resuls => no good.
This patch solves the issue by having changing the semantics of
FirstBatch and the moment when the state transition happens. FirstBatch
never had a proper definition; it has now been renamed to AutoRetry and
gets clear semantics - this state is like Open, plus, if a retriable
error happens while we're in this state, we can auto-retry the
transaction (more exactly, we can retry the transaction by retrying the
part of the current batch that is part of the current transaction).
The txnState.autoRetry field is gone - a big win in itself. The state
alone now dictates whether we can auto-retry or not.
Transactions always start in AutoRetry and move to Open at the end of a
batch.

### DIFF
--- a/pkg/cli/sql.go
+++ b/pkg/cli/sql.go
@@ -505,7 +505,7 @@ func (c *cliState) refreshTransactionStatus() {
 	case sql.RestartWait.String():
 		c.lastKnownTxnStatus = " RETRY"
 	case sql.Open.String():
-		// The state FirstBatch is reported by the server as Open, so no need to
+		// The state AutoRetry is reported by the server as Open, so no need to
 		// handle it here.
 		c.lastKnownTxnStatus = "  OPEN"
 	}
--- a/pkg/sql/distsql_running.go
+++ b/pkg/sql/distsql_running.go
@@ -214,7 +214,7 @@ func (dsp *distSQLPlanner) Run(
 	return nil
 }
 
-// distSQLReceiver is a RowReceiver that sends results to a StatementResultWriter.
+// distSQLReceiver is a RowReceiver that writes results to a rowResultWriter.
 // This is where the DistSQL execution meets the SQL Session - the RowContainer
 // comes from a client Session.
 //
@@ -260,8 +260,8 @@ type distSQLReceiver struct {
 	updateClock func(observedTs hlc.Timestamp)
 }
 
-// rowResultWriter is a subset of StatementResultWriter for use in the
-// distSQLReceiver -- also implemented bufferedRowWriter as a one off.
+// rowResultWriter is a subset of StatementResult to be used with the
+// distSQLReceiver. It's implemented by RowResultWriter.
 type rowResultWriter interface {
 	// AddRow takes the passed in row and adds it to the current result.
 	AddRow(ctx context.Context, row parser.Datums) error
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -299,7 +299,7 @@ func (*ExecutorTestingKnobs) ModuleTestingKnobs() {}
 
 // StatementFilter is the type of callback that
 // ExecutorTestingKnobs.StatementFilter takes.
-type StatementFilter func(context.Context, string, ResultWriter, error) error
+type StatementFilter func(context.Context, string, ResultsWriter, error) error
 
 // ExecutorTestingKnobs is part of the context used to control parts of the
 // system during testing.
@@ -330,7 +330,7 @@ type ExecutorTestingKnobs struct {
 	// AfterExecute is like StatementFilter, but it runs in the same goroutine of the
 	// statement.
 	AfterExecute func(
-		ctx context.Context, stmt string, statementResultWriter StatementResultWriter, err error,
+		ctx context.Context, stmt string, res StatementResult, err error,
 	)
 
 	// DisableAutoCommit, if set, disables the auto-commit functionality of some
@@ -573,7 +573,7 @@ func (e *Executor) ExecuteStatementsBuffered(
 	session *Session, stmts string, pinfo *parser.PlaceholderInfo, expectedNumResults int,
 ) (StatementResults, error) {
 	b := newBufferedWriter(session.makeBoundAccount())
-	session.ResultWriter = b
+	session.ResultsWriter = b
 	err := e.ExecuteStatements(session, stmts, pinfo)
 	res := b.results()
 	if err != nil {
@@ -595,7 +595,7 @@ func (e *Executor) ExecuteStatementsBuffered(
 	return res, nil
 }
 
-// ExecuteStatements executes the given statement(s) and returns a response.
+// ExecuteStatements executes the given statement(s).
 func (e *Executor) ExecuteStatements(
 	session *Session, stmts string, pinfo *parser.PlaceholderInfo,
 ) error {
@@ -753,7 +753,7 @@ func (e *Executor) execParsed(
 ) error {
 	var avoidCachedDescriptors bool
 	txnState := &session.TxnState
-	resultWriter := session.ResultWriter
+	resultWriter := session.ResultsWriter
 
 	if len(stmts) == 0 {
 		resultWriter.SetEmptyQuery()
@@ -764,7 +764,10 @@ func (e *Executor) execParsed(
 		// Each iteration consumes a transaction's worth of statements. Any error
 		// that is encountered resets stmts.
 
-		groupResultWriter := resultWriter.NewGroupResultWriter()
+		// TODO(andrei): NewResultsGroup() should only be called if inTxn == false.
+		// Also, we're currently not always closing this txnResults, which is
+		// against the contract.
+		txnResults := resultWriter.NewResultsGroup()
 		inTxn := txnState.State() != NoTxn
 		execOpt := client.TxnExecOptions{}
 		// Figure out the statements out of which we're going to try to consume
@@ -834,11 +837,11 @@ func (e *Executor) execParsed(
 			}
 
 			// Some results may have been produced by a previous attempt.
-			groupResultWriter.Reset(session.Ctx())
+			txnResults.Reset(session.Ctx())
 			var err error
 			remainingStmts, err = runTxnAttempt(
 				e, session, stmtsToExec, pinfo, origState, opt,
-				!inTxn /* txnPrefix */, avoidCachedDescriptors, automaticRetryCount, groupResultWriter)
+				!inTxn /* txnPrefix */, avoidCachedDescriptors, automaticRetryCount, txnResults)
 
 			// TODO(andrei): Until #7881 fixed.
 			if err == nil && txnState.State() == Aborted {
@@ -919,7 +922,7 @@ func (e *Executor) execParsed(
 		// If we're no longer in a transaction, finish the trace.
 		if txnState.State() == NoTxn {
 			txnState.finishSQLTxn(session)
-			groupResultWriter.End()
+			txnResults.Close()
 		}
 
 		// Verify that the metadata callback fails, if one was set. This is
@@ -989,6 +992,7 @@ func (e *Executor) execParsed(
 //
 // txnPrefix: set if the statements represent the first batch of statements in a
 // 	txn. Used to trap nested BEGINs.
+// txnResults: used to pass query results to the caller.
 func runTxnAttempt(
 	e *Executor,
 	session *Session,
@@ -999,7 +1003,7 @@ func runTxnAttempt(
 	txnPrefix bool,
 	avoidCachedDescriptors bool,
 	automaticRetryCount int,
-	groupResultWriter GroupResultWriter,
+	txnResults ResultsGroup,
 ) (StatementList, error) {
 
 	// Ignore the state that might have been set by a previous try of this
@@ -1014,7 +1018,7 @@ func runTxnAttempt(
 
 	remainingStmts, err := e.execStmtsInCurrentTxn(
 		session, stmts, pinfo,
-		txnPrefix, avoidCachedDescriptors, automaticRetryCount, groupResultWriter)
+		txnPrefix, avoidCachedDescriptors, automaticRetryCount, txnResults)
 
 	if opt.AutoCommit && len(remainingStmts) > 0 {
 		panic("implicit txn failed to execute all stmts")
@@ -1047,6 +1051,7 @@ func runTxnAttempt(
 // avoidCachedDescriptors: set if the statement execution should avoid
 //  using cached descriptors.
 // automaticRetryCount: increases with each retry; 0 for the first attempt.
+// txnResults: used to pass query results to the caller.
 //
 // Returns:
 //  - the statements that haven't been executed because the transaction has
@@ -1063,7 +1068,7 @@ func (e *Executor) execStmtsInCurrentTxn(
 	txnPrefix bool,
 	avoidCachedDescriptors bool,
 	automaticRetryCount int,
-	groupResultWriter GroupResultWriter,
+	txnResults ResultsGroup,
 ) (StatementList, error) {
 	txnState := &session.TxnState
 	if txnState.State() == NoTxn {
@@ -1122,23 +1127,23 @@ func (e *Executor) execStmtsInCurrentTxn(
 			stmtStrBefore = stmt.String()
 		}
 
-		statementResultWriter := groupResultWriter.NewStatementResultWriter()
+		stmtResult := txnResults.NewStatementResult()
 		var err error
 		// Run SHOW TRANSACTION STATUS in a separate code path so it is
 		// always guaranteed to execute regardless of the current transaction state.
 		if _, ok := stmt.AST.(*parser.ShowTransactionStatus); ok {
-			err = runShowTransactionState(session, statementResultWriter)
+			err = runShowTransactionState(session, stmtResult)
 		} else {
 			switch txnState.State() {
 			case Open, FirstBatch:
 				err = e.execStmtInOpenTxn(
 					session, stmt, pinfo, txnPrefix && (i == 0), /* firstInTxn */
-					avoidCachedDescriptors, automaticRetryCount, statementResultWriter,
-					groupResultWriter.ResultsSentToClient)
+					avoidCachedDescriptors, automaticRetryCount, stmtResult,
+					txnResults.ResultsSentToClient)
 			case Aborted, RestartWait:
-				err = e.execStmtInAbortedTxn(session, stmt, groupResultWriter)
+				err = e.execStmtInAbortedTxn(session, stmt, txnResults)
 			case CommitWait:
-				err = e.execStmtInCommitWaitTxn(session, stmt, statementResultWriter)
+				err = e.execStmtInCommitWaitTxn(session, stmt, stmtResult)
 			default:
 				panic(fmt.Sprintf("unexpected txn state: %s", txnState.State()))
 			}
@@ -1151,7 +1156,7 @@ func (e *Executor) execStmtsInCurrentTxn(
 			}
 		}
 		if filter := e.cfg.TestingKnobs.StatementFilter; filter != nil {
-			if err := filter(session.Ctx(), stmt.String(), session.ResultWriter, err); err != nil {
+			if err := filter(session.Ctx(), stmt.String(), session.ResultsWriter, err); err != nil {
 				return nil, err
 			}
 		}
@@ -1191,15 +1196,15 @@ func getTransactionState(txnState *txnState) string {
 }
 
 // runShowTransactionState returns the state of current transaction.
-func runShowTransactionState(session *Session, statementResultWriter StatementResultWriter) error {
-	statementResultWriter.BeginResult((*parser.ShowTransactionStatus)(nil))
-	statementResultWriter.SetColumns(sqlbase.ResultColumns{{Name: "TRANSACTION STATUS", Typ: parser.TypeString}})
+func runShowTransactionState(session *Session, res StatementResult) error {
+	res.BeginResult((*parser.ShowTransactionStatus)(nil))
+	res.SetColumns(sqlbase.ResultColumns{{Name: "TRANSACTION STATUS", Typ: parser.TypeString}})
 
 	state := getTransactionState(&session.TxnState)
-	if err := statementResultWriter.AddRow(session.Ctx(), parser.Datums{parser.NewDString(state)}); err != nil {
+	if err := res.AddRow(session.Ctx(), parser.Datums{parser.NewDString(state)}); err != nil {
 		return err
 	}
-	return statementResultWriter.EndResult()
+	return res.CloseResult()
 }
 
 // execStmtInAbortedTxn executes a statement in a txn that's in state
@@ -1208,9 +1213,9 @@ func runShowTransactionState(session *Session, statementResultWriter StatementRe
 // - ROLLBACK TO SAVEPOINT / SAVEPOINT: reopens the current transaction,
 //   allowing it to be retried.
 func (e *Executor) execStmtInAbortedTxn(
-	session *Session, stmt Statement, groupResultWriter GroupResultWriter,
+	session *Session, stmt Statement, txnResults ResultsGroup,
 ) error {
-	statementResultWriter := groupResultWriter.NewStatementResultWriter()
+	res := txnResults.NewStatementResult()
 	txnState := &session.TxnState
 	if txnState.State() != Aborted && txnState.State() != RestartWait {
 		panic("execStmtInAbortedTxn called outside of an aborted txn")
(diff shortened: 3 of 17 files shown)


andreimatei:
You can ignore the first commit; it's part of a larger effort to document and improve the result streaming interfaces that's still in progress. It's only here because it contains some interface renames used in the following commits; I was too lazy to cherry-pick those without the renames but I'll do it later.

knz:
This first commit is actually pretty good. Thanks for doing it.
The others are also pretty good. Only minor nits.

This looks good to me! Although perhaps Nathan should indeed have another look, so we get more confidence.

---

Reviewed 12 of 12 files at r1, 1 of 1 files at r2, 5 of 5 files at r3.

---

*[pkg/sql/executor.go, line 1114 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnEYgHjrdOPivuETMi:-KsnEYgIhUdcQxVPDp2e:b-2u6r85) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L1114)):*
> ```Go
> 			// Note that the remaining statements include the current one (i.e.
> 			// index i - stmt), since we stopped short of executing it.
> 			return stmts[i:], nil
> ```

Yeah that's actually much more elegant. :100:

---

*[pkg/sql/results_writer.go, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnBc6CSCHvTpicS6JY:-KsnBc6CSCHvTpicS6JZ:bf505io) ([raw file](https://github.com/cockroachdb/cockroach/blob/343c7bb0e507e04944c280a0f41256fe541308b4/pkg/sql/results_writer.go#L40)):*
> ```Go
> }
> 
> // ResultsWriter the interface used to by the Executor to produce results for
> ```

+ "is"

---

*[pkg/sql/results_writer.go, line 41 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnBftH-kySut_MSSPY:-KsnBftH-kySut_MSSPZ:b-j0dsl9) ([raw file](https://github.com/cockroachdb/cockroach/blob/343c7bb0e507e04944c280a0f41256fe541308b4/pkg/sql/results_writer.go#L41)):*
> ```Go
> 
> // ResultsWriter the interface used to by the Executor to produce results for
> // query execution for a SQL client. The main implementater is v3Conn, which
> ```

s/implementater/implementer

---

*[pkg/sql/session.go, line 809 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnD-B2HkOhdcg4W8ql:-KsnD-B2HkOhdcg4W8qm:b-hq5hll) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L809)):*
> ```Go
> 	// automatically retry the 2nd batch even through the statements in the first
> 	// batch will not be executed again and their results have already been sent
> 	// to the clients. We can do this because some statements are special in that
> ```

to the consumer (i.e. SQL client)

---

*[pkg/sql/session.go, line 813 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnD4hYMZQvFZQjG9Io:-KsnD4hYMZQvFZQjG9Ip:bi8mwly) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L813)):*
> ```Go
> 	// beginning of this 2nd batch that the FirstBatch state is useful: automatic
> 	// retries are enabled or disabled for the continuation of a previous
> 	// transaction based on we're in FirstBatch or not.
> ```

.. based on whether we're ...

---

*[pkg/sql/session.go, line 975 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnDLJpwsKoktJWDku2:-KsnDLJpwsKoktJWDku3:bywv55j) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/session.go#L975)):*
> ```Go
> 	ts.sqlTimestamp = sqlTimestamp
> 	ts.implicitTxn = implicitTxn
> 	ts.txnResults = s.ResultsWriter.NewResultsGroup()
> ```

Maybe here assert there's no ongoing result group that's not closed yet.

---

*[pkg/sql/pgwire/v3.go, line 1160 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnE31ZV8egrBhhUxH0:-KsnE31ZV8egrBhhUxH1:bj1gvw0) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/pgwire/v3.go#L1160)):*
> ```Go
> 
> // Flush implements the ResultsGroup interface.
> func (c *v3Conn) Flush(ctx context.Context) {
> ```

Ok this implements the interface, but maybe it's worth also actually flushing the buffer towards the network connection and resetting the buffer altogether (and set s.txnStartIdx to 0).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-:-KsnCp9W1G1N5q2y0L0j:b5vrz3s)*

nvanbenschoten:
Reviewed 1 of 1 files at r2, 1 of 5 files at r3.

---

*[pkg/sql/executor.go, line 754 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoNzKeL3oepR27l2yR:-KsoNzKeL3oepR27l2yS:bjtc44r) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L754)):*
> ```Go
> 	// or not it can be automatically retried depends on the state that the
> 	// previous batch has left the txn in: if it left it in FirstBatch, then we
> 	// can automatically retry it (by definition; that's what the FirstBatch state
> ```

Is `FirstBatch` the wrong name for this state then? What about `AutoRetryable`?

In fact, why do we even need to maintain `txnState.autoRetry` separately from `txnState.State()`? Can it just be a local in this function?

---

*[pkg/sql/executor.go, line 809 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoZF3eyZkI_re606j_:-KsoZF3eyZkI_re606ja:bx87k2z) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L809)):*
> if we're in state FirstBatch, we don't want to auto-retry

We don't want to retry, or we don't expect to ever need to?

---

*[pkg/sql/executor.go, line 814 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoOhY6MCUG46lLa9zH:-KsoOhY6MCUG46lLa9zI:bmtfqjz) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L814)):*
> ```Go
> 		// that point the next group will run with auto-retries. If we're in every
> 		// other state then Open, then again retriable errors don't apply.
> 		execOpt.AutoRetry = txnState.state == Open && txnState.autoRetry
> ```

`txnState.State() == Open`?

---

*[pkg/sql/session.go, line 807 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoMximbuvJnaEju061:-KsoMximbuvJnaEju062:b-2o2wlv) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L807)):*
> ```Go
> 	// example, is we get a batch with "BEGIN; SET TRANSACTION ISOLATION LEVEL
> 	// foo; SAVEPOINT cockroach_restart;" followed by a 2nd batch, we can
> 	// automatically retry the 2nd batch even through the statements in the first
> ```

`s/through/though/`

---

*[pkg/sql/session.go, line 810 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoN4krsGJElLBegYyM:-KsoN4krsGJElLBegYyN:b-i0ej0j) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L810)):*
> ```Go
> 	// batch will not be executed again and their results have already been sent
> 	// to the clients. We can do this because some statements are special in that
> 	// all their executions generate exactly the same results. It is at the
> ```

`s/all their executions generate/their execution always generates/`

---

*[pkg/sql/session.go, line 815 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoNK9Z6eOFLhYKU0dx:-KsoNK9Z6eOFLhYKU0dy:bcv6kd4) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L815)):*
> ```Go
> 	// transaction based on we're in FirstBatch or not.
> 	//
> 	// A BEGIN statement makes the transaction from NoTxn to FirstBatch. As soon
> ```

`s/transaction from/transaction transition from/`

---

*[pkg/sql/logictest/testdata/logic_test/show_trace, line 16 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoZsv2D2Ku5OO5auYq:-KsoZsv2D2Ku5OO5auYr:b-mhhqg3) ([raw file](https://github.com/cockroachdb/cockroach/blob/6151997b9ca391811144e89e5544ede2a13ddac5/pkg/sql/logictest/testdata/logic_test/show_trace#L16)):*
> ```
> (1,0)  === SPAN START: sql txn ===           sql txn implicit
> (1,0)  executing 1/5: BEGIN TRANSACTION      sql txn implicit
> (1,0)  executing 2/5: SELECT 1               sql txn implicit
> ```

I get why this happened, but do we want to try to retain the old behavior in the eyes of traces?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020)*

andreimatei:
The commit changed some (for the better :) ), PTAL.

---

Review status: 5 of 17 files reviewed at latest revision, 14 unresolved discussions.

---

*[pkg/sql/executor.go, line 754 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoNzKeL3oepR27l2yR:-KsobpfuWWXTAamV4gby:bxr9db5) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L754)):*

Is `FirstBatch` the wrong name for this state then? What about `AutoRetryable`?

In fact, why do we even need to maintain `txnState.autoRetry` separately from `txnState.State()`? Can it just be a local in this function?
</blockquote></details>

no... OK yeah... *(1 day later)* no, screw you Nathan *(1 day later)* done; I'm glad you brought it up.

---

*[pkg/sql/executor.go, line 814 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoOhY6MCUG46lLa9zH:-KsvEB5IrHJfEwj3hK2w:bvwq4dp) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L814)):*

`txnState.State() == Open`?
</blockquote></details>

I believe the main goroutine can access that without the atomic. But done, and also in other places.

---

*[pkg/sql/results_writer.go, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnBc6CSCHvTpicS6JY:-KsoKbeYJCqkUqtVsqE_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/343c7bb0e507e04944c280a0f41256fe541308b4/pkg/sql/results_writer.go#L40)):*

+ "is"
</blockquote></details>

Done.

---

*[pkg/sql/results_writer.go, line 41 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnBftH-kySut_MSSPY:-KsoKc0ZOvzx8BpCXl7m:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/343c7bb0e507e04944c280a0f41256fe541308b4/pkg/sql/results_writer.go#L41)):*

s/implementater/implementer
</blockquote></details>

Done.

---

*[pkg/sql/session.go, line 807 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoMximbuvJnaEju061:-KsvET2Z6MAp7xMBIhCd:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L807)):*

`s/through/though/`
</blockquote></details>

Done.

---

*[pkg/sql/session.go, line 809 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnD-B2HkOhdcg4W8ql:-KsoL1sSd2tBENQlDoG7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L809)):*

to the consumer (i.e. SQL client)
</blockquote></details>

Done.

---

*[pkg/sql/session.go, line 813 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnD4hYMZQvFZQjG9Io:-KsoL50A3PI8DKd76bcz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L813)):*

.. based on whether we're ...
</blockquote></details>

Done.

---

*[pkg/sql/session.go, line 975 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnDLJpwsKoktJWDku2:-KsoLx4PJGtHhUs479ca:b-8y75h9) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/session.go#L975)):*

Maybe here assert there's no ongoing result group that's not closed yet.
</blockquote></details>

ok, added this condition to the assertion at the top of this method.
Well, technically, here we assert that the `txnState.txnResults` is nil. This doesn't necessarily ensure that there's no open group on the `ResultsWriter` - there's no way to assert that through the interface. Implementers should individually assert this. `v3Conn` currently doesn't; I want to concentrate on the `v3Conn` implementation separately, so stay tuned.

---

*[pkg/sql/pgwire/v3.go, line 1160 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsnE31ZV8egrBhhUxH0:-KsoaTDxJmE3vLs_8uDI:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/pgwire/v3.go#L1160)):*

Ok this implements the interface, but maybe it's worth also actually flushing the buffer towards the network connection and resetting the buffer altogether (and set s.txnStartIdx to 0).
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/show_trace, line 16 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoZsv2D2Ku5OO5auYq:-KsvEZSNsYCEu-BS8eGu:bmq04b5) ([raw file](https://github.com/cockroachdb/cockroach/blob/6151997b9ca391811144e89e5544ede2a13ddac5/pkg/sql/logictest/testdata/logic_test/show_trace#L16)):*

I get why this happened, but do we want to try to retain the old behavior in the eyes of traces?
</blockquote></details>

This now reverted back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-:-KsvEjqz23JBZL5d7lYz:b-vylxve)*

knz:
I'm still good on this.

---

Reviewed 2 of 12 files at r1, 1 of 6 files at r4, 12 of 12 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-:-KsxU0p82Exlsd6SeB4n:b-lqjxdo)*

nvanbenschoten:
:lgtm: Though I'm still curious about the guarantee not to rerun `canStayInAutoRetryState` statements during a retry. Did you remove that guarantee but not update the comment? I know we had gone back and forth about this.

---

Reviewed 3 of 12 files at r5.

---

*[pkg/sql/executor.go, line 754 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoNzKeL3oepR27l2yR:-KsyfDRZtZ0uNHJaXjd9:bgc7vqv) ([raw file](https://github.com/cockroachdb/cockroach/blob/262e5e0570b44565e3e4994d957558df34ed6dc4/pkg/sql/executor.go#L754)):*

no... OK yeah... *(1 day later)* no, screw you Nathan *(1 day later)* done; I'm glad you brought it up.
</blockquote></details>

:heart: 

---

*[pkg/sql/executor.go, line 809 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsyimEqgPsRLsd5ZTMQ:-KsyimEqgPsRLsd5ZTMR:b-n7ds8j) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L809)):*
> ```Go
> 			)
> 		}
> 		execOpt.AutoRetry = txnState.State() == AutoRetry
> ```

I'm sending you a peer ack next week specifically for this line.

---

*[pkg/sql/executor.go, line 1084 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-Ksyfv9-5qZ0Y8WEKQos:-Ksyfv9-5qZ0Y8WEKQot:bdlcho1) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L1084)):*
> ```Go
> 	}
> 
> 	// If we're still in AutoRetry but we've seen statements that would need to be
> ```

Beautiful :100: 

---

*[pkg/sql/executor.go, line 1109 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-Ksyg3jOu4q1Tz2_3jhY:-Ksyg3jPnKHDxnLcq0-q:b-dbvxn6) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L1109)):*
> ```Go
> // Args:
> // session:    The session to execute the statement in.
> // stmt: 		   The statement to execute.
> ```

Weird spacing here I think.

---

*[pkg/sql/executor.go, line 2261 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-Ksygno9uj3YEgr4AShd:-Ksygno9uj3YEgr4AShe:b-339z7o) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L2261)):*
> ```Go
> 
> // canStayInAutoRetryState returns true if the statement, by itself, should not
> // cause the session to transition from AutoRetry>Open at the end of the
> ```

`s/AutoRetry>Open/AutoRetry->Open/`

---

*[pkg/sql/executor.go, line 2267 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsyhguTF83C7Wz1YZe2:-KsyhguTF83C7Wz1YZe3:b-5f1qta) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L2267)):*
> ```Go
> // AutoRetry->Open transition that ends of batches usually perform.
> //
> // Statements in this category may not be executed again in case of an automatic
> ```

How do you assure this? I must be missing something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-:-Ksyj9go5RhkAvIREGbk:bz3wfbg)*

andreimatei:
Thanks for the review!
Does my response below sound good?

---

---

*[pkg/sql/executor.go, line 2267 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsyhguTF83C7Wz1YZe2:-Ksyk5SVGdCUrg6hYJZq:b-h1q595) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L2267)):*

How do you assure this? I must be missing something.
</blockquote></details>

Well the comment now says "_may_ not be executed".
The behavior stays as it was before this PR - they'll be executed again if they happen to be part of the batch that caught the retryable error, and will not be re-executed if they were part of a previous batch.

The situation will improve once there's no more "batches" to speak of.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-:-KsykZztrNSZRIcrYpTk:bdne3yh)*

nvanbenschoten:

---

*[pkg/sql/executor.go, line 2267 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsyhguTF83C7Wz1YZe2:-KsymOQHeuLR3WqZpU_U:b-cnmpox) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L2267)):*

Well the comment now says "_may_ not be executed".
The behavior stays as it was before this PR - they'll be executed again if they happen to be part of the batch that caught the retryable error, and will not be re-executed if they were part of a previous batch.

The situation will improve once there's no more "batches" to speak of.

</blockquote></details>

I misinterpreted the "may" as a directive. This sounds good to me now. Thanks for clarifying.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020)*

andreimatei:
Review status: 3 of 17 files reviewed at latest revision, 10 unresolved discussions.

---

*[pkg/sql/executor.go, line 1109 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-Ksyg3jOu4q1Tz2_3jhY:-KsytOU3QF-z8uKMghYf:b-xmqp6z) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L1109)):*

Weird spacing here I think.
</blockquote></details>

I'm never sure what the spacing here looks good... Changed it some.

---

*[pkg/sql/executor.go, line 2261 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-Ksygno9uj3YEgr4AShd:-KsytZtOy6KlXzEJQGGb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/beb347717118686bf9d2d02a54f336ea29c254de/pkg/sql/executor.go#L2261)):*

`s/AutoRetry>Open/AutoRetry->Open/`
</blockquote></details>

Done.

---

*[pkg/sql/session.go, line 810 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoN4krsGJElLBegYyM:-Ksytiuj-uRZ5qL1722r:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L810)):*

`s/all their executions generate/their execution always generates/`
</blockquote></details>

Done.

---

*[pkg/sql/session.go, line 815 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020#-KsoNK9Z6eOFLhYKU0dx:-Ksytts1SbaD4yntARx7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4d3b05438d60263e8122cfdaaa9b8d9ac58a398/pkg/sql/session.go#L815)):*

`s/transaction from/transaction transition from/`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18020)*

Issue #17921, opened 2017-08-25
DELETE FROM ... [returning nothing] crashes a node


AnyCPU:
A cluster has 3 nodes running on Cockroachdb 1.0.5 (Linux, 64bit)
A db has about 195 millions of records.
Scheme is 
```
CREATE TABLE stats (
	numa STRING(32) NULL,
	bytes INT NULL,
	amount DECIMAL NULL,
	ts TIMESTAMP NULL,
	da INT NULL,
	INDEX idx_stats_ts (ts ASC) STORING (numa, bytes, amount, da),
	FAMILY "primary" (numa, bytes, amount, ts, da, rowid)
)
```
The db has data of an one month.
I want to delete data starting from second day of month, so
I run either  `delete from stats where ts > '2017-01-01 23:59:59'::timestamp;` or `delete from stats where ts > '2017-01-01 23:59:59'::timestamp returning nothing;`.
I use the `cockroach sql --insecure --host = ...` command to run my query.
After some minutes an used node dies.
A db driver returns - bad connection.
And there are a lot of messages like that `context canceled while in command queue: ResolveIntent...` in the log.

All nodes have free space.

Do I have to delete data by chunks?

Thanks.

AnyCPU:
I cannot even remove a half a day.

AnyCPU:
I have tried to remove data for an one hour but I got
```pq: kv/txn_coord_sender.go:428: transaction is too large to commit: 478692 intents```

AnyCPU:
It is possible to remove data by ten-minutes chunks :-(

tschottdorf:
Hey @AnyCPU, you're running into the fact that as of the time of writing, support for large writing transactions is not ideal in CockroachDB. You've pretty much experienced all the problems: them never succeeding, or failing with the "too large to commit" error. While deleting in small chunks is currently the best option, rest assured that we're working on improving this (see the issue above). We have workarounds in place for the case in which you want to drop/truncate the whole table (which we can do more efficiently by essentially swapping out the table with a new one), but that's clearly not going to help you.

Tracking issue: https://github.com/cockroachdb/cockroach/issues/15849

AnyCPU:
Thank you @tschottdorf

tschottdorf:
Despite many improvements, this is still an issue in `master` at the time of writing, though we now have a test that reproduces this reliably: https://github.com/cockroachdb/cockroach/pull/22876

tschottdorf:
[...]

PR #18023, opened 2017-08-30
sql: add support for LIMIT with DELETE

### LINKED ISSUE #18006: sql: support LIMIT for DELETEs
MySQL has a `LIMIT` clause for `DELETE` that allows one to incrementally delete rows from a table. Instead of deleting all rows at once, one can execute `DELETE ... LIMIT N` until there are no more rows to be deleted. This reduces the duration of the transaction containing the `DELETE`. (See the [MySQL DELETE docs](https://dev.mysql.com/doc/refman/5.7/en/delete.html) for more details.)

This is a workaround for large `DELETE`s until we have support for [support for larger transactions](https://github.com/cockroachdb/cockroach/issues/15849). This would improve the user experience described in #17921.

The one downside I can think of this approach is that [Postgres's DELETE](https://www.postgresql.org/docs/9.6/static/sql-delete.html) does not support the `LIMIT` clause, which means that ORMs' Postgres dialects won't generate queries that use `LIMIT`s.

knz:
Suggested by @cuongdo:

> MySQL has a LIMIT clause for DELETE that allows one to incrementally
> delete rows from a table. Instead of deleting all rows at once, one
> can execute DELETE ... LIMIT N until there are no more rows to be
> deleted. This reduces the duration of the transaction containing the
> DELETE.

Fixes #18006.
Alleviates #17921.


tschottdorf:
Very happy you put this together so quickly! Definitely has my blessing.

pkg/sql/logictest/testdata/logic_test/delete
```diff
@@ -173,3 +173,54 @@ CREATE TABLE indexed (id int primary key, value int, other int, index (value))
...
+0  ·       from   indexed
+1  limit   ·      ·
+2  render  ·      ·
+3  scan    ·      ·
+3  ·       table  indexed@primary
+3  ·       spans  ALL
+3  ·       limit  10
+
+query ITTT
+EXPLAIN DELETE FROM indexed WHERE value = 5 LIMIT 10 RETURNING id
```
justinj: Could you also put in a test of actually executing `DELETE...LIMIT`? If the order being undefined means you can't check the exact results you can check the `COUNT(*)`, right?

pkg/sql/parser/sql.y
```diff
@@ -1425,12 +1425,19 @@ create_stmt:
...
+//               [LIMIT <expr>]
+//               [RETURNING <exprs...>]
 // %SeeAlso: WEBDOCS/delete.html
 delete_stmt:
-  opt_with_clause DELETE FROM relation_expr_opt_alias where_clause returning_clause
+  opt_with_clause DELETE FROM relation_expr_opt_alias where_clause opt_limit_clause returning_clause
   {
-    $$.val = &Delete{Table: $4.tblExpr(), Where: newWhere(astWhere, $5.expr()), Returning: $6.retClause()}
+    $$.val = &Delete{
+               Table: $4.tblExpr(),
```
justinj: nit: I think the indentation here is a little wonky.

justinj:
Hah, what a pleasantly small change!

knz:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 215 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18023#-KsnKbrePCfihPE3jMQM-r1-215:-KsnTuA2KUfAQoD4apLi:b-8vskqr) ([raw file](https://github.com/cockroachdb/cockroach/blob/b221ad0524cdb33b97cc9a3b922f3ee1deeb1e76/pkg/sql/logictest/testdata/logic_test/delete#L215)):*

Could you also put in a test of actually executing `DELETE...LIMIT`? If the order being undefined means you can't check the exact results you can check the `COUNT(*)`, right?
</blockquote></details>

The result tag contains the number of rows affected, or, if RETURNING was used, the rows actually deleted. The client has plenty of information to work with. Unfortunately, we can't check the tag in logic tests, but I put a RETURNING test here to showcase. Also indeed, the rows are non-deterministic so we need count(*).

---

*[pkg/sql/parser/sql.y, line 1436 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18023#-KsnKbrePCfihPE3jMQR-r1-1436:-KsnTsfJQRtMAFfAkRZb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b221ad0524cdb33b97cc9a3b922f3ee1deeb1e76/pkg/sql/parser/sql.y#L1436)):*

nit: I think the indentation here is a little wonky.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18023)*

justinj:
LGTM!

knz:
TFYR!

bdarnell:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 215 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18023#-KsnKbrePCfihPE3jMQM-r1-215:-Kso0Df4xXPm-vLKBRTX:baheigp) ([raw file](https://github.com/cockroachdb/cockroach/blob/b221ad0524cdb33b97cc9a3b922f3ee1deeb1e76/pkg/sql/logictest/testdata/logic_test/delete#L215)):*

The result tag contains the number of rows affected, or, if RETURNING was used, the rows actually deleted. The client has plenty of information to work with. Unfortunately, we can't check the tag in logic tests, but I put a RETURNING test here to showcase. Also indeed, the rows are non-deterministic so we need count(*).
</blockquote></details>

It's probably not worth supporting here, but mysql also extends DELETE with an ORDER BY clause to make the results of a LIMITed DELETE deterministic. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18023)*

PR #18024, opened 2017-08-30
storage: avoid transaction ID pointers in timestamp cache API

tamird:
d0cf7d8 changed the timestamp cache internals to avoid retaining
pointers for reasons of memory retention; this commit propagates that
change to the TS cache API surface, slightly simplifying various code
paths.

-----

@petermattis: was there a reason you avoided this in d0cf7d8?


petermattis:
:lgtm:

> was there a reason you avoided this in d0cf7d8?

Yes. It seemed like a lot of changes for little benefit.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18024#-:-Kso1uNycb6RiJe8eCih:b-x2qrao)*

PR #18030, opened 2017-08-30
sql: support ORDER BY / LIMIT clauses with DELETE and UPDATE statements

knz:
Suggested by @bdarnell in https://github.com/cockroachdb/cockroach/pull/18023#issuecomment-326049415:

> MySQL also extends DELETE with an ORDER BY clause to make the
> results of a LIMITed DELETE deterministic.

Actually this helps making the output of RETURNING deterministic in
general, whether or not LIMIT is present.


bdarnell:
:lgtm:

Since these are apparently so easy, I'll also mention that MySQL also adds `ORDER BY` and `LIMIT` to `UPDATE`. 

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 183 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18030#-Kso4dyXLR0pFsdSGjdl:-Kso4dyXLR0pFsdSGjdm:bz7wzeq) ([raw file](https://github.com/cockroachdb/cockroach/blob/896dd3f4a5db156e53c57f8d14d062962a9e7d52/pkg/sql/logictest/testdata/logic_test/delete#L183)):*
> ```
> 
> query ITTT
> EXPLAIN DELETE FROM unindexed WHERE v = 5 ORDER BY v LIMIT 10
> ```

Keep the old test to ensure that DELETE with LIMIT but without ORDER BY doesn't insert a `sort` step.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18030#-:-Kso50xnJqjd1BYT0qYX:bg9mkow)*

knz:
Done, PTAL.

PR #18037, opened 2017-08-30
storageccl: turn default import batch size down to 32MB

benesch:
63MB seemed like a safe size for the import batch size, but resulted in errors
like

    command is too large: 68166516 bytes (max: 67108864)

on some restores. Turn the size down to 32MB to avoid these errors.

### DIFF
--- a/pkg/ccl/storageccl/import.go
+++ b/pkg/ccl/storageccl/import.go
@@ -42,7 +42,7 @@ var importRequestLimiter = makeConcurrentRequestLimiter(importRequestLimit)
 var importBatchSize = settings.RegisterByteSizeSetting(
 	"kv.import.batch_size",
 	"",
-	63<<20,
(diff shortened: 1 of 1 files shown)


danhhz:
:lgtm:

I think it's worth merging this as-is. But we may want to consider removing the maxImportBatchSize logic as a followup. Doesn't seem super useful anymore

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18037#-:-KsoXk_WocbXNLUd88gL:b-1g9cfw)*

benesch:
Agreed, but I think it's worth understanding why it's broken first. Concerning that Raft seems to be adding 10MB of overhead to the batch request.

danhhz:
Wanna file a 1.2 issue to follow up?

PR #18040, opened 2017-08-30
cli: default --cache and --max-sql-memory to 128MiB

### LINKED ISSUE #14512: perf: RocksDB block cache grows over time
The RocksDB block cache will eventually grow over time to reach its configured limit, even in an idle cluster. This is a combination of our time-series data storage and the fact that RocksDB does not purge old data from the cache until it reaches its memory limit, so the cache can use more memory than the on-disk storage (because it holds data for sstables that have been compacted away). This is surprising and results in a lot of user reactions like "CockroachDB is leaking memory! I started a node, forgot about it for a week, and now it's using 6GB of RAM".

I'm filing this issue in part to give us a canonical thread to point to for this issue, and also to discuss things that we might do to make it better. Possible mitigations include
- Document this issue more prominently and encourage running with a reduced `--cache-size` flag and/or in a memory-limited cgroup (the latter would also reduce the sql memory budget and give us more flexibility to tweak memory allocation between different pools in the future, but it appears to be more complicated)
- Change the default cache size. The current default of 25% of physical memory (or the cgroup limit) was chosen on the assumption that cockroach would have the machine to itself. Maybe it's more important to avoid excessive memory consumption when there's other stuff on the machine than to maximize the cache size on dedicated machines. This shifts the burden so that deployments on dedicated machines would need to set `--cache-size` for best performance. (we should probably also benchmark different cache sizes to see how much it matters)
- Change RocksDB to purge compacted sstables from the block cache. This would still allow the cache to grow over time to its full limit, but it would slow the growth and limit the size of the cache to the size of data on disk.
- Change RocksDB to preallocate the entire block cache at startup. This will make the issue of cache size more visible and make it clear that it's a fixed-size cache instead of a leak.

petermattis:
Previously, these settings defaulted to 1/4 of physical memory. That is
a reasonable default for production, but not for development or an
environment where cockroach is sharing resources with other processes on
the machine. For example, it was previously very easy to run a 5-node
cluster locally on a single machine and have each cockroach node
configured to use 1/4 of physical memory for cache. Leave it running too
long and the cluster will crash.

Using a relatively small cache size by default also avoids the perceived
memory leak where cockroach will grow to fill the cache even when idle
due to the writing of timeseries data.

The downside to this change is that we need to educate users to increase
--cache and --max-sql-memory when deploying to production. This change
adds a warning to the logs when running with the default settings.

```
  Using the default setting for --cache (128 MiB).
  A significantly larger value is usually needed for good performance.
  If you have a dedicated server a reasonable setting is --cache=25% (4.0 GiB).
```

A future enhancement can add a warning to the admin UI.

Allow specifying a percentage of physical memory to --cache and
--max-sql-memory. This allows an easy fallback to the previous behavior
by specifying --cache=25% and --max-sql-memory=25%.

Fixes #14512

### DIFF
--- a/pkg/cli/cliflags/flags.go
+++ b/pkg/cli/cliflags/flags.go
@@ -95,8 +95,8 @@ accept requests.`,
 Total size in bytes available for use to store temporary data for SQL
 clients, including prepared queries and intermediate data rows during
 query execution. Size suffixes are supported (e.g. 1GB and 1GiB). If
-left unspecified, defaults to 25% of the physical memory, or 512MB if
-the memory size cannot be determined.`,
+left unspecified, defaults to 128MiB. A percentage of physical memory
+can also be specified (e.g. 25%).`,
 	}
 
 	Cache = FlagInfo{
@@ -104,8 +104,8 @@ the memory size cannot be determined.`,
 		Description: `
 Total size in bytes for caches, shared evenly if there are multiple
 storage devices. Size suffixes are supported (e.g. 1GB and 1GiB).
-If left unspecified, defaults to 25% of the physical memory, or
-512MB if the memory size cannot be determined.`,
+If left unspecified, defaults to 128MiB. A percentage of physical memory
+can also be specified (e.g. 25%).`,
 	}
 
 	ClientHost = FlagInfo{
--- a/pkg/cli/debug.go
+++ b/pkg/cli/debug.go
@@ -70,7 +70,7 @@ func parseRangeID(arg string) (roachpb.RangeID, error) {
 }
 
 func openStore(cmd *cobra.Command, dir string, stopper *stop.Stopper) (*engine.RocksDB, error) {
-	cache := engine.NewRocksDBCache(512 << 20)
+	cache := engine.NewRocksDBCache(server.DefaultCacheSize)
 	defer cache.Release()
 	maxOpenFiles, err := server.SetOpenFileLimitForOneStore()
 	if err != nil {
--- a/pkg/cli/flags.go
+++ b/pkg/cli/flags.go
@@ -30,7 +30,6 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/cli/cliflags"
 	"github.com/cockroachdb/cockroach/pkg/security"
 	"github.com/cockroachdb/cockroach/pkg/util/envutil"
-	"github.com/cockroachdb/cockroach/pkg/util/humanizeutil"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/cockroachdb/cockroach/pkg/util/log/logflags"
 )
@@ -247,12 +246,8 @@ func init() {
 		varFlag(f, &serverCfg.JoinList, cliflags.Join)
 
 		// Engine flags.
-		setDefaultSizeParameters(&serverCfg)
-		cacheSize := humanizeutil.NewBytesValue(&serverCfg.CacheSize)
-		varFlag(f, cacheSize, cliflags.Cache)
-
-		sqlSize := humanizeutil.NewBytesValue(&serverCfg.SQLMemoryPoolSize)
-		varFlag(f, sqlSize, cliflags.SQLMem)
+		varFlag(f, cacheSizeValue, cliflags.Cache)
+		varFlag(f, sqlSizeValue, cliflags.SQLMem)
 	}
 
 	for _, cmd := range certCmds {
--- a/pkg/cli/interactive_tests/test_sql_mem_monitor.tcl
+++ b/pkg/cli/interactive_tests/test_sql_mem_monitor.tcl
@@ -44,7 +44,7 @@ send "ulimit -v [ expr {2*$vmem+400} ]\r"
 eexpect ":/# "
 
 # Start a server with this limit set. The server will now run in the foreground.
-send "$argv start --insecure --no-redirect-stderr -s=path=logs/db \r"
+send "$argv start --insecure --max-sql-memory=25% --no-redirect-stderr -s=path=logs/db \r"
 eexpect "restarted pre-existing node"
 sleep 1
 
--- a/pkg/cli/start.go
+++ b/pkg/cli/start.go
@@ -27,6 +27,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strconv"
 	"strings"
 	"syscall"
 	"text/tabwriter"
@@ -47,6 +48,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/envutil"
 	"github.com/cockroachdb/cockroach/pkg/util/grpcutil"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
+	"github.com/cockroachdb/cockroach/pkg/util/humanizeutil"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/cockroachdb/cockroach/pkg/util/log/logflags"
 	"github.com/cockroachdb/cockroach/pkg/util/stop"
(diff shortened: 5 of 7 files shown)


knz:
The default should be the min(1/4 * Ram, 128MiB), not just 128MiB.

bdarnell:
:lgtm: 

> The default should be the min(1/4 * Ram, 128MiB), not just 128MiB.

I don't think we can run in under 512MiB (we have reports of OOMs with 1GiB of RAM, which is why we list 2GiB as our minimum requirement).

---

Reviewed 6 of 6 files at r1.

---

*[pkg/cli/debug.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18040#-KsoeG4_stWj9xcNAQVp:-KsoeG4_stWj9xcNAQVq:b-tjbxen) ([raw file](https://github.com/cockroachdb/cockroach/blob/618fcaa9e95e3ae6ad7a236583c68861cc167fbe/pkg/cli/debug.go#L73)):*
> ```Go
> 
> func openStore(cmd *cobra.Command, dir string, stopper *stop.Stopper) (*engine.RocksDB, error) {
> 	cache := engine.NewRocksDBCache(128 << 20)
> ```

Refer to the constant in the `server` package? 

---

*[pkg/cli/start.go, line 532 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18040#-KsoeZ0URePModvj4hif:-KsoeZ0URePModvj4hig:b-2k44zq) ([raw file](https://github.com/cockroachdb/cockroach/blob/618fcaa9e95e3ae6ad7a236583c68861cc167fbe/pkg/cli/start.go#L532)):*
> ```Go
> 	}
> 	fmt.Fprintf(&buf, ".")
> 	log.Shout(context.Background(), log.Severity_WARNING, buf.String())
> ```

I don't like shouting for this; I think it's going to be too intrusive for development use (where you're OK with the defaults), and not visible enough for production use (where stderr might not go anywhere useful). I think I'd just let this go to the logs. 

Also, while we know that increasing the cache size improves performance, increasing `--max-sql-memory` doesn't. It just increases the size of the queries you can run, so many users should be fine leaving it at the default. (alternately, since a too-high value of `--max-sql-memory` doesn't have any cost until you issue a large query, it might be OK to leave it at the former default of 25% of physical memory). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18040#-:-KsofS5DD6scJLko_lWd:b-yyqfrw)*

petermattis:
> I don't think we can run in under 512MiB (we have reports of OOMs with 1GiB of RAM, which is why we list 2GiB as our minimum requirement).

It's not hard to do as @knz suggests. Or perhaps we should warn if physical memory is too small. 

---

Review status: 3 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/cli/debug.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18040#-KsoeG4_stWj9xcNAQVp:-KsogfdiUQSLYbaEwtuL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/618fcaa9e95e3ae6ad7a236583c68861cc167fbe/pkg/cli/debug.go#L73)):*

Refer to the constant in the `server` package? 
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 532 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18040#-KsoeZ0URePModvj4hif:-KsoggLiElSeLBzLEb21:b-57yq1g) ([raw file](https://github.com/cockroachdb/cockroach/blob/618fcaa9e95e3ae6ad7a236583c68861cc167fbe/pkg/cli/start.go#L532)):*

I don't like shouting for this; I think it's going to be too intrusive for development use (where you're OK with the defaults), and not visible enough for production use (where stderr might not go anywhere useful). I think I'd just let this go to the logs. 

Also, while we know that increasing the cache size improves performance, increasing `--max-sql-memory` doesn't. It just increases the size of the queries you can run, so many users should be fine leaving it at the default. (alternately, since a too-high value of `--max-sql-memory` doesn't have any cost until you issue a large query, it might be OK to leave it at the former default of 25% of physical memory). 
</blockquote></details>

Ok, changed to `log.Warning`. Feels weird to leave `--max-sql-memory` at 25% of physical memory while lowering `--cache` to 128MiB. I'd prefer to lower both and get rid of the suggestion to increase `--max-sql-memory`. @knz Thoughts? How much memory do queries typically use know that we have external storage?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18040#-:-KsohmAWHfyVmMGv7s0n:b-agwgc9)*

a-robinson:
I'm pretty disappointed by the effect that this has on production-ready deployment guides. Commands that used to be copy-and-pasteable from docs directly to terminal will now require the user to manually specify an amount of memory for the cache, making for more manual work. Blog posts and tutorials about how to use Cockroach get more complicated. There's also the fact that needing to make the decision at all adds cognitive overhead. Each extra decision a user has to make when onboarding makes it more likely that they'll give up along the way.

This also breaks our Kubernetes and Swarm configs since we can't possibly know in advance how much memory is available in users' clusters but they expect our configs to be production-ready. The most likely fix for them would be to wrap the cockroach invocation in some bash that does what `setDefaultSizeParameters` previously did, which isn't pretty.

It would take a little more work, but I think that the above problems would be lessened somewhat if the `--cache` flag was capable of taking a percentage rather than an absolute amount of memory. Then our config files could continue being agnostic of what kind of machines you're using, and deployment guides targeted at production use cases could more easily generalize. It could add a little confusion if someone uses one of these production guides on their local dev machine, but at least in that case the command-line flag would make it a little more apparent what the memory is being used for.

a-robinson:
To clarify for my incorrect use of "rather than" above, I don't mean to suggest `--cache` only take a percentage, but that ideally it'd allow either a percentage or the humanized bytes format it allows today.

petermattis:
@a-robinson Adding `--cache=25%` and `--max-sql-memory=25%` is fairly easy. The default would still be 128MiB, though, right?

a-robinson:
Right, I'm not arguing against changing the default.

jseldess:
I agree with @a-robinson's sentiment that it's not ideal to make running in production more complex. Why are we prioritizing local clusters over prod clusters? 

> The downside to this change is that we need to educate users to increase --cache and --max-sql-memory when deploying to production. This change adds a warning to the logs when running with the default settings.

An alternative would be to educate users to decrease `--cache` and `--max-sql-memory` when running locally...

petermattis:
> An alternative would be to educate users to decrease --cache and --max-sql-memory when running locally...

Isn't that what we've been doing so far? Doesn't seem to be working. 

@a-robinson Added a commit to support percentages for `--cache` and `--max-sql-memory`.

jseldess:
We could directly incorporate the flag into the local deployment commands. Right now, it's just a callout that people mostly seem to miss. In any case, this isn't a huge change to prod deployment docs, but extra flags do make things look more complex.

petermattis:
@jseldess Agreed this is a large change to prod deployment docs. We shouldn't do this lightly and this PR was intended to focus discussion. I think we've generated a number of negative impressions by the perceived memory leak from not setting `--cache` in development environments. Might this change generate more and severe negative impressions from the performance hit in production environments? One point arguing for this is that setting the cache size is a pretty obvious optimization knob for databases.

jseldess:
I see. Didn't know about the negative impressions we've been getting here. It'll be easy to include these flags, set to 25%, in deployment docs.

jseldess (approved):
LGTM

petermattis:
I'm going to let this bake for a day, but plan on squashing+merging Thursday afternoon if I don't here additional objections.

knz:
LGTM. An alternative to consider is a pair `--mode=dev` and `--mode=prod` which would auto-tune many defaults at the same time: cache size, sql memory size, but also possibly `--insecure`, `--host`, `--logtostderr`, etc.

petermattis:
> An alternative to consider is a pair --mode=dev and --mode=prod which would auto-tune many defaults at the same time: cache size, sql memory size, but also possibly --insecure, --host, --logtostderr, etc.

Acknowledging the alternative as something we should consider for the future. We did have a `--dev` flag at one point that enabled `--insecure`. Didn't seem particularly useful for controlling one setting, but perhaps when there are multiple settings that it controls.

PR #18042, opened 2017-08-30
storage/engine,storageccl: account for entire SST size

benesch:
We weren't properly accounting for the full size of the SST when
determining where to chunk SSTs for import. This matters now that we
need to keep the size of an AddSSTable command from exceeding the
maximum Raft command size. For context, a DataSize of 53MB can translate
to a FileSize of 64MB.

### DIFF
--- a/c-deps/libroach/db.cc
+++ b/c-deps/libroach/db.cc
@@ -2484,11 +2484,12 @@ DBStatus DBSstFileWriterOpen(DBSstFileWriter* fw) {
   return kSuccess;
 }
 
-DBStatus DBSstFileWriterAdd(DBSstFileWriter* fw, DBKey key, DBSlice val) {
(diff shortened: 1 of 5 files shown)


danhhz (approved):
nice find!

bdarnell:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18042#-:-KstDK7J4-UkxQ2MH_L6:bnfp4nl)*

benesch:
Urgh, RocksDB reports the file size as 0 until the first block is filled, which is the reason that half the backup/restore tests broke from this change. Pushed an update, but might be worth waiting on 1.2 for this. Using a chunk size of 32MB (instead of 64MB - FUDGE) papers over this problem.

benesch:
Using 32MiB as the limit has proven to work quite well in practice. Calling this unnecessary.

PR #18043, opened 2017-08-30
sqlccl: improve scatter usage in RESTORE

danhhz:
Lowers BenchmarkRestore2TB from ~3.5h to ~55m. This commit is a very
close derivative of an idea that benesch had, so all credit goes to
him.

Splits and scatters are now run with some amount of parallelism but also
staying as close to the order in importSpans as possible (the more out
of order, the more work is done if a RESTORE job loses its lease and has
to be restarted).

At a high level, this is accomplished by splitting and scattering large
"chunks" from the front of importEntries in one goroutine, each of which
are in turn passed to one of many worker goroutines that split and
scatter the individual entries.

image

image

### DIFF
--- a/pkg/ccl/sqlccl/restore.go
+++ b/pkg/ccl/sqlccl/restore.go
@@ -716,17 +716,9 @@ func restore(
 		// leaseholders, so presplit and scatter the ranges to balance the work
 		// among many nodes.
 		for i, importSpan := range importSpans {
-			var newSpan roachpb.Span
-			{
-				var ok bool
-				newSpan.Key, ok, _ = kr.RewriteKey(append([]byte(nil), importSpan.Key...))
-				if !ok {
-					return errors.Errorf("could not rewrite key: %s", importSpan.Key)
-				}
-				newSpan.EndKey, ok, _ = kr.RewriteKey(append([]byte(nil), importSpan.EndKey...))
-				if !ok {
-					return errors.Errorf("could not rewrite key: %s", importSpan.EndKey)
-				}
+			newSpan, err := kr.RewriteSpan(importSpan.Span)
+			if err != nil {
+				return err
 			}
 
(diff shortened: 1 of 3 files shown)


petermattis:
A 75% speed-up!!! That's huge.

benesch:
:lgtm_strong: 

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18043#-:-Ksp07bjJYgvZXJAoWVZ:b-uwkowd)*

benesch:
Wait, hang on, doesn't this still interact poorly with progress? Won't you
be importing a little bit of each chunk at once? Seems like you want all 2
* NCPUs split/scatter workers to be operating on the same chunk, and only
move to the next chunk when the first chunk is exhausted.

No longer have the code in front of me so perhaps I'm crazy and it already
works like that.

benesch:
But my proposal would break the optimization. Never mind. Perhaps the
chunks just need to be smaller? Sqrt(16k) * 30 if you have a 15 node 2TB
restore seems like a lot of progress to lose. I'd be okay with hardcoding a
chunk size of 50ish. Not sure it actually makes sense to scale chunk size
with the size of the restore.

danhhz:
Yeah, sqrt is not quite right because that's a lot of progress to lose, but you do kind of want it to scale. At least, you don't want to hamstring small restores by hardcoding 50. How do you feel about merging as-is and testing some tweaks to the chunk size calculation as followups (I'm also thinking about log)? That's even the kind of thing we could merge after the 1.1 branch is cut if we come up with something amazing.

benesch:
SGTM!

danhhz:
Okay, then here we go...

benesch:
🎉

PR #18045, opened 2017-08-30
sql: adjust cost function for index selection

### LINKED ISSUE #15649: sql: wrong INDEX chosen by SELECT ORDER BY LIMIT 10000
Config Info:
````
I170502 18:34:55.174939 1 util/log/clog.go:990  [config] file created at: 2017/05/02 18:34:55
I170502 18:34:55.174939 1 util/log/clog.go:990  [config] running on machine: gzns-dba-sz-ndbtest-02
I170502 18:34:55.174939 1 util/log/clog.go:990  [config] binary: CockroachDB CCL v1.0-rc.1-dirty (linux amd64, built 2017/05/01 18:33:34, go1.8.1)
I170502 18:34:55.174939 1 util/log/clog.go:990  [config] arguments: [/home/ndbadmin/ndb/bin//cockroach start --port=26257 --http-port=8080 --store=path=/ssd/ndb-data,attrs=ssd --cache=48G --attrs=gzns --insecure --join=gzns-dba-sz-ndbtest-02.gzns:26257]
````

Observed behavior:
wrong index chosen by `SELECT LIMIT 10000` statement.
````
+-------+-------------------------------------------------+
| Table |                   CreateTable                   |
+-------+-------------------------------------------------+
| test  | CREATE TABLE test (_                            |
|       |     id INT NOT NULL,_                           |
|       |     midname STRING NULL,_                       |
|       |     name STRING NULL,_                          |
|       |     CONSTRAINT "primary" PRIMARY KEY (id ASC),_ |
|       |     INDEX test_name_idx (name ASC),_            |
|       |     FAMILY "primary" (id, midname, name)_       |
|       | )                                               |
+-------+-------------------------------------------------+

root@:26257/test> EXPLAIN SELECT id FROM test ORDER BY id asc LIMIT 10 offset 100;
+-------+--------+-------+--------------+
| Level |  Type  | Field | Description  |
+-------+--------+-------+--------------+
|     0 | limit  |       |              |
|     1 | render |       |              |
|     2 | scan   |       |              |
|     2 |        | table | test@primary |
|     2 |        | spans | ALL          |
|     2 |        | limit |          110 |
+-------+--------+-------+--------------+

root@:26257/test> EXPLAIN SELECT id FROM test ORDER BY id asc LIMIT 10 offset 10000;
+-------+--------+----------+--------------------+
| Level |  Type  |  Field   |    Description     |
+-------+--------+----------+--------------------+
|     0 | limit  |          |                    |
|     1 | sort   |          |                    |
|     1 |        | order    | +id                |
|     1 |        | strategy | top 10010          |
|     2 | render |          |                    |
|     3 | scan   |          |                    |
|     3 |        | table    | test@test_name_idx |
|     3 |        | spans    | ALL                |
+-------+--------+----------+--------------------+
````

justinj:
Fixes #15649.

This is a pretty arbitrary change, I debated making this cost add
instead of multiply, but I figured this made more sense as all the
existing cost changes are multiplicative. Not sure how to gain
confidence that this won't regress any existing index selections,
besides the fact that this passes the logic test for index selection.

### DIFF
--- a/pkg/sql/index_selection.go
+++ b/pkg/sql/index_selection.go
@@ -17,6 +17,7 @@ package sql
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"sort"
 
 	"golang.org/x/net/context"
@@ -30,6 +31,7 @@ import (
 )
 
 const nonCoveringIndexPenalty = 10
+const nonMatchingCostFactor = 2
 
 // analyzeOrderingFn is the interface through which the index selection code
 // discovers how useful is the ordering provided by a certain index. The higher
@@ -435,7 +437,13 @@ func (v *indexInfo) analyzeOrdering(
 		v.reverse = true
 	}
 	weight := float64(orderCols+1) / float64(match+1)
-	v.cost *= weight
+
+	// This is largely arbitrary - we up the priority of this slightly so that it
+	// overcomes the cost of having an index with a large number of columns.
+	// See #15649 for a case where this was an issue.
+	// This uses Pow to be consistent with the other cost modifications which are
+	// all multiplicative.
+	v.cost *= math.Pow(weight, nonMatchingCostFactor)
 
 	if match == orderCols && preferOrderMatching {
 		// Offset the non-covering index cost penalty.
--- a/pkg/sql/logictest/testdata/logic_test/select_index
+++ b/pkg/sql/logictest/testdata/logic_test/select_index
@@ -592,6 +592,38 @@ SELECT * FROM t12022@i WHERE (c1, c2) < (2, true) ORDER BY (c1, c2);
 1  true
 2  false
 
+# Regression tests for #15649.
+
+statement ok
+CREATE TABLE test15649 (
+  id INT NOT NULL,
+  midname STRING NULL,
+  name STRING NULL,
+  CONSTRAINT "primary" PRIMARY KEY (id ASC),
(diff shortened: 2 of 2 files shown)


knz:
Comments please. In code and commit message.

justinj:
OK, PTAL, the more I think about this change the more anxious it makes me. Advice on how to test this more extensively is welcome.

justinj:
Random idea, perhaps we could generate schemas and queries against them, and then compare the ranking of how fast the queries run across different indexes to the ranking given by this code...

knz:
Maybe you can touch base with Alberto about this tomorrow?

RaduBerinde:
Not sure squaring this value makes sense, something that has a lot of columns could be penalized a lot.

Perhaps we want to change `preferOrderMatching` to be a continuum rather than an on/off thing. It could be a value that says "if the order matches, the cost gets scaled by this factor". The factor should be similar to `1/nonCoveringIndexPenalty` for small limits. For no limits, it should be something to account for the benefit of not sorting (e.g. 0.5). A "full" order match scales the cost by that factor, a partial match would interpolate between the two values to pick something in-between.

justinj:
After spending some time on this, I don't really feel comfortable making this change right before the code freeze, so I'm going to bump the corresponding issue to 1.2.

PR #18047, opened 2017-08-30
sqlccl: support relative directories in nodelocal

madelynnblue:
A previous change broke relative directories for `load csv` because
nodelocal didn't treat relative directories correctly. The URLs,
when parsed, would use the first directory part as the URL host,
which was ignored, and thus broke the remainder of the path.

Due to the new checking in nodelocal about Host, this uncovered a
bug where we were double appending `nodelocal://` to some test data.

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -1577,6 +1577,13 @@ func TestAsOfSystemTimeOnRestoredData(t *testing.T) {
 	defer cleanupFn()
 	sqlDB.Exec(`DROP TABLE data.bank`)
 
+	// backupRestoreTestSetup sets nodelocal:// on dir, so strip it off.
+	url, err := url.Parse(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	dir = url.Path
+
 	const numAccounts = 10
 	bankData := sampledataccl.BankRows(numAccounts)
 	backup, err := sampledataccl.ToBackup(t, bankData, filepath.Join(dir, "backup"))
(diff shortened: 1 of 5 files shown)


dt:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/ccl/utilccl/sampledataccl/bankdata.go, line 167 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18047#-Ksp74byY5h8H3uxdhlR:-Ksp74byY5h8H3uxdhlS:bq4hbkm) ([raw file](https://github.com/cockroachdb/cockroach/blob/aeaebbe1d135dd8a82f14cf86e962f3dd67b22cc/pkg/ccl/utilccl/sampledataccl/bankdata.go#L167)):*
> ```Go
> 	// TODO(dan): The csv load will be less overhead, use it when we have it.
> 	ts := hlc.Timestamp{WallTime: hlc.UnixNano()}
> 	desc, err := sqlccl.Load(ctx, db, &stmts, `data`, `nodelocal://`+dir, ts, chunkBytes, tempDir)
> ```

Don't think `Load` is prefixing this, so either want to remove this, or add the same treatment as LoadCSV

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18047#-:-Ksp6gP0kqWHnDkMqEyU:bnfp4nl)*

mjibson:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/ccl/utilccl/sampledataccl/bankdata.go, line 167 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18047#-Ksp74byY5h8H3uxdhlR:-Ksp8is0fmHUHbCEcAhv:b-5urd9g) ([raw file](https://github.com/cockroachdb/cockroach/blob/aeaebbe1d135dd8a82f14cf86e962f3dd67b22cc/pkg/ccl/utilccl/sampledataccl/bankdata.go#L167)):*

Don't think `Load` is prefixing this, so either want to remove this, or add the same treatment as LoadCSV
</blockquote></details>

This was removed because `dir` already has nodelocal at the front (from https://github.com/cockroachdb/cockroach/blob/7d914b11163f69e56e1603df160419ad73d75ca1/pkg/ccl/sqlccl/backup_test.go#L129). That's what the uri.Host check caught.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18047)*

mjibson:
Had to do some annoying stuff to fix the single test that depended on weird behavior. It's not great but it works for all tests now. PTAL.

dt (approved):

PR #17145, opened 2017-07-20
cli/server: Use AdvertiseAddr instead of Addr where appropriate

a-robinson:
**cli: Fix internal client connections to use AdvertiseAddr, not Addr**

As far as I can tell, this wasn't actually breaking anything, since
getClientGRPCConn, the only caller of addrWithDefaultHost, isn't called
from within the start command, which is the only command where
AdvertiseAddr can differ from Addr.

Fixes #17143

**server: Use AdvertiseAddr instead of Addr HTTP<->gRPC connections**

Fixes #10374

PR #18049, opened 2017-08-30
cli: Fix internal client connections to use AdvertiseAddr, not Addr

### LINKED ISSUE #17143: possible bug unused variable in /pkg/cli/start.go
Just browsing the code and noticed on [/pkg/cli/start.go#L626]( https://github.com/cockroachdb/cockroach/blob/master/pkg/cli/start.go#L626) `func addrWithDefaultHost` has an `addr` variable that is not being used anywhere in the function, but it looks like [/pkg/cli/start.go#L649](https://github.com/cockroachdb/cockroach/blob/master/pkg/cli/start.go#L649) is expecting that variable to be the `serverCfg.AdvertiseAddr`. I am not confident enough with the domain to try to fix it , so I am just leaving this here as an FYI.

a-robinson:
As far as I can tell, this wasn't actually breaking anything, since
getClientGRPCConn, the only caller of addrWithDefaultHost, isn't called
from within the start command, which is the only command where
AdvertiseAddr can differ from Addr.

Fixes #17143

Pulled out of #17145, which also included a much more controversial commit.

### DIFF
--- a/pkg/cli/start.go
+++ b/pkg/cli/start.go
@@ -620,7 +620,7 @@ func setupAndInitializeLoggingAndProfiling(ctx context.Context) (*stop.Stopper, 
 }
 
 func addrWithDefaultHost(addr string) (string, error) {
-	host, port, err := net.SplitHostPort(baseCfg.Addr)
+	host, port, err := net.SplitHostPort(addr)
 	if err != nil {
 		return "", err
 	}


petermattis:
:lgtm:, though perhaps `getClientGRPCConn` should also be passing `serverCfg.Addr` to `addrWithDefaultHost`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18049#-:-Ksp2GADn314sDq-KGIb:b-ppnpu3)*

PR #18053, opened 2017-08-30
sqlccl: add some child spans in dist csv path

dt:


mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 995 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-KspWQMbqnBbgzB_JbAO:-KspWQMbqnBbgzB_JbAP:b-jj3ppv) ([raw file](https://github.com/cockroachdb/cockroach/blob/68a270237ade497bd93acbdb851bed83ace1d178/pkg/ccl/sqlccl/csv.go#L995)):*
> ```Go
> 	// Read CSV into CSV records
> 	group.Go(func() error {
> 		gCtx, span := tracing.ChildSpan(gCtx, "readCSVProcessor-readCSV")
> ```

Why do we need `readCSVProcessor-` here if it's already part of the parent span?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 995 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-KspWQMbqnBbgzB_JbAO:-KsqYYjM0Bibl92UVmdw:b-gt6jb) ([raw file](https://github.com/cockroachdb/cockroach/blob/68a270237ade497bd93acbdb851bed83ace1d178/pkg/ccl/sqlccl/csv.go#L995)):*

Why do we need `readCSVProcessor-` here if it's already part of the parent span?
</blockquote></details>

my initial choice was just driven by grep-ability -- given one of these span names, even without looking up its parents I can just grep and jump straight to a single, unique line of code -- but I don't feel strongly about it if you prefer non-prefixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053)*

mjibson:
:lgtm:

---

---

*[pkg/ccl/sqlccl/csv.go, line 995 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-KspWQMbqnBbgzB_JbAO:-Ksq_kSJXbyz6tvgITM7:b2bia5v) ([raw file](https://github.com/cockroachdb/cockroach/blob/68a270237ade497bd93acbdb851bed83ace1d178/pkg/ccl/sqlccl/csv.go#L995)):*

my initial choice was just driven by grep-ability -- given one of these span names, even without looking up its parents I can just grep and jump straight to a single, unique line of code -- but I don't feel strongly about it if you prefer non-prefixed.
</blockquote></details>

I'd prefer to remove the prefix because I think this sets a bad precedent. May make sense to change convert to convertRecords or something, to improve its grep-ability.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-:-Ksq_x0Cd7hhLlclcm0s:bnfp4nl)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 995 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-KspWQMbqnBbgzB_JbAO:-Kssu81VVDsc0JHW6pA6:b-ftzgs9) ([raw file](https://github.com/cockroachdb/cockroach/blob/68a270237ade497bd93acbdb851bed83ace1d178/pkg/ccl/sqlccl/csv.go#L995)):*

I'd prefer to remove the prefix because I think this sets a bad precedent. May make sense to change convert to convertRecords or something, to improve its grep-ability.
</blockquote></details>

Fair enough... @RaduBerinde as far as precedent is concerned, do our tracing people have any best-practices  guidance here, as far as duplicative prefixes vs uniqueness/grep-ability? /cc @RaduBerinde @danhhz 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053)*

RaduBerinde:

---

*[pkg/ccl/sqlccl/csv.go, line 995 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-KspWQMbqnBbgzB_JbAO:-KssuxiAm7yjU1wTSyUb:b-cm6g9w) ([raw file](https://github.com/cockroachdb/cockroach/blob/68a270237ade497bd93acbdb851bed83ace1d178/pkg/ccl/sqlccl/csv.go#L995)):*

Fair enough... @RaduBerinde as far as precedent is concerned, do our tracing people have any best-practices  guidance here, as far as duplicative prefixes vs uniqueness/grep-ability? /cc @RaduBerinde @danhhz 
</blockquote></details>

We don't use these kind of prefixes; the tracing infra maintains the child relations. But I don't see the problem with it if it's necessary for uniqueness - in the vast majority of cases you can find one callsite that creates a span based on just the name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 995 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053#-KspWQMbqnBbgzB_JbAO:-KsswHD3-IxzFTJqt3Oy:b-nkd5ju) ([raw file](https://github.com/cockroachdb/cockroach/blob/68a270237ade497bd93acbdb851bed83ace1d178/pkg/ccl/sqlccl/csv.go#L995)):*

We don't use these kind of prefixes; the tracing infra maintains the child relations. But I don't see the problem with it if it's necessary for uniqueness - in the vast majority of cases you can find one callsite that creates a span based on just the name.
</blockquote></details>

Thanks! de-prefixed, but added `csv` to most of them since e.g. `read`  or `convert` aren't very grep-able.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18053)*

madelynnblue (approved):

PR #18054, opened 2017-08-30
libroach: emulate DB::SyncWAL on Windows using WriteBatch::LogData

petermattis:
Using WriteBatch::LogData avoids adding a key to the memtable for every
sync (a small performance win) and also avoids the need for
StoreSyncKey().

### DIFF
--- a/c-deps/libroach/db.cc
+++ b/c-deps/libroach/db.cc
@@ -203,7 +203,6 @@ struct DBIterator {
 // kKeyLocalRangePrefix are the mvcc-encoded prefixes.
 const rocksdb::Slice kKeyLocalRangeIDPrefix("\x01i", 2);
 const rocksdb::Slice kKeyLocalMax("\x02", 1);
-const rocksdb::Slice kKeyLocalStoreSync("\x01ssync\x00", 7);
 
 const DBStatus kSuccess = { NULL, 0 };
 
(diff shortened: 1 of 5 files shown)


petermattis:
@BramGruneir I'm reasonably confident this does the right thing during `DB::Write()`, but it would be great if you have some way of verifying on Windows that a sync is being performed. Alternately, I can enable this code path on Unix temporarily and verify the syncing there.

petermattis:
> Alternately, I can enable this code path on Unix temporarily and verify the syncing there.

Ok, I did this and RocksDB is definitely calling `Sync` on the log file when writing the batch containing the `LogData`.

bdarnell:
:lgtm: assuming it works as expected on windows.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18054#-:-KstGqJZNmbNYeLSfFF_:bcgzkvb)*

BramGruneir:
I don't have a readily available way of checking this in windows, besides instrumenting some code in rocksdb directly.  But I can check to make sure that it will run on windows.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18054#-:-KstH_X9o7F-OLSOqfIe:b-y3nylh)*

BramGruneir:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18054)*

petermattis:
Ok. I'm pretty confident it will call Sync on Windows because I verified that this new code has that effect on Unix. Won't the Windows build happen automatically after I merge?

BramGruneir:
I think only on release.  But post merge, I'll just build one locally and make sure it runs.

BramGruneir:
Just to followup on this, I tested a latest build on Windows and everything seemed correct.

PR #18066, opened 2017-08-31
sql: avoid integer overflow in generate_series

### LINKED ISSUE #18021: sql: func exec timeout GENERATE_SERIES
`panic: func exec timeout: SELECT GENERATE_SERIES(9223372036854775807::int, -9223372036854775807::int, -9223372036854775807::int)`

The above error (it's not a real panic, just a timeout enforced by RSG) occurred. It's possible this isn't a bug and that function is supposed to take >10s to complete (see https://github.com/cockroachdb/cockroach/blob/6151997b9ca391811144e89e5544ede2a13ddac5/pkg/sql/rsg_test.go#L204). But filing here for triage or closure.

https://teamcity.cockroachdb.com/viewLog.html?buildId=338204&buildTypeId=Cockroach_Nightlies_RandomSyntaxTests

Found with RSG.

madelynnblue:
Fixes #18021

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/generators
+++ b/pkg/sql/logictest/testdata/logic_test/generators
@@ -234,3 +234,11 @@ SELECT * from generate_series(1, (select * from generate_series(1, 0)))
 query T
 SELECT unnest((select current_schemas((select isnan((select round(3.4, (select generate_series(1, 0)))))))));
 ----
+
+# Regression for #18021.
+query I
+SELECT GENERATE_SERIES(9223372036854775807::int, -9223372036854775807::int, -9223372036854775807::int)
+----
+9223372036854775807
+0
+-9223372036854775807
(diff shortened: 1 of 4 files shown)


pkg/sql/logictest/testdata/logic_test/generators
```diff
@@ -234,3 +234,10 @@ SELECT * from generate_series(1, (select * from generate_series(1, 0)))
 query T
 SELECT unnest((select current_schemas((select isnan((select round(3.4, (select generate_series(1, 0)))))))));
 ----
+
+# Regression for #18021.
+query I
+SELECT GENERATE_SERIES(9223372036854775807::int, -9223372036854775807::int, -9223372036854775807::int)
+----
+9223372036854775807
+0
```
knz: I'm not sure that's correct. the right bound for `generate_series` is inclusive.
If we can't reach that value at all, then the call should produce an error instead.

knz (changes requested):

pkg/sql/logictest/testdata/logic_test/generators
mjibson: Good catch. Updated to fix this.

knz:
:lgtm:

---

Reviewed 3 of 4 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18066#-:-KstLAlh2zwrj2-UuUWV:bnfp4nl)*

knz (approved):

Issue #9797, opened 2016-10-07
stability: debug endpoint for command queue introspection


tschottdorf:
In https://github.com/cockroachdb/cockroach/issues/9749, we were seeing the command queue on time series Replicas overflooded and performing poorly as a result.

We should build a debug endpoint which exposes data that helps debug these issues the future. As for details on what should be exposed, I defer to @nvanbenschoten.

dianasaur323:
@nvanbenschoten I'm cleaning out Q1 2017 milestones. I think I want to move this to Q3 2017 if it's still relevant. Feel free to adjust if you disagree.

nvanbenschoten:
Q3 seems appropriate to me. Thanks @dianasaur323.

petermattis:
This should probably live on the /debug/range page. Or maybe summarized on the /debug/range page with a link for more details about the commands sitting in the queue for a specified range.

cuongdo:
@BramGruneir is this still happening for 1.1?

BramGruneir:
If it's just a metric that we're adding to the debug range page, than yes.
But if it's a new endpoint and specialized page, I think we should push it out.

petermattis:
I think there is both short term and long term work. Short term let's add a metric for the size of the command queue. I think this can get into 1.1. Long term I think we want a new debug page for introspection of what the command queue contains. Let's push this to 1.2.

BramGruneir:
SGTM

a-robinson:
> Long term I think we want a new debug page for introspection of what the command queue contains.

<img src="http://i3.kym-cdn.com/photos/images/facebook/000/104/953/do_want.jpg">

cuongdo:
@nvanbenschoten Can you fill in some details regarding what's needed here? I'm intending for this to be a starter project to be worked on starting next week.

nvanbenschoten:
I think @BramGruneir actually just got started on this.

At a minimum, we'd want information about the size of the `CommandQueue`, similar to what we have in [`updateCommandQueueGauges`](https://github.com/cockroachdb/cockroach/blob/c01bef22d97156f2d125ea9caa16ee28b5655614/pkg/storage/store.go#L4048), but on a per-Replica basis.

[...]

PR #18067, opened 2017-08-31
server, storage, ui: Add command queue metrics to the range debug page

BramGruneir:
This adds write_commands, read_commands, max_overlaps_seen and tree_size for both local and global command queues.

Part of #9797.

screen shot 2017-08-31 at 00 58 52


tschottdorf:
These metrics are all-time, right? Then this may not be so useful -- when looking at the range page, we'd be interested more in a current snapshot. `write_commands` and `read_commands` as an ever-increasing counter kind of work -- you can refresh the page and guess at the rate at which they're growing -- but `max_overlaps_seen` isn't useful once it's large as then you don't know whether that's happening now or happened ten days ago.

Some of the existing fields have similar problems (Raft dropped cmds, etc), but those are counters, so you can work around it -- tracking a max is more difficult.

Apologies if these metrics aren't all-time -- I've only skimmed the diff. Either way, great to see work in this area! I've often wondered about the command queue internals.

BramGruneir:
max_overlaps_seen is of all time, but all the rest are current. 

When a new command is added, it is added to read or write.
When a command is removed, it is subtracted from them.  Take a look at the `remove` function in the command queue.

tree_size is also current, however unlike the others, it is calculated on the fly.

tschottdorf:
Cool, so only max overlaps is awkward. I'm sure it's still better than not having it. Thanks for the clarifications!

nvanbenschoten:
Reviewed 7 of 12 files at r1.
Review status: 7 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/storage/replica.go, line 5334 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-Kst_ZpuCcEuHlFaKv6Z:-Kst_ZpuCcEuHlFaKv6_:bv5ritc) ([raw file](https://github.com/cockroachdb/cockroach/blob/974aa52b2289be51b635778dd97c4f93708ab808/pkg/storage/replica.go#L5334)):*
The locking notes on `Replica.cmdQMu` say:

> Locking notes: Replica.mu < Replica.cmdQMu

Since we don't want to risk getting an inconsistent metric state, let's put this lock inside `r.mu.RLock()`.

---

*[pkg/storage/replica.go, line 5335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-Kst_pRvtqUsF_-AJxSM:-Kst_pRvtqUsF_-AJxSN:b-6djpaa) ([raw file](https://github.com/cockroachdb/cockroach/blob/974aa52b2289be51b635778dd97c4f93708ab808/pkg/storage/replica.go#L5335)):*
> ```Go
> 
> 	r.cmdQMu.Lock()
> 	commandQueueMetricsLocal := r.cmdQMu.queues[spanLocal].metrics()
> ```

nit everywhere: we generally use variables called `cmdQ` instead of `commandQueue`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067)*

BramGruneir:
Comments addressed.

Here's an updated view:


---

Review status: 2 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/storage/replica.go, line 5334 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-Kst_ZpuCcEuHlFaKv6Z:-KstiXNnuExZ2Mfjmnvi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/974aa52b2289be51b635778dd97c4f93708ab808/pkg/storage/replica.go#L5334)):*

The locking notes on `Replica.cmdQMu` say:

> Locking notes: Replica.mu < Replica.cmdQMu

Since we don't want to risk getting an inconsistent metric state, let's put this lock inside `r.mu.RLock()`.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 5335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-Kst_pRvtqUsF_-AJxSM:-KstiXjyOAGf4235Wnog:b-an0ur8) ([raw file](https://github.com/cockroachdb/cockroach/blob/974aa52b2289be51b635778dd97c4f93708ab808/pkg/storage/replica.go#L5335)):*

nit everywhere: we generally use variables called `cmdQ` instead of `commandQueue`.
</blockquote></details>

Done.

I've kept the proto message type as CommandQueueMetrics, but everything else has been shortened.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-:-KstoycCerTGUe5BBZK2:b6pp4jo)*

bdarnell:
:lgtm:

---

Review status: 2 of 13 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-:-KsuJ_BZ_QtIbK5BEBiw:bnfp4nl)*

nvanbenschoten:
:lgtm: Thanks Bram!

---

Review status: 2 of 13 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18067#-:-KsuKPiNrHMlYiF2mhQQ:b-51rtoj)*

PR #18069, opened 2017-08-31
server, ui: Fixes to the raft debug page and endpoint.

BramGruneir:
- fix the `/_status/raft/` endpoint from falsely claiming that all range descriptors are different
- fix the links to the range debug pages for ranges listed on the raft debug page


BramGruneir:
Who doesn't like a commit that is literally only removing 2 characters, a `#` and an `&`.  

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18069#-:-Ksqzx1LbF4cfIfWEAfJ:b1xbgke)*

tschottdorf:
💯 would review again, LGTM!

PR #18073, opened 2017-08-31
log: retire the proto field FileDetail.severity (we now only have 1 log file)

### LINKED ISSUE #14327: logs: suggested improvements
Following today's discussion:

- [x] send all logs to a single file (proposed by peter, #13905)
- [x] fix the entry reader to load the severity back from the individual log lines.
- [x] fix unwanted side effects of 31232153983ba65a74b39e83a0bd0c53099ec873, which causes *everything* otherwise printed to stderr to go *only* to the logs: - #14364
   - [x] ensure that when a TestLogScope is activated, the message "the log files can be found in ..." is printed to the real stderr
   - [x] install a defer/recover in the top-level start function so that a message gets printed to the real stderr informing the user to look a the logs when a panic occurs
   - [x] ~~tee the text of panics both to the log and stderr (potentially with a special case to avoid duplication if logs are going to stderr).~~ (not easy to do; panics go directly to fd 2 and that is piped to the log file.)
- [x] ~~make `-show-logs` become the default when the output is a TTY (e.g. `make test` at the shell prompt)~~ (not possible, `go test` messes with the file descriptors)
  - [x] tell the user about the flag `-show-logs` when the log scope is activated - #14364

- [x] investigate whether we can capture artifacts and panics for `testrace` (not currently available in TC, as found on #14317)
- [ ] suggested by Tamir: ensure that the now-unused "level"/"severity" fields in the admin RPC protobufs are marked "reserved" and update the UI code accordingly #18073

knz:
Suggested by @tamird.
Fixes #14327.


tamird:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18073#-:-KsslEJhMAltXWFN2PlB:bnfp4nl)*

knz:
TFYR!

PR #18074, opened 2017-08-31
sqlccl: batch work through channels

dt:
```
name      old time/op    new time/op    delta
Import-4    9.02ms ±13%    7.98ms ±12%  -11.54%  (p=0.011 n=10+10)

name      old alloc/op   new alloc/op   delta
Import-4    2.40MB ± 3%    3.07MB ± 1%  +28.18%  (p=0.000 n=10+8)

name      old allocs/op  new allocs/op  delta
Import-4     41.3k ± 1%     42.3k ± 0%   +2.32%  (p=0.000 n=10+8)
```

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -135,11 +135,11 @@ func doLocalCSVTransform(
 ) (csvCount, kvCount, sstCount int64, err error) {
 	// Some channels are buffered because reads happen in bursts, so having lots
 	// of pre-computed data improves overall performance.
-	const chanSize = 10000
+	chanSize := 10 * runtime.NumCPU()
 
 	group, gCtx := errgroup.WithContext(ctx)
-	recordCh := make(chan csvRecord, chanSize)
-	kvCh := make(chan roachpb.KeyValue, chanSize)
+	recordCh := make(chan csvRecordBatch, chanSize)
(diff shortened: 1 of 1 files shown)


dt:
need to benchmark this on an real machine before I'm confident in the numbers (my little MacBook does a lot of thermal throttling), but wanted to get ci/review opinions started in the meantime.

dt:
my azure worker's run doesn't indicate a conclusive win either:
```
name      old time/op    new time/op    delta
Import-4    6.50ms ±14%    6.66ms ±30%     ~     (p=0.662 n=23+24)

name      old alloc/op   new alloc/op   delta
Import-4    2.44MB ± 4%    3.11MB ± 3%  +27.30%  (p=0.000 n=24+23)

name      old allocs/op  new allocs/op  delta
Import-4     41.5k ± 2%     42.5k ± 3%   +2.47%  (p=0.000 n=24+24)
```

mjibson:
Well the code looked good. Thanks for doing this test.

dt:
it might still bear out with some further tweaking, but doesn't seem worth merging as-is.

mjibson:
I did lots of tests with chan buffer sizes for single node before deciding on the current value. I recall doing some tests with exactly what this PR did, but I don't remember the results (maybe because they were a wash as you found, and just added more complexity).

PR #18081, opened 2017-08-31
log: allow fully-verbose exfiltration

tbg:
While debugging production clusters, it has occurred frequently that
we needed to bump `--vmodule` to get at relevant log traffic from, say,
the bowels of Raft. In production settings and in particular with clusters
that we don't have direct access to, this has proven difficult and time-
consuming.

The idea here is to have an easily accessible endpoint that can "siphon off"
small chunks of completely verbose log traffic for short amounts of time.

Whether this endpoint is accessible from the cli, sql shell, debug ui or all
of them remains to be decided. For now, I went with the easiest option:

```
$ curl http://localhost:8080/debug/logspy
[... messages stream for ~5s ...]
```

I like this, but before making it a real thing I wanted to gauge the initial
reaction.

It will be straightforward to add vmodule-style filtering, grep, and basically
anything else that we'd like. Actually, if I lift the invocation of the
interceptor up one level to `addStructured`, we'll have access to the original
logging context (so can filter by range, node, etc, without bending over
backwards).

### DIFF
--- a/pkg/server/debug.go
+++ b/pkg/server/debug.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/cockroachdb/cockroach/pkg/settings"
 	"github.com/cockroachdb/cockroach/pkg/settings/cluster"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/pkg/errors"
 	"github.com/rcrowley/go-metrics"
 	"github.com/rcrowley/go-metrics/exp"
@@ -37,6 +38,8 @@ import (
 // debugEndpoint is the prefix of golang's standard debug functionality
 // for access to exported vars and pprof tools.
 const debugEndpoint = "/debug/"
+
+const debugLogSpyEndpoint = "/debug/logspy"
 
 // We use the default http mux for the debug endpoint (as pprof and net/trace
 // register to that via import, and go-metrics registers to that via exp.Exp())
@@ -149,6 +152,11 @@ func init() {
 </html>
 `)
 	})
+
+	spy := logSpy{
+		setIntercept: log.Intercept,
+	}
+	debugServeMux.HandleFunc(debugLogSpyEndpoint, spy.handleDebugLogSpy)
 
 	// This registers a superset of the variables exposed through the /debug/vars endpoint
 	// onto the /debug/metrics endpoint. It includes all expvars registered globally and
--- a/pkg/util/log/clog.go
+++ b/pkg/util/log/clog.go
@@ -36,6 +36,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/net/context"
+
 	"github.com/cockroachdb/cockroach/pkg/build"
 	"github.com/cockroachdb/cockroach/pkg/util/caller"
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
@@ -437,7 +439,7 @@ func formatHeader(
 	if line < 0 {
 		line = 0 // not a real line number, but acceptable to someDigits
 	}
-	if s > Severity_FATAL {
+	if s > Severity_FATAL || s <= Severity_UNKNOWN {
 		s = Severity_INFO // for safety.
 	}
 
@@ -462,6 +464,9 @@ func formatHeader(
 	// Lyymmdd hh:mm:ss.uuuuuu file:line
 	tmp[n] = severityChar[s-1]
 	n++
+	if year < 2000 {
+		year = 2000
+	}
 	n += buf.twoDigits(n, year-2000)
 	n += buf.twoDigits(n, int(month))
 	n += buf.twoDigits(n, day)
@@ -644,6 +649,8 @@ type loggingT struct {
 	verbosity level         // V logging level, the value of the --verbosity flag/
 	exitFunc  func(int)     // func that will be called on fatal errors
 	gcNotify  chan struct{} // notify GC daemon that a new log file was created
+
+	interceptor atomic.Value // InterceptorFn
 }
 
 // buffer holds a byte Buffer for reuse. The zero value is ready for use.
@@ -708,19 +715,18 @@ func (l *loggingT) putBuffer(b *buffer) {
 // the data to the log files. If a trace location is set, stack traces
 // are added to the entry before marshaling.
 func (l *loggingT) outputLogEntry(s Severity, file string, line int, msg string) {
+	// Set additional details in log entry.
+	now := time.Now()
+	entry := MakeEntry(s, now.UnixNano(), file, line, msg)
+
+	if f, ok := l.interceptor.Load().(InterceptorFn); ok && f != nil {
+		f(entry)
+		return
+	}
+
 	// TODO(tschottdorf): this is a pretty horrible critical section.
 	l.mu.Lock()
 
-	// Set additional details in log entry.
-	now := time.Now()
-	entry := Entry{
-		Severity:  s,
-		Time:      now.UnixNano(),
-		Goroutine: goid.Get(),
-		File:      file,
-		Line:      int64(line),
-		Message:   msg,
-	}
 	// On fatal log, set all stacks.
 	var stacks []byte
 	if s == Severity_FATAL {
@@ -1145,14 +1151,45 @@ func v(level level) bool {
 	return VDepth(level, 1)
 }
 
+// InterceptorFn is the type of function accepted by Intercept().
+type InterceptorFn func(entry Entry)
+
+// Intercept diverts log traffic to the given function `f`. When `f` is not nil,
+// the logging package begins operating at full verbosity (i.e. `V(n) == true`
+// for all `n`) but nothing will be printed to the logs. Instead, `f` is invoked
+// for each log entry.
+//
+// To end log interception, invoke `Intercept()` with `f == nil`. Note that
+// interception does not terminate atomically, that is, the originally supplied
+// callback may still be invoked after a call to `Intercept` with `f == nil`.
+func Intercept(ctx context.Context, f InterceptorFn) {
+	logging.Intercept(ctx, f)
+}
+
+func (l *loggingT) Intercept(ctx context.Context, f InterceptorFn) {
+	// TODO(tschottdorf): restore sanity so that all methods have a *loggingT
+	// receiver.
+	if f != nil {
+		logDepth(ctx, 0, Severity_WARNING, "log traffic is now intercepted; log files will be incomplete", nil)
+	}
+	l.interceptor.Store(f) // intentionally also when f == nil
+	if f == nil {
+		logDepth(ctx, 0, Severity_INFO, "log interception is now stopped; normal logging resumes", nil)
+	}
+}
+
 // VDepth reports whether verbosity at the call site is at least the requested
 // level.
 func VDepth(level level, depth int) bool {
 	// This function tries hard to be cheap unless there's work to do.
-	// The fast path is two atomic loads and compares.
+	// The fast path is three atomic loads and compares.
 
 	// Here is a cheap but safe test to see if V logging is enabled globally.
 	if logging.verbosity.get() >= level {
+		return true
+	}
+
+	if f, ok := logging.interceptor.Load().(InterceptorFn); ok && f != nil {
 		return true
 	}
 
--- a/pkg/acceptance/debug_remote_test.go
+++ b/pkg/acceptance/debug_remote_test.go
@@ -75,6 +75,7 @@ func testDebugRemote(t *testing.T) {
 				"/debug/requests",
 				"/debug/range?id=1",
 				"/debug/certificates",
+				"/debug/logspy?duration=1ns",
 			} {
 				t.Run(url, func(t *testing.T) {
 					resp, err := cluster.HTTPClient.Get(l.URL(ctx, 0) + url)
--- a/pkg/server/debug_logspy.go
+++ b/pkg/server/debug_logspy.go
@@ -1,0 +1,237 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strconv"
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/cockroachdb/cockroach/pkg/util/caller"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
+	"github.com/pkg/errors"
+)
+
+// regexpAsString wraps a *regexp.Regexp for better printing and
+// JSON unmarshaling.
+type regexpAsString struct {
+	re *regexp.Regexp
+}
+
+func (r regexpAsString) String() string {
+	if r.re == nil {
+		return ".*"
+	}
+	return r.re.String()
+}
+
+func (r *regexpAsString) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+	var err error
+	(*r).re, err = regexp.Compile(s)
+	return err
+}
+
+// intAsString wraps an int that can be populated from a JSON string.
+type intAsString int
+
+func (i *intAsString) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+	var err error
+	*(*int)(i), err = strconv.Atoi(s)
+	return err
+}
+
+// durationAsString wraps a time.Duration that can be populated from a JSON
+// string.
+type durationAsString time.Duration
+
+func (d *durationAsString) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+	var err error
+	*(*time.Duration)(d), err = time.ParseDuration(s)
+	return err
+}
+
+func (d durationAsString) String() string {
+	return time.Duration(d).String()
+}
+
+const (
+	logSpyDefaultDuration = durationAsString(10 * time.Second)
+	logSpyMaxDuration     = durationAsString(time.Minute)
+	logSpyDefaultCount    = 1000
+	logSpyChanCap         = 4096
+)
+
+type logSpyOptions struct {
+	Count    intAsString
+	Duration durationAsString
+	Grep     regexpAsString
+}
+
+func logSpyOptionsFromValues(values url.Values) (logSpyOptions, error) {
+	rawValues := map[string]string{}
+	for k, vals := range values {
+		if len(vals) > 0 {
+			rawValues[k] = vals[0]
+		}
+	}
+	data, err := json.Marshal(rawValues)
+	if err != nil {
+		return logSpyOptions{}, err
+	}
+	var opts logSpyOptions
+	if err := json.Unmarshal(data, &opts); err != nil {
+		return logSpyOptions{}, err
+	}
+	if opts.Duration == 0 {
+		if opts.Count > 0 {
+			opts.Duration = logSpyMaxDuration
+		} else {
+			opts.Duration = logSpyDefaultDuration
+		}
+	} else if opts.Duration > logSpyMaxDuration {
+		return logSpyOptions{}, errors.Errorf("duration %s is too large (limit is %s)", opts.Duration, logSpyMaxDuration)
+	}
+
+	if opts.Count == 0 {
+		opts.Count = logSpyDefaultCount
+	}
+	return opts, nil
+}
+
+type logSpy struct {
+	active       int32 // updated atomically between 0 and 1
+	setIntercept func(ctx context.Context, f log.InterceptorFn)
+}
+
+func (spy *logSpy) handleDebugLogSpy(w http.ResponseWriter, r *http.Request) {
+	opts, err := logSpyOptionsFromValues(r.URL.Query())
+	if err != nil {
+		http.Error(w, "while parsing options: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if swapped := atomic.CompareAndSwapInt32(&spy.active, 0, 1); !swapped {
+		http.Error(w, "a log interception is already in progress", http.StatusInternalServerError)
+		return
+	}
+	defer atomic.StoreInt32(&spy.active, 0)
+
+	w.Header().Add("Content-type", "text/plain; charset=UTF-8")
+	ctx := r.Context()
+	if err := spy.run(ctx, w, opts); err != nil {
+		// This is likely a broken HTTP connection, so nothing too unexpected.
+		log.Info(ctx, err)
+	}
+}
+
+func (spy *logSpy) run(ctx context.Context, w io.Writer, opts logSpyOptions) (err error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(opts.Duration))
+	defer cancel()
+
+	var countDropped int32
+	defer func() {
+		if err == nil {
+			if dropped := atomic.LoadInt32(&countDropped); dropped > 0 {
+				f, l, _ := caller.Lookup(0)
+				entry := log.MakeEntry(
+					log.Severity_WARNING, timeutil.Now().UnixNano(), f, l,
+					fmt.Sprintf("%d messages were dropped", dropped))
+				err = entry.Format(w) // modify return value
+			}
+		}
+	}()
+
+	// Note that in the code below, this channel is never closed. This is
+	// because we don't know when that is safe. This is sketchy in general but
+	// OK here since we don't have to guarantee that the channel is fully
+	// consumed.
+	entries := make(chan log.Entry, logSpyChanCap)
+
+	{
+		f, l, _ := caller.Lookup(0)
+		entry := log.MakeEntry(
+			log.Severity_INFO, timeutil.Now().UnixNano(), f, l,
+			fmt.Sprintf("intercepting logs with options %+v", opts))
+		entries <- entry
+	}
+
+	spy.setIntercept(ctx, func(entry log.Entry) {
+		if re := opts.Grep.re; re != nil && !re.MatchString(entry.Message) && !re.MatchString(entry.File) {
+			return
+		}
+
+		select {
+		case entries <- entry:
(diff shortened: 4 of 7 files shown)


pkg/server/debug.go
```diff
@@ -150,6 +154,35 @@ func init() {
...
+				return false
+			default:
+				entries <- entry
+				return true
+			}
+		})
+
+		for entry := range entries {
+			fmt.Fprintln(w, entry.Message)
+		}
```
petermattis: There is probably a means of flushing each message.

Rather than explicitly filtering on range ID or other attributes, I'd add a regex which can filter what is returned. Then you can `grep -w r<someRangID>`.

pkg/util/log/clog.go
```diff
@@ -1156,6 +1181,10 @@ func VDepth(level level, depth int) bool {
 		return true
 	}
 
+	if f, ok := logging.interceptor.Load().(InterceptorFn); ok && f != nil {
+		return true
+	}
+
```
petermattis: This only gets verbose logs. Wouldn't you want all logs?

petermattis (commented):
Interesting. I had a thought yesterday that we should and an endpoint that can inspect the Raft debug messages for a particular range. This is somewhat more general.

a-robinson:
> I like this, but before making it a real thing I wanted to gauge the initial
> reaction.

I'm a fan. This would definitely come in handy.

bdarnell:
This seems like a decent idea, but we should probably start running tests with `--v=9` to shake out any panics related to rarely-used log messages.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/server/debug.go, line 160 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KsuRvXiSrFWFP8rY9jW:-KsuRvXiSrFWFP8rY9jX:b-ey7i64) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/server/debug.go#L160)):*
> ```Go
> 		// The explicit header is necessary or (at least Chrome) will try to
> 		// download a gzipped file (Content-type comes back application/x-gzip).
> 		w.Header().Add("Content-type", "text/plain") // FIXME(tschottdorf): check this is correct
> ```

Should be `text/plain; charset=UTF-8` I think. 

---

*[pkg/server/debug.go, line 183 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KstBV8_BVD9aU19IjZI-r1-183:-KsuSXY5EoNtnujNA-9Z:b-z30y4s) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/server/debug.go#L183)):*

There is probably a means of flushing each message.

Rather than explicitly filtering on range ID or other attributes, I'd add a regex which can filter what is returned. Then you can `grep -w r<someRangID>`. 
</blockquote></details>

The means of flushing is to test if the ResponseWriter implements `http.Flusher` and if so to call its `Flush()` method.

---

*[pkg/util/log/clog.go, line 1184 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KsuSv3pauylsZ2_PswO:-KsuSv3pauylsZ2_PswP:b-sntlgo) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/util/log/clog.go#L1184)):*
> ```Go
> 	}
> 
> 	if f, ok := logging.interceptor.Load().(InterceptorFn); ok && f != nil {
> ```

So this will cause all V(N) logs to go to disk too, right? That might be an annoying amount of additional disk I/O. It would be nice to separate this so that the verbose logs only go to the interceptor. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-KsuTIr6BpoXJY_mF6x9:bkky0gp)*

tschottdorf:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/util/log/clog.go, line 1184 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KsuSv3pauylsZ2_PswO:-KsxS6TxNghEDt-W0qOb:b-bpiqdx) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/util/log/clog.go#L1184)):*

So this will cause all V(N) logs to go to disk too, right? That might be an annoying amount of additional disk I/O. It would be nice to separate this so that the verbose logs only go to the interceptor. 
</blockquote></details>

no, in this code when an interceptor is active, it *diverts* the log stream. I set it up this way because

1. it's nice to completely avoid the logging mutex overhead
1. we can't afford to spill V(\infty) into the logs
1. there's currently no way to detect whether a `log.Something` is inside of a `V(something)`; we'd first have to change the API to something like `if log.V(something, func(logger log.Logger) { ... })` which is clunky.

Ideally what we would want is to let messages that would've gone to the log without the interceptor continue to do so. But that needs the above refactor so that logging calls under `V(x)` get associated with that `V(x)`.

---

*[pkg/util/log/clog.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KstBV8_BVD9aU19IjZJ-r1-1187:-KsxSj2m80oKMKH_2u2d:bof7krb) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/util/log/clog.go#L1187)):*

This only gets verbose logs. Wouldn't you want all logs?
</blockquote></details>

This part of the code only makes `V(x)` return true when an interceptor is active. While an interceptor is active, it also diverts *all* log traffic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081)*

tschottdorf:
Ok, looks like we may want to do this. The major question here, I think, is what behavior we want with respect to what goes to the log files while an interceptor is active. The options are

1. exactly what would go to the log file if there weren't an interceptor
1. everything (i.e. whatever the interceptor sees, the log sees)
1. nothing (i.e. the interceptor reroutes)

We currently do 3) because 2) seems likely to lead to contention and potentially a downward spiral + quick demise. 1) is likely what we *really* want but it requires a refactor as illustrated by the snippet below.

```go
if V(5) { // while an interceptor is active, `V(5)` must return true or we never try to log
  log.Info(ctx, "something") // when processing this, have no idea whether it's connected to the `V(5)`
}
```

would have to be (something like)

```go
log.V(ctx, 5, func(ctx context.Context) {
    log.Info(ctx, "something")
})
```

where `V()`, if it decides to log, puts something in the context so that we can connect the call to `Info` to the fact that it's inside of a `V(5)`.

I think that above API isn't bad, actually. But we definitely always allocate a closure. *Maybe* that's ok? Or something like this:

```go
if ctx := log.V(ctx, 5); ctx != nil {
    log.Info(ctx, "something")
}
```

which does roughly the same but avoids the closure. Either way, some token from the `V()` call must be passed to the logging incantations guarded by it.

I think my preference is the second option.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/server/debug.go, line 160 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KsuRvXiSrFWFP8rY9jW:-KsxWja6QJB6E0y40pTf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/server/debug.go#L160)):*

Should be `text/plain; charset=UTF-8` I think. 
</blockquote></details>

Done.

---

*[pkg/server/debug.go, line 183 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KstBV8_BVD9aU19IjZI-r1-183:-KsxWSgRRZs--hZOFOlf:bczp5ez) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/server/debug.go#L183)):*

The means of flushing is to test if the ResponseWriter implements `http.Flusher` and if so to call its `Flush()` method.
</blockquote></details>

Added flushing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-KsxWk_edfR09AxL-j6o:b-ud60c5)*

bdarnell:
OK, I agree that sending nothing to the logs is better than sending everything, and it's not worth the refactoring to change how log.V works right now.

Another interface option we could consider is to add a combined `log.VInfof` (similar to `VEventf`). This would be slightly less efficient than the current `if log.V` interface because its arguments would always be evaluated, but maybe it would be OK outside of hotspots.

tschottdorf:
SGTM. I'll polish up this change as is next week then. As a first pass, will probably just add `?regexp=something&seconds=t` query parameters and call it a day, so I can easily cherry-pick into 1.1. We can iterate on it later.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-Ksxoevx9kGqvsVCryaP:bjrskr5)*

petermattis:
Always allocating a closure is likely a non-starter.  A `log.VInfof` makes sense. For hot-spots you could still do:

```
if log.V(5) {
  log.VInfof(ctx, 5, ...)
}
```

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/util/log/clog.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KstBV8_BVD9aU19IjZJ-r1-1187:-KsxxW0XFjTGkGQL2J6F:bnjgua1) ([raw file](https://github.com/cockroachdb/cockroach/blob/b45ea557e0c80497189021e6b0955d289599ed6f/pkg/util/log/clog.go#L1187)):*

This part of the code only makes `V(x)` return true when an interceptor is active. While an interceptor is active, it also diverts *all* log traffic.
</blockquote></details>

Ah, I should spend more time understanding a change before commenting. Ignore me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-Ksxxs84HIX03AA1X1PK:b-7bhc1b)*

tschottdorf:
Small updates pending, but have code that's worth reviewing. No `V(x)` refactor yet as I'll want to cherry-pick this into 1.1. It can be done separately if we feel strongly that we should have one (I don't).

---

Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-Kt4m7_ItqCRxY0_IbUQ:bvku9nj)*

petermattis:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[pkg/server/debug_logspy.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt81bcgSPgVTjvda0lm:-Kt81bcgSPgVTjvda0ln:b-rxpfrp) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L93)):*
> ```Go
> 	Duration durationAsString
> 	Message  regexpType
> 	FileLine regexpType
> ```

Is it really necessary to have separate regexps for the message and the fileline? I was thinking there would be a single regex and you'd let an entry through if it matched either the message or the fileline. 

---

*[pkg/server/debug_logspy.go, line 180 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt83GcAo1tEB7zefD9a:-Kt83GcAo1tEB7zefD9b:b44fe1k) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L180)):*
> ```Go
> 		case entries <- entry:
> 		default:
> 			// Consumer fell behind, just drop the message.
> ```

This could be unfortunate during debugging. Rather than buffering to a channel, perhaps logspy should buffer to a temporary file on disk. 

---

*[pkg/server/debug_logspy.go, line 192 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt829g_hNsZUx8cgYkd:-Kt829g_hNsZUx8cgYke:b-qa90nu) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L192)):*
> ```Go
> 			strings.TrimSpace(entry.Message),
> 		)
> 	}
> ```

I think we should use `log.Entry.Format` instead. The delta time is potentially interesting, but the above is leaving out a bunch of info such as the absolute time, the severity and the goroutine ID.

---

*[pkg/util/log/clog.go, line 1177 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt82wXCd7GtoYcROhjv:-Kt82wXCd7GtoYcROhjw:bhpf975) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/util/log/clog.go#L1177)):*
> ```Go
> 	// TODO(tschottdorf): restore sanity so that all methods have a *loggingT
> 	// receiver.
> 	logDepth(ctx, 0, Severity_WARNING, "log traffic is now intercepted; log files will be incomplete", nil)
> ```

This should be wrapped in `if f != nil { ... }`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081)*

bdarnell:
Reviewed 4 of 5 files at r2, 1 of 1 files at r3.

---

*[pkg/server/debug.go, line 42 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDHfpsl0TviDWegMN-:-KtDHfpsl0TviDWegMN0:b-jsbs2a) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug.go#L42)):*
> ```Go
> const debugEndpoint = "/debug/"
> 
> const debugLogSpyEndpoint = "/debug/logspy"
> ```

Add this to acceptance/debug_remote_test.go too. 

---

*[pkg/server/debug_logspy.go, line 38 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDJJa2iXFewrsVINd2:-KtDJJa2iXFewrsVINd3:b-twkev5) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L38)):*
> ```Go
> )
> 
> type regexpType struct {
> ```

`regexpAsString` for consistency with the following types? Could also use a brief comment.

---

*[pkg/server/debug_logspy.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt81bcgSPgVTjvda0lm:-KtDLbmAbOQ4-pXvHjLc:b2s4tqz) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L93)):*

Is it really necessary to have separate regexps for the message and the fileline? I was thinking there would be a single regex and you'd let an entry through if it matched either the message or the fileline. 
</blockquote></details>

Or if we're going to separate FileLine from message, why not have separate "file" and "line" arguments so we don't have to format a `file:line` string for every line? (I'd be fine with a single regexp for the final string too)

---

*[pkg/server/debug_logspy.go, line 103 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDKHwZ3hIuMkJbjM0C:-KtDKHwZ3hIuMkJbjM0D:b-au3a7x) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L103)):*
> ```Go
> 		}
> 	}
> 	data, err := json.Marshal(rawValues)
> ```

Seems a little silly to convert our url-encoded parameters to json and parse them again (especially when we're not using standard json types anywhere). It doesn't seem like this saves much work compared to just looping over the `url.Values`. 

---

*[pkg/server/debug_logspy.go, line 180 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt83GcAo1tEB7zefD9a:-KtDLFHhZ2CYla9PjBdJ:bop0tc0) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L180)):*

This could be unfortunate during debugging. Rather than buffering to a channel, perhaps logspy should buffer to a temporary file on disk. 
</blockquote></details>

Or check the `len()` of the channel, and insert a "discarding messages" entry before we overflow. 

---

*[pkg/util/log/clog.go, line 1188 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDFslANEOy89iWk1u7:-KtDFslANEOy89iWk1u8:bpqhtzp) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/util/log/clog.go#L1188)):*
> ```Go
> func VDepth(level level, depth int) bool {
> 	// This function tries hard to be cheap unless there's work to do.
> 	// The fast path is two atomic loads and compares.
> ```

s/two/three/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081)*

tschottdorf:
Review status: 1 of 7 files reviewed at latest revision, 10 unresolved discussions.

---

*[pkg/server/debug.go, line 42 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDHfpsl0TviDWegMN-:-KtEcDEFuMj-dn7bV_KV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug.go#L42)):*

Add this to acceptance/debug_remote_test.go too. 
</blockquote></details>

Done.

---

*[pkg/server/debug_logspy.go, line 38 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDJJa2iXFewrsVINd2:-KtEcZ4rZfLqJzv1VmOw:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L38)):*

`regexpAsString` for consistency with the following types? Could also use a brief comment.
</blockquote></details>

Done.

---

*[pkg/server/debug_logspy.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt81bcgSPgVTjvda0lm:-KtCkvZkLXFsuREUka1G:b-i3kofo) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L93)):*

Or if we're going to separate FileLine from message, why not have separate "file" and "line" arguments so we don't have to format a `file:line` string for every line? (I'd be fine with a single regexp for the final string too)
</blockquote></details>

I'll stick with the current code unless consensus emerges that we should add we should add `(*Entry).FormatString() string` and match on the whole string before sending it (the string) to the channel. Probably converting everything to string is fine, but I'm not sure it buys us that much. It does allow matching for goroutine ID etc though, but then again this is also possible by just grepping the output of a suitably restricted `?grep=.`.

---

*[pkg/server/debug_logspy.go, line 103 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDKHwZ3hIuMkJbjM0C:-KtEcgNdwlkAdrbA6Z1r:b-aqsr0s) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L103)):*

Seems a little silly to convert our url-encoded parameters to json and parse them again (especially when we're not using standard json types anywhere). It doesn't seem like this saves much work compared to just looping over the `url.Values`. 
</blockquote></details>

Yeah. Since I've already written what's here and it's not better or worse than writing everything manually, I'd prefer to leave as is unless you feel strongly.

---

*[pkg/server/debug_logspy.go, line 180 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt83GcAo1tEB7zefD9a:-KtCVDWydF9wVOV8eJI9:bh1fm9j) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L180)):*

Or check the `len()` of the channel, and insert a "discarding messages" entry before we overflow. 
</blockquote></details>

Isn't checking `len()` racy and can end up blocking?. I've changed the code so that whenever we drop messages, the consumer gets a notification at the end of their received log. If we want to do more, let's file something (I don't expect that we'll need it).

---

*[pkg/server/debug_logspy.go, line 192 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt829g_hNsZUx8cgYkd:-KtCkBpBMWrg50IYBB8A:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L192)):*

I think we should use `log.Entry.Format` instead. The delta time is potentially interesting, but the above is leaving out a bunch of info such as the absolute time, the severity and the goroutine ID.
</blockquote></details>

Done.

---

*[pkg/util/log/clog.go, line 1177 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt82wXCd7GtoYcROhjv:-KtCkIeiYccUdOvyVkoy:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/util/log/clog.go#L1177)):*

This should be wrapped in `if f != nil { ... }`.
</blockquote></details>

Done.

---

*[pkg/util/log/clog.go, line 1188 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDFslANEOy89iWk1u7:-KtEdLhJYBV9IKm_Rkt7:b-p9zc3j) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/util/log/clog.go#L1188)):*

s/two/three/
</blockquote></details>

Done. We could get it back down to two if we really wanted to, but probably not worth it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081)*

petermattis:
:lgtm:

---

Review status: 1 of 7 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[pkg/server/debug_logspy.go, line 200 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtHFDAcQCZfyKvY0js6:-KtHFDAcQCZfyKvY0js7:b-mbp8j1) ([raw file](https://github.com/cockroachdb/cockroach/blob/872e2fed76c998e75e7f8259fcb87efb3f73d322/pkg/server/debug_logspy.go#L200)):*
> ```Go
> 				fmt.Sprintf("%d messages were dropped", dropped))
> 			if err := entry.Format(w); err != nil {
> 				log.Warning(ctx, err)
> ```

This will call back to the interceptor which doesn't seem good.

---

*[pkg/server/debug_logspy.go, line 214 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtHFJin1uarOGF3e_qN:-KtHFJioCIKRMrgtM9qP:b9alpzx) ([raw file](https://github.com/cockroachdb/cockroach/blob/872e2fed76c998e75e7f8259fcb87efb3f73d322/pkg/server/debug_logspy.go#L214)):*
> ```Go
> 		case entry := <-entries:
> 			if err := entry.Format(w); err != nil {
> 				log.Warning(ctx, err)
> ```

Ditto about this calling back to the interceptor. Perhaps return the error here and log it in the caller.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-KtHFk7zjS5pgdSEJgmu:bnfp4nl)*

tschottdorf:
TFTRs! I added a test about the error case returns too.

---

Review status: 1 of 7 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/server/debug_logspy.go, line 200 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtHFDAcQCZfyKvY0js6:-KtHWdyJqU7eRMu8UbuM:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/872e2fed76c998e75e7f8259fcb87efb3f73d322/pkg/server/debug_logspy.go#L200)):*

This will call back to the interceptor which doesn't seem good.
</blockquote></details>

Done.

---

*[pkg/server/debug_logspy.go, line 214 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtHFJin1uarOGF3e_qN:-KtHWcaFN36I9A6C_6GJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/872e2fed76c998e75e7f8259fcb87efb3f73d322/pkg/server/debug_logspy.go#L214)):*

Ditto about this calling back to the interceptor. Perhaps return the error here and log it in the caller.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-KtHWf3JygwmbSd7PF50:b74qq8i)*

bdarnell:
:lgtm:

---

Reviewed 4 of 6 files at r4, 2 of 2 files at r5.

---

*[pkg/server/debug_logspy.go, line 103 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-KtDKHwZ3hIuMkJbjM0C:-KtHsIHFEZr40cQGr1Hz:b-sn6gr9) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L103)):*

Yeah. Since I've already written what's here and it's not better or worse than writing everything manually, I'd prefer to leave as is unless you feel strongly.
</blockquote></details>

I don't feel strongly; leaving it as-is is fine.

---

*[pkg/server/debug_logspy.go, line 180 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-Kt83GcAo1tEB7zefD9a:-KtHslsro2TDDhjC9uE0:b-lygszs) ([raw file](https://github.com/cockroachdb/cockroach/blob/d78db838b7e10cbf0d9e084d9453cd460af3ead9/pkg/server/debug_logspy.go#L180)):*

Isn't checking `len()` racy and can end up blocking?. I've changed the code so that whenever we drop messages, the consumer gets a notification at the end of their received log. If we want to do more, let's file something (I don't expect that we'll need it).
</blockquote></details>

`len()` is racy, but could be made non-racy if the interceptor were called under the log mutex. Tracking a count is good too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18081#-:-KtHtHMoA5bDC3YYxfyS:bnfp4nl)*

PR #18083, opened 2017-08-31
sqlbase: improve the NON-NULL column panic message

vivekmenezes:


jordanlewis:
:lgtm: good idea

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18083#-:-KstIyEZD3Nd4cax1wWA:bnaseej)*

RaduBerinde:
Do you think we should dump the table and index descriptors as well? Would be interesting to see if there are ongoing mutations

PR #17378, opened 2017-08-02
sql: ensure that a table cannot be read before its creation time

vivekmenezes:

PR #18086, opened 2017-08-31
sqlccl: set CSV table desc creation timestamp

madelynnblue:


vivekmenezes:
More accurately we should be using the timestamp used in the backup representation for each K:V as the ModificationTime for the table descriptor. Thoughts?

mjibson:
We are. That walltime is plumbed through the entire creation process and is set as the backup descriptor time, KV MVCC time, and now the create table time.

danhhz:
What's the context here?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18086#-:-KstkUi7koYmfek67AEn:b-gibgt8)*

mjibson:
Added a link to the original issue and removed the TODO referencing this.

mjibson:
https://github.com/cockroachdb/cockroach/pull/17378#issuecomment-320968079 contains just a bit more about this. Although it's unclear to us exactly why we need to set a timestamp (besides it being part of the function call).

vivekmenezes:
LGTM

danhhz:
> #17378 (comment) contains just a bit more about this. Although it's unclear to us exactly why we need to set a timestamp (besides it being part of the function call).

👍

PR #18094, opened 2017-08-31
sql: add support for subqueries with ANY/SOME/ALL operations

### LINKED ISSUE #17662: sql: ANY/ALL with subqueries
We currently support `ANY` and `ALL` with ARRAYs:

```
root@:26257/> SELECT 3 = ALL ARRAY[3,3,3];
+------------------------+
| 3 = ALL ARRAY[3, 3, 3] |
+------------------------+
| true                   |
+------------------------+
(1 row)

Time: 1.991091ms

root@:26257/> SELECT 3 = ALL ARRAY[3, 3, 4];
+------------------------+
| 3 = ALL ARRAY[3, 3, 4] |
+------------------------+
| false                  |
+------------------------+
(1 row)

Time: 532.323µs
```

We should also support this with (uncorrelated) subqueries:

```
root@:26257/> SELECT 3 = ALL (SELECT 3);
pq: unsupported comparison operator: 3 = ALL (SELECT 3): op ALL array requires array on right side
root@:26257/> SELECT 3 = ANY (SELECT 3);
pq: unsupported comparison operator: 3 = ANY (SELECT 3): op ANY array requires array on right side
```

cc @knz @nvanbenschoten

richardwu:
Previously we only supported arrays for `ANY` and `ALL` operations:
```
SELECT 1 = ANY(ARRAY[1, 2, 3]);

SELECT 1 = ANY(ARRAY[1, 2, 3]);
+--------------------------+
| 1 = ANY (ARRAY[1, 2, 3]) |
+--------------------------+
| true                     |
+--------------------------+
(1 row)

Time: 508.392µs
```

This PR allows us to use subqueries as well in the RHS of the predicate:
```
SELECT 1 = ANY(SELECT * FROM generate_series(1,3));
+--------------------------------+
| 1 = ANY (SELECT * FROM generate_series(1, 3)) |
+--------------------------------+
| true                           |
+--------------------------------+
(1 row)

Time: 648.021µs
```

Fixes #17662


pkg/sql/parser/eval.go
```diff
@@ -1750,36 +1756,27 @@ func evalArrayCmp(
...
-		if res {
-			anyTrue = true
-		} else {
-			allTrue = false
-		}
-	}
-
-	if all {
-		if !allTrue {
+		if res && any {
```
richardwu: I optimized this loop through each `Datum` in our slice of `Datum`s such that it short circuits/early terminates when:

1. `ANY`: result is encountered that results in `true` (formally, there exists a right such that the predicate evaluates to `true`, "any" is true, done)
2. `ALL`: result is encountered that results in `false` (there exists a `right` such that the predicate evaluates to `false`, not all are true, we are finished)

This results in statements such as this
```
SELECT 1 = ALL(SELECT * FROM generate_series(1,1000000))
```
to terminate after evaluating `1 = 2` instead of iterating until the end.

justinj:
This looks great so far! I left some comments, but @knz should probably take a look as well!

---

Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/suboperators, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksto8nelysLPTT8pZj6:-Ksto8nelysLPTT8pZj7:bh0pzu6) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/logictest/testdata/logic_test/suboperators#L1)):*
> ```
> # LogicTest: default parallel-stmts distsql
> ```

So thorough :thumbsup: could we also have a couple tests involving `NULL`s?

---

*[pkg/sql/parser/eval.go, line 1742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstll0tlxy1wL8DO0pR:-Kstll0tlxy1wL8DO0pS:bilgo2v) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L1742)):*
> ```Go
> 	any := !all
> 	sawNull := false
> 	for _, elem := range right {
> ```

Perhaps we could just branch once at the top for `any` vs. `all` and then have a different `for` within each branch (maybe pulling each into helper functions)? It might simplify this a little and remove the need for the explanatory short-circuiting comments.

---

*[pkg/sql/parser/eval.go, line 2726 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KstkZCl3hI37bacSbQH:-KstkZCl3hI37bacSbQI:b-f8s3oi) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L2726)):*
> ```Go
> 	if op.hasSubOperator() {
> 		// Branch to different helper comparison functions
> 		// depending on whether a subquery or an array follows
> ```

nit: we generally put periods at the end of comments (also applies to some other comments here), also, is this comment accurate? we don't seem to branch to different functions, just different ways of extracting the `Datums`

---

*[pkg/sql/parser/eval.go, line 2733 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstqpje8iOl_TwmUNSh:-KstkcAUnACoRnXRDaPf:bdup3cg) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L2733)):*
> ```Go
> 			datums = array.Array
> 		}
> 		// Type checking in parser ensures rightType is either TypeTuple or TypeArray
> ```

consider adding a `panic` in an `else` clause with this as the error message instead? if this is violated we probably don't want to continue.

---

*[pkg/sql/parser/type_check.go, line 1011 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KstrCxTSYxnLa3oMC5i:-KstrCxTSYxnLa3oMC5j:b-vkq92b) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/type_check.go#L1011)):*
> ```Go
> 
> 		rightUnwrapped := UnwrapType(rightReturn)
> 		switch rightUnwrapped.(type) {
> ```

if you replace this with a [type switch](https://tour.golang.org/methods/16) you won't have to do the explicit casts below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-:-KstpcmNIcujBsXk-8jV:bax9ew)*

knz:
I'll let you process and iterate with Justin first. Once you've settled together I'll do a round too. I don't think it's useful to have 2+ people reviewing at the same time for now. Let me know.

Please however do respect the rule 1 concern = 1 commit, see my comment below.

---

Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/eval.go, line 1759 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstg5fnyM0vLESsbXtO-r1-1759:-KstxTgf_zFu__PBJG-b:bnxirb5) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L1759)):*

I optimized this loop through each `Datum` in our slice of `Datum`s such that it short circuits/early terminates when:

1. `ANY`: result is encountered where `res = true` (formally, there exists a right such that the predicate evaluates to `true`, "any" is true)
2. `ALL`: result is encountered where `res = false` (there exists a `right` such that the predicate evaluates to `false`, not all are true)

This results in statements such as this
```
SELECT 1 = ALL(SELECT * FROM generate_series(1,1000000))
```
to terminate after evaluating `1 = 2` instead of iterating until the end.
</blockquote></details>

Please split this change into a different commit. Add tests that involve NULL in the subquery.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-:-Ksty2gc0mCoSevedAcg:b-emcb89)*

richardwu:
Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/suboperators, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksto8nelysLPTT8pZj6:-KsuJmjWcozrKgjGIE9E:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/logictest/testdata/logic_test/suboperators#L1)):*

So thorough :thumbsup: could we also have a couple tests involving `NULL`s?
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval.go, line 1742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstll0tlxy1wL8DO0pR:-KsuJp4H1px2wCID3aEc:b-1pbv9f) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L1742)):*

Perhaps we could just branch once at the top for `any` vs. `all` and then have a different `for` within each branch (maybe pulling each into helper functions)? It might simplify this a little and remove the need for the explanatory short-circuiting comments.
</blockquote></details>

As per @knz's comment, I will move this refactor to a separate PR.

---

*[pkg/sql/parser/eval.go, line 2726 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KstkZCl3hI37bacSbQH:-KsuJzV4RiWF_2o0uByW:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L2726)):*

nit: we generally put periods at the end of comments (also applies to some other comments here), also, is this comment accurate? we don't seem to branch to different functions, just different ways of extracting the `Datums`
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval.go, line 2733 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstqpje8iOl_TwmUNSh:-KsuJ9eOpUZbQQxGCeaV:bfk3exq) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L2733)):*

consider adding a `panic` in an `else` clause with this as the error message instead? if this is violated we probably don't want to continue.
</blockquote></details>

I decided to propagate an internal PG error so we don't crash the instance.

---

*[pkg/sql/parser/type_check.go, line 1011 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KstrCxTSYxnLa3oMC5i:-KsuJ5kPp449wiPG0FLO:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/type_check.go#L1011)):*

if you replace this with a [type switch](https://tour.golang.org/methods/16) you won't have to do the explicit casts below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

justinj:
This all looks good to me (modulo a squash and making sure the final commit name has the `sql:` prefix)! All yours @knz.

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/parser/eval.go, line 1742 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstll0tlxy1wL8DO0pR:-KsuL5Pcb6Y79giezA86:bdpste2) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L1742)):*

As per @knz's comment, I will move this refactor to a separate PR.
</blockquote></details>

sgtm :thumbsup: 

---

*[pkg/sql/parser/eval.go, line 2733 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstqpje8iOl_TwmUNSh:-KsuM9CBZxshobWGOx--:bszhcem) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L2733)):*

I decided to propagate an internal PG error so we don't crash the instance.
</blockquote></details>

sounds reasonable. I don't think we log `CodeInternalError`s to sentry like we do panics, but we probably should.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-:-KsuLeSb3rO5LpxAjbnu:b-yxlq0d)*

knz:
Reviewed 5 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.

---

*[pkg/sql/parser/eval.go, line 2733 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kstqpje8iOl_TwmUNSh:-KsuWA0Z7mqhJSb1vMiJ:b-3wl1q5) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/eval.go#L2733)):*

sounds reasonable. I don't think we log `CodeInternalError`s to sentry like we do panics, but we probably should.
</blockquote></details>

I think we do actually -- or if we don't there's an issue already for doing so, scheduled for 1.2. all good.

---

*[pkg/sql/parser/type_check.go, line 998 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsuWJBkfKlLQjMn-QAc:-KsuWJBkfKlLQjMn-QAd:b-kj4xr2) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/type_check.go#L998)):*
> ```Go
> 		cmpTypeLeft = leftTyped.ResolvedType()
> 
> 		// Try to type the right expression as an Array of the left's type.
> ```

"as an array" - capital not needed

---

*[pkg/sql/parser/type_check.go, line 1017 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsuWs7Hx33vyJ1vtDiP:-KsuWs7Hx33vyJ1vtDiQ:b-g6d2ag) ([raw file](https://github.com/cockroachdb/cockroach/blob/6bed6596421e7af79253c576398f476bae52ca26/pkg/sql/parser/type_check.go#L1017)):*
> ```Go
> 			// Subqueries are expected to return 1 column of values
> 			// (see planner.analyzeExpr in analyze.go).
> 			cmpTypeRight = rightUnwrapped[0]
> ```

Woah, hold your horses. We can produce tuples in SQL in other ways than using queries!
What do you think should happen with `1 = ANY(3,1,3)` ? Right now your code says "false" but I think it should be "true". Moreover, we can generate zero-tuples with "`ROW()`", e.g. `1 = ANY(ROW())`, I fear you're getting  "array out of bounds" panic here.

I don't think you'll be getting away without special casing subqueries out of other kinds of tuples.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

justinj:

---

*[pkg/sql/parser/type_check.go, line 1017 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsuWs7Hx33vyJ1vtDiP:-KsuZp8uhcl3wCTCPwH5:b-a701ri) ([raw file](https://github.com/cockroachdb/cockroach/blob/6bed6596421e7af79253c576398f476bae52ca26/pkg/sql/parser/type_check.go#L1017)):*

Woah, hold your horses. We can produce tuples in SQL in other ways than using queries!
What do you think should happen with `1 = ANY(3,1,3)` ? Right now your code says "false" but I think it should be "true". Moreover, we can generate zero-tuples with "`ROW()`", e.g. `1 = ANY(ROW())`, I fear you're getting  "array out of bounds" panic here.

I don't think you'll be getting away without special casing subqueries out of other kinds of tuples.
</blockquote></details>

Ah, you're right, sorry for missing that @richardwu!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
Review status: 0 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/parser/type_check.go, line 998 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsuWJBkfKlLQjMn-QAc:-Ksy_oY04KUzgag5_21-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f4fea7b7c531b48c3e5beadca6295b724c91ee6/pkg/sql/parser/type_check.go#L998)):*

"as an array" - capital not needed
</blockquote></details>

Done.

---

*[pkg/sql/parser/type_check.go, line 1017 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsuWs7Hx33vyJ1vtDiP:-Ksy_vlfHYjCCnlyMj4P:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6bed6596421e7af79253c576398f476bae52ca26/pkg/sql/parser/type_check.go#L1017)):*

Ah, you're right, sorry for missing that @richardwu!
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
So I ran into a few failed tests that resulted from our use of type checking on `sql.subquery`.

The "lifetime" of the new `parser.SubqueryPlaceholder` that wraps `sql.subquery` begins at `replaceSubqueries` and lives until `TypeCheck` (where `TypeCheck` is delegated to the nested `sql.subquery) in `analyzeExpr` https://github.com/cockroachdb/cockroach/blob/1ae8e374f9c8bbeae2e41cfb4375b2c5a7b42ccd/pkg/sql/analyze.go#L1642#L1667.

I tried my best to go through all the instances of type checking for `sql.subquery` during the lifespan of `parser.SubqueryPlaceholder` https://github.com/cockroachdb/cockroach/pull/18094/files#diff-06230dfb99f17a44b32848b107b7fd56R310 and https://github.com/cockroachdb/cockroach/pull/18094/files#diff-06230dfb99f17a44b32848b107b7fd56R481, but I think it ought be best if someone with more knowledge about this part of code comment on this approach.

knz:
Reviewed 1 of 6 files at r6, 1 of 1 files at r7, 1 of 1 files at r10, 5 of 5 files at r11.

---

*[pkg/sql/subquery.go, line 333 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsylLP6fWo853m9jscI:-KsylLP6fWo853m9jscJ:b-ukfou4) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/subquery.go#L333)):*
> ```Go
> 	switch expr.(type) {
> 	// We already replaced this one; do nothing.
> 	case *subquery, *parser.SubqueryPlaceholder:
> ```

I don't think you can observe `*subquery` any more.

---

*[pkg/sql/update.go, line 310 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsyjtlrM32mdgzxDd2U:-Ksyjtlt1f42C_KQLOkf:b-iuhszw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/update.go#L310)):*
> ```Go
> 					currentUpdateIdx++
> 				}
> 			case *subquery, *parser.SubqueryPlaceholder:
> ```

I think because of your change it becomes impossible to encounter a naked `*subquery` here. Can you confirm?

---

*[pkg/sql/update.go, line 481 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsyjywOS9u0TQjT572L:-KsyjywOS9u0TQjT572M:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/update.go#L481)):*
> ```Go
> 			n := -1
> 			switch t := expr.Expr.(type) {
> 			case *subquery, *parser.SubqueryPlaceholder:
> ```

ditto

---

*[pkg/sql/logictest/testdata/logic_test/suboperators, line 327 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksyk3C-8T1NsOZA8d2x:-Ksyk3C-8T1NsOZA8d2y:bqxhc6z) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/logictest/testdata/logic_test/suboperators#L327)):*
> ```
> SELECT 1 = ANY ROW()
> 
> query error unsupported comparison operator: 1 = ANY \(1, 2, 3\)
> ```

I fail to agree here -- it seems to me that we can compare a value with the members of a tuple, when the tuple values match in type.

---

*[pkg/sql/parser/expr.go, line 737 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsykZH_uXCVHlOwrAt3:-KsykZH_uXCVHlOwrAt4:bmq6k0b) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/expr.go#L737)):*
> ```Go
> }
> 
> func (s *SubqueryPlaceholder) String() string { return s.Sq.String() }
> ```

This String method does not belong here. Place it at the end of `expr.go` with the others.

---

*[pkg/sql/parser/expr.go, line 749 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsykJlG5jwwCSgfiqEg:-KsykJlG5jwwCSgfiqEh:b-xee43n) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/expr.go#L749)):*
> ```Go
> // TypeCheck implements the Expr interface.
> func (s *SubqueryPlaceholder) TypeCheck(ctx *SemaContext, desired Type) (TypedExpr, error) {
> 	return s.Sq.TypeCheck(ctx, desired)
> ```

Unless you do the deferred construction thing I explained earlier, this method should do `return s, nil`.

The entire Placeholder must implement the Expr interface, which in turn means it also needs an Eval and ResolvedType methods. These do recurse into `Sq`, to be provided by `sql.subquery`. 

---

*[pkg/sql/parser/type_check.go, line 1019 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksyl5fDw7pUVgZwYKIE:-Ksyl5fDw7pUVgZwYKIF:brtic61) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/type_check.go#L1019)):*
> ```Go
> 			// Subqueries are expected to return 1 column of values
> 			// (see planner.analyzeExpr in analyze.go).
> 			if _, ok := right.(*SubqueryPlaceholder); !ok {
> ```

You need two separate pieces of logic, one for tuples generated by non-subqueries (where the RHS is the entire tuple) and one for tuples geneerated by subqueries.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

knz:

---

*[pkg/sql/parser/expr.go, line 734 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksym7-BUGaOxVjZpUru:-Ksym7-CcZh4LstqEEFb:bn6wj2b) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/expr.go#L734)):*
> ```Go
> // during type checking.
> type SubqueryPlaceholder struct {
> 	Sq Expr
> ```

`Sq TypedExpr` btw, the subquery already has its types.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
*[pkg/sql/subquery.go, line 333 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsylLP6fWo853m9jscI:-KszW1r_fIU3l8X5DPEP:bmxiqxd) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/subquery.go#L333)):*

I don't think you can observe `*subquery` any more.
</blockquote></details>

I have decided to not introduce the SubqueryPlaceholder in this PR since it is not strictly required (since we are supporting tuples with ANY/ALL, we do not need to special case subqueries as they evaluate to tuples). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
Review status: 2 of 6 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[pkg/sql/parser/type_check.go, line 1019 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksyl5fDw7pUVgZwYKIE:-KszWMxHXUochMpFVknd:bdnpd1d) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/type_check.go#L1019)):*

You need two separate pieces of logic, one for tuples generated by non-subqueries (where the RHS is the entire tuple) and one for tuples geneerated by subqueries.
</blockquote></details>

Ditto above comment

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
Review status: 2 of 6 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[pkg/sql/update.go, line 310 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsyjtlrM32mdgzxDd2U:-KszWa6pgxrCwFUlm-kC:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/update.go#L310)):*

I think because of your change it becomes impossible to encounter a naked `*subquery` here. Can you confirm?
</blockquote></details>

Done.

---

*[pkg/sql/update.go, line 481 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsyjywOS9u0TQjT572L:-KszWYgCxQN6OLWAjOaP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/update.go#L481)):*

ditto
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr.go, line 734 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Ksym7-BUGaOxVjZpUru:-KszWZRUitDbB0uQfaEj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/expr.go#L734)):*

`Sq TypedExpr` btw, the subquery already has its types.
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr.go, line 737 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsykZH_uXCVHlOwrAt3:-KszWZyr-nMR9-o-YP2l:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/expr.go#L737)):*

This String method does not belong here. Place it at the end of `expr.go` with the others.
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr.go, line 749 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KsykJlG5jwwCSgfiqEg:-KszW_VTW6N7pdROLOP3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2af1a0248e59ad513a8fb856a1fbaff32cc2332/pkg/sql/parser/expr.go#L749)):*

Unless you do the deferred construction thing I explained earlier, this method should do `return s, nil`.

The entire Placeholder must implement the Expr interface, which in turn means it also needs an Eval and ResolvedType methods. These do recurse into `Sq`, to be provided by `sql.subquery`. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

knz:
It's nice to see how unifying the tuple and subquery cases also simplify the code. I like where this is headed at last.

One last bug to correct and you should be good to go.

---

Reviewed 6 of 6 files at r12.

---

*[pkg/sql/parser/type_check.go, line 1000 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3bTUxskVuc1Orq_ut:-Kt3bTUymW62wYfCd5pP:b-nu2541) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1000)):*
> ```Go
> 
> 		// Try to type the right expression as an array of the left's type.
> 		// If right is an sql.subquery Expr, it should already be typed.
> ```

Add:
`// TODO(richardwu): If right is a subquery, we should really propagate the left type as a desired type for the result column.`

---

*[pkg/sql/parser/type_check.go, line 1029 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3dEQawKw2PgHNgpGq:-Kt3dEQbzvOj2RP-xT4E:b61l5kr) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1029)):*
> ```Go
> 	var ok bool
> 	for _, cmpTypeRight := range cmpTypesRight {
> 		if fn, ok = ops.lookupImpl(cmpTypeLeft, cmpTypeRight); !ok {
> ```

This is incorrect, and the following statement will demonstrate it is incorrect by crashing the server:
`SELECT 4 = any(1, '2', 3);`

What you need to do:
- revert back `cmpTypesRight []Type` to `cmpTypeRight Type` above
- above in the TTuple case, check that *all the tuple element types are identical*, and save that (unique) type into `cmpTypeRight`
- restore the previous code that was looking up `fn`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-:-Kt3dj6ziBRE1dwUVHfB:byvp9e3)*

knz:

---

*[pkg/sql/parser/type_check.go, line 1000 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3bTUxskVuc1Orq_ut:-Kt3dzpbvgoSMrcum2rb:b-mucx19) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1000)):*

Add:
`// TODO(richardwu): If right is a subquery, we should really propagate the left type as a desired type for the result column.`
</blockquote></details>

Also, change this to use `typeCheckSameTypedTupleExprs` if the right side is a tuple, so as to maximize the chance the tuple has the right element types.

---

*[pkg/sql/parser/type_check.go, line 1029 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3dEQawKw2PgHNgpGq:-Kt3dtADrodYgdVamGxy:b-404nbq) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1029)):*

This is incorrect, and the following statement will demonstrate it is incorrect by crashing the server:
`SELECT 4 = any(1, '2', 3);`

What you need to do:
- revert back `cmpTypesRight []Type` to `cmpTypeRight Type` above
- above in the TTuple case, check that *all the tuple element types are identical*, and save that (unique) type into `cmpTypeRight`
- restore the previous code that was looking up `fn`
</blockquote></details>

Also, add a test -- the example I gave should properly return an error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
*[pkg/sql/parser/type_check.go, line 1029 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3dEQawKw2PgHNgpGq:-Kt3qqlm_Egfk3ynnLCR:b-qpr8m0) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1029)):*

Also, add a test -- the example I gave should properly return an error.
</blockquote></details>

So I played around with the current implementation and it does correctly return an error for the statement `SELECT 4 = any(1, '2', 3)` (it notices it can't compare an `int` with a `string`).

If I were to check all tuple element types to be identical, then the following statement would return a type mismatch error `SELECT 4 = any(1, 2.5)`. This will be inconsistent with how we interpret tuples with `IN`, that is

```
SELECT 1 IN (1, 2.5);
+------------------+
| 1 IN (1, 2.5) |
+------------------+
| true             |
+------------------+
(1 row)

Time: 615.825µs
```

You can check out the logic test I've added with your example https://github.com/cockroachdb/cockroach/pull/18094/files#diff-7021787014e75d2e5d4600ab91b6475aR353.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
*[pkg/sql/parser/type_check.go, line 1000 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3bTUxskVuc1Orq_ut:-Kt4-5LkgK3XmIx-b17o:brdqrs8) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1000)):*

Also, change this to use `typeCheckSameTypedTupleExprs` if the right side is a tuple, so as to maximize the chance the tuple has the right element types.
</blockquote></details>

So I did some investigation with how our `TypeCheck` works, and it looks like `TypeCheck` does recurse down to `Tuple` and `Array`. In fact, `Tuple.TypeCheck` does type check every element of the tuple https://github.com/cockroachdb/cockroach/blob/16971acd9046d8e07a02d544a0b3262e7a2463a9/pkg/sql/parser/type_check.go#L737L748.

I've updated the statement to simply type check `cmpTypeLeft` (instead of `TArray{cmpTypeLeft}` which was wholly unnecessary) https://github.com/cockroachdb/cockroach/pull/18094/files#diff-f5c8f9ab30e6b17710413b03d2169349R1003.

I've verified we correctly type check expressions in the logs for both arrays and tuples

```
SELECT 1 = ANY (ARRAY[1,2] || 3)

// Before typing 
W170902 19:49:14.446858 311 sql/parser/type_check.go:1005  &parser.ParenExpr{Expr:(*parser.BinaryExpr)(0xc42064e090), typeAnnotation:parser.typeAnnotation{typ:parser.Type(nil)}}
// After typing
W170902 19:49:14.446942 311 sql/parser/type_check.go:1011  &parser.ParenExpr{Expr:(*parser.BinaryExpr)(0xc42064e090), typeAnnotation:parser.typeAnnotation{typ:parser.TArray{Typ:parser.tInt{}}}}

SELECT 1 = ANY((1, 2, 3.5))

// Before typing
W170902 19:49:48.748120 311 sql/parser/type_check.go:1005  &parser.ParenExpr{Expr:(*parser.Tuple)(0xc4208358c0), typeAnnotation:parser.typeAnnotation{typ:parser.Type(nil)}}
// After typing
W170902 19:49:48.748160 311 sql/parser/type_check.go:1011  &parser.ParenExpr{Expr:(*parser.Tuple)(0xc4208358c0), typeAnnotation:parser.typeAnnotation{typ:parser.TTuple{parser.tInt{}, parser.tInt{}, parser.tDecimal{}}}}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

knz:
Review status: 4 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/type_check.go, line 1029 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3dEQawKw2PgHNgpGq:-KtC5Zs_hX6D_hLzXNKC:b-gra098) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1029)):*

So I played around with the current implementation and it does correctly return an error for the statement `SELECT 4 = any(1, '2', 3)` (it notices it can't compare an `int` with a `string`).

If I were to check all tuple element types to be identical, then the following statement would return a type mismatch error `SELECT 4 = any(1, 2.5)`. This will be inconsistent with how we interpret tuples with `IN`, that is

```
SELECT 1 IN (1, 2.5);
+------------------+
| 1 IN (1, 2.5) |
+------------------+
| true             |
+------------------+
(1 row)

Time: 615.825µs
```

You can check out the logic test I've added with your example https://github.com/cockroachdb/cockroach/pull/18094/files#diff-7021787014e75d2e5d4600ab91b6475aR353.
</blockquote></details>

If you use `typeCheckSameTypedTupleExprs` you can use it to propagate the type on the left to all the elements on the right. It's ok if `1 = any(1, 2.5)` fails, but `1::decimal = any(1, 2.5)` should work.

Also check with Jordan what is going on with placeholders. Right now the tests fail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

richardwu:
*[pkg/sql/parser/type_check.go, line 1029 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-Kt3dEQawKw2PgHNgpGq:-KtI8OkTBQ5j8L0UrFhp:b-3z2h5l) ([raw file](https://github.com/cockroachdb/cockroach/blob/0e3424bc379ac39efb1ba0b91dcfb3eae752c7b5/pkg/sql/parser/type_check.go#L1029)):*

If you use `typeCheckSameTypedTupleExprs` you can use it to propagate the type on the left to all the elements on the right. It's ok if `1 = any(1, 2.5)` fails, but `1::decimal = any(1, 2.5)` should work.

Also check with Jordan what is going on with placeholders. Right now the tests fail.
</blockquote></details>

As discussed offline, `typeCheckSameTypedTupleExprs` is useful for multiple `Tuple` expressions, but does not check (let alone require) individual elements in a single `Tuple` expression.

I've written a helper function `typeCheckAndRequireTupleElems` (https://github.com/cockroachdb/cockroach/pull/18094/files#diff-f5c8f9ab30e6b17710413b03d2169349R912) to check and **require** individual elements of a tuple (using the existing `typeCheckAndRequire` helper function).

I've also updated logic tests for the correct behavior in @knz's comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094)*

knz:
Ok :lgtm: 
Well done!

---

Reviewed 1 of 5 files at r13, 1 of 1 files at r17, 3 of 3 files at r18.

---

*[pkg/sql/parser/type_check.go, line 917 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-KtMVjhEx2NSDknFoefB:-KtMVjhFUGNEuOKhUAtZ:bc5miw8) ([raw file](https://github.com/cockroachdb/cockroach/blob/25ca38f56bf18f7ae2f8f8b37073aea06e6aef0d/pkg/sql/parser/type_check.go#L917)):*
> ```Go
> 	for i, subExpr := range tuple.Exprs {
> 		// Require that the sub expression is equivalent (or may be inferred) to the required type.
> 		typedExpr, err := typeCheckAndRequire(ctx, subExpr, required, "TUPLE element")
> ```

lowercase "tuple"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18094#-:-KtMWa1W4r3qaplDI0Ii:b-5eci72)*

PR #18098, opened 2017-08-31
storage: account for sideloaded data in Replica.mu.raftLogSize

### LINKED ISSUE #18077: sideloading: 2x disk usage after a restore
When restoring X bytes of data, you need 2X (plus some headroom) of available space on the cluster. This is because (until the new tables get some traffic) the raft log for the new ranges is not truncated, so an entire copy of the restored data ends up sitting in raft logs.

We fixed this once by making the raft log truncation work by both number of entries and total size, but it seems there's a regression when it comes to sideloading. Perhaps the truncation threshold logic doesn't take into account the fully hydrated command size?

petermattis:
Fixes #18077

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -3273,15 +3273,18 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 	if len(rd.Entries) > 0 {
 		// All of the entries are appended to distinct keys, returning a new
 		// last index.
-		thinEntries, err := r.maybeSideloadEntriesRaftMuLocked(ctx, rd.Entries)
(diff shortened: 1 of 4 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18098#-:-KsuTnTv7EUg4IVzj86a:bnfp4nl)*

petermattis:
@cuongdo Can you test another restore and see if this reduces the resulting data size multiple?

cuongdo:
Yes, doing that now.

cuongdo:
👍 

The RESTOREd data is now 17 GB, exactly as expected.

PR #18100, opened 2017-08-31
sqlccl: test point in time recovery via BACKUP & RESTORE

danhhz:
If an operator issues a bad query or if a deploy contains a bug that
corrupts data, it should be possible to return to a previous point in
time before the badness. This adds a test for two ways to do this using
BACKUP & RESTORE.

If no previous BACKUPs have been taken, a new one can be taken using `AS
OF SYSTEM TIME` with a timestamp before the badness started. This can
then be RESTORE'd into a temporary database. The operator can manually
reconcile the current data with the restored data before finally
RENAME-ing the table into the final location.

If there is a recent BACKUP (either full or incremental), then it will
likely be faster to make an incremental BACKUP from that latest BACKUP
and RESTORE using that. Everything else works the same as above.

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -1953,3 +1953,83 @@ func TestBackupAzureAccountName(t *testing.T) {
 		t.Fatalf("unexpected error %v", err)
 	}
 }
+
+// If an operator issues a bad query or if a deploy contains a bug that corrupts
+// data, it should be possible to return to a previous point in time before the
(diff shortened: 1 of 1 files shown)


benesch (approved):
It does seem silly that we can't use `INSERT INTO foo2 [SELECT * FROM foo AS OF SYSTEM TIME xxx]`, but this works today, I suppose. LGTM!

madelynnblue (approved):

danhhz:
Thanks all!

PR #18101, opened 2017-08-31
sqlccl: add missing ctx select in splitAndScatter

danhhz:
Caught by Radu accidentally during some testing. The restore deadlocked
between `g.Wait` and a splitAndScatter worker sending on this channel
with a missing ctx.Done() select.

Also add a `defer cancel()` catchall in case something like this happens
again.

### DIFF
--- a/pkg/ccl/sqlccl/restore.go
+++ b/pkg/ccl/sqlccl/restore.go
@@ -639,7 +639,12 @@ func splitAndScatter(
 						// throughput.
 						log.Errorf(ctx, "failed to scatter %d: %s", idx, pErr.GoError())
 					}
-					readyForImportCh <- importSpan
+
+					select {
+					case <-ctx.Done():
+						return ctx.Err()
+					case readyForImportCh <- importSpan:
+					}
 				}
 			}
 			return nil
@@ -772,7 +777,9 @@ func restore(
 		return failed, errors.Wrapf(err, "making import requests for %d backups", len(backupDescs))
 	}
 
-	restoreCtx, cancel := context.WithCancel(restoreCtx)
+	var cancel func()
+	restoreCtx, cancel = context.WithCancel(restoreCtx)
+	defer cancel()
 	if err := job.Created(restoreCtx, cancel); err != nil {
 		return failed, err
 	}


pkg/ccl/sqlccl/restore.go
```diff
@@ -772,7 +777,9 @@ func restore(
 		return failed, errors.Wrapf(err, "making import requests for %d backups", len(backupDescs))
 	}
 
-	restoreCtx, cancel := context.WithCancel(restoreCtx)
+	var cancel func()
+	restoreCtx, cancel = context.WithCancel(restoreCtx)
+	defer cancel()
```
benesch: Was there a reason you needed to switch from shadowing `restoreCtx` to reassigning it?

benesch (approved):

RaduBerinde:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18101#-:-Kstuux1YXMF_HHrpfOB:bnfp4nl)*

pkg/ccl/sqlccl/restore.go
mjibson: I'm not sure why this was legal before. The spec seems to say it shouldn't have worked.

> Unlike regular variable declarations, a short variable declaration may redeclare variables provided they were originally declared earlier in the same block (or the parameter lists if the block is the function body) with the same type, and at least one of the non-blank variables is new. As a consequence, redeclaration can only appear in a multi-variable short declaration. Redeclaration does not introduce a new variable; it just assigns a new value to the original.

https://golang.org/ref/spec#Short_variable_declarations

madelynnblue (approved):

danhhz:

---

*[pkg/ccl/sqlccl/restore.go, line 782 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18101#-Kstr9e9k76Mg-iDDe53-r1-782:-KstzGWm7mqq7cDMUuHZ:b-yvscjx) ([raw file](https://github.com/cockroachdb/cockroach/blob/822bd176cc725c6b50905ea615023200b395e14f/pkg/ccl/sqlccl/restore.go#L782)):*

I'm not sure why this was legal before. The spec seems to say it shouldn't have worked.

> Unlike regular variable declarations, a short variable declaration may redeclare variables provided they were originally declared earlier in the same block (or the parameter lists if the block is the function body) with the same type, and at least one of the non-blank variables is new. As a consequence, redeclaration can only appear in a multi-variable short declaration. Redeclaration does not introduce a new variable; it just assigns a new value to the original.

https://golang.org/ref/spec#Short_variable_declarations
</blockquote></details>

no particular reason, just dislike this sort of shadowing on principle and I was already here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18101)*

danhhz:
TFTRs!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18101#-:-KstzcHmFrc2yEMnoMeL:bz5mair)*

benesch:

---

*[pkg/ccl/sqlccl/restore.go, line 782 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18101#-Kstr9e9k76Mg-iDDe53-r1-782:-Ksu3vTvueB5Fd4Zep3i:b-fdci2e) ([raw file](https://github.com/cockroachdb/cockroach/blob/822bd176cc725c6b50905ea615023200b395e14f/pkg/ccl/sqlccl/restore.go#L782)):*
> Unlike regular variable declarations, a short variable declaration may redeclare variables provided they were originally declared earlier in the same block (or the parameter lists if the block is the function body) with the same type, and at least one of the non-blank variables is new.

Er, doesn't it say that it _should_ work? Because "cancel" is new?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18101)*

PR #18104, opened 2017-08-31
pgwire: remove bogus empty commandComplete

andreimatei:
This patch removes the sending of a funky commandComplete message with
an empty command tag. The current code doesn't make much sense - it send
the message if results have been produced by the last transaction.
Before the recent move to results streaming, the code was sending this
message if, after executing a _batch_, it didn't get any results. I
can't see how that case would ever happen (and checked that it didn't
happen in our tests), because we have special handling of query strings
that don't result in any statements.
Also, the protocol doesn't say anything about such a message with an
empty tag.

### DIFF
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -1270,6 +1270,10 @@ func (e *Executor) execStmtInAbortedTxn(
 			return err
 		}
 
+		statementResultWriter.BeginResult((*parser.RollbackToSavepoint)(nil))
+		if err := statementResultWriter.EndResult(); err != nil {
+			return err
+		}
 		if txnState.State() == RestartWait {
 			// Reset the state to FirstBatch. We're in an "open" txn again.
 			txnState.SetState(FirstBatch)
@@ -1529,7 +1533,11 @@ func (e *Executor) execStmtInOpenTxn(
 	case *parser.Prepare:
 		// This must be handled here instead of the common path below
 		// because we need to use the Executor reference.
-		return e.PrepareStmt(session, s)
+		if err := e.PrepareStmt(session, s); err != nil {
+			return err
+		}
+		statementResultWriter.BeginResult((*parser.Prepare)(nil))
+		return statementResultWriter.EndResult()
 
 	case *parser.Execute:
 		// Substitute the placeholder information and actual statement with that of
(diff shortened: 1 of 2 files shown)


jordanlewis:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104#-:-Ksu1Ul3BvCXGtygAlpd:bnfp4nl)*

andreimatei:
Added a prequel commit. PTAL
cc @knz

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104#-:-KsuCrutdD1L-H5jfTn8:bh9dpqz)*

jordanlewis:
So you found some cases where the original commit's code would fire? In that case, it should be easy to write tests that fail without the "prequel commit" - Let's get those in!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104#-:-KsuDNZqIiSW4nVlkS7-:b-tr5628)*

andreimatei:
I'm not sure I follow. Without the first commit, but with the 2nd commit, we have plenty of tests that fail - because the statements fixed in the 1st commit wouldn't send any results to the client.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104#-:-KsuEfAZTTSOoRtfZj5b:b5tqjb2)*

jordanlewis:
Oh, in that case, all's well - I only commented because you had said that you ran all the tests without encountering the codepath you removed. Presumably the tests did in fact hit that codepath, then?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104#-:-KsuFeMrUr3tdwujfogu:b-f27i5j)*

andreimatei:
Before result streaming, the funky commandComplete was not hit. After result streaming, it was inadvertently hit.
With enough error tolerance, all the tests pass :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104#-:-KsuG5lK7uI1hqWG4OTL:bj82zez)*

jordanlewis:
Reviewed 2 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18104)*

Issue #17666, opened 2017-08-15
ui: better "no results" page


kuanluo:
Our current "no results" page is a bit under-designed. Perhaps we can make it more interesting? One idea below:

Current | Proposal
------------ | -------------
![image](https://user-images.githubusercontent.com/1486483/29330290-faf37058-81b5-11e7-951d-89dd5228ea02.png) | 
![no results](https://user-images.githubusercontent.com/1486483/29331486-ce27d998-81b9-11e7-9f87-aa2e84d1c039.png)

cc: @benesch @dianasaur323

dianasaur323:
I like this new design, but since @mrtracy is strapped for time finishing up some features, going to aggressively move this into 1.2

dianasaur323:
cc @nstewart this is the first outstanding issue that no results looks weird.

BramGruneir:
@kuanluo can you send me the no results image please, spider and text included!

kuanluo:
@BramGruneir I just slacked it to you. Github doesn't support SVG files. :)

PR #18107, opened 2017-08-31
ui: updates to the jobs page

### LINKED ISSUE #17861: ui: incorporate a "loading" page status when pages are waiting to load
@mrtracy based on our discussion today, it seems like the jobs table shows the "no results" view instead of a loading view. Could we create a loading view? 

cc @kuanluo for design.

We also need to consider if this is a big change vs a bug fix, so please let me know. Thanks! If this is a bug fix, let's get this in by 1.1

cc @nstewart 
cc @cuongdo

BramGruneir:
- Add a loading spinner to sorted and sortable tables.
- Use the spinner for when the jobs results are still not available.
- Adds a no results image when the jobs query returns with no results.

Closes #17861, #17666.

screen shot 2017-08-31 at 16 59 06
screen shot 2017-08-31 at 16 58 17


BramGruneir:
cc: @dianasaur323, @kuanluo

kuanluo:
Thank you, @BramGruneir ! Two nits: 

1. The loading circle is too big. Half of current size would be great. 
2. Both the loading circle and the No Results image should be about 10% above the waist of the screen, so just a bit higher would be great.

pkg/ui/src/views/shared/components/sortedtable/index.tsx
```diff
@@ -121,15 +131,19 @@ export class SortedTable<T> extends React.Component<SortedTableProps<T>, {}> {
   );
 
   render() {
-    const { data, sortSetting, onChangeSortSetting } = this.props;
+    const { data, sortSetting, onChangeSortSetting, loading } = this.props;
+    if (!_.isNil(loading) && loading) {
```
mrtracy: this can just be `if (loading)`, `!_.isNil()` is guaranteed true if loading is true.

But also, see above, this should be a separate component which wraps the table.

pkg/ui/src/views/shared/components/sortedtable/index.tsx
```diff
@@ -121,15 +131,19 @@ export class SortedTable<T> extends React.Component<SortedTableProps<T>, {}> {
...
-                            onChangeSortSetting={onChangeSortSetting}
-                            columns={this.columns(this.props)}
-                            rowClass={this.rowClass(this.props)}
-                            className={this.props.className} />;
+        sortSetting={sortSetting}
+        onChangeSortSetting={onChangeSortSetting}
+        columns={this.columns(this.props)}
+        rowClass={this.rowClass(this.props)}
+        className={this.props.className}
+        loading={this.props.loading} />;
```
mrtracy: You're duplicating work here; both sortedTable and sortableTable can display a 'spinner' gif section, it should either be one or the other.

But again, as stated above, it should be a separate component.

mrtracy:
Okay, apparently my review dropped a comment: the "loading' functionality should not be added to the table directly, it can be a separate component which displays either a loading spinner or its children.  That component could wrap an arbitrary child, not just a table.

dianasaur323:
Great! As long as we are using the same components for loading that @mrtracy is using for the graphs, it's fine with me! Thanks!

BramGruneir:
So I created a wrapper that I call `Loading` but I actually use it twice, once for loading and once for the fancy no results page.  

The spinner is now the correct size and both images are slightly above the fold.  




---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/ui/src/views/shared/components/sortedtable/index.tsx, line 135 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18107#-Ksu51DQ3Tx33PxQSeHn-r1-135:-Ksw4oVSoJJFoDoqxPbO:b6y01s1) ([raw file](https://github.com/cockroachdb/cockroach/blob/1b0f33d8e082a5749d3a8df4c7c6faa7a5328a8b/pkg/ui/src/views/shared/components/sortedtable/index.tsx#L135)):*

this can just be `if (loading)`, `!_.isNil()` is guaranteed true if loading is true.

But also, see above, this should be a separate component which wraps the table.
</blockquote></details>

It's now a separate component.

---

*[pkg/ui/src/views/shared/components/sortedtable/index.tsx, line 145 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18107#-Ksu51DQ3Tx33PxQSeHn-r1-145:-Ksw4ksjDalzlB5pK5U4:b6y01s1) ([raw file](https://github.com/cockroachdb/cockroach/blob/1b0f33d8e082a5749d3a8df4c7c6faa7a5328a8b/pkg/ui/src/views/shared/components/sortedtable/index.tsx#L145)):*

You're duplicating work here; both sortedTable and sortableTable can display a 'spinner' gif section, it should either be one or the other.

But again, as stated above, it should be a separate component.
</blockquote></details>

It's now a separate component.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18107#-:-Ksw4r9n6M49-A4gUrRe:b-ib6v7f)*

dianasaur323:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18107#-:-KsxhNTEP2pXBVxekwS_:bnfp4nl)*

kuanluo:
LGTM! Thank you, @BramGruneir @mrtracy !

mrtracy:
:lgtm:, thanks for doing that, much better now

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18107#-:-Ksy8h1cfmvj0nRwUB4q:b-21c0s4)*

PR #13319, opened 2017-01-31
docs: propose an RFC to clarify namespaces

knz:
cc @bdarnell @jseldess  @nvanbenschoten

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach/13319)

PR #18108, opened 2017-08-31
sql: correct behaviour of current_schemas(false)

### LINKED ISSUE #13997: sql: built-in `current_schemas` does not do what it should do
in psql:

```
kena=# set search_path=foa,pg_catalog;
SET
kena=# select current_schemas(true);
 current_schemas  
------------------
 {foa,pg_catalog}
(1 row)

kena=# select current_schemas(false);
 current_schemas  
------------------
 {foa,pg_catalog}
(1 row)
```

Note: **if pg_catalog was set explicitly in the search path by the user, `current_schema` will report it regardless of its argument**

In CockroachDB:

```
root@:26257/t> set search_path=foa,pg_catalog;
SET
root@:26257/> SELECT current_schemas(true);
+-----------------------+
| current_schemas(true) |
+-----------------------+
| {foa,pg_catalog}      |
+-----------------------+
(1 row)
root@:26257/> SELECT current_schemas(false);
+------------------------+
| current_schemas(false) |
+------------------------+
| {foa}                  |
+------------------------+
(1 row)
```

In other words the logic that has been added to `current_schemas()` in 951539ac6115d9edf7034d174ed80ceb83187b46 is incomplete.

justinj:
Fixes #13997.

The iteration of the session search path is more complex than just
looping through a list - it always includes pg_catalog, but there is an
observable difference between pg_catalog existing within the actual
search path itself and not.

Previously, we would always explicitly include pg_catalog within the
search path. This change abstracts the iteration of the search path to
add the pg_catalog when necessary, and makes the fields on SearchPath
private so that one doesn't accidentally iterate it manually.

Worth noting that this still isn't 100% accurate - Postgres'
current_schemas doesn't show entries in the search_path which are not
actual schemas.

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -250,7 +250,7 @@ func makeCSVTableDescriptor(
 		ctx,
 		nil, /* txn */
 		sql.NilVirtualTabler,
-		nil, /* SearchPath */
+		parser.SearchPath{},
 		create,
 		parentID,
 		tableID,
--- a/pkg/ccl/sqlccl/load.go
+++ b/pkg/ccl/sqlccl/load.go
@@ -148,7 +148,7 @@ func Load(
 			// only uses txn for resolving FKs and interleaved tables, neither of which
 			// are present here.
 			var txn *client.Txn
-			desc, err := sql.MakeTableDesc(ctx, txn, sql.NilVirtualTabler, nil, s, dbDesc.ID, 0 /* table ID */, ts, privs, affected, dbDesc.Name, &evalCtx)
+			desc, err := sql.MakeTableDesc(ctx, txn, sql.NilVirtualTabler, parser.SearchPath{}, s, dbDesc.ID, 0 /* table ID */, ts, privs, affected, dbDesc.Name, &evalCtx)
 			if err != nil {
 				return BackupDescriptor{}, errors.Wrap(err, "make table desc")
 			}
--- a/pkg/sql/create.go
+++ b/pkg/sql/create.go
@@ -1161,7 +1161,7 @@ func (n *createViewNode) makeViewTableDesc(
 			columnTableDef.Name = columnNames[i]
 		}
 		// We pass an empty search path here because there are no names to resolve.
-		col, _, err := sqlbase.MakeColumnDefDescs(&columnTableDef, nil, evalCtx)
+		col, _, err := sqlbase.MakeColumnDefDescs(&columnTableDef, parser.SearchPath{}, evalCtx)
 		if err != nil {
 			return desc, err
 		}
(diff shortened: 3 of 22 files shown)


knz:
I also want to look into this. Unfortunately my brain is fried for today.

jordanlewis:
Generally :lgtm:, but I'd like to raise my usual anxiety about not having any equivalent to the default 'public' value that's in Postgres' default search path. Even though we don't have actual schema support, it might be misleading to tools and users that names can still be resolved despite having nothing in the search path.

---

Review status: 0 of 19 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18108#-:-KsuCgm5vHSucEPCs_ub:b-dgebri)*

knz:
Can you have a cursory look at https://airtable.com/tblD3oZPLJgGhCmch/viwrzXTNNjgK78B4k/recmkB89sOzx3FDw1

The salient paragraph is the description of what was envisioned, the outcome of a discussion I had with Jordan:

> Change the SQL interfaces that allow apps to introspect the catalog to reveal a schema in-between database and table. Make each database in CockroachDB have a single main schema called "public". Keep our current notion of "virtual schema" as schema not database, and make virtual schemas appear in every database.

> This project is *not* about allowing clients to create multiple schemas side-by-side inside a single database! (that's https://airtable.com/tblD3oZPLJgGhCmch/viwrzXTNNjgK78B4k/recRy5ObWF8IzLKgA / #13319) In this iteration each database would have a single container schema for user-defined tables called "public". From a code perspective the descriptor structure would stay the same -- all the work would be in name resolution and the construction of the virtual tables "pg_catalog" and "information_schema".

justinj:
@knz I've separated out the search path construction and added explicit unit tests for it to make future changes a little easier, I'll open an issue for the change described in that link (and add the GitHub link to Airtable) if that's acceptable to you? I'd like to get this change in for 1.1 but I think that one might be ambitious before the code freeze.

justinj:
@knz would you be ok with me cherry-picking this into 1.1 and leaving the larger schema/database reimagining to 1.2?

jordanlewis:
:lgtm: I like these tests. I think we should cherrypick this for sure.

---

Review status: 0 of 22 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18108#-:-KtMiNRUHFGcwuPpdOVP:b-fof9l9)*

jordanlewis:
Reviewed 10 of 19 files at r1, 12 of 12 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18108)*

knz:
Sorry it took me a while to look at this.
I love your new SearchPath object and the idea to iterate from -1. This made me smile and feel smarter.

:lgtm_strong: 

---

Reviewed 10 of 19 files at r1, 12 of 12 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18108#-:-KtXi06BNMgRmgJg-aBa:b-29nm3n)*

knz:
And yes please cherry-pick to 1.1

knz:
@jordanlewis I'm ok with having this PR go in for now, and then when/if a client requires it revisit the introduction of a `public` schema.

jordanlewis:
SGTM @knz

PR #18111, opened 2017-08-31
sql: expose job leaseholder node in crdb_internal.jobs

danhhz:
Omit it from SHOW JOBS in case we change our minds.

@benesch I'm actually leaning toward adding this to SHOW JOBS as well. Thoughts?


danhhz:
We decided offline to also include this in SHOW JOBS. Updating...

benesch (approved):

danhhz:
There was an inconclusive-ish internal poll on whether to call this column node_id, coordinator_id, or something else. coordinator_id won it 2 to 1, so if anyone has strong opinions on this, please speak up

@benesch This is now ready for review

dt (approved):

danhhz:
TFTRs!

PR #18118, opened 2017-09-01
sqlccl: save full EntryCounts in BackupDescriptor_File

danhhz:
No reason to drop data here. Not used for anything now, but this gives
us the flexibility to later show row counts per table, for example.


mjibson:

---

*[pkg/ccl/sqlccl/backup.go, line 537 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18118#-KsvCLG8_aH8Emb0Knbs:-KsvCLG8_aH8Emb0Knbt:b-6wigjh) ([raw file](https://github.com/cockroachdb/cockroach/blob/064d691705ed54decccd216e3381cf7dbb6ff5ad/pkg/ccl/sqlccl/backup.go#L537)):*
> ```Go
> 					Path:     file.Path,
> 					Sha512:   file.Sha512,
> 					DataSize: uint64(file.Exported.DataSize),
> ```

Why don't we care about this anymore?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18118)*

danhhz:

---

*[pkg/ccl/sqlccl/backup.go, line 537 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18118#-KsvCLG8_aH8Emb0Knbs:-KsxLavK2SFiqLKPV0wN:bu5wswk) ([raw file](https://github.com/cockroachdb/cockroach/blob/064d691705ed54decccd216e3381cf7dbb6ff5ad/pkg/ccl/sqlccl/backup.go#L537)):*

Why don't we care about this anymore?
</blockquote></details>

We do. It's also present in EntryCounts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18118)*

dt (approved):
Should we teach `SHOW BACKUP` about the other counts?

danhhz:
Hmm, good point. rows would be nice. we can probably skip index_entries and system_records. I'm making a change to crdb_internal.jobs and maybe SHOW JOBS over in https://github.com/cockroachdb/cockroach/pull/18111 so I'm inclined to do it there and merge this once teamcity goes green

danhhz:
> I'm making a change to crdb_internal.jobs and maybe SHOW JOBS over in #18111 so I'm inclined to do it there and merge this once teamcity goes green

ugh. disregard the second half of that comment. I have no idea what I was thinking

danhhz:
added rows to SHOW BACKUP. ptal

dt:
:lgtm:

We also mentioned including the buildInfo in `SHOW BACKUP` earlier this week -- If we still like that idea, I'm happy to follow this with that, unless you find you need to eat another cycle and really want to do it while you're here.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18118#-:-KsxbtMW22Y6n2Ksu_du:b-p8swul)*

danhhz:
TFTR!

My primary concern here is the part that writes new backup descriptors, so I'd like to make sure this is in. I've got a green build so I'll go ahead and merge.

Please do feel free to followup with more SHOW BACKUP changes

PR #18125, opened 2017-09-01
ui/server: Analytics Identify Event

mrtracy:
Admin UI analytics, when enabled, now dispatches a single "identify"
event per browser session. This message includes the clusterID, the
version of CockroachDB that is running (the build tag), the userAgent of
the browser being used, and a boolean indicating if enterprise features
are enabled on the cluster.

Adds an "enterprise_enabled" field to the Admin "Cluster" endpoint.


mrtracy:
cc @dianasaur323

knz:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/server/admin.go, line 971 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-KswmwkITAdfSWAL2BwX:-KswmwkJDYb1dc0wf4kx:b-adogaq) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/server/admin.go#L971)):*
> ```Go
> 	// between different features.
> 	enterpriseEnabled := false
> 	organization := sql.ClusterOrganization.Get(&s.server.st.SV)
> ```

no need to export `sql.clusterOrganization`; either:

- use`s.server.sqlExecutor.cfg.Organization()`  here, or
- move clusterOrganization to the server package

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125)*

pkg/server/admin.go
```diff
@@ -962,9 +963,22 @@ func (s *adminServer) Cluster(
 	if clusterID == (uuid.UUID{}) {
 		return nil, grpc.Errorf(codes.Unavailable, "cluster ID not yet available")
 	}
+
+	// Check if enterprise features are enabled.  We currently test for the
```
dt: maybe something like https://github.com/cockroachdb/cockroach/commit/64fd4d3aacc102cb1021bfac4100c2cefb4ff23c

dt: another option, that I like slightly more but don't feel that strongly about, would be to have some string->string `analyticsTags` or such that isn't "enterprise" or "license" specific, with a ccl hook that populates a "enterprise" tag, to avoid "enterprise" handling in the pure FLOSS code. That said, our commitment thus far has just been that we will ensure that it is possible to build a pure FLOSS binary, and there's already places where it shows that it was extracted from the full CCL codebase (e.g. the enterprise statements have to live in the FLOSS grammar and parser ASTS, even their execution isn't implemented), so I don't think it matters that much.

dianasaur323:
:lgtm: Thank you! I'm assuming it's more work than we have time for to put in lifetime, in which case we can figure out how to join registration cluster info with the info here?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-:-KsxcmXrhFslPvA1guQ5:bd4ftuj)*

dt:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/server/admin.go, line 971 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-KswmwkITAdfSWAL2BwX:-KsxdJsPRyVN4bwu7Lc_:bwh6osw) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/server/admin.go#L971)):*
> move clusterOrganization to the server package
:thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125)*

BramGruneir:
Frontend code LGTM

---

Reviewed 3 of 10 files at r1.
Review status: 3 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/ui/src/redux/analytics.ts, line 8 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-Ksy4SFW3JHFRBk83Otu:-Ksy4SFW3JHFRBk83Otv:blunbn2) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/ui/src/redux/analytics.ts#L8)):*
> ```XML
> import * as protos from "src/js/protos";
> import { store, history, AdminUIState } from "src/redux/state";
> import { versionsSelector } from "src/redux/alerts";
> ```

alphabetize please

---

*[pkg/ui/src/redux/analytics.ts, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-Ksy4pEFseUKFRZ13Aa9:-Ksy4pEFseUKFRZ13AaA:b-fffi5z) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/ui/src/redux/analytics.ts#L116)):*
> ```XML
>         }
> 
>         // If the cluster ID is not yet available, queue the location to be
> ```

This comment is not correct.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-:-Ksy58Io2vGQwKepSE-B:bhykela)*

mrtracy:
How can I move ClusterOrganization to the SQL package?  It's needed in the sql package, I can't import it from the server package

I also can't use `s.server.sqlExecutor.cfg.Organization()` here, or anywhere, because `s` is defined in the server package and `sqlExecutor` in the sql package.

I think it just has to be exported from the sql package.

mrtracy:
Review status: 1 of 12 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/server/admin.go, line 967 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-KswAgKJBLbaPqcekbEq-r1-967:-Ksyg8thwuNYs6sIunbI:b-xqg0fw) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/server/admin.go#L967)):*

another option, that I like slightly more but don't feel that strongly about, would be to have some string->string `analyticsTags` or such that isn't "enterprise" or "license" specific, with a ccl hook that populates a "enterprise" tag, to avoid "enterprise" handling in the pure FLOSS code. That said, our commitment thus far has just been that we will ensure that it is possible to build a pure FLOSS binary, and there's already places where it shows that it was extracted from the full CCL codebase (e.g. the enterprise statements have to live in the FLOSS grammar and parser ASTs, even though their execution isn't implemented), so I don't think it matters that much.
</blockquote></details>

I've taken the first route you suggested of adding a `LicenseCheckFn`to the server package.  I've done this because, while adding a tag would indicate if the binary was compiled with CCL features, I don't think it would indicate if they were actually enabled via a license.

---

*[pkg/server/admin.go, line 971 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-KswmwkITAdfSWAL2BwX:-Ksyfe4K-jLHM9WR1xtV:b7a7e1y) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/server/admin.go#L971)):*

> move clusterOrganization to the server package
:thumbsup: 
</blockquote></details>

I don't think this can be done easily, because clusterOrganization is used in the sql package; it can't just be moved to the server package and used from sql, it would need to be moved to a lower package.  But server already depends on sql, so I think it's in the right place, it just needs to be exported.

---

*[pkg/ui/src/redux/analytics.ts, line 8 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-Ksy4SFW3JHFRBk83Otu:-Ksy_4iU5QIFf4obAcZc:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/ui/src/redux/analytics.ts#L8)):*

alphabetize please
</blockquote></details>

Done.

---

*[pkg/ui/src/redux/analytics.ts, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-Ksy4pEFseUKFRZ13Aa9:-Ksycp2uTxW_e4X9AEx8:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/ui/src/redux/analytics.ts#L116)):*

This comment is not correct.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125)*

dt:
:lgtm:

---

Review status: 1 of 12 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/server/admin.go, line 967 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-KswAgKJBLbaPqcekbEq-r1-967:-KsyjrOF41KnA6KCsEIM:bjav9ps) ([raw file](https://github.com/cockroachdb/cockroach/blob/402aac0d5180fbd7291c30b05b3952ad773e2585/pkg/server/admin.go#L967)):*

I've taken the first route you suggested of adding a `LicenseCheckFn`to the server package.  I've done this because, while adding a tag would indicate if the binary was compiled with CCL features, I don't think it would indicate if they were actually enabled via a license.
</blockquote></details>

:thumbsup: 

fwiw, i didn't mean a build tag, but rather a func that when called can return a map[string]string or something. We'd call it in the endpoint and let it return whatever it wants. it'd just call the lic check itself. functionally doesn't matter much, it'd just keep some "enterprise"-specific code out of the FLOSS codebase. anyway, like i said, i don't care much because that ship sailed elsewhere already.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18125#-:-KsykFXcmH01jwjKBb3q:bnfp4nl)*

PR #18127, opened 2017-09-01
cli/sql: new options to help troubleshooting

### LINKED ISSUE #18120: sql: auto-retries are hard to test using the CLI because of the injection of extra statements
Consider (one batch per line):
```
BEGIN;
SELECT crdb_internal.force_retry(interval '10s');
```

We have a nice feature that allows for the second batch to be retried even though the transaction technically began on the previous batch. However, if you do perform these queries in the CLI, you'll notice that the 2nd batch is not retried automatically. That's because the shell injects statements behind batches to show the current database and transaction status. The statements break the inter-batch magic.
I'd like an option to have the shell not run statements under me.

And btw, I've noticed that the pg protocol passes the transaction status back and forth transparently, so in principle we don't need to run a `show transaction status` to get it. Except I guess that doesn't support our extended `RestartWait` state?

knz:
Fixes #18120.

cc @cockroachdb/sql-execution 

See the two commits for details.

### DIFF
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -691,6 +691,8 @@ func Example_sql() {
 	c.RunWithArgs([]string{"sql", "-d", "nonexistent", "-e", "create database nonexistent; create table foo(x int); select * from foo"})
 	// COPY should return an intelligible error message.
 	c.RunWithArgs([]string{"sql", "-e", "copy t.f from stdin"})
+	// --echo-sql should print out the SQL statements.
+	c.RunWithArgs([]string{"user", "ls", "--echo-sql"})
 
 	// Output:
 	// sql -e show application_name
@@ -747,6 +749,10 @@ func Example_sql() {
 	// x
 	// sql -e copy t.f from stdin
 	// woops! COPY has confused this client! Suggestion: use 'psql' for COPY
+	// user ls --echo-sql
+	// > SELECT username FROM system.users
+	// 0 rows
+	// username
 }
 
 func Example_sql_format() {
--- a/pkg/cli/cliflags/flags.go
+++ b/pkg/cli/cliflags/flags.go
@@ -153,6 +153,12 @@ with a non-zero status code and further statements are not executed. The
 results of each SQL statement are printed on the standard output.`,
 	}
 
+	EchoSQL = FlagInfo{
+		Name: "echo-sql",
+		Description: `
+Reveal the SQL statements sent implicitly by the command-line utility.`,
+	}
+
 	UnsafeUpdates = FlagInfo{
 		Name: "unsafe-updates",
 		Description: `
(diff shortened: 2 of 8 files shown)


andreimatei:
:lgtm: 

Thanks a lot!

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/cli/sql.go, line 81 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18127#-Ksy4v4FHfmS20h5IU_e:-Ksy4v4FHfmS20h5IU_f:b-m1xz1d) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c845132bd3db00b89f74fe01c0fff0b3a74c03d/pkg/cli/sql.go#L81)):*
> ```Go
> 	// properly.
> 	normalizeHistory bool
> 	// smartPrompt indicates whether to update the prompt using queries to the server.
> ```

nit: give a hint as to what the queries and the prompt updates would be

---

*[pkg/cli/interactive_tests/test_local_cmds.tcl, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18127#-Ksy4mlHMBNTJvtu0yEd:-Ksy4mlHMBNTJvtu0yEe:b-rwdo6y) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bb58b1dd22f24372f43ea7d135e2ffdbaf0049b/pkg/cli/interactive_tests/test_local_cmds.tcl#L67)):*
> ```Tcl
> start_test "Check that \\set without argument prints the current options"
> send "\\set\r"
> eexpect "5 rows"
> ```

how come this changed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18127#-:-Ksy50r0CrQLjgv9ZCmO:b-945yy1)*

knz:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/cli/sql.go, line 81 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18127#-Ksy4v4FHfmS20h5IU_e:-KsyfHS93NqBX7nE91ma:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c845132bd3db00b89f74fe01c0fff0b3a74c03d/pkg/cli/sql.go#L81)):*

nit: give a hint as to what the queries and the prompt updates would be
</blockquote></details>

Done.

---

*[pkg/cli/interactive_tests/test_local_cmds.tcl, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18127#-Ksy4mlHMBNTJvtu0yEd:-KsyfIv7PsdJJcH2yu56:bp4g551) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bb58b1dd22f24372f43ea7d135e2ffdbaf0049b/pkg/cli/interactive_tests/test_local_cmds.tcl#L67)):*

how come this changed?
</blockquote></details>

The test doesn't need to check how many options there are in total, just that they are printed. I figured it's silly to have to update this counter here every time we add a new option.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18127)*

PR #18128, opened 2017-09-01
storage: disable raft pre-vote

irfansharif:
In light of issues `blue` and `adriatic` ran into this week hitting 0
QPS and maintaining non-zero MsgPreVote{,Resp} message counts thereon
forth. Disable it ahead of the v1.1 branch cut.

+cc @bdarnell, merge when approved, I'll be AFK today morning.


a-robinson (approved):
LGTM, even if only due to the rolling upgrade concerns not being worked out.

We'll still need to understand what's been happening, though.

PR #18141, opened 2017-09-01
storage: Remove very chatty log line that looks accidental

a-robinson:
Added in 45396f80916f6bbff015de4658fe4a8e54cf7f7b

@petermattis this looks like it was some debug logging that you left in given how chatty it is.


petermattis (approved):
LGTM

Oops.

PR #18148, opened 2017-09-01
storage: remove empty sideloaded directories

bsnyder788:
@tschottdorf first time contributing to Cockroach, but I think this PR is ready for review.

### DIFF
--- a/pkg/storage/replica_sideload_disk.go
+++ b/pkg/storage/replica_sideload_disk.go
@@ -125,12 +125,11 @@ func (ss *diskSideloadStorage) Clear(_ context.Context) error {
 }
 
 func (ss *diskSideloadStorage) TruncateTo(ctx context.Context, index uint64) error {
-	// TODO(tschottdorf): if we end up removing *all* files, we could also
-	// remove the directory.
 	matches, err := filepath.Glob(filepath.Join(ss.dir, "i*.t*"))
 	if err != nil {
 		return err
 	}
+	var deleted int
 	for _, match := range matches {
 		base := filepath.Base(match)
 		if len(base) < 1 || base[0] != 'i' {
@@ -147,6 +146,14 @@ func (ss *diskSideloadStorage) TruncateTo(ctx context.Context, index uint64) err
 		}
 		if err := ss.purgeFile(ctx, match); err != nil {
 			return errors.Wrapf(err, "while purging %q", match)
+		}
+		deleted++
+	}
+
+	if deleted == len(matches) {
+		err = os.Remove(ss.dir)
+		if !os.IsNotExist(err) {
+			return errors.Wrapf(err, "while purging %q", ss.dir)
 		}
 	}
 	return nil
--- a/pkg/storage/replica_sideload_test.go
+++ b/pkg/storage/replica_sideload_test.go
@@ -41,6 +41,9 @@ import (
 	"github.com/coreos/etcd/raft/raftpb"
 	"github.com/kr/pretty"
 	"github.com/pkg/errors"
+	"math"
+	"os"
+	"path/filepath"
 )
 
 func entryEq(l, r raftpb.Entry) error {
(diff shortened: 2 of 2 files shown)


petermattis:
@bsnyder788 Thanks for you contribution. We'll need a test before being able to merge this. In addition to checking expected conditions, you should verify sane behavior if the sideloaded directory contains an unexpected file and thus can't be removed.

---

---

*[pkg/storage/replica_sideload_disk.go, line 150 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-KsysSayDC3KFIk6kNe1:-KsysSayDC3KFIk6kNe2:b-2lwhkw) ([raw file](https://github.com/cockroachdb/cockroach/blob/2df04395e5eefb2887ce50b8c0a62f45621e3ede/pkg/storage/replica_sideload_disk.go#L150)):*
> ```Go
> 		}
> 	}
> 	err = ss.purgeDirectoryIfEmpty(ctx)
> ```

@tschottdorf will have to speak up here, but the sideloaded directory only contains `i*.t*` files and thus if `len(matches) == 0` the directory is empty and can be removed. I think this can be simplified to:

```
if len(matches) == 0 {
  err = os.Remove(ss.dir)
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-:-KsyskObUgtuDHpJusaW:b-xsjfdk)*

bsnyder788:
*[pkg/storage/replica_sideload_disk.go, line 150 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-KsysSayDC3KFIk6kNe1:-KszKt1E8baqm6Pm59jd:b-lsboph) ([raw file](https://github.com/cockroachdb/cockroach/blob/2df04395e5eefb2887ce50b8c0a62f45621e3ede/pkg/storage/replica_sideload_disk.go#L150)):*

@tschottdorf will have to speak up here, but the sideloaded directory only contains `i*.t*` files and thus if `len(matches) == 0` the directory is empty and can be removed. I think this can be simplified to:

```
if len(matches) == 0 {
  err = os.Remove(ss.dir)
}
```
</blockquote></details>

I see, that would indeed be much simpler.  If @tschottdorf can confirm that is the case I will go ahead and make the changes and add the appropriate tests you suggested. Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148)*

petermattis:

---

*[pkg/storage/replica_sideload_disk.go, line 150 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-KsysSayDC3KFIk6kNe1:-KszMSR544TKppnh9UFN:bcv3fvp) ([raw file](https://github.com/cockroachdb/cockroach/blob/2df04395e5eefb2887ce50b8c0a62f45621e3ede/pkg/storage/replica_sideload_disk.go#L150)):*
<details><summary><i>Previously, bsnyder788 (Brett Snyder) wrote…</i></summary><blockquote>

I see, that would indeed be much simpler.  If @tschottdorf can confirm that is the case I will go ahead and make the changes and add the appropriate tests you suggested. Thanks!
</blockquote></details>

Heh, it's simple because it's wrong. We would need to account for deleting files in the loop:

```
  var deleted int
  for _, match := range matches {
    ...
    if err := ss.purgeFile(); err != nil {
      ...
    }
    deleted++
  }

  if deleted == len(matches) {
    err = os.Remove(ss.dir)
  }
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148)*

bsnyder788:
*[pkg/storage/replica_sideload_disk.go, line 150 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-KsysSayDC3KFIk6kNe1:-KtHEVqdKJuOEtv_t2RR:b-2anv9e) ([raw file](https://github.com/cockroachdb/cockroach/blob/2df04395e5eefb2887ce50b8c0a62f45621e3ede/pkg/storage/replica_sideload_disk.go#L150)):*

Heh, it's simple because it's wrong. We would need to account for deleting files in the loop:

```
  var deleted int
  for _, match := range matches {
    ...
    if err := ss.purgeFile(); err != nil {
      ...
    }
    deleted++
  }

  if deleted == len(matches) {
    err = os.Remove(ss.dir)
  }
```
</blockquote></details>

thanks!  I will take some time and update the PR accordingly, this week.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148)*

bsnyder788:
*[pkg/storage/replica_sideload_disk.go, line 150 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-KsysSayDC3KFIk6kNe1:-Ku4upo0UJeEk2p9Pj4b:bxc4b5s) ([raw file](https://github.com/cockroachdb/cockroach/blob/2df04395e5eefb2887ce50b8c0a62f45621e3ede/pkg/storage/replica_sideload_disk.go#L150)):*
<details><summary><i>Previously, bsnyder788 (Brett Snyder) wrote…</i></summary><blockquote>

thanks!  I will take some time and update the PR accordingly, this week.
</blockquote></details>

I believe I have addressed all comments now. Let me know if I should update anything else. Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148)*

petermattis:
:lgtm: modulo a few minor comments in the test.

---

---

*[pkg/storage/replica_sideload_test.go, line 273 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-Ku5ktpBMjwnxLGH50tA:-Ku5ktpBMjwnxLGH50tB:b-59asrp) ([raw file](https://github.com/cockroachdb/cockroach/blob/e49bfddcc11e0bae0fadc190d9cda75f45585db6/pkg/storage/replica_sideload_test.go#L273)):*
> ```Go
> 	if !isInMem {
> 		// first add a file that shouldn't be in the sideloaded storage to ensure
> 		// sane behaviour when directory can't be removed after full truncate
> ```

We try and use proper capitalization and punctuation in our comments (i.e. `s/first/First/g` and add a period after `truncate`). This applies to the comments below as well. Apologies for the nit, but consistency is nice.

---

*[pkg/storage/replica_sideload_test.go, line 294 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-Ku5lWyxr04JCUHLkwfP:-Ku5lWyxr04JCUHLkwfQ:b-ftxdwv) ([raw file](https://github.com/cockroachdb/cockroach/blob/e49bfddcc11e0bae0fadc190d9cda75f45585db6/pkg/storage/replica_sideload_test.go#L294)):*
> ```Go
> 		}
> 
> 		if err := ss.TruncateTo(ctx, math.MaxUint64); err != nil {
> ```

I believe this is only test the case where `filepath.Glob` returned 0 matches. It might also be good to test that the directory gets removed when a non-zero number of matches are all deleted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-:-Ku5lkcviZG0CR7TOP4i:b-koq4pb)*

bsnyder788:
*[pkg/storage/replica_sideload_test.go, line 273 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-Ku5ktpBMjwnxLGH50tA:-Ku5mehDtKaokno11Q3e:b8a1a9u) ([raw file](https://github.com/cockroachdb/cockroach/blob/e49bfddcc11e0bae0fadc190d9cda75f45585db6/pkg/storage/replica_sideload_test.go#L273)):*

We try and use proper capitalization and punctuation in our comments (i.e. `s/first/First/g` and add a period after `truncate`). This applies to the comments below as well. Apologies for the nit, but consistency is nice.
</blockquote></details>

No, apology necessary, thanks for the good feedback. I will make the changes and resubmit shortly!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148)*

bsnyder788:
*[pkg/storage/replica_sideload_test.go, line 294 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-Ku5lWyxr04JCUHLkwfP:-Ku5x6FwWjRpHjjaYv-1:b-bcbm9m) ([raw file](https://github.com/cockroachdb/cockroach/blob/e49bfddcc11e0bae0fadc190d9cda75f45585db6/pkg/storage/replica_sideload_test.go#L294)):*

I believe this is only test the case where `filepath.Glob` returned 0 matches. It might also be good to test that the directory gets removed when a non-zero number of matches are all deleted.
</blockquote></details>

@petermattis, PR has been updated to address this, 3 unit tests (seem to be flaky according to CI) have failed though, is there an easy way to re-trigger them to run without pushing a new commit?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148)*

petermattis:
I think we can merge on green.

---

---

*[pkg/storage/replica_sideload_test.go, line 294 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-Ku5lWyxr04JCUHLkwfP:-Ku6IgEvSh5M0II-Jqfe:bngn3hz) ([raw file](https://github.com/cockroachdb/cockroach/blob/e49bfddcc11e0bae0fadc190d9cda75f45585db6/pkg/storage/replica_sideload_test.go#L294)):*
<details><summary><i>Previously, bsnyder788 (Brett Snyder) wrote…</i></summary><blockquote>

@petermattis, PR has been updated to address this, 3 unit tests (seem to be flaky according to CI) have failed though, is there an easy way to re-trigger them to run without pushing a new commit?
</blockquote></details>

I've re-triggered a CI run. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18148#-:-Ku6IqYSEhaFsiTrRfJ0:b-2ourck)*

bsnyder788:
@petermattis should be good to go now, CI build was green after you re-triggered it!

PR #18157, opened 2017-09-01
sqlccl: include runnable, `trasform_only` IMPORT stmt in jobs text

dt:
while the original stmt doesn’t make sense as it also runs a RESTORE that already has its own job, if we add `transform_only`, we should have an essentially runnable stmt that corresponds to the job’s actual tracked work (storage URI sanitation notwithstanding).

it’s still a bit wonky having two jobs with one that looks very similar, but its the best we can do for now.

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -55,6 +55,11 @@ const (
 	importOptionTemp          = "temp"
 )
 
+var valuelessImportOptions = map[string]bool{
+	importOptionTransformOnly: true,
+	importOptionDistributed:   true,
+}
+
 // LoadCSV converts CSV files into enterprise backup format.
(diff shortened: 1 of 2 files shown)


madelynnblue (approved):
Oh this is not pretty, but I can't think of a better way.

dianasaur323:
Thanks @dt for coming up with a workaround here!

dianasaur323:
Thanks for figuring out a workaround for this! Although it's awkward to have two jobs, I think this approach seems fine for now. The new description text looks great to me. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18157#-:-KsylCciXAyqFzRVlF0Q:bcqxtn)*

benesch:
Merging at @dt's request!

PR #18158, opened 2017-09-01
sqlccl: let the reusable helper filter valueless opts

dt:
(ignore first commit, its in #18157)


madelynnblue (approved):

dt:
@mjibson revised to expand the map to include all options, and require options be in the map and return an error otherwise.

mjibson:
Could you add a test for the unexpected option error?

dt:
@mjibson done.

madelynnblue (approved):

Issue #14231, opened 2017-03-17
storage: panic in raftGroup.Step while running under stressrace


a-robinson:
Running at head (150b52daca1127372c6e925a164b9370586a570b), I was stressing `TestRemovePlaceholderRace` on an azworker to try to repro #12574 while looking into #14193. After running 5606 runs for over an hour, I hit a different (but possibly related) panic:

```
E170317 15:36:04.978177 166 vendor/github.com/coreos/etcd/raft/log.go:191  [s2,r1/24:{-},@c4208f9880] tocommit(123) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost?
panic: tocommit(123) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost? [recovered]
	panic: tocommit(123) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost?

goroutine 166 [running]:
github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).Recover(0xc42001ebe0)
	/home/alex/go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:198 +0xd8
panic(0x2135780, 0xc420543df0)
	/home/alex/go1.8/src/runtime/panic.go:489 +0x2f0
github.com/cockroachdb/cockroach/pkg/storage.(*raftLogger).Panicf(0xc420543b50, 0x23bf79b, 0x5d, 0xc420c8c460, 0x2, 0x2)
	/home/alex/go/src/github.com/cockroachdb/cockroach/pkg/storage/raft.go:111 +0x11d
github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft.(*raftLog).commitTo(0xc420c95500, 0x7b)
	/home/alex/go/src/github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft/log.go:191 +0x1fe
github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft.(*raft).handleHeartbeat(0xc4201524b0, 0x8, 0x18, 0x1, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/alex/go/src/github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft/raft.go:1100 +0x7a
[...]

PR #18172, opened 2017-09-01
storage: Log a help message if raft data is missing

a-robinson:
A hacky stopgap to try to explain what's happening if someone runs
into #14231. It's not great, but seems better than leaving users with
a hard-to-understand raft error if they make a configuration mistake.

This results in stderr output like this when I reproduce locally:

```
*
* ERROR: [n1,s1,r3/1:/System/NodeLiveness{-Max}] Server crashing due to missing data. Was the server restarted with a different store directory than before? A --join parameter must be specified in order to make restarts with a new store directory safe. Please try again with a new directory and a valid --join flag. If this server was started with its old data or with a valid --join parameter and you are still seeing this message, please report an issue at https://github.com/cockroachdb/cockroach/issues/new
*
*
* ERROR: [n1,s1] a panic has occurred!
*
tocommit(52) is out of range [lastIndex(20)]. Was the raft log corrupted, truncated, or lost?
```

I'm not a huge fan of this, but it's [happened quite a lot since 1.0 was released](https://sentry.io/cockroach-labs/cockroachdb/issues/270588925/) so it seems worth trying to give users some sort of explanation.

### DIFF
--- a/pkg/storage/raft.go
+++ b/pkg/storage/raft.go
@@ -18,6 +18,7 @@ package storage
 import (
 	"bytes"
 	"fmt"
+	"strings"
 
 	"golang.org/x/net/context"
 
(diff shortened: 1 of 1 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18172#-:-Kt8CXKsbHtyvV1NoSLQ:bnfp4nl)*

pkg/storage/raft.go
```diff
@@ -99,14 +100,42 @@ func (r *raftLogger) Fatalf(format string, v ...interface{}) {
...
+// understand them.
+//
+// TODO(#14231): Remove this method once the issue is fixed (hopefully in 1.2).
+func is14231Error(s string) bool {
+	return strings.Contains(s, "is out of range [lastIndex(") &&
+		strings.Contains(s, "Was the raft log corrupted, truncated, or lost?")
+}
+
+// TODO(#14231): Remove this once the issue is fixed (hopefully in 1.2).
+const is14231HelpMessage = "Server crashing due to missing data. Was the server restarted with a " +
```
knz: log.Shout supports multi-line messages. You can wrap this one.

petermattis:
Ugh, this feels so dirty, though it is a definite improvement on the current error message.

pkg/storage/raft.go
a-robinson: Done.

a-robinson (commented):
I don't disagree, but it's better than nothing.

knz (approved):

Issue #17995, opened 2017-08-29
TestDecommission is flaky


tschottdorf:
it times out during

```
// Run a second time to wait until the replicas have all been GC'ed.
// Note that we specify "all" because even though the first node is
// now running, it may not be live by the time the command runs.
o, err = decommission(ctx, c, 2, target, "decommission", "--wait", "all")
```

https://teamcity.cockroachdb.com/repository/download/Cockroach_UnitTests_Acceptance/337493:id/acceptance/acceptance.log

Issue #14250, opened 2017-03-18
stability: cobalt wedged after ~5h.


mberhault:
`cobalt` was wiped and restarted as usual by continuous deployment, with sha 43f24c9
```
I170318 05:01:50.445249 1 cli/start.go:325  CockroachDB CCL 43f24c9-dirty (linux amd64, built 2017/03/18 04:35:44, go1.8)
```

It wedged about 4h45 later. All nodes are still up and running, but liveness dropped to 0 and all ranges are unavailable:
<img width="1434" alt="screen shot 2017-03-18 at 3 42 09 pm" src="https://cloud.githubusercontent.com/assets/5901480/24073074/952cdcc6-0bf1-11e7-8e62-9992a9d6c630.png">

Nothing shocking in the logs, just disconnects:
first 10 warning/errors per node at 9:45, +/- 5 minutes:
```
1: cockroach@cockroach-cobalt-0001.crdb.io
E170318 09:45:48.791445 1179 rpc/context.go:265  [n1] removing connection to cockroach-cobalt-06:26257 due to error: rpc error: code = 4 desc = context deadline exceeded
W170318 09:45:48.796514 1497406 storage/raft_transport.go:442  [n1] raft transport stream to node 6 failed: rpc error: code = 13 desc = transport is closing
W170318 09:45:49.668196 26151470 storage/replica.go:2255  [n1,s1,r88/3:/Table/51/1/29{52735…-88943…},@c421648a80] context cancellation after 0.0s of attempting command [txn: e75cf19a], BeginTransaction [/Table/51/1/2974175659068629674/"f40c3fd8-fd4e-41de-8d24-c87c531489c1"/211563/0,/Min), ConditionalPut [/Table/51/1/2974175659068629674/"f40c3fd8-fd4e-41de-8d24-c87c531489c1"/211563/0,/Min), EndTransaction [/Table/51/1/2974175659068629674/"f40c3fd8-fd4e-41de-8d24-c87c531489c1"/211563/0,/Min)
W170318 09:45:58.054749 124 storage/node_liveness.go:293  [n1,hb] slow heartbeat took 3.1s
[...]

PR #18178, opened 2017-09-02
server,storage,sql: record node {de,re}commissioning in the event log

### LINKED ISSUE #17677: ui: add de / re commissioning nodes to event log
We want to surface important events in the event log - we currently include node joins, and should also include de / re commissioning events. 

This is the final outstanding item in #6198 (comment).

benesch:
Fixes #17677.

/cc @a-robinson @tschottdorf can one or both of you review? This wasn't quite as simple as I'd hoped. The naive solution spams the event log with duplicate entries because `./node decommission FOO` fires off many decommissioning requests in quick succession.

Also, note that `TestDecommission` is currently `t.Skip`ped, so the test won't run until #17995 is fixed. (I did manually verify my addition by replacing the `decommission --wait all` that hangs forever with a `decommission --wait none` and verifying that the `CheckQueryResults` succeeds.)

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1156,9 +1156,30 @@ func (s *Server) Undrain(off []serverpb.DrainMode) []serverpb.DrainMode {
 
 // Decommission idempotently sets the decommissioning flag for specified nodes.
 func (s *Server) Decommission(ctx context.Context, setTo bool, nodeIDs []roachpb.NodeID) error {
+	eventLogger := sql.MakeEventLogger(s.leaseMgr)
+	eventType := sql.EventLogNodeDecommissioned
+	if !setTo {
+		eventType = sql.EventLogNodeRecommissioned
+	}
 	for _, nodeID := range nodeIDs {
-		if err := s.nodeLiveness.SetDecommissioning(ctx, nodeID, setTo); err != nil {
+		changeCommitted, err := s.nodeLiveness.SetDecommissioning(ctx, nodeID, setTo)
+		if err != nil {
 			return errors.Wrapf(err, "during liveness update %d -> %t", nodeID, setTo)
+		}
+		if changeCommitted {
+			// If we die right now or if this transaction fails to commit, the
+			// commissioning event will not be recorded to the event log. While we
+			// could insert the event record in the same transaction as the liveness
+			// update, this would force a 2PC and potentially leave write intents in
+			// the node liveness range. Better to make the event logging best effort
+			// than to slow down future node liveness transactions.
+			if err := s.db.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
+				return eventLogger.InsertEventRecord(
+					ctx, txn, eventType, int32(nodeID), int32(s.NodeID()), struct{}{},
+				)
+			}); err != nil {
+				log.Errorf(ctx, "unable to record %s event for node %d: %s", eventType, nodeID, err)
+			}
 		}
 	}
 	return nil
--- a/pkg/ui/src/util/eventTypes.ts
+++ b/pkg/ui/src/util/eventTypes.ts
@@ -33,11 +33,15 @@ export const FINISH_SCHEMA_CHANGE = "finish_schema_change";
 export const NODE_JOIN = "node_join";
 // Recorded when an existing node rejoins the cluster after being offline.
 export const NODE_RESTART = "node_restart";
+// Recorded when a node is marked as decommissioning.
+export const NODE_DECOMMISSIONED = "node_decommissioned";
+// Recorded when a decommissioned node is recommissioned.
+export const NODE_RECOMMISSIONED = "node_recommissioned";
 // Recorded when a cluster setting is changed.
 export const SET_CLUSTER_SETTING = "set_cluster_setting";
 
 // Node Event Types
-export const nodeEvents = [NODE_JOIN, NODE_RESTART];
+export const nodeEvents = [NODE_JOIN, NODE_RESTART, NODE_DECOMMISSIONED, NODE_RECOMMISSIONED];
 export const databaseEvents = [CREATE_DATABASE, DROP_DATABASE];
 export const tableEvents = [CREATE_TABLE, DROP_TABLE, ALTER_TABLE, CREATE_INDEX,
   DROP_INDEX, CREATE_VIEW, DROP_VIEW, REVERSE_SCHEMA_CHANGE, FINISH_SCHEMA_CHANGE];
--- a/pkg/acceptance/decommission_test.go
+++ b/pkg/acceptance/decommission_test.go
@@ -15,6 +15,7 @@
 package acceptance
 
 import (
+	"fmt"
 	"regexp"
 	"strconv"
 	"testing"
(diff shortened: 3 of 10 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-:-Kt8NR3Ytbc0KacQmSor:bnfp4nl)*

pkg/server/server.go
```diff
@@ -1156,10 +1156,31 @@ func (s *Server) Undrain(off []serverpb.DrainMode) []serverpb.DrainMode {
...
+			// could insert the event record in the same transaction as the liveness
+			// update, this would force a 2PC and potentially leave write intents in
+			// the node liveness range. Better to make the event logging best effort
+			// than to slow down future node liveness transactions.
+			if err := s.db.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
+				return eventLogger.InsertEventRecord(
+					ctx, txn, eventType, int32(nodeID), int32(nodeID), struct{}{},
+				)
+			}); err != nil {
+				log.Errorf(ctx, "unable to record commissioning event for node %d: %s", nodeID, err)
```
a-robinson: nit, but it'd be nice to include the event type here as well.

pkg/ui/src/util/eventTypes.ts
```diff
@@ -33,11 +33,16 @@ export const FINISH_SCHEMA_CHANGE = "finish_schema_change";
 export const NODE_JOIN = "node_join";
 // Recorded when an existing node rejoins the cluster after being offline.
 export const NODE_RESTART = "node_restart";
+// Recorded when a node is marked as decommissioning.
+export const NODE_DECOMMISSIONED = "node_decommissioned";
+// EventLogNodeRecommissioned is recorded when a decommissioned node is
```
a-robinson: s/EventLogNodeRecommissioned is recorded/Recorded/

pkg/server/server.go
```diff
@@ -1156,10 +1156,31 @@ func (s *Server) Undrain(off []serverpb.DrainMode) []serverpb.DrainMode {
...
+		if changeCommitted {
+			// If we die right now or if this transaction fails to commit, the
+			// commissioning event will not be recorded to the event log. While we
+			// could insert the event record in the same transaction as the liveness
+			// update, this would force a 2PC and potentially leave write intents in
+			// the node liveness range. Better to make the event logging best effort
+			// than to slow down future node liveness transactions.
+			if err := s.db.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
+				return eventLogger.InsertEventRecord(
+					ctx, txn, eventType, int32(nodeID), int32(nodeID), struct{}{},
```
a-robinson: The `reporting` argument shouldn't be the `nodeID` being commissioned, it should be `s.NodeID()`.

a-robinson (approved):
Nice solution! LGTM, just a few small comments.

tschottdorf:
:lgtm:

---

---

*[pkg/server/server.go, line 1172 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-KtEgPSrG7GZzb29_5bh:-KtEgPSrG7GZzb29_5bi:b-6vnv9x) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d72faac3f8b076ef3c92b511119cac6f866abd9/pkg/server/server.go#L1172)):*
> ```Go
> 			// If we die right now or if this transaction fails to commit, the
> 			// commissioning event will not be recorded to the event log. While we
> 			// could insert the event record in the same transaction as the liveness
> ```

There's also the problem that intents on the node liveness range are not allowed. Not 100% sure that's true, though I think I remember it.

---

*[pkg/ui/src/views/cluster/containers/events/index.tsx, line 110 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-KtEhTg8IkKLgx75ncWJ:-KtEhTg8IkKLgx75ncWK:b-xq7g4t) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d72faac3f8b076ef3c92b511119cac6f866abd9/pkg/ui/src/views/cluster/containers/events/index.tsx#L110)):*
> ```XML
>       break;
>     case eventTypes.NODE_DECOMMISSIONED:
>       content = <span>Node Decommissioned: Node {targetId} was decommissioned</span>;
> ```

The message seems oddly repetitive, but apparently that's true for its neighbors as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-:-KtEhZRzkkT9qJeaKDY-:bnfp4nl)*

pkg/server/server.go
benesch: Nice catch! Done.

pkg/server/server.go
benesch: Done.

pkg/ui/src/util/eventTypes.ts
benesch: Thanks. Done.

tschottdorf:
Review status: 8 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/server/server.go, line 1172 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-KtEgPSrG7GZzb29_5bh:-KtHOntwbr139odJJkVj:b7jezsg) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d72faac3f8b076ef3c92b511119cac6f866abd9/pkg/server/server.go#L1172)):*

Is that enforced somewhere other than this block of code? https://github.com/cockroachdb/cockroach/blob/d65ed58d30ddf48ab50a9095260353f8056d50ae/pkg/storage/node_liveness.go#L647-L660
</blockquote></details>

Yep, that's the location.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178)*

benesch:
Ok, merging before someone touches embedded.go!

---

Review status: 8 of 11 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/server/server.go, line 1172 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-KtEgPSrG7GZzb29_5bh:-KtHHkRmUstuIXbOus04:b68a50w) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d72faac3f8b076ef3c92b511119cac6f866abd9/pkg/server/server.go#L1172)):*

Yep, that's the location.
</blockquote></details>

Let me know if you have a rewording for this comment, @tschottdorf. It seems correct to me, even with the above comment in mind.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-:-KtHg0FjF6oBrZc35Yr3:bxb2ffb)*

tschottdorf:
Review status: 8 of 11 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/server/server.go, line 1172 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178#-KtEgPSrG7GZzb29_5bh:-KtHpwtk7HRugvo4WvQt:b-4f783k) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d72faac3f8b076ef3c92b511119cac6f866abd9/pkg/server/server.go#L1172)):*

Let me know if you have a rewording for this comment, @tschottdorf. It seems correct to me, even with the above comment in mind.
</blockquote></details>

It's incorrect in that it suggests that it would slow down node liveness when in fact it would break it. But don't think that matters here.
See https://github.com/cockroachdb/cockroach/issues/14250#issuecomment-287839979 for archaeology.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18178)*

PR #18181, opened 2017-09-02
build: don't put .node_modules/bin on the path

benesch:
Since Yarn 0.25 (yarnpkg/yarn#3310), Yarn transitively symlinks binaries
into node_modules/.bin. (Previously, it would only sync top-level
binaries--i.e., binaries that we directly depend on.) The new behavior
results in adding all sorts of junk to the path--notably, a JavaScript
which reimplementation that shadows the system-provided which. Ew.

Avoid putting node_modules/.bin on the PATH by using full paths to the
binaries within, or using `yarn run` if within pkg/ui.

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -48,14 +48,7 @@ GO_INSTALL := GOBIN='$(LOCAL_BIN)' $(GO) install
 
 # Prefer tools we've installed with go install and Yarn to those elsewhere on
 # the PATH.
-#
-# Usually, we could use the yarn run command to avoid changing the PATH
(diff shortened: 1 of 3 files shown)


knz (approved):
Thanks, this makes me happy today. :sparkling_heart:

bdarnell:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18181#-:-Kt8NpbFKNX5Nq1CMcys:bnfp4nl)*

PR #18182, opened 2017-09-02
storage: fix quiescence when not leaseholder

### LINKED ISSUE #17741: stability: Zero QPS on cyan
The cyan (continuous deployment) cluster has been seeing zero qps for most of the last four hours. 

image

The start of the problems corresponds to an upgrade spanning the following range of commits (more than we'd like to see in continuous deployment, but the CI server was having trouble): https://github.com/cockroachdb/cockroach/compare/8cf21414b57403d4fe48598889ea26194b699555...c13339d6609bf3ed596aa187e8c92f40a901b150

petermattis:
Fix check for whether we own a valid lease.

Fixes #17741

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -1142,6 +1142,10 @@ func (r *Replica) getLeaseRLocked() (roachpb.Lease, *roachpb.Lease) {
 func (r *Replica) OwnsValidLease(ts hlc.Timestamp) bool {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
+	return r.ownsValidLeaseRLocked(ts)
+}
+
+func (r *Replica) ownsValidLeaseRLocked(ts hlc.Timestamp) bool {
 	return r.mu.state.Lease.OwnedBy(r.store.StoreID()) &&
 		r.leaseStatus(*r.mu.state.Lease, ts, r.mu.minLeaseProposedTS).state == leaseValid
 }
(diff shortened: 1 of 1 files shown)


bdarnell:
:lgtm:

We should go ahead and merge this (and assuming it fixes the problem, patch it in for the beta), but we need a follow-up issue to add more testing for quiescence and specifically a test that would be sensitive to this issue. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18182#-:-Kt3Al01McS-Ebdii1zR:b8drpdv)*

petermattis:
Merging and filing an issue to test this area of quiescence. I won't be able to cherry-pick into the release branch until later tonight (after 8pm).

bdarnell:
OK, I'll do the cherry-pick in #18184

bdarnell:
`adriatic` had been in the zero qps state and recovered when this change was deployed (not conclusive since it doesn't always get stuck, but a good sign). `omega` and `cerulean` were fine before this change was deployed and fine after. `omega` appeared to suffer more disruption to traffic during the push than usual.

petermattis:
@bdarnell `adriatic` and `omega` are both running 8633ca46d53c987b229562c0652694c53f3bae79 which does have your cherry-pick (3cab35b14efad1d86f8847eade2c0615f47fd4e5).

bdarnell:
> adriatic and omega are both running 8633ca4 which does have your cherry-pick (3cab35b).

That's what I mean. I was describing the push of 8633ca4 to the test clusters.

PR #18190, opened 2017-09-03
storage: avoid leaking temp dirs with in-memory stores

### LINKED ISSUE #18189: storage: in-mem rocksdb causes directory leak
- Each time an in-mem `RocksDB` is created, it creates both `/tmp/.inmemrocksdbNNNN` *and* `/tmp/.inmemrocksdbNNN/.auxiliary`
- Each time an in-mem `RocksDB` is closed, it only deletes the latter, `/tmp/.inmemrocksdbNNN/.auxiliary`

All the `/tmp/.inmemrocksdbNNN` are preserved.

For example when running tests this causes the number of (hidden) subdirectories in `/tmp` to increase quickly.

This, in turn, in many filesystems will cause filesystem accesses to `/tmp` to become slower and slower, due to the decreasing speed of directory lookups.

knz:
Fixes  #18189.

### DIFF
--- a/pkg/storage/engine/rocksdb.go
+++ b/pkg/storage/engine/rocksdb.go
@@ -377,11 +377,11 @@ func newMemRocksDB(
 		cache: cache.ref(),
 	}
 
-	tempDir, err := ioutil.TempDir(os.TempDir(), ".inmemrocksdb")
+	auxDir, err := ioutil.TempDir(os.TempDir(), "cockroach-auxiliary")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18190#-:-Kt46Ftb2MtGNcow0FY3:bnfp4nl)*

tbg (approved):
Thanks!

PR #18155, opened 2017-09-01
storage: bump intent resolution timeout and respect caller's timeout

nvb:
Fixes #18015.

Previously, the timeout for intent resolution was too low. It was only
`1*NetworkTimeout`, even though, at a minimum, intent resolution
performs two operations. We bump this timeout up by an order of
magnitude.

We now also take a context as an argument in `processIntentsAsync`. This
allows intent resolution to respect the caller's timeout when processing
intents synchronously.

PR #18191, opened 2017-09-03
storage: don't retry non-queueable PushTxnRequests

petermattis:
Don't retry non-queueable PushTxnRequests which push immediately, such
as PUSH_TOUCH requests. Doing so causes a fast failure loop where the
PushTxnRequest continually fails and retries until it eventually times
out. This was the cause of TestReplicaDanglingMetaIntent taking 60s and
TestStoreReadInconsistent taking 30s. In both cases, the IntentResolver
had issued a PushTxnRequest with a 30s timeout which was failing and
then getting retried in a tight loop by Store.Send. This shaves 90s off
the time it takes to run the `storage` tests.

Note that #18155 bumped the intent resolver timeout from 3s to 30s which
exacerbated the test time, but the underlying issue has been present for
a long time.

### DIFF
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -2561,13 +2561,17 @@ func (s *Store) Send(
 		// retry. Other errors are returned to caller.
 		switch pErr.GetDetail().(type) {
 		case *roachpb.TransactionPushError:
-			// On a transaction push error, retry immediately. This will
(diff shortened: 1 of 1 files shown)


tbg (approved):
Thanks!

nvb (approved):

PR #18196, opened 2017-09-03
cherrypick: storage: don't retry non-queueable PushTxnRequests

petermattis:


petermattis:
While I noticed the problem when investigating the speed of a test, this could definitely affect any `PUSH_TOUCH` `PushTxnRequest`.

bdarnell (approved):
LGTM, but remember to select the target branch when submitting cherry-picks.

PR #18197, opened 2017-09-04
storage: fork the context span for range lookups

### LINKED ISSUE #18115: testrace: race detected in  TestReplicateReAddAfterDown
Found in https://teamcity.cockroachdb.com/viewLog.html?buildId=340677&

```
==================
Write at 0x00c420493c28 by goroutine 122:
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.(*trace).reset()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:721 +0x182
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.freeTrace()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:879 +0x57
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.(*trace).unref()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:838 +0x9b
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.(*traceBucket).Add()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:578 +0x133
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.(*trace).Finish()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:386 +0x41f
  github.com/cockroachdb/cockroach/pkg/util/tracing.(*span).FinishWithOptions()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/tracing/tracer_span.go:262 +0x192
  github.com/cockroachdb/cockroach/pkg/util/tracing.(*span).Finish()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/tracing/tracer_span.go:246 +0x63
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).StartHeartbeat.func1.1()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:349 +0x53d
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).StartHeartbeat.func1()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:349 +0x37c
  github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).RunWorker.func1()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:198 +0x160

Previous read at 0x00c420493c28 by goroutine 1112:
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.(*trace).addEvent()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:767 +0x1b0
  github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace.(*trace).LazyPrintf()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/net/trace/trace.go:801 +0xdd
  github.com/cockroachdb/cockroach/pkg/util/tracing.(*span).LogFields()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/tracing/tracer_span.go:340 +0xce2
  github.com/cockroachdb/cockroach/pkg/util/log.eventInternal()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/log/trace.go:131 +0x2ca
  github.com/cockroachdb/cockroach/pkg/util/log.Event()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/log/trace.go:157 +0x78
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeDescriptorCache).lookupRangeDescriptorInternal()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_cache.go:353 +0xcaa
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeDescriptorCache).LookupRangeDescriptor()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_cache.go:241 +0xb3
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeDescriptorCache).performRangeLookup()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_cache.go:412 +0x6c9
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeDescriptorCache).lookupRangeDescriptorInternal.func3()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_cache.go:284 +0x11e
  github.com/cockroachdb/cockroach/vendor/golang.org/x/sync/singleflight.(*Group).doCall()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/sync/singleflight/singleflight.go:93 +0x4c
[23:44:27][TestReplicateReAddAfterDown] [Test Output]
Goroutine 122 (running) created at:
  github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).RunWorker()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:199 +0xba
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).StartHeartbeat()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:358 +0x2bc
  github.com/cockroachdb/cockroach/pkg/storage_test.(*multiTestContext).addStore()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/client_test.go:824 +0x1b30
  github.com/cockroachdb/cockroach/pkg/storage_test.(*multiTestContext).Start()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/client_test.go:296 +0xc06
  github.com/cockroachdb/cockroach/pkg/storage_test.TestReplicateReAddAfterDown()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/client_raft_test.go:2985 +0x127
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:657 +0x107

Goroutine 1112 (finished) created at:
  github.com/cockroachdb/cockroach/vendor/golang.org/x/sync/singleflight.(*Group).DoChan()
      /go/src/github.com/cockroachdb/cockroach/vendor/golang.org/x/sync/singleflight/singleflight.go:86 +0x517
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeDescriptorCache).lookupRangeDescriptorInternal()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_cache.go:334 +0x7d3
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeDescriptorCache).LookupRangeDescriptor()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_cache.go:241 +0xb3
  github.com/cockroachdb/cockroach/pkg/kv.(*DistSender).getDescriptor()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/dist_sender.go:416 +0xc9
  github.com/cockroachdb/cockroach/pkg/kv.(*RangeIterator).Seek()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/range_iter.go:171 +0x4c5
  github.com/cockroachdb/cockroach/pkg/kv.(*DistSender).divideAndSendBatchToRanges()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/dist_sender.go:710 +0x35f
  github.com/cockroachdb/cockroach/pkg/kv.(*DistSender).Send()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/dist_sender.go:597 +0x3ed
  github.com/cockroachdb/cockroach/pkg/kv.(*TxnCoordSender).Send()
      /go/src/github.com/cockroachdb/cockroach/pkg/kv/txn_coord_sender.go:444 +0x281
  github.com/cockroachdb/cockroach/pkg/internal/client.(*DB).send()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/db.go:545 +0x209
  github.com/cockroachdb/cockroach/pkg/internal/client.(*Txn).Send()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/txn.go:908 +0x577
  github.com/cockroachdb/cockroach/pkg/internal/client.(*Txn).Send-fm()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/txn.go:486 +0xa3
  github.com/cockroachdb/cockroach/pkg/internal/client.sendAndFill()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/db.go:454 +0x155
  github.com/cockroachdb/cockroach/pkg/internal/client.(*Txn).Run()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/txn.go:486 +0x111
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).updateLivenessAttempt.func1()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:662 +0x598
  github.com/cockroachdb/cockroach/pkg/internal/client.(*DB).Txn.func1()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/db.go:507 +0x5e
  github.com/cockroachdb/cockroach/pkg/internal/client.(*Txn).Exec()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/txn.go:722 +0xc5
  github.com/cockroachdb/cockroach/pkg/internal/client.(*DB).Txn()
      /go/src/github.com/cockroachdb/cockroach/pkg/internal/client/db.go:508 +0x174
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).updateLivenessAttempt()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:663 +0x18c
[23:44:27][TestReplicateReAddAfterDown] [Test Output]
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).updateLiveness()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:613 +0x7d
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).heartbeatInternal()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:438 +0x4a9
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).StartHeartbeat.func1.1()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:338 +0x30d
  github.com/cockroachdb/cockroach/pkg/storage.(*NodeLiveness).StartHeartbeat.func1()
      /go/src/github.com/cockroachdb/cockroach/pkg/storage/node_liveness.go:349 +0x37c
  github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).RunWorker.func1()
      /go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:198 +0x160
==================
```

petermattis:
Deflakes TestReplicateReAddAfterDown.

Fixes #18115

### DIFF
--- a/pkg/kv/range_cache.go
+++ b/pkg/kv/range_cache.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 
 	"github.com/biogo/store/llrb"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	"golang.org/x/sync/singleflight"
@@ -281,6 +282,18 @@ func (rdc *RangeDescriptorCache) lookupRangeDescriptorInternal(
 
 	requestKey := makeLookupRequestKey(key, evictToken, useReverseScan)
 	resC := rdc.lookupRequests.DoChan(requestKey, func() (interface{}, error) {
+		// We're inside a goroutine here which makes it an error to use a span
+		// embedded inside ctx. But we want to remain hooked up to the parent
+		// context for cancellation. The nature of the singleflight interface
+		// prevents us from using tracing.ForkCtxSpan() because we're not
+		// guaranteed that this closure is run nor do we get a signal that the
+		// closure is not run. As a workaround, we hide the span in the context.
+		//
+		// TODO(nvanbenschoten): Figure out some way to not lose tracing continuity
+		// here.
+		ctx := ctx // disable shadows linter
+		ctx = opentracing.ContextWithSpan(ctx, nil)
+
 		rs, preRs, err := rdc.performRangeLookup(ctx, key, useReverseScan)
 		if err != nil {
 			return nil, err


petermattis:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-Kt9JKwK6fXEWEuIqYAs:bic7s7w) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*
> ```Go
> 	reqCtx, reqSpan := tracing.ForkCtxSpan(ctx, "range lookup")
> 	resC := rdc.lookupRequests.DoChan(requestKey, func() (interface{}, error) {
> 		defer tracing.FinishSpan(reqSpan)
> ```

I'm almost positive this isn't correct, but I'm not sure how to actually finish the span at the correct time. We could wait until the read from `resC`, but that doesn't work in the presence of the context being cancelled.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

nvanbenschoten:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-Kt9QYWuU6Gfo3h-XRnC:bsdcscv) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

I'm almost positive this isn't correct, but I'm not sure how to actually finish the span at the correct time. We could wait until the read from `resC`, but that doesn't work in the presence of the context being cancelled.
</blockquote></details>

Yeah, I don't think this works because not all requests will actually enter this closure. Why were you unsure about forking the context inside here instead? That seems like the correct approach to me. Are you trying to share the span between all coalesced requests or something?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

petermattis:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtCHOcYaOLKfhSjscfq:bnfh6mv) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

Yeah, I don't think this works because not all requests will actually enter this closure. Why were you unsure about forking the context inside here instead? That seems like the correct approach to me. Are you trying to share the span between all coalesced requests or something?
</blockquote></details>

The other uses of `ForkCtxSpan` call that method before creating the goroutine. I can't tell from the implementation or documentation of `ForkCtxSpan` if it is kosher to call it from within the goroutine. My guess is know because doing so accesses the span that is tied to the parent goroutine. @raduberinde?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

bdarnell:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtDQ4cH7jiKgcFeRi6K:byasi0j) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

The other uses of `ForkCtxSpan` call that method before creating the goroutine. I can't tell from the implementation or documentation of `ForkCtxSpan` if it is kosher to call it from within the goroutine. My guess is know because doing so accesses the span that is tied to the parent goroutine. @raduberinde?
</blockquote></details>

Yeah, the goroutine can be delayed arbitrarily so I don't think it's safe to put `ForkCtxSpan` inside it. The `singleflight` `DoChan` API should really be extended with an indication of whether your callback will be called or if it was abandoned. Without that, I think the only way to make this safe is to do the cleanup after reading from the returned channel (spawning a goroutine if necessary in the event of context cancellation). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

nvanbenschoten:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtDj16O6QqF8FnMz7TK:b-v144ur) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

Yeah, the goroutine can be delayed arbitrarily so I don't think it's safe to put `ForkCtxSpan` inside it. The `singleflight` `DoChan` API should really be extended with an indication of whether your callback will be called or if it was abandoned. Without that, I think the only way to make this safe is to do the cleanup after reading from the returned channel (spawning a goroutine if necessary in the event of context cancellation). 
</blockquote></details>

Ah yeah, I see what you're saying @bdarnell. If `singleflight` returned whether the callback will be called (which it has [direct access to](https://github.com/golang/sync/blob/1ae7c7b29e06598039be46c5083819ba6fd7a97e/singleflight/singleflight.go#L67)), we could `FinishSpan` after either select case below. We could avoid the goroutine in the event of context cancellation because we'd only need the `<-ctx.Done()` case for follower requests.

Actually, if we only allowed the `<-ctx.Done()` case for follower requests (because leader requests would propagate their context cancellation error through `res = <-resC`), would we need to use `ForkCtxSpan` at all? That would guarantee that `DoChan's` goroutine never outlives the leader's time in `lookupRangeDescriptorInternal`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

petermattis:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtEOUrT4bAFZ1X4Ll9U:b-9zcnt0) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

Ah yeah, I see what you're saying @bdarnell. If `singleflight` returned whether the callback will be called (which it has [direct access to](https://github.com/golang/sync/blob/1ae7c7b29e06598039be46c5083819ba6fd7a97e/singleflight/singleflight.go#L67)), we could `FinishSpan` after either select case below. We could avoid the goroutine in the event of context cancellation because we'd only need the `<-ctx.Done()` case for follower requests.

Actually, if we only allowed the `<-ctx.Done()` case for follower requests (because leader requests would propagate their context cancellation error through `res = <-resC`), would we need to use `ForkCtxSpan` at all? That would guarantee that `DoChan's` goroutine never outlives the leader's time in `lookupRangeDescriptorInternal`.
</blockquote></details>

As a stop-gap solution, can I create a new span before creating the goroutine using `opentracing.ContextWithSpan(ctx, nil)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

nvanbenschoten:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtEUerZK9JkqKvhvu0A:b3cinx2) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

As a stop-gap solution, can I create a new span before creating the goroutine using `opentracing.ContextWithSpan(ctx, nil)`?
</blockquote></details>

You mean a new `Context`? Sure, but could you leave me a TODO to clean this up so that we don't lose tracing continuity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

petermattis:

---

*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtEVFIk2vjDrgaqFbuP:bmtus86) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

You mean a new `Context`? Sure, but could you leave me a TODO to clean this up so that we don't lose tracing continuity?
</blockquote></details>

Yep, a new `Context`. This is done. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

nvanbenschoten:
:lgtm:

---

---

*[pkg/kv/range_cache.go, line 292 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-KtEcy0AaiF7nhXykrWL:-KtEcy0AaiF7nhXykrWM:b51n7qd) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7ebce4fe7a4b0f2e2d81283ce3976d3a39fc266/pkg/kv/range_cache.go#L292)):*
> ```Go
> 		// closure is not run. As a workaround, we hide the span in the context.
> 		//
> 		// TODO(nathan): Figure out some way to not lose tracing continuity here.
> ```

`s/nathan/nvanbenschoten/`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-:-KtEd68mCCwUobDaeLQC:bnfp4nl)*

petermattis:

---

*[pkg/kv/range_cache.go, line 292 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-KtEcy0AaiF7nhXykrWL:-KtEdVE_U8q-LaO_U7z8:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7ebce4fe7a4b0f2e2d81283ce3976d3a39fc266/pkg/kv/range_cache.go#L292)):*

`s/nathan/nvanbenschoten/`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

RaduBerinde:
*[pkg/kv/range_cache.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197#-Kt9JKwK6fXEWEuIqYAr:-KtHC3nZXRU-7V0HsVeo:brfthej) ([raw file](https://github.com/cockroachdb/cockroach/blob/6fa8db1a8618d995bcd9be85c594f8142cc0b065/pkg/kv/range_cache.go#L289)):*

Yep, a new `Context`. This is done. PTAL.
</blockquote></details>

You could get the span context and then use `StartSpan` with `FollowsFrom` with that context instead of `ForkCtxSpan`. Basically what `ForkCtxSpan` but with the `Context()` call done earlier. The span context should still be valid even if the span it was derived from finishes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18197)*

PR #17758, opened 2017-08-19
sql/ir: teach the generator about field packing and primitive types

knz:
**Note to reviewers: the patch appears large, but most of it is auto-generated code. Don't be intimidated!**

There are four complementary parts in this patch:

- the generator now predefines the common Go primitive types for
  every input definition file (bool, string, int64, etc.)

- the code generator is extended to take a few options on the command
  line that control its behavior. The main() function is refactored to
  make it more readable.

- the code generator is taught about field packing, i.e. storing
  multiple small numeric values in a variable of a larger size. See a copy
  of the explanation below.

- the Makefile is modified to use multiple configurations to generate
  test IR environments. This is used to exercise different
  combinations of the code generation parameters, to ensure that all
  of them produce valid code.

A copy of the explanatory comment, that outlines the allocation of
memory slots for IR struct types, follows.

“The interesting part of code generation is fitting structs into nodes.

Each [IR] node consists of slots, i.e. spaces in memory where to put
values. The goal of slot allocation is to decide which struct field goes
to which slot.

### Overview

There are four kinds of slots: numeric, string, references and extra.
The numeric, string and reference slots are called "dedicated".
Dedicated slots come in finite amount!
For example, in the default configuration, there are 2 numeric slots,
1 string slot and 2 reference slots.

In general, we prefer a dedicated slot. When dedicated slots are
exhausted for a particular type (e.g. when encountering the 3rd
numeric field in a struct in the default configuration), we spill
to the extra slots. Extra slots expand on demand without limit.

We support two modes: packed and unpacked.

Understanding unpacked mode can serve as foundation to better
understand packed mode.

### How unpacked mode works

In that mode, each numeric field uses one numeric slot; each
reference to a struct uses one reference slot; and each reference
to a sum uses both a numeric slot (for the tag) and a reference
slot (for the value). Every other type uses an extra
slot. When dedicated slots are exhausted, an extra slot is also
used. For example:

```go
type BinExprValue struct {
	Left  Expr
	Op    BinOp
	Right Expr
}
type BinExpr struct { *node }

//// Packing with 3 numeric slots:
func (x BinExprValue) R(a Allocator) BinExpr {
	node := a.new()
	node.nums[0] = numvalslot(x.Left.tag)
	node.nums[1] = numvalslot(x.Op)
	node.nums[2] = numvalslot(x.Right.tag)
	node.refs[0] = x.Left.ref
	node.refs[1] = x.Right.ref
	return BinExpr{node}
}
[...]

PR #18198, opened 2017-09-04
sql/ir: auto-generate an S-expression parser

knz:
This PR is based off the commits in #17758. Once that merges, this one will be rebased.

Previously an
[S-expression](https://en.wikipedia.org/wiki/S-expression)
pretty-printer was already generated. This patch extends that
previous code with the corresponding (super fast) parser.

This makes it possible to enter arbitrary trees of nodes using a
human-friendly string-based format and round-trip existing trees.

For example given the following mini-language generated in package
`ir`:

```
enum BinOp {  Add = 1;  Mul = 2 }
struct ConstExpr {  int64 Datum = 1 }
struct BinExpr {  Expr Left = 1;  BinOp Op = 2;  Expr Right = 3 }
sum Expr {  ConstExpr = 1;  BinExpr = 2 }
```

The following API is generated:

```go
package ir
type Parser
func (p *Parser) ParseExpr() (Expr, error)
func (p *Parser) ParseBinExpr() (BinExpr, error)
func (p *Parser) ParseConstExpr() (ConstExpr, error)
func (p *Parser) ParseBinOp() (BinOp, error)
```

And thus one can call:

```go
  s := `
  (bin-expr
     :op mul
     :left (const-expr :datum 3)
     :right (bin-expr
               :op add
               :left (const-expr :datum 1)
               :right (const-expr :datum 2)))
  `

  p := ir.MakeParser(s, irAlloc)
  expr, err := p.ParseExpr()
```

And then verify that `expr.String()` (which uses `FormatSExpr`
internally) has the same structure as the initial input:

```
(bin-expr :left (const-expr :datum 3) :op mul :right (bin-expr :left (const-expr :datum 1) :op add :right (const-expr :datum 2)))
```

(The fields inside a struct can be arbitrarily reordered on input, but
will be printed in definition order by `FormatSExpr`.)


petermattis:
S-expressions are human friendly? :smiley: 

---

Review status: 0 of 40 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18198#-:-KtHJQjQdnvE7A5YVZwk:b7vpbia)*

bdarnell:
@petermattis I was expecting that comment to come from a vim-using heathen, not a fellow emacs user :)

knz:
Superseded by #19135.

PR #18203, opened 2017-09-04
cherrypick: storage: Log a help message if raft data is missing

a-robinson:
A hacky stopgap to try to explain what's happening if someone runs
into #14231. It's not great, but seems better than leaving users with
a hard-to-understand raft error if they make a configuration mistake.

Cherrypicks #18172


a-robinson:
@cockroachdb/release

bdarnell (approved):

PR #10055, opened 2016-10-18
RFC [WIP] an intermediate representation for SQL

knz:
Reaction to last week's meeting (with @irfansharif @arjunravinarayan @petermattis @RaduBerinde @andreimatei)

@eisenstatdavid I'd like you to help me shape this up before we submit it for discussion to a larger group. Some special questions for you: 
- choice of HLL, would something else read better to the uneducated reader?
- how much more than this do we need to be able to define tree transformations in the HLL and have this translated to Go automatically?
- do you recall the algorithm to check whether a pattern match is exhaustive, or a reference to such an algorithm?

---

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach/10055)

PR #18204, opened 2017-09-04
rfc: auxiliary leaf data in SQL abstract trees

knz:
Discussed last week with @jordanlewis. 

Needed for #10055.

### DIFF
--- a/docs/RFCS/20170926_sql_aux_data.md
+++ b/docs/RFCS/20170926_sql_aux_data.md
@@ -1,0 +1,469 @@
+- Feature Name: Auxiliary leaf data in SQL abstract trees
+- Status: in-progress
+- Start Date: 2017-09-04
+- Authors: knz, jordan
+- RFC PR: #18204
+- Cockroach Issue: (one or more # from the issue tracker)
+
+# Summary
+
+This RFC proposes to extract leaf data from the tree data
+structures used to represent SQL syntax and logical plans, and instead:
+
+- host the values in slices in a context data structure passed as
+  argument to the functions where the values are needed/used;
+- in the tree, instead store an index into that slice.
+
+Why: this generally increases performance in several areas, and
+incidentally+serendipitously removes the cause for a sore point that
+prevented progress on the IR RFC (#10055).
+
+How: this is a mechanical, easy to review code substitution in the
+`sql/parser` and `sql` packages.
+
+Impact: performance + enables further IR work.
+
+# Motivation
+
+tl;dr: hosting leaf data outside of the logical tree makes
+things generally simpler and cleaner, which is desirable overall.
+
+Why is this so? Granted, it is hard to recognize this to be true in
+general. Concretely, there are four different sorts of leaf
+data items in SQL trees, and for each of them the motivation for this
+RFC can be phrased in a different way, which I detail below. However,
+after reading the four motivations the reader can satisfy themselves
+that the "tl'dr" summary above is adequate.
+
+The four sorts are, in decreasing order of "how well they justify this
+RFC and the corresponding changes":
+
+- placeholders
+- subqueries
+- column data references ("indexed vars")
+- datums
+
+## Motivation for placeholders
+
+Currently in CockroachDB placeholders in the input (`$1`, `$2` etc. in
+prepared queries) are replaced by a `Placeholder` instance in the
+tree:
+
+```go
+type Placeholder struct {
+	Name string
+	typeAnnotation
+}
+```
+
+During type checking, the `typeAnnotation` member gets filled in with
+the inferred/specified type for the placeholder. This field is
+subsequently used by the `ResolvedType()` method whenever the type is
+needed.
+
+Then, when the prepared query is executed, the entire AST is
+*rewritten* so that each `Placeholder` instance is replaced by the
+`Datum` for the value provided by the client. Since CockroachDB avoids
+in-place modifications to AST nodes, this rewrite requires
+re-allocating a fresh object on the heap for every ancestor of a
+placeholder up to the root of the query AST.
+
+What's wrong with this?
+
+- when a placeholder is used multiple times (e.g. `select $1 where x >
+  $1`), the type information is stored multiple times in the AST.
+- the tree rewrite at each execute of a prepared query is expensive
+  and puts pressure on Go's heap allocator and GC.
+
+## Motivation for subqueries
+
+Subqueries in SQL can be of two sorts:
+
+- subqueries as data sources, e.g. `select * from (select * from
+  kv)` - these are simply nested selects and are treated very
+  efficiently: their logical plan is simply embedded as the suitable
+  operand in the enclosing query's own plan.
+- subqueries as expressions. For example with `select * from kv limit
+  (select v from kv where k=2)`. Here the value of the result of
+  evaluating the subquery must be known before the rest of the plan
+  can be executed.
+
+This latter case is currently handled as follows:
+
+- when a query plan is initially constructed, first all the
+  expressions are traversed, and any sub-query expression is replaced
+  by an instance of `sql.subquery` (or `parser.SubqueryPlaceholder`
+  once #18094 is merged). *This object is embedded as leaf
+  data in the expression tree!*
+- when execution of a plan starts, again all expressions are
+  traversed, and any sub-query is executed in turn to completion, and
+  the query's results are used to *replace* the `subquery` object in
+  the expression tree. Since CockroachDB avoids in-place modifications
+  to AST nodes, this rewrite requires re-allocating a fresh object on
+  the heap for every ancestor of a subquery up to the root of the
+  query AST.
+- the EXPLAIN statement must also traverse all expressions in a logical
+  plan to "fish out" the logical plans of subqueries and embed them in the EXPLAIN output.
+
+What's wrong with this?
+
+- the various stages of subquery handling are really algorithms of the
+  form "for every subquery, *no matter where it appears*, do X". It is
+  silly to have to recursively traverse all expressions, *including
+  those that definitely do not contain subqueries*, to apply these
+  algorithms.
+- the rewrite of the subquery results, when applicable, is expensive
+  towards Go's heap allocator and GC.
+
+## Motivation for column data referencs ("indexed vars")
+
+Currently in CockroachDB an early transform called "name resolution"
+will replace any column reference by name (e.g. "`k`" in `select k
+from foo`) or by ordinal reference (e.g. `@1` in `select @1 from foo`)
+by an instance of `IndexedVar`:
+
+```go
+type IndexedVar struct {
+	Idx       int
+	container IndexedVarContainer
+}
+```
+
+This object contains the index of the column inside the "current data
+source context" (usually: the current table; this is only more complex
+with joins and UPSERT). It *also* contains a pointer to some other object, somewhere, that is able to:
+
+- serve the type of the column (via the `ResolvedType()` method);
+- serve the value of the column for the current row (via the `Eval()` method);
+- serve a representation of the column reference (via the `Format()` method).
+
+This `container` fields requires some care during transforms: if an
+expression is migrated from one level of a logical plan to another (a
+common occurrence during optimizations), the `container` field must be
+suitably rewritten.
+
+This happens pretty often actually (`BindIfUnbound()` and `Rebind()`). Two aspects of note:
+
+- during initial name resolution, `BindIfUnbound()` is called, and
+  this actually overwrites `container` in-place. This incidentally
+  violates the rule that ASTs should be immutable once constructed!
+  And has caused bugs (and probably is causing bugs, due to our
+  inability to assert immutability).
+- `Rebind()` is called many times for the same expression in
+  moderately complex queries (most notoriously during filter
+  propagation), each time requires a full traversal of the entire
+  expression tree, and if an expression is found requires a
+  substitution. And as you can expect if you've read the two sections
+  above: since CockroachDB avoids in-place modifications to AST nodes,
+  this rewrite requires re-allocating a fresh object on the heap for
+  every ancestor of a subquery up to the root of the query AST.
+
+What's wrong with this?
+
+- `BindIfUnbound()` violates the immutability contract;
+- if the same column reference is used multiple times at the same
+  level of a logical plan, there are multiple redundant references
+  (`container` copies) to the data source.
+- `Rebind()` exerts pressure on the Go heap allocator and GC.
+
+## Motivation for Datums
+
+(This motivating section is a bit less evident for the casual reader
+with less experience with CockroachDB's SQL codebase. It is also not
+terribly important since the 3 sections above already motivate the
+common underlying pattern. Feel free to skip to the next sub-section.)
+
+CockroachDB currently defines 17 elementary Go types to hold SQL
+values (`DInt`, `DString`, `DInterval`, etc.), increasing to 18
+when #18171 merges, and possibly increasing further as we extend pg
+compatibility. Moreover, we are eventually planning to let users
+define their own value types.
+
+It is certainly instructive to ask: "Why?"
+
+For example, TiDB uses a single Go struct for all the values, storing
+the actual value in different fields of that struct depending on the
+semantic type of the SQL expression. (This is not where this RFC is
+going but this observation highlights that the current approach in
+CockroachDB is not trivially necessary.)
+
+There are two motivations for using separate Go types:
+
+- Go requires that values that are composed of a reference to
+  something else (e.g. `DString`, or a slice reference as in `DByte`),
+  and values that are composed of simple "value bits" (e.g. `DInt`,
+  `DFloat`) are stored in variables that have distinct Go types,
+  because Go's GC needs to differentiate statically value and
+  reference variables.
+- in many expression transformation algorithms in CockroachDB we must
+  discriminate in conditionals based on the semantic type of SQL
+  values. It so happens that Go is rather good at branching on type
+  tags ("type switches" in Go's jargon) so this vaguely suggests
+  separate Go types for each SQL value type.
+
+What's wrong with this?
+
+- `Datum` types are actually embedded in SQL expression trees using an
+  interface reference (`Datum` and/or `Expr`). This means that even
+  "value" types are never embedded as-is in the expression tree and
+  must be allocated on the heap instead. Worse even, a SQL value that
+  is itself a reference (e.g. `DBytes`) then requires a double
+  indirection: once to retrieve the Datum object from the Expr
+  reference, then another one to access the data behind that datum's
+  reference. Given the ubiquity of values this merits some attention
+  if a simplification is possible.
+- the value types are also used to report values in *result rows*
+  during plan execution, behind a `Datum` reference. In Go this means
+  that each time a row tuple is constructed, Go must assemble a
+  reference to the value together with the vtable pointer of the
+  `Datum` interface, to construct a `Datum` reference in each position
+  of the result tuple. Then whenever this value is consumed somewhere
+  else (either at a different level in the logical plan or when the
+  results are converted towards pgwire), Go must check that the type
+  cast is valid, this constitutes run-time overhead. One should note
+  here that this type dance is entirely *unnecessary*: from one row to
+  the next, the type of the SQL value for a given result column is
+  *always the same* (*). Storing then checking the Go type information in
+  memory for every cell of every row in a plan's result column is
+  horrendously redundant.
+
+(*) Regarding the handling of NULL values: NULL is a value that
+inhabits every type. That is, `NULL::int` and `NULL::string` are two
+different things in SQL. Even in a column where NULL *values* can
+occur, all values in the columns should still be of the same type.
+This is not currently true in CockroachDB, some suggestions are made
+below to arrive there.
+
+## Synthesis: what's the underlying problematic pattern?
+
+Discriminating different things in a tree data structure in Go using
+different Go types implementing a common "node interface" is a
+textbook design pattern. It is simple to understand, simple to
+implement, simple to recognize, and "just works".
+
+However it breaks down when any of the following conditions apply:
+
+- the application domain strongly favors/recommends immutable trees,
+  but the logical values in the trees require replacement between
+  uses. This requires an expensive "rewrite via new allocation" traversal.
+- some algorithms need to perform an action for every node in a tree of a
+  specific type, without knowning in advance where these values are,
+  and this types happens to be rather uncommon in trees in
+  practice. This causes "work for nothing": the common case pays the
+  price of full tree traversals even when the application domain tells us
+  in advance that these full traversals are usually unnecessary.
+- when many objects *are known to have the same type* (e.g. values for
+  the same column in many rows, multiple instances of the same
+  placeholders in different places of an expression) it is wasteful to
+  duplicate this type information across all instances.
+
+In CockroachDB, all three conditions apply. So there's a problem.
+
+# Guide-level explanation
+
+When this proposal is implemented, CockroachDB will replace the
+"embedding" of special values in IR trees by a simple integer value,
+to serve as index in an array of appropriate type outside of the tree.
+
+For example, before:
+
+```go
+e := BinExpr{Left: IndexedVar{Idx:1}, Right: IndexedVar{Idx:2}, Op: Add}
+
+// Link IndexedVars to containers.
+ivarHelper := MakeIndexedVarHelper(container)
+e = ivarHelper.Rebind(e)
+
+// Use the Expr.
+fmt.Println(e.String()) // shows "x + y" instead of "@1 + @2"
+
+// This uses:
+// type IndexedVar struct {Idx int; container IndexedVarContainer}
+// type Expr interface { Format(buf *bytes.Buffer); String() string };
+// (X).String() calls (X).Format() for every X implementing Expr;
+// (iv *IndexedVar).Format() calls iv.container.FormatIndexedVar(iv.Idx).
+```
+
+After:
+
+```go
+e := BinExpr{Left: IndexedVarIdx(1), Right: IndexedVarIdx(2), Op: Add}
+
+// Use the Expr.
+fmt.Println(Format(e, container)) // shows "x + y" instead of "@1 + @2"
+
+// This uses:
+type IndexedVarIdx int
+type Expr { Format(buf *bytes.Buffer, container IndexedVarContainer) ... }
+func Format(e, container) {
+  e.Format(buf, container)
+}
+func (iv IndexedVarIdx) Format(buf, container) {
+  container.FormatIndexedVar(int(iv))
+}
+```
+
+In general, the tree structure will store only an integer, and the
+resolution of that integer to the "thing" that it logically refers to
+is only performed at the point of use, not stored in the tree
+directly. This enables minimal data storage in the tree and efficient
+substitutions of the corresponding values without having to mutate the
+tree.
+
+# Reference-level explanation
+
+## Detailed design
+
+| Current code | New code | Notes |
+|--------------|----------|-------|
+| `type IndexedVar struct {...}` | `type IndexedVarIdx int` | |
+| `type Placeholder struct {name, typ}` | `type PlaceholderIdx string` | (1) |
+| `type subquery struct {...}` | `type subquery int` | |
+| `type Datum interface { Expr; ... }` | `type Datum int` | (4) |
+| `(NodeFormatter).Format(buf, f)` | `(NodeFormatter).Format(ctx, buf)` | (2) (3) |
+| `(TypedExpr).ResolvedType()` | `(TypedExpr).ResolvedType(ctx)` | (3) |
+| `(Datum).AmbiguousFormat()` | `(Datum).AmbiguousFormat(ctx)` | (3) |
+| `(Datum).Prev()` | `(Datum).Prev(ctx)` | (3) |
+| `(Datum).Next()` | `(Datum).Prev(ctx)` | (3) |
+| `(Datum).IsMin()` | `(Datum).IsMin(ctx)` | (3) |
+| `(Datum).IsMax()` | `(Datum).IsMax(ctx)` | (3) |
+| `(Datum).min()` | `(Datum).min(ctx)` | (3) |
+| `(Datum).max()` | `(Datum).max(ctx)` | (3) |
+| `(Datum).Size()` | `(Datum).Size(ctx)` | (3) |
+
+Notes:
+
+1. about placeholders: CockroachDB currently identifies placeholders
+   by name. This is because the Postgres protocol, in principle,
+   allows placeholders with arbitrary names not just numbers. In
+   practice however, we never encountered a client that does so, and
+   the CockroachDB code even contains an assertion on the initial
+   lexing of placeholder names to force them to be numerical. Perhaps
+   it is time to drop the idea to name placeholders and instead number
+   them, which will in turn make the data structure even smaller and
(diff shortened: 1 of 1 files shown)


RaduBerinde:

---

*[docs/RFCS/sql_aux_data.md, line 226 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtHuE0Wez_cmqUJafQu:-KtHuE0Wez_cmqUJafQv:bautx9f) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L226)):*
> ```Markdown
>   here that this type dance is entirely *unnecessary*: from one row to
>   the next, the type of the SQL value for a given result column is
>   *always the same*. Storing then checking the Go type information in
> ```

One blemish here is `DNull` which is a separate type. So on some rows we may have `Dint` (for example), and on others `DNull`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

knz:
If there's an int, -1 can signal DNull.

knz:
And for result rows we can have a bitmap.

docs/RFCS/sql_aux_data.md
```diff
@@ -0,0 +1,449 @@
...
+## Motivation for Datums
+
+(This motivating section is a bit less evident for the casual reader
+with less experience with CockroachDB's SQL codebase. It is also not
+terribly important since the 3 sections above already motivate the
+common underlying pattern. Feel free to skip to the next sub-section.)
+
+CockroachDB currently defines 17 elementary Go types to hold SQL
+values (`DInt`, `DString`, `DInterval`, etc.), increasing to 18 when
+#18171 merges, and possibly increasing further as we extend pg
```
tschottdorf: You might have to escape this `#`.

tbg (commented):
Informative read, thanks for the write up!

jordanlewis:

---

*[docs/RFCS/sql_aux_data.md, line 389 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtI2lQtC77brDH8x8AQ:-KtI2lQu5DSttsMoSNw6:b-qazj6p) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L389)):*
> ```Markdown
> 
> - Slight deviation from 100-level data structure textbooks. (Although
>   the new propsoed pattern is not really unknown to 200- or 300-level
> ```

proposed*

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

petermattis:

---

*[docs/RFCS/sql_aux_data.md, line 422 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtIUDp7Upan0nIL23fY:-KtIUDp7Upan0nIL23fZ:b-97lhvj) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L422)):*
> ```Markdown
> 
> - do nothing: further SQL development and run-time performance slowly
>   grind to a halt.
> ```

Heh, that's a bit of hyperbole. Also, `s/grind/grinds/g`.

---

*[docs/RFCS/sql_aux_data.md, line 426 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtIa-_3D5Pa6A1HD25N:-KtIa-_3D5Pa6A1HD25O:btvrgqj) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L426)):*
> ```Markdown
>   which enables programmers to control allocations and avoid paying
>   run-time overhead for typing when it's known to be safe. This
>   alternative is serious but a longer-term endeavour. Also even if the
> ```

I think this actually minimizes the work involved. In addition to rewriting a ton of code, there would be significant technical obstacles involved in glueing it together with Go (efficiently). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

knz:
I extended the text to highlight how to treat NULLs properly. PTAL. If there are no more comments I'll start the FCP this week.

---

---

*[docs/RFCS/sql_aux_data.md, line 178 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtE7eAZExHT8UTZoEL_-r1-178:-KtfqxQY7MQ38i-tRpBu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L178)):*

You might have to escape this `#`.
</blockquote></details>

Done.

---

*[docs/RFCS/sql_aux_data.md, line 226 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtHuE0Wez_cmqUJafQu:-KtfrHZhYrfXIKlifdTE:b-6x11g7) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L226)):*

One blemish here is `DNull` which is a separate type. So on some rows we may have `Dint` (for example), and on others `DNull`.
</blockquote></details>

Added a paragraph to clarify here and below.

---

*[docs/RFCS/sql_aux_data.md, line 389 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtI2lQtC77brDH8x8AQ:-Ktfs1Gx3_npXBgPuFtc:bt2vr1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L389)):*

proposed*
</blockquote></details>

Fixed.

---

*[docs/RFCS/sql_aux_data.md, line 422 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtIUDp7Upan0nIL23fY:-KtfsAtq_CwrbWlXKPAL:b-2ixzkr) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L422)):*

Heh, that's a bit of hyperbole. Also, `s/grind/grinds/g`.
</blockquote></details>

Thanks! Fixed, and added a note to that extent.

---

*[docs/RFCS/sql_aux_data.md, line 426 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KtIa-_3D5Pa6A1HD25N:-KtfsMsiWu79YqQiVdyH:b-3ts5go) ([raw file](https://github.com/cockroachdb/cockroach/blob/5597e94576c0e1cdbf479182bcbd2ab4a1cef402/docs/RFCS/sql_aux_data.md#L426)):*

I think this actually minimizes the work involved. In addition to rewriting a ton of code, there would be significant technical obstacles involved in glueing it together with Go (efficiently). 
</blockquote></details>

Changed the phrasing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-:-KtfsY1A6aVc-3likghD:bjefhco)*

docs/RFCS/sql_aux_data.md
```diff
@@ -0,0 +1,464 @@
...
+in-place modifications to AST nodes, this rewrite requires
+re-allocating a fresh object on the heap for every ancestor of a
+placeholder up to the root of the query AST.
+
+What's wrong with this?
+
+- when a placeholder is used multiple times (e.g. `select $1 where x >
+  $1`), the type information is stored multiple times in the AST.
+- the tree rewrite at each execute of a prepared query is expensive
+  and puts pressure on Go's heap allocator and GC.
```
vivekmenezes: sounds like a big deal!

vivekmenezes (commented):
The motivation section is very well written! Awesome doc, thanks

LGTM

knz:
final comment period.

couchand:
Reviewed 1 of 1 files at r2.

---

*[docs/RFCS/sql_aux_data.md, line 207 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPwjdYf4q3Ux0PCGcg:-KuPwjdZ-f57x4zNDtO8:b-oz5lrf) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L207)):*
> ```Markdown
> 
> - `Datum` types are actually embedded in SQL expression trees using an
>   interface reference (`Datum` and/of `Expr`). This means that even
> ```

s/of/or/

---

*[docs/RFCS/sql_aux_data.md, line 228 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPxK4RQCWv4nfs_hxp:-KuPxK4SagPtpKNhVKlF:bdv5df2) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L228)):*
> ```Markdown
>   *always the same* (*). Storing then checking the Go type information in
>   memory for every cell of every row in a plan's result column is
>   horrendously redundant.
> ```

Isn't this essentially the interpreter/compiler tradeoff?  Perhaps I'm missing something in this discussion, but I don't see how the SQL execution engine could be written to be both general (supports all legal statements at runtime) and optimally efficient (not require boxing primitive data types).  I suppose maybe it would involve transposing the standard row-wise computation of results into a column-wise version that allows building a query plan that operates on just the data types of that particular sequence?  Or maybe JITting?

---

*[docs/RFCS/sql_aux_data.md, line 231 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPyURA85m_pnJYI8Vq:-KuPyURA85m_pnJYI8Vr:b-fkgsuc) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L231)):*
> ```Markdown
> 
> (*) Regarding the handling of NULL values: NULL is a value that
> inhabits every type. That is, `NULL::int` and `NULL::string` are two
> ```

Well, it might be nice to have nullable types defined to be distinct type (or use a Maybe type constructor), to, for instance, help enforce that non-nullable columns never receive NULL as a value.  But that's not really relevant to this discussion.

---

*[docs/RFCS/sql_aux_data.md, line 246 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzCRVrDGcX_qPvz9U:-KuPzCRVrDGcX_qPvz9V:byjb1kw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L246)):*
> ```Markdown
> - the application domain strongly favors/recommends immutable trees,
>   but the logical values in the trees require replacement between
>   uses. This requires an expensive "rewrite via new allocation" traversal.
> ```

If the "logical values" change, in what sense is the tree immutable?  If I can't replace the object at the leaf node, but I can still change it's values, aren't I committing just as grave of a sin?

---

*[docs/RFCS/sql_aux_data.md, line 256 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzUbWlPUPFk3AW9Fz:-KuPzUbWlPUPFk3AW9G-:b-s5mx42) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L256)):*
> ```Markdown
>   the same column in many rows, multiple instances of the same
>   placeholders in different places of an expression) it is wasteful to
>   duplicate this type information across all instances.
> ```

I think I see what you're getting at now.  In the case of `Datum`, even primitives would still be hidden behind the interface, but the type switch would no longer be necessary?  That doesn't seem like all that big of a win, but maybe I'm missing something.

---

*[docs/RFCS/sql_aux_data.md, line 308 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzpOkcLfRbUT_jcXy:-KuPzpOkcLfRbUT_jcXz:bdwboli) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L308)):*
> ```Markdown
> is only performed at the point of use, not stored in the tree
> directly. This enables minimal data storage in the tree and efficient
> substitutions of the corresponding values without having to mutate the
> ```

As I asked above, what's the real difference between "substitution" and "mutation" here?  Won't the same safety/bug questions arise from either strategy?

---

*[docs/RFCS/sql_aux_data.md, line 342 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuQ-Q5cnGspnT967bkd:-KuQ-Q5cnGspnT967bke:bty8vfs) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L342)):*
> ```Markdown
>    it is time to drop the idea to name placeholders and instead number
>    them, which will in turn make the data structure even smaller and
>    more efficient to use (lookups using an array instead of a map).
> ```

Do the placeholders have to be dense?  In practice they probably are.  But it seems worth noting the difference between the current enforcement of numeric identifiers and enforcing the smallest possible numeric identifiers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions.

---

*[docs/RFCS/sql_aux_data.md, line 207 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPwjdYf4q3Ux0PCGcg:-KuwLYrerBKn2tVINL7s:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L207)):*

s/of/or/
</blockquote></details>

Done.

---

*[docs/RFCS/sql_aux_data.md, line 228 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPxK4RQCWv4nfs_hxp:-KuwLizrSZDJ9IFkjqmH:b-q06uhd) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L228)):*
> Isn't this essentially the interpreter/compiler tradeoff? 

No, it's about choosing the right data structures. This always pays off regardless of the evaluation strategy of the language itself.

> Perhaps I'm missing something in this discussion, but I don't see how the SQL execution engine could be written to be both general (supports all legal statements at runtime) and optimally efficient (not require boxing primitive data types).  

You're right however I don't think this is relevant in this discussion.

> I suppose maybe it would involve transposing the standard row-wise computation of results into a column-wise version [...]

Yes eventually this is where the correct solution to that particular instance of the general problem is to be found. We have a roadmap item to go there eventually.

---

*[docs/RFCS/sql_aux_data.md, line 231 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPyURA85m_pnJYI8Vq:-KuwMEOTqx5Ic6JIXJzy:b-pbdqzg) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L231)):*

Well, it might be nice to have nullable types defined to be distinct type (or use a Maybe type constructor), to, for instance, help enforce that non-nullable columns never receive NULL as a value.  But that's not really relevant to this discussion.
</blockquote></details>

Agreed!

---

*[docs/RFCS/sql_aux_data.md, line 246 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzCRVrDGcX_qPvz9U:-KuwMJ_hq8Kn2j26eHtf:bonancw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L246)):*

If the "logical values" change, in what sense is the tree immutable?  If I can't replace the object at the leaf node, but I can still change it's values, aren't I committing just as grave of a sin?
</blockquote></details>

The issue is that the data structure instances in memory are/will be shared across multiple versions. E.g. with AST caching we want to reuse the tree data as much as possible even though each particular execution needs to rewrite the tree.

---

*[docs/RFCS/sql_aux_data.md, line 256 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzUbWlPUPFk3AW9Fz:-KuwMaQjvJRjfnw6IibA:brfo4qt) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L256)):*

I think I see what you're getting at now.  In the case of `Datum`, even primitives would still be hidden behind the interface, but the type switch would no longer be necessary?  That doesn't seem like all that big of a win, but maybe I'm missing something.
</blockquote></details>

Eventually the Datum interface would disappear.

---

*[docs/RFCS/sql_aux_data.md, line 308 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzpOkcLfRbUT_jcXy:-KuwMfLUhoU66ovmOQ6x:b-mno1mz) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L308)):*

As I asked above, what's the real difference between "substitution" and "mutation" here?  Won't the same safety/bug questions arise from either strategy?
</blockquote></details>

Tree updates by substitution allow sharing the nodes between multiple versions of the tree, see above.

---

*[docs/RFCS/sql_aux_data.md, line 342 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuQ-Q5cnGspnT967bkd:-KuwMmv_nNHyEQKas07L:bntyvi0) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/sql_aux_data.md#L342)):*

Do the placeholders have to be dense?  In practice they probably are.  But it seems worth noting the difference between the current enforcement of numeric identifiers and enforcing the smallest possible numeric identifiers.
</blockquote></details>

Added a sentence to clarify.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

couchand:
Reviewed 2 of 2 files at r3.

---

*[docs/RFCS/20170926_sql_aux_data.md, line 246 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzCRVrDGcX_qPvz9U:-Kuyf-FzUxTbvDSvftgB:b-3enqi3) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L246)):*

The issue is that the data structure instances in memory are/will be shared across multiple versions. E.g. with AST caching we want to reuse the tree data as much as possible even though each particular execution needs to rewrite the tree.
</blockquote></details>

I think I still take issue with considering this to still be an immutable tree.  We can actively make the performance/purity tradeoff, because sometimes it makes sense to sacrifice purity for the sake of performance.  But let's not pretend like we're not sacrificing purity.  And as long as we're at it, let's just go ahead and mutate the tree for other types of changes we're making.  I fail to see how splitting the difference here (unchanging tree structure with changing leaf values) maintains any of the bug-squashing advantages of immutable trees anymore.

---

*[docs/RFCS/20170926_sql_aux_data.md, line 256 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzUbWlPUPFk3AW9Fz:-Kuyfgly2FHN81t4wsy0:b-5bs1jm) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L256)):*

Eventually the Datum interface would disappear.
</blockquote></details>

I think my confusion here is linked to my comment above about the compiler/interpreter divide.  If there's no general interface, doesn't that demand compilation?

---

*[docs/RFCS/20170926_sql_aux_data.md, line 308 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzpOkcLfRbUT_jcXy:-KuyfxKC3qflCc5f5vTI:b-apayz0) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L308)):*

Tree updates by substitution allow sharing the nodes between multiple versions of the tree, see above.
</blockquote></details>

See my comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

knz:

---

*[docs/RFCS/20170926_sql_aux_data.md, line 246 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzCRVrDGcX_qPvz9U:-Kuz3LpUsnUstx09_QtZ:b-y74w1i) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L246)):*

I think I still take issue with considering this to still be an immutable tree.  We can actively make the performance/purity tradeoff, because sometimes it makes sense to sacrifice purity for the sake of performance.  But let's not pretend like we're not sacrificing purity.  And as long as we're at it, let's just go ahead and mutate the tree for other types of changes we're making.  I fail to see how splitting the difference here (unchanging tree structure with changing leaf values) maintains any of the bug-squashing advantages of immutable trees anymore.
</blockquote></details>

Please use more words in your discussion here. What is the problem you are trying to solve?
1) this RFC acknowledge that the current code is heavy in tree mutations (either by substitution, or erroneously in-place sometimes), and overall proposes to do away with that so the trees become properly immutable. Are you suggesting this proposal is not useful?
2) the trade-off between in-place mutation and update-by-copy is extremely well-known and practiced either one way or another in every language/library under the sun. In CockroachDB we made the choice to avoid in-place mutations for a variety of reasons that are still valid but outside of the scope of the discussion. What do you want me to add here?

---

*[docs/RFCS/20170926_sql_aux_data.md, line 256 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzUbWlPUPFk3AW9Fz:-Kuz3znYjo7GsZ4HIuSV:bt3yoxz) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L256)):*

I think my confusion here is linked to my comment above about the compiler/interpreter divide.  If there's no general interface, doesn't that demand compilation?
</blockquote></details>

Not necessarily. You can also  eliminate dynamic dispatch by specializing (parts of) the traversal code to the value types. Also, note that this RFC does not aim to propose the technical mechanisms to achieve those things, and merely states we need to change the data structure upfront to enable further work on the solution(s).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

couchand:

---

*[docs/RFCS/20170926_sql_aux_data.md, line 246 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzCRVrDGcX_qPvz9U:-KuzE9NHLFbrL2wFuvcr:bdoof47) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L246)):*

Please use more words in your discussion here. What is the problem you are trying to solve?
1) this RFC acknowledge that the current code is heavy in tree mutations (either by substitution, or erroneously in-place sometimes), and overall proposes to do away with that so the trees become properly immutable. Are you suggesting this proposal is not useful?
2) the trade-off between in-place mutation and update-by-copy is extremely well-known and practiced either one way or another in every language/library under the sun. In CockroachDB we made the choice to avoid in-place mutations for a variety of reasons that are still valid but outside of the scope of the discussion. What do you want me to add here?
</blockquote></details>

I mean to say, either strategy is fine, but we should pick one and commit to it.  I agree that there are lots of benefits to avoiding mutations.  It seems like, to actually realize those benefits, you have to be strict about preventing mutations anywhere (a point that seems to be echoed by your comment "erroneously in-place sometimes").  I don't think I understand how requiring the structure to be immutable but allowing the leaf values to be mutable is sufficiently strict that it will afford the advantages of treating the trees as immutable.  Specifically, wouldn't a change ("substitution") to a leaf value "pull the rug out" from under the assumption that the tree won't change?  How is it that a substitution doesn't count as a mutation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

knz:

---

*[docs/RFCS/20170926_sql_aux_data.md, line 246 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KuPzCRVrDGcX_qPvz9U:-KuzRjyRFlLVLmlCouXJ:b-jvqbms) ([raw file](https://github.com/cockroachdb/cockroach/blob/5dfe2550ef2f35cab18c9bb2e8dcf0cb87a46476/docs/RFCS/20170926_sql_aux_data.md#L246)):*

I mean to say, either strategy is fine, but we should pick one and commit to it.  I agree that there are lots of benefits to avoiding mutations.  It seems like, to actually realize those benefits, you have to be strict about preventing mutations anywhere (a point that seems to be echoed by your comment "erroneously in-place sometimes").  I don't think I understand how requiring the structure to be immutable but allowing the leaf values to be mutable is sufficiently strict that it will afford the advantages of treating the trees as immutable.  Specifically, wouldn't a change ("substitution") to a leaf value "pull the rug out" from under the assumption that the tree won't change?  How is it that a substitution doesn't count as a mutation?
</blockquote></details>

Oh now I see what you mean.
So the goal of this entire RFC is to propose that the actual *values* of leaf data do not live in the tree directly. In particular, the decision of *what is the particular value* of a leaf item (i.e. what is the result of its "evaluation") should not be a property of the tree any more, but a property of the evaluation context. Once this is true, we can change the value of evaluating a tree (by changing the evaluation of leaf data) without having to modify the tree any more. Does this clarify? Perhaps I should add a sentence to that effect, let me know what works for you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

RaduBerinde:
LGTM. This makes a lot of sense to me. In addition to the performance concerns, the "kind of immutable but not really" tree we have now is error-prone (e.g. we rely that all code plays nice and makes modifications that don't affect a retry of the statement).

---

---

*[docs/RFCS/20170926_sql_aux_data.md, line 233 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KvXJg6E0II4iN5Yhe6I:-KvXJg6E0II4iN5Yhe6J:bk7txgc) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7acc21dbe149ef29c8478346f9ddafbd929a3fe/docs/RFCS/20170926_sql_aux_data.md#L233)):*
> ```Markdown
> inhabits every type. That is, `NULL::int` and `NULL::string` are two
> different things in SQL. Even in a column where NULL *values* can
> occur, all values in the columns are still of the same type.
> ```

I agree with this at a high level, but it's not the case in the current implementation (all values in a column are not of the same *Go* type).

I see the proposal below does describe handling of NULLs, so maybe here we should just say here that NULLs complicate things a bit but the proposal below addresses the issue.

---

*[docs/RFCS/20170926_sql_aux_data.md, line 317 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KvXKtDVmTNQl5WXTeLO:-KvXKtDVmTNQl5WXTeLP:b-piu038) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7acc21dbe149ef29c8478346f9ddafbd929a3fe/docs/RFCS/20170926_sql_aux_data.md#L317)):*
> ```Markdown
> | Current code | New code | Notes |
> |--------------|----------|-------|
> | `type IndexedVar struct {...}` | `type IndexedVar int` | |
> ```

[nit] We may want to rename these as `IndexedVarIdx`, `PlaceholderIdx` etc

---

*[docs/RFCS/20170926_sql_aux_data.md, line 382 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KvXLyr-ZDOiuObYw50S:-KvXLyr-ZDOiuObYw50T:b-1fcakr) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7acc21dbe149ef29c8478346f9ddafbd929a3fe/docs/RFCS/20170926_sql_aux_data.md#L382)):*
> ```Markdown
> measures/decides which one to keep.
> 
> This strongly suggest to implement the link between logical plans not
> ```

[nit] suggests

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-:-KvXMEwPR_pdS0BtjDSJ:b-1gr3fr)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions.

---

*[docs/RFCS/20170926_sql_aux_data.md, line 233 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KvXJg6E0II4iN5Yhe6I:-Kw53GH7c-josFx8KtuO:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7acc21dbe149ef29c8478346f9ddafbd929a3fe/docs/RFCS/20170926_sql_aux_data.md#L233)):*

I agree with this at a high level, but it's not the case in the current implementation (all values in a column are not of the same *Go* type).

I see the proposal below does describe handling of NULLs, so maybe here we should just say here that NULLs complicate things a bit but the proposal below addresses the issue.
</blockquote></details>

Done.

---

*[docs/RFCS/20170926_sql_aux_data.md, line 317 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KvXKtDVmTNQl5WXTeLO:-Kw53Suvl0ur84K9Ec9w:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7acc21dbe149ef29c8478346f9ddafbd929a3fe/docs/RFCS/20170926_sql_aux_data.md#L317)):*

[nit] We may want to rename these as `IndexedVarIdx`, `PlaceholderIdx` etc
</blockquote></details>

Done.

---

*[docs/RFCS/20170926_sql_aux_data.md, line 382 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204#-KvXLyr-ZDOiuObYw50S:-Kw53VgxWnJxi-aeEm3f:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7acc21dbe149ef29c8478346f9ddafbd929a3fe/docs/RFCS/20170926_sql_aux_data.md#L382)):*

[nit] suggests
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18204)*

PR #18212, opened 2017-09-05
build: adjust 3->5 allocator test success criteria

cuongdo:
I recently recreated the snapshot for the "small" nightly allocator
tests. In the process, the # of ranges increased without a
proportionate increase the maximum allowed standard deviation for
replica counts. This fixes that oversight.

### DIFF
--- a/build/teamcity-nightly-acceptance.sh
+++ b/build/teamcity-nightly-acceptance.sh
@@ -37,11 +37,11 @@ case $TESTNAME in
     ;;
   TestRebalance_3To5Small)
     TESTTIMEOUT=4h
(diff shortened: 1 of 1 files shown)


benesch:
Does this supersede #18185?

cuongdo:
@benesch Yes, a max stddev of 20 (as is specified in #18185) isn't sufficient for 841 ranges, since we try to rebalance such that replica counts are within 5% of the mean. `floor(841*.05) = 42`.

benesch (approved):

benesch:
LGTM. On a meta-level, it sounds like stddev is the wrong thing to be checking?

cuongdo:
Yes, though I'm not inclined to fix this immediately, because re-enabling stats-based rebalancing (planned for 1.2) will require significant changes to the allocator test success criteria anyway. Until then, stddev is not ideal but sufficient.

benesch:
Makes sense.

vivekmenezes:
LGTM!

a-robinson (commented):
LGTM

Issue #18219, opened 2017-09-05
storage: node liveness' IncrementEpoch can clobber "newer" state


tschottdorf:
See the following analysis from #17995:

- node 1 gets killed
- node 3 increments node 1's epoch
- everything from now on happens on node 2
- calls IncrementEpoch, which stalls for now
- sets node 1's decommissioning flag
- gossip fires and updates node1's liveness map entry
- IncrementEpoch continues
- it fails (since the decommissioning flag mismatches)
- [blindly updates the liveness][1], effectively clobbering its in-memory state
- calls to Decommission for node1 now turn into no-ops on node2, but
  when the liveness is queried, it looks like node1 isn't decommissioning.

[1]: https://github.com/cockroachdb/cockroach/blob/a9ef8342681baa571ab552d2dc93185ac8eccb49/pkg/storage/node_liveness.go#L566-L572

tschottdorf:
We should add a sequence number to the node liveness field. There are now various hacks in place that "sort of" try to establish an ordering between different liveness versions floating around, but that doesn't work any more now that the `draining` and `decommissioning` flags are present.

It should be possible to migrate this in from 1.1->1.2.

Alternatively, and I'd like that, we can make gossip the only source of liveness updates. I *think* the main reason we do ad-hoc updates of our in-memory state is to short-circuit episodes of clobbering. To mitigate that, we could circuit break. But... sequence numbers may just be more straightforward and you can always use the liveness entries you see, no matter the source, assuming that their sequence number is higher than the previous one.

tschottdorf:
I encountered a problem likely related to this in #18478 today. A dead node had been run `--decommission` against, but the process wouldn't finish and the node remained dead in the ui (not decommissioned).

Investigating one of the live nodes showed that this node had the correct decommissioned liveness for the target node, but yet the decommissioning command wasn't terminating. This suggests that the liveness cache hadn't properly updated itself to account for the new decommissioning status.

Rebooting that node (thus having the node liveness cache reloaded from the system span) fixed the problem, corroborating the above assumption.

tschottdorf:
This happened again, https://forum.cockroachlabs.com/t/decomissioning-dead-1-0-6-nodes-is-hanging/999

Gonna bump this higher on my priority list since it's a big nuisance.

tschottdorf:
This seems to work better since #19422. Closing.

PR #18220, opened 2017-09-05
acceptance: de-flake TestDecommission

### LINKED ISSUE #17995: TestDecommission is flaky

tbg:
The test frequently got stuck during `decommission --wait` for a
freshly killed node, due to the following:

- node 1 gets killed
- node 3 increments node 1's epoch
- everything from now on happens on node 2
- calls IncrementEpoch, which stalls for now
- sets node 1's decommissioning flag
- gossip fires and updates node1's liveness map entry
- IncrementEpoch continues
- it fails (since the decommissioning flag mismatches)
- blindly updates the liveness[1], effectively clobbering its in-memory state
- calls to Decommission for node1 now turn into no-ops on node2, but
  when the liveness is queried, it looks like node1 isn't decommissioning.

Interestingly, when I looked at the liveness output as seen by the other two
nodes, they always had the same problem. Perhaps something else is at work on
top of the above? I filed an issue to investigate & fix the problematic behavior
of IncrementEpoch:

https://github.com/cockroachdb/cockroach/issues/18219

For this PR, I am simply using the semaphore to avoid interleavings between
IncrementEpoch and Decommission, and was able to run the test 10x without it
hanging (usually would hang every other time before, maybe more often).

Fixes #17995.

[1]: https://github.com/cockroachdb/cockroach/blob/a9ef8342681baa571ab552d2dc93185ac8eccb49/pkg/storage/node_liveness.go#L566-L572

### DIFF
--- a/pkg/acceptance/decommission_test.go
+++ b/pkg/acceptance/decommission_test.go
@@ -15,7 +15,7 @@
 package acceptance
 
 import (
-	"fmt"
+	"reflect"
 	"regexp"
 	"strconv"
 	"testing"
@@ -33,6 +33,8 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/httputil"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/cockroachdb/cockroach/pkg/util/retry"
+	"github.com/kr/pretty"
+	"github.com/pkg/errors"
 )
 
 // TestDecommission starts up an >3 node cluster and decomissions and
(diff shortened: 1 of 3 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18220#-:-KtIHEU6GvM4tMZL1Hcn:bnfp4nl)*

tschottdorf:
Unfortunately hit the same flake despite the fix. I can see from gossip that all nodes have received the right info:

```
$ for port in 59743  59751 59761; do ./cockroach debug gossip-values --insecure --port $port | grep -F liveness:1; done
"liveness:1": {NodeID:1 Epoch:1 Expiration:1504645273.001476803,0 Draining:false Decommissioning:true}
"liveness:1": {NodeID:1 Epoch:1 Expiration:1504645273.001476803,0 Draining:false Decommissioning:true}
"liveness:1": {NodeID:1 Epoch:1 Expiration:1504645273.001476803,0 Draining:false Decommissioning:true}
```

but yet, none of the nodes reflect this in `node status --decommission`:

```
$ for port in 59743  59751 59761; do ./cockroach node status 1 --decommission --insecure --port $port; done
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
| id |     address     |      build      |     updated_at      |     started_at      | is_live | gossiped_replicas | is_decommissioning | is_draining |
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
|  1 | 127.0.0.1:59734 | 20aa80063-dirty | 2017-09-05 17:01:03 | 2017-09-05 17:01:03 | false   |                 1 | false              | false       |
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
(1 row)
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
| id |     address     |      build      |     updated_at      |     started_at      | is_live | gossiped_replicas | is_decommissioning | is_draining |
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
|  1 | 127.0.0.1:59734 | 20aa80063-dirty | 2017-09-05 17:01:03 | 2017-09-05 17:01:03 | false   |                 1 | false              | false       |
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
(1 row)
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
| id |     address     |      build      |     updated_at      |     started_at      | is_live | gossiped_replicas | is_decommissioning | is_draining |
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
|  1 | 127.0.0.1:59734 | 20aa80063-dirty | 2017-09-05 17:01:03 | 2017-09-05 17:01:03 | false   |                 1 | false              | false       |
+----+-----------------+-----------------+---------------------+---------------------+---------+-------------------+--------------------+-------------+
(1 row)
```

Looking.

tschottdorf:
Ok, added a commit that hopefully addresses the flakiness. It also explains why I was seeing the odd behavior on all nodes -- we were simply throwing the gossip update away in the common case.

tschottdorf:
OK, addressed one more flake introduced by https://github.com/cockroachdb/cockroach/pull/18178, though I wonder if it's entirely expected that these RPC errors leak to the client.

PTAL!

bdarnell:
:lgtm:

---

---

*[pkg/acceptance/decommission_test.go, line 288 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18220#-KtMmPXfAIuk2YuiB4S8:-KtMmPXfAIuk2YuiB4S9:bhiz3hk) ([raw file](https://github.com/cockroachdb/cockroach/blob/d03759982495f26d7845b912221d7be589292432/pkg/acceptance/decommission_test.go#L288)):*
> ```Go
> 			// send RPCs to the (relatively recently) down node:
> 			//
> 			// pq: rpc error: code = Unavailable desc = grpc: the connection is
> ```

No, we don't want to leak these errors to the client. For idempotent operations, we should retry internally; for non-idempotent operations we should wrap them in AmbiguousResultError. The DistSender does this; is this error coming from somewhere other than the DistSender?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18220#-:-KtMml_RilZ_l8XGsVaS:bnfp4nl)*

tschottdorf:

---

*[pkg/acceptance/decommission_test.go, line 288 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18220#-KtMmPXfAIuk2YuiB4S8:-KtNjkCdb2h8xSiqquGC:bio9l0x) ([raw file](https://github.com/cockroachdb/cockroach/blob/d03759982495f26d7845b912221d7be589292432/pkg/acceptance/decommission_test.go#L288)):*

No, we don't want to leak these errors to the client. For idempotent operations, we should retry internally; for non-idempotent operations we should wrap them in AmbiguousResultError. The DistSender does this; is this error coming from somewhere other than the DistSender?
</blockquote></details>

I think it would come from DistSender, but it's hard to tell. We should annotate these errors better. Filed https://github.com/cockroachdb/cockroach/issues/18264.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18220)*

PR #18228, opened 2017-09-05
storage: Fix off-by-one error in getEstimatedBehindCountRLocked

a-robinson:
The intent of this code was to return an estimated count that would
disable rebalancing, but it was actually returning the maximum value
that wouldn't disable rebalancing, making it so that two replicas had
to be behind before rebalancing would be disabled by
Store.updateReplicationGauges.

This wasn't truly breaking anything, but it was making issues
like #17375 appear less frequently than they otherwise would.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -1082,10 +1082,10 @@ func (r *Replica) getEstimatedBehindCountRLocked(raftStatus *raft.Status) int64 
 		// We haven't heard from the leader, assume we're far behind. This is the
 		// case that is commonly hit when a node restarts. In particular, we hit
 		// this case until an election timeout passes and canCampaignIdleReplica
-		// starts to return true. The results it that a restarted node will
+		// starts to return true. The result is that a restarted node will
 		// consider its replicas far behind initially which will in turn cause it
 		// to reject rebalances.
-		return prohibitRebalancesBehindThreshold
+		return prohibitRebalancesBehindThreshold + 1
 	}
 	if r.mu.estimatedCommitIndex >= r.mu.state.RaftAppliedIndex {
 		return int64(r.mu.estimatedCommitIndex - r.mu.state.RaftAppliedIndex)


tbg (approved):
LGTM, but looks like that code path is untested.

petermattis:
:lgtm:

---

---

*[pkg/storage/replica.go, line 1088 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18228#-KtIdLHpRZAgwSAA_sQ2:-KtIdLHqlS8JRj_W-soO:b-nzcgds) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2e749c84b73d7cad0f1e0313f6ae0ea86e418e0/pkg/storage/replica.go#L1088)):*
> ```Go
> 		// consider its replicas far behind initially which will in turn cause it
> 		// to reject rebalances.
> 		return prohibitRebalancesBehindThreshold + 1
> ```

Or perhaps the code in `updateReplicationGauges` should be:

```
	if selfBehindCount >= prohibitRebalancesBehindThreshold {
```

Not sure why I'm mentioning that. I don't have a strong preference.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18228#-:-KtIdeCfR1gEL7lXaJrx:bnfp4nl)*

a-robinson:

---

*[pkg/storage/replica.go, line 1088 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18228#-KtIdLHpRZAgwSAA_sQ2:-KtIeJPoiQj1lV5Hxa4U:b-4a6j3c) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2e749c84b73d7cad0f1e0313f6ae0ea86e418e0/pkg/storage/replica.go#L1088)):*

Or perhaps the code in `updateReplicationGauges` should be:

```
	if selfBehindCount >= prohibitRebalancesBehindThreshold {
```

Not sure why I'm mentioning that. I don't have a strong preference.
</blockquote></details>

Indeed. I considered that but it required rewording the comment on `prohibitRebalancesBehindThreshold` and seemed less clear overall since a `+1` is more self-documenting (and less likely to be accidentally changed in the future) than using `>=` instead of `>`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18228)*

Issue #18084, opened 2017-08-31
storage: panic in processRaft after handleRaftReady


a-robinson:
https://sentry.io/cockroach-labs/cockroachdb/issues/268859810/

```
*errors.errorString: *errors.fundamental store.go:3411
  File "github.com/cockroachdb/cockroach/pkg/storage/store.go", line 3411, in processReady
  File "github.com/cockroachdb/cockroach/pkg/storage/scheduler.go", line 217, in worker
  File "github.com/cockroachdb/cockroach/pkg/storage/scheduler.go", line 168, in func2
  File "github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go", line 215, in func1

*errors.fundamental store.go:3411
```

a-robinson:
@cuongdo this is one of our most frequent panics in Sentry. Seems worth investigation for 1.1, even if it requires changes past the code freeze. Possibly related to #18082

tschottdorf:
We're panicking while applying a preemptive snapshot. I don't have much else to go on, but #18229 should give us something more actionable as folks upgrade to v1.6.

tschottdorf:
We have something in place for 1.0.6 (#18272, #18229) so hopefully we know soon what the actual crash is about. I'll have to put something similar together for 1.1.

tschottdorf:
Moving to 1.2 now that we'll get better insight into the error on sentry.

tschottdorf:
Marking as resolved, as after https://github.com/cockroachdb/cockroach/pull/20104 we'll see the actual RocksDB messages, and currently (i.e. without that PR) we see lots of errors emanating from RocksDB. Anything else will be tracked under new issues in sentry already.

PR #18229, opened 2017-09-05
storage: storage: expose site of error return in preemptive snapshot path

tbg:
Hopefully this can shed some light on #18084.

----

I intend to add a corresponding patch to the master branch and to cherry-pick it into release-1.1 as well.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -26,6 +26,7 @@ import (
 	"math/rand"
 	"reflect"
 	"sort"
+	"strconv"
 	"sync/atomic"
 	"time"
 	"unsafe"
@@ -48,6 +49,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/storage/engine"
 	"github.com/cockroachdb/cockroach/pkg/storage/engine/enginepb"
 	"github.com/cockroachdb/cockroach/pkg/storage/storagebase"
+	"github.com/cockroachdb/cockroach/pkg/util/caller"
 	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 	"github.com/cockroachdb/cockroach/pkg/util/envutil"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
@@ -2708,7 +2710,10 @@ type handleRaftReadyStats struct {
 // are ready to read, be saved to stable storage, committed or sent to other
 // peers. It takes a non-empty IncomingSnapshot to indicate that it is
 // about to process a snapshot.
-func (r *Replica) handleRaftReady(inSnap IncomingSnapshot) (handleRaftReadyStats, error) {
+//
+// The returned string is nonzero whenever an error is returned to give a
+// non-sensitive cue as to what happened.
+func (r *Replica) handleRaftReady(inSnap IncomingSnapshot) (handleRaftReadyStats, string, error) {
 	r.raftMu.Lock()
 	defer r.raftMu.Unlock()
 	return r.handleRaftReadyRaftMuLocked(inSnap)
@@ -2716,9 +2721,12 @@ func (r *Replica) handleRaftReady(inSnap IncomingSnapshot) (handleRaftReadyStats
 
 // handleRaftReadyLocked is the same as handleRaftReady but requires that the
 // replica's raftMu be held.
+//
+// The returned string is nonzero whenever an error is returned to give a
+// non-sensitive cue as to what happened.
 func (r *Replica) handleRaftReadyRaftMuLocked(
 	inSnap IncomingSnapshot,
-) (handleRaftReadyStats, error) {
+) (handleRaftReadyStats, string, error) {
 	var stats handleRaftReadyStats
 
 	ctx := r.AnnotateCtx(context.TODO())
@@ -2737,11 +2745,12 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 	})
 	r.mu.Unlock()
 	if err != nil {
-		return stats, err
+		const expl = "while checking raft group for Ready"
+		return stats, expl, errors.Wrap(err, expl)
 	}
 
 	if !hasReady {
-		return stats, nil
+		return stats, "", nil
 	}
 
 	logRaftReady(ctx, rd)
@@ -2767,7 +2776,8 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 	if !raft.IsEmptySnap(rd.Snapshot) {
 		snapUUID, err := uuid.FromBytes(rd.Snapshot.Data)
 		if err != nil {
-			return stats, errors.Wrap(err, "invalid snapshot id")
+			const expl = "invalid snapshot id"
+			return stats, expl, errors.Wrap(err, expl)
 		}
 		if inSnap.SnapUUID == (uuid.UUID{}) {
 			log.Fatalf(ctx, "programming error: a snapshot application was attempted outside of the streaming snapshot codepath")
@@ -2777,7 +2787,8 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 		}
 
 		if err := r.applySnapshot(ctx, inSnap, rd.Snapshot, rd.HardState); err != nil {
-			return stats, err
+			const expl = "while applying snapshot"
+			return stats, expl, errors.Wrap(err, expl)
 		}
 
 		if err := func() error {
@@ -2787,16 +2798,15 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 			if r.store.removePlaceholderLocked(r.RangeID) {
 				atomic.AddInt32(&r.store.counts.filledPlaceholders, 1)
 			}
-			if err := r.store.processRangeDescriptorUpdateLocked(ctx, r); err != nil {
-				return errors.Wrap(err, "could not processRangeDescriptorUpdate after applySnapshot")
-			}
-			return nil
+			return r.store.processRangeDescriptorUpdateLocked(ctx, r)
 		}(); err != nil {
-			return stats, err
+			const expl = "could not processRangeDescriptorUpdate after applySnapshot"
+			return stats, expl, errors.Wrap(err, expl)
 		}
 
 		if lastIndex, err = r.stateLoader.loadLastIndex(ctx, r.store.Engine()); err != nil {
-			return stats, err
+			const expl = "while loading last index"
+			return stats, expl, errors.Wrap(err, expl)
 		}
 		// We refresh pending commands after applying a snapshot because this
 		// replica may have been temporarily partitioned from the Raft group and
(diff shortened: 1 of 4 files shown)


a-robinson:
Reviewed 4 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica.go, line 2792 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtIpi9o06bCprf4dfhs:-KtIpi9o06bCprf4dfht:bamni5k) ([raw file](https://github.com/cockroachdb/cockroach/blob/5fb8ef669dccee4d663a4c76a5b8fd3c9ffe6fc8/pkg/storage/replica.go#L2792)):*
> ```Go
> 			}
> 			if err := r.store.processRangeDescriptorUpdateLocked(ctx, r); err != nil {
> 				return util.NotSensitive(errors.Wrap(err, "could not processRangeDescriptorUpdate after applySnapshot"))
> ```

I don't think this is safe. It looks like `processRangeDescriptorUpdateLocked` includes a key in one of its returned errors.

Along those lines, this case was easy, but it's pretty tough to judge whether all these instances are actually safe in general. Like is `r.stateLoader.loadLastIndex` actually safe? It calls down into `engine.MVCCGet`, among other things, and I for one don't know if that code path can include anything sensitive in its errors.

I think it'll be tough to maintain instances like this where we're marking an error returned by one of our other functions as `NotSensitive`. It seems safer to only use `NotSensitive` for what you might call "leaf" errors, where we're either creating a new error or wrapping/returning one from a third-party library that doesn't have access to sensitive data. That would mean that using `errors.Wrap` would get tougher (and probably have to be forked in our repo), but would avoid accidentally leaking something by adding sensitive info to an error that isn't obviously marked as `NotSensitive`.

---

*[pkg/util/error.go, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtIootVwgwY9NmEWsX_:-KtIootVwgwY9NmEWsXa:b-makdhz) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ea390a98d54d94bafa9afc0f07e35067f4ea511/pkg/util/error.go#L31)):*
> ```Go
> // RedactedValue implements log.Redactable.
> func (e safeError) RedactedValue() string {
> 	return e.error.Error()
> ```

Not a big deal at all, but is there a reason not to just use `e.Error()`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229)*

tschottdorf:
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica.go, line 2792 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtIpi9o06bCprf4dfhs:-KtIuDt0JPRodBazCaJE:b-9dioa5) ([raw file](https://github.com/cockroachdb/cockroach/blob/5fb8ef669dccee4d663a4c76a5b8fd3c9ffe6fc8/pkg/storage/replica.go#L2792)):*

I don't think this is safe. It looks like `processRangeDescriptorUpdateLocked` includes a key in one of its returned errors.

Along those lines, this case was easy, but it's pretty tough to judge whether all these instances are actually safe in general. Like is `r.stateLoader.loadLastIndex` actually safe? It calls down into `engine.MVCCGet`, among other things, and I for one don't know if that code path can include anything sensitive in its errors.

I think it'll be tough to maintain instances like this where we're marking an error returned by one of our other functions as `NotSensitive`. It seems safer to only use `NotSensitive` for what you might call "leaf" errors, where we're either creating a new error or wrapping/returning one from a third-party library that doesn't have access to sensitive data. That would mean that using `errors.Wrap` would get tougher (and probably have to be forked in our repo), but would avoid accidentally leaking something by adding sensitive info to an error that isn't obviously marked as `NotSensitive`.


</blockquote></details>

I agree that this venue is risky. In an ideal world (1.2?) we'll do something like you suggested (making leaf errors `NotSensitive` and making sure it plays ball with `errors.Wrap`), but that's too large a refactor for 1.0.6 (or master at the moment). I'll change the `return errors.Wrap(err, str)`s here to `log.Warning(ctx, err); return errors.New(str)` so that we can at least pinpoint where the errors originate from. Or do you have a better idea?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229)*

a-robinson:
Reviewed 1 of 1 files at r2.

---

*[pkg/storage/replica.go, line 2792 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtIpi9o06bCprf4dfhs:-KtJ2mf6rLEKW8YVgSDY:b23a5ve) ([raw file](https://github.com/cockroachdb/cockroach/blob/5fb8ef669dccee4d663a4c76a5b8fd3c9ffe6fc8/pkg/storage/replica.go#L2792)):*

I agree that this venue is risky. In an ideal world (1.2?) we'll do something like you suggested (making leaf errors `NotSensitive` and making sure it plays ball with `errors.Wrap`), but that's too large a refactor for 1.0.6 (or master at the moment). I'll change the `return errors.Wrap(err, str)`s here to `log.Warning(ctx, err); return errors.New(str)` so that we can at least pinpoint where the errors originate from. Or do you have a better idea?
</blockquote></details>

Yeah, that'll be ok. I'd just ask that you either do the appropriate digging on the other methods whose errors we're wrapping here or do the same for them.

---

*[pkg/storage/replica.go, line 2938 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtJ2d0GcnMk80TOo-X8:-KtJ2d0GcnMk80TOo-X9:b5sqwd0) ([raw file](https://github.com/cockroachdb/cockroach/blob/5fb8ef669dccee4d663a4c76a5b8fd3c9ffe6fc8/pkg/storage/replica.go#L2938)):*
> ```Go
> 	// has changed. Or do we need more locking to guarantee that replica
> 	// ID cannot change during handleRaftReady?
> 	return stats, r.withRaftGroup(func(raftGroup *raft.RawNode) (bool, error) {
> ```

I think you also want to wrap the return value of `r.withRaftGroup`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229)*

tschottdorf:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/storage/replica.go, line 2792 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtIpi9o06bCprf4dfhs:-KtM87GA47HSWvrYZ3LM:b6fsleh) ([raw file](https://github.com/cockroachdb/cockroach/blob/5fb8ef669dccee4d663a4c76a5b8fd3c9ffe6fc8/pkg/storage/replica.go#L2792)):*

Yeah, that'll be ok. I'd just ask that you either do the appropriate digging on the other methods whose errors we're wrapping here or do the same for them.
</blockquote></details>

I wasn't going to return any of their errors up the stack, see the current code.

---

*[pkg/storage/replica.go, line 2938 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtJ2d0GcnMk80TOo-X8:-KtMGq9wCxAIY84An4ZX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5fb8ef669dccee4d663a4c76a5b8fd3c9ffe6fc8/pkg/storage/replica.go#L2938)):*

I think you also want to wrap the return value of `r.withRaftGroup`
</blockquote></details>

Done.

---

*[pkg/util/error.go, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtIootVwgwY9NmEWsX_:-KtMGk-jTfawcFeKZ7P3:b-b3jbv4) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ea390a98d54d94bafa9afc0f07e35067f4ea511/pkg/util/error.go#L31)):*

Not a big deal at all, but is there a reason not to just use `e.Error()`?
</blockquote></details>

Obsolete.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229)*

a-robinson:
:lgtm:

---

Reviewed 3 of 5 files at r3, 3 of 3 files at r4.

---

*[pkg/storage/replica.go, line 2960 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtMh7GrVH1NeB69xZvL:-KtMh7GrVH1NeB69xZvM:bhbalm6) ([raw file](https://github.com/cockroachdb/cockroach/blob/5db4412830467c034c935880c94207b8119d126f/pkg/storage/replica.go#L2960)):*
> ```Go
> 		return true, nil
> 	}); err != nil {
> 		return stats, expl, errors.Wrap(err, expl)
> ```

Setting `expl` here would be more consistent, but arguably not worth burning another CI cycle over.

---

*[pkg/storage/store.go, line 3432 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtMhcQ5_2YK6_jeF5AJ:-KtMhcQ5_2YK6_jeF5AK:b-xt85dy) ([raw file](https://github.com/cockroachdb/cockroach/blob/01bccfb79c8c62b9298d2502dd39a0cce863229a/pkg/storage/store.go#L3432)):*
> ```Go
> 		if err != nil {
> 			log.Error(s.AnnotateCtx(context.TODO()), err)
> 			panic(log.Safe{V: expl}) // TODO(bdarnell)
> ```

Any reason not to do something similar with printing the replica state, message type, etc here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-:-KtMiC8hxi2LHMFQj-rg:bnfp4nl)*

tschottdorf:
TFTR!

---

---

*[pkg/storage/replica.go, line 2960 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtMh7GrVH1NeB69xZvL:-KtNtP3aJQD6HJB8ugvL:bcr6y2f) ([raw file](https://github.com/cockroachdb/cockroach/blob/5db4412830467c034c935880c94207b8119d126f/pkg/storage/replica.go#L2960)):*

Setting `expl` here would be more consistent, but arguably not worth burning another CI cycle over.
</blockquote></details>

Oops. Yeah, I'll save the CI.

---

*[pkg/storage/store.go, line 3432 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtMhcQ5_2YK6_jeF5AJ:-KtNsx9mtCqFT4rAd24-:bxbymo5) ([raw file](https://github.com/cockroachdb/cockroach/blob/01bccfb79c8c62b9298d2502dd39a0cce863229a/pkg/storage/store.go#L3432)):*

Any reason not to do something similar with printing the replica state, message type, etc here?
</blockquote></details>

No, just keeping the diff small. The panic we're seeing in sentry is consistently at line 3111.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-:-KtNtSbQ-2z35ciOkOEA:b-814g8k)*

tschottdorf:

---

*[pkg/storage/store.go, line 3432 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229#-KtMhcQ5_2YK6_jeF5AJ:-KtNz6EE6_Al_2FmbiUE:bewthx3) ([raw file](https://github.com/cockroachdb/cockroach/blob/01bccfb79c8c62b9298d2502dd39a0cce863229a/pkg/storage/store.go#L3432)):*

No, just keeping the diff small. The panic we're seeing in sentry is consistently at line 3111.
</blockquote></details>

Oh wait, the other line has an issue too and it's assigned to me as well. Will send an update.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18229)*

tschottdorf:
cc @cockroachdb/release

PR #18235, opened 2017-09-05
sql/pgwire: simplify COPY ... FROM STDIN data flow

### LINKED ISSUE #17941: ERROR:  internal/client/txn.go:836: attempting to use transaction with wrong status or finalized
**BUG REPORT**

1. cockroach version
Build Tag:    v1.1-alpha.20170817
Build Time:   2017/08/17 22:27:51
Distribution: CCL
Platform:     darwin amd64
Go Version:   go1.8.3
C Compiler:   4.2.1 Compatible Clang 3.8.0 (tags/RELEASE_380/final)
Build SHA-1:  9a357a9f553a21ec08235492c7b916219652581a
Build Type:   release

- What did you expect to see?

works correctly on 1.0.4

$ psql "$DB dbname=$DBNAME" < pgbench.ddl
CREATE TABLE
CREATE TABLE
CREATE TABLE
CREATE TABLE
roberts-MBP:pgbench rslee$ cat test.sql |psql "$DB dbname=$DBNAME"
# filling table pgbench_branches (1)
COPY 1
# filling table pgbench_tellers (10)
COPY 10
# filling table pgbench_accounts (100000)
COPY 100000
roberts-MBP:pgbench rslee$ cockroach version
Build Tag:    v1.0.4
Build Time:   2017/07/27 17:52:31
Distribution: CCL
Platform:     darwin amd64
Go Version:   go1.8.3
C Compiler:   4.2.1 Compatible Clang 3.8.0 (tags/RELEASE_380/final)
Build SHA-1:  5b757262d33d814bda1deb2af20161a1f7749df3
Build Type:   release

- What did you see instead?

psql "$DB" -c "create database $DBNAME"
CREATE DATABASE
roberts-MBP:pgbench rslee$ psql "$DB dbname=$DBNAME" < pgbench.ddl
CREATE TABLE
CREATE TABLE
CREATE TABLE
CREATE TABLE
roberts-MBP:pgbench rslee$ cat test.sql |psql "$DB dbname=$DBNAME"
# filling table pgbench_branches (1)
COPY 1
# filling table pgbench_tellers (10)
COPY 10
# filling table pgbench_accounts (100000)
ERROR:  internal/client/txn.go:836: attempting to use transaction with wrong status or finalized
[pgbench.ddl.zip](https://github.com/cockroachdb/cockroach/files/1254060/pgbench.ddl.zip)
[test.sql.zip](https://github.com/cockroachdb/cockroach/files/1254061/test.sql.zip)

: COMMITTED true

### LINKED ISSUE #18133: ERROR:  internal/client/txn.go:836: attempting to use transaction with wrong status or finalized: COMMITTED true
Please use TBL and schema from https://github.com/cockroachdb/cockroach/issues/18076

**BUG REPORT**

$ cockroach version
Build Tag:    391d19f
Build Time:   2017/08/31 16:02:30
Distribution: CCL
Platform:     darwin amd64
Go Version:   go1.8.3
C Compiler:   4.2.1 Compatible Clang 3.8.0 (tags/RELEASE_380/final)
Build SHA-1:  391d19f65fc1583fba632c203c82985442322545
Build Type:   release

[logs.zip](https://github.com/cockroachdb/cockroach/files/1270422/logs.zip)

- What did you do?

cat > copy.lineitem.sql <<EOF
copy lineitem from stdin;
EOF
cat copy.lineitem.sql lineitem.tbl | psql "port=26257 user=root host=127.1 dbname=tpch"

- What did you expect to see?

Data loaded.

- What did you see instead?

$ cat copy.lineitem.sql lineitem.tbl | psql "port=26257 user=root host=127.1 dbname=tpch"
ERROR:  internal/client/txn.go:836: attempting to use transaction with wrong status or finalized: COMMITTED true

Committed 100 rows instead of not committing.

$ wc -l lineitem.tbl
 6001215 lineitem.tbl

> select count(*) from lineitem;
+----------+
| count(*) |
+----------+
|      100 |
+----------+

### LINKED ISSUE #18175: sql: pgwire copy is broken by results streaming
Here's an excellent one: when doing pgwire copy operations, we call `Executor.execRequest()` recursively. This is very bad; the `Executor` is not re-entrant and all sorts of session state management goes wrong.
Besides, I think now every packet of data to copy means one recursive call, so there's probably only so much data we handle until the stack overflows.

Here's the flow: during a copy, sending a result leads to this `CopyIn()` call which in turn calls back into the `Executor`:
https://github.com/cockroachdb/cockroach/blob/ef021b3e921aa1130b8fb4c949e115ee09b49012/pkg/sql/pgwire/v3.go#L1253

I believe this is new with result streaming. Before, sending results was done out of the `Executor`. Now, it's done while a query is running.

One way to see bad things happen is by running `TestCopyError` and observing that this assertion that was put in for #7881 fires:
https://github.com/cockroachdb/cockroach/blob/ef021b3e921aa1130b8fb4c949e115ee09b49012/pkg/sql/executor.go#L850

I think this is a biggie. Besides copy being broken, this is also preventing me from doing some other fixes I need to do to the `Executor`, in the sense that things currently don't crash because we've papered over the assertion listed above and we do something random and broken to recover from it, but I have to change this state of affairs.

I don't know anything about this copy protocol, but I'm happy to consult about finding a new structure. @mjibson are you a good person for this one?

jordanlewis:
COPY FROM STDIN in a post-streaming-results world had an inadvertent
re-entrancy into the main executor flow. This was confusing and possibly
incorrect. The way it worked before was that running `CloseResult` on
the `COPY FROM` statement itself would initialize and run the entire
COPY state machine, which needs to re-enter the executor leading to
that possible incorrectness.

This refactor removes the recursive flow. Now, `CloseResult` on `COPY
FROM` merely sends the magic "ready for copy in" message to the client.
Then, once we return from the executor for `COPY FROM`, the pgwire
implementation switches from its ordinary state machine to the COPY
state machine and runs until completion, at which point it returns
control to the ordinary state machine.

Also, a small refactor to store a `Session` inside of the `copyNode` instead of a `planner`, which was not necessary.

Fixes #18175.
Fixes #17941.
Fixes #18133.

### DIFF
--- a/pkg/acceptance/psql_test.go
+++ b/pkg/acceptance/psql_test.go
@@ -38,6 +38,7 @@ psql -c "select 1" | grep "1 row"
 # Check that COPY works outside of a transaction (#13395)
 psql -d testdb <<EOF
 CREATE DATABASE IF NOT EXISTS testdb;
+CREATE TABLE ints (a INTEGER NOT NULL);
 CREATE TABLE playground (
     equip_id integer NOT NULL,
     type character varying(50) NOT NULL,
@@ -82,6 +83,16 @@ echo "SELECT 'hooray'" >> import.sql
 psql -d testdb < import.sql | grep hooray
 # Assert the junk line wasn't added.
 psql -d testdb -c "SELECT * from playground WHERE type='junk'" | grep "0 rows"
+
+# Test that large COPY FROM STDIN commands don't create a bad connection status.
+# See issue #17941.
+echo 'COPY ints FROM stdin;' > import.sql
+for i in {1..1000}; do
+    echo $i >> import.sql
+done
+echo "\." >> import.sql
+psql -d testdb < import.sql
+psql -d testdb -c "SELECT COUNT(*) FROM ints" | grep "1000"
 
 exit 0
 `
--- a/pkg/sql/copy.go
+++ b/pkg/sql/copy.go
@@ -41,7 +41,7 @@ import (
 //
 // See: https://www.postgresql.org/docs/9.5/static/sql-copy.html
 type copyNode struct {
-	p             *planner
+	session       *Session
 	table         parser.TableExpr
 	columns       parser.UnresolvedNames
 	resultColumns sqlbase.ResultColumns
@@ -54,7 +54,7 @@ func (*copyNode) Values() parser.Datums        { return nil }
 func (*copyNode) Next(runParams) (bool, error) { return false, nil }
 
 func (n *copyNode) Close(ctx context.Context) {
-	n.rowsMemAcc.Wsession(n.p.session).Close(ctx)
+	n.rowsMemAcc.Wsession(n.session).Close(ctx)
 }
 
 // CopyFrom begins a COPY.
@@ -81,7 +81,7 @@ func (p *planner) CopyFrom(ctx context.Context, n *parser.CopyFrom) (planNode, e
 	for i, c := range cols {
 		cn.resultColumns[i] = sqlbase.ResultColumn{Typ: c.Type.ToDatumType()}
 	}
-	cn.p = p
+	cn.session = p.session
 	cn.rowsMemAcc = p.session.OpenAccount()
 	return cn, nil
 }
@@ -90,10 +90,10 @@ func (p *planner) CopyFrom(ctx context.Context, n *parser.CopyFrom) (planNode, e
 func (n *copyNode) Start(runParams) error {
 	// Should never happen because the executor prevents non-COPY messages during
 	// a COPY.
-	if n.p.session.copyFrom != nil {
+	if n.session.copyFrom != nil {
 		return fmt.Errorf("COPY already in progress")
 	}
-	n.p.session.copyFrom = n
+	n.session.copyFrom = n
 	return nil
 }
 
@@ -174,7 +174,7 @@ func (n *copyNode) addRow(ctx context.Context, line []byte) error {
 		return fmt.Errorf("expected %d values, got %d", len(n.resultColumns), len(parts))
 	}
 	exprs := make(parser.Exprs, len(parts))
-	acc := n.rowsMemAcc.Wsession(n.p.session)
+	acc := n.rowsMemAcc.Wsession(n.session)
 	for i, part := range parts {
 		s := string(part)
 		if s == nullString {
@@ -194,7 +194,7 @@ func (n *copyNode) addRow(ctx context.Context, line []byte) error {
 				return err
 			}
 		}
-		d, err := parser.ParseStringAs(n.resultColumns[i].Typ, s, n.p.session.Location)
+		d, err := parser.ParseStringAs(n.resultColumns[i].Typ, s, n.session.Location)
 		if err != nil {
 			return err
 		}
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -688,6 +688,7 @@ func (e *Executor) CopyDone(session *Session) error {
 
 // CopyEnd ends the COPY mode. Any buffered data is discarded.
 func (s *Session) CopyEnd(ctx context.Context) {
+	s.copyFrom.Close(ctx)
 	s.copyFrom = nil
 }
 
--- a/pkg/sql/pgwire/v3.go
+++ b/pkg/sql/pgwire/v3.go
@@ -544,6 +544,31 @@ func (c *v3Conn) serve(ctx context.Context, draining func() bool, reserved mon.B
 		if err != nil {
 			return err
 		}
+
+		if c.streamingState.copyIn {
+			tag := append(c.tagBuf[:0], c.streamingState.pgTag...)
+			tag = append(tag, ' ')
+
+			// Run the copy pgwire state machine. This runs until its finished or
+			// there was an error, returning the number of rows inserted.
+			rowsInserted, err := c.copyIn(ctx, c.streamingState.columns)
+			c.streamingState.copyIn = false
+			if err != nil {
+				if err := c.setError(err); err != nil {
+					return err
+				}
+			}
+
+			// Send the number of rows inserted, along with the COPY tag. Then we're
+			// ready to return to normal operation.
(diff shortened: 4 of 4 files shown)


andreimatei:
I'll look carefully tomorrow. I'm really excited that this change is small :thumbsup: 

You should say "Fixes #18175" in the commit.
Also probably fixes #18133 and #17941. Would you mind checking if that's the case?

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-:-KtK7d7NVS2sM6AJek7u:b-jjndru)*

jordanlewis:
Yeah, I wasn't able to reproduce either of those two issues from @robert-s-lee after this patch. Updated the PR message.

madelynnblue (approved):

andreimatei:
LGTM, but can we add a test based on those bugs?

nit: update commit message too

---

---

*[pkg/sql/copy.go, line 197 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-KtMigWA3L1aFr6Sa16-:-KtMigWA3L1aFr6Sa160:b-mlyast) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97604fee82ae2a39f059ed1a76d56cb27377d34/pkg/sql/copy.go#L197)):*
> ```Go
> 			}
> 		}
> 		d, err := parser.ParseStringAs(n.resultColumns[i].Typ, s, n.session.Location)
> ```

nit: is there an `EvalCtx` around that you can use to get the location from? If so, I think that's be nicer.

---

*[pkg/sql/pgwire/v3.go, line 1107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-KtMo6gP3KCGShPjXZAh:-KtMo6gP3KCGShPjXZAi:b-awkbrr) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97604fee82ae2a39f059ed1a76d56cb27377d34/pkg/sql/pgwire/v3.go#L1107)):*
> ```Go
> // copyIn processes COPY IN data and returns the number of rows inserted.
> // See: https://www.postgresql.org/docs/current/static/protocol-flow.html#PROTOCOL-COPY
> func (c *v3Conn) copyIn(ctx context.Context, columns []sqlbase.ResultColumn) (int64, error) {
> ```

please document what this returns

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-:-KtMhcqaj8OQMH_t6DWz:b-ox6x77)*

jordanlewis:
Added a test that fails before this patch. Turns out all you need to do is send a COPY large enough that it takes more than 1 insert statement to run.

Personally I prefer to leave the Fixes declarations in the PR description only, since they're a GitHub thing. Do we have a standard here?

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/copy.go, line 197 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-KtMigWA3L1aFr6Sa16-:-KtMsM_0fzWROMpxikmk:buk1ljl) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97604fee82ae2a39f059ed1a76d56cb27377d34/pkg/sql/copy.go#L197)):*

nit: is there an `EvalCtx` around that you can use to get the location from? If so, I think that's be nicer.
</blockquote></details>

Not really.

---

*[pkg/sql/pgwire/v3.go, line 1107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-KtMo6gP3KCGShPjXZAh:-KtMsWDAb4H_dTmHw2f3:b-sxl9uy) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97604fee82ae2a39f059ed1a76d56cb27377d34/pkg/sql/pgwire/v3.go#L1107)):*

please document what this returns
</blockquote></details>

It already is documented (returns the number of rows inserted) - did you mean something more?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-:-KtNFbjUq687J3y3R1GC:bhkcdeq)*

andreimatei:
> Personally I prefer to leave the Fixes declarations in the PR description only, since they're a GitHub thing. Do we have a standard here?

By this rationale, we shouldn't have issue references in our code, but we do, of course. Also, what if you have multiple commits in a PR?
The standard that I've heard and have been following says that for single-commit PRs, the PR message and the commit message should be the same.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-:-KtNxMm7pK46OVypK4RC:be14mw3)*

jordanlewis:
@andreimatei that's reasonable. Another reason I prefer to keep the issue number out of commit messages is that otherwise the issue gets modified once per force-push, which just seems unnecessarily shaming to those of us who push a few times to get CI to pass 😊

knz:
Many conciseness, Such elegance. Wow. Amaze. :dog:
:lgtm: but there's a little case of bad grammar to fix now!!! :sweat_smile: 

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2.

---

*[pkg/sql/pgwire/v3.go, line 552 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-KtRJwZOfohTSrt97z3h:-KtRJwZP8lTv5qrjiDHk:bb7ibks) ([raw file](https://github.com/cockroachdb/cockroach/blob/c814889106bf7d35cd57e6016a684c540fe9dcc7/pkg/sql/pgwire/v3.go#L552)):*
> ```Go
> 			tag = append(tag, ' ')
> 
> 			// Run the copy pgwire state machine. This runs until its finished or
> ```

it's

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18235#-:-KtRK2wHRDWwyDJmhJMM:b-q04mnk)*

Issue #17432, opened 2017-08-03
DistSQL runs on one node when running count(*) on 150k row table


nstewart:
NODE 1
I170803 20:51:31.732120 1 util/log/clog.go:975  [config] file created at: 2017/08/03 20:51:31
I170803 20:51:31.732120 1 util/log/clog.go:975  [config] running on machine: ubuntu-512mb-sfo1-01
I170803 20:51:31.732120 1 util/log/clog.go:975  [config] binary: CockroachDB CCL 8e4b34a (linux amd64, built 2017/08/03 16:48:03, go1.8.3)
I170803 20:51:31.732120 1 util/log/clog.go:975  [config] arguments: [/usr/local/bin/cockroach start --insecure --advertise-host=10.134.8.165 --join=10.134.9.162]

NODE 2
I170803 20:53:50.138391 1 util/log/clog.go:975  [config] file created at: 2017/08/03 20:53:50
I170803 20:53:50.138391 1 util/log/clog.go:975  [config] running on machine: ubuntu-512mb-sfo1-02
I170803 20:53:50.138391 1 util/log/clog.go:975  [config] binary: CockroachDB CCL 8e4b34a (linux amd64, built 2017/08/03 16:48:03, go1.8.3)
I170803 20:53:50.138391 1 util/log/clog.go:975  [config] arguments: [cockroach start --insecure --join=10.134.8.165 --advertise-host=10.134.9.162]

NODE 3
I170803 20:55:35.508466 1 util/log/clog.go:975  [config] file created at: 2017/08/03 20:55:35
I170803 20:55:35.508466 1 util/log/clog.go:975  [config] running on machine: ubuntu-512mb-sfo1-03
I170803 20:55:35.508466 1 util/log/clog.go:975  [config] binary: CockroachDB CCL 8e4b34a (linux amd64, built 2017/08/03 16:48:03, go1.8.3)
I170803 20:55:35.508466 1 util/log/clog.go:975  [config] arguments: [/usr/local/bin/cockroach start --insecure --join=10.134.8.165 --advertise-host=10.134.4.235]

Please describe the issue you observed:

- What did you do? 
loaded tpch data into a 3 node cluster. 
EXPLAIN(DISTSQL) SELECT count(*) FROM customer; (150k rows)
EXPLAIN(DISTSQL) SELECT count(*) FROM lineitems; (6M rows)

- What did you expect to see? 
I expected to see a query plan that spanned multiple nodes in both cases.

[...]

Issue #16317, opened 2017-06-05
backup/restore: extra empty range created on restore


danhhz:
Backup selects the spans to export by collecting all indexes from targeted tables. When these are restored, we presplit ranges along the backup boundaries, which results in an extra empty range between /Table/51 and /Table/51/1 (where 51 is the table id and 1 is the primary index id). This happens for each non-interleaved (or interleave-root) table in the restore. Range quiescence mitigates this a bit, but it's still unfortunate and we should avoid this permanently empty range

danhhz:
This is not going to make 1.1

a6802739:
@danhhz  could I have a see on it?

a6802739:
@danhhz  if you could give me more reference, maybe I could do this.

danhhz:
@a6802739 thanks for the offer, but this will have to wait until we open master to 1.2 development (which happens in early september)

mjibson:
I'm not able to reproduce this:

```
root@:26257/a> alter table d.t split at values (2);
+--------------+---------------+
|     key      |    pretty     |
+--------------+---------------+
| "\u0109\x8a" | /Table/60/1/2 |
+--------------+---------------+
(1 row)

Time: 33.326645ms

root@:26257/a> select * from d.t;
+---+
| i |
+---+
| 1 |
| 3 |
+---+
(2 rows)

Time: 2.153415ms

root@:26257/a> show testing_ranges from table d.t;
+-----------+---------+----------+----------+--------------+
| Start Key | End Key | Range ID | Replicas | Lease Holder |
+-----------+---------+----------+----------+--------------+
| NULL      | /2      |       43 | {1}      |            1 |
| /2        | NULL    |       47 | {1}      |            1 |
+-----------+---------+----------+----------+--------------+
(2 rows)

Time: 11.029681ms

root@:26257/a> BACKUP d.* TO 'nodelocal:///home/mjibson/backup';
+--------------------+-----------+--------------------+------+---------------+----------------+-------+
|       job_id       |  status   | fraction_completed | rows | index_entries | system_records | bytes |
+--------------------+-----------+--------------------+------+---------------+----------------+-------+
| 285210659025780737 | succeeded |                  1 |    2 |             0 |             25 |  7224 |
+--------------------+-----------+--------------------+------+---------------+----------------+-------+
(1 row)

Time: 111.203283ms

[...]

PR #18236, opened 2017-09-05
sql: minor fix to `SHOW TESTING_RANGES`

RaduBerinde:
When showing ranges for a table, if we happen to have a split exactly at the
index start key `/tableID/priIdxID` we are showing an extra range. The issue is
that we are looking up the meta range for the span start key; the fix is to look
for the next key so we skip any range that ends exactly at this key.

Before:
```
root@:26257/tpch> show testing_ranges from table customer;
+-----------+---------+----------+--------------+
| Start Key | End Key | Replicas | Lease Holder |
+-----------+---------+----------+--------------+
| NULL      |         | {1,2,3}  |            3 |
|           | NULL    | {1,2,3}  |            3 |
+-----------+---------+----------+--------------+
```

After:
```
root@:26257/tpch> SHOW TESTING_RANGES FROM TABLE customer;
+-----------+---------+----------+--------------+
| Start Key | End Key | Replicas | Lease Holder |
+-----------+---------+----------+--------------+
| NULL      | NULL    | {1,2,3}  |            1 |
+-----------+---------+----------+--------------+
```

Relevant to the second issue described in #17432.

### DIFF
--- a/pkg/sql/show_ranges.go
+++ b/pkg/sql/show_ranges.go
@@ -154,7 +154,7 @@ func (n *showRangesNode) Close(_ context.Context) {
 func scanMetaKVs(
 	ctx context.Context, txn *client.Txn, span roachpb.Span,
 ) ([]client.KeyValue, error) {
-	metaStart := keys.RangeMetaKey(keys.MustAddr(span.Key))
+	metaStart := keys.RangeMetaKey(keys.MustAddr(span.Key).Next())
 	metaEnd := keys.RangeMetaKey(keys.MustAddr(span.EndKey))
 
 	kvs, err := txn.Scan(ctx, metaStart, metaEnd, 0)


RaduBerinde:
Note that #16317 is relevant here - when we restore a table, we have an empty range that ends in the primary index start key.

knz:
I'm confused; if a request is made to the start key, which range serves the query? 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18236#-:-KtMa2KLbnP47DomCn5U:be01nf5)*

RaduBerinde:
The query is not to the range itself, its to the `meta` range corresponding to that key.

RaduBerinde:
To be more specific - every range `[StartKey, EndKey)` has a corresponding `Meta2Prefix/EndKey` key with the range descriptor as value.

andreimatei:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18236#-:-KtMgYzMLA3eM8G5p9Ck:bnfp4nl)*

RaduBerinde:
@petermattis thoughts on cross-porting this to 1.1? It's cosmetic but the fix is trivial (and we're using this command internally to understand distsql plans, e.g. #17432)

petermattis:
@RaduBerinde I'm fine with cherry-picking this to 1.1.

Issue #18240, opened 2017-09-06
ci: nightly stress runs are noops


tschottdorf:
https://teamcity.cockroachdb.com/viewLog.html?buildId=344333&buildTypeId=Cockroach_Nightlies_Stress&tab=buildLog

We should've known something is up since there haven't been any nightly failures in a while. Example:

https://teamcity.cockroachdb.com/repository/download/Cockroach_Nightlies_Stress/344285:id/stress.log

I think it's enough to bump the `stress` dependency which I'm trying to do right now.

jordanlewis:
Did this fix the issue?

tschottdorf:
Definitely in local testing, but we'll know tomorrow - I'm very sure we get failures every night.

tschottdorf:
(re: how I found out -- I was running `make stress PKG=./pkg/server TESTS=MyFlakyTest` and it printed `SUCCESS` && exit 0 right after a fat exploding stacktrace).

tschottdorf:
judging by the 28 new stress failures opened tonight, there's little doubt that it now works again.

PR #18241, opened 2017-09-06
vendor: update stress

tbg:
The previous version reported success on failures.

See https://github.com/cockroachdb/cockroach/issues/18240.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -52,3 +52,7 @@ ignored = [
 [[constraint]]
   name = "github.com/coreos/etcd"
   branch = "master"
+
+[[constraint]]
+  name = "github.com/cockroachdb/stress"
+  branch = "master"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -656,7 +656,7 @@
 
 [[projects]]
   name = "google.golang.org/grpc"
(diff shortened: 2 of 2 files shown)


tschottdorf:
I tried to avoid a full-on dep bump by running `dep ensure -v github.com/cockroachdb/stress@29b5d31b4c3a949cb3a726750bc34c4d58ec15e8` but `dep ensure` (which we verify in CI) doesn't want to move `stress` to `master`, so I added a constraint.

@dt mind checking if the diff looks ok?

Gopkg.toml
```diff
@@ -52,3 +52,7 @@ ignored = [
 [[constraint]]
   name = "github.com/coreos/etcd"
   branch = "master"
+
+[[constraint]]
+  name = "github.com/cockroachdb/stress"
+  branch = "master"
```
dt: you could also just tag a version in `stress` since dep default is latest tag

dt (approved):
Yep, that's expected `dep` behavior -- default is latest tagged release, absent some constraint (like `master`). We add `master` constraints like this where we want to run ahead of an upstream's release cycle (or when they don't tag releases), or we prod them to tag a release including whatever commit we're interested. The `master` constraint certainly works here, or, since we have some amount of influence on upstream in this case, we could also try to influence their release cycle :p

tschottdorf:
Thanks @dt! I'll keep this pinned to master because this particular package is very unlikely to be updated when we don't want it to be pulled in.

PR #18243, opened 2017-09-06
distsql: enable sending arrays over distsql

justinj:
🔬🐶. Relevant info for someone who knows more about distSQL is that
arrays have a value encoding but no key encoding.

### DIFF
--- a/pkg/sql/distsql_physical_planner.go
+++ b/pkg/sql/distsql_physical_planner.go
@@ -251,8 +251,7 @@ func (dsp *distSQLPlanner) checkSupportForNode(node planNode) (distRecommendatio
 
 	case *renderNode:
 		for i, e := range n.render {
-			if typ := n.columns[i].Typ; typ.FamilyEqual(parser.TypeTuple) ||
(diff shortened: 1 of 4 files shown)


RaduBerinde:
This check needs to be enhanced to always force a `VALUE` encoding for arrays: https://github.com/cockroachdb/cockroach/blob/8ac54961ab8edc641e41880f8e410032dc1843eb/pkg/sql/distsqlrun/stream_encoder.go#L107

If tests are passing, then there is no test where the `preferredEconding` is used. I think something where we render an array instead of reading it would do it, something like `SELECT ARRAY[k] FROM kv` where `kv` has a bunch of keys.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18243#-:-KtM_Rk9GsWL-8o-87Yg:b-1zcrnx)*

justinj:
I see! OK, PTAL, the test I added did indeed fail like you suspected.

RaduBerinde:
LGTM

justinj:
TFTR!

PR #18252, opened 2017-09-06
storage: improve error log about passing no candidates to RemoveTarget

a-robinson:
This wasn't actually causing problems because allocator.RemoveTarget
does a good job of verifying its inputs, but it's better form to handle
the error outside of RemoveTarget and to provide more information in the
error message than one RemoveTarget could -- its error message is
"must supply at least one candidate replica to allocator.RemoveTarget()"


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18252#-:-KtN-tjSefetfOyGmCZu:bnfp4nl)*

PR #15903, opened 2017-05-13
sql/parser: permit untyped string constant coercion to numeric types

benesch:
It seems that @nvanbenschoten would be the obvious choice as reviewer, but @knz, since he's on vacation, could you triage and/or review?

@cuongdo, now that I've drawn this up, it seems to me that it's a better solution than cockroachdb/sequelize-cockroachdb#2.

---

Allow untyped string literals to coerce to `INT`, `DECIMAL`, and `FLOAT`. (For
prior art, see #13685, which allowed string literals to coerce into
booleans.)

Essentially, anywhere an `INT`, `DECIMAL`, or `FLOAT` is required, a string
constant encoding the appropriate type can be used instead. The
following query now works, for example:

    CREATE TABLE t (a INT DEFAULT '1');

As does:

    SELECT * FROM foo LIMIT '1';

This brings us closer in line with Postgres's semantics. Importantly,
it solves a problem with Sequelize (see cockroachdb/sequelize-cockroachdb#2),
where Sequelize expects to be able to store a string literal
representing an integer (instead of an integer literal) into an INT
column.

PR #18258, opened 2017-09-06
distsql: fix serialization of negative ints

### LINKED ISSUE #15617: distsql: roundtrip to string does not respect precedence of ::: and -
Run
```
CREATE TABLE t (c int PRIMARY KEY);
INSERT INTO t VALUES (1), (2), (3);
SELECT (-9223372036854775808) ::: INT;
SELECT c FROM t WHERE (c, c) < (1, -9223372036854775808);
```
and get
```
Error: pq: ($0, $0) < (1, - 9223372036854775808:::INT): tuples ($0, $0), (1, - 9223372036854775808:::INT) are not comparable at index 2: numeric constant out of int64 range
```
Turn off distsql and the problem goes away.

RaduBerinde:
Correcting the serialization of negative integers from `-5:::INT` to
`'-5':::INT`.

Fixes #15617.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/distsql_expr
+++ b/pkg/sql/logictest/testdata/logic_test/distsql_expr
@@ -1,0 +1,14 @@
+# LogicTest: default distsql
+
+statement ok
+CREATE TABLE t (c int PRIMARY KEY);
+
+statement ok
+INSERT INTO t VALUES (1), (2), (3);
+
+# Verify we serialize the negative constant correctly (see #15617).
+query I rowsort
+SELECT c FROM t WHERE (c, c) > (2, -9223372036854775808);
+----
+2
+3
--- a/pkg/sql/parser/datum.go
+++ b/pkg/sql/parser/datum.go
@@ -422,7 +422,16 @@ func (*DInt) AmbiguousFormat() bool { return true }
 
 // Format implements the NodeFormatter interface.
 func (d *DInt) Format(buf *bytes.Buffer, f FmtFlags) {
+	// If the number is negative, we need to use parens or the `:::INT` type hint
+	// will take precedence over the negation sign.
+	quote := f.disambiguateDatumTypes && *d < 0
+	if quote {
+		buf.WriteByte('(')
+	}
 	buf.WriteString(strconv.FormatInt(int64(*d), 10))
+	if quote {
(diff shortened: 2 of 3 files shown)


benesch:
Oof, that's a gotcha. This seems to rely on https://github.com/cockroachdb/cockroach/pull/15903, though, which @knz only grudgingly LGTM'd. Is 

```
(-9223372036854775808):::int;
```

perhaps purer? 

Might also be worth adding a more targeted test case in sql/parser/format_test.go.

Anyway, don't bother with either until @knz weighs in.

RaduBerinde:
Other datums like DDecimal also use quotes. I don't think #15903 is relevant, we always have a `:::INT` specifier (which is close to a cast). I think that with parens we would get an extra `ParenExpr` node when we parse.

benesch:
I believe `DDecimal` and `DFloat` only use quotes for `NaN` and `Inf`, which aren't SQL identifiers and so have to be string literals. I see what you mean about `:::int` forcing the conversion with or without #15903, though. Consider that concern striken.

knz:
DFLoat and DDecimal need quotes because they may be represented by "Inf" or "NaN". 
DInt has no such limitation, so the parentheses are adequate. Also there *is* overhead in handling this with quotes.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18258#-:-KtRFWgRX0omt2yYjgU-:bq9cgy9)*

knz:
Nice find though!!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18258#-:-KtRFligYPNlUtkwy53Y:bnyjzes)*

knz:
This must be cherry-picked to 1.1 when it mergs.

RaduBerinde:
Understood, switched to parens.

knz (approved):

RaduBerinde:
Also added a format_test as suggested. Any thoughts on negative decimals? They work without parens (even constructions like `x - -largeconstant`) so it's not strictly necessary.

knz:
Yes I am not concerned about decimals (and floats).

RaduBerinde:
Thanks for the reviews!

PR #18260, opened 2017-09-06
ui: Adjust Threshold for Suspect Nodes

### LINKED ISSUE #17840: ui: nodes sometimes appear as suspect when the cluster is healthy
This seems to be a known bug that we should find a way to improve, since it confuses users who see suspect nodes, expect something to be wrong, but nothing is wrong with the cluster. Stability of the cluster is an important value prop for us, so we should make sure not to confuse our users.

This issue has come up in the following issues: https://github.com/cockroachdb/cockroach/issues/17775, https://github.com/cockroachdb/cockroach/issues/17132

mrtracy:
The criteria for marking a node as "suspect" in the UI may have been too
strict, resulting in occasions where working nodes are being marked as
suspect. This commit is assuming that the issue arises from a
combination of latency and clock skew.

+ Previously, the UI marked nodes as "suspect" if their liveness
expiration time had elapsed as of the timestamp when liveness records
were recieved. This commit now compares liveness expirations to the
timestamp when liveness records were *requested*, which should eliminate
latency from the round trip call as a factor.
+ Additionally, a five second grace period has been added to the suspect
computation.

We expect that the combination of these two measures should relieve the
issue.

Resolves #17840

### DIFF
--- a/pkg/ui/src/redux/cachedDataReducer.spec.ts
+++ b/pkg/ui/src/redux/cachedDataReducer.spec.ts
@@ -58,6 +58,7 @@ describe("basic cachedDataReducer", function () {
         state = reducer(state, testReducerObj.requestData());
         expected = new CachedDataReducerState<Response>();
         expected.inFlight = true;
+        expected.requestedAt = testMoment;
         assert.deepEqual(state, expected);
       });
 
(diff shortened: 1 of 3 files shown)


BramGruneir:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18260#-:-KtN_angEHqxcenpciQ9:bnfp4nl)*

PR #18261, opened 2017-09-06
sql: added short-circuiting when evaluating ANY/ALL operations

richardwu:
Instead of iterating through the entire set of elements we can short-circuit when we know already that:

1. ANY: there exists an element that evaluates to true, so we can immediately return true
2. ALL: there exists an element that evaluates to false, so we can immediately return false

This was identified earlier in #18094

### DIFF
--- a/pkg/sql/parser/eval.go
+++ b/pkg/sql/parser/eval.go
@@ -1723,7 +1723,12 @@ func makeEvalTupleIn(typ Type) CmpOp {
 }
 
 // evalDatumsCmp evaluates Datums (slice of Datum) using the provided
-// sub-operator type and its CmpOp with the left Datum.
+// sub-operator type (ANY/SOME, ALL) and its CmpOp with the left Datum.
+// It returns the result of the ANY/SOME/ALL predicate.
+//
+// A NULL result is returned if there exists a NULL element and:
+//   ANY/SOME: no comparisons evaluate to true
+//   ALL: no comparisons evaluate to false
 //
 // For example, given 1 < ANY (SELECT * FROM GENERATE_SERIES(1,3))
 // (right is a DTuple), evalTupleCmp would be called with:
@@ -1735,8 +1740,7 @@ func evalDatumsCmp(
 	ctx *EvalContext, op, subOp ComparisonOperator, fn CmpOp, left Datum, right Datums,
 ) (Datum, error) {
 	all := op == All
-	allTrue := true
-	anyTrue := false
+	any := !all
 	sawNull := false
 	for _, elem := range right {
 		if elem == DNull {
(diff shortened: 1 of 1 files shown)


justinj:
I still sort of feel like the structure of this function is sort of wonky, and might make more sense as two separate functions (one for ANY/SOME, one for ALL), wdyt?

---

---

*[pkg/sql/parser/eval.go, line 1734 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-KtNcCcs8AIV-HPkTyPV:-KtNcCcs8AIV-HPkTyPW:bqfdvr0) ([raw file](https://github.com/cockroachdb/cockroach/blob/230cd83c0d7f2707e5a0f3beb5a4258dab16cd12/pkg/sql/parser/eval.go#L1734)):*
> ```Go
> // evalArrayCmp would be called with:
> //   evalDatumsCmp(ctx, LT, Any, CmpOp(LT, leftType, rightParamType), leftDatum, rightArray.Array).
> func evalDatumsCmp(
> ```

I think this function could maybe do with an explanation of the `NULL` behaviour here, since it's not especially obvious what is supposed to happen.

---

*[pkg/sql/parser/eval.go, line 1759 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-KtNbbkt_w5_RNHEkPr2:-KtNbbkt_w5_RNHEkPr3:b-6n1tmc) ([raw file](https://github.com/cockroachdb/cockroach/blob/230cd83c0d7f2707e5a0f3beb5a4258dab16cd12/pkg/sql/parser/eval.go#L1759)):*
> ```Go
> 		res := *b != DBool(not)
> 		if any && res {
> 			// There exist 1 element that evaluates to true, therefore ANY
> ```

nit: `exists one`, but I'm actually not sure this comment is really necessary.

---

*[pkg/sql/parser/eval.go, line 1763 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-KtNbfbUrr1fmO8Vz0Pv:-KtNbfbUrr1fmO8Vz0Pw:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/230cd83c0d7f2707e5a0f3beb5a4258dab16cd12/pkg/sql/parser/eval.go#L1763)):*
> ```Go
> 			return DBoolTrue, nil
> 		} else if all && !res {
> 			// There exist 1 element that evaluates to false, therefore not ALL
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-:-KtNbTgvnI9dhYE8-a4i:b-955gt5)*

richardwu:
It's not my favorite function in the world, but the alternative would result in duplicated code which is arguably less readable and harder to maintain.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-:-KtNtYfxd_Y0QtqgJNpu:bwtvmht)*

richardwu:

---

*[pkg/sql/parser/eval.go, line 1734 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-KtNcCcs8AIV-HPkTyPV:-KtNuW7ePLUHJf505M9e:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/230cd83c0d7f2707e5a0f3beb5a4258dab16cd12/pkg/sql/parser/eval.go#L1734)):*

I think this function could maybe do with an explanation of the `NULL` behaviour here, since it's not especially obvious what is supposed to happen.
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval.go, line 1759 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-KtNbbkt_w5_RNHEkPr2:-KtNuWoQoDHzUXyE7BdB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/230cd83c0d7f2707e5a0f3beb5a4258dab16cd12/pkg/sql/parser/eval.go#L1759)):*

nit: `exists one`, but I'm actually not sure this comment is really necessary.
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval.go, line 1763 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-KtNbfbUrr1fmO8Vz0Pv:-KtNuXFtGi2YirHzyP2F:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/230cd83c0d7f2707e5a0f3beb5a4258dab16cd12/pkg/sql/parser/eval.go#L1763)):*

ditto
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261)*

justinj:
Ok, LGTM then!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18261#-:-KtNz_lXBCM5k9ITpSLI:b-2xlms3)*

PR #18263, opened 2017-09-06
sql: don't rollback txn on missing table name; improve pg_table_is_visible

### LINKED ISSUE #18257: sql: builtins that use the EvalPlanner have unexpected behavior
Opening this as an investigation. I noticed that the query that `psql` uses to evaluate its `\d` command (print all tables) is very slow against CockroachDB:

```
root@:26257/test> SELECT n.nspname AS "Schema", c.relname AS "Name", CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 'f' THEN 'foreign table' END AS "Type", pg_catalog.pg_get_userbyid(c.relowner) AS "Owner" FROM pg_catalog.pg_class AS c LEFT JOIN pg_catalog.pg_namespace AS n ON n.oid = c.relnamespace WHERE ((((c.relkind IN ('r', 'v', 'm', 'S', 'f', '')) AND (n.nspname != 'pg_catalog')) AND (n.nspname != 'information_schema')) AND (n.nspname !~ '^pg_toast')) AND pg_catalog.pg_table_is_visible(c.oid) ORDER BY 1, 2;
+--------+----------+-------+-------+
| Schema |   Name   | Type  | Owner |
+--------+----------+-------+-------+
| test   | accounts | table | NULL  |
| test   | derp     | table | NULL  |
| test   | empty    | table | NULL  |
| test   | fam      | table | NULL  |
| test   | famp     | table | NULL  |
| test   | fampb    | table | NULL  |
| test   | inz      | table | NULL  |
| test   | j1_tbl   | table | NULL  |
| test   | j2_tbl   | table | NULL  |
| test   | m        | table | NULL  |
| test   | nofam    | table | NULL  |
| test   | o        | table | NULL  |
| test   | t        | table | NULL  |
| test   | t1       | table | NULL  |
| test   | t2       | table | NULL  |
| test   | t3       | table | NULL  |
| test   | v        | table | NULL  |
| test   | x        | table | NULL  |
| test   | y        | table | NULL  |
+--------+----------+-------+-------+
(19 rows)

Time: 182.812436ms
```

The trace for this query is 3600 lines long and contains many "rolling back transaction" log lines, like:

```
| 2017-09-06 17:59:17.842303+00:00 | 134ms922µs730ns | === SPAN START: /cockroach.roachpb.Internal/Batch ===                                                                                                                                                        | NULL                                                                           | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842307+00:00 | 134ms926µs494ns | 1 Get                                                                                                                                                                                                        | [n1]                                                                           | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842311+00:00 | 134ms930µs153ns | executing 1 requests                                                                                                                                                                                         | [n1,s1]                                                                        | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842314+00:00 | 134ms933µs805ns | read-only path                                                                                                                                                                                               | [n1,s1,r5/1:/Table/{SystemCon…-11}]                                            | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842318+00:00 | 134ms937µs621ns | command queue                                                                                                                                                                                                | [n1,s1,r5/1:/Table/{SystemCon…-11}]                                            | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842322+00:00 | 134ms941µs513ns | waiting for read lock                                                                                                                                                                                        | [n1,s1,r5/1:/Table/{SystemCon…-11}]                                            | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842348+00:00 | 134ms967µs635ns | read completed                                                                                                                                                                                               | [n1,s1,r5/1:/Table/{SystemCon…-11}]                                            | /cockroach.roachpb.Internal/Batch | (0,228) |
| 2017-09-06 17:59:17.842369+00:00 | 134ms988µs867ns | [client=[::1]:59270,user=root,n1] rolling back transaction                                                                                                                                                   | [client=[::1]                                                                  | sql txn implicit                  | (0,0)   |
```

Removing the call to `pg_table_is_visible`, which is a builtin that uses the `EvalPlanner` to run a query, produces a small trace of only 56 rows without any of these "rolling back" messages.

Not sure yet what's going on but I'm suspicious that there is something messed up about the transactions here.

jordanlewis:
Previously, a failure to find a table name in a database would cause an internal rolled back transaction. This was unnecessary and was misleading in traces. Now, the failure propagates without failing the transaction.

Also, `pg_table_is_visible` didn't qualify the database of the
tables `pg_namespace` and `pg_class` in an internal query, leading to
many pointless failed table lookups and potential bad behavior. Now it does.

Fixes #18257.

### DIFF
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -1247,20 +1247,25 @@ func (m *LeaseManager) resolveName(
 ) (sqlbase.ID, error) {
 	nameKey := tableKey{dbID, tableName}
 	key := nameKey.Key()
-	var id sqlbase.ID
-	err := m.db.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
+	id := sqlbase.InvalidID
+	if err := m.db.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
 		txn.SetFixedTimestamp(timestamp)
 		gr, err := txn.Get(ctx, key)
 		if err != nil {
 			return err
 		}
 		if !gr.Exists() {
-			return sqlbase.ErrDescriptorNotFound
+			return nil
 		}
 		id = sqlbase.ID(gr.ValueInt())
 		return nil
-	})
-	return id, err
+	}); err != nil {
+		return id, err
+	}
+	if id == sqlbase.InvalidID {
+		return id, sqlbase.ErrDescriptorNotFound
+	}
+	return id, nil
 }
 
 // Acquire acquires a read lease for the specified table ID valid for
--- a/pkg/sql/parser/pg_builtins.go
+++ b/pkg/sql/parser/pg_builtins.go
@@ -316,7 +316,7 @@ var pgBuiltins = map[string][]Builtin{
 			fn: func(ctx *EvalContext, args Datums) (Datum, error) {
 				oid := args[0]
 				t, err := ctx.Planner.QueryRow(ctx.Ctx(),
-					"SELECT nspname FROM pg_class c JOIN pg_namespace n ON c.relnamespace=n.oid "+
+					"SELECT nspname FROM pg_catalog.pg_class c JOIN pg_catalog.pg_namespace n ON c.relnamespace=n.oid "+
 						"WHERE c.oid=$1 AND nspname=ANY(current_schemas(true));", oid)
 				if err != nil {
 					return nil, err
--- a/pkg/sql/sqlbase/structured.go
+++ b/pkg/sql/sqlbase/structured.go
@@ -38,6 +38,9 @@ import (
 // ID is a custom type for {Database,Table}Descriptor IDs.
 type ID parser.ID
 
+// InvalidID is the uninitialised descriptor id.
+const InvalidID ID = 0
+
 // IDs is a sortable list of IDs.
 type IDs []ID
 
@@ -82,7 +85,7 @@ const (
 	InterleavedFormatVersion
 )
 
-// MutationID is custom type for TableDescriptor mutations.
+// MutationID is a custom type for TableDescriptor mutations.
 type MutationID uint32
 
 // InvalidMutationID is the uninitialised mutation id.


justinj:

---

*[pkg/sql/lease.go, line 1263 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18263#-KtNi-q0DbWUqKuCZKSv:-KtNi-q0DbWUqKuCZKSw:bib9xj9) ([raw file](https://github.com/cockroachdb/cockroach/blob/60c6e46109613fd50de698550ab1ea86b7d490e8/pkg/sql/lease.go#L1263)):*
> ```Go
> 		return nil
> 	})
> 	if id == sqlbase.InvalidID {
> ```

it's not super clear to me if this is the result of
1. the `id = sqlbase.ID(gr.ValueInt())` line not getting run, or
2. `gr.ValueInt()` returning 0,
or both, could we initialize `id` to `InvalidID` if 1 is the case?
Also is it an issue that we're potentially throwing away the `err` result from `m.db.Txn`? Seems like we could potentially throw away the result of `txn.Get`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18263)*

jordanlewis:

---

*[pkg/sql/lease.go, line 1263 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18263#-KtNi-q0DbWUqKuCZKSv:-KtNjywhgrYvH5lLxYoM:bm3bzsy) ([raw file](https://github.com/cockroachdb/cockroach/blob/60c6e46109613fd50de698550ab1ea86b7d490e8/pkg/sql/lease.go#L1263)):*

it's not super clear to me if this is the result of
1. the `id = sqlbase.ID(gr.ValueInt())` line not getting run, or
2. `gr.ValueInt()` returning 0,
or both, could we initialize `id` to `InvalidID` if 1 is the case?
Also is it an issue that we're potentially throwing away the `err` result from `m.db.Txn`? Seems like we could potentially throw away the result of `txn.Get`.
</blockquote></details>

Hmm yeah, good point. I think we do this kind of weird looking zero-value manipulation all over the place in the codebase, so it's not really out of place, but I agree that it's generally confusing. Now `id` is explicitly initialized to `InvalidID`.

Yes I was accidentally throwing away the error. Good catch, thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18263)*

justinj:
:lgtm:!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18263#-:-KtNmyIONtPKO367YUbY:bg7igkw)*

PR #18266, opened 2017-09-06
sql: properly handle parenthesized/nested tuples as operands for ANY/ALL operations

richardwu:
Type checking for tuples in ANY/ALL operations were added in #18094.

It was initially implemented such that it assumed the `right Expr` was either a literal `tuple` (of which it would require every element be equivalent to the type of `left` _after_ upcasting), an array, or a subquery.

We did not however check for tuples nested within parentheses, thus a semantically-incorrect expression such as
```
SELECT 1 = ANY (((1.0, 1.1)))
```
would not type error but succeed.


justinj:
Can you see if there's a way this can be done without `AnnotateParens` like how we are able to typecheck something like `(1.2) + ((1))` without it? It seems fishy to me that we need this extra mechanism now but we didn't before. If not could you add a (brief) comment explaining why to `AnnotateParens`?

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/expr.go, line 272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtO88W1jRkBPRWZQD64:-KtO88W1jRkBPRWZQD65:btbg449) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr.go#L272)):*
> ```Go
> 	exprParen := expr
> 	for {
> 		if p, ok := exprParen.(*ParenExpr); ok {
> ```

I think you can maybe write this with fewer jumpy statements:

```
for {
  if ... ; ok {
    p.typ = typ
    exprParen = p.Expr
  } else {
    return expr
  }
}
```
?

Also, you never use the return value of this so consider either making this not return a value or using its return value. Since it's mutative I would lean towards not returning a value to emphasize that.

---

*[pkg/sql/parser/expr_test.go, line 237 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtO5pYvjRkns_P1UAi4:-KtO5pYvjRkns_P1UAi5:b3emd55) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr_test.go#L237)):*
> ```Go
> 	}
> 
> 	for i, test := range testExprs {
> ```

you can make these subtests (which can give you slightly cleaner output) if you use `t.Run...`

---

*[pkg/sql/parser/expr_test.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOCOIbPhLs-LjNq1rB:-KtOCOIbPhLs-LjNq1rC:bwnthz) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr_test.go#L244)):*
> ```Go
> 		annotated := AnnotateParens(expr, typ)
> 
> 		n := 0
> ```

I think you can put this on the `for` line

---

*[pkg/sql/parser/type_check.go, line 913 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtO610xVcFlYj2yIHyB:-KtO610xVcFlYj2yIHyC:b-8so11) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/type_check.go#L913)):*
> ```Go
> // For example, (1, 2.5) with required TypeDecimal would raise a sane error whereas (1.0, 2.5) with required TypeDecimal would pass.
> func typeCheckAndRequireTupleElems(ctx *SemaContext, expr Expr, required Type) (TypedExpr, error) {
> 	// Retrieve reference to (possibly nested) tuple expression
> ```

nit: period

---

*[pkg/sql/parser/type_check.go, line 925 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOCYH0LS12JwpDfeK9:-KtOCYH0LS12JwpDfeKA:bz9o5is) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/type_check.go#L925)):*
> ```Go
> 		tuple.types[i] = typedExpr.ResolvedType()
> 	}
> 	// We cannot simply return tuple since we want the original expression returned and with
> ```

super nit: wrap "tuple" with backticks so it's clearer that this is referring to a variable

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtOB3oKe7O9lBRPxVXg:bs8roon)*

richardwu:
Thanks for bringing this up! I decided to leave in `AnnotateParens` since another part of the code seems to use it; my guess is that codepath could also be refactored to do the same.

I've verified that the types are propagated to the surrounding `ParenExpr`s using log statements. I don't think there's a clean/kosher way to test that the type annotations propagate, thoughts?

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/expr.go, line 272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtO88W1jRkBPRWZQD64:-KtORZk03tXe87IuMBQp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr.go#L272)):*

I think you can maybe write this with fewer jumpy statements:

```
for {
  if ... ; ok {
    p.typ = typ
    exprParen = p.Expr
  } else {
    return expr
  }
}
```
?

Also, you never use the return value of this so consider either making this not return a value or using its return value. Since it's mutative I would lean towards not returning a value to emphasize that.
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr_test.go, line 237 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtO5pYvjRkns_P1UAi4:-KtOSaziM7In5_PNayol:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr_test.go#L237)):*

you can make these subtests (which can give you slightly cleaner output) if you use `t.Run...`
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr_test.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOCOIbPhLs-LjNq1rB:-KtOSbn8d2sxVXUR7qWi:b-smdtlg) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr_test.go#L244)):*

I think you can put this on the `for` line
</blockquote></details>

I needed to scope the `n` so I can compare it with `test.count` afterwards.

---

*[pkg/sql/parser/type_check.go, line 913 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtO610xVcFlYj2yIHyB:-KtOSnZMLC7ytOwFo58H:bhd2qz) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/type_check.go#L913)):*

nit: period
</blockquote></details>

Removed comment so n/a anymore.

---

*[pkg/sql/parser/type_check.go, line 925 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOCYH0LS12JwpDfeK9:-KtOSte7Ku0slKjKu3PH:bhd2qz) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/type_check.go#L925)):*

super nit: wrap "tuple" with backticks so it's clearer that this is referring to a variable
</blockquote></details>

Removed comment so n/a anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtOQaWUpNjw_UDunBAa:b-ugpvxg)*

justinj:
(I think you didn't push up your changes)

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtOUYP4IaXU5P5YC8mE:b-usyww7)*

justinj:
One more thing I don't fully understand - why do we need to type every layer of parentheses rather than just typing the outermost ones?

Otherwise LGTM but @knz might have comments so wait for him to take a look.

---

---

*[pkg/sql/parser/expr.go, line 278 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOVmL_9HDxyx5c44V7:-KtOVmL_9HDxyx5c44V8:bf5cqla) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/expr.go#L278)):*
> ```Go
> 			return
> 		}
> ```

nit: extra line

---

*[pkg/sql/parser/expr_test.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOCOIbPhLs-LjNq1rB:-KtOVwjVcLLR-LIowE_W:b-hedxkb) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a7e08e9e0477a4409517c95f729e2a37b0867b7/pkg/sql/parser/expr_test.go#L244)):*

I needed to scope the `n` so I can compare it with `test.count` afterwards.
</blockquote></details>

Ah yeah, my bad!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtOW56JxpwQ_AM62LU1:bdfaghx)*

knz:
I think you're still missing one intellectual step about the power of functional rewrites.
The entire TypeCheck function is free to *replace* arbitrary sub-trees of an expression by another.

In particular, it is valid to remove any remaining ParenExpr here. So strip away! And don't bother to "annotate types on them". See my comments below for some suggestions, but please also grep for `ParenExpr` throughout the `parser` and `sql` packages, and replace all such loops by `StripParens`. There is no honor in respecting parentheses. They are a purely grammatical construct, they have no semantic value whatsoever and do not deserve to be preserved.

---

Reviewed 1 of 4 files at r1, 2 of 3 files at r2.
Review status: 3 of 4 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/expr.go, line 269 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRGr2qXTSA_QNWHgwH:-KtRGr2qXTSA_QNWHgwI:bfbj23y) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/expr.go#L269)):*
> ```Go
> // AnnotateParens annotates the parentheses surrounding the first non-parentheses
> // expression with the given type.
> func AnnotateParens(expr Expr, typ Type) {
> ```

wut? no!!  

:laughing: 

---

*[pkg/sql/parser/type_check.go, line 918 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRG3dxYJrLU_w5jcFw:-KtRG3dxYJrLU_w5jcFx:bge9825) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L918)):*
> ```Go
> 	// If the expression is wrapped in parentheses, do a recursive call
> 	// and type annotate the parentheses with the resulting tuple type.
> 	if parenExpr, ok := expr.(*ParenExpr); ok {
> ```

Man you're making this so complicated!

Replace this entire conditional by
```
expr = StripParens(expr)
```

bam!

---

*[pkg/sql/parser/type_check.go, line 980 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHgI9Y7_D-hTMhIs3:-KtRHgI9Y7_D-hTMhIs4:b-yuf2fr) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L980)):*
> ```Go
> 	var cmpTypeLeft Type
> 	var cmpTypesRight []Type
> 	var leftTyped, rightTyped TypedExpr
> ```

Insert `left = StripParens(left); right = StripParens(right)` here.

---

*[pkg/sql/parser/type_check.go, line 981 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHjefRwz1NZjsLVju:-KtRHjefRwz1NZjsLVjv:b-jsqtn0) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L981)):*
> ```Go
> 	var cmpTypesRight []Type
> 	var leftTyped, rightTyped TypedExpr
> 	if array, isConstructor := StripParens(right).(*Array); isConstructor {
> ```

Remove the call to `StripParens` here.

---

*[pkg/sql/parser/type_check.go, line 1006 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHn805ocjpwHRDtcV:-KtRHn805ocjpwHRDtcW:b-ekltov) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L1006)):*
> ```Go
> 		array.typ = TArray{retType}
> 
> 		AnnotateParens(right, array.typ)
> ```

Remove (see above -- the parentheses are gone if you arrive here.)

---

*[pkg/sql/parser/type_check.go, line 1024 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHsbfNTunSl-9Bmup:-KtRHsbfNTunSl-9Bmuq:ba2czar) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L1024)):*
> ```Go
> 		cmpTypeLeft = leftTyped.ResolvedType()
> 
> 		if _, ok := StripParens(right).(*Tuple); ok {
> ```

remove the call to StripParens here, already done above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtRHyTZnX5Xq5Uk44_-:bz74zgd)*

richardwu:
I'm wondering if we might run into bugs later on with precedence when we serialize it for distsql. For example, this came up with negative numbers https://github.com/cockroachdb/cockroach/issues/15617. 

I can't think of an example off the top of my head (perhaps someone more familiar with SQL grammar gotchas would fare better) where serializing the functional rewrite of an expression without parentheses would cause mis-parsing, but I think you're absolutely correct!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtReUVfsqLT_GUGcmwJ:b-ogfb41)*

richardwu:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[pkg/sql/parser/type_check.go, line 918 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRG3dxYJrLU_w5jcFw:-KtRw-dbCR5djonT4yVO:bhjcdnk) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L918)):*

Man you're making this so complicated!

Replace this entire conditional by
```
expr = StripParens(expr)
```

bam!

</blockquote></details>

I've decided to require that the expression passed in be strictly a Tuple and let parentheses unwrapping be handled by its callers.

---

*[pkg/sql/parser/type_check.go, line 980 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHgI9Y7_D-hTMhIs3:-KtRvzdyw3jYLe2OifmJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L980)):*

Insert `left = StripParens(left); right = StripParens(right)` here.
</blockquote></details>

Done.

---

*[pkg/sql/parser/type_check.go, line 981 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHjefRwz1NZjsLVju:-KtRwDnM1jxaTMgh5iEr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L981)):*

Remove the call to `StripParens` here.
</blockquote></details>

Done.

---

*[pkg/sql/parser/type_check.go, line 1006 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHn805ocjpwHRDtcV:-KtRwEYDvrgtmR8_vUyU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L1006)):*

Remove (see above -- the parentheses are gone if you arrive here.)
</blockquote></details>

Done.

---

*[pkg/sql/parser/type_check.go, line 1024 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRHsbfNTunSl-9Bmup:-KtRwFte7fKRlSukwGW6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/type_check.go#L1024)):*

remove the call to StripParens here, already done above
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr.go, line 269 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtRGr2qXTSA_QNWHgwH:-KtRvpiTckJ9eaUKtSvM:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/expr.go#L269)):*

wut? no!!  

:laughing: 
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr.go, line 278 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-KtOVmL_9HDxyx5c44V7:-KtRwHKREbem-Q6hQhHi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8e23bafe3c1cd47797e8719a7ba266cf759bc67/pkg/sql/parser/expr.go#L278)):*

nit: extra line
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266)*

richardwu:
In order to keep things consistent across the board, I've decided to strip ParenExpr from all expressions in general during type checking.

It would be great to get another look/LGTM if this is okay (@knz, @justinj). I can also move this to another PR if necessary!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtSiIJYR0M4hbjlwhDp:bec1hpq)*

knz:
:lgtm:

---

Reviewed 5 of 5 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18266#-:-KtXfxg1N0gJuL2M7FiG:bnfp4nl)*

PR #18267, opened 2017-09-06
distsql: knobs for temp storage size

### LINKED ISSUE #17663: distsql: expose knob for limiting query disk space usage
Currently, temporary disk space usage is hardcoded to 25% of total disk space. This is reasonable, but results in queries taking a long time to fail, when they are planned pathologically. Case in point: [TPC-H query 9](https://github.com/cockroachdb/loadgen/blob/master/tpch/query9.go) is a 6 way join. The intention of the query is to suss out if we are ordering our joins cleverly, to keep the intermediate state to a manageable level. Of course, we are not, which currently results in us generating gigantic intermediate cross-join state, until we hit the disk accounting limit and error out.

On a recent run on @asubiotto's laptop, it generated 62 gigabytes of state, running for 2 hours, before hitting the limit. [1] This gives us the sense that this knob could be tuned down by default, to give more responsive answers to pathological cross-join queries, and we can allow users to turn this up if they're explicitly running large queries with very large working sets. We could make this knob an environment variable, or a cluster setting.

As an aside, it is worth noting that bad things happen if a Cockroach node hits 100% disk space utilization. And with the current 25% limit, a 75% full disk appears relatively safe to an operator, but is in fact at the risk of a single adversarial query killing the entire cluster (which distsql will helpfully schedule on all nodes).

[1] As an aside, this is impressive throughput!

RaduBerinde:
Adding cluster settings for max temp disk space usage, both as a percentage of
total capacity and as an absolute number (whichever is smaller).

The default limit is MIN(10% of total disk capacity, 32GB).

Fixes #17663.

### DIFF
--- a/pkg/sql/distsqlrun/hashjoiner.go
+++ b/pkg/sql/distsqlrun/hashjoiner.go
@@ -161,7 +161,7 @@ func (h *hashJoiner) Run(ctx context.Context, wg *sync.WaitGroup) {
 	}
 
 	st := h.flowCtx.Settings
-	useTempStorage := (distSQLUseTempStorage.Get(&st.SV) && distSQLUseTempStorageJoins.Get(&st.SV)) ||
+	useTempStorage := settingUseTempStorageJoins.Get(&st.SV) ||
 		h.flowCtx.testingKnobs.MemoryLimitBytes > 0 ||
 		h.testingKnobMemFailPoint != unset
 	rowContainerMon := h.flowCtx.EvalCtx.Mon
@@ -170,7 +170,7 @@ func (h *hashJoiner) Run(ctx context.Context, wg *sync.WaitGroup) {
 		// The hashJoiner will overflow to disk if this limit is not enough.
 		limit := h.flowCtx.testingKnobs.MemoryLimitBytes
 		if limit <= 0 {
-			limit = workMemBytes
+			limit = settingWorkMemBytes.Get(&st.SV)
 		}
 		limitedMon := mon.MakeMonitorInheritWithLimit("hashjoiner-limited", limit, rowContainerMon)
 		limitedMon.Start(ctx, rowContainerMon, mon.BoundAccount{})
--- a/pkg/sql/distsqlrun/hashjoiner_test.go
+++ b/pkg/sql/distsqlrun/hashjoiner_test.go
@@ -686,7 +686,7 @@ func TestHashJoinerDrain(t *testing.T) {
 	// Since the use of external storage overrides h.initialBufferSize, disable
 	// it for this test.
 	settings := cluster.MakeTestingClusterSettings()
-	distSQLUseTempStorage.Override(&settings.SV, false)
+	settingUseTempStorageJoins.Override(&settings.SV, false)
 	flowCtx := FlowCtx{Settings: settings, EvalCtx: evalCtx}
 
 	post := PostProcessSpec{Projection: true, OutputColumns: outCols}
--- a/pkg/sql/distsqlrun/server.go
+++ b/pkg/sql/distsqlrun/server.go
@@ -15,6 +15,7 @@
 package distsqlrun
 
 import (
+	"fmt"
 	"io"
 	"time"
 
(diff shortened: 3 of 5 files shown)


asubiotto:
It might be interesting to also add a way to override the default. As in, imagine an operator that wants to set the disk space usage to x bytes where x > 10% of disk capacity. It might be a bit surprising to have to change the percentage setting as well for the x bytes to be observed. However, this can get a bit messy (which setting is the one that actually gets used in the end?), what do you think?

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlrun/server.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtNwC2SQkKnChP7urmV:-KtNwC2SQkKnChP7urmW:bir8iie) ([raw file](https://github.com/cockroachdb/cockroach/blob/1871a1efe77c9bc11ab8a8930109cf0b2088648d/pkg/sql/distsqlrun/server.go#L98)):*
> ```Go
> )
> 
> var settingWorkMemBytes = settings.RegisterByteSizeSetting(
> ```

Why make this a cluster setting? If there are clusters with a bunch of nodes with varying RAM I'm not sure it makes sense to have all of them be able to observe a single cluster setting. It doesn't seem like this would be generally useful.

---

*[pkg/sql/distsqlrun/server.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtNwu0mFeq5cSOHvhcg:-KtNwu0mFeq5cSOHvhch:b-yugtgo) ([raw file](https://github.com/cockroachdb/cockroach/blob/1871a1efe77c9bc11ab8a8930109cf0b2088648d/pkg/sql/distsqlrun/server.go#L116)):*
> ```Go
> )
> 
> var settingDiskBudgetAbsolute = settings.RegisterByteSizeSetting(
> ```

This is another thing I would make an env var due to potentially differing disk sizes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-:-KtNxQjeWtE0dESPoFkb:b-dspsvx)*

RaduBerinde:
I added a "see the other setting" note in the descriptions. I don't see a clean way to do it and still have this smaller-of-the-two default. It's easy enough to set the percent to 100 if necessary.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/server.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtNwC2SQkKnChP7urmV:-KtO48tyXTk66kF-XaST:b-fze1gs) ([raw file](https://github.com/cockroachdb/cockroach/blob/1871a1efe77c9bc11ab8a8930109cf0b2088648d/pkg/sql/distsqlrun/server.go#L98)):*

Why make this a cluster setting? If there are clusters with a bunch of nodes with varying RAM I'm not sure it makes sense to have all of them be able to observe a single cluster setting. It doesn't seem like this would be generally useful.
</blockquote></details>

It's useful because you can change it easily, unlike an env var (and can actually change it while the server is running). E.g. for a quick test you can increase or decrease it and see what happens.

The value is pretty small compared to the amount of RAM we expect on machines, realistically I don't think we will have many cases where we want different values across machines.

---

*[pkg/sql/distsqlrun/server.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtNwu0mFeq5cSOHvhcg:-KtO4d2U_Sip3h4n7D0S:bc31ruf) ([raw file](https://github.com/cockroachdb/cockroach/blob/1871a1efe77c9bc11ab8a8930109cf0b2088648d/pkg/sql/distsqlrun/server.go#L116)):*

This is another thing I would make an env var due to potentially differing disk sizes.
</blockquote></details>

Like with work mem, this is a fairly small value compared to typical disk sizes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-:-KtO5HCR5SPvYpxKWEqX:bjt50cb)*

couchand:
Bikeshedding about the setting name: it might be a good idea to treat `sql.distsql.tempstorage` as a pure namespace and move the setting to enable/disable to something like `sql.distsql.tempstorage.enabled`.  Makes it a little more self-documenting.

asubiotto:
That's a good idea. @jesse, putting these changes on your radar.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/server.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtNwC2SQkKnChP7urmV:-KtRlIFITK66VqWbK977:b1rfxpc) ([raw file](https://github.com/cockroachdb/cockroach/blob/1871a1efe77c9bc11ab8a8930109cf0b2088648d/pkg/sql/distsqlrun/server.go#L98)):*

It's useful because you can change it easily, unlike an env var (and can actually change it while the server is running). E.g. for a quick test you can increase or decrease it and see what happens.

The value is pretty small compared to the amount of RAM we expect on machines, realistically I don't think we will have many cases where we want different values across machines.
</blockquote></details>

Sounds good. Do you want to keep the env var in that case?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-:-KtRlPtDMMNHu7uNXyC-:bwr976c)*

RaduBerinde:
*[pkg/sql/distsqlrun/server.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtNwC2SQkKnChP7urmV:-KtRrpdK_PzkWWW86K9D:b-jffatk) ([raw file](https://github.com/cockroachdb/cockroach/blob/1871a1efe77c9bc11ab8a8930109cf0b2088648d/pkg/sql/distsqlrun/server.go#L98)):*

Sounds good. Do you want to keep the env var in that case?
</blockquote></details>

Good question, I'm leaning towards removing it because it might not work as expected (it is just a default value so it's effectively ignored once the cluster is set up)..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267)*

RaduBerinde:
@couchand yeah I agree. Though I think we can just remove `sql.distsql.tempstorage`, we don't need a "master switch" when we have switches for each use (sorts, join)

asubiotto:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/server.go, line 211 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtS1ZEP6jEp3V3A_FpR:-KtS1ZEP6jEp3V3A_FpS:b-nw8t2j) ([raw file](https://github.com/cockroachdb/cockroach/blob/91e33813b4fa69312e11a79928c480296dd36190/pkg/sql/distsqlrun/server.go#L211)):*
> ```Go
> 		diskMonitorBudget = maxBytes
> 	}
> 	ds.diskMonitor = mon.MakeMonitor(
> ```

I don't think this would work as expected. Previous to this pr the percentage was for *all* queries (this monitor is used by all processors). If you want to make this per-query you would have to create a child monitor with another limit in the processor. To mirror the memory monitor this would be have a percentage limit defined for all queries here and then create a child monitor with the absolute limit in the processors. This would effectively respect MIN(percentage, absolute).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267)*

RaduBerinde:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/server.go, line 211 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267#-KtS1ZEP6jEp3V3A_FpR:-KtS3_z4B6d1cOVrUh57:bdrxil8) ([raw file](https://github.com/cockroachdb/cockroach/blob/91e33813b4fa69312e11a79928c480296dd36190/pkg/sql/distsqlrun/server.go#L211)):*

I don't think this would work as expected. Previous to this pr the percentage was for *all* queries (this monitor is used by all processors). If you want to make this per-query you would have to create a child monitor with another limit in the processor. To mirror the memory monitor this would be have a percentage limit defined for all queries here and then create a child monitor with the absolute limit in the processors. This would effectively respect MIN(percentage, absolute).
</blockquote></details>

Ah, my mistake, I will update the descriptions. I think we always want a global limit even if we will also add per-query limits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18267)*

asubiotto (approved):

RaduBerinde:
@arjunravinarayan are you ok with the new cluster settings and defaults?

PR #18268, opened 2017-09-06
sql: don't auto-retry if we've already sent results to the client

andreimatei:
See individual commits.

cc @jordanlewis

### DIFF
--- a/pkg/internal/client/db_test.go
+++ b/pkg/internal/client/db_test.go
@@ -407,6 +407,7 @@ func TestCommonMethods(t *testing.T) {
 		{txnType, "IsSerializableRestart"}:           {},
 		{txnType, "NewBatch"}:                        {},
 		{txnType, "Exec"}:                            {},
+		{txnType, "PrepareForRetry"}:                 {},
 		{txnType, "ResetDeadline"}:                   {},
 		{txnType, "Run"}:                             {},
 		{txnType, "Send"}:                            {},
@@ -427,6 +428,8 @@ func TestCommonMethods(t *testing.T) {
 		{txnType, "UserPriority"}:                    {},
 		{txnType, "AnchorKey"}:                       {},
 		{txnType, "ID"}:                              {},
+		{txnType, "IsAborted"}:                       {},
+		{txnType, "IsCommitted"}:                     {},
 	}
 
 	for b := range omittedChecks {
--- a/pkg/internal/client/txn.go
+++ b/pkg/internal/client/txn.go
@@ -154,10 +154,21 @@ func (txn *Txn) IsFinalized() bool {
 	return txn.mu.finalized
 }
 
+// status returns the txn proto status field.
 func (txn *Txn) status() roachpb.TransactionStatus {
 	txn.mu.Lock()
 	defer txn.mu.Unlock()
 	return txn.mu.Proto.Status
+}
+
+// IsCommitted returns true if the transaction has the committed status.
+func (txn *Txn) IsCommitted() bool {
+	return txn.status() == roachpb.COMMITTED
+}
+
+// IsAborted returns true if the transaction has the aborted status.
+func (txn *Txn) IsAborted() bool {
+	return txn.status() == roachpb.ABORTED
 }
 
 // SetUserPriority sets the transaction's user priority. Transactions default to
@@ -709,6 +720,14 @@ func (e *AutoCommitError) Error() string {
 // to clean up the transaction before returning an error. In case of
 // TransactionAbortedError, txn is reset to a fresh transaction, ready to be
 // used.
+//
+// TODO(andrei): The SQL Executor was the most complex user of this interface.
+// It needed fine control by using TxnExecOptions. Now SQL no longer uses this
+// interface, so it's time to see how it can be simplified. TxnExecOptions can
+// probably go away, and so can AutoCommitError. The method should also be
+// documented to not allow calls concurrent with any other txn use, so that the
+// Commit() call inside it is clearly correct (as in, it won't run concurrently
+// with other txn calls).
 func (txn *Txn) Exec(
 	ctx context.Context, opt TxnExecOptions, fn func(context.Context, *Txn, *TxnExecOptions) error,
 ) (err error) {
@@ -722,11 +741,7 @@ func (txn *Txn) Exec(
 		err = fn(ctx, txn, &opt)
 
 		if err == nil && opt.AutoCommit {
-			// Copy the status out of the Proto under lock. Making decisions on
-			// this later is not thread-safe, but the commutativity property of
-			// transactions assure that reasoning about the situation is straightforward.
 			status := txn.status()
-
 			switch status {
 			case roachpb.ABORTED:
 				// TODO(andrei): Until 7881 is fixed.
@@ -766,13 +781,20 @@ func (txn *Txn) Exec(
 			break
 		}
 
-		txn.commitTriggers = nil
-
-		log.VEventf(ctx, 2, "automatically retrying transaction: %s because of error: %s",
-			txn.DebugName(), err)
+		txn.PrepareForRetry(ctx, err)
 	}
 
 	return err
+}
+
+// PrepareForRetry needs to be called before an retry to perform some
+// book-keeping.
+//
+// TODO(andrei): I think this is called in the wrong place. See #18170.
+func (txn *Txn) PrepareForRetry(ctx context.Context, err error) {
+	txn.commitTriggers = nil
+	log.VEventf(ctx, 2, "automatically retrying transaction: %s because of error: %s",
+		txn.DebugName(), err)
 }
 
 // IsRetryableErrMeantForTxn returns true if err is a retryable
--- a/pkg/sql/copy.go
+++ b/pkg/sql/copy.go
@@ -346,4 +346,4 @@ func (CopyDataBlock) StatementType() parser.StatementType { return parser.RowsAf
 
 // StatementTag returns a short string identifying the type of statement.
 func (CopyDataBlock) StatementTag() string { return "" }
-func (CopyDataBlock) String() string       { return "" }
+func (CopyDataBlock) String() string       { return "CopyDataBlock" }
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -344,6 +344,19 @@ type ExecutorTestingKnobs struct {
 
 	// DistSQLPlannerKnobs are testing knobs for distSQLPlanner.
 	DistSQLPlannerKnobs DistSQLPlannerTestingKnobs
+
+	// BeforeAutoCommit is called when the Executor is about to commit the KV
+	// transaction after running a statement in an implicit transaction, allowing
+	// tests to inject errors into that commit.
+	// If an error is returned, that error will be considered the result of
+	// txn.Commit(), and the txn.Commit() call will not actually be
+	// made. If no error is returned, txn.Commit() is called normally.
+	//
+	// Note that this is not called if the SQL statement representing the implicit
+	// transaction has committed the KV txn itself (e.g. if it used the 1-PC
+	// optimization). This is only called when the Executor is the one doing the
+	// committing.
+	BeforeAutoCommit func(ctx context.Context, stmt string) error
 }
 
 // DistSQLPlannerTestingKnobs is used to control internals of the distSQLPlanner
@@ -569,6 +582,9 @@ func (e *Executor) Prepare(
 // Note that we will only receive an error even if we run a successful statement
 // followed by a statement which has an error then the caller will only receive
 // the error, however the first statement will have been executed.
+//
+// If no error is returned, the caller has to call Close() on the returned
+// StatementResults.
 func (e *Executor) ExecuteStatementsBuffered(
 	session *Session, stmts string, pinfo *parser.PlaceholderInfo, expectedNumResults int,
 ) (StatementResults, error) {
@@ -741,7 +757,7 @@ func (e *Executor) execRequest(
 		// statements or only one, so just pretend there was one.
 		if txnState.mu.txn != nil {
 			// Rollback the txn.
-			txnState.updateStateAndCleanupOnErr(err, e)
+			err = txnState.updateStateAndCleanupOnErr(err, e)
 		}
 		return err
 	}
@@ -767,7 +783,6 @@ func (e *Executor) execParsed(
 		// that is encountered resets stmts.
 
 		inTxn := txnState.State() != NoTxn
-		execOpt := client.TxnExecOptions{}
 		// Figure out the statements out of which we're going to try to consume
 		// this iteration. If we need to create an implicit txn, only one statement
 		// can be consumed.
@@ -775,13 +790,18 @@ func (e *Executor) execParsed(
 		// If protoTS is set, the transaction proto sets its Orig and Max timestamps
 		// to it each retry.
 		var protoTS *hlc.Timestamp
+
+		// autoCommit will be set if we're now starting an implicit transaction and
+		// thus should automatically commit after we've run the statement.
+		autoCommit := false
+
 		// If we're not in a transaction, then the next statement is part of a new
 		// transaction (implicit txn or explicit txn). We do the corresponding state
 		// reset.
 		if !inTxn {
 			// Detect implicit transactions - they need to be autocommitted.
 			if _, isBegin := stmts[0].AST.(*parser.BeginTransaction); !isBegin {
-				execOpt.AutoCommit = true
+				autoCommit = true
 				stmtsToExec = stmtsToExec[:1]
 				// Check for AS OF SYSTEM TIME. If it is present but not detected here,
 				// it will raise an error later on.
@@ -800,14 +820,13 @@ func (e *Executor) execParsed(
 			}
 			txnState.resetForNewSQLTxn(
 				e, session,
-				execOpt.AutoCommit, /* implicitTxn */
-				false,              /* retryIntent */
+				autoCommit, /* implicitTxn */
+				false,      /* retryIntent */
 				e.cfg.Clock.PhysicalTime(), /* sqlTimestamp */
 				session.DefaultIsolationLevel,
 				roachpb.NormalUserPriority,
 			)
 		}
-		execOpt.AutoRetry = txnState.State() == AutoRetry
 
 		if txnState.State() == NoTxn {
 			panic("we failed to initialize a txn")
@@ -818,15 +837,9 @@ func (e *Executor) execParsed(
 
 		// Track if we are retrying this query, so that we do not double count.
 		automaticRetryCount := 0
-		txnClosure := func(ctx context.Context, txn *client.Txn, opt *client.TxnExecOptions) error {
+
+		txnClosure := func(ctx context.Context) (transitionToOpen bool, _ error) {
 			defer func() { automaticRetryCount++ }()
-			if txnState.TxnIsOpen() && txnState.mu.txn != txn {
-				panic(fmt.Sprintf("closure wasn't called in the txn we set up for it."+
-					"\ntxnState.mu.txn:%+v\ntxn:%+v\ntxnState:%+v", txnState.mu.txn, txn, txnState))
-			}
-			txnState.mu.Lock()
-			txnState.mu.txn = txn
-			txnState.mu.Unlock()
 
 			if protoTS != nil {
 				txnState.mu.txn.SetFixedTimestamp(*protoTS)
@@ -835,7 +848,7 @@ func (e *Executor) execParsed(
 			// Some results may have been produced by a previous attempt.
 			txnState.txnResults.Reset(session.Ctx())
 			var err error
-			remainingStmts, err = runTxnAttempt(
+			remainingStmts, transitionToOpen, err = runTxnAttempt(
 				e, session, stmtsToExec, pinfo, origState,
 				!inTxn /* txnPrefix */, avoidCachedDescriptors, automaticRetryCount, txnState.txnResults)
 
@@ -854,42 +867,129 @@ func (e *Executor) execParsed(
 				}
 			}
 
-			return err
+			return transitionToOpen, err
 		}
-		// This is where the magic happens - we ask db to run a KV txn and possibly retry it.
-		txn := txnState.mu.txn // this might be nil if the txn was already aborted.
-		err := txn.Exec(session.Ctx(), execOpt, txnClosure)
+
+		// We now run the prefix of stmtsToExec that belong to the current
+		// transaction via txnClosure. We do this in a loop because we may want to
+		// do automatic retries, if possible. Whether or not we can do auto-retries
+		// depends on the state before we execute statements in the batch.
+		//
+		// TODO(andrei): It's unfortunate that we're keeping track of what the state
+		// was before running the statements. It'd be great if the state in which we
+		// find ourselves after running the statements told us if we can auto-retry.
+		// A way to do that is to introduce another state called AutoRestartWait,
+		// similar to RestartWait. We'd enter this state whenever we're in state
+		// AutoRetry and we get a retriable error.
+		txnCanBeAutoRetried := txnState.State() == AutoRetry
+		var err error
+		for {
+			// transitionToOpen will tell us if we need to transition from AutoRetry
+			// to Open after executing the statements.
+			var transitionToOpen bool
+
+			// Run some statements.
+			transitionToOpen, err = txnClosure(session.Ctx())
+
+			// Sanity checks.
+			if err != nil && txnState.TxnIsOpen() {
+				log.Fatalf(session.Ctx(), "statement(s) generated error but state is %s. err: %s",
+					txnState.State(), err)
+			} else {
+				state := txnState.State()
+				if err == nil && (state == Aborted || state == RestartWait) {
+					crash := true
+					if len(stmtsToExec) > 0 {
+						// SHOW TRANSACTION STATUS statements are always allowed regardless
+						// of the transaction state.
+						if _, ok := stmtsToExec[0].AST.(*parser.ShowTransactionStatus); ok {
+							crash = false
+						}
+					}
+					if crash {
+						log.Fatalf(session.Ctx(), "no error, but we're in error state: %s", state)
+					}
+				}
+			}
+
+			// Check if we need to auto-commit. If so, we end the transaction now; the
+			// transaction was only supposed to exist for the statement that we just
+			// ran.
+			if autoCommit {
+				if err == nil {
+					txn := txnState.mu.txn
+					if txn == nil {
+						log.Fatalf(session.Ctx(), "implicit txn returned with no error and yet "+
+							"the kv txn is gone. No state transition should have done that. State: %s",
+							txnState.State())
+					}
+
+					// We're were told to autoCommit. The KV txn might already be
+					// committed (planNodes are free to do that when running an implicit
+					// transaction, and some try to do it to take advantage of 1-PC txns).
+					// If it is, then there's nothing to do. If it isn't, then we commit
+					// it here.
+					//
+					// NOTE(andrei): It bothers me some that we're peeking at txn to
+					// figure out whether we committed or not, where SQL could already
+					// know that - individual statements could report this back.
+					if txn.IsAborted() {
+						// TODO(andrei): Until 7881 is fixed.
+						log.Fatalf(session.Ctx(), "#7881: the statement we just ran didn't generate an error "+
+							"but the txn proto is aborted. This should never happen. txn: %+v",
+							txn)
+					}
+
+					if !txn.IsCommitted() {
+						var skipCommit bool
+						if e.cfg.TestingKnobs.BeforeAutoCommit != nil {
+							err = e.cfg.TestingKnobs.BeforeAutoCommit(session.Ctx(), stmtsToExec[0].String())
+							skipCommit = err != nil
+						}
+						if !skipCommit {
+							err = txn.Commit(session.Ctx())
+						}
+						log.Eventf(session.Ctx(), "AutoCommit. err: %v\ntxn: %+v", err, txn.Proto())
+						if err != nil {
+							err = txnState.updateStateAndCleanupOnErr(err, e)
+						}
+					}
+				}
+
+				// After autoCommit, unless we're in RestartWait, we leave the
+				// transaction in NoTxn, regardless of whether we executed the query
+				// successfully or we encountered an error.
+				if txnState.State() != RestartWait {
+					txnState.resetStateAndTxn(NoTxn)
+				}
+			}
+
+			// If we've been told by runTxnAttempt that we should move to Open, do it
+			// now.
+			if err == nil && (txnState.State() == AutoRetry) && transitionToOpen {
+				txnState.SetState(Open)
+			}
+
+			if _, ok := err.(*roachpb.UnhandledRetryableError); ok {
+				// We sent transactional requests, so the TxnCoordSender was supposed to
+				// turn retryable errors into HandledRetryableTxnError.
+				log.Fatalf(session.Ctx(), "unexpected UnhandledRetryableError at the Executor level: %s", err)
+			}
+
+			shouldAutoRetry := txnState.State() == RestartWait && txnCanBeAutoRetried
+			if !shouldAutoRetry {
+				break
+			}
+			txnState.mu.txn.PrepareForRetry(session.Ctx(), err)
+		}
+		// We're now done with auto-retries.
+
 		if err != nil && (log.V(2) || logStatementsExecuteEnabled.Get(&e.cfg.Settings.SV)) {
 			log.Infof(session.Ctx(), "execParsed: error: %v. state: %s", err, txnState.State())
 		}
 
-		// Update the Err field of the last result if the error was coming from
-		// auto commit. The error was generated outside of the txn closure, so it was not
-		// set in any result.
-		if err != nil {
-			if aErr, ok := err.(*client.AutoCommitError); ok {
-				// TODO(andrei): Until #7881 fixed.
-				{
-					if txnState.mu.txn != nil {
-						log.Eventf(session.Ctx(), "executor got AutoCommitError: %s\n"+
-							"txn: %+v\nexecOpt.AutoRetry %t, execOpt.AutoCommit:%t, stmts %+v, remaining %+v",
-							aErr, txnState.mu.txn.Proto(), execOpt.AutoRetry, execOpt.AutoCommit, stmts,
-							remainingStmts)
-					} else {
-						log.Errorf(session.Ctx(), "7881: AutoCommitError on nil txn: %s, "+
-							"txnState %+v, execOpt %+v, stmts %+v, remaining %+v, txn captured before executing batch: %+v",
-							aErr, txnState, execOpt, stmts, remainingStmts, txn)
-						txnState.sp.SetBaggageItem(keyFor7881Sample, "sample me please")
-					}
-				}
-				e.TxnAbortCount.Inc(1)
-				// TODO(andrei): Once 7881 is fixed, this should be
-				// txnState.mu.txn.CleanupOnError().
-				txn.CleanupOnError(session.Ctx(), err)
-			}
-		}
-
-		// Sanity check about not leaving KV txns open on errors.
+		// Sanity check about not leaving KV txns open on errors (other than
+		// retriable errors).
 		if err != nil && txnState.mu.txn != nil && !txnState.mu.txn.IsFinalized() {
 			if _, retryable := err.(*roachpb.HandledRetryableTxnError); !retryable {
 				log.Fatalf(session.Ctx(), "got a non-retryable error but the KV "+
@@ -898,23 +998,15 @@ func (e *Executor) execParsed(
 			}
 		}
 
-		// Now make sense of the state we got into and update txnState.
-		if (txnState.State() == RestartWait || txnState.State() == Aborted) &&
-			txnState.commitSeen {
-			// A COMMIT got an error (retryable or not). Too bad, this txn is toast.
+		if txnState.commitSeen && txnState.State() == Aborted {
+			// A COMMIT got an error (retryable or not); we'll move to state NoTxn.
 			// After we return a result for COMMIT (with the COMMIT pgwire tag), the
 			// user can't send any more commands.
-			e.TxnAbortCount.Inc(1)
-			txn.CleanupOnError(session.Ctx(), err)
 			txnState.resetStateAndTxn(NoTxn)
 		}
 
-		if execOpt.AutoCommit {
-			// If execOpt.AutoCommit was set, then the txn no longer exists at this point.
-			txnState.resetStateAndTxn(NoTxn)
-		}
-
-		// If we're no longer in a transaction, finish the trace.
+		// If we're no longer in a transaction, close the transaction-scoped
+		// resources.
 		if txnState.State() == NoTxn {
 			txnState.finishSQLTxn(session)
 		}
@@ -971,7 +1063,7 @@ func (e *Executor) execParsed(
 		// Figure out what statements to run on the next iteration.
 		if err != nil {
 			return convertToErrWithPGCode(err)
-		} else if execOpt.AutoCommit {
+		} else if autoCommit {
 			stmts = stmts[1:]
 		} else {
 			stmts = remainingStmts
@@ -981,25 +1073,27 @@ func (e *Executor) execParsed(
 	return nil
 }
 
-// runTxnAttempt is used in the closure we pass to txn.Exec(). It will be called
-// possibly multiple times with the same set to statements.
-// It takes in a batch of statements and executes a prefix of them corresponding
-// to one transaction. More technically, it runs them one by one until either:
+// runTxnAttempt takes in a batch of statements and executes a prefix of them
+// corresponding to one transaction. It is called multiple times with the same
+// set of statements when performing auto-retries.
+//
+// More technically, it run statements one by one until either:
 // - a statement returns an error
 // - the end of the the batch is reached
 // - a statement transition the state to a non-"open" one, meaning that we're
 // not in a transaction any more (or we are in a transaction but in an error
 // state, although this also triggers the error condition above).
 //
-// This method may transition the state from AutoRetry to Open if the end of the
-// batch is reached.
-//
 // Args:
 // txnPrefix: set if the start of the batch corresponds to the start of the
 // current transaction. Used to trap nested BEGINs.
 // txnResults: used to push query results.
 //
-// It returns all the statements that were not executed.
+// It returns:
+// remainingStmts: all the statements that were not executed.
+// transitionToOpen: specifies if the caller should move from state AutoRetry to
+// state Open. This will be false if the state is not AutoRetry when this
+// returns.
 func runTxnAttempt(
 	e *Executor,
 	session *Session,
@@ -1010,7 +1104,7 @@ func runTxnAttempt(
 	avoidCachedDescriptors bool,
 	automaticRetryCount int,
 	txnResults ResultsGroup,
-) (StatementList, error) {
+) (remainingStmts StatementList, transitionToOpen bool, _ error) {
 
 	txnState := &session.TxnState
 
@@ -1048,7 +1142,7 @@ func runTxnAttempt(
 			txnPrefix && i == 0, /* firstInTxn */
 			avoidCachedDescriptors, automaticRetryCount, stmtResult,
 		); err != nil {
-			return nil, err
+			return nil, false, err
 		}
 
 		// If the transaction is done, stop executing more statements.
@@ -1062,20 +1156,19 @@ func runTxnAttempt(
 	// statements. We also want future ResultsSentToClient() calls to return
 	// false.
 	if err := txnState.txnResults.Flush(session.Ctx()); err != nil {
-		txnState.updateStateAndCleanupOnErr(err, e)
-		return nil, err
+		err = txnState.updateStateAndCleanupOnErr(err, e)
+		return nil, false, err
 	}
 
 	// If we're still in AutoRetry but we've seen statements that would need to be
 	// retried if we'd later want to retry the transaction, we transition to Open.
 	// Automatic retries are not possible beyond this point.
-	if txnState.State() == AutoRetry && encounteredNonAutoRetryStmt {
-		txnState.SetState(Open)
-	}
+	transitionToOpen = txnState.State() == AutoRetry && encounteredNonAutoRetryStmt
+
 	// Return the statements that weren't executed in the current txn, so they can
 	// be executed in the context of other transactions. The set can be empty if
 	// we executed everything that was passed in.
-	return stmts[i+1:], nil
+	return stmts[i+1:], transitionToOpen, nil
 }
 
 // execSingleStatement executes one statement by dispatching according to the
@@ -1277,14 +1370,14 @@ func (e *Executor) execStmtInAbortedTxn(
 		}
 		if err := parser.ValidateRestartCheckpoint(spName); err != nil {
 			if txnState.State() == RestartWait {
-				txnState.updateStateAndCleanupOnErr(err, e)
+				err = txnState.updateStateAndCleanupOnErr(err, e)
 			}
 			return err
 		}
 		if !txnState.retryIntent {
 			err := fmt.Errorf("SAVEPOINT %s has not been used", parser.RestartSavepointName)
 			if txnState.State() == RestartWait {
-				txnState.updateStateAndCleanupOnErr(err, e)
+				err = txnState.updateStateAndCleanupOnErr(err, e)
 			}
 			return err
 		}
@@ -1322,8 +1415,7 @@ func (e *Executor) execStmtInAbortedTxn(
 			// there seems to be little point in staying in RestartWait (plus,
 			// higher-level code asserts that we're only in RestartWait when returning
 			// retryable errors to the client).
-			txnState.updateStateAndCleanupOnErr(err, e)
-			return err
+			return txnState.updateStateAndCleanupOnErr(err, e)
 		}
 		return sqlbase.NewTransactionAbortedError("" /* customMsg */)
 	}
@@ -1411,7 +1503,7 @@ func (e *Executor) execStmtInOpenTxn(
 			if !txnState.TxnIsOpen() {
 				panic(fmt.Sprintf("unexpected txnState when cleaning up: %v", txnState.State()))
 			}
-			txnState.updateStateAndCleanupOnErr(err, e)
+			err = txnState.updateStateAndCleanupOnErr(err, e)
 
 			if firstInTxn && isBegin(stmt) {
 				// A failed BEGIN statement that was starting a txn doesn't leave the
@@ -1455,6 +1547,7 @@ func (e *Executor) execStmtInOpenTxn(
 				// client.Txn.
 				roachpb.Transaction{},
 			)
+			err = txnState.updateStateAndCleanupOnErr(err, e)
 		}
 	}()
 
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -831,158 +831,23 @@ func (e *Executor) execParsed(
 		if txnState.State() == NoTxn {
 			panic("we failed to initialize a txn")
 		}
-		// Now actually run some statements.
-		var remainingStmts StatementList
-		origState := txnState.State()
 
-		// Track if we are retrying this query, so that we do not double count.
-		automaticRetryCount := 0
-
-		txnClosure := func(ctx context.Context) (transitionToOpen bool, _ error) {
-			defer func() { automaticRetryCount++ }()
-
-			if protoTS != nil {
-				txnState.mu.txn.SetFixedTimestamp(*protoTS)
-			}
-
-			// Some results may have been produced by a previous attempt.
-			txnState.txnResults.Reset(session.Ctx())
-			var err error
-			remainingStmts, transitionToOpen, err = runTxnAttempt(
-				e, session, stmtsToExec, pinfo, origState,
-				!inTxn /* txnPrefix */, avoidCachedDescriptors, automaticRetryCount, txnState.txnResults)
-
-			// TODO(andrei): Until #7881 fixed.
-			if err == nil && txnState.State() == Aborted {
-				doWarn := true
-				if len(stmtsToExec) > 0 {
-					if _, ok := stmtsToExec[0].AST.(*parser.ShowTransactionStatus); ok {
-						doWarn = false
-					}
-				}
-				if doWarn {
-					log.Errorf(ctx,
-						"7881: txnState is Aborted without an error propagating. stmtsToExec: %s, "+
-							"remainingStmts: %s, txnState: %+v", stmtsToExec, remainingStmts, txnState)
-				}
-			}
-
-			return transitionToOpen, err
-		}
-
-		// We now run the prefix of stmtsToExec that belong to the current
-		// transaction via txnClosure. We do this in a loop because we may want to
-		// do automatic retries, if possible. Whether or not we can do auto-retries
-		// depends on the state before we execute statements in the batch.
-		//
-		// TODO(andrei): It's unfortunate that we're keeping track of what the state
-		// was before running the statements. It'd be great if the state in which we
-		// find ourselves after running the statements told us if we can auto-retry.
-		// A way to do that is to introduce another state called AutoRestartWait,
-		// similar to RestartWait. We'd enter this state whenever we're in state
-		// AutoRetry and we get a retriable error.
-		txnCanBeAutoRetried := txnState.State() == AutoRetry
 		var err error
-		for {
-			// transitionToOpen will tell us if we need to transition from AutoRetry
-			// to Open after executing the statements.
-			var transitionToOpen bool
-
-			// Run some statements.
-			transitionToOpen, err = txnClosure(session.Ctx())
-
-			// Sanity checks.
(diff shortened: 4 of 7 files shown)


jordanlewis:
The bugfix in r3 LGTM.

---

Reviewed 6 of 6 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.

---

*[pkg/internal/client/txn.go, line 723 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOJEnaS2fmf8erKc-B:-KtOJEnbSxWlMb_KrDVV:b5jf600) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L723)):*
> ```Go
> // It needed fine control by using TxnExecOptions. Now SQL no longer uses this
> // interface, so it's time to see how it can be simplified. TxnExecOptions can
> // probably go way, and so can AutoCommitError.
> ```

s/way/away/

---

*[pkg/sql/copy.go, line 349 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOJXqp_hBUv4gIIDmj:-KtOJXqp_hBUv4gIIDmk:bcwmojh) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/copy.go#L349)):*
> ```Go
> // StatementTag returns a short string identifying the type of statement.
> func (CopyDataBlock) StatementTag() string { return "" }
> func (CopyDataBlock) String() string       { return "Copy" }
> ```

This node is not really a copy - it's merely a marker struct. It's misleading to call it a Copy - I"m not sure when this will even be stringified, but it'd be more clear to name this CopyDataBlock.

---

*[pkg/sql/executor.go, line 890 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOKSxgZzIv-f70ij9k:-KtOKSxgZzIv-f70ij9l:b-i1qgid) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L890)):*
> ```Go
> 					crash := true
> 					if len(stmtsToExec) > 0 {
> 						if _, ok := stmtsToExec[0].AST.(*parser.ShowTransactionStatus); ok {
> ```

This special case could use a clarifying comment. I think it's saying that "show transaction status" statements are always allowed regardless of the transaction state. I'd find it helpful to see a comment to that effect above this whole stanza

---

*[pkg/sql/executor.go, line 916 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOLBf0zfYepmrd45iA:-KtOLBf1omQ5D-iEdecQ:b4joh8l) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L916)):*
> ```Go
> 					// later is not thread-safe, but the commutativity property of
> 					// transactions assure that reasoning about the situation is
> 					// straightforward.
> ```

Not your comment, but to me this isn't super straightforward - if you understand what's going on here, I think adding 1 or two sentences to explain this reasoning would be useful.

---

*[pkg/sql/executor.go, line 926 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOLeZVgBHiUaeATzU1:-KtOLeZVgBHiUaeATzU2:bqnhatc) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L926)):*
> ```Go
> 					case roachpb.PENDING:
> 						err = txn.Commit(session.Ctx())
> 						log.Eventf(session.Ctx(), "AutoCommit. err: %v\ntxn: %+v", err, txn.Proto())
> ```

nit: won't this make traces hard to read if an event log has a newline in the middle of it? Actually, I'm also surprised that this doesn't cause any of the tracing logic tests to fail. Does that indicate a lack of test coverage for this kind of statement?

---

*[pkg/sql/executor.go, line 982 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOM4OiZWOL8km7J95O:-KtOM4OiZWOL8km7J95P:b-7qfu0o) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L982)):*
> ```Go
> 			// After we return a result for COMMIT (with the COMMIT pgwire tag), the
> 			// user can't send any more commands.
> 			txnState.resetStateAndTxn(NoTxn)
> ```

Is it no longer correct to bump the abort count here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-:-KtONFHDL-L7CbNRB6cf:b-w81m1h)*

andreimatei:

---

*[pkg/internal/client/txn.go, line 723 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOJEnaS2fmf8erKc-B:-KtS_IoCEk_6OZHgLJF7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L723)):*

s/way/away/
</blockquote></details>

Done.

---

*[pkg/sql/copy.go, line 349 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOJXqp_hBUv4gIIDmj:-KtS_AThJZcOsXZ3DjmL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/copy.go#L349)):*

This node is not really a copy - it's merely a marker struct. It's misleading to call it a Copy - I"m not sure when this will even be stringified, but it'd be more clear to name this CopyDataBlock.
</blockquote></details>

Done.

---

*[pkg/sql/executor.go, line 890 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOKSxgZzIv-f70ij9k:-KtS__bxMLmbFtRx7G_Z:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L890)):*

This special case could use a clarifying comment. I think it's saying that "show transaction status" statements are always allowed regardless of the transaction state. I'd find it helpful to see a comment to that effect above this whole stanza
</blockquote></details>

Done.

---

*[pkg/sql/executor.go, line 916 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOLBf0zfYepmrd45iA:-KtSfoY1FTxwrPhV9-s7:b-oxgj7k) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L916)):*

Not your comment, but to me this isn't super straightforward - if you understand what's going on here, I think adding 1 or two sentences to explain this reasoning would be useful.
</blockquote></details>

This comment was FUD. Removed from here and `txn.Exec()`. Thanks for saying something.

---

*[pkg/sql/executor.go, line 926 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOLeZVgBHiUaeATzU1:-KtSh7_qJjuirqgX6L5a:b-j8zuu1) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L926)):*

nit: won't this make traces hard to read if an event log has a newline in the middle of it? Actually, I'm also surprised that this doesn't cause any of the tracing logic tests to fail. Does that indicate a lack of test coverage for this kind of statement?
</blockquote></details>

I don't know about hard to read; I'll leave it alone for the moment.
The one test that I can find that would care about this line hilariously does:
```
# Inspect the trace: we exclude messages containing newlines as these
# may contain non-deterministic txn object descriptions.
# This also checks that the span column properly reports separate
# SQL transactions.
SELECT span, message, operation FROM [SHOW TRACE FOR SESSION] WHERE message NOT LIKE e'%\n%'
```

Anyway, I don't know about a "lack of test coverage" - we have no promises about tracing coverage for tests :)
But as it happens @knz seems very gung-ho about adding more tracing to the state transitions in preparation for further state machine refactoring, so good stuff might happen.

---

*[pkg/sql/executor.go, line 982 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOM4OiZWOL8km7J95O:-KtSsuThmOrzR_PhSUZd:b-9c5axo) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L982)):*

Is it no longer correct to bump the abort count here?
</blockquote></details>

I think that the code before was double-counting non-retriable errors; the increment is done when we transition to the `Aborted` state. And now we also transition to the aborted case on retriable errors that won't be auto-retried (in `updateStateAndCleanupOnErr`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

jordanlewis:
Reviewed 5 of 5 files at r4.
Review status: 3 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/executor.go, line 926 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtOLeZVgBHiUaeATzU1:-KtSvqu5cZahrTYfruS_:b4c9lyr) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/sql/executor.go#L926)):*

I don't know about hard to read; I'll leave it alone for the moment.
The one test that I can find that would care about this line hilariously does:
```
# Inspect the trace: we exclude messages containing newlines as these
# may contain non-deterministic txn object descriptions.
# This also checks that the span column properly reports separate
# SQL transactions.
SELECT span, message, operation FROM [SHOW TRACE FOR SESSION] WHERE message NOT LIKE e'%\n%'
```

Anyway, I don't know about a "lack of test coverage" - we have no promises about tracing coverage for tests :)
But as it happens @knz seems very gung-ho about adding more tracing to the state transitions in preparation for further state machine refactoring, so good stuff might happen.

</blockquote></details>

:thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

nvanbenschoten:
Reviewed 3 of 6 files at r1, 3 of 5 files at r4, 1 of 1 files at r5, 3 of 3 files at r6.

---

*[pkg/internal/client/txn.go, line 163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtStHdYse-__nadL3_0:-KtStHdZPxx8JTHt2Ovz:b-jc22nj) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L163)):*
> ```Go
> }
> 
> // Status returns the txn proto status field.
> ```

Why have `status` and `Status`?

---

*[pkg/internal/client/txn.go, line 788 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtStQq5OW_ZmhNPbWGR:-KtStQq5OW_ZmhNPbWGS:bo5splb) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L788)):*
> ```Go
> 
> // PrepareForRetry needs to be called before an retry to perform some
> // book-keeping.
> ```

What book-keeping?

---

*[pkg/sql/executor.go, line 887 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT4RLK2A2QHTVjR677:-KtT4RLK2A2QHTVjR678:b-8q913x) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L887)):*
> ```Go
> 			} else {
> 				state := txnState.State()
> 				if err == nil && (state == Aborted || state == RestartWait) {
> ```

Will this sanity check work correctly with parallel statement execution? Couldn't a statement effect these concurrently?

---

*[pkg/sql/executor.go, line 948 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT5KOgPPztoqWeySDQ:-KtT5KOhoUOT8PCfz31t:brnx7dn) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L948)):*
> ```Go
> 			// If we've been told by runTxnAttempt that we should move to Open, do it
> 			// now.
> 			if err == nil && (txnState.State() == AutoRetry) && transitionToOpen {
> ```

Is `transitionToOpen == true && txnState.State() != AutoRetry` possible?

---

*[pkg/sql/executor.go, line 1525 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7QYIxxBY7NbnHT1W:-KtT7QYIxxBY7NbnHT1X:b219azt) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L1525)):*
> ```Go
> 				roachpb.Transaction{},
> 			)
> 			txnState.updateStateAndCleanupOnErr(&err, e)
> ```

Why is this now needed but wasn't before?

---

*[pkg/sql/executor.go, line 926 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8N1zw1PvDf0lVms2:-KtT8N1zw1PvDf0lVms3:bsjplea) ([raw file](https://github.com/cockroachdb/cockroach/blob/31edafee49b0f2c9874889ad652395bd75ddabb0/pkg/sql/executor.go#L926)):*
> ```Go
> // transaction. It deals with auto-retries: when possible, the statements are
> // retried in case of retriable errors. It also deals with "autoCommit" - if the
> // current transaction only consists a single statement (i.e. it's an "implicit
> ```

`s/consists a/consists of a/`

---

*[pkg/sql/executor.go, line 986 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT95OIT_Aedg7lyHD1:-KtT95OIT_Aedg7lyHD2:bnwq59q) ([raw file](https://github.com/cockroachdb/cockroach/blob/31edafee49b0f2c9874889ad652395bd75ddabb0/pkg/sql/executor.go#L986)):*
> ```Go
> 
> 		if protoTS != nil {
> 			txnState.mu.txn.SetFixedTimestamp(*protoTS)
> ```

Do we need to lock `txnState`?

---

*[pkg/sql/session.go, line 1194 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT6agpAnVZz6aZrVP9:-KtT6agpAnVZz6aZrVPA:b1yh11f) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L1194)):*
> ```Go
> // and the KV txn is cleaned up.
> //
> // This method takes err by pointer because it might replace it: a retriable
> ```

Just return an error. The following is much more idiomatic:
```
err = functionCall(err)
```

---

*[pkg/sql/txn_restart_test.go, line 1061 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7CCP4PQUOH5AhOj1:-KtT7CCP4PQUOH5AhOj2:bwn2ing) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/txn_restart_test.go#L1061)):*
> ```Go
> 	}
> 	for _, tc := range testCases {
> 		t.Run(fmt.Sprintf("retryIntent:%t", tc.retryIntent), func(t *testing.T) {
> ```

`s/retryIntent:%t/retryIntent=%t/`

---

*[pkg/sql/txn_restart_test.go, line 1784 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9tE1sIM2mY4RTFv-:-KtT9tE1sIM2mY4RTFv0:bpl7gqn) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1784)):*
> Once we're on go 1.9, this test should use the new db.Conn()

I like the sound of that

---

*[pkg/sql/txn_restart_test.go, line 1796 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTA9d_YRU4mJJf1udN:-KtTA9d_YRU4mJJf1udO:b-6x2wnt) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1796)):*
> ```Go
> 			var savepoint string
> 			if tc.clientDirectedRetries {
> 				savepoint = "SAVEPOINT cockroach_restart;"
> ```

`declareSavepoint.SQLCommand()`?

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-KtTANk4_gL84tuHuUy4:bior3v1) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*
> ```Go
> 			}
> 			var expectedState string
> 			if tc.clientDirectedRetries {
> ```

Put these in the test case struct as a `sql.TxnStateEnum` field

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

knz:
Reviewed 5 of 5 files at r4, 1 of 1 files at r5, 3 of 3 files at r6.

---

*[pkg/sql/copy.go, line 349 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT6yvY42txBt8mpFyz:-KtT6yvZ965R5Xf5ykgl:b-1tyqeu) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/copy.go#L349)):*
> ```Go
> // StatementTag returns a short string identifying the type of statement.
> func (CopyDataBlock) StatementTag() string { return "" }
> func (CopyDataBlock) String() string       { return "CopyDataBlock" }
> ```

This does not seem to belong to the right commit.

---

*[pkg/sql/executor.go, line 883 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7NkJviT7uZKPyPL8:-KtT7NkJviT7uZKPyPL9:b94slmd) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L883)):*
> ```Go
> 			// Sanity checks.
> 			if err != nil && txnState.TxnIsOpen() {
> 				log.Fatalf(session.Ctx(), "statement(s) generated error but state is %s. err: %s",
> ```

A statement can error out without the kv txn bailing. For example a SQL logic error. I don't think a `log.Fatalf` here is safe. However it is the case that if `err != nil` and the txn is still open, it must be closed.

---

*[pkg/sql/executor.go, line 890 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7n3yiguWljiMGfBU:-KtT7n3yiguWljiMGfBV:b-gd6s9j) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L890)):*
> ```Go
> 					crash := true
> 					if len(stmtsToExec) > 0 {
> 						// SHOW TRANSACTION STATUS statements are always allowed regardless
> ```

If `err` is nil here and the txn is aborted, I don't think SHOW TRANSACTION STATUS should mask this. It is a strong inconsistency that suggests a bug.

---

*[pkg/sql/executor.go, line 918 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8s5NtfHcxAxfYmUn:-KtT8s5OsgqGdfoSJxh_:b-yshe2q) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L918)):*
> ```Go
> 					// to figure out whether a 1-PC transaction ran under us, where SQL
> 					// could already know that - individual statements decide whether to
> 					// use 1-PC or not, so they could report this back.
> ```

It's not so clear to me. It's not because a statement issues a 1PC batch that the kv txn will effectively be 1PC. For example you could have some schema access in the same txn, invisible from that statement's logic, which causes the KV layer to "upgrade" the txn to 2PC. That's why I think KV knows best. However, we should really have an accessor method `(*Txn).Is1PC()` that does this test instead of here.

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-KtT9z0fKFsBQGstBptE:bve0b05) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*
> ```Go
> 	// statement execution by the infrastructure added to support CancelRequest,
> 	// we should try to actively drain this queue instead of passively waiting
> 	// for it to drain. (andrei, 2017/09) - We now have support for statement
> ```

@nvanbenschoten can you share your thoughts about what needs to be done here?

---

*[pkg/sql/session.go, line 1197 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTA8BObx9yEYH-MVo1:-KtTA8BObx9yEYH-MVo2:bo2tqlc) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L1197)):*
> ```Go
> // error meant for another txn is replaced with a non-retriable error because
> // higher layers are not supposed to consider it retriable.
> func (ts *txnState) updateStateAndCleanupOnErr(err *error, e *Executor) {
> ```

Make it return the new err instead, please.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

andreimatei:

---

*[pkg/internal/client/txn.go, line 163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtStHdYse-__nadL3_0:-KtTCqGOt-9__-tBM36q:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L163)):*

Why have `status` and `Status`?
</blockquote></details>

Done.

---

*[pkg/internal/client/txn.go, line 788 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtStQq5OW_ZmhNPbWGR:-KtTBoUMP0iO0LgAWrxT:bs6rt41) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L788)):*

What book-keeping?
</blockquote></details>

`txn.commitTriggers = nil`

---

*[pkg/sql/copy.go, line 349 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT6yvY42txBt8mpFyz:-KtTS0K5iGAj76fly0q7:b-oqwm5y) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/copy.go#L349)):*

This does not seem to belong to the right commit.
</blockquote></details>

moved to its own commit. But let's not discourage drive-bys too much :)

---

*[pkg/sql/executor.go, line 883 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7NkJviT7uZKPyPL8:-KtTD50HCnfWoRkRA4aG:b-y6ft2a) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L883)):*

A statement can error out without the kv txn bailing. For example a SQL logic error. I don't think a `log.Fatalf` here is safe. However it is the case that if `err != nil` and the txn is still open, it must be closed.
</blockquote></details>

any execution error moves the state to Aborted and rolls-back the KV txn, below this level. There's no error recovery in our language.

---

*[pkg/sql/executor.go, line 887 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT4RLK2A2QHTVjR677:-KtTDaAB3nAhvumYjLcL:blrxho5) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L887)):*

Will this sanity check work correctly with parallel statement execution? Couldn't a statement effect these concurrently?
</blockquote></details>

I think it's correct. Parallel stmt execution does not change the state async. All the state changes go through here.

---

*[pkg/sql/executor.go, line 890 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7n3yiguWljiMGfBU:-KtTDsAb5L93jxxvd7Xh:b9osm5c) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L890)):*

If `err` is nil here and the txn is aborted, I don't think SHOW TRANSACTION STATUS should mask this. It is a strong inconsistency that suggests a bug.
</blockquote></details>

the idea is that, for better or for worse, `SHOW TRANSACTION STATUS` runs when we're in any state, and never generates an error. It is special like that. So this sanity check doesn't apply to it.
There's got to be a better solution about running it outside of the state machine, but I didn't pull in that direction here.

---

*[pkg/sql/executor.go, line 918 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8s5NtfHcxAxfYmUn:-KtTEDdFX0XWHj6WhydE:b-4entkw) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L918)):*

It's not so clear to me. It's not because a statement issues a 1PC batch that the kv txn will effectively be 1PC. For example you could have some schema access in the same txn, invisible from that statement's logic, which causes the KV layer to "upgrade" the txn to 2PC. That's why I think KV knows best. However, we should really have an accessor method `(*Txn).Is1PC()` that does this test instead of here.
</blockquote></details>

1PC is not technically correct, but I think it works for the purpose of this comment. The idea is that, for implicit transactions, we tell plan nodes that they're free to add a `EndTxn` to the batch(es) that they run, and so they might commit the KV transaction from under us. The point of this is to enable 1PC by producing batches that contain both the `BeginTxn` and the `EndTxn`. It indeed doesn't mean that the transaction ran as 1PC (the DistSender can split it if it spans ranges, etc), but one way or another the transaction is committed if the SQL statement put a `EndTxn` in it.

What you're saying about schema changes I don't think makes sense. Schema changes run outside of (KV) transactions; there's no "upgrade" around them. 

---

*[pkg/sql/executor.go, line 948 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT5KOgPPztoqWeySDQ:-KtTJHRQiEWL__uWnHvf:b-8ox2zh) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L948)):*

Is `transitionToOpen == true && txnState.State() != AutoRetry` possible?
</blockquote></details>

yes, because of the transition to `NoTxn` done by autoCommit. The alternative is to pipe the `autoCommit` all the way to `runTxnAttempt()` and have it not return `transitionToOpen` in that case, but I don't like that very much. 

---

*[pkg/sql/executor.go, line 1525 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7QYIxxBY7NbnHT1W:-KtTKt9IJHfQuxrvggYQ:b-48n9lq) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L1525)):*

Why is this now needed but wasn't before?
</blockquote></details>

I think it should have always been there, but it didn't matter very much technically when auto-retries were being done by `txn.Exec()` based on the error type - the state would have been overwritten when we retried anyway. But now auto-retries are driven by state, so it matters.

---

*[pkg/sql/executor.go, line 926 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8N1zw1PvDf0lVms2:-KtTQE9IXNTVIGytlx_7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/31edafee49b0f2c9874889ad652395bd75ddabb0/pkg/sql/executor.go#L926)):*

`s/consists a/consists of a/`
</blockquote></details>

Done.

---

*[pkg/sql/executor.go, line 986 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT95OIT_Aedg7lyHD1:-KtTQM5FzKt6T-b3gYG_:b-wsgu77) ([raw file](https://github.com/cockroachdb/cockroach/blob/31edafee49b0f2c9874889ad652395bd75ddabb0/pkg/sql/executor.go#L986)):*

Do we need to lock `txnState`?
</blockquote></details>

no; reads from the main goroutine don't need synchronization (because this goroutine is the only writer)

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-KtTMTlQ9n67qsSCjyhK:b-iv2xtd) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*

@nvanbenschoten can you share your thoughts about what needs to be done here?
</blockquote></details>

If you rephrase the TODO, I'll copy-paste.

---

*[pkg/sql/session.go, line 1194 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT6agpAnVZz6aZrVP9:-KtTMiCQpnuh-0ybfB6B:bp2ak45) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L1194)):*

Just return an error. The following is much more idiomatic:
```
err = functionCall(err)
```
</blockquote></details>

but... but... ok fine.

---

*[pkg/sql/session.go, line 1197 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTA8BObx9yEYH-MVo1:-KtTMqibXArod2fUTO2W:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L1197)):*

Make it return the new err instead, please.
</blockquote></details>

Done.

---

*[pkg/sql/txn_restart_test.go, line 1061 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7CCP4PQUOH5AhOj1:-KtTP0YBQgwPEaYWmKeV:bah89up) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/txn_restart_test.go#L1061)):*

`s/retryIntent:%t/retryIntent=%t/`
</blockquote></details>

Done.
At least put a "[nit]" in front of it :)

---

*[pkg/sql/txn_restart_test.go, line 1796 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTA9d_YRU4mJJf1udN:-KtTO78BPz5rkb47ujkP:b3xh1zt) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1796)):*

`declareSavepoint.SQLCommand()`?
</blockquote></details>

meh, that constant serves other purposes in other tests. I don't see a need to use it here.

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-KtTOKbgxz6GqElZIoUX:b7obzqk) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

Put these in the test case struct as a `sql.TxnStateEnum` field
</blockquote></details>

I thought about doing that, but I think it would make the test less readable - you'd have to jump back to the tc def to see what you expect to happen (and it's not like you have all the information about the test at the tc definition point such that it'd be useful to see it from the get-go).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

knz:
Reviewed 5 of 5 files at r7, 5 of 5 files at r8, 1 of 1 files at r9, 3 of 3 files at r10.
Review status: all files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

*[pkg/sql/executor.go, line 918 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8s5NtfHcxAxfYmUn:-KtXcyX1LQKFQaOzVm0L:b-9nww1y) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L918)):*

1PC is not technically correct, but I think it works for the purpose of this comment. The idea is that, for implicit transactions, we tell plan nodes that they're free to add a `EndTxn` to the batch(es) that they run, and so they might commit the KV transaction from under us. The point of this is to enable 1PC by producing batches that contain both the `BeginTxn` and the `EndTxn`. It indeed doesn't mean that the transaction ran as 1PC (the DistSender can split it if it spans ranges, etc), but one way or another the transaction is committed if the SQL statement put a `EndTxn` in it.

What you're saying about schema changes I don't think makes sense. Schema changes run outside of (KV) transactions; there's no "upgrade" around them. 
</blockquote></details>

Then call it by the name that describes what you say (I'd vote for "statement-controlled txn") and still create a bool somewhere that gives you this info.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

andreimatei:
Review status: 1 of 7 files reviewed at latest revision, 14 unresolved discussions.

---

*[pkg/sql/executor.go, line 918 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8s5NtfHcxAxfYmUn:-KtZCb3kLDnuaSiAE6IY:b-oq3912) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L918)):*

Then call it by the name that describes what you say (I'd vote for "statement-controlled txn") and still create a bool somewhere that gives you this info.
</blockquote></details>

Done, PTAL

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

knz:
Reviewed 1 of 5 files at r11, 2 of 6 files at r13, 1 of 1 files at r14, 1 of 2 files at r15.
Review status: 5 of 7 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

*[pkg/sql/executor.go, line 918 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT8s5NtfHcxAxfYmUn:-Ktb70qxgRtDGXkUEI4o:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L918)):*

Done, PTAL
</blockquote></details>

:+1:

---

*[pkg/sql/executor.go, line 1068 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-Ktb6SMaPfOC7byreiJk:-Ktb6SMbqN5jmeL4OSW-:b-a9ucq0) ([raw file](https://github.com/cockroachdb/cockroach/blob/dfeec26ffa00673faa4e475e18aae1b3e57594b1/pkg/sql/executor.go#L1068)):*
> ```Go
> 				if !txn.IsCommitted() {
> 					var skipCommit bool
> 					if e.cfg.TestingKnobs.BeforeAutoCommit != nil {
> ```

I think the BeforeAutoCommit knob should only be called here if skipCommit is false.
(caveat: I haven't yet read the source of the tests that use it)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

nvanbenschoten:
Reviewed 1 of 5 files at r4, 3 of 5 files at r11, 1 of 1 files at r12, 5 of 6 files at r13, 1 of 1 files at r14, 1 of 2 files at r15.

---

*[pkg/internal/client/txn.go, line 788 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtStQq5OW_ZmhNPbWGR:-KtkxTqNrZor_O44zlfS:bo0rrdv) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L788)):*

`txn.commitTriggers = nil`
</blockquote></details>

Spell that out in the comment

---

*[pkg/sql/executor.go, line 887 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT4RLK2A2QHTVjR677:-Ktl1Jg5JzqU8p8d7fQu:b7guhpi) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L887)):*

I think it's correct. Parallel stmt execution does not change the state async. All the state changes go through here.
</blockquote></details>

Ah ok, so it's ok to rely on the `txnState.State` only being changed here, but not the `tnx.state`, right?

---

*[pkg/sql/executor.go, line 1217 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtkyvkjQ9fCK8DO_hZf:-KtkyvkjQ9fCK8DO_hZg:b-w41zp4) ([raw file](https://github.com/cockroachdb/cockroach/blob/9873bb7e20486586368858cf36eee5663e5f7c56/pkg/sql/executor.go#L1217)):*
> ```Go
> 	// If implicitTxn is set, the state is supposed to be AutoRetry. However, we
> 	// test the state too, just in case we have a state machine bug, in which case
> 	// we don't want this code to hide a bug.
> ```

Should this be an assertion then? Aren't we still hiding any bugs?

---

*[pkg/sql/executor.go, line 947 at r13](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtkzsX7FV-OQpvKAIFT:-KtkzsX7FV-OQpvKAIFU:b-k12jys) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee6f0d9f81bda7f8730c9401ef339f2d4ba71685/pkg/sql/executor.go#L947)):*
> If an error is returned, skipCommit has to be true

Let's assert that this is used correctly

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-Ktky6UOhxQaMnebgWir:b-l3gvi9) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*

If you rephrase the TODO, I'll copy-paste.
</blockquote></details>

Do we have support for statement cancellation? Last time I checked we still don't have per-statement contexts.

---

*[pkg/sql/txn_restart_test.go, line 1061 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7CCP4PQUOH5AhOj1:-KtkyLuSI6-HZ-neTTsq:b-jbkypi) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/txn_restart_test.go#L1061)):*

Done.
At least put a "[nit]" in front of it :)
</blockquote></details>

@tamird wouldn't call it a nit, so neither will I! As we discussed in person, let's add this to the style guide.

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-Ktl0TQtBQAM7u8v2STr:b7597ok) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

I thought about doing that, but I think it would make the test less readable - you'd have to jump back to the tc def to see what you expect to happen (and it's not like you have all the information about the test at the tc definition point such that it'd be useful to see it from the get-go).
</blockquote></details>

I disagree. The whole point of table-driven testing is that you factor out all differences between test cases so that you can see how each input configuration affects each output result. In terms of the `expectedState` string, this is a clear win.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

andreimatei:

---

*[pkg/internal/client/txn.go, line 788 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtStQq5OW_ZmhNPbWGR:-Ktl7237bVnJK3tPq9-K:b-lmhugn) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570b5abdf24a594cf3228bb1c806290629c27d3/pkg/internal/client/txn.go#L788)):*

Spell that out in the comment
</blockquote></details>

Well first of all I think this `commitTriggers` member is reset in the wrong place currently; I filed a bug about that. So I don't wanna comment to much about it yet.
Second, I don't think a comment about that internal member belongs in the method comment anyway; it's not something all callers should understand.

---

*[pkg/sql/executor.go, line 887 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT4RLK2A2QHTVjR677:-Ktl9gvxHzunY_D7-aIl:b8fr4k8) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/executor.go#L887)):*

Ah ok, so it's ok to rely on the `txnState.State` only being changed here, but not the `tnx.state`, right?
</blockquote></details>

You mean "not `txn.status()`. Yes, that's correct. `txn.status()` can perhaps do some transitions async, but that's not relevant here.

---

*[pkg/sql/executor.go, line 1217 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtkyvkjQ9fCK8DO_hZf:-KtlB5teLb-C9BjWc_b8:b-8wngkn) ([raw file](https://github.com/cockroachdb/cockroach/blob/9873bb7e20486586368858cf36eee5663e5f7c56/pkg/sql/executor.go#L1217)):*

Should this be an assertion then? Aren't we still hiding any bugs?
</blockquote></details>

What this comment is trying to say is that `SHOW TRANSACTIONS STATUS` is supposed to be useful for debugging, so we don't want it itself to panic or have complex logic; it's useful for it to give you the straight dope.

---

*[pkg/sql/executor.go, line 947 at r13](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtkzsX7FV-OQpvKAIFT:-KtlQHUPg-f_aZAXcSw7:byisbca) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee6f0d9f81bda7f8730c9401ef339f2d4ba71685/pkg/sql/executor.go#L947)):*

> If an error is returned, skipCommit has to be true

Let's assert that this is used correctly
</blockquote></details>

I remove `skipCommit` per Rafa's request. He called it "cruft" :).

---

*[pkg/sql/executor.go, line 1068 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-Ktb6SMaPfOC7byreiJk:-KtlQCt23FA8GWvK1hqY:b7jqr51) ([raw file](https://github.com/cockroachdb/cockroach/blob/dfeec26ffa00673faa4e475e18aae1b3e57594b1/pkg/sql/executor.go#L1068)):*

I think the BeforeAutoCommit knob should only be called here if skipCommit is false.
(caveat: I haven't yet read the source of the tests that use it)
</blockquote></details>

removed `skipCommit` as discussed offline

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-KtlEAw9CyJc-Yl9s8UG:bunlqf2) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*

Do we have support for statement cancellation? Last time I checked we still don't have per-statement contexts.
</blockquote></details>

We do have statement cancellation. Indeed, we do not have per-statement contexts - statement cancellation really implemented by transaction cancellation, but I don't think that's relevant here. And in fact here you want the biggest hammer available, so in fact transaction cancellation is more appropriate.

---

*[pkg/sql/txn_restart_test.go, line 1061 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7CCP4PQUOH5AhOj1:-KtlEf0YAsZxcF07G0IL:b4os9rn) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/txn_restart_test.go#L1061)):*

@tamird wouldn't call it a nit, so neither will I! As we discussed in person, let's add this to the style guide.
</blockquote></details>

shoutout  to our shareholder homie!

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-KtlGmprschuNZvYrjft:b-6oulny) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

I disagree. The whole point of table-driven testing is that you factor out all differences between test cases so that you can see how each input configuration affects each output result. In terms of the `expectedState` string, this is a clear win.
</blockquote></details>

I respectfully disagree. I'm pretty sure that I (at least) would find it less readable if it were written that way.
I don't believe that the point of table-driven testing is to "factor out _all_ differences", at least not to the detriment of readability. The point, as I see it, is to reduce some repetition. If you're saying that there should never be any conditionals dependent on the test case in the test body, then I think you're taking it too far. For example, would you also put the line `savepoint = "SAVEPOINT cockroach_restart;` into the test definition?
My principle is that something makes it to the test definition if there's enough context at the point where the tests table is placed to intuit what that member is about and, if it's a test result rather than a test input, you should be able to intuit why is that result expected. In the particular case of `expectedState`, I don't think that the condition is met: the `expectedState` only makes sense once you've seen the test body and understand what the test does.
Another thing to note about `expectedState` is that that's not the main assertion of these tests; it's a somewhat complementary/incidental assertion. The main assertion is common among all sub-tests and is the fact that the batch gets a retriable error. I see this as also adding to the case for not elevating this `expectedState` to a more prominent position.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

knz:
Reviewed 3 of 6 files at r18, 1 of 1 files at r19, 2 of 2 files at r20.

---

*[pkg/sql/executor.go, line 947 at r13](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtkzsX7FV-OQpvKAIFT:-KtqvQALKIhgwGM5Qa9L:b4qaftc) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee6f0d9f81bda7f8730c9401ef339f2d4ba71685/pkg/sql/executor.go#L947)):*

I remove `skipCommit` per Rafa's request. He called it "cruft" :).
</blockquote></details>

Thx!! :beers: 

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-Ktqv_gAseYph44BKFHx:ba0zhj1) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*

We do have statement cancellation. Indeed, we do not have per-statement contexts - statement cancellation really implemented by transaction cancellation, but I don't think that's relevant here. And in fact here you want the biggest hammer available, so in fact transaction cancellation is more appropriate.
</blockquote></details>

So, the comment is still incomplete. If a transaction is being cancelled, we should have all parallel statements cancelled and not wait for them to complete.

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-KtqvvV--VbED8P6OLFw:b8tgar2) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

I respectfully disagree. I'm pretty sure that I (at least) would find it less readable if it were written that way.
I don't believe that the point of table-driven testing is to "factor out _all_ differences", at least not to the detriment of readability. The point, as I see it, is to reduce some repetition. If you're saying that there should never be any conditionals dependent on the test case in the test body, then I think you're taking it too far. For example, would you also put the line `savepoint = "SAVEPOINT cockroach_restart;` into the test definition?
My principle is that something makes it to the test definition if there's enough context at the point where the tests table is placed to intuit what that member is about and, if it's a test result rather than a test input, you should be able to intuit why is that result expected. In the particular case of `expectedState`, I don't think that the condition is met: the `expectedState` only makes sense once you've seen the test body and understand what the test does.
Another thing to note about `expectedState` is that that's not the main assertion of these tests; it's a somewhat complementary/incidental assertion. The main assertion is common among all sub-tests and is the fact that the batch gets a retriable error. I see this as also adding to the case for not elevating this `expectedState` to a more prominent position.


</blockquote></details>

I still disagree. With enough parameterization the body of the test is so short that it will fit in your editor window together with the testData definition above. Your argument doesn't really hold water :)

Feel free to do as you wish, but I'll PR that cleanup afterwards if you don't.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

andreimatei:

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-KtvoNAUbt7k2nuf8kz0:b-bpwl1h) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*

So, the comment is still incomplete. If a transaction is being cancelled, we should have all parallel statements cancelled and not wait for them to complete.
</blockquote></details>

well, I believe we should cancel them *and* wait for them to complete afterwards, otherwise I'm not sure that various cleanups done here can be done concurrently with running the statements. The comment mentions memory monitor problems, maybe there's others too.
Anyway, I think the comment + TODO + note are sufficient until someone wants to actually improve the code :)

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-KtvpD8NSxe4S43Z0NgB:bj1jzav) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

I still disagree. With enough parameterization the body of the test is so short that it will fit in your editor window together with the testData definition above. Your argument doesn't really hold water :)

Feel free to do as you wish, but I'll PR that cleanup afterwards if you don't.
</blockquote></details>

:) if this would be the threshold at which people decide that cleanups are imperative, our codebase would be much cleaner.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

tamird:

---

*[pkg/sql/txn_restart_test.go, line 1061 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT7CCP4PQUOH5AhOj1:-Ktvw7eHxGx6ONOoSCbr:blbqnec) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/txn_restart_test.go#L1061)):*

shoutout  to our shareholder homie!
</blockquote></details>

:wave: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268)*

nvanbenschoten:
:lgtm:

---

---

*[pkg/sql/session.go, line 533 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtT9z0fKFsBQGstBptD:-Ku3Tm8kgVjep4tel4Pa:b-8fx2u7) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ac70cd9839b832e01f402eab4a66a671ee61a6f/pkg/sql/session.go#L533)):*

well, I believe we should cancel them *and* wait for them to complete afterwards, otherwise I'm not sure that various cleanups done here can be done concurrently with running the statements. The comment mentions memory monitor problems, maybe there's others too.
Anyway, I think the comment + TODO + note are sufficient until someone wants to actually improve the code :)
</blockquote></details>

:+1: 

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-Ku3TsXahrCLqahoVHnt:b-dz80c2) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

:) if this would be the threshold at which people decide that cleanups are imperative, our codebase would be much cleaner.
</blockquote></details>

If @knz doesn't, I will. The `savepoint = "SAVEPOINT cockroach_restart` case and the `expectedState` case are not the same thing. `savepoint` is a function of the `clientDirectedRetries` field; the two aren't independent variables. They both represent the same idea and `savepoint` could just as easily be a method on the testCase type (if it wasn't anonymous). `expectedState` on the other hand is an independent variable that you're asserting will correlate in a certain way with the other variables in the test cases. 

So yes, you're right. The point of table-driven testing is not to "factor out _all_ differences". There's more to this than just blind parameterization. The main point it's to factor out the independent input and output variables for whatever is being tested. This allows the test to be **much** more readable because the test cases are declarative and can stand on their own. They allow us to easily see how changing the inputs affects the outputs without even needing to understand the implementation details of the test. If we're just trying to understand desired behavior then, for all we care, `t.Run` is some black-box mapping function from the input state to the output state and we're simply asserting that `tc.output = t.Run(tc.input)`.

Also, if you're worried about not having enough context about what `expectedState` is, name it better. Perhaps `expectedTxnStateAfterRetry`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-:-Ku3b8gHNxQ54G-gHMUQ:bnfp4nl)*

knz:
Go for it, Nathan and I will clean up.

andreimatei:
TFTRs !

---

---

*[pkg/sql/txn_restart_test.go, line 1812 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-KtTANk4_gL84tuHuUy3:-Ku5uIH35Fi-EqV8KL2q:b-i7uqoh) ([raw file](https://github.com/cockroachdb/cockroach/blob/83facbcf9af3ad2c34a5e3f24bcbbc9fdaf793dc/pkg/sql/txn_restart_test.go#L1812)):*

If @knz doesn't, I will. The `savepoint = "SAVEPOINT cockroach_restart` case and the `expectedState` case are not the same thing. `savepoint` is a function of the `clientDirectedRetries` field; the two aren't independent variables. They both represent the same idea and `savepoint` could just as easily be a method on the testCase type (if it wasn't anonymous). `expectedState` on the other hand is an independent variable that you're asserting will correlate in a certain way with the other variables in the test cases. 

So yes, you're right. The point of table-driven testing is not to "factor out _all_ differences". There's more to this than just blind parameterization. The main point it's to factor out the independent input and output variables for whatever is being tested. This allows the test to be **much** more readable because the test cases are declarative and can stand on their own. They allow us to easily see how changing the inputs affects the outputs without even needing to understand the implementation details of the test. If we're just trying to understand desired behavior then, for all we care, `t.Run` is some black-box mapping function from the input state to the output state and we're simply asserting that `tc.output = t.Run(tc.input)`.

Also, if you're worried about not having enough context about what `expectedState` is, name it better. Perhaps `expectedTxnStateAfterRetry`.
</blockquote></details>

I'll let you two race to ~~cleanup~~ obfuscate this :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-:-Ku5ueLpxpltt9L5nxLS:bodi8in)*

andreimatei:
For posterity, I messed up the rebase at some point in the middle of the review. The bug fix that was intended for the final commit inadvertently moved to a previous commit. The way the last commit was intended to look can be seen at an older revision, e.g. r10 - b896d3d 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18268#-:-KuKuniy_qVsuKmETqgb:b-j9e4os)*

PR #18271, opened 2017-09-06
sql: refactor unnecessary bookkeeping of every type in tuples when type checking ANY/ALL operations

richardwu:
In #18094, I introduced an intermediary step where we keep track of all the types in a tuple such that we can verify there exists a comparison function between each element and the left expression.

I forgot to remove this unnecessary bookkeeping after introducing a helper method that type checks and requires every element in a tuple be equivalent to the left.

As pointed out by @knz in #18094 (https://reviewable.io/reviews/cockroachdb/cockroach/18094#-Kt3dEQawKw2PgHNgpGq), I should have reverted this back to the original style: https://github.com/cockroachdb/cockroach/pull/18094/files#diff-f5c8f9ab30e6b17710413b03d2169349L1019.

### DIFF
--- a/pkg/sql/parser/type_check.go
+++ b/pkg/sql/parser/type_check.go
@@ -961,8 +961,7 @@ func typeCheckComparisonOpWithSubOperator(
 	foldedOp, _, _, _, _ := foldComparisonExpr(subOp, nil, nil)
 	ops := CmpOps[foldedOp]
 
-	var cmpTypeLeft Type
-	var cmpTypesRight []Type
+	var cmpTypeLeft, cmpTypeRight Type
 	var leftTyped, rightTyped TypedExpr
 	if array, isConstructor := StripParens(right).(*Array); isConstructor {
 		// If the right expression is an (optionally nested) array constructor, we
@@ -999,7 +998,7 @@ func typeCheckComparisonOpWithSubOperator(
 			break
 		}
 		rightTyped = right.(TypedExpr)
-		cmpTypesRight = append(cmpTypesRight, retType)
+		cmpTypeRight = retType
 
 		// Return early without looking up a CmpOp if the comparison type is TypeNull.
 		if retType == TypeNull {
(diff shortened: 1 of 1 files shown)


justinj:

---

*[pkg/sql/logictest/testdata/logic_test/suboperators, line 261 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-KtO-7oF9opmhKFk1LPc:-KtO-7oF9opmhKFk1LPd:b-1gfwle) ([raw file](https://github.com/cockroachdb/cockroach/blob/f945f5fb40d942ace154b76a2ae3ef2db69a541d/pkg/sql/logictest/testdata/logic_test/suboperators#L261)):*
> ```
> SELECT 1 = ALL(ARRAY['foo', 'bar'])
> 
> query error unsupported comparison operator: <int> = ALL <string\[\]>
> ```

how come this change? these errors seemed more helpful before to me.

---

*[pkg/sql/parser/type_check.go, line 1047 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-KtO0RC6niMqCPJ_KDaa:-KtO0RC6niMqCPJ_KDab:bxq3vu9) ([raw file](https://github.com/cockroachdb/cockroach/blob/12bdae3122a74489513f0c9fedcd51f5b03bf685/pkg/sql/parser/type_check.go#L1047)):*
> ```Go
> 		case TTuple:
> 			if len(rightUnwrapped) == 0 {
> 				// Literal tuple or tuple from subquery contains no elements.
> ```

kind of unfortunate that we can't handle this case :/ (since it has a definite answer as long as there's some overload matching the left side), but I guess there's complications with the typing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271)*

richardwu:
*[pkg/sql/logictest/testdata/logic_test/suboperators, line 261 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-KtO-7oF9opmhKFk1LPc:-KtOKr8P4mFI6ORdH0df:b-de5bp1) ([raw file](https://github.com/cockroachdb/cockroach/blob/f945f5fb40d942ace154b76a2ae3ef2db69a541d/pkg/sql/logictest/testdata/logic_test/suboperators#L261)):*

how come this change? these errors seemed more helpful before to me.
</blockquote></details>

Good point. I naively followed what the previous codepath did with these error messages, but it only handled arrays before whereas now tuples are also supported.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271)*

richardwu:
*[pkg/sql/parser/type_check.go, line 1047 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-KtO0RC6niMqCPJ_KDaa:-KtOMO18vg4lK-dc7YMh:b-uvtc0m) ([raw file](https://github.com/cockroachdb/cockroach/blob/12bdae3122a74489513f0c9fedcd51f5b03bf685/pkg/sql/parser/type_check.go#L1047)):*

kind of unfortunate that we can't handle this case :/ (since it has a definite answer as long as there's some overload matching the left side), but I guess there's complications with the typing.
</blockquote></details>

One option is to move checking for empty tuples to `typeCheckAndRequireTupleElems`, but this would be a different behavior than the other type checking methods. This case in fact only covers literal tuples since subqueries are verified for 1 and only 1 column during `analyzeExpr` (analyze.go). I've updated the comment to reflect that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271)*

knz:
LGTM modulo comment request

---

Reviewed 1 of 2 files at r2.

---

*[pkg/sql/parser/type_check.go, line 1052 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-KtRIwD6LXzevP-QTx0J:-KtRIwD6LXzevP-QTx0K:b-hxzk6w) ([raw file](https://github.com/cockroachdb/cockroach/blob/6d3a82812ec2752a82680cf4003a951d6c5204d6/pkg/sql/parser/type_check.go#L1052)):*
> ```Go
> 				return nil, nil, CmpOp{}, subOpCompError(cmpTypeLeft, rightReturn, subOp, op)
> 			}
> 			cmpTypeRight = rightUnwrapped[0]
> ```

Add a comment here to remind the reader why rightUnwrapped never contains more than 1 element.
Also, it's perhaps worth extending the condition above so that it also serves as an assertion this is true.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-:-KtRJ8VNR26S1AF4imo1:b-ez7edx)*

richardwu:
*[pkg/sql/parser/type_check.go, line 1052 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271#-KtRIwD6LXzevP-QTx0J:-KtRZY92SxU4Bb7MkEWw:b-rk55vn) ([raw file](https://github.com/cockroachdb/cockroach/blob/6d3a82812ec2752a82680cf4003a951d6c5204d6/pkg/sql/parser/type_check.go#L1052)):*

Add a comment here to remind the reader why rightUnwrapped never contains more than 1 element.
Also, it's perhaps worth extending the condition above so that it also serves as an assertion this is true.
</blockquote></details>

Well rightUnwrapped would contain more than 1 element if it's a literal tuple (but all elements will have been type inferred to equivalent types).

I've added a comment to reflect this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18271)*

PR #18278, opened 2017-09-07
server: allow eager idle replica campaigning on the bootstrapped node

### LINKED ISSUE #18246: sql: TestTrace: slow ALTER TABLE ... SPLIT AT
From [a comment](https://github.com/cockroachdb/cockroach/issues/17798#issuecomment-327346140) on #17798:

> I see waiting for replication take ~500ms, creating the database and tables taking <100ms and splitting the table taking >3s. I wonder why the split is so slow. 

In other testing recently, I've seen clusters performing hundreds of splits per second. Seems likely this is some test anomaly, but worth tracking down in case it isn't.

petermattis:
Campaigning of idle replicas is allowed when sufficient time (a Raft
election cycle) has passed since the cluster has started. This
campaigning of an idle replica allows the replica to quickly be
rejuvenated when it is dormant. We also provide special handling for allowing
campaigning of idle replicas immediately if we just bootstrapped a
store. Unfortunately, this handling of bootstrapped stores had a hole:
when bootstrapping a cluster we first create a temporary store to
bootstrap on node 1 and then we create the real store and we were
failing to consider this real store bootstrapped. If we subsequently had
to campaign an idle replica we were forced to wait for the Raft election
timeout (3s) before campaigning. The hole is now plugged by explicitly
calling Store.NotifyBootstrapped() if the node has just performed
cluster bootstrapping.

Reduces the total running time for all tests by ~15% with the savings
coming from ccl/sqlccl (50s), sql (21s) and storage (7s).

Fixes #18246

### DIFF
--- a/pkg/server/node.go
+++ b/pkg/server/node.go
@@ -380,6 +380,14 @@ func (n *Node) start(
 		return err
 	}
 
+	if n.initialBoot {
+		// The cluster was just bootstrapped by this node, explicitly notify the
+		// stores that they were bootstrapped.
+		for _, s := range stores {
+			s.NotifyBootstrapped()
+		}
+	}
+
 	if err := n.startStores(ctx, stores, n.stopper); err != nil {
 		return err
 	}
--- a/pkg/server/node_test.go
+++ b/pkg/server/node_test.go
@@ -155,7 +155,8 @@ func createAndStartTestNode(
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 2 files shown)


petermattis:
Cc @andreimatei

a-robinson (approved):
It'd be nicer to just have the `Node` object track whether `bootstrap` has been called on it than to burden callers with another parameter to `start`, but very nice find!

petermattis:
> It'd be nicer to just have the Node object track whether bootstrap has been called on it than to burden callers with another parameter to start, but very nice find!

Good point. Turns out there was already `node.initialBoot` which tracked exactly that.

PR #18280, opened 2017-09-07
sql: fix two bugs in table LeaseManager

vivekmenezes:


vivekmenezes:
@cockroachdb/release

andreimatei:
The code in the first commit looks good, but I don't see what the bug was. Was it an inconsistency between the `tableNameCache` being populated earlier than the rest of the `tableState`? Consider improving the commit message explaining what the bug was exactly and how it was fixed.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/lease.go, line 686 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSN1zKLo4bxPIiIsHZ:-KtSN1zL0aN7T_Lrjxsq:b-tyk9el) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease.go#L686)):*
> ```Go
> // acquireFromStoreLocked acquires a new lease from the store and inserts it
> // into the active set. t.mu must be locked. If the table descriptor version
> // doesn't exist an error is returned (the descriptor version >= minVersion).
> ```

please take the occasion to document what this method does when called while an acquisition is already in progress - particularly that no new lease necessarily has been put in the `tableState` when this returns.
In fact, this behavior seems very surprising to me given the name of this method. I'd suggest returning an error in the `acquireWait` case (something like `errAgain`), which would mean that the caller needs to re-evaluate the `tableState`'s state and possibly retry this call.
But actually, the more I think about it, the more it seems to me like this is the wrong level to deal with `t.acquireWait()`. It bothers me that `acquireFromStoreLocked()` would return an error indicating to the caller that *the caller's* state has changed; I don't know how to describe the semantics of this method.
So I think it'd be better if the caller kept track of `tableState` changes itself and didn't rely on `acquireFromStoreLocked` to signal them. So basically I'd lift the `t.acquireWait()` call, at which point this method doesn't need to exist any more. I'd also lift the `t.mu.acquiring` signaling out of `acquireNodeLease` - at which point I think that method too can go away.

---

*[pkg/sql/lease.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSWe6l1qAwUXZHOI_h:-KtSWe6l1qAwUXZHOI_i:b-igo1nd) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease.go#L690)):*
> ```Go
> 	// Ensure there is no lease acquisition in progress.
> 	if t.acquireWait() {
> 		// There was a lease acquisition in progress; accept the lease just
> ```

strike the "accept the lease just acquired" - that doesn't mean anything, as far as I can tell.
For your addition to the comment, I'd lift it to a method comment (if this method remains) and I wouldn't talk about dropping a table - I think that'll only confuse us very soon.

---

*[pkg/sql/lease_internal_test.go, line 512 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSW3aspEhSmJB9TbAH:-KtSW3aspEhSmJB9TbAI:bsmhopm) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease_internal_test.go#L512)):*
> ```Go
> }
> 
> func TestLeaseManagerAcquireDuringDrop(t *testing.T) {
> ```

please add a comment to this test explaining what its point is

---

*[pkg/sql/lease_internal_test.go, line 549 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSWE59-CTcnR-DImG5:-KtSWE59-CTcnR-DImG6:bvvttl) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease_internal_test.go#L549)):*
> ```Go
> 		go func() {
> 			defer wg.Done()
> 			table, _, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
> ```

I'm confused: if we use an old timestamp for this `Acquire()` call, older than the DROP, why exactly would any call fail with `table is being dropped`? Shouldn't the `LeaseManager` decide that it needs to return a descriptor below the current version (the one that's somehow marked for dropping)?

Also, given that the timestamp apparently doesn't matter much, I'm wondering how come we're expecting "table is being dropped" errors, but no errors about the table not being found because it has been dropped and there's no recollection of it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-:-KtSK7A0-TLmePSVzMQD:b-39voki)*

vivekmenezes:
Update the commit message to discuss the bug and how it is fixed.

---

---

*[pkg/sql/lease.go, line 686 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSN1zKLo4bxPIiIsHZ:-KtUEQjW8tK7qWUdHrKk:b-vf0l0s) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease.go#L686)):*

please take the occasion to document what this method does when called while an acquisition is already in progress - particularly that no new lease necessarily has been put in the `tableState` when this returns.
In fact, this behavior seems very surprising to me given the name of this method. I'd suggest returning an error in the `acquireWait` case (something like `errAgain`), which would mean that the caller needs to re-evaluate the `tableState`'s state and possibly retry this call.
But actually, the more I think about it, the more it seems to me like this is the wrong level to deal with `t.acquireWait()`. It bothers me that `acquireFromStoreLocked()` would return an error indicating to the caller that *the caller's* state has changed; I don't know how to describe the semantics of this method.
So I think it'd be better if the caller kept track of `tableState` changes itself and didn't rely on `acquireFromStoreLocked` to signal them. So basically I'd lift the `t.acquireWait()` call, at which point this method doesn't need to exist any more. I'd also lift the `t.mu.acquiring` signaling out of `acquireNodeLease` - at which point I think that method too can go away.
</blockquote></details>

I've restructured the code to now not need acquireWait() to return a boolean. I also used acquireFreshestFromStoreLocked() in ensureVersion() because that is what is expected there.

---

*[pkg/sql/lease.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSWe6l1qAwUXZHOI_h:-KtUEvmlgHD1CF8s0ib1:b-6t2dn7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease.go#L690)):*

strike the "accept the lease just acquired" - that doesn't mean anything, as far as I can tell.
For your addition to the comment, I'd lift it to a method comment (if this method remains) and I wouldn't talk about dropping a table - I think that'll only confuse us very soon.
</blockquote></details>

This method has been deleted

---

*[pkg/sql/lease_internal_test.go, line 512 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSW3aspEhSmJB9TbAH:-KtUEtp12iAS2VB9M8Ek:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease_internal_test.go#L512)):*

please add a comment to this test explaining what its point is
</blockquote></details>

Done.

---

*[pkg/sql/lease_internal_test.go, line 549 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSWE59-CTcnR-DImG5:-KtUF7dP0b1rNDRCUWKZ:brvzfwg) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease_internal_test.go#L549)):*

I'm confused: if we use an old timestamp for this `Acquire()` call, older than the DROP, why exactly would any call fail with `table is being dropped`? Shouldn't the `LeaseManager` decide that it needs to return a descriptor below the current version (the one that's somehow marked for dropping)?

Also, given that the timestamp apparently doesn't matter much, I'm wondering how come we're expecting "table is being dropped" errors, but no errors about the table not being found because it has been dropped and there's no recollection of it?
</blockquote></details>

I no longer use a DROP TABLE in this test  and have a simpler test that fails without the fix.

I've created issue https://github.com/cockroachdb/cockroach/issues/18354 to discuss the problem you reported here with reads with a timestamp before a table DROP failing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-:-KtUEJbP-TlfiJ3EGROO:b-4bpgzx)*

vivekmenezes:
TFTR! Ready for another review.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-:-KtUFXuBSioFE4nRWMEX:b-ikvmdr)*

andreimatei:
:lgtm:

---

---

*[pkg/sql/lease.go, line 686 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-KtSN1zKLo4bxPIiIsHZ:-KtXTZFFRMV03hSYcMQI:b-9e0wsk) ([raw file](https://github.com/cockroachdb/cockroach/blob/d594dbaf056a56157ee65fe50f023e1e86239320/pkg/sql/lease.go#L686)):*

I've restructured the code to now not need acquireWait() to return a boolean. I also used acquireFreshestFromStoreLocked() in ensureVersion() because that is what is expected there.
</blockquote></details>

:thumbsup: thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18280#-:-KtXTlG5ZzVFxScIypzd:bnfp4nl)*

Issue #16298, opened 2017-06-03
*: provide better means for users to op-out of auto-reporting


knz:
The issue here is that we already have some reports that users do not like the auto-reporting being opt-out, and find the frequency of once per hour too intrusive.

**The fact that these complaints are emerging is a signal that documentation, tooltips or other such written notices about how to opt-out are insufficient.**

Side constraints:

- having auto-reporting enabled at many user sites is pretty much essential to the success of Cockroach Labs' business, and a major way for us to keep a handle on quality 
- currently it is technically unwise to reduce the reporting frequency when reporting is enabled, because doing so may cause memory usage in each node to grow excessively (due to more data accumulated between reporting events). 

Observation: *it is not **necessary** for auto-reporting to be enabled at **all** user sites by default.* 

We can (and this issue argues we should) provide a choice to those users who care about this.

Some proposals:
- make pre-built binaries have auto-report *enabled* by default, but **change the build rules so that manually built binaries have auto-report disabled by default**. Users that "go for the source" for "more confidence/trust in what they are using" also get the more conservative program behavior.
- (my preference) **publish two sets of pre-built binaries**, with two download buttons under each other on the download page; the bigger one with auto-reporting enabled and the 2nd set "privacy conscious" be just the same build with the different default.

knz:
cc @dt @dianasaur @petermattis

dt:
I mean, given that you can opt-out via a setting or env var, we already "provide a choice", without fragmenting binaries with different features.

I think we can communicate here that this data only used _in aggregate_ e.g. `70% of users are using foreign keys but aren't interleaving`, or `70% of deployments have more than 3 nodes`, etc which lets us prioritize what features to focus on, document, etc.

w.r.t. the frequency... nothing says we need to _report_ stats every time we _clear_ collected stats -- quick-fix here is we just switch back to our previous daily reporting interval, and, yes, we'll just clear and discard stats the other 23h. eventually, we randomize which period we sample for reporting to make what we get more representative, but that can come later.

-1 on fragmenting with different builds for different behavior: flags/settings/env-vars are for tuning runtime behavior -- another dimension to debug on, possible confusion, etc.  That said, non-"release" binaries (e.g. if you just run `make build`) already skip crash reporting but I generally think the binaries we distribute should behave as close as possible to what we build/test/develop day-to-day -- every difference we introduces just add more opportunities for bugs to hide in the discrepancy.

dt:
[...]

PR #18335, opened 2017-09-07
cherrypick-1.0: Redact hostnames in crash reports

bdarnell:
For increased privacy, omit the hostnames that are by default included
in crash reports. Instead, construct a `server_id` tag that combines the
last eight characters of the cluster ID and the node ID.

See #16298 for details.

### DIFF
--- a/pkg/util/log/crash_reporting.go
+++ b/pkg/util/log/crash_reporting.go
@@ -175,6 +175,10 @@ func sendCrashReport(ctx context.Context, r interface{}, depth int) {
 
 	ex := raven.NewException(err, raven.NewStacktrace(depth+1, contextLines, crdbPaths))
 	packet := raven.NewPacket(err.Error(), ex)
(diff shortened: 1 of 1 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18335#-:-KtSacyPTgDX8U95xzx9:bnfp4nl)*

a-robinson (approved):

bdarnell:
I'm removing some of the tests from this commit because they depend on other changes that aren't in `release-1.0`.

bdarnell:
Had to trim this down some more because there's a race in `raven-go` and I don't want to deal with a dependency update to fix it. So now this PR just redacts the hostname without introducing anything to replace it. We'll get the new server id string in 1.1.

PR #18337, opened 2017-09-07
sql/parser: avoid noop ReplaceAll

dt:
ReplaceAll appears to always allocate a new copy, in addition to whatever work it does internally.
We only need to call it if we acutally detect a match.

### DIFF
--- a/pkg/sql/parser/datum.go
+++ b/pkg/sql/parser/datum.go
@@ -1263,13 +1263,15 @@ var (
 func parseTimestampInLocation(s string, loc *time.Location, typ Type) (time.Time, error) {
 	origS := s
 	l := len(s)
-	// HACK: go doesn't handle offsets that are not zero-padded from psql/jdbc.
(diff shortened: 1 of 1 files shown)


knz (approved):

dt:
rebased locally on #18381 to see what this did there:
```
ConvertRecord-8    22.1µs ± 6%    20.5µs ±10%   -7.22%    (p=0.002 n=9+9)

name             old speed      new speed      delta
ConvertRecord-8  5.44MB/s ± 5%  5.86MB/s ± 9%   +7.89%    (p=0.002 n=9+9)

ConvertRecord-8    2.73kB ± 0%    2.09kB ± 0%  -23.29%  (p=0.000 n=10+10)

ConvertRecord-8       119 ± 0%       101 ± 0%  -15.13%  (p=0.000 n=10+10)
```

jordanlewis:
LGTM, thanks @dt!

madelynnblue (approved):

PR #18338, opened 2017-09-07
cherrypick-1.1: storage: invalidate cached lastTerm on snapshots

nvb:
Addresses the current issue in #17524.

I'll open an issue to properly test this, but that might take some time.
Fow now this seems like a clear fix that we should get in sooner rather
than later.

@cockroachdb/release


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18338#-:-KtSmHRFqp6rxEuFEw0w:bnfp4nl)*

PR #18346, opened 2017-09-07
build: Skip docker publish step for beta releases

bdarnell:
Temporarily disable a test step that stopped the docker publish from
happening.

### DIFF
--- a/build/teamcity-publish-artifacts.sh
+++ b/build/teamcity-publish-artifacts.sh
@@ -9,23 +9,27 @@ export BUILDER_HIDE_GOPATH_SRC=1
 
 build/teamcity-publish-s3-binaries.sh "$@"
 
-if [[ "$TC_BUILD_BRANCH" != *alpha* ]] && [ "$TEAMCITY_BUILDCONF_NAME" == 'Publish Releases' ]; then
-	image=docker.io/cockroachdb/cockroach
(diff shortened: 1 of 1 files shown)


build/teamcity-publish-artifacts.sh
```diff
@@ -22,8 +25,9 @@ if [[ "$TC_BUILD_BRANCH" != *alpha* ]] && [ "$TEAMCITY_BUILDCONF_NAME" == 'Publi
 	    ;;
 	esac
 
-	build/builder.sh make TYPE=$TYPE testbuild TAGS=acceptance PKG=./pkg/acceptance
-	(cd pkg/acceptance && ./acceptance.test -i $image -b /cockroach/cockroach -nodes 4 -test.v -test.timeout -5m)
+        # TODO(bdarnell): This has broken with recent changes to the acceptance tests.
```
benesch: Extraneous space?

benesch (approved):

build/teamcity-publish-artifacts.sh
bdarnell: Hmm, this file is indented with a mix of tabs and spaces (and my editor only interprets tabs as 2 spaces in go-mode; in other languages it uses python rules in which tabs are always 8 spaces). Untabified the whole file.

PR #18262, opened 2017-09-06
ui: lint config files

benesch:
These JS config files were causing Codacy failures, but weren't being
linted by our TeamCity. Enable linting of our config files (and fix the
lint errors within).

PR #18349, opened 2017-09-07
ui: remove generated files from version control

### LINKED ISSUE #18735: ui: reduce impact of embedded.go bottleneck
I'm personally a bit confused about the scope of this issue, but since the Airtable record was becoming a discussion about implementation anyway, I wanted to open this up, assign @BramGruneir and @benesch to it, and move the discussion here.

Since we are actively going to start making some bigger changes to the admin UI starting in around a month, I'm assuming it would be nice to have this in by then.

cc @cuongdo

benesch:
Ignore the first commit; that's #18262.

From the last commit message:

```
Checking in src/js/protos.js and embedded.go causes several major
problems:

  * Every change to the UI, even if the diff is only one line,
    entirely rewrites embedded.go, which is upwards of 12MB in size.
    This adds a lot of weight to the Git repository.

  * Every commit that touches the UI or protobuf definitions must be
    serialized, as diffs in protos.js and embedded.go cannot be
    automatically merged.

  * Including embedded.go in a pull request breaks GitHub's "request
    review" feature.

Removing these files from version control solves all these problems with
only two downsides:

  * Every developer will need Node and Yarn installed. This seems
    reasonable to expect of our engineers and external contributors.
    This commit teaches our source archives to include a pre-generated
    embedded.go so that end users who simply want to build from source
    without Node and Yarn can do so.

  * Developers will likely need to rebuild the UI a few times per day
    as they switch between branches that span UI changes. This commit
    brings the UI compilation time down to about 5s when only
    first-party sources have been updated and 30s when UI dependencies
    are bumped (infrequently), so this also seems reasonable.
```

Closes #18349.
Closes #18735.


bdarnell:
This approach SGTM but don't forget to update CONTRIBUTING.md (and anywhere else we list build requirements).

benesch:
@bdarnell good point, done.

BramGruneir:
We should get someone who develops on linux to ensure this works.  @knz or perhaps @m-schneider.  Can one of you see if this works?

---

Reviewed 1 of 6 files at r2, 14 of 18 files at r4, 5 of 6 files at r5, 1 of 1 files at r6.
Review status: 14 of 20 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[CONTRIBUTING.md, line 26 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXqbcbwYOoKAa1UiY9:-KtXqbcbwYOoKAa1UiYA:bm06gal) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/CONTRIBUTING.md#L26)):*
> ```Markdown
>    - CMake 3.1+
>    - Autoconf 2.68+
>    - NodeJS 6+ and Yarn 0.22.0+.
> ```

Might as well bump yarn to 1.0 +

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXpSIhTHb1h4yZ8jjm:b41x4zf) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*
> ```Makefile
> 	yarn run -- tslint -c tslint.json -p tsconfig.json --type-check
> 	@# TODO(benesch): Invoke tslint just once when palantir/tslint#2827 is fixed.
> 	yarn run -- tslint -c tslint.json *.js
> ```

For all of these run commands, I'm getting a yarn warning that the `--` is being depreciated.  Perhaps update your version of yarn?

Can we check at the start of make that `node --version > 6` and `yarn --version` > 1.0 ?  This would make things clearer for those you don't RTFM.

---

*[pkg/ui/tslint.json, line 119 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXrcP7EeGNnpq8KuPq:-KtXrcP7EeGNnpq8KuPr:btvpp8o) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/tslint.json#L119)):*
> ```JSON
>     ]
>   },
>   "jsRules": {
> ```

Does our JS pass with these rules?  Do these rules matter, since all of our JS is generated?

---

*[pkg/ui/webpack.vendor.js, line 5 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXrBbRkRpBHL1fIWCR:-KtXrBbRkRpBHL1fIWCS:b-6tcvx2) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/webpack.vendor.js#L5)):*
> ```JavaScript
> const path = require("path");
> const webpack = require("webpack");
> ```

nit: extra blank line

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-KtXrk7mFQ1sekiSJW5X:bdv9wjl)*

benesch:
Review status: 14 of 20 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[CONTRIBUTING.md, line 26 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXqbcbwYOoKAa1UiY9:-KtXsuZA8wPm-FqiDQ40:b-n10pll) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/CONTRIBUTING.md#L26)):*

Might as well bump yarn to 1.0 +
</blockquote></details>

Hmm, ok. Done.

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXtLPW0PJb0waerG4_:bmu1him) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

For all of these run commands, I'm getting a yarn warning that the `--` is being depreciated.  Perhaps update your version of yarn?

Can we check at the start of make that `node --version > 6` and `yarn --version` > 1.0 ?  This would make things clearer for those you don't RTFM.
</blockquote></details>

```
time yarn --version
1.0.1

real	0m0.569s
```

We definitely don't want to check at the beginning of Make, as it'll slow down no-op compiles by .5s. Let me see if I can come up with something else.

---

*[pkg/ui/tslint.json, line 119 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXrcP7EeGNnpq8KuPq:-KtXtYgIa-hsnRQrw_GJ:b8nqwpn) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/tslint.json#L119)):*

Does our JS pass with these rules?  Do these rules matter, since all of our JS is generated?
</blockquote></details>

Heh, not all of our JS is generated. Specifically, not our JS config files. P.S. this commit is in a standalone PR that's ready to merge after review: https://github.com/cockroachdb/cockroach/pull/18262

---

*[pkg/ui/webpack.vendor.js, line 5 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXrBbRkRpBHL1fIWCR:-KtXtp2HclnzxbtZVtAn:b-q30swn) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/webpack.vendor.js#L5)):*

nit: extra blank line
</blockquote></details>

Intentional separation of first-party and third-party imports, though I guess I'm not sure we actually settled on that as our style moving forward.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

BramGruneir:
*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXubZpPstjCLD6vT0C:b30b0rx) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

```
time yarn --version
1.0.1

real	0m0.569s
```

We definitely don't want to check at the beginning of Make, as it'll slow down no-op compiles by .5s. Let me see if I can come up with something else.
</blockquote></details>

why would checking a version take so bloody long?  That's nuts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

BramGruneir:
*[pkg/ui/webpack.vendor.js, line 5 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXrBbRkRpBHL1fIWCR:-KtXuiR2gxMQ1ua85wgG:b-95nijw) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/webpack.vendor.js#L5)):*

Intentional separation of first-party and third-party imports, though I guess I'm not sure we actually settled on that as our style moving forward.
</blockquote></details>

Ah, I didn't notice.  carry on!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

bdarnell:
Review status: 14 of 20 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXtNUXKu57qpZzTyEq:bhdg7ao) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

why would checking a version take so bloody long?  That's nuts.
</blockquote></details>

Yarn 1.0 is 3 days old. It's reasonable to require non-JS developers to have node installed, but if this is going to require frequent environment updates then we're going to have to rethink this change. Don't increase minimum versions of tools that we don't vendor just to encourage people to keep up to date; only do it when there is a specific reason. (We don't mandate specific versions of gcc/clang, for example, just something that supports c++11)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

BramGruneir:
Reviewed 1 of 18 files at r4, 6 of 6 files at r7, 6 of 6 files at r8, 1 of 1 files at r9.

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXvykFDy5aVBNPfBLH:b-e1g45e) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

Yarn 1.0 is 3 days old. It's reasonable to require non-JS developers to have node installed, but if this is going to require frequent environment updates then we're going to have to rethink this change. Don't increase minimum versions of tools that we don't vendor just to encourage people to keep up to date; only do it when there is a specific reason. (We don't mandate specific versions of gcc/clang, for example, just something that supports c++11)
</blockquote></details>

I had no idea it was so new.  And I agree entirely, but I think moving to a 1.0 release over a 0.2 makes sense.  To be fair, the only reason I even noticed this was because of the warnings about the change in the run command semantics.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

benesch:
Review status: 0 of 19 files reviewed at latest revision, 4 unresolved discussions.

---

*[CONTRIBUTING.md, line 26 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXqbcbwYOoKAa1UiY9:-KtXwauhpoIlnEI396jK:b-unn24s) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/CONTRIBUTING.md#L26)):*

Hmm, ok. Done.
</blockquote></details>

Actually, this will require rebuilding the builder image, which I'm really not looking forward to doing. If you don't mind, I'll leave this as allowing 0.22 for the moment, and save the bump to 1.0 later.

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXwoVRnMRpCyT_Bioe:b-gbuvp4) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

I had no idea it was so new.  And I agree entirely, but I think moving to a 1.0 release over a 0.2 makes sense.  To be fair, the only reason I even noticed this was because of the warnings about the change in the run command semantics.
</blockquote></details>

Yeah, `yarn run -- COMMAND` is deprecated in favor of `yarn run COMMAND` in 1.0. I think we can live with the warnings for now. `yarn run` is kind of useless anyway, given how slow it is; in another PR I can replace every use with `node_modules/.bin/COMMAND` for a nice speed boost.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

benesch:
Also, I should mention, I added a commit in here to enforce Yarn/Node versions. The restrictions are loose right now (just what CONTRIBUTING.md dictates), but this gives us a path forward when we need to mandate a version upgrade.

---

Review status: 0 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-KtXyDCzvPxorwbqSTo9:b-m5le7a)*

BramGruneir:
Reviewed 1 of 18 files at r4, 19 of 20 files at r10.
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[CONTRIBUTING.md, line 26 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXqbcbwYOoKAa1UiY9:-KtXy6kr6nl8UhCaYPYu:bfv3195) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/CONTRIBUTING.md#L26)):*

Actually, this will require rebuilding the builder image, which I'm really not looking forward to doing. If you don't mind, I'll leave this as allowing 0.22 for the moment, and save the bump to 1.0 later.
</blockquote></details>

Gotcha.  Ok, leave it as is.

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtXy9dvlrh2XZLqfAe7:b-p170ev) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

Yeah, `yarn run -- COMMAND` is deprecated in favor of `yarn run COMMAND` in 1.0. I think we can live with the warnings for now. `yarn run` is kind of useless anyway, given how slow it is; in another PR I can replace every use with `node_modules/.bin/COMMAND` for a nice speed boost.
</blockquote></details>

SGTM.  Perhaps open an issue for this and the yarn bump?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

knz:
I'm getting the following error over here: 
```
installing jsdoc@^3.4.2
installing minimist@^1.2.0
installing espree@^3.4.1
child_process.js:634
    throw err;
    ^

Error: Command failed: npm --silent install jsdoc@^3.4.2 minimist@^1.2.0 espree@^3.4.1
    at checkExecSyncError (child_process.js:591:13)
    at Object.execSync (child_process.js:631:13)
    at modInstall (/data/home/kena/src/go/src/github.com/cockroachdb/cockroach/pkg/ui/node_modules/protobufjs/cli/util.js:129:19)
    at Object.exports.setup (/data/home/kena/src/go/src/github.com/cockroachdb/cockroach/pkg/ui/node_modules/protobufjs/cli/util.js:156:5)
    at Object.<anonymous> (/data/home/kena/src/go/src/github.com/cockroachdb/cockroach/pkg/ui/node_modules/protobufjs/cli/pbjs.js:7:6)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
```

BramGruneir:
:lgtm:

Once someone with linux tests this, I'll feel a bit better.

---

Reviewed 3 of 18 files at r13, 1 of 6 files at r14, 1 of 1 files at r15.
Review status: 5 of 19 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-KtXz28HXWtQ0k4aIiqd:b-p1p0tj)*

a-robinson:
It's awesome to see this, @benesch!

I'd also point out that developers who rely on the builder image won't need node and yarn installed.

knz:
The error above is due to a missing `npm`. Please document that `npm` is required in the contributor guide. LGTM otherwise (FreeBSD).

knz:
Oh and another data point: the newest yarn package I can get my hands on right now is 0.27. I am not keen on seeing the requirement bumped to 1.0 until my package manager has it.

benesch:
@knz, I believe I've eliminated the dependency on NPM. If you have a sec, could you double-check?

---

Review status: 3 of 20 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/ui/Makefile, line 38 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KtXpSIgMPaSYqgMJ_ov:-KtYBiIeH4XUEpCttg32:b-yr2ban) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a6bff674b85a2ff8bf689f0df11fbc659905a36/pkg/ui/Makefile#L38)):*

SGTM.  Perhaps open an issue for this and the yarn bump?
</blockquote></details>

Done. #18378 and #18379

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-KtYC9tYOLxgNZ78bGbG:bjhaf7p)*

knz:
yep, npm not needed any more.

mrtracy:
Reviewed 3 of 18 files at r19.
Review status: 3 of 20 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[Makefile, line 163 at r19](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuKXlEU95pCFX9Djxz_:-KuKXlEU95pCFX9Djxza:b-9775b7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d299852a8a196a5c23323c34fa2f6726138deff3/Makefile#L163)):*
> ```Makefile
> 
> # The build.utcTime format must remain in sync with TimeFormat in pkg/build/info.go.
> $(COCKROACH) build buildoss go-install check test testshort testrace bench: override LINKFLAGS += \
> ```

Assuming this was just some sort of overdue cleanup to change these to gotestdashi?

---

*[pkg/ui/webpack.app.js, line 92 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuKZqZteIyktrhNswRF:-KuKZqZteIyktrhNswRG:blpwaxl) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.app.js#L92)):*
> ```JavaScript
> 
>   plugins: [
>     // new ForkTsCheckerWebpackPlugin({ checkSyntacticErrors: true }),
> ```

Should this be commented out, or just removed entirely?  I think this is only needed if we use thread-loader, and you've removed thread-loader here.

---

*[pkg/ui/webpack.app.js, line 115 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuKaFsF1UsvZ8zh_IBq:-KuKaFsF1UsvZ8zh_IBr:bhp86p0) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.app.js#L115)):*
> ```JavaScript
>       appMountId: "react-layout",
>     }),
>     new webpack.DllReferencePlugin({
> ```

This optimization is very helpful, but also a bit savage; this is no longer straightforward to understand, especially with the magic invocation you've added to the makefile to generate vendor and vendor-manifest.  If you could add a comment either here, or in the makefile when vendor/vendor-manifest are  built, or maybe both, explaining this process... that might be helpful to anyone else tooling around in here in the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

couchand:
Good work, sir.  This is great.

It seems like `embedded.go` should be removed entirely, right?

---

Review status: 3 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/ui/package.json, line 19 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL-qKmDsN3tTLf5Lex:-KuL-qKmDsN3tTLf5Ley:b-b1vyuj) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/package.json#L19)):*
> ```JSON
>     "react": "^15.4.2",
>     "react-addons-create-fragment": "^15.4.2",
>     "react-addons-test-utils": "^15.4.2",
> ```

Should this be in `devDependencies`?

---

*[pkg/ui/package.json, line 28 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL-wfPLIqkXNFf3SjP:-KuL-wfPLIqkXNFf3SjQ:b-rqxvo7) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/package.json#L28)):*
> ```JSON
>     "react-select": "^1.0.0-rc.4",
>     "react-sticky": "^5.0.5",
>     "react-test-renderer": "^15.5.4",
> ```

ditto.

---

*[pkg/ui/webpack.vendor.js, line 13 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL1Q47dD0flmHPiAxW:-KuL1Q47dD0flmHPiAxX:b-7hrqj3) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L13)):*
> ```JavaScript
>   entry: {
>     vendor: [
>       "./src/js/protos",
> ```

I'm wondering if it might be worth the overhead of having three bundles, one that's just the protos?  It seems like those will be changing much more frequently than our dependencies, it's not obvious that protos should live in `vendor`...

---

*[pkg/ui/webpack.vendor.js, line 36 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2BX8scYZwx7GsSwt:-KuL2BX8scYZwx7GsSwu:b-c2vvpi) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L36)):*
> ```JavaScript
>         // the specific packages in node_modules that actually use ES6.
>         // https://github.com/babel/babel-loader/issues/171
>         exclude: /node_modules\/(?!analytics-node)/,
> ```

(I understand this isn't your change, but this is the first I've seen of this, so...) It's not at all clear that we should be using this on the client side.  It would seem that this library is intended for server side use, and we should instead be bundling their `analytics.js` file.

---

*[pkg/ui/webpack.vendor.js, line 37 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2ySdygOk-xUiRzOd:-KuL2ySeNbmwxJ7iSQxF:b-dz9h1t) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L37)):*
> ```JavaScript
>         // https://github.com/babel/babel-loader/issues/171
>         exclude: /node_modules\/(?!analytics-node)/,
>         use: ["cache-loader", "thread-loader", "babel-loader"],
> ```

Are we using `thread-loader` at all?  I didn't think `babel-loader` could run in it...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-KuL-bEgiP5unSd0G0eq:b-54puyg)*

mrtracy:
Review status: 3 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/ui/webpack.vendor.js, line 36 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2BX8scYZwx7GsSwt:-KuL3c5LF-4y3dt4pKv1:b-sfeyye) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L36)):*

(I understand this isn't your change, but this is the first I've seen of this, so...) It's not at all clear that we should be using this on the client side.  It would seem that this library is intended for server side use, and we should instead be bundling their `analytics.js` file.
</blockquote></details>

Analytics.js is designed for a website you are hosting, not for a distributable client. It does several automatic things we don't want; primarily, collecting the URL on all requests - We can't collect the URL automatically, because we don't own the server (its a client application) and thus the collected URL would likely contain PII.  Using analytics-node lets us control exactly what is collected; it also fits into the build system better, as analytics.js is implemented as an embeddable snippet.

AFAIK, segment is aware that many people are using analytics-node for client-side code in this fashion and they will not drop support lightly: https://github.com/segmentio/analytics-node/issues/91#issuecomment-319812988

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

mrtracy:
Review status: 3 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/ui/webpack.vendor.js, line 36 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2BX8scYZwx7GsSwt:-KuL5VSJAValYXhc6Rtu:bvg4qvk) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L36)):*

Analytics.js is designed for a website you are hosting, not for a distributable client. It does several automatic things we don't want; primarily, collecting the URL on all requests - We can't collect the URL automatically, because we don't own the server (its a client application) and thus the collected URL would likely contain PII.  Using analytics-node lets us control exactly what is collected; it also fits into the build system better, as analytics.js is implemented as an embeddable snippet.

AFAIK, segment is aware that many people are using analytics-node for client-side code in this fashion and they will not drop support lightly: https://github.com/segmentio/analytics-node/issues/91#issuecomment-319812988
</blockquote></details>

Used some weird language there, and reviewable doesn't seem to support editing; what I meant by "it's a client application" is, the Admin UI is being hosted directly from the client's deployment of CRDB, so the URL would be for their server, not any server owned by us.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

couchand:
Reviewed 1 of 18 files at r4, 1 of 18 files at r13, 14 of 18 files at r19, 4 of 6 files at r20, 1 of 1 files at r21, 2 of 2 files at r22.

---

*[pkg/ui/webpack.vendor.js, line 36 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2BX8scYZwx7GsSwt:-KuPe0Bng5xtzUdxi1dE:baueeu7) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L36)):*

Used some weird language there, and reviewable doesn't seem to support editing; what I meant by "it's a client application" is, the Admin UI is being hosted directly from the client's deployment of CRDB, so the URL would be for their server, not any server owned by us.
</blockquote></details>

I'm going to disagree here for several reasons:

- There are two ways to change the default URL that is collected - setting the page's canonical URL (https://segment.com/docs/sources/website/analytics.js/#page) or explicitly overriding it, which can also be used for other defaults (https://segment.com/docs/sources/website/analytics.js/#default-properties).
- It's not support that I'm thinking about (although the official recommendation for browser usage does seem to be analytics.js), but rather the additional dependencies and browser shims that this will include in our binary.
- analytics.js can indeed be installed as an NPM module (https://www.npmjs.com/package/analytics.js), but
- I'm not sure we even want to install this library as a regular dependency.  It bloats our binary with something that is available better from a CDN, since it might already be cached.

All that being said, if this set up and working already, it's probably not worth going back and changing now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

benesch:
Yes, `embedded.go` should be gone! Is it not?

Anyway, this is ready for re-review! Sorry for the high latency.

---

Review status: 0 of 21 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[Makefile, line 163 at r19](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuKXlEU95pCFX9Djxz_:-Kuz3TVb-7kMHuTPbltD:bd29v4q) ([raw file](https://github.com/cockroachdb/cockroach/blob/d299852a8a196a5c23323c34fa2f6726138deff3/Makefile#L163)):*

Assuming this was just some sort of overdue cleanup to change these to gotestdashi?
</blockquote></details>

Yessir. It's in its own commit with rationale: https://github.com/cockroachdb/cockroach/pull/18349/commits/c8d3aa3721eac0de748aaa995292a256ba4bc224.

---

*[pkg/ui/package.json, line 19 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL-qKmDsN3tTLf5Lex:-KuzR_obCq4_8zupQjtA:b-yasr8y) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/package.json#L19)):*

Should this be in `devDependencies`?
</blockquote></details>

Yup, thanks. Done.

---

*[pkg/ui/package.json, line 28 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL-wfPLIqkXNFf3SjP:-KuzRaaxpTiMFs7sMqrH:b5t6jp4) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/package.json#L28)):*

ditto.
</blockquote></details>

Also done.

---

*[pkg/ui/webpack.app.js, line 92 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuKZqZteIyktrhNswRF:-Kuz3j5DxcKo4Zgh93s0:b-onwwsl) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.app.js#L92)):*

Should this be commented out, or just removed entirely?  I think this is only needed if we use thread-loader, and you've removed thread-loader here.
</blockquote></details>

Yup, good catch. Removed.

---

*[pkg/ui/webpack.app.js, line 115 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuKaFsF1UsvZ8zh_IBq:-Kuz3mHHB32XePcuddmL:b-29hyz9) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.app.js#L115)):*

This optimization is very helpful, but also a bit savage; this is no longer straightforward to understand, especially with the magic invocation you've added to the makefile to generate vendor and vendor-manifest.  If you could add a comment either here, or in the makefile when vendor/vendor-manifest are  built, or maybe both, explaining this process... that might be helpful to anyone else tooling around in here in the future.
</blockquote></details>

Added a note to the README, plus two cross-references.

---

*[pkg/ui/webpack.vendor.js, line 13 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL1Q47dD0flmHPiAxW:-Kv-EM0eO9zG_F86lzx2:b-qu6n94) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L13)):*

I'm wondering if it might be worth the overhead of having three bundles, one that's just the protos?  It seems like those will be changing much more frequently than our dependencies, it's not obvious that protos should live in `vendor`...
</blockquote></details>

Good point. Done.

---

*[pkg/ui/webpack.vendor.js, line 36 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2BX8scYZwx7GsSwt:-Kuz3vsmczHU34tffPRY:bfnmxyk) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L36)):*

I'm going to disagree here for several reasons:

- There are two ways to change the default URL that is collected - setting the page's canonical URL (https://segment.com/docs/sources/website/analytics.js/#page) or explicitly overriding it, which can also be used for other defaults (https://segment.com/docs/sources/website/analytics.js/#default-properties).
- It's not support that I'm thinking about (although the official recommendation for browser usage does seem to be analytics.js), but rather the additional dependencies and browser shims that this will include in our binary.
- analytics.js can indeed be installed as an NPM module (https://www.npmjs.com/package/analytics.js), but
- I'm not sure we even want to install this library as a regular dependency.  It bloats our binary with something that is available better from a CDN, since it might already be cached.

All that being said, if this set up and working already, it's probably not worth going back and changing now.
</blockquote></details>

Ack. Going to leave as is now and let you and @mrtracy sort it out. :)

---

*[pkg/ui/webpack.vendor.js, line 37 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2ySdygOk-xUiRzOd:-Kv-ENRN_9DsyVPbCzgx:b-s31slc) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L37)):*

Are we using `thread-loader` at all?  I didn't think `babel-loader` could run in it...
</blockquote></details>

¯\_(ツ)_/¯  seems to work just fine! Also seems to halve the build time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-Kv-EwQi_2XdWyWj8p1Y:b-u6b3jm)*

couchand:
Indeed `embedded.go` is gone, it was just Reviewable being weird.

Three minor things, and I think we want to wait to merge until a few weeks after the 1.1 release, but otherwise LGTM.  Thank you so much for your work on this!

---

Reviewed 1 of 18 files at r4, 1 of 18 files at r13, 1 of 18 files at r19, 10 of 20 files at r26, 2 of 6 files at r27, 1 of 1 files at r28, 2 of 2 files at r29.
Review status: 14 of 21 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/ui/webpack.protos.js, line 14 at r29](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-Kv2_Oeo6CXdiK1yn4rt:-Kv2_Oeo6CXdiK1yn4ru:b-39oh8c) ([raw file](https://github.com/cockroachdb/cockroach/blob/16eca953948b0ea50d98aa02b070b0766792830a/pkg/ui/webpack.protos.js#L14)):*
> ```JavaScript
>   output: {
>     filename: "protos.dll.js",
>     path: __dirname + "/dist",
> ```

maybe `path.resolve(__dirname, "dist")`?

---

*[pkg/ui/webpack.vendor.js, line 36 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2BX8scYZwx7GsSwt:-Kv2_bnadWEM9QEb1rZf:bv37kww) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L36)):*

Ack. Going to leave as is now and let you and @mrtracy sort it out. :)
</blockquote></details>

Out of scope for this PR :)

---

*[pkg/ui/webpack.vendor.js, line 37 at r22](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-KuL2ySdygOk-xUiRzOd:-Kv2YY4IGKa6Mmw8ZBcL:b-83ta3j) ([raw file](https://github.com/cockroachdb/cockroach/blob/76fbd79b3510bfea7c937af3711a4c5ba2231d5d/pkg/ui/webpack.vendor.js#L37)):*

¯\_(ツ)_/¯  seems to work just fine! Also seems to halve the build time.
</blockquote></details>

OK np

---

*[pkg/ui/webpack.vendor.js, line 14 at r29](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-Kv2_DoybeLpxDDdVr_1:-Kv2_DoybeLpxDDdVr_2:bsmnsy5) ([raw file](https://github.com/cockroachdb/cockroach/blob/16eca953948b0ea50d98aa02b070b0766792830a/pkg/ui/webpack.vendor.js#L14)):*
> ```JavaScript
>   entry: {
>     vendor: [
>       "./src/js/protos",
> ```

I think you want to delete this line.

---

*[pkg/ui/webpack.vendor.js, line 21 at r29](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-Kv2_7iSeWYQZVuwCb87:-Kv2_7iTENSVLtzlsEaS:b-39oh8c) ([raw file](https://github.com/cockroachdb/cockroach/blob/16eca953948b0ea50d98aa02b070b0766792830a/pkg/ui/webpack.vendor.js#L21)):*
> ```JavaScript
>   output: {
>     filename: "vendor.dll.js",
>     path: __dirname + "/dist",
> ```

maybe `path.resolve(__dirname, "dist")`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-Kv2Y7f1uyKp3tZwY8CT:bdelm6q)*

couchand:
Just documenting here a brief conversation from the Admin UI standup today: @mrtracy pointed out that, since this shouldn't really affect the ability to cherrypick bugfixes into 1.1, we can go ahead and merge this whenever it's ready, rather than waiting as I had suggested.

benesch:
Review status: 0 of 21 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[pkg/ui/webpack.protos.js, line 14 at r29](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-Kv2_Oeo6CXdiK1yn4rt:-Kv4ycCcUi5zSaCTvYi1:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/16eca953948b0ea50d98aa02b070b0766792830a/pkg/ui/webpack.protos.js#L14)):*

maybe `path.resolve(__dirname, "dist")`?
</blockquote></details>

Done.

---

*[pkg/ui/webpack.vendor.js, line 14 at r29](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-Kv2_DoybeLpxDDdVr_1:-Kv4youjyROHZfbS8eXb:bfbq99i) ([raw file](https://github.com/cockroachdb/cockroach/blob/16eca953948b0ea50d98aa02b070b0766792830a/pkg/ui/webpack.vendor.js#L14)):*

I think you want to delete this line.
</blockquote></details>

Good point, done.

---

*[pkg/ui/webpack.vendor.js, line 21 at r29](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-Kv2_7iSeWYQZVuwCb87:-Kv4yoVrY_HzxoFUeGpb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/16eca953948b0ea50d98aa02b070b0766792830a/pkg/ui/webpack.vendor.js#L21)):*

maybe `path.resolve(__dirname, "dist")`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349)*

couchand:
:shipit:

---

Reviewed 1 of 18 files at r4, 1 of 18 files at r13, 1 of 18 files at r19, 1 of 20 files at r26, 15 of 20 files at r33, 4 of 6 files at r34, 1 of 1 files at r35, 2 of 2 files at r36.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-Kv7v5PqhA-lE0QOYTjO:b-xr9eiq)*

mrtracy:
This is LGTM

bdarnell:
`node-run.sh` LGTM. This class of issue is news to me, but I think the unix-hater's handbook could use a section on file descriptors vs file descriptions. 

---

Review status: 6 of 20 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18349#-:-KxAaiTAB7jknfILwz28:bvffkrv)*

benesch:
:tada: thanks for the reviews, everyone! Landing this while it's green.

PR #18350, opened 2017-09-07
ui: remove generated files from version control

benesch:
Github choked on https://github.com/cockroachdb/cockroach/pull/18349, so reopening under a different branch name.

Ignore the first commit; that's #18262.

From the last commit message:

```
Checking in src/js/protos.js and embedded.go causes several major
problems:

  * Every change to the UI, even if the diff is only one line,
    entirely rewrites embedded.go, which is upwards of 12MB in size.
    This adds a lot of weight to the Git repository.

  * Every commit that touches the UI or protobuf definitions must be
    serialized, as diffs in protos.js and embedded.go cannot be
    automatically merged.

   * Including embedded.go in a pull request breaks GitHub's "request
     review" feature.

Removing these files from version control solves all these problems with
only two downsides:

  * Every developer will need Node and Yarn installed. This seems
    reasonable to expect of our engineers and external contributors.
    This commit teaches our source archives to include a pre-generated
    embedded.go so that end users who simply want to build from source
    without Node and Yarn can do so.

  * Developers will likely need to rebuild the UI a few times per day
    as they switch between branches that span UI changes. This commit
    brings the UI compilation time down to about 5s when only
    first-party sources have been updated and 30s when UI dependencies
    are bumped (infrequently), so this also seems reasonable.
```


benesch:
Hmm, this one's broken too. Sigh.

PR #18351, opened 2017-09-07
kv: fix (the horribly broken) prev()

tbg:
prev() was broken in two ways:

- it had a one-off where `prev([a, b), b)` would  return `b` instead of `KeyMin`.
- in one path it was assembling the return from the start keys, not the end keys, which is
  wrong. When keys `>= c` have been handled, and your remaining batch is `[a, b)`, it would
  erroneously return `a` instead of `b`.

### DIFF
--- a/pkg/kv/batch.go
+++ b/pkg/kv/batch.go
@@ -125,7 +125,14 @@ func truncate(ba roachpb.BatchRequest, rs roachpb.RSpan) (roachpb.BatchRequest, 
 }
 
 // prev gives the right boundary of the union of all requests which don't
-// affect keys larger than the given key.
+// affect keys larger than the given key. Note that a right boundary is
+// exclusive, that is, the returned RKey is to be used as the exclusive
+// right endpoint in finding the next range to query.
+//
+// Informally, a call `prev(ba, k)` means: we've already executed the parts
+// of `ba` that intersect `[k, KeyMax)`; please tell me how far to the
+// left the next relevant request begins.
+//
 // TODO(tschottdorf): again, better on BatchRequest itself, but can't pull
 // 'keys' into 'roachpb'.
 func prev(ba roachpb.BatchRequest, k roachpb.RKey) (roachpb.RKey, error) {
@@ -140,31 +147,63 @@ func prev(ba roachpb.BatchRequest, k roachpb.RKey) (roachpb.RKey, error) {
 		if err != nil {
 			return nil, err
 		}
-		eAddr, err := keys.AddrUpperBound(h.EndKey)
+		endKey := h.EndKey
+		if len(endKey) == 0 {
+			// This is unintuitive, but if we have a point request at `x=k` then that request has
+			// already been satisfied (since the batch has already been executed for all keys `>=
+			// k`). We treat `k` as `[k,k)` which does the right thing below. It also does when `x >
+			// k` and `x < k`, so we're good.
+			//
+			// Note that if `x` is /Local/k/something, then AddrUpperBound below will turn it into
+			// `k\x00`, and so we're looking at the key range `[k, k\x00)`. This is exactly what we
+			// want since otherwise the result would be `k` and so the caller would restrict itself
+			// to `key < k`, but that excludes `k` itself and thus all local keys attached to it.
+			//
+			// See TestBatchPrevNext for a test case with commentary.
+			endKey = h.Key
+		}
+		eAddr, err := keys.AddrUpperBound(endKey)
 		if err != nil {
 			return nil, err
 		}
-		if len(eAddr) == 0 {
-			eAddr = addr.Next()
-		}
 		if !eAddr.Less(k) {
-			if !k.Less(addr) {
+			// EndKey is k or higher.
+			//           [x-------y)    !x.Less(k) -> skip
+			//         [x-------y)      !x.Less(k) -> skip
+			//      [x-------y)          x.Less(k) -> return k
+			//  [x------y)               x.Less(k) -> return k
+			// [x------y)                not in this branch
+			//          k
+			if addr.Less(k) {
 				// Range contains k, so won't be able to go lower.
+				// Note that in the special case in which the interval
+				// touches k, we don't take this branch. This reflects
+				// the fact that `prev(k)` means that all keys >= k have
+				// been handled, so a request `[k, x)` should simply be
+				// skipped.
 				return k, nil
 			}
 			// Range is disjoint from [KeyMin,k).
 			continue
 		}
+		// Current candidate interval is strictly to the left of `k`.
 		// We want the largest surviving candidate.
-		if candidate.Less(addr) {
-			candidate = addr
+		if candidate.Less(eAddr) {
+			candidate = eAddr
 		}
 	}
 	return candidate, nil
 }
 
-// next gives the left boundary of the union of all requests which don't
-// affect keys less than the given key.
+// next gives the left boundary of the union of all requests which don't affect
+// keys less than the given key. Note that the left boundary is inclusive, that
+// is, the returned RKey is the inclusive left endpoint of the keys the request
+// should operate on next.
+//
+// Informally, a call `next(ba, k)` means: we've already executed the parts of
+// `ba` that intersect `[KeyMin, k)`; please tell me how far to the right the
+// next relevant request begins.
+//
 // TODO(tschottdorf): again, better on BatchRequest itself, but can't pull
 // 'keys' into 'proto'.
 func next(ba roachpb.BatchRequest, k roachpb.RKey) (roachpb.RKey, error) {
--- a/pkg/keys/keys.go
+++ b/pkg/keys/keys.go
@@ -352,20 +352,25 @@ func IsLocal(k roachpb.Key) bool {
 	return bytes.HasPrefix(k, localPrefix)
 }
 
-// Addr returns the address for the key, used to lookup the range containing
-// the key. In the normal case, this is simply the key's value. However, for
-// local keys, such as transaction records, range-spanning binary tree node
-// pointers, the address is the inner encoded key, with the local key prefix
-// and the suffix and optional detail removed. This address unwrapping is
-// performed repeatedly in the case of doubly-local keys. In this way, local
-// keys address to the same range as non-local keys, but are stored separately
-// so that they don't collide with user-space or global system keys.
+// Addr returns the address for the key, used to lookup the range containing the
+// key. In the normal case, this is simply the key's value. However, for local
+// keys, such as transaction records, the address is the inner encoded key, with
+// the local key prefix and the suffix and optional detail removed. This address
+// unwrapping is performed repeatedly in the case of doubly-local keys. In this
+// way, local keys address to the same range as non-local keys, but are stored
+// separately so that they don't collide with user-space or global system keys.
+//
+// Logically, the keys are arranged as follows:
+//
+// k1 /local/k1/KeyMin ... /local/k1/KeyMax k1\x00 /local/k1/x00/KeyMin ...
 //
 // However, not all local keys are addressable in the global map. Only range
 // local keys incorporating a range key (start key or transaction key) are
 // addressable (e.g. range metadata and txn records). Range local keys
 // incorporating the Range ID are not (e.g. abort cache entries, and range
 // stats).
+//
+// See AddrEndKey which is to be used when `k` is the EndKey of an interval.
 func Addr(k roachpb.Key) (roachpb.RKey, error) {
 	if !IsLocal(k) {
 		return roachpb.RKey(k), nil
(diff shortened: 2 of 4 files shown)


tschottdorf:
@nvanbenschoten please give this a super thorough review, and if you have the local-key archaeology handy, I'd appreciate pointers (I can do it myself but haven't so far).

pkg/kv/batch.go
```diff
@@ -126,6 +126,11 @@ func truncate(ba roachpb.BatchRequest, rs roachpb.RSpan) (roachpb.BatchRequest,
 
 // prev gives the right boundary of the union of all requests which don't
 // affect keys larger than the given key.
+//
+// Informally, a call `prev(ba, k)` means: we've already executed the parts
+// of `ba` that intersect `[key, KeyMax)`; please tell me how far to the
```
bdarnell: s/key/k/

pkg/kv/batch.go
```diff
@@ -140,24 +145,38 @@ func prev(ba roachpb.BatchRequest, k roachpb.RKey) (roachpb.RKey, error) {
 		if err != nil {
 			return nil, err
 		}
+		// FIXINPR(tschottdorf,nvanbenschoten): yikes, this is totally paged out.
```
bdarnell: Range-local keys exist in the "gaps" between regular keys. `keys.Addr` finds the lower bound: the greatest regular key that is less than or equal to the input, and AddrUpperBound finds the smallest regular key that is greater than or equal to the input. 

Here, we're trying to expand `[StartKey, EndKey)` to find the keys that form the tightest (potential) range boundaries around the given keys. So we must use Addr on StartKey and AddrUpperBound on EndKey to avoid leaving EndKey outside the new bounds.

bdarnell (approved):

pkg/kv/batch.go
tschottdorf: Done.

pkg/kv/batch.go
tschottdorf: Thanks for the reminder, added misc commentary.

tschottdorf:
Ok, ready for final pass. summoning @nvanbenschoten

pkg/kv/batch.go
tschottdorf: Done.

pkg/kv/batch.go
tschottdorf: Done.

nvanbenschoten:
Reviewed 2 of 3 files at r2.
Review status: 2 of 4 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

*[pkg/keys/keys.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXfDB0iRC6vlWTRQaF:-KtXfDB1hp5emJXX_qPf:b42uvm9) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/keys/keys.go#L357)):*
> ```Go
> // Addr returns the address for the key, used to lookup the range containing
> // the key. In the normal case, this is simply the key's value. However, for
> // local keys, such as transaction records, range-spanning binary tree node
> ```

somewhat unrelated: the `range-spanning binary tree node pointers` is gone, right?

---

*[pkg/keys/keys.go, line 415 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXfqkfz_F1yJo_Nl6u:-KtXfqkgCr3MnYIlYIu7:b-aog851) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/keys/keys.go#L415)):*
> ```Go
> // but is suitable for use as the EndKey of a span involving a range-local key.
> //
> // Logically, the keys are arranged as follows:
> ```

Mind adding a similar example to `Addr`?

---

*[pkg/kv/batch.go, line 128 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXk3NNgLIcTp8U8acc:-KtXk3NNgLIcTp8U8acd:b-4lfr8x) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L128)):*
> ```Go
> 
> // prev gives the right boundary of the union of all requests which don't
> // affect keys larger than the given key.
> ```

I think it would be valuable here and below to remind the reader that a right boundary is exclusive but a left boundary is inclusive. That's the subtlety that causes the logic in each to be different and what caused the one-off you fixed here.

---

*[pkg/kv/batch.go, line 130 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXg_vN1CByFg25KTJV:-KtXg_vN1CByFg25KTJW:b-puwre5) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L130)):*
> ```Go
> // affect keys larger than the given key.
> //
> // Informally, a call `prev(ba, k)` means: we've already executed the parts
> ```

Add something like this to `next` please.

---

*[pkg/kv/batch.go, line 153 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXh2Y7Q0S_0fgNoelm:-KtXh2Y7Q0S_0fgNoeln:b-387uv8) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L153)):*
> ```Go
> 		}
> 		if len(eAddr) == 0 {
> 			// This is unintuitive, but if we have a point request at `x=k` then we're done
> ```

`we're done with k` is vague. Let's expand this comment now before we all forget how anything in here works for another 18 months.

---

*[pkg/kv/batch.go, line 158 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXj40EQ4M6VoKOEPaT:-KtXj40EQ4M6VoKOEPaU:b474qxy) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L158)):*
> ```Go
> 			eAddr = addr
> 		}
> 		if !eAddr.Less(k) {
> ```

Examples for each branch here would make this a lot easier to skim. Your call.

---

*[pkg/kv/batch_test.go, line 45 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXl46PHMInP9lbQjSh:-KtXl46PHMInP9lbQjSi:b-fotmvu) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch_test.go#L45)):*
> ```Go
> 	max, min := string(roachpb.RKeyMax), string(roachpb.RKeyMin)
> 	abc := span("a", "", "b", "", "c", "")
> 	testCases := []struct {
> ```

None of these test cases test multiple eligible `candidates`, right?

---

*[pkg/kv/batch_test.go, line 136 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXlwK7FHgveM8dMwJk:-KtXlwK7FHgveM8dMwJl:bn6rw4y) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch_test.go#L136)):*
> ```Go
> 			expBW: min,
> 		},
> 		{spans: span(loc("b"), ""), key: "b\x00",
> ```

Is this case right? I would have expected `expBW=b\x00`. I must be mixed up somewhere.

I would think we'd want:

```
endKey := h.EndKey
if len(endKey) == 0 {
    endKey = h.Key
}
eAddr, err := keys.AddrUpperBound(endKey)
if err != nil {
    return nil, err
}
```

instead of

```
eAddr, err := keys.AddrUpperBound(h.EndKey)
if err != nil {
    return nil, err
}
if len(eAddr) == 0 {
    eAddr = addr
}
```

---

*[pkg/kv/batch_test.go, line 143 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXlHBWEnPbyteQC2uS:-KtXlHBWEnPbyteQC2uT:batsvgs) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch_test.go#L143)):*
> ```Go
> 
> 	for _, test := range testCases {
> 		t.Run("", func(t *testing.T) {
> ```

No subtest name?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351)*

tschottdorf:
TFTR! PR is updated.

---

Review status: 0 of 4 files reviewed at latest revision, 11 unresolved discussions.

---

*[pkg/keys/keys.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXfDB0iRC6vlWTRQaF:-KtY5BRIYz8fUvJV0TOH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/keys/keys.go#L357)):*

somewhat unrelated: the `range-spanning binary tree node pointers` is gone, right?
</blockquote></details>

Done.

---

*[pkg/keys/keys.go, line 415 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXfqkfz_F1yJo_Nl6u:-KtY6gD_8oD3yt_eKVIe:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/keys/keys.go#L415)):*

Mind adding a similar example to `Addr`?
</blockquote></details>

Done.

---

*[pkg/kv/batch.go, line 128 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXk3NNgLIcTp8U8acc:-KtY7HZvmpkulr78uvSC:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L128)):*

I think it would be valuable here and below to remind the reader that a right boundary is exclusive but a left boundary is inclusive. That's the subtlety that causes the logic in each to be different and what caused the one-off you fixed here.
</blockquote></details>

Done.

---

*[pkg/kv/batch.go, line 130 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXg_vN1CByFg25KTJV:-KtY7I-eDDZ1ESDH3a9T:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L130)):*

Add something like this to `next` please.
</blockquote></details>

Done.

---

*[pkg/kv/batch.go, line 153 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXh2Y7Q0S_0fgNoelm:-KtY7dSOR5jd9GRNpAuD:b-haa1be) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L153)):*

`we're done with k` is vague. Let's expand this comment now before we all forget how anything in here works for another 18 months.
</blockquote></details>

But it's so much fun to find out again! Done.

---

*[pkg/kv/batch.go, line 158 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXj40EQ4M6VoKOEPaT:-KtY95q6pP6goaE2jU3V:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch.go#L158)):*

Examples for each branch here would make this a lot easier to skim. Your call.
</blockquote></details>

Done.

---

*[pkg/kv/batch_test.go, line 45 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXl46PHMInP9lbQjSh:-KtYBjda-X_gmWGANqsM:bjfwunu) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch_test.go#L45)):*

None of these test cases test multiple eligible `candidates`, right?
</blockquote></details>

They do now.

---

*[pkg/kv/batch_test.go, line 136 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXlwK7FHgveM8dMwJk:-KtY9gTsik23PL3wqcrv:bs8duvo) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch_test.go#L136)):*

Is this case right? I would have expected `expBW=b\x00`. I must be mixed up somewhere.

I would think we'd want:

```
endKey := h.EndKey
if len(endKey) == 0 {
    endKey = h.Key
}
eAddr, err := keys.AddrUpperBound(endKey)
if err != nil {
    return nil, err
}
```

instead of

```
eAddr, err := keys.AddrUpperBound(h.EndKey)
if err != nil {
    return nil, err
}
if len(eAddr) == 0 {
    eAddr = addr
}
```

</blockquote></details>

Yikes, you're right. Fixed and added a boatload more comments.

---

*[pkg/kv/batch_test.go, line 143 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-KtXlHBWEnPbyteQC2uS:-KtYAqm8Ujurrq_rfZK6:bqytoto) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0860f35323c1a4dfa30e3cd6ee3cce8db039d62/pkg/kv/batch_test.go#L143)):*

No subtest name?
</blockquote></details>

No, apparently this is a thing. The test harness will enumerate it like `#%d`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-:-KtYDJ1w0KKQ0Jm22vHY:bjf48fz)*

nvanbenschoten:
:lgtm_strong: thanks for adding all the comments! Hopefully we won't have to touch this code again for a long long time. 

---

Reviewed 1 of 3 files at r1, 1 of 3 files at r2, 2 of 3 files at r3.
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18351#-:-KtYGXTk31ChY4eTswjv:b-xv6b5g)*

knz:
This needs to go to 1.1, but perhaps 1.0 too (if we ever release 1.0.7)

tschottdorf:
@knz ack, l'll see to it.

Issue #11987, opened 2016-12-04
storage: test replicate queue behavior


petermattis:
#11982 fixed an issue where the replicate queue was retrying lease transfers for a range repeatedly because the available target replica was behind and thus not a candidate. The replicate queue code does not make testing this easy. We should either refactor it or create infrastructure to make such tests feasible.

petermattis:
@a-robinson tossing your way as this likely ties in with other rebalancing work you're doing.

tschottdorf:
Is there anything left to do here?

a-robinson:
Well, we're still missing the tests and test infrastructure for the replicate queue that were desired when the issue was created. I think it's fair to leave open, especially given that the purpose of the replicate queue is changing slightly as part of https://github.com/cockroachdb/cockroach/issues/26059

PR #18364, opened 2017-09-08
storage: Consider which replica will be removed when adding a replica…

### LINKED ISSUE #17971: storage: Consider which replica will be removed when adding a replica to improve balance
In clusters with multiple localities, we try to maintain diversity by spreading the replicas for each range as evenly across localities as possible. This means that in a cluster with 3 localities and 3 replicas per range, we'll try to keep ` replica in each locality for each range.

When in a rebalancing state where one locality has 1 replicas and the other localities each have 1, we'll always remove a replica from the locality that has 2. This is good.

What isn't as good is that we don't consider this when deciding whether to rebalance in the first place. Our rebalancing logic will kick in if a possible new destination is a better fit than any of the existing replicas, even if the new destination is in a different locality and thus won't actually be a direct replacement for the existing replica that isn't a great fit.

I haven't seen this cause massive problems by itself, but in combination with another problem (I've seen it flare up with both #17879 and #17970) it can make for rebalance thrashing, where we repeatedly add and remove a replica on the same 1 or 2 nodes.

Not all situations will be as straightforward as the 3-locality example above, so we'll have to make sure the fix for this is somewhat more general than just making sure that a potential replica to add is in the same locality as the worst existing replica.

a6802739:
@a-robinson 

Fixes #17971. @a-robinson , Thanks for your guidance, I just simulate `(*Allocator).RemoveTarget` before we actually add the replica during the rebalance.

Still don't know how to add a test for this.

### DIFF
--- a/pkg/storage/allocator.go
+++ b/pkg/storage/allocator.go
@@ -184,12 +184,14 @@ type RangeInfo struct {
 }
 
 func rangeInfoForRepl(repl *Replica, desc *roachpb.RangeDescriptor) RangeInfo {
-	writesPerSecond, _ := repl.writeStats.avgQPS()
-	return RangeInfo{
-		Desc:            desc,
-		LogicalBytes:    repl.GetMVCCStats().Total(),
-		WritesPerSecond: writesPerSecond,
+	info := RangeInfo{
+		Desc:         desc,
+		LogicalBytes: repl.GetMVCCStats().Total(),
 	}
+	if writesPerSecond, dur := repl.writeStats.avgQPS(); dur >= MinStatsDuration {
+		info.WritesPerSecond = writesPerSecond
+	}
+	return info
 }
 
 // Allocator tries to spread replicas as evenly as possible across the stores
@@ -374,6 +376,25 @@ func (a *Allocator) AllocateTarget(
 	}
 }
 
+func (a Allocator) simulateRemoveTarget(
+	ctx context.Context,
+	targetStore roachpb.StoreID,
+	constraints config.Constraints,
+	candidates []roachpb.ReplicaDescriptor,
+	rangeInfo RangeInfo,
+) (roachpb.ReplicaDescriptor, string, error) {
+	// Update statistics first
+	// TODO(a-robinson): This could theoretically interfere with decisions made by other goroutines,
+	// but as of October 2017 calls to the Allocator are mostly serialized by the ReplicateQueue
+	// (with the main exceptions being Scatter and the status server's allocator debug endpoint).
+	// Try to make this interfere less with other callers.
+	a.storePool.updateLocalStoreAfterRebalance(targetStore, rangeInfo, roachpb.ADD_REPLICA)
+	defer func() {
+		a.storePool.updateLocalStoreAfterRebalance(targetStore, rangeInfo, roachpb.REMOVE_REPLICA)
+	}()
+	return a.RemoveTarget(ctx, constraints, candidates, rangeInfo)
+}
+
 // RemoveTarget returns a suitable replica to remove from the provided replica
 // set. It first attempts to randomly select a target from the set of stores
 // that have greater than the average number of replicas. Failing that, it
@@ -444,7 +465,11 @@ func (a Allocator) RemoveTarget(
 // rebalance. This helps prevent a stampeding herd targeting an abnormally
 // under-utilized store.
 func (a Allocator) RebalanceTarget(
-	ctx context.Context, constraints config.Constraints, rangeInfo RangeInfo, filter storeFilter,
+	ctx context.Context,
+	constraints config.Constraints,
+	raftStatus *raft.Status,
+	rangeInfo RangeInfo,
+	filter storeFilter,
 ) (*roachpb.StoreDescriptor, string) {
 	sl, _, _ := a.storePool.getStoreList(rangeInfo.Desc.RangeID, filter)
 
@@ -490,6 +515,37 @@ func (a Allocator) RebalanceTarget(
 		candidates, existingCandidates, target)
 	if target == nil {
 		return nil, ""
+	}
+	// We could make a simulation here to verify whether we'll remove the target we'll rebalance to.
+	for len(candidates) > 0 {
+		if raftStatus == nil || raftStatus.Progress == nil {
+			break
+		}
+		newReplica := roachpb.ReplicaDescriptor{
+			NodeID:    target.store.Node.NodeID,
+			StoreID:   target.store.StoreID,
+			ReplicaID: rangeInfo.Desc.NextReplicaID,
+		}
+		desc := *rangeInfo.Desc
+		desc.Replicas = append(desc.Replicas, newReplica)
+		rangeInfo.Desc = &desc
+
+		replicaCandidates := simulateFilterUnremovableReplicas(raftStatus, desc.Replicas, newReplica.ReplicaID)
+
+		removeReplica, _, err := a.simulateRemoveTarget(ctx, target.store.StoreID, constraints, replicaCandidates, rangeInfo)
+		if err != nil {
+			log.Warningf(ctx, "simulating RemoveTarget failed: %s", err)
+			return nil, ""
+		}
+		if removeReplica.StoreID != target.store.StoreID {
+			break
+		}
+		newTargets := candidates.removeCandidate(*target)
+		newTarget := newTargets.selectGood(a.randGen)
+		if newTarget == nil {
+			return nil, ""
+		}
+		target = newTarget
 	}
 	details, err := json.Marshal(decisionDetails{
 		Target:               target.String(),
@@ -875,6 +931,16 @@ func filterBehindReplicas(
 		}
 	}
 	return candidates
+}
+
+func simulateFilterUnremovableReplicas(
+	raftStatus *raft.Status,
+	replicas []roachpb.ReplicaDescriptor,
+	brandNewReplicaID roachpb.ReplicaID,
+) []roachpb.ReplicaDescriptor {
+	status := *raftStatus
+	status.Progress[uint64(brandNewReplicaID)] = raft.Progress{Match: 0}
+	return filterUnremovableReplicas(&status, replicas, brandNewReplicaID)
 }
 
 // filterUnremovableReplicas removes any unremovable replicas from the supplied
--- a/pkg/storage/allocator_test.go
+++ b/pkg/storage/allocator_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/rpc"
 	"github.com/cockroachdb/cockroach/pkg/settings/cluster"
+	"github.com/cockroachdb/cockroach/pkg/storage/engine/enginepb"
 	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/gossiputil"
 	"github.com/cockroachdb/cockroach/pkg/util"
@@ -657,6 +658,7 @@ func TestAllocatorRebalance(t *testing.T) {
 		result, _ := a.RebalanceTarget(
 			ctx,
 			config.Constraints{},
+			nil,
 			testRangeInfo([]roachpb.ReplicaDescriptor{{StoreID: 3}}, firstRange),
 			storeFilterThrottled,
 		)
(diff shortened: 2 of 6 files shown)


a-robinson:
Hmm, I think we may need to take this a little farther. `RebalanceTarget` might internally find multiple good stores to rebalance a range to. It only returns the one that it thinks is best, but it may have found others that would be really good fits. In such cases, we should rebalance to one of the other candidates, but this code will just return without doing a rebalance. I think we need to do this sort of filtering inside `RebalanceTarget` so that it can return a different candidate if there is one that would work instead.

As for testing, the replicate queue is unfortunately in need of some test love (#11987). If you could write one that covers this code, that would be great, otherwise some manual testing may be needed. The best tests for it currently are the allocator acceptance tests and running the `cmd/allocsim` tool and inspecting the output by hand.

---

---

*[pkg/storage/replicate_queue.go, line 498 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364#-Ktso9nkcPen3-l0uuLK:-Ktso9nl_Dq7DXk0EApI:b8pwnk6) ([raw file](https://github.com/cockroachdb/cockroach/blob/321f8b26853abac4528504007b20eccff2b77b59/pkg/storage/replicate_queue.go#L498)):*
> ```Go
> 					StoreID: rebalanceReplica.StoreID,
> 				}
> 				candidates := filterUnremovableReplicas(repl.RaftStatus(), desc.Replicas, 0)
> ```

I think we'd be best off if we simulate the removal process as closely as possible, which would mean including the new replica in the `RaftStatus` with a `Progress` of 0 and passing an ID for it as the `brandNewReplicaID` to `filterUnremovableReplicas` rather than appending it to the list of candidates afterwards.

---

*[pkg/storage/replicate_queue.go, line 500 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364#-KtsotNJ-2N2m-lKVSVU:-KtsotNKWOMxIjoizm9j:b-8etsfn) ([raw file](https://github.com/cockroachdb/cockroach/blob/321f8b26853abac4528504007b20eccff2b77b59/pkg/storage/replicate_queue.go#L500)):*
> ```Go
> 				candidates := filterUnremovableReplicas(repl.RaftStatus(), desc.Replicas, 0)
> 				candidates = append(candidates, rep)
> 				removeReplica, _, err := rq.allocator.RemoveTarget(ctx, zone.Constraints, candidates, rangeInfo)
> ```

I'm a little concerned about how this will interact with #18425, since we aren't using the updated stats here. Do you have any thoughts on how to make this more accurately reflect the updated stats? I think we may be best off not doing anything about it, but I imagine this discrepancy will probably be a problem at some point.

---

*[pkg/storage/replicate_queue.go, line 505 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364#-KtspxPE6PMeWDSpIhK5:-KtspxPE6PMeWDSpIhK6:biw5jg9) ([raw file](https://github.com/cockroachdb/cockroach/blob/321f8b26853abac4528504007b20eccff2b77b59/pkg/storage/replicate_queue.go#L505)):*
> ```Go
> 				}
> 				if removeReplica.StoreID == rebalanceReplica.StoreID {
> 					log.VEventf(ctx, 1, "we may immediately remove the same replica after we add this replica for rebalancing")
> ```

If you don't mind me rewording this, I'd phrase it as "not rebalancing to candidate %+v because it would be removed immediately after being added"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364#-:-KtsqSKznU6TP0MIqQEl:bcycgon)*

a6802739:
@a-robinson , Sorry, I couldn't login in Reviewable right Now.

If we already pass the new replica ID as `brandNewReplicaID` to `filterUnremovableReplicas`, why we still include the new replica in the RaftStatus with a Progress of 0?


And what do you mean `I'm a little concerned about how this will interact with #18425, since we aren't using the updated stats here`, I'm sorry, I don't quite understand here? Do you mean we should add the update stats first when we try to simulate the remove procedure?

a-robinson:
Sorry, I'll try to stick to Github comments if Reviewable is inconvenient for you.

> If we already pass the new replica ID as `brandNewReplicaID` to `filterUnremovableReplicas`, why we still include the new replica in the RaftStatus with a Progress of 0?

Because it makes a difference in what `filterUnremovableReplicas` will do. If the new replica isn't included in the `RaftStatus` then `filterUnremovableReplicas` will behave as if it doesn't exist (even if we pass it in as the `brandNewReplicaID`). If we add the new replica to the `RaftStatus` but don't pass it in as the `brandNewReplicaID`, then `filterUnremovableReplicas` will consider the new replica to be behind, which will make it likely to be selected.

> And what do you mean `I'm a little concerned about how this will interact with #18425, since we aren't using the updated stats here`, I'm sorry, I don't quite understand here? Do you mean we should add the update stats first when we try to simulate the remove procedure?

The goal here is to simulate the actual removal process as accurately as possible. If the actual removal process will be acting on updated stats and this simulated removal process isn't, then the simulated process may be incorrect sometimes. Modifying the stats for the simulation in a way that's easy to roll back and doesn't affect anything else happening in the system may be difficult, though. What do you think?

a6802739:
> The goal here is to simulate the actual removal process as accurately as possible. If the actual removal process will be acting on updated stats and this simulated removal process isn't, then the simulated process may be incorrect sometimes. Modifying the stats for the simulation in a way that's easy to roll back and doesn't affect anything else happening in the system may be difficult, though. What do you think?

I think the actual intent is just to simulate `RemoveTarget`, but we didn't actually did the rebalance before the simulation. If we want to be very close to the real `RemoveTarget`, I think we we just to to update the local copy of target rebalance store, I mean  add `WritesPerSecond` for the target store, and recover it after the simulation?

a-robinson:
That would be ideal, yes. It'll probably require a fair amount of refactoring, so use your judgment, but that would be best.

a6802739:
@a-robinson, Thank you for your review. 
I have made a simulation in `RemoveTarget`. Still not sure how to add a test for it.

pkg/storage/allocator.go
```diff
@@ -491,6 +511,33 @@ func (a Allocator) RebalanceTarget(
 	if target == nil {
 		return nil, ""
 	}
+	// We could make a simulation here to verify whether we'll remove the target we'll rebalance to.
```
a-robinson: Do you think we could avoid this code duplication with what's in `replicate_queue.go` and instead create a shared function they can both use? If they're separate like this, it seems likely that they'll accidentally diverge in the future.

pkg/storage/allocator.go
```diff
@@ -491,6 +511,33 @@ func (a Allocator) RebalanceTarget(
...
+			log.Warningf(ctx, "simulating RemoveTarget failed: %s", err)
+			return nil, ""
+		}
+		if removeReplica.StoreID == target.store.StoreID {
+			newTargets := candidates.removeCandidate(*target)
+			newTarget := newTargets.selectGood(a.randGen)
+			if newTarget == nil {
+				return nil, ""
+			}
+			target = newTarget
```
a-robinson: I think we actually want this logic to happen in a loop. What if the `newTarget` is also going to be immediately removed?

a-robinson (commented):
I'm sorry for forgetting about this for a while, @a6802739! We definitely still want this to get in.

Regarding testing, we have a few options of increasing difficulty:

1. Do something like in #18425, where we create a configuration that would previously have caused thrashing and test that `RebalanceTarget` will prefer not rebalancing over rebalancing to something that will immediately be removed.
1. Create a test cluster with multiple nodes via something like `multiTestContext` or `TestCluster` with localities configured in a way that would normally trigger this bug and ensure that the rebalancing settles down. For an example configuration, see #19013. To summarize that issue, a cluster with 2 nodes in locality X, 1 node in locality Y, and 1 node in locality Z will exhibit the kind of thrashing that this PR should fix.
1. Most complicated would be to randomly generate cluster/replica configurations to run through a unit test that tries runs `RebalanceTarget`, makes (or simulates) the requested change, and then runs `RemoveTarget`. The test would fail if any configuration chose to remove the replica that it just added.

a-robinson:
Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364)*

pkg/storage/allocator.go
a6802739: Hmm, Sorry, I couldn't quite understand what do you mean here. I didn't make any change to `replicate_queue.go`.
Would you please point out the  code duplication with `replicate_queue.go`? Thank you very much.

pkg/storage/allocator.go
```diff
@@ -374,6 +374,22 @@ func (a *Allocator) AllocateTarget(
 	}
 }
 
+func (a Allocator) simulationRemoveTarget(
```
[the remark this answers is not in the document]
a6802739: Great. Done.

pkg/storage/allocator.go
```diff
@@ -374,6 +374,22 @@ func (a *Allocator) AllocateTarget(
...
+	ctx context.Context,
+	replica *Replica,
+	targetStore roachpb.StoreID,
+	constraints config.Constraints,
+	candidates []roachpb.ReplicaDescriptor,
+	rangeInfo RangeInfo,
+) (roachpb.ReplicaDescriptor, string, error) {
+	// Update statics first
```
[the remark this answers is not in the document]
a6802739: My mistake. Done.

pkg/storage/allocator.go
```diff
@@ -374,6 +374,22 @@ func (a *Allocator) AllocateTarget(
...
+	a.storePool.updateLocalStoreAfterRebalance(targetStore, replica, roachpb.ADD_REPLICA)
```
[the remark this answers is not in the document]
a6802739: Sure, Done.

pkg/storage/allocator.go
a6802739: Yeah, You are right. My mistake, Done.

a6802739:
@a-robinson, Thank you for your reference about unit test. Please have a review again. Thank you very much.

a6802739:
@a-robinson , Sorry, I don't know why  the teamcity failed. And I didn't see any important information from the teamcity.

pkg/storage/allocator.go
```diff
@@ -444,7 +464,12 @@ func (a Allocator) RemoveTarget(
 // rebalance. This helps prevent a stampeding herd targeting an abnormally
 // under-utilized store.
 func (a Allocator) RebalanceTarget(
-	ctx context.Context, constraints config.Constraints, rangeInfo RangeInfo, filter storeFilter,
+	ctx context.Context,
+	constraints config.Constraints,
+	repl *Replica,
```
a-robinson: Passing in a `Replica` is pretty broad. It looks like we could shrink this down to just using the `RangeInfo` if we made `simulateRemoveTarget` just take a `RangeInfo` instead of a `Replica`.

pkg/storage/allocator_test.go
```diff
@@ -685,9 +694,164 @@ func TestAllocatorRebalance(t *testing.T) {
 	}
 }
 
+// TestAllocatorRebalanceTarget could help us to verify whether we'll rebalance to a target that
+// we'll immediately remove.
+func TestAllocatorRebalanceTarget(t *testing.T) {
```
a-robinson: Can you confirm that this test passes with your change but fails without it?

pkg/storage/allocator_test.go
```diff
@@ -685,9 +694,164 @@ func TestAllocatorRebalance(t *testing.T) {
...
+			Capacity: roachpb.StoreCapacity{
+				Capacity:        100,
+				Available:       30,
+				RangeCount:      150,
+				LogicalBytes:    80,
+				WritesPerSecond: 100,
+			},
+		},
+		{
+			StoreID: 5,
```
a-robinson: Can you add a comment explaining these stores and the fake replica stats below? I'm not really sure why this fifth store is here or why are the values are set the way that they are. And if I'm not sure, then future readers are even less likely to understand what you were thinking while creating the test.

pkg/storage/allocator_test.go
```diff
@@ -685,9 +694,164 @@ func TestAllocatorRebalance(t *testing.T) {
...
+				LogicalBytes:    70,
+				WritesPerSecond: 60,
+			},
+		},
+	}
+	sg := gossiputil.NewStoreGossiper(g)
+	sg.GossipStores(stores, t)
+
+	st := a.storePool.st
+	EnableStatsBasedRebalancing.Override(&st.SV, true)
```
a-robinson: Why? In 1.1, this problem occurs even if stats-based rebalancing is disabled. Testing the default setting seems more valuable than testing the non-default setting. Keeping this false would also let us remove the capacity, available, logicalbytes, and writespersecond fields from all the store descriptors above.

pkg/storage/allocator_test.go
```diff
@@ -685,9 +694,164 @@ func TestAllocatorRebalance(t *testing.T) {
...
+	for i := 0; i < 10; i++ {
+		result, _ := a.RebalanceTarget(
+			context.Background(),
+			config.Constraints{},
+			repl,
+			status,
+			testRangeInfo(replicas, firstRange),
+			storeFilterThrottled,
+		)
+		if expected := roachpb.StoreID(5); result.StoreID != expected {
```
a-robinson: I don't think having a fifth store here is any better than just having 4 stores and verifying that we don't want to move a replica to the fourth.

a-robinson (commented):
I think that test failure in teamcity is just a flake - https://github.com/cockroachdb/cockroach/issues/18554

pkg/storage/allocator.go
a-robinson: I was referring to the code for the `AllocatorRemove` case in `processOneChange`, but on looking at it again it doesn't look like there's much that could be de-duped. Sorry for the confusion!

a-robinson:
Reviewed 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364)*

pkg/storage/allocator_test.go
a6802739: Yes, I could confirm that.

pkg/storage/allocator_test.go
a6802739: Hmm, my original thinking is that if we didn't add the fifth store, it will have no store to make an rebalance, So I just add a fifth store here. But you are right, it seems it have no need to add the fifth store here, it could just return a nil target for `RebalanceTarget`.

pkg/storage/allocator_test.go
a6802739: Yeah, you are right.

pkg/storage/allocator_test.go
a6802739: hmm, Yeah, that will be more easier. Thank you very much.

pkg/storage/allocator_test.go
a6802739: Yeah, you are right.  4 stores will be enough.

pkg/storage/allocator.go
a6802739: So we should guarantee the duration for `WritesPerSecond` in `RangeInfo` exceeds `MinStatsDuration`?

a-robinson: Yes, I suppose so. Since this code only runs on the leaseholder, it pretty much always should be, anyway.

pkg/storage/allocator.go
```diff
@@ -491,6 +516,37 @@ func (a Allocator) RebalanceTarget(
 	if target == nil {
 		return nil, ""
 	}
+	// We could make a simulation here to verify whether we'll remove the target we'll rebalance to.
+	for len(candidates) >= 0 {
```
a-robinson: Sorry for not noticing this last time, but `for len(candidates) >= 0` is a pretty meaningless loop condition. Did you mean `for len(candidates) > 0`? Or to just make this an infinite loop?

pkg/storage/allocator_test.go
```diff
@@ -685,6 +689,124 @@ func TestAllocatorRebalance(t *testing.T) {
...
+// TestAllocatorRebalanceTarget could help us to verify whether we'll rebalance to a target that
+// we'll immediately remove.
+func TestAllocatorRebalanceTarget(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+	manual := hlc.NewManualClock(123)
+	clock := hlc.NewClock(manual.UnixNano, time.Nanosecond)
+	// We're going to manually mark stores dead in this test.
+	stopper, g, _, a, _ := createTestAllocator( /* deterministic */ false)
+	defer stopper.Stop(context.Background())
+	stores := []*roachpb.StoreDescriptor{
```
a-robinson: Thanks for explaining in the code review, but a comment in the code explaining this configuration of stores would be very helpful to future maintainers.

a-robinson (commented):
Just a couple small comments left. Should be good to go after the last couple changes :)

a-robinson:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364)*

pkg/storage/allocator_test.go
a6802739: Sorry for that. I have already added a comment about the configuration of stores.

pkg/storage/allocator.go
a6802739: Yeah, I just thought it will break in the loop condition. I have changed it as your suggestion.

a-robinson (approved):
LGTM, thanks!

a-robinson:
Reviewed 5 of 5 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18364)*

a-robinson (commented):
It looks like it fails to build, though, according to teamcity:

```
[TestStyle/TestErrCheck] style_test.go:819: err=exit status 2, stderr=/go/src/github.com/cockroachdb/cockroach/pkg/storage/store_pool_test.go:434:45: cannot use replica (variable of type *Replica) as RangeInfo value in argument to sp.updateLocalStoreAfterRebalance
```

a6802739:
@a-robinson, Thank you very much. I have changed it already. Wait for a successful CI of the teamcity. :)

PR #18368, opened 2017-09-08
cherrypick-1.1: build: Skip docker publish step for beta releases

bdarnell:
Temporarily disable a test step that stopped the docker publish from
happening.


benesch (approved):
LGTM!

PR #18372, opened 2017-09-08
server: Fix DistSQL query plan debug API

lgo:
The keyword JSON is not valid as a column, likely because of a change to
the YACC grammar, causing this query to error. This fixes that and adds
a smoke test to ensure the endpoint doesn't fail.

### DIFF
--- a/pkg/server/admin.go
+++ b/pkg/server/admin.go
@@ -1101,7 +1101,7 @@ func (s *adminServer) QueryPlan(
 	}
 
 	explain := fmt.Sprintf(
-		"SELECT JSON FROM [EXPLAIN (distsql) %s]",
+		"SELECT \"JSON\" FROM [EXPLAIN (distsql) %s]",
 		strings.Trim(req.Query, ";"))
 	r, err := s.server.sqlExecutor.ExecuteStatementsBuffered(session, explain, nil, 1)
 	if err != nil {
--- a/pkg/server/admin_test.go
+++ b/pkg/server/admin_test.go
@@ -1159,6 +1159,40 @@ func TestAdminAPIJobs(t *testing.T) {
 	}
 }
 
+func TestAdminAPIQueryPlan(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+
+	s, conn, _ := serverutils.StartServer(t, base.TestServerArgs{})
+	defer s.Stopper().Stop(context.TODO())
+	sqlDB := sqlutils.MakeSQLRunner(t, conn)
+
+	sqlDB.Exec(`CREATE DATABASE api_test`)
+	sqlDB.Exec(`CREATE TABLE api_test.t1 (id int primary key, name string)`)
+	sqlDB.Exec(`CREATE TABLE api_test.t2 (id int primary key, name string)`)
+
+	testCases := []struct {
+		query string
+		exp   []string
+	}{
+		{"SELECT sum(id) FROM api_test.t1", []string{"nodeNames\":[\"1\"]", "Out: @1"}},
+		{"SELECT sum(1) FROM api_test.t1 JOIN api_test.t2 on t1.id = t2.id", []string{"nodeNames\":[\"1\"]", "Out: @1", "MergeJoiner"}},
+	}
+	for i, testCase := range testCases {
+		var res serverpb.QueryPlanResponse
+		queryParam := url.QueryEscape(testCase.query)
+		if err := getAdminJSONProto(s, fmt.Sprintf("queryplan?query=%s", queryParam), &res); err != nil {
+			t.Errorf("%d: got error %s", i, err)
+		}
+
+		for _, exp := range testCase.exp {
+			if !strings.Contains(res.DistSQLPhysicalQueryPlan, exp) {
+				t.Errorf("%d: expected response %s to contain %s", i, res, exp)
+			}
+		}
+	}
+
+}
+
 func TestAdminAPIRangeLog(t *testing.T) {
 	defer leaktest.AfterTest(t)()
 	s, _, _ := serverutils.StartServer(t, base.TestServerArgs{})


BramGruneir:
Reviewed 2 of 2 files at r1.

---

*[pkg/server/admin_test.go, line 1172 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXg9G5mX0Lkkd_7YGb:-KtXg9G5mX0Lkkd_7YGc:bmpsxqu) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1172)):*
> ```Go
> 	sqlDB.Exec(`CREATE TABLE api_test.query_plan_test (id int primary key, name string)`)
> 
> 	const invalidJobType = math.MaxInt32
> ```

What is this used for?

---

*[pkg/server/admin_test.go, line 1176 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgHD1Ukm0OuFh6YhW:-KtXgHD1Ukm0OuFh6YhX:by4tfx6) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1176)):*
> ```Go
> 	testCases := []struct {
> 		query        string
> 		expectedJSON string
> ```

might as well just call this expected

---

*[pkg/server/admin_test.go, line 1178 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgB3h3a_Cr45u4L2B:-KtXgB3h3a_Cr45u4L2C:b-r13m4n) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1178)):*
> ```Go
> 		expectedJSON string
> 	}{
> 		{"SELECT sum(id) FROM api_test.query_plan_test", ""},
> ```

Only one test case?  And one with out expected result?

---

*[pkg/server/admin_test.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgN-PxstlruyH8ttk:-KtXgN-PxstlruyH8ttl:b-4nwsy4) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1187)):*
> ```Go
> 		}
> 	}
> 	sqlDB.Exec(`DROP TABLE api_test.query_plan_test`)
> ```

hmm... perhaps put this in a defer?
And why not drop db too while we're at it?
And is this even needed at all?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372)*

BramGruneir:
Oh, and this should be in `ui:` at all.  It should be under `server:`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-:-KtXgaJm5Wt5KaILh2F-:b9oixfm)*

lego:
:thumbsup: 

---

Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/server/admin_test.go, line 1172 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXg9G5mX0Lkkd_7YGb:-KtXi8xJO3bdoYg-ixnK:b7zqn0h) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1172)):*

What is this used for?
</blockquote></details>

Fixed. That was something carried over from copying the `Job` tests.

---

*[pkg/server/admin_test.go, line 1176 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgHD1Ukm0OuFh6YhW:-KtXjMi7cZ2eQxHmBRSl:b-349zs1) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1176)):*

might as well just call this expected
</blockquote></details>

See below

---

*[pkg/server/admin_test.go, line 1178 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgB3h3a_Cr45u4L2B:-KtXiGS0hMU5mTkT_IPc:b-75ewpr) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1178)):*

Only one test case?  And one with out expected result?
</blockquote></details>

I wasn't sure what to test in addition to this, as it is DistSQL query plan results.

I've removed the expected bit outright because the result is JSON and it seems brittle to test for anything within in. What do you think? The JSON looks like:


```json
{"nodeNames":["1"],"processors":[{"nodeIdx":0,"inputs":[],"core":{"title":"TableReader","details":["primary@u","Render: 1:::INT"]},"outputs":[],"stage":1},{"nodeIdx":0,"inputs":[],"core":{"title":"Response","details":[]},"outputs":[],"stage":0}],"edges":[{"sourceProc":0,"sourceOutput":0,"destProc":1,"destInput":0}]}
```

---

*[pkg/server/admin_test.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgN-PxstlruyH8ttk:-KtXiRsa2jKpfijUQ6CI:b-b9peeq) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1187)):*

hmm... perhaps put this in a defer?
And why not drop db too while we're at it?
And is this even needed at all?
</blockquote></details>

Yep, you're right, turns out removing this outright is fine.

Although another test creates an `api_test` DB, so I'm surprised it's not broken. I don't know how these tests clean up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-:-KtXjFNsrcoKDOiSMIRQ:b4c9lyr)*

BramGruneir:
Reviewed 1 of 1 files at r2.

---

*[pkg/server/admin_test.go, line 1178 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgB3h3a_Cr45u4L2B:-KtY9lr_pB4bNeov82Br:b7i53bs) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1178)):*

I wasn't sure what to test in addition to this, as it is DistSQL query plan results.

I've removed the expected bit outright because the result is JSON and it seems brittle to test for anything within in. What do you think? The JSON looks like:


```json
{"nodeNames":["1"],"processors":[{"nodeIdx":0,"inputs":[],"core":{"title":"TableReader","details":["primary@u","Render: 1:::INT"]},"outputs":[],"stage":1},{"nodeIdx":0,"inputs":[],"core":{"title":"Response","details":[]},"outputs":[],"stage":0}],"edges":[{"sourceProc":0,"sourceOutput":0,"destProc":1,"destInput":0}]}
```
</blockquote></details>

There's no problem checking it against an expected string.  It does mean that if the query plan changes, this will have to be updated, but I don't think that's necessarily a bad thing, especially if these are simple queries.

You could also just look for some expected values, like nodeNames, processor, etc.

---

*[pkg/server/admin_test.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgN-PxstlruyH8ttk:-KtY9O5n3-2yQ7axTTRU:b-uel076) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1187)):*

Yep, you're right, turns out removing this outright is fine.

Although another test creates an `api_test` DB, so I'm surprised it's not broken. I don't know how these tests clean up.
</blockquote></details>

I'm pretty sure the whole cluster is completely torn down between each test's run.  
If you're worried, you could always add a uuid or something like that to the database, and table names. 

To make sure this isn't a problem, you should test it under stress for a good 10 mins, see if it fails.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372)*

lego:

---

*[pkg/server/admin_test.go, line 1178 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgB3h3a_Cr45u4L2B:-KuKzQa5QBjvjam3etEm:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1178)):*

There's no problem checking it against an expected string.  It does mean that if the query plan changes, this will have to be updated, but I don't think that's necessarily a bad thing, especially if these are simple queries.

You could also just look for some expected values, like nodeNames, processor, etc.
</blockquote></details>

Done.

---

*[pkg/server/admin_test.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgN-PxstlruyH8ttk:-KuKz_JHJcomKD56D0Pl:bb68bwn) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1187)):*

I'm pretty sure the whole cluster is completely torn down between each test's run.  
If you're worried, you could always add a uuid or something like that to the database, and table names. 

To make sure this isn't a problem, you should test it under stress for a good 10 mins, see if it fails.
</blockquote></details>

It seems no other tests that use the similar test utilities explicitly cleans up, so it should be fine.

Out of curiosity, how would you stress test this? Is there a command stress test specific tests?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372)*

BramGruneir:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---

*[pkg/server/admin_test.go, line 1187 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-KtXgN-PxstlruyH8ttk:-KuL4OgQQbJhyPTr-JSy:b-ufa9md) ([raw file](https://github.com/cockroachdb/cockroach/blob/71862790dfd56d09da6f94f3f21fa1cdbf926cb9/pkg/server/admin_test.go#L1187)):*

It seems no other tests that use the similar test utilities explicitly cleans up, so it should be fine.

Out of curiosity, how would you stress test this? Is there a command stress test specific tests?
</blockquote></details>

Take a look at`make stress`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18372#-:-KuL4gnw9t-UJqDBow_6:bnfp4nl)*

Issue #18159, opened 2017-09-01
server: Range debug page "there was an error retrieving the range log"


petermattis:
This is currently happening for every range I look at on `blue`:

![screen shot 2017-09-01 at 2 57 26 pm](https://user-images.githubusercontent.com/6201335/29983910-eb2caa9a-8f25-11e7-8cd8-e99de312509f.png)

That error message isn't particularly helpful as I'm not sure what node's logs to look at.

a-robinson:
It looks like the request is timing out before it completes. The range log on blue is quite large, and the endpoint powering that table tries to get the 100 latest entries involving the given range. Running the query involved by hand takes more than a minute.

https://github.com/cockroachdb/cockroach/issues/17781 would fix this, but there may be something wrong with the query plan/execution. Converting the query from doing a sort with a limit to just doing a `COUNT(*)` makes the query only take 2.5 seconds and indicates that there are only 151 rows, which definitely shouldn't be taking 1m20s to sort and limit.

Is it bad that filtering and sorting aren't showing up in the query plan below, or is that expected?

```sql
root@:26257/> SELECT * FROM system.rangelog where "rangeID" = 1 OR "otherRangeID" = 1 ORDER BY timestamp DESC LIMIT 100;
[...]
(100 rows)

Time: 1m20.377738093s

root@:26257/> EXPLAIN SELECT * FROM system.rangelog where "rangeID" = 1 OR "otherRangeID" = 1 ORDER BY timestamp DESC LIMIT 100;
+-------+---------+-------+------------------+
| Level |  Type   | Field |   Description    |
+-------+---------+-------+------------------+
|     0 | limit   |       |                  |
|     1 | revscan |       |                  |
|     1 |         | table | rangelog@primary |
|     1 |         | spans | ALL              |
+-------+---------+-------+------------------+
(4 rows)

Time: 14.666615ms

root@:26257/> SELECT COUNT(*) FROM system.rangelog where "rangeID" = 1 OR "otherRangeID" = 1;
+----------+
| count(*) |
+----------+
|      151 |
+----------+
(1 row)

Time: 2.41550115s
```

Skipping the sort makes things at least reasonably paced:

```
root@:26257/> SELECT * FROM system.rangelog where "rangeID" = 1 OR "otherRangeID" = 1 LIMIT 100;
[...]
(100 rows)

Time: 6.355918614s
```

a-robinson:
[...]

PR #18376, opened 2017-09-08
server, ui: Fix the range log history in the range debug page.

BramGruneir:
As reported in #18159, the range log query can take a long time to return. This solves this problem.

- The rangeLog has been removed from the range debug endpoint so we can rely entirely on the preexisting rangelog endpoint. This required augmenting the endpoint to return pretty printed strings for all elements of the info as this cannot be done on the javascript side.
- The range debug page now calls into this endpoint.
- The orderBy clause has been removed from the query for now, since it was slowing down the select.  This is currently being performed on the javascript side.
- The limit is specificly removed from the request originating from the range debug page, as it also was impacting the performance of the query.
- Since this new endpoint might still take a while to return, I gave the ajax call a rediculous amount of time before it times out and added the loading dialog to show that the request was still in progrss.
- Moved the range log and simulated allocator to below connections, since those connections are different and the range log might now be quite large and will typically take longer to render.
- Also, since some of the styles and fonts have changed, I updated the table's date column's width accordingly.


BramGruneir:
There are only very small changes to looks in this change, so I'm not going to add a before/after.  Just note that I've moved things around a bit on the range debug page, so that the elements that will take longer to populate, the range log and the allocator are now at the bottom.

a-robinson:
:lgtm_strong: 

---

Reviewed 17 of 17 files at r1.

---

*[pkg/server/admin.go, line 752 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmOpHv4CLUAVyfS4Mx:-KtmOpHv4CLUAVyfS4My:bcovxyu) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/admin.go#L752)):*
> ```Go
> 		q.Append(`WHERE "rangeID" = $ OR "otherRangeID" = $`, rangeID, rangeID)
> 	}
> 	// TODO(bram): Add this ordering back in after #18159 is resolved.
> ```

Is #18159 the issue you meant to point to? This PR will effectively resolve that issue, unless you plan to change what it's tracking.

---

*[pkg/server/admin.go, line 753 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmP-sX3cJw_97l1tF-:-KtmP-sX3cJw_97l1tF0:b-74vfj3) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/admin.go#L753)):*
> ```Go
> 	}
> 	// TODO(bram): Add this ordering back in after #18159 is resolved.
> 	// q.Append("ORDER BY timestamp DESC ")
> ```

I think we only want to remove the `ORDER BY` if there's no limit -- otherwise we'll be returning a fairly arbitrary set of records.

---

*[pkg/server/serverpb/admin.proto, line 465 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmPY1YchKzIZzpyMeG:-KtmPY1YchKzIZzpyMeH:b-5cqjqq) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/serverpb/admin.proto#L465)):*
> ```protobuf
>     PrettyInfo pretty_info = 2 [(gogoproto.nullable) = false];
>   }
>   reserved 1;
> ```

Nit, but personally I'd find it nice to also leave a comment indicating that this field has already been used in the past (as opposed to be reserved for some future use). Maybe that's just me, though.

---

*[pkg/server/serverpb/status.proto, line 499 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmPl6GuS0objFFCxgV:-KtmPl6HLpvO0cVEMZ8x:bs4q9op) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/serverpb/status.proto#L499)):*
> ```protobuf
>     (gogoproto.nullable) = false
>   ];
>   reserved 4;
> ```

Ditto.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-:-KtmR-BDiTLl12P6dOuk:b-uwkowd)*

BramGruneir:
Review status: 9 of 17 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/server/admin.go, line 752 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmOpHv4CLUAVyfS4Mx:-Ku0ie5ru5Jj72U2VzTC:bmgnkdo) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/admin.go#L752)):*

Is #18159 the issue you meant to point to? This PR will effectively resolve that issue, unless you plan to change what it's tracking.
</blockquote></details>

I was thinking this was a temporary fix and we should re-add this once `order by desc` works correctly.  

I've removed the comment, and if you ask for a limit, it now orders.

---

*[pkg/server/admin.go, line 753 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmP-sX3cJw_97l1tF-:-Ku0iP4iSNxq6DfuRIFB:bv3esr2) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/admin.go#L753)):*

I think we only want to remove the `ORDER BY` if there's no limit -- otherwise we'll be returning a fairly arbitrary set of records.
</blockquote></details>

hmm... good point.  I've added it back in, but only when the results are limited.

---

*[pkg/server/serverpb/admin.proto, line 465 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmPY1YchKzIZzpyMeG:-Ku0j64WEkn_QtyGuPBK:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/serverpb/admin.proto#L465)):*

Nit, but personally I'd find it nice to also leave a comment indicating that this field has already been used in the past (as opposed to be reserved for some future use). Maybe that's just me, though.
</blockquote></details>

Done.

---

*[pkg/server/serverpb/status.proto, line 499 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmPl6GuS0objFFCxgV:-Ku0jDNoArOhxVPkE7Q4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/serverpb/status.proto#L499)):*

Ditto.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376)*

a-robinson:
Reviewed 1 of 8 files at r2.
Review status: 10 of 17 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/server/admin.go, line 752 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmOpHv4CLUAVyfS4Mx:-Ku0skw4vvp6ZzGhC7NY:bpam5qc) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/admin.go#L752)):*

I was thinking this was a temporary fix and we should re-add this once `order by desc` works correctly.  

I've removed the comment, and if you ask for a limit, it now orders.
</blockquote></details>

`ORDER BY DESC` works correctly as of #18351, but it's still a lot slower than the unordered version (as tested on blue). So are we going to re-add it? If not now, when?

The improved speed of not doing the ordering is fantastic, so keeping out the ordering is nice. But I've also benefitted multiple times from being able to load the `/_admin/v1/rangelog` and see the events ordered, because having them ordered makes thrashing very obvious.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376)*

BramGruneir:
*[pkg/server/admin.go, line 752 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376#-KtmOpHv4CLUAVyfS4Mx:-Ku0vE6vZr4kTGYxAzHD:bikjbzi) ([raw file](https://github.com/cockroachdb/cockroach/blob/691e8c6382137ce14ec03dd68586c6046b1a0a2f/pkg/server/admin.go#L752)):*

`ORDER BY DESC` works correctly as of #18351, but it's still a lot slower than the unordered version (as tested on blue). So are we going to re-add it? If not now, when?

The improved speed of not doing the ordering is fantastic, so keeping out the ordering is nice. But I've also benefitted multiple times from being able to load the `/_admin/v1/rangelog` and see the events ordered, because having them ordered makes thrashing very obvious.
</blockquote></details>

Yep.  Going to leave this as is, and since the default still orders, I think we're good.  If the amount of data coming back to the ui starts to get too large to display in the dom, we can start limiting it in the ui as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18376)*

a-robinson:
Make sure to cherry pick this, please :)

BramGruneir:
Of course!

PR #18377, opened 2017-09-08
sqlmigrations,logictest: use placeholders for set cluster setting

### LINKED ISSUE #18224: sql: use placeholders where appropriate with SET CLUSTER SETTING
A TODO around running a `SET CLUSTER SETTING` is referencing the now-merged #17591.
https://github.com/cockroachdb/cockroach/blob/bff2545cf4de06b5335b8548555ff80c3404167c/pkg/sqlmigrations/migrations.go#L463

I have reason to believe that @knz intended to do something where @tschottdorf left such references, but I guess this one was missed.
17591 is also referenced in a logic test.

knz:
Fixes #18224.


tbg (approved):
Thanks @knz!

PR #18381, opened 2017-09-08
sqlccl: benchmark convertRecord on TPCH lineitem data

dt:
```
name             time/op
ConvertRecord-8    10.7µs ± 7%

name             speed
ConvertRecord-8  11.2MB/s ± 7%

name             alloc/op
ConvertRecord-8    1.83kB ± 0%

name             allocs/op
ConvertRecord-8      93.0 ± 0%
```


dt:
@mjibson switched to run multiple convert workers to more accurately match the real usage, RFAL.

madelynnblue (approved):

PR #18384, opened 2017-09-08
ui: network latency debug page bug fixes

BramGruneir:
- bring back styles for dead/stale nodes
- fix the react missing keys error

### DIFF
--- a/pkg/ui/src/views/reports/containers/network/index.tsx
+++ b/pkg/ui/src/views/reports/containers/network/index.tsx
@@ -112,8 +112,8 @@ function noConnectionTable(noConnections: NoConnection[]) {
               </td>
           </tr>
           {
-            _.map(noConnections, (noConn, k) => (
-              <tr className="failure-table__row" key={k}>
+            _.map(noConnections, (noConn) => (
+              <tr className="failure-table__row" key={`${noConn.from.nodeID}-${noConn.to.nodeID}`}>
                 <td className="failure-table__cell">
                   n{noConn.from.nodeID}
                 </td>
@@ -142,7 +142,7 @@ function noConnectionTable(noConnections: NoConnection[]) {
 }
 
 // createHeaderCell creates and decorates a header cell.
-function createHeaderCell(staleIDs: Set<number>, id: Identity, key: number) {
+function createHeaderCell(staleIDs: Set<number>, id: Identity, key: string) {
   const node = `n${id.nodeID.toString()}`;
   const title = _.join([node, id.address, id.locality], "\n");
   let className = "network-table__cell network-table__cell--header";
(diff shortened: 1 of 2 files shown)


vilterp:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384#-KtYmfuFQ6kiXeQEeVIa:-KtYmfuFQ6kiXeQEeVIb:b-w7msgj) ([raw file](https://github.com/cockroachdb/cockroach/blob/002e63d316285beccc9003085d2300badf1c7e87/pkg/ui/src/views/reports/containers/network/index.tsx#L61)):*
> ```XML
>           {
>             _.map(staleIdentities, (staleIdentity, key) => (
>               <tr className="failure-table__row" key={key}>
> ```

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318); using the node id seems better

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 200 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384#-KtYn2_r7yOJqjqQlyBL:-KtYn2_r7yOJqjqQlyBM:b-47fafm) ([raw file](https://github.com/cockroachdb/cockroach/blob/002e63d316285beccc9003085d2300badf1c7e87/pkg/ui/src/views/reports/containers/network/index.tsx#L200)):*
> ```XML
>     // getLatencyCell creates and decorates a cell based on it's latency.
>     function getLatencyCell(nodeIDa: number, nodeIDb: number) {
>       const key = `${nodeIDa}-${nodeIDb}`;
> ```

looks good

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384)*

vilterp:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384#-KtYmfuFQ6kiXeQEeVIa:-KtYnSFo4ReO7HhW1A_O:by7fj5) ([raw file](https://github.com/cockroachdb/cockroach/blob/002e63d316285beccc9003085d2300badf1c7e87/pkg/ui/src/views/reports/containers/network/index.tsx#L61)):*

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318); using the node id seems better
</blockquote></details>

sorry, first time using reviewable… Meant to say:

Why was this change necessary? Using an index as a key [can cause bugs](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318); using node id seems better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384)*

vilterp:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384)*

BramGruneir:
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384#-KtYmfuFQ6kiXeQEeVIa:-KtYxHloUJ4j31dPHUZP:bs90akj) ([raw file](https://github.com/cockroachdb/cockroach/blob/002e63d316285beccc9003085d2300badf1c7e87/pkg/ui/src/views/reports/containers/network/index.tsx#L61)):*

sorry, first time using reviewable… Meant to say:

Why was this change necessary? Using an index as a key [can cause bugs](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318); using node id seems better.
</blockquote></details>

It wasn't.  Left over from my investigation.  Reverting.  And fixed the other one.

I've read that article, and calling it an anti-pattern is a bit harsh.  But it should be avoided if possible.  But this depends on what you're displaying.  These keys are used for redrawing, and as of right now, the debug pages have to be manually refreshed, so there's no real advantage.  That being said, it still is a good idea to use a more descriptive key, even if it just future proofs it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384)*

vilterp:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18384#-:-KtZS43tmMZX9aS7O1mU:bnfp4nl)*

BramGruneir:
TFTR!

PR #18385, opened 2017-09-08
cherrypick-1.0: kv: fix (the horribly broken) prev()

tbg:
kv: fix (the horribly broken) prev()

prev() was broken in two ways:

- it had a one-off where `prev([a, b), b)` would  return `b` instead of `KeyMin`.
- in one path it was assembling the return from the start keys, not the end keys, which is
  wrong. When keys `>= c` have been handled, and your remaining batch is `[a, b)`, it would
  erroneously return `a` instead of `b`.

Note on the cherry-pick: I had to rip out the newly added test that exercised the "truncation resulted in empty batch" path since it uses features not available in 1.0 (testing knobs, etc).

cc @cockroachdb/release


bdarnell (approved):

bdarnell:
Merging for inclusion in 1.0.6.

PR #18390, opened 2017-09-09
[DNM] sql: allow configuring rounding mode for decimals

madelynnblue:
This is a proof-of-concept to see if we think this is a good idea. It allows for users to configure how decimals work, since that's a feature we can provide. The unfortunate part about this is there's a really high amount of plumbing for a tiny benefit and pretty niche feature, so it barely seems worth it. We do the same thing for time zones and some other stuff, including copying all this stuff around into distsql (which this PR doesn't support yet).

Do we think these kinds of features are worth it?

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/builtin_function
+++ b/pkg/sql/logictest/testdata/logic_test/builtin_function
@@ -778,6 +778,35 @@ SELECT round('inf'::decimal)
 query error invalid operation
 SELECT round(1::decimal, 3000)
 
+# Test rounding modes.
+
+statement error unknown rounding: blah
+SET decimal_round = 'blah'
+
+statement ok
+SET decimal_round = 'DOWN'
+
+query RRRR
+SELECT round(-2.5::decimal), round(-1.5::decimal), round(1.5::decimal), round(2.5::decimal)
+----
+-2  -1  1  2
+
+statement ok
+SET decimal_round = 'HALF_EVEN'
+
+query RRRR
+SELECT round(-2.5::decimal), round(-1.5::decimal), round(1.5::decimal), round(2.5::decimal)
+----
+-2  -2  2  2
+
+statement ok
+RESET decimal_round
+
+query RRRR
+SELECT round(-2.5::decimal), round(-1.5::decimal), round(1.5::decimal), round(2.5::decimal)
+----
+-3  -2  2  3
+
 query III
 SELECT sign(-2), sign(0), sign(2)
 ----
--- a/pkg/sql/parser/builtins.go
+++ b/pkg/sql/parser/builtins.go
@@ -1380,14 +1380,19 @@ CockroachDB supports the following flags:
 		floatBuiltin1(func(x float64) (Datum, error) {
 			return NewDFloat(DFloat(round(x))), nil
 		}, "Rounds `val` to the nearest integer using half to even (banker's) rounding."),
-		decimalBuiltin1(func(x *apd.Decimal) (Datum, error) {
-			return roundDecimal(x, 0)
-		}, "Rounds `val` to the nearest integer, half away from zero: "+
-			"ROUND(+/-2.4) = +/-2, ROUND(+/-2.5) = +/-3."),
+		Builtin{
+			Types:      ArgTypes{{"input", TypeDecimal}},
+			ReturnType: fixedReturnType(TypeDecimal),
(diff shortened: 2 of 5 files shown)


bdarnell:
> Do we think these kinds of features are worth it?

I'd say "probably not", but I'm willing to be convinced. Is anyone asking for this feature? Do other databases implement it?

Is a more self-contained implementation possible? (for example, adding an optional `mode` argument to the `round()` function instead of a session variable) Session variables have a significant plumbing cost (both now and in the future), and for one whose effect is this subtle it's easy to omit one without realizing it. Implementing this feature means committing to keep it well-tested as the system changes. Implementing it with caveats like "doesn't work in distsql" is probably worse than not having it at all.

knz:
So I'm mildly open to implementing this, but then not using a conditional checked on every call to the round() function. Instead, the overload resolution should select one or another implementation of round depending on the behavior-modifying parameter.

The idea of a session variable is actually sound, and so would be the solution to introduce two (or more) differently named round functions side-by-side (i.e. not a single function taking a mode parameter). 

In terms of implementation overhead, perhaps the better solution for now is differently named builtin functions side by side. If/when users say "I can't change my application code *and* I want a different rounding mode" we can revisit.

mjibson:
Well maybe this specific example of round is a bad one. (Also ignore the lack of DistSQL implementation, I'm not worried about doing that and would, I just wanted to have a discussion before doing that work.)

Consider instead changing the internal precision of decimal operations. Currently decimal operations are rounded to 20 digits, which isn't that many more than floats. What if someone wanted many more? Having separate functions for all of these would not be a good way to implement it: a session setting would be much better. I haven't seen anyone ask for this specifically yet, but I am curious about this feature in other DBs.

dianasaur323:
perhaps either @nstewart or andy when he gets in could investigate this for you.

knz:
I believe the precision is a different matter because it doesn't require a different code path. The issue I had with the proposed code is the introduction of a conditional on the compute path of the function. The precision would use the same compute path just with a different input, so that would be entirely kosher for me.

nstewart:
> Do we think these kinds of features are worth it? 
I'd agree with Ben on "probably not" (especially if you think there is a high cost and low reward). There is also the usability question Ben mentioned around making sure it works without a lot of caveats. 

@mjibson -- For my understanding, what is the real-world problem this solves?

knz:
@nstewart rounding to some fixed number of decimals after a computation is essential in finance and science. 

In finance people usually require round to be half-up.

In science instead most common is round half-to-even to eliminate numeric biases, but actually different numerical domains call for different rounding modes.

A general-purpose tool like SQL which aims to cater to a variety of numerical applications should offer at least both.

https://en.wikipedia.org/wiki/Rounding

nstewart:
Thanks @knz -- surprised it hasn't come up from any financial customers yet (to my knowledge). I think the main question is going to be the high-cost low reward that Matt and Ben discussed. If we want to spend time rolling it out we'll need to treat it like any other feature cc @awoods187

knz:
Yes we currently implement half-up if I'm not mistaken so that's appropriate for finance. @mjibson can you confirm?

bdarnell:
> In finance people usually require round to be half-up.

Is that true? Half-to-even rounding is also called "banker's rounding". 

From a brief search, it looks like [most major databases](https://www.postgresql.org/message-id/CAEZATCWGaeLoHdsWk7Yz4jFMFOH8f%3DcapC6SjMa0HYBK5_xu5Q%40mail.gmail.com) default to half-away-from-zero, and few of them offer an option to change this, so there doesn't seem to be much demand for it. (maybe people who need half-to-even rounding do their rounding outside of SQL?)

I'd be fine with adding a set of functions like `round_half_even()` to support rounding with an explicit mode. I'd rather not add session settings for decimal options until there's demand for it (but if people are asking for it then a session variable seems like the appropriate way to change decimal precision).

mjibson:
Float uses half up and decimal uses half even (banker's rounding). We were pretty surprised when we found out that the types determined the rounding mode in postgres. I think that for rounding it makes sense to add some new round overrides that have an additional argument, which is a string for the rounding mode. For precision a session setting is the best for sure.

However I agree with the other comments here that I'm dubious about the use of these. I kind of want to talk to a bank or two and a few scientific computing folks to see if there's features like this that would help them out. There is some teeny tiny remote chance that we could position distsql to be a distributed computation framework if we addressed these kinds of issues.

PR #18397, opened 2017-09-09
cherrypick-1.0: storage/engine: reproduce RocksDB DeleteRange bug

### LINKED ISSUE #18371: c-deps: Investigate RocksDB DeleteRange bug
The TiDB team found a data corruption bug in RocksDB ([blog post](https://pingcap.github.io/blog/2017/09/08/rocksdbbug/), [issue](https://github.com/facebook/rocksdb/issues/2752), [PR](https://github.com/facebook/rocksdb/pull/2799)). We need to determine whether we are affected by this; if so we need to patch in the fix for 1.1. 

The bug relates to the `DeleteRange` operation and snapshots; we use both features and based on the test case in facebook/rocksdb#2799 I don't see any reason to assume that we're immune. The symptom is data that was deleted by DeleteRange could reappear.

petermattis:
Manually apply the fix to our RocksDB repo. The presence of the test
will prevent upgrading to a version of RocksDB that does not contain the
fix.

Fixes #18371


petermattis:
Turns out the patch did apply cleanly, I was just doing it wrong.

bdarnell (approved):

petermattis:
Feel free to merge on green. I may not be online then.

PR #18398, opened 2017-09-09
sql/ir: use a special character to highlight metavars in templates

knz:


petermattis:
The use of the special character is interesting, though I'm not sure if the need is compelling enough to justify the usage. A unique prefix would also suffice.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18398#-:-KtvslT_0AEN6ine3AbR:bl6p377)*

knz:
Do you have a better suggestion?

petermattis:
Does the template code need to be valid go code? If not, I'd use something like `<numRefsPerNode>`. If the template code does need to be valid go code, perhaps something like `META_numRefsPerNode`. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18398#-:-Ku4zheZZqsoqtgFlERA:bz2atx1)*

justinj:
I can see an argument against having "normal" characters given that metavariables could occur anywhere, including in strings. Does seem like a bit of a pain to work with though... I don't have a suggestion for something that would solve both problems short of moving to a homoiconic language :) so I'll give a weak +1 to using the unicode character, personally.

knz:
Superseded by #19135.

PR #18406, opened 2017-09-10
storage: clear Entry before reuse

tbg:
See:

https://github.com/cockroachdb/cockroach/issues/18324#issuecomment-328364081

Touches #18324.

### DIFF
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -3429,6 +3429,7 @@ func sendSnapshot(
 	{
 		var ent raftpb.Entry
 		for i := range logEntries {
+			ent.Reset()
 			if err := ent.Unmarshal(logEntries[i]); err != nil {
 				return err
 			}


pkg/storage/store.go
```diff
@@ -3429,6 +3429,7 @@ func sendSnapshot(
 	{
 		var ent raftpb.Entry
 		for i := range logEntries {
+			ent.Reset()
```
petermattis: `ent.Unmarshal` resets internally. Doesn't it?

tschottdorf: No, see the discussion linked to in this PR. Unless I really messed up both the unit test there and my code inspection.

RaduBerinde:
LGTM

pkg/storage/store.go
petermattis: Oh, it doesn't! That's surprising. Is that a bug in gogoproto? I'm pretty sure `proto.Unmarshal` clears the destination.

petermattis (approved):
LGTM

tschottdorf:
Yes, `proto.Unmarshal` does and that's also what I looked at first. I'm not sure this is a bug. I'll definitely file something to investigate. This is likely an issue elsewhere.

petermattis:
`git grep '\.Unmarshal(' | grep -v _test.go | grep -v 'proto.Unmarshal(' | grep -v "Binary file" | grep -v 'jsonpb.Unmarshal(' | grep -v '^cmd/'` reveals 29 calls to `\.Unmarshal(`. A quick inspection doesn't show any of them to be problematic because they are all unmarshaling on pristine variables. Of course, I could have missed something. 

Pretty sure this is a bug in gogoproto. From the `proto.Unmarshal` docs:

```
// Unmarshal resets pb before starting to unmarshal, so any
// existing data in pb is always removed. Use UnmarshalMerge
// to preserve and append to existing data.
```

tschottdorf:
Yep, that's the  comment that led me to believe everything was fine until I wrote the unit test. Filed https://github.com/cockroachdb/cockroach/issues/18407 for you (which links to an upstream issue).

PR #18425, opened 2017-09-11
storage: immediately update target store write stats after rebalance

### LINKED ISSUE #17970: storage: Delayed update of per-store write stats can cause rebalance thrashing
After a rebalance has happened, the `LogicalBytes` and `WritesPerSecond` logically change for the store that the replica was added to or removed from. If the `LogicalBytes` stat changes by a large amount, then the store will re-gossip its `StoreCapacity` ahead of schedule. While it isn't guaranteed by any means, in practice this does a mostly decent job of getting updated information spread throughout the cluster quickly enough that a bunch more rebalance operations aren't based on outdated information.

The same is not true for `WritesPerSecond`, though. We don't start counting `WritesPerSecond` stats until a replica has been on a new store for 5 seconds, so a good deal more rebalancing decisions can be made without considering the additional writes on the node. In some circumstances (as seen on indigo), this can make for rebalance thrashing where we add a replica to a node, then decide that the replica isn't a good fit for it. If the store's `WritesPerSecond` stat had been updated, we wouldn't move the replica.

As seen on indigo, this was closely intertwined with #17971.

@BramGruneir has previously suggested passing along `WritesPerSecond` stats as part of a rebalance to combat this. I'm still not sure about that, but the leaseholder that made the rebalance decisions should be capable of updating its own local copy of the other store's descriptor so that it has a more accurate view when deciding which replica to remove as part of a rebalance.

a6802739:
@a-robinson, Fixes #17970.

I just update the local copy for target store after the rebalance immediately.

still don't know how to add a test for it.

Please help me review it. Thank you very much.

### DIFF
--- a/pkg/storage/replicate_queue.go
+++ b/pkg/storage/replicate_queue.go
@@ -572,7 +572,11 @@ func (rq *replicateQueue) addReplica(
 	if dryRun {
 		return nil
 	}
-	return repl.changeReplicas(ctx, roachpb.ADD_REPLICA, target, desc, priority, reason, details)
+	if err := repl.changeReplicas(ctx, roachpb.ADD_REPLICA, target, desc, priority, reason, details); err != nil {
+		return err
+	}
+	rq.allocator.storePool.updateLocalStoreAfterRebalance(target.StoreID, repl, roachpb.ADD_REPLICA)
+	return nil
 }
 
 func (rq *replicateQueue) removeReplica(
@@ -587,7 +591,11 @@ func (rq *replicateQueue) removeReplica(
 	if dryRun {
 		return nil
 	}
-	return repl.ChangeReplicas(ctx, roachpb.REMOVE_REPLICA, target, desc, reason, details)
+	if err := repl.ChangeReplicas(ctx, roachpb.REMOVE_REPLICA, target, desc, reason, details); err != nil {
+		return err
+	}
+	rq.allocator.storePool.updateLocalStoreAfterRebalance(target.StoreID, repl, roachpb.REMOVE_REPLICA)
+	return nil
 }
 
 func (rq *replicateQueue) canTransferLease() bool {
--- a/pkg/storage/store_pool.go
+++ b/pkg/storage/store_pool.go
@@ -327,6 +327,38 @@ func (sp *StorePool) deadReplicasGossipUpdate(_ string, content roachpb.Value) {
 	detail.deadReplicas = deadReplicas
 }
 
+// updateLocalStoreAfterRebalance is used to update local copy of target store
+// after we make an rebalance immediately.
+func (sp *StorePool) updateLocalStoreAfterRebalance(
+	storeID roachpb.StoreID, repl *Replica, changeType roachpb.ReplicaChangeType,
+) {
+	sp.detailsMu.Lock()
+	defer sp.detailsMu.Unlock()
+	detail := *sp.getStoreDetailLocked(storeID)
+	switch changeType {
+	case roachpb.ADD_REPLICA:
+		detail.desc.Capacity.LogicalBytes += repl.GetMVCCStats().Total()
+		if qps, dur := repl.writeStats.avgQPS(); dur >= MinStatsDuration {
+			detail.desc.Capacity.WritesPerSecond += qps
+		}
+	case roachpb.REMOVE_REPLICA:
+		total := repl.GetMVCCStats().Total()
+		if detail.desc.Capacity.LogicalBytes <= total {
+			detail.desc.Capacity.LogicalBytes = 0
+		} else {
+			detail.desc.Capacity.LogicalBytes -= total
+		}
+		if qps, dur := repl.writeStats.avgQPS(); dur >= MinStatsDuration {
+			if detail.desc.Capacity.WritesPerSecond <= qps {
+				detail.desc.Capacity.WritesPerSecond = 0
+			} else {
+				detail.desc.Capacity.WritesPerSecond -= qps
+			}
+		}
+	}
+	sp.detailsMu.storeDetails[storeID] = &detail
+}
+
 // newStoreDetail makes a new storeDetail struct. It sets index to be -1 to
 // ensure that it will be processed by a queue immediately.
 func newStoreDetail() *storeDetail {
--- a/pkg/storage/store_pool_test.go
+++ b/pkg/storage/store_pool_test.go
@@ -21,6 +21,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/cockroachdb/cockroach/pkg/storage/engine"
+	"github.com/cockroachdb/cockroach/pkg/storage/engine/enginepb"
+
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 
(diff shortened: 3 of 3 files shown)


pkg/storage/store_pool.go
```diff
@@ -327,6 +327,20 @@ func (sp *StorePool) deadReplicasGossipUpdate(_ string, content roachpb.Value) {
...
+// updateLocalStoreAfterRebalance is used to update local copy of target store
+// after we make an rebalance immediately.
+func (sp *StorePool) updateLocalStoreAfterRebalance(storeID roachpb.StoreID, repl *Replica) {
+	sp.detailsMu.Lock()
+	defer sp.detailsMu.Unlock()
+	detail := sp.getStoreDetailLocked(storeID)
+	// update LogicalBytes.
+	detail.desc.Capacity.LogicalBytes += repl.GetMVCCStats().Total()
+	// update WritesPerSecond
+	qps, _ := repl.writeStats.avgQPS()
```
a-robinson: Only use the qps returned if the duration returned is greater than `MinStatsDuration`

pkg/storage/store_pool.go
```diff
@@ -327,6 +327,20 @@ func (sp *StorePool) deadReplicasGossipUpdate(_ string, content roachpb.Value) {
 	detail.deadReplicas = deadReplicas
 }
 
+// updateLocalStoreAfterRebalance is used to update local copy of target store
+// after we make an rebalance immediately.
+func (sp *StorePool) updateLocalStoreAfterRebalance(storeID roachpb.StoreID, repl *Replica) {
+	sp.detailsMu.Lock()
+	defer sp.detailsMu.Unlock()
+	detail := sp.getStoreDetailLocked(storeID)
```
a-robinson: I think you'll have to make a copy here and reinsert the copy into the `StorePool` rather than modifying the returned detail in place, since there could be other parts of the code accessing the detail without`sp.detailsMu` held.

pkg/storage/store_pool.go
```diff
@@ -327,6 +327,20 @@ func (sp *StorePool) deadReplicasGossipUpdate(_ string, content roachpb.Value) {
...
+	// update LogicalBytes.
+	detail.desc.Capacity.LogicalBytes += repl.GetMVCCStats().Total()
+	// update WritesPerSecond
+	qps, _ := repl.writeStats.avgQPS()
+	detail.desc.Capacity.WritesPerSecond += qps
+	detail.lastUpdatedTime = sp.clock.PhysicalTime()
```
a-robinson: I don't think we should actually update this, since it would mess up calculations of whether the store is live or dead. This is only meant to be modified when new gossip is received from the store.

pkg/storage/replicate_queue.go
```diff
@@ -508,6 +508,7 @@ func (rq *replicateQueue) processOneChange(
 				); err != nil {
 					return false, err
 				}
+				rq.allocator.storePool.updateLocalStoreAfterRebalance(rebalanceReplica.StoreID, repl)
```
a-robinson: We should just call this from within `addReplica` and `removeReplica` so that we always have more accurate estimates. That'll mean modifying `updateLocalStoreAfterRebalance` to support removing stats as well as adding them.

a6802739: @a-robinson  Do you mean we should add the `WritesPerSecond` when we `addReplica`, and we should minus the `WritesPerSecond` when we `removeReplica`?

pkg/storage/store_pool.go
a6802739: Why should we guarantee this?Our intent is to immediately update the write stats, right?

pkg/storage/store_pool.go
a6802739: So we should do it like: `detail := *sp.getStoreDetailLocked(storeID)`, then we update the `detail` and reinsert it into  the `StorePool`?

pkg/storage/store_pool.go
a6802739: My mistake.

pkg/storage/replicate_queue.go
a-robinson: Yeah, that's what I was thinking.

pkg/storage/store_pool.go
a-robinson: Yes, something with those semantics. I'm worried it may create data races to modify the `detail` object in place.

pkg/storage/store_pool.go
a-robinson: The reason for the check is that taking an average over a very small duration makes large errors pretty likely. I saw issues in my testing that were caused by using the stats returned shortly after a replica was initialized, because the stats returned were often significantly higher than their true steady state value.

There may be other ways to get around this, so suggestions are appreciated. Otherwise, though, I think it's a reasonable practice to wait at least a couple seconds before using these stats.

In almost all cases the duration will have been long enough here, since this code is only executing on the leaseholder.

pkg/storage/replicate_queue.go
a6802739: Done.

pkg/storage/store_pool.go
a6802739: Done.

pkg/storage/store_pool.go
a6802739: Done.

pkg/storage/store_pool.go
```diff
@@ -327,6 +327,29 @@ func (sp *StorePool) deadReplicasGossipUpdate(_ string, content roachpb.Value) {
...
+	defer sp.detailsMu.Unlock()
+	detail := *sp.getStoreDetailLocked(storeID)
+	switch changeType {
+	case roachpb.ADD_REPLICA:
+		detail.desc.Capacity.LogicalBytes += repl.GetMVCCStats().Total()
+		if qps, dur := repl.writeStats.avgQPS(); dur >= MinStatsDuration {
+			detail.desc.Capacity.WritesPerSecond += qps
+		}
+	case roachpb.REMOVE_REPLICA:
+		detail.desc.Capacity.LogicalBytes -= repl.GetMVCCStats().Total()
```
a-robinson: Nit, but it'd be nice to make sure we aren't reducing these values to less than 0 here. It's very unlikely, but is possible for us to have already received updated gossip from the store which would technically make it possible for subtracting these values to end up at negative values.

a-robinson (commented):
Nice! Now that the logic is nicely segregated, adding a unit test to `ReplicateQueue` or to `StorePool` shouldn't be too hard.

pkg/storage/store_pool.go
a6802739: Done.

a6802739:
@a-robinson , Already add a test for this PR, I don't if it's appropriate.

pkg/storage/store_pool_test.go
```diff
@@ -318,6 +321,101 @@ func TestStorePoolGetStoreList(t *testing.T) {
...
+			},
+		},
+	}
+	sg.GossipStores(stores, t)
+	node := roachpb.NodeDescriptor{NodeID: roachpb.NodeID(1)}
+	eng := engine.NewInMem(roachpb.Attributes{}, 1<<20)
+	stopper.AddCloser(eng)
+	cfg := TestStoreConfig(clock)
+	cfg.Transport = NewDummyRaftTransport(cfg.Settings)
+	store := NewStore(cfg, eng, &node)
```
a-robinson: It seems like some of this setup could be skipped (e.g. it doesn't look like we use store 3 for anything; do we really need to create a real engine and store for the replica that never has to actually do anything?).

a-robinson (approved):
LGTM

pkg/storage/store_pool_test.go
a6802739: I have removed store 3.
If we  didn't create a real engine and store, when we want to create a replica with `NewReplica()`, it will always coredump.

a-robinson (approved):
Looks good! Feel free to merge

PR #18426, opened 2017-09-11
storage : disable stats-based rebalancing when we use SCATTER

### LINKED ISSUE #17671: storage: stats-based rebalancing can't handle large numbers of splits and scatters
Discovered in #17644. Here's a RESTORE with stats-based rebalancing enabled:

screenshot 2017-08-15 13 20 53

And without:

screenshot 2017-08-15 12 50 10

a6802739:
@a-robinson , Fixes #17671.
I just disable stats-based rebalancing when we use SQL statement `ALTER TABLE ... SCATTER ...`
Still don't know how to add a test for it. Thank you for your guidance.
Please help me review this.

### DIFF
--- a/pkg/storage/allocator_scorer.go
+++ b/pkg/storage/allocator_scorer.go
@@ -77,8 +77,8 @@ var statRebalanceThreshold = settings.RegisterNonNegativeFloatSetting(
 	0.20,
 )
 
-func statsBasedRebalancingEnabled(st *cluster.Settings) bool {
-	return EnableStatsBasedRebalancing.Get(&st.SV) && st.Version.IsActive(cluster.VersionStatsBasedRebalancing)
+func statsBasedRebalancingEnabled(st *cluster.Settings, disableStatsBasedRebalance bool) bool {
+	return EnableStatsBasedRebalancing.Get(&st.SV) && st.Version.IsActive(cluster.VersionStatsBasedRebalancing) && !disableStatsBasedRebalance
 }
 
 type balanceDimensions struct {
@@ -307,6 +307,7 @@ func allocateCandidates(
 	rangeInfo RangeInfo,
 	existingNodeLocalities map[roachpb.NodeID]roachpb.Locality,
 	deterministic bool,
+	disableStatsBasedRebalance bool,
 ) candidateList {
 	var candidates candidateList
 	for _, s := range sl.stores {
@@ -321,7 +322,7 @@ func allocateCandidates(
 			continue
 		}
 		diversityScore := diversityScore(s, existingNodeLocalities)
-		balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo)
+		balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance)
 		candidates = append(candidates, candidate{
 			store:           s,
 			valid:           true,
@@ -349,6 +350,7 @@ func removeCandidates(
 	rangeInfo RangeInfo,
 	existingNodeLocalities map[roachpb.NodeID]roachpb.Locality,
 	deterministic bool,
+	disableStatsBasedRebalance bool,
 ) candidateList {
 	var candidates candidateList
 	for _, s := range sl.stores {
@@ -370,9 +372,9 @@ func removeCandidates(
 			continue
 		}
 		diversityScore := diversityRemovalScore(s.Node.NodeID, existingNodeLocalities)
-		balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo)
+		balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance)
 		var convergesScore int
-		if !rebalanceFromConvergesOnMean(st, sl, s.Capacity, rangeInfo) {
+		if !rebalanceFromConvergesOnMean(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance) {
 			// If removing this candidate replica does not converge the store
 			// stats to their means, we make it less attractive for removal by
 			// adding 1 to the constraint score. Note that when selecting a
@@ -411,6 +413,7 @@ func rebalanceCandidates(
 	rangeInfo RangeInfo,
 	existingNodeLocalities map[roachpb.NodeID]roachpb.Locality,
 	deterministic bool,
+	disableStatsBasedRebalance bool,
 ) (candidateList, candidateList) {
 	// Load the exiting storesIDs into a map to eliminate having to loop
 	// through the existing descriptors more than once.
@@ -446,7 +449,7 @@ func rebalanceCandidates(
 	if !rebalanceConstraintsCheck {
 		for _, store := range sl.stores {
 			if _, ok := existingStoreIDs[store.StoreID]; ok {
-				if shouldRebalance(ctx, st, store, constraintsOkStoreList, rangeInfo) {
+				if shouldRebalance(ctx, st, store, constraintsOkStoreList, rangeInfo, disableStatsBasedRebalance) {
 					shouldRebalanceCheck = true
 					break
 				}
@@ -484,9 +487,9 @@ func rebalanceCandidates(
 				continue
 			}
 			diversityScore := diversityRemovalScore(s.Node.NodeID, existingNodeLocalities)
-			balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo)
+			balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance)
 			var convergesScore int
-			if !rebalanceFromConvergesOnMean(st, sl, s.Capacity, rangeInfo) {
+			if !rebalanceFromConvergesOnMean(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance) {
 				// Similarly to in removeCandidates, any replica whose removal
 				// would not converge the range stats to their means is given a
 				// constraint score boost of 1 to make it less attractive for
@@ -506,9 +509,9 @@ func rebalanceCandidates(
 			if !storeInfo.ok || !maxCapacityOK {
 				continue
 			}
-			balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo)
+			balanceScore := balanceScore(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance)
 			var convergesScore int
-			if rebalanceToConvergesOnMean(st, sl, s.Capacity, rangeInfo) {
+			if rebalanceToConvergesOnMean(st, sl, s.Capacity, rangeInfo, disableStatsBasedRebalance) {
 				// This is the counterpart of !rebalanceFromConvergesOnMean from
 				// the existing candidates. Candidates whose addition would
 				// converge towards the range count mean are promoted.
@@ -552,6 +555,7 @@ func shouldRebalance(
 	store roachpb.StoreDescriptor,
 	sl StoreList,
 	rangeInfo RangeInfo,
+	disableStatsBasedRebalance bool,
 ) bool {
 	if store.Capacity.FractionUsed() >= maxFractionUsedThreshold {
 		log.VEventf(ctx, 2, "s%d: should-rebalance(disk-full): fraction-used=%.2f, capacity=(%v)",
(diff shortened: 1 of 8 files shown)


a-robinson (commented):
Before worrying about how to write a unit test for this, have you tested it manually? That could be done, for example, by splitting a table into hundreds of ranges and then calling scatter on them. Without any code changes, we'd expect little rebalancing to be done. With correct code changes, we'd expect the range counts on each node to all be roughly equal.

I don't think this change will do what we want. The `Settings` object here doesn't get passed down to the replicas that process the `adminScatter` request. I think we'll need `adminScatter` to pass a custom settings object into the replicate queue.

petermattis:
> I don't think this change will do what we want. The Settings object here doesn't get passed down to the replicas that process the adminScatter request. I think we'll need adminScatter to pass a custom settings object into the replicate queue.

Do we want to be passing a custom `Settings` object to `replicateQueue`? I was imagining there would be some sort of `replicateQueueOptions` structure we'd pass. Perhaps that is what you meant be a custom settings object.

a-robinson:
> Do we want to be passing a custom Settings object to replicateQueue? I was imagining there would be some sort of replicateQueueOptions structure we'd pass. Perhaps that is what you meant be a custom settings object.

That'd be fine as well. It'll just have to override the existing `Settings` object where it's currently used within the Allocator.

a6802739:
@a-robinson , I just change the logic now and did a test manually.

Here is the result after I run `ALTER TABLE sh_test SCATTER;`:

7e9ec4e68c6d99a3e3d9a810920cf490

And here is the cluster setting for `kv.allocator.stat_based_rebalancing.enabled`:

281b31e0c64ed577abcd53fb5ce9c0dc

If this PR is in a right direction, I will try to add a test then.

pkg/roachpb/api.proto
```diff
@@ -943,6 +943,7 @@ message AdminScatterRequest {
   option (gogoproto.equal) = true;
 
   optional Span header = 1 [(gogoproto.nullable) = false, (gogoproto.embed) = true];
+  optional bool disableStatsBasedRebalance = 2 [(gogoproto.nullable) = false];
```
a-robinson: Is there a reason the API change is needed? The implementation of `AdminScatter` can just always disable stats-based rebalancing internally.

pkg/storage/allocator_scorer.go
```diff
@@ -541,19 +544,20 @@ func shouldRebalance(
...
+	disableStatsBasedRebalance bool,
 ) bool {
 	if store.Capacity.FractionUsed() >= maxFractionUsedThreshold {
 		log.VEventf(ctx, 2, "s%d: should-rebalance(disk-full): fraction-used=%.2f, capacity=(%v)",
 			store.StoreID, store.Capacity.FractionUsed(), store.Capacity)
 		return true
 	}
 
-	if !statsBasedRebalancingEnabled(st) {
+	if !statsBasedRebalancingEnabled(st) || disableStatsBasedRebalance {
```
a-robinson: All these different conditionals would be less error prone if we just passed `disableStatsBasedRebalancing` into `statsBasedRebalancingEnabled`.

pkg/storage/allocator_test.go
```diff
@@ -659,6 +667,7 @@ func TestAllocatorRebalance(t *testing.T) {
 			config.Constraints{},
 			testRangeInfo([]roachpb.ReplicaDescriptor{{StoreID: 3}}, firstRange),
 			storeFilterThrottled,
+			true,
```
a-robinson: We should just be able to pass `false` into all of these test call sites, since the `Settings` objects are already configured properly.

pkg/storage/replicate_queue.go
```diff
@@ -94,6 +98,7 @@ func makeReplicateQueueMetrics() ReplicateQueueMetrics {
 // additional replica to their range.
 type replicateQueue struct {
 	*baseQueue
+	options           replicateQueueOptions
```
a-robinson: I don't think we'll ever want this field to be set, so I'd just remove it.

pkg/storage/replica_command.go
```diff
@@ -4020,6 +4020,7 @@ func (r *Replica) adminScatter(
 		Multiplier:     2,
 		MaxRetries:     5,
 	}
+	rq.options.disableStatsBasedRebalance = args.DisableStatsBasedRebalance
```
a-robinson: This is permanently changing the setting on the `ReplicateQueue`, which is definitely not what we want, since it'll disable stats based rebalancing on this node until the next time the node restarts. All we really need is to just pass in whether stats should be disabled to `processOneChange` rather than storing it as state in the `ReplicateQueue`.

a-robinson (commented):
I have a few comments, but yes this approach is good and mostly ready to test.

Also, I can live with it, but all the additional plumbing in `allocator_scorer.go` bugs me. You don't have to do it, but I think we could get rid of it by extracting the few setting values that we care about into a custom struct in the `allocator.go` methods to pass into the `allocator_scorer.go` methods. I'll play around with the idea after this merges if you don't want to deal with it.

pkg/roachpb/api.proto
a6802739: Hmm, You are right. My mistake.

pkg/storage/allocator_scorer.go
a6802739: Done.

pkg/storage/allocator_test.go
a6802739: Done.

pkg/storage/replicate_queue.go
a6802739: Done.

pkg/storage/replica_command.go
a6802739: Yes, My mistake. Done.

a6802739:
@a-robinson  Sorry, I couldn't quite understand what do you mean ` all the additional plumbing in allocator_scorer.go bugs me`.

Do you mean we should move the function `statsBasedRebalancingEnabled`, and global setting value `EnableStatsBasedRebalancing` to `allocator.go`?  The we only need to pass a parameter `disableStatsBasedRebalance` to `allocator_scorer.go`?

And about the test, I wonder how could we verify the number of replicas of each store is pretty close after we `Scatter` the table?

a-robinson (commented):
> @a-robinson Sorry, I couldn't quite understand what do you mean all the additional plumbing in allocator_scorer.go bugs me.

Sorry, bad choice of words. I'm referring to having to pass the `disableStatsBasedRebalance` variable into almost all of the allocator functions now.

> Do you mean we should move the function statsBasedRebalancingEnabled, and global setting value EnableStatsBasedRebalancing to allocator.go? The we only need to pass a parameter disableStatsBasedRebalance to allocator_scorer.go?

`EnableStatsBasedRebalancing` isn't the only setting used in `allocator_scorer.go`, so we'd also have to grab the other couple of settings used in `allocator_scorer.go` and combine them all into a struct in order to actually reduce the number of arguments being passed around.

> And about the test, I wonder how could we verify the number of replicas of each store is pretty close after we Scatter the table?

That would work. We could also just test that calling `processOneChange` or the `Allocator` methods with `disableStatsBasedRebalance` returns the decisions that we expect for a small sampling of existing replica configurations.

a6802739:
> EnableStatsBasedRebalancing isn't the only setting used in allocator_scorer.go, so we'd also have to grab the other couple of settings used in allocator_scorer.go and combine them all into a struct in order to actually reduce the number of arguments being passed around.

Yeah, I could understand what do you mean now. I could try to fix it in another PR.

And I also add a test in `allocator_test.go`, I'm not sure if it's appropriate.

pkg/storage/allocator_test.go
```diff
@@ -1913,6 +1924,79 @@ func TestAllocatorComputeAction(t *testing.T) {
...
+					ReplicaID: 2,
+				},
+				{
+					StoreID:   3,
+					NodeID:    3,
+					ReplicaID: 3,
+				},
+			},
+		}
+		action, _ := a.ComputeAction(ctx, zoneConfig, RangeInfo{Desc: &desc}, true)
```
a-robinson: It'd be nice to run this for both possible input values of the `disableStatsBasedRebalance` parameter to verify that there's a difference in behavior (i.e. that the test is testing what we want it to).

pkg/storage/allocator_test.go
```diff
@@ -1913,6 +1924,79 @@ func TestAllocatorComputeAction(t *testing.T) {
 	}
 }
 
+func TestAllocatorComputeActionDisableStatsRebalance(t *testing.T) {
```
a-robinson: This test needs to set the `EnableStatsBasedRebalancing` cluster setting to true to be interesting at all.

pkg/storage/allocator_test.go
```diff
@@ -1913,6 +1924,79 @@ func TestAllocatorComputeAction(t *testing.T) {
...
+	defer leaktest.AfterTest(t)()
+	ctx := context.Background()
+	stopper, g, _, a, _ := createTestAllocator( /* deterministic */ false)
+	defer stopper.Stop(ctx)
+	stores := []*roachpb.StoreDescriptor{
```
a-robinson: A comment explaining why the different constants (the `RangeCount` values, the `us-east` constraint, etc.) and what this test is testing would be very helpful for anyone who has to debug it in the future.

a6802739:
@a-robinson, Sorry, I have thought about this test, and I found no matter what configure of stores I make, if I run `ComputeAction` to test the action we want to try, it will always be `AllocatorConsiderRebalance`.

pkg/storage/allocator_test.go
a6802739: Done.

a-robinson:
> @a-robinson, Sorry, I have thought about this test, and I found no matter what configure of stores I make, if I run ComputeAction to test the action we want to try, it will always be AllocatorConsiderRebalance.

I'm sorry, I don't quite understand. Do you mean that the line 2137 in allocator_test.go isn't ever passing for you?

a6802739:
@a-robinson, I mean, if I use `ComputeAction` to verify the action we expected, we found no matter what initial stores I set, it will always be `AllocatorConsiderRebalance`.

What I expected is if I set `disableStatsBasedRebalance` to be false, it will be `AllocatorNoop`. Then I found it will return `AllocatorNoop` in this situation:

```
	if len(liveReplicas) < quorum {
		// Do not take any removal action if we do not have a quorum of live
		// replicas.
		log.VEventf(ctx, 1, "unable to take action - live replicas %v don't meet quorum of %d",
			liveReplicas, quorum)
		return AllocatorNoop, 0
	}
``` 

So I mean it seems it's wrong way to test this PR with `ComputeAction`, but I still can't find another way to test it.

a-robinson (commented):
It looks like you should be testing `RebalanceTarget`, since that's what'll make the decision about whether we should rebalance.

a6802739:
@a-robinson , Yeah, I think it works now. 
Thank you very much!

pkg/storage/allocator_test.go
```diff
@@ -2046,6 +2058,108 @@ func TestAllocatorComputeAction(t *testing.T) {
...
+		},
+	}
+	for i := 0; i < 50; i++ {
+		target, _ := a.RebalanceTarget(
+			context.Background(),
+			config.Constraints{},
+			nil,
+			testRangeInfo(desc.Replicas, desc.RangeID),
+			storeFilterThrottled,
+			false,
```
a-robinson: Going by the recent addition to our style guide (https://github.com/cockroachdb/cockroach/pull/19185), this should be `false /* disableStatsBasedRebalance */`

pkg/storage/allocator_test.go
```diff
@@ -2046,6 +2058,108 @@ func TestAllocatorComputeAction(t *testing.T) {
...
+	}
+
+	for i := 0; i < 50; i++ {
+		target, _ := a.RebalanceTarget(
+			context.Background(),
+			config.Constraints{},
+			nil,
+			testRangeInfo(desc.Replicas, desc.RangeID),
+			storeFilterThrottled,
+			true,
```
a-robinson: Going by the recent addition to our style guide (https://github.com/cockroachdb/cockroach/pull/19185), this should be `true /* disableStatsBasedRebalance */`

a-robinson (approved):
LGTM other than a small style comment. Thanks @a6802739!

pkg/storage/allocator_test.go
a6802739: Done now. Thanks!

pkg/storage/allocator_test.go
a6802739: Done now. Thanks!

PR #18374, opened 2017-09-08
storage/engine: reproduce RocksDB DeleteRange bug

petermattis:
See #18371

PR #18436, opened 2017-09-11
c-deps/rocksdb: update to the actual upstream fix for the DeleteRange bug

petermattis:
PR #18374 manually applied an upstream fix for a `DeleteRange`
bug. Unfortunately, the patch applied was slightly different from the
commit applied upstream. THere seems to be some difference between the
github PR and the actual merged commit which I don't know how to
explain (the commit in the PR doesn't exist in their github repo).


petermattis:
Apparently the difference was [just an optimization](https://github.com/facebook/rocksdb/pull/2799#pullrequestreview-59383254). I lean against cherrypicking this to 1.0 and towards cherrypicking it for 1.1.

petermattis:
Cc @tamird

tamird (approved):

tbg (approved):
Thanks for noticing, @tamird!

PR #18437, opened 2017-09-11
sqlccl: disallow errors after chan close in CSV load

### LINKED ISSUE #17336: sqlccl: unique violation error may be swalloed, needs refactor
https://teamcity.cockroachdb.com/viewLog.html?buildId=310737&buildTypeId=Cockroach_UnitTests_Testrace

madelynnblue:
Previously there was a race condition where a go routine in a group
could error, run its defer to close a chan, and then another go routine
would see that close before the first go routine could send its error
back to the errgroup, thus returning the wrong error. Prevent this
by refactoring the code such that no errors can be produced after a
chan is closed from a defer.

We don't lose any concurrency here because the final step was waiting
for all of the SSTs to be written.

Fixes #17336

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -161,10 +161,21 @@ func doLocalCSVTransform(
 	// of pre-computed data improves overall performance.
 	const chanSize = 10000
 
-	group, gCtx := errgroup.WithContext(ctx)
 	recordCh := make(chan csvRecord, chanSize)
 	kvCh := make(chan roachpb.KeyValue, chanSize)
 	contentCh := make(chan sstContent)
+	var backupDesc *BackupDescriptor
+	conf, err := storageccl.ExportStorageConfFromURI(dest)
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	es, err := storageccl.MakeExportStorage(ctx, conf)
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	defer es.Close()
+
+	group, gCtx := errgroup.WithContext(ctx)
 	group.Go(func() error {
 		defer close(recordCh)
 		var err error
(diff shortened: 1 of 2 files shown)


danhhz:
:lgtm:

---

---

*[pkg/ccl/sqlccl/csv.go, line 575 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18437#-KuL7zq_7n24TnQyHQeJ:-KuL7zq_7n24TnQyHQeK:bw2qcsq) ([raw file](https://github.com/cockroachdb/cockroach/blob/f701865e65f259d6530fb5e3390734360c1d5fa6/pkg/ccl/sqlccl/csv.go#L575)):*
> ```Go
> ) (int64, error) {
> 	backupDesc := BackupDescriptor{
> 		FormatVersion: BackupFormatInitialVersion,
> ```

I can see why the finalizeCSVBackup stuff moved, but why did this? seems more natural for it to stay here if that's possible

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18437#-:-KuL8UMMzvWfNtNmEEkP:bnfp4nl)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 575 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18437#-KuL7zq_7n24TnQyHQeJ:-KuLHJB8QJFj7gmm1vHv:b-mw8isp) ([raw file](https://github.com/cockroachdb/cockroach/blob/f701865e65f259d6530fb5e3390734360c1d5fa6/pkg/ccl/sqlccl/csv.go#L575)):*

I can see why the finalizeCSVBackup stuff moved, but why did this? seems more natural for it to stay here if that's possible
</blockquote></details>

Moved this back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18437)*

PR #18449, opened 2017-09-12
storage: cleanup updating of Replica.mu.lastTerm after snapshots

### LINKED ISSUE #18336: storage: clean up lastTerm caching and improve testing
#18327 fixed [an issue](https://github.com/cockroachdb/cockroach/issues/17524) with `r.mu.lastTerm`, but didn't include any testing to accompany the change because we wanted a contained cherry-pick. We should improve testing around raft snapshots and `lastTerm` caching such that the bug fixed in #18327 would have been caught.

While doing so, there may be some room for cleaning up our handling of cached `lastTerm` and `lastIndex`. See @tschottdorf's comments in #18327.

petermattis:
Cleanup the updating of `Replica.mu.lastTerm` after applying a
snapshot. `Replica.applySnapshot` updates
`Replica.mu.{lastTerm,lastIndex,raftLogSize}` and we need to copy those
udpates into local variables in
`Replica.handleRaftReadyRaftMuLocked`. Failure to do this updating
properly led to a bug in which a replica was persistently unavailable
because the cached last term was incorrect. Changed `applySnapshot` to
determine the actual last term and to verify this is updated correctly
in `TestSnapshotAfterTruncation`.

Fixes #18336

### DIFF
--- a/pkg/storage/client_raft_test.go
+++ b/pkg/storage/client_raft_test.go
@@ -660,7 +660,7 @@ func TestSnapshotAfterTruncation(t *testing.T) {
 			defer mtc.Stop()
 			mtc.Start(t, 3)
 			const stoppedStore = 1
-			repl, err := mtc.stores[0].GetReplica(1)
+			repl0, err := mtc.stores[0].GetReplica(1)
 			if err != nil {
 				t.Fatal(err)
 			}
@@ -693,7 +693,7 @@ func TestSnapshotAfterTruncation(t *testing.T) {
 
 			mtc.waitForValues(key, []int64{incAB, incA, incAB})
 
-			index, err := repl.GetLastIndex()
+			index, err := repl0.GetLastIndex()
 			if err != nil {
 				t.Fatal(err)
 			}
@@ -770,6 +770,30 @@ func TestSnapshotAfterTruncation(t *testing.T) {
 				mtc.restartStore(stoppedStore)
 			}
 			mtc.waitForValues(key, []int64{incAB, incAB, incAB})
+
+			// Verify that the cached index and term (Replica.mu.last{Index,Term}))
+			// on all of the replicas is the same. #18327 fixed an issue where the
+			// cached term was left unchanged after applying a snapshot leading to a
+			// persistently unavailable range.
+			repl0, err = mtc.stores[0].GetReplica(1)
+			if err != nil {
+				t.Fatal(err)
+			}
+			expectedLastIndex, _ := repl0.GetLastIndex()
+			expectedLastTerm := repl0.GetCachedLastTerm()
+
+			for i := 1; i < len(mtc.stores); i++ {
+				repl1, err := mtc.stores[i].GetReplica(1)
+				if err != nil {
+					t.Fatal(err)
+				}
+				if lastIndex, _ := repl1.GetLastIndex(); expectedLastIndex != lastIndex {
+					t.Fatalf("%d: expected last index %d, but found %d", i, expectedLastIndex, lastIndex)
+				}
+				if lastTerm := repl1.GetCachedLastTerm(); expectedLastTerm != lastTerm {
+					t.Fatalf("%d: expected last term %d, but found %d", i, expectedLastTerm, lastTerm)
+				}
+			}
 		})
 	}
 }
--- a/pkg/storage/helpers_test.go
+++ b/pkg/storage/helpers_test.go
@@ -339,14 +339,22 @@ func (r *Replica) GetTimestampCacheLowWater() hlc.Timestamp {
 
 // GetRaftLogSize returns the raft log size.
 func (r *Replica) GetRaftLogSize() int64 {
-	r.mu.Lock()
-	defer r.mu.Unlock()
+	r.mu.RLock()
+	defer r.mu.RUnlock()
 	return r.mu.raftLogSize
 }
 
+// GetCachedLastTerm returns the cached last term value. May return
+// invalidLastTerm if the cache is not set.
+func (r *Replica) GetCachedLastTerm() uint64 {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.mu.lastTerm
+}
+
 func (r *Replica) IsRaftGroupInitialized() bool {
-	r.mu.Lock()
-	defer r.mu.Unlock()
+	r.mu.RLock()
+	defer r.mu.RUnlock()
 	return r.mu.internalRaftGroup != nil
 }
 
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -3250,14 +3250,14 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 			return stats, err
 		}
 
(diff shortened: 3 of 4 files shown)


nvanbenschoten:
Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/storage/client_raft_test.go, line 774 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtrVbw3jdNCMvt0E6Ja:-KtrVbw3jdNCMvt0E6Jb:b-7pun1e) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4004a62d13271789dcbd29fe89ab74a7df6cd17/pkg/storage/client_raft_test.go#L774)):*
> ```Go
> 			mtc.waitForValues(key, []int64{incAB, incAB, incAB})
> 
> 			// Verify that the cached term (r.mu.lastTerm) on the stopped store has
> ```

Should we make this more general and assert that all three replicas have the same cached `lastIndex`, `lastTerm`, and `raftLogSize`? 

---

*[pkg/storage/helpers_test.go, line 342 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtrVDrgvn7s0WRQn4G5:-KtrVDrgvn7s0WRQn4G6:b-6vlh44) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4004a62d13271789dcbd29fe89ab74a7df6cd17/pkg/storage/helpers_test.go#L342)):*
> ```Go
> // GetRaftLogSize returns the raft log size.
> func (r *Replica) GetRaftLogSize() int64 {
> 	r.mu.Lock()
> ```

Not touched here, but can't `GetRaftLogSize` and `IsRaftGroupInitialized` also use `r.mu.RLock()`?

---

*[pkg/storage/replica.go, line 3259 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtrVRwz_uXZPdSx1Hjw:-KtrVRwz_uXZPdSx1Hjx:b-bwzk14) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4004a62d13271789dcbd29fe89ab74a7df6cd17/pkg/storage/replica.go#L3259)):*
> ```Go
> 		lastIndex = r.mu.lastIndex
> 		lastTerm = r.mu.lastTerm
> 		raftLogSize = r.mu.raftLogSize
> ```

Was not updating this before another bug?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449)*

petermattis:
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/storage/client_raft_test.go, line 774 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtrVbw3jdNCMvt0E6Ja:-Ktrfzk9xaPmo2QkTyCe:b8c9t7x) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4004a62d13271789dcbd29fe89ab74a7df6cd17/pkg/storage/client_raft_test.go#L774)):*

Should we make this more general and assert that all three replicas have the same cached `lastIndex`, `lastTerm`, and `raftLogSize`? 
</blockquote></details>

Partially done. `raftLogSize` can differ between replicas because it isn't persisted across restarts. We hit the exact case when restarting a store in this test.

---

*[pkg/storage/helpers_test.go, line 342 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtrVDrgvn7s0WRQn4G5:-KtreqDbMOW1AN8t6Xlc:brgq1k0) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4004a62d13271789dcbd29fe89ab74a7df6cd17/pkg/storage/helpers_test.go#L342)):*

Not touched here, but can't `GetRaftLogSize` and `IsRaftGroupInitialized` also use `r.mu.RLock()`?
</blockquote></details>

Sure. Done.

---

*[pkg/storage/replica.go, line 3259 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtrVRwz_uXZPdSx1Hjw:-Ktrf62XPCiay5YOcdVL:b-7s9ash) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4004a62d13271789dcbd29fe89ab74a7df6cd17/pkg/storage/replica.go#L3259)):*

Was not updating this before another bug?
</blockquote></details>

Yes, though the effects were minor. We were either not truncating the Raft log soon enough or truncating before it reached the desired size. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449)*

nvanbenschoten:
:lgtm:

---

Reviewed 1 of 4 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-:-Ktrj_td__umrMe8HnWk:bnfp4nl)*

bdarnell:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2.

---

*[pkg/storage/replica_raftstorage.go, line 778 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtvqRwGwwKj9EDlnyYn:-KtvqRwGwwKj9EDlnyYo:b-mvkpe0) ([raw file](https://github.com/cockroachdb/cockroach/blob/350cd609fe95fc9aa6c9ad46dbc0b1d428d29732/pkg/storage/replica_raftstorage.go#L778)):*
> ```Go
> 			return err
> 		}
> 		raftLogSize += sideloadedEntriesSize
> ```

raftLogSize is always zero at this point, so it can be `=` instead of `+=`. 

Don't we need to set `r.mu.raftLogSize` even when applying a preemptive snapshot? Or do we recompute the log size when the replica id is set?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449)*

petermattis:

---

*[pkg/storage/replica_raftstorage.go, line 778 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449#-KtvqRwGwwKj9EDlnyYn:-KtvsATFdCPPnr9O3v6q:bn2pukx) ([raw file](https://github.com/cockroachdb/cockroach/blob/350cd609fe95fc9aa6c9ad46dbc0b1d428d29732/pkg/storage/replica_raftstorage.go#L778)):*

raftLogSize is always zero at this point, so it can be `=` instead of `+=`. 

Don't we need to set `r.mu.raftLogSize` even when applying a preemptive snapshot? Or do we recompute the log size when the replica id is set?
</blockquote></details>

True about `=` vs `+=`, but it is slightly future proof to use `+=` and I'm also not feeling up to sending a PR to change it. 

When applying a preemptive snapshot, we'll compute the size of the "fat" entries in the call to `Replica.append` below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18449)*

PR #18459, opened 2017-09-12
all: require go 1.9

RaduBerinde:
We will start using go 1.9 features (like `math/bits` or `testing.T.Helper()`)
soon.

### DIFF
--- a/.go-version
+++ b/.go-version
@@ -1,1 +1,1 @@
-GOVERS = go1\.[89].*
+GOVERS = go1\.9.*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Before you start contributing, review these [basic guidelines](https://www.cockr
    - A Go environment with a recent 64-bit version of the toolchain. Note that
     the Makefile enforces the specific version required, as it is updated
     frequently.
-   - Git 1.8+
+   - Git 1.9+
    - Bash (4+ is preferred)
    - GNU Make (3.81+ is known to work)
    - CMake 3.1+
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -101,17 +101,14 @@ RUN git clone --depth 1 https://github.com/tpoechtrager/osxcross.git \
  && mv osxcross/target /x-tools/x86_64-apple-darwin13 \
  && rm -rf osxcross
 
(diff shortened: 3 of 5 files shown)


RaduBerinde:
How should I test the Dockerfile changes? :dog:

tschottdorf:
I'm curious. Once we use these features, cherry-picking might become annoying. Also in general, it'll make switching the branches annoying -- can we just hold off on this until 1.1.1 is out or is there any pressing reason?

RaduBerinde:
Cherry-picking is only a problem if we use any of these features in code that's intended to go into 1.1 as well. Switching branches should not be annoying at all, go 1.9 should work for both.

That said, I can of course hold off.

RaduBerinde:
It looks like TeamCity needs to upgrade the go version before this goes in @jordanlewis

RaduBerinde:
Any tips on how can I build and upload an updated docker image?

petermattis:
I'm fine with merging this, but I'd hold off on using new feature of go1.9 significantly for a little bit. In particular, let's not do a massive code change to use `testing.T.Helper()` until mid-way in the 1.2 cycle. I suppose this falls in the general guidelines to not do massive refactorings early in a cycle.

RaduBerinde:
Agreed, I wasn't planning on refactoring existing code any time soon.

jordanlewis:
@RaduBerinde I think `build/builder.sh init; build/builder.sh push` does it. That will build the new image and send it to Docker Hub. You may need to get an account there first.

Then, you'll need to update the `version` variable in `build/builder.sh` to get TeamCity to pick up the new image. I think that's all there is to do, but I could be forgetting something.

jordanlewis:
Hmm actually you might need to update the version number before you init && push. I'm not sure.

benesch:
Nope, the version number is assigned by `push` (and printed to stdout after the image is successfully uploaded to Docker Hub). Beware: it'll take a few hours to rebuild the builder if you don't have any of the intermediate state cached from a prior rebuild.

RaduBerinde:
Thanks @jordanlewis and @benesch, I think I got it.

RaduBerinde:
@tschottdorf - let me know if you're ok with merging this now or if I should hold off for a couple weeks.

tschottdorf:
Let's do it!
-- 

-- Tobias

benesch (approved):
Awesome, thanks for doing this @RaduBerinde!

PR #18460, opened 2017-09-12
sql: omit vars in replaced filter expressions

jordanlewis:
Previously, when a filter expression is completely replaced by an index
constraint, the expression tree was not properly reanalyzed to see which
columns were necessary to decode during row fetching. This led to
unnecessary extra column decodes.

This patch makes sure that the filter expression is properly reanalyzed.
Removing these extra decodes gives a 15-20% performance improvement for
filtered `SELECT COUNT(*)` queries on an index prefix. For example:

```
CREATE TABLE t (a INT, b INT, PRIMARY KEY (a, b));

SELECT COUNT(*) FROM t WHERE a = 0;
```

Performance on the above query on a single node with 1000000 rows in the
table where `a = 0` improved by 18% on my machine.

cc @danhhz since we were talking about this

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/needed_columns
+++ b/pkg/sql/logictest/testdata/logic_test/needed_columns
@@ -182,3 +182,19 @@ EXPLAIN (VERBOSE) SELECT count(*) FROM (SELECT "name", age FROM a);
 3  scan    ·            ·             ("name"[omitted], age[omitted], rowid[hidden,omitted])  ·
 3  ·       table        a@primary     ·                                                       ·
 3  ·       spans        ALL           ·                                                       ·
+
+# Ensure that variables within filter conditions are omitted (not decoded) if
+# the filter condition is replaced by an index search.
+
+statement ok
+CREATE TABLE ab (a INT, b INT, PRIMARY KEY (a, b));
+
+query ITTTTT
+EXPLAIN (VERBOSE) SELECT count(*) FROM ab WHERE a=1
+----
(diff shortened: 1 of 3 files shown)


pkg/sql/logictest/testdata/logic_test/needed_columns
```diff
@@ -171,7 +171,7 @@ statement ok
 CREATE TABLE a ("name" string, age int);
 
 query ITTTTT
-EXPLAIN (VERBOSE) SELECT count(*) FROM (SELECT "name", age FROM a);
+EXPLAIN (VERBOSE) SELECT COUNT(*) FROM (SELECT "name", age FROM a);
```
knz: we do built-in functions in lowercase now. This needn't change.

pkg/sql/logictest/testdata/logic_test/needed_columns
```diff
@@ -182,3 +182,19 @@ EXPLAIN (VERBOSE) SELECT count(*) FROM (SELECT "name", age FROM a);
...
 3  ·       spans        ALL           ·                                                       ·
+
+# Ensure that variables within filter conditions are omitted (not decoded) if
+# the filter condition is replaced by an index search.
+
+statement ok
+CREATE TABLE ab (a INT, b INT, PRIMARY KEY (a, b));
+
+query ITTTTT
+EXPLAIN (VERBOSE) SELECT COUNT(*) FROM ab WHERE a=1
```
knz: ditto

knz (approved):
LGTM with nit

PR #18461, opened 2017-09-12
util: improvements to the union-find data structure

RaduBerinde:
I've been working on adding equivalency groups to `sql.orderingInfo` and some
shortcomings of the union-find implementation became apparent.

The group representative chosen by the structure depends on the set of union
operations; this complicates some things (like displaying information or
checking if two structures are equal). This change augments the structure to
keep track of the minimum element in each group and always return that as the
representative.

In addition, the interface is improved to operate on ints and eliminate uint32
casts; finally, we add Copy and Equals methods.

### DIFF
--- a/pkg/util/union_find.go
+++ b/pkg/util/union_find.go
@@ -14,6 +14,11 @@
 
 package util
 
+import (
+	"fmt"
+	"math"
+)
+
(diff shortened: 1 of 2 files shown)


knz:
LGTM

PR #18462, opened 2017-09-12
storage: fix two snapshot-related sideloading problems

tbg:
I'll cherry-pick these when they're in.

```
    storage: don't clobber sideloaded storage

    `tryGetOrCreateReplica` was using the incoming `replicaID` to create the
    sideloaded storage. This was incorrect because that `Replica` may already
    have had a replicaID in which case we were silently switching out the
    sideloaded storage. The particular instance in which this was observed as
    a problem was the following:

    1. Replica has nonzero replicaID
    1. it receives incoming preemptive snapshot message
    1. that message erroneously replaces the sideloaded storage with one for
       replicaID zero
    1. a sideloaded proposal applies; gets written to replicaID zero storage
    1. a regular Raft message arrives and again resets the sideload storage (to its
       correct replicaID)
    1. when this Replica attempts to send a snapshot, it won't find the sideloaded
       file for that log entry.

    Added a test that fails before, succeeds after.
```

```
    storage: handle inlined entries when inlining sideloaded data

    When a preemptive snapshot is applied, we (have to) write "fat" entries into the
    log because we don't have a sideloaded storage location yet (no replicaID).

    However, when that node itself later tries to create snapshot, it ends up in the
    situation in which an entry has the sideloading raft command version, but isn't
    actually on disk (but instead already inlined). We weren't handling that case
    correctly, resulting in (the first half of)
    https://github.com/cockroachdb/cockroach/issues/18324#issuecomment-328929342.

    Remedied that and updated TestRaftSSTableSideloadingInline so that it fails before
    the fix and passes after.
```

### DIFF
--- a/pkg/storage/client_raft_test.go
+++ b/pkg/storage/client_raft_test.go
@@ -1121,14 +1121,24 @@ func TestStoreRangeUpReplicate(t *testing.T) {
 	mtc.stores[0].ForceReplicationScanAndProcess()
 
 	// The range should become available on every node.
+	var r *storage.Replica // from the last store
 	testutils.SucceedsSoon(t, func() error {
+		rs := roachpb.RSpan{
+			Key: roachpb.RKey("a"), EndKey: roachpb.RKey("z"),
+		}
 		for _, s := range mtc.stores {
-			r := s.LookupReplica(roachpb.RKey("a"), roachpb.RKey("b"))
+			r = s.LookupReplica(rs.Key, rs.EndKey)
 			if r == nil {
-				return errors.Errorf("expected replica for keys \"a\" - \"b\"")
+				return errors.Errorf("expected replica for span %v", rs)
 			}
 			if n := s.ReservationCount(); n != 0 {
 				return errors.Errorf("expected 0 reservations, but found %d", n)
+			}
+			if len(r.Desc().Replicas) != 3 {
+				// This fails even after the preemptive snapshot has arrived and
+				// only goes through once the replica has properly caught up to
+				// the fully replicated descriptor.
+				return errors.Errorf("not fully initialized")
 			}
 		}
 		return nil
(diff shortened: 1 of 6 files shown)


bdarnell:
:lgtm:

---

Reviewed 3 of 3 files at r2, 3 of 3 files at r3.

---

*[pkg/storage/replica.go, line 850 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18462#-KtsQ9QUrVvuvOHVCQmA:-KtsQ9QUrVvuvOHVCQmB:b-wjx52l) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5220b3368f5875962036e3a7e9e79698a485896/pkg/storage/replica.go#L850)):*
> ```Go
> 	// }
> 
> 	// Initialize the new sideloaded storage.
> ```

If we already had sideloaded storage from our previous replicaID, don't we need to copy any sideloaded files from that replica ID to our new one?

---

*[pkg/storage/replica_sideload.go, line 233 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18462#-KtsP7eeWsmUGZTIJgXm:-KtsP7eeWsmUGZTIJgXn:bz2h01) ([raw file](https://github.com/cockroachdb/cockroach/blob/6b05d5b841fbbf530709228b6013a0c18e1ea90a/pkg/storage/replica_sideload.go#L233)):*
> ```Go
> 	// Out of luck, for whatever reason the inlined proposal isn't in the cache.
> 	cmdID, data := DecodeRaftCommand(ent.Data)
> 	ent.Data = nil // no reuse of potentially shared slice
> ```

Did something change that makes this safe to remove now while it was necessary before, or has it been unnecessary for a while?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18462#-:-KtsQFWx0Lw_61rs72np:bnfp4nl)*

tschottdorf:

---

*[pkg/storage/replica.go, line 850 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18462#-KtsQ9QUrVvuvOHVCQmA:-Ktt-jwz6qZHSam0OdM-:bnkztho) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5220b3368f5875962036e3a7e9e79698a485896/pkg/storage/replica.go#L850)):*

If we already had sideloaded storage from our previous replicaID, don't we need to copy any sideloaded files from that replica ID to our new one?
</blockquote></details>

We discussed this offline and basically Ben is right. I *hope* this is relatively straightforward but since it's a separate bug that needs some testing, I'll fix it separately.

---

*[pkg/storage/replica_sideload.go, line 233 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18462#-KtsP7eeWsmUGZTIJgXm:-Ktt-rlKWAOJH8tsZtjD:b-rsx0cw) ([raw file](https://github.com/cockroachdb/cockroach/blob/6b05d5b841fbbf530709228b6013a0c18e1ea90a/pkg/storage/replica_sideload.go#L233)):*

Did something change that makes this safe to remove now while it was necessary before, or has it been unnecessary for a while?
</blockquote></details>

I had introduced it back then to make it less likely below to accidentally change the wrong thing. Removed it now because it didn't make much sense and we have tests that ensure the original memory doesn't get mutated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18462)*

PR #18467, opened 2017-09-13
storage: preserve sideloaded files on replicaID bump

tbg:
If a replica gets removed from and subsequently re-added to a store without
being garbage collected in the meantime, its replicaID will change but it will
otherwise keep most of its state.

Before this commit, we would create a new sideloaded storage for the new
replicaID, effectively losing the old one. Now, the contents are moved along as
well.

Observed by @bdarnell in
https://github.com/cockroachdb/cockroach/pull/18462#issuecomment-329013272

I'll cherry-pick this into 1.1.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"os"
 	"reflect"
 	"sort"
 	"sync/atomic"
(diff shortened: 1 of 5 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/storage/replica.go, line 870 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18467#-KtveBB2XxHAnbOm0-wE:-KtveBB34XBh8eLq-2HH:buafxtr) ([raw file](https://github.com/cockroachdb/cockroach/blob/77feec3202cffebd799b1fed7089def6680f71ef/pkg/storage/replica.go#L870)):*
> ```Go
> 	}
> 	if prevSideloadedDir != "" {
> 		if _, err := os.Stat(prevSideloadedDir); err != nil {
> ```

Rather than calling `os.Stat`, can you blindly call `os.Rename` here and look for `os.IsNotExist(err)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18467#-:-KtveMuk_yfouK7kx22o:bnfp4nl)*

tschottdorf:
TFTR! I'll let @bdarnell get a look in, too.

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/storage/replica.go, line 870 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18467#-KtveBB2XxHAnbOm0-wE:-KtvlGhDS-B2dHIErE6l:b-7jxvs3) ([raw file](https://github.com/cockroachdb/cockroach/blob/77feec3202cffebd799b1fed7089def6680f71ef/pkg/storage/replica.go#L870)):*

Rather than calling `os.Stat`, can you blindly call `os.Rename` here and look for `os.IsNotExist(err)`?
</blockquote></details>

I was hesitant to do that because the documentation isn't clear whether `os.IsNotExist` would actually catch a "not found" error from `os.Rename`, and I additionally wasn't sure whether we might end up catching "not found"-like errors for the destination. Have a slight preference for leaving as is (especially seeing how which errors are returned is system-specific). That said, it would *probably* just work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18467#-:-Ktvlcqw7WV1wX2NiBgw:b375lc2)*

bdarnell:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18467#-:-KtvwBB7Oqi-b1QWtTWz:bnfp4nl)*

PR #18472, opened 2017-09-13
cli: simplify and improve dump performance

madelynnblue:
A significant percentage (near 50%) of CPU time during dump is spent
converting strings returned from the SELECT into SQL-safe strings and
writing that to the output. Much of this work can be done concurrently,
which improves the performance, and gets it close to that of piping
a `SELECT *` to a file. This is done using an errgroup and passing
various stages over channels.

We can also greatly simplify dump now since cockroach streams
results. There's no need anymore to page through the data.

See cockroachdb/docs#1674

### DIFF
--- a/pkg/cli/dump.go
+++ b/pkg/cli/dump.go
@@ -24,8 +24,10 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
+
 	"github.com/cockroachdb/cockroach/pkg/sql/parser"
-	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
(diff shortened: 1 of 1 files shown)


knz (commented):
I think this is fine, but don't we want to throttle this? I don't recall that Go guarantees unlimited concurrent goroutines. @petermattis what do you think?

mjibson:
Throttle what? There's only 3 go routines. It never went above 200% CPU during my tests (on a 12-core machine). The buffered chan only has ~100 strings in it. Am I missing something?

knz:
Ugh, I'm sorry for the noise. I had wrongly read that there was one goroutine per row. LGTM as-is.

PR #18479, opened 2017-09-13
libroach: log estimated-pending-compaction-bytes

petermattis:
RocksDB will slow down writes when pending-compaction-bytes exceeds
64GB. This shouldn't happen under normal operation, but appears to be
happening during a restore.

See #17902


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18479#-:-Ku-pc_S71h7LgDEqmof:bnfp4nl)*

PR #18482, opened 2017-09-13
storage: improve error message for failed splits

petermattis:
When a split fails because the underlying descriptor changed provide a
more informative error messages.

Before:

```
unexpected value:
raw_bytes:"I\205B\247\003\010\001\022\000\032\001a\"\006\010\001\020\001\030\001(\002"
timestamp:<wall_time:1505319352553640907 logical:0 >
```

After:

```
descriptor changed: r1:/M{in-ax} [(n1,s1):1, next=2] != r1:{/Min-a} [(n1,s1):1, next=2]
```

See #17180

### DIFF
--- a/pkg/storage/replica_command.go
+++ b/pkg/storage/replica_command.go
@@ -2613,6 +2613,21 @@ func (r *Replica) AdminSplit(
 	return roachpb.AdminSplitResponse{}, roachpb.NewError(ctx.Err())
 }
 
+func maybeDescriptorChangedError(desc *roachpb.RangeDescriptor, err error) (string, bool) {
+	if detail, ok := err.(*roachpb.ConditionFailedError); ok {
+		// Provide a better message in the common case that the range being changed
+		// was already changed by a concurrent transaction.
+		var actualDesc roachpb.RangeDescriptor
+		if err := detail.ActualValue.GetProto(&actualDesc); err == nil {
+			if desc.RangeID == actualDesc.RangeID && !desc.Equal(actualDesc) {
+				return fmt.Sprintf("descriptor changed: [expected] %s != [actual] %s",
+					desc, actualDesc), true
+			}
+		}
+	}
+	return "", false
+}
+
 // adminSplitWithDescriptor divides the range into into two ranges, using
 // either args.SplitKey (if provided) or an internally computed key that aims
 // to roughly equipartition the range by size. The split is done inside of a
@@ -2784,10 +2799,13 @@ func (r *Replica) adminSplitWithDescriptor(
 		// range descriptors are picked outside the transaction. Return
 		// ConditionFailedError in the error detail so that the command can be
 		// retried.
-		if _, ok := err.(*roachpb.ConditionFailedError); ok {
-			return reply, true, roachpb.NewError(err)
+		pErr := roachpb.NewError(err)
+		if msg, ok := maybeDescriptorChangedError(desc, err); ok {
+			pErr.Message = fmt.Sprintf("split at key %s failed: %s", splitKey, msg)
+		} else {
+			pErr.Message = fmt.Sprintf("split at key %s failed: %s", splitKey, err)
 		}
-		return reply, true, roachpb.NewErrorf("split at key %s failed: %s", splitKey, err)
+		return reply, true, pErr
 	}
 	return reply, true, nil
 }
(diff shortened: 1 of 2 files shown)


petermattis:
Should probably do this for `ChangeReplicas` as well.

pkg/storage/replica_command.go
```diff
@@ -2784,8 +2784,17 @@ func (r *Replica) adminSplitWithDescriptor(
...
-		if _, ok := err.(*roachpb.ConditionFailedError); ok {
-			return reply, true, roachpb.NewError(err)
+		if detail, ok := err.(*roachpb.ConditionFailedError); ok {
+			pErr := roachpb.NewError(err)
+			// Provide a better message in the common case that the range being split
+			// changed during the split.
+			var actualDesc roachpb.RangeDescriptor
+			if err := detail.ActualValue.GetProto(&actualDesc); err == nil {
+				if desc.RangeID == actualDesc.RangeID && !desc.Equal(actualDesc) {
+					pErr.Message = fmt.Sprintf("descriptor changed: %s != %s", desc, actualDesc)
```
a-robinson: It's a minor thing, but it'll take some thought when seeing this in the logs which value is the expected and which is the actual. That could probably be made clearer.

a-robinson (approved):

pkg/storage/replica_command.go
petermattis: Changed. PTAL.

petermattis:
I extended this change to also provide a better error message on failed change-replicas operations.

bdarnell:
:lgtm:

---

---

*[pkg/storage/replica_command.go, line 2618 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18482#-Ku-q8dZC1g7Qh6p4SOI:-Ku-q8dZC1g7Qh6p4SOJ:b-kgka6n) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7798125e51d805472ef395d1efa0942532e1716/pkg/storage/replica_command.go#L2618)):*
> ```Go
> func maybeDescriptorChangedError(desc *roachpb.RangeDescriptor, err error) (string, bool) {
> 	if detail, ok := err.(*roachpb.ConditionFailedError); ok {
> 		// Provide a better message in the common case that the range being split
> ```

s/split/changed/

---

*[pkg/storage/replica_command.go, line 2619 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18482#-Ku-qARU7pnSk4DZhBXl:-Ku-qARU7pnSk4DZhBXm:b-ebtre2) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7798125e51d805472ef395d1efa0942532e1716/pkg/storage/replica_command.go#L2619)):*
> ```Go
> 	if detail, ok := err.(*roachpb.ConditionFailedError); ok {
> 		// Provide a better message in the common case that the range being split
> 		// changed during the split.
> ```

"was also changed by a concurrent transaction"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18482#-:-Ku-qLu8C9HkbafEE_WH:bnfp4nl)*

pkg/storage/replica_command.go
```diff
@@ -3702,6 +3720,9 @@ func (r *Replica) changeReplicas(
 		return nil
 	}); err != nil {
 		log.Event(ctx, err.Error())
+		if msg, ok := maybeDescriptorChangedError(desc, err); ok {
+			return errors.Wrapf(err, "change replicas of r%d failed: %s", rangeID, msg)
```
a-robinson: Is wrapping here useful? It'll just include the original `ConditionFailedError` again after the improved error message

a-robinson (approved):

petermattis:

---

*[pkg/storage/replica_command.go, line 2618 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18482#-Ku-q8dZC1g7Qh6p4SOI:-Ku-rQozAZKu_h79ms2D:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7798125e51d805472ef395d1efa0942532e1716/pkg/storage/replica_command.go#L2618)):*

s/split/changed/
</blockquote></details>

Done.

---

*[pkg/storage/replica_command.go, line 2619 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18482#-Ku-qARU7pnSk4DZhBXl:-Ku-rQNv4CO8tNXTsRtb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7798125e51d805472ef395d1efa0942532e1716/pkg/storage/replica_command.go#L2619)):*

"was also changed by a concurrent transaction"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18482)*

pkg/storage/replica_command.go
petermattis: It isn't useful at the moment and the additional error message is a bit confusing and worthless though it does contain a bit of additional information. The one benefit of wrapping is we can call `errors.Cause()` to extract the original error, though we don't currently need that on this error path.

PR #18483, opened 2017-09-13
sql/sqlbase: don't read idx key if not necessary.

jordanlewis:
During row fetching, if we don't need any of the columns contained
within the index key and the index is not interleaved, we can completely
skip reading and processing the index key.

This gives a nice speed boost for count(*) queries. I observe a ~10-15%
improvement on a million row dataset on a local 1 node cluster.

cc @nvanbenschoten on @knz's recommendation

### DIFF
--- a/pkg/server/settingsworker.go
+++ b/pkg/server/settingsworker.go
@@ -46,7 +46,7 @@ func (s *Server) refreshSettings() {
 		// First we need to decode the setting name field from the index key.
 		{
 			nameRow := []sqlbase.EncDatum{{Type: tbl.Columns[0].Type}}
-			_, matches, err := sqlbase.DecodeIndexKey(a, &tbl, tbl.PrimaryIndex.ID, nameRow, nil, kv.Key)
+			_, matches, err := sqlbase.DecodeIndexKey(a, &tbl, &tbl.PrimaryIndex, nameRow, nil, kv.Key)
 			if err != nil {
 				return errors.Wrap(err, "failed to decode key")
 			}
--- a/pkg/server/settingsworker.go
+++ b/pkg/server/settingsworker.go
@@ -31,7 +31,7 @@ import (
 
 // RefreshSettings starts a settings-changes listener.
 func (s *Server) refreshSettings() {
-	tbl := sqlbase.SettingsTable
+	tbl := &sqlbase.SettingsTable
 
 	a := &sqlbase.DatumAlloc{}
 	settingsTablePrefix := keys.MakeTablePrefix(uint32(tbl.ID))
@@ -46,7 +46,7 @@ func (s *Server) refreshSettings() {
 		// First we need to decode the setting name field from the index key.
 		{
 			nameRow := []sqlbase.EncDatum{{Type: tbl.Columns[0].Type}}
-			_, matches, err := sqlbase.DecodeIndexKey(a, &tbl, &tbl.PrimaryIndex, nameRow, nil, kv.Key)
+			_, matches, err := sqlbase.DecodeIndexKey(a, tbl, &tbl.PrimaryIndex, nameRow, nil, kv.Key)
 			if err != nil {
 				return errors.Wrap(err, "failed to decode key")
 			}
--- a/pkg/sql/sqlbase/rowfetcher.go
+++ b/pkg/sql/sqlbase/rowfetcher.go
@@ -278,7 +278,7 @@ func prettyEncDatums(vals []EncDatum) string {
 
 // ReadIndexKey decodes an index key for the fetcher's table.
 func (rf *RowFetcher) ReadIndexKey(k roachpb.Key) (remaining []byte, ok bool, err error) {
-	return DecodeIndexKey(rf.alloc, rf.desc, rf.index.ID, rf.keyVals,
+	return DecodeIndexKey(rf.alloc, rf.desc, rf.index, rf.keyVals,
 		rf.indexColumnDirs, k)
 }
 
--- a/pkg/sql/sqlbase/rowfetcher.go
+++ b/pkg/sql/sqlbase/rowfetcher.go
@@ -61,6 +61,10 @@ type RowFetcher struct {
 	// The set of ColumnIDs that are required.
 	neededCols util.FastIntSet
 
+	// True if the index key must be decoded. This is only true if there are no
+	// needed columns and the table has no interleave children.
+	mustDecodeIndexKey bool
+
 	// Map used to get the index for columns in cols.
 	colIdxMap map[ColumnID]int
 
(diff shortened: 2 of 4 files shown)


jordanlewis:
Hmm, don't bother reviewing this yet. It is broken.

jordanlewis:
PTAL.

knz:
Reviewed 4 of 4 files at r1, 1 of 1 files at r2.

---

*[pkg/sql/sqlbase/rowfetcher.go, line 245 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku4JPcoWA0z0m9mrAUP:-Ku4JPcqbaCdR7-z7aIt:b-8sewvi) ([raw file](https://github.com/cockroachdb/cockroach/blob/5e4e35220d0dc8f89296d9f547298ba5590be01d/pkg/sql/sqlbase/rowfetcher.go#L245)):*
> ```Go
> 		// reading work if we need values from outside of the key, but not from
> 		// inside of the key.
> 		if len(rf.index.Interleave.Ancestors) > 0 || len(rf.index.InterleavedBy) > 0 || !rf.neededCols.Empty() {
> ```

1) change the order of the conditional. `!rf.neededCols.Empty()` is the most common case and thus should be handled first (this is on the hot path...)

2) this condition does not change from one row to the next. You can precompute the boolean and reuse it here. This would enable you to compute a slightly more precise condition (neededCols only lists non-PK columns).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483)*

jordanlewis:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/sqlbase/rowfetcher.go, line 245 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku4JPcoWA0z0m9mrAUP:-Ku5UEQa9A9UUgQAkeRM:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5e4e35220d0dc8f89296d9f547298ba5590be01d/pkg/sql/sqlbase/rowfetcher.go#L245)):*

1) change the order of the conditional. `!rf.neededCols.Empty()` is the most common case and thus should be handled first (this is on the hot path...)

2) this condition does not change from one row to the next. You can precompute the boolean and reuse it here. This would enable you to compute a slightly more precise condition (neededCols only lists non-PK columns).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483)*

knz:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-:-Ku5hjJoKFcDyyKyuJvA:bnfp4nl)*

nvanbenschoten:

---

*[pkg/server/settingsworker.go, line 49 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku5nHYWfz2p1NmVfR55:-Ku5nHYWfz2p1NmVfR56:b-yfki83) ([raw file](https://github.com/cockroachdb/cockroach/blob/5e4e35220d0dc8f89296d9f547298ba5590be01d/pkg/server/settingsworker.go#L49)):*
> ```Go
> 		{
> 			nameRow := []sqlbase.EncDatum{{Type: tbl.Columns[0].Type}}
> 			_, matches, err := sqlbase.DecodeIndexKey(a, &tbl, &tbl.PrimaryIndex, nameRow, nil, kv.Key)
> ```

This will probably allocate if we don't change `tbl := sqlbase.SettingsTable` to `tbl := &sqlbase.SettingsTable`.

---

*[pkg/sql/sqlbase/rowfetcher.go, line 65 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku5o7246mCMs5LTXYS0:-Ku5o7246mCMs5LTXYS1:bacrsv) ([raw file](https://github.com/cockroachdb/cockroach/blob/b16b108e2b34fa7b8d24275d7ee4ce07db10e649/pkg/sql/sqlbase/rowfetcher.go#L65)):*
> ```Go
> 
> 	// True if the index key must be decoded. This is only true if there are no
> 	// needed columns and the table has no interleave children.
> ```

`interleave`? `interleaved`? `interleaven`?

---

*[pkg/sql/sqlbase/rowfetcher.go, line 149 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku5pFsNFxYEe27BgWkM:-Ku5pFsNFxYEe27BgWkN:b-jts1xy) ([raw file](https://github.com/cockroachdb/cockroach/blob/b16b108e2b34fa7b8d24275d7ee4ce07db10e649/pkg/sql/sqlbase/rowfetcher.go#L149)):*
> ```Go
> 	}
> 
> 	if !rf.neededCols.Empty() || len(rf.index.InterleavedBy) > 0 || len(rf.index.Interleave.Ancestors) > 0 {
> ```

The first part of this condition is intuitive, but could you leave a comment explaining the other two? I think we have that below, but wouldn't it make more sense here since it's where we're setting `mustDecodeIndexKey`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483)*

jordanlewis:

---

*[pkg/server/settingsworker.go, line 49 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku5nHYWfz2p1NmVfR55:-Ku7PphfVXqAAmx4Tftf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5e4e35220d0dc8f89296d9f547298ba5590be01d/pkg/server/settingsworker.go#L49)):*

This will probably allocate if we don't change `tbl := sqlbase.SettingsTable` to `tbl := &sqlbase.SettingsTable`.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/rowfetcher.go, line 65 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku5o7246mCMs5LTXYS0:-Ku7Ps2jf-Aghl6IBEjv:bfdjkui) ([raw file](https://github.com/cockroachdb/cockroach/blob/b16b108e2b34fa7b8d24275d7ee4ce07db10e649/pkg/sql/sqlbase/rowfetcher.go#L65)):*

`interleave`? `interleaved`? `interleaven`?
</blockquote></details>

To me, `interleave children` makes more sense than `interleaved children` even though it's not really English grammatical. I think that's because `interleaved children` doesn't make it obvious (to me anyway) that we're talking about the CockroachDB concept of an `interleave` - `interleave children` says to me `children` in the scope of CRDB `interleaves`.

---

*[pkg/sql/sqlbase/rowfetcher.go, line 149 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483#-Ku5pFsNFxYEe27BgWkM:-Ku7QN4OzZFpC6oR6ho1:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b16b108e2b34fa7b8d24275d7ee4ce07db10e649/pkg/sql/sqlbase/rowfetcher.go#L149)):*

The first part of this condition is intuitive, but could you leave a comment explaining the other two? I think we have that below, but wouldn't it make more sense here since it's where we're setting `mustDecodeIndexKey`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18483)*

PR #18485, opened 2017-09-13
cloud: Update kubernetes configs for v1.1

a-robinson:
Being sure to add on the cache and max-sql-memory flags that are now
important for production-ready configs.

Sending out mostly so that I don't forget to update the cache flag. Not to be submitted until we actually have a usable v1.1.0 docker image.


a-robinson:
Superseded by https://github.com/cockroachdb/cockroach/pull/19218

PR #18487, opened 2017-09-13
cherrypick-1.1: distsql: knobs for temp storage size

### LINKED ISSUE #17663: distsql: expose knob for limiting query disk space usage

RaduBerinde:
Adding cluster settings for max temp disk space usage, both as a percentage of
total capacity and as an absolute number (whichever is smaller).

The default limit is MIN(10% of total disk capacity, 32GB).

Fixes #17663.

CC @cockroachdb/release

### DIFF
--- a/pkg/sql/distsqlrun/hashjoiner.go
+++ b/pkg/sql/distsqlrun/hashjoiner.go
@@ -161,7 +161,7 @@ func (h *hashJoiner) Run(ctx context.Context, wg *sync.WaitGroup) {
 	}
 
 	st := h.flowCtx.Settings
-	useTempStorage := (distSQLUseTempStorage.Get(&st.SV) && distSQLUseTempStorageJoins.Get(&st.SV)) ||
+	useTempStorage := settingUseTempStorageJoins.Get(&st.SV) ||
 		h.flowCtx.testingKnobs.MemoryLimitBytes > 0 ||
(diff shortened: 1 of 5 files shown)


andreimatei:
LGTM

PR #17816, opened 2017-08-22
pgwire, cli/sql: support tables with no columns properly

knz:
Fixes #13769.
Supersedes #14973.

PR #18491, opened 2017-09-14
cli: fix pretty print alignment

### LINKED ISSUE #10042: cli: strings that look like numbers are aligned incorrectly
Looks like some pretty printer somewhere is overzealous in recognizing numbers, and aligns them right instead of left. This most annoyingly happens when the apparent numbers are a result of a `substr()` operation, and some results suddenly get aligned differently than the rest.

We should look at the type of the result, if at all possible, or at least at all the results in that column.

```
root@:26258> SELECT '0'::text as XXXXX;
+-------+
| XXXXX |
+-------+
|     0 |
+-------+
(1 row)
root@:26258> SELECT '0s'::text as XXXXX;
+-------+
| XXXXX |
+-------+
| 0s    |
+-------+
```

pkit:
This patch modifies the `pretty` table format implementation so that
caller scan specify alignment of results inside the table. The
alignment directive supports "l", "r" and "c" for left, right and
center horizontal alignment, respectively.

Alignment in results printed by the `sql` client have alignment
decided automatically based on column type:

- strings and unicode: `left`
- ints and other numericals: `right`
- bools: `center`

This patch also adds proper alignment to `node ls` and other
"ad-hoc/non-table" queries.

Fixes #10042.

Release note (cli change): command-line utilities that print results
with the `pretty` format now use consistent horizontal alignment for
every row of the result.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -523,7 +523,7 @@
   branch = "master"
   name = "github.com/lib/pq"
   packages = [".","oid"]
-  revision = "b609790bd85edf8e9ab7e0f8912750a786177bcf"
+  revision = "27ea5d92de30060e7121ddd543fe14e9a327e0cc"
 
 [[projects]]
   name = "github.com/lightstep/lightstep-tracer-go"
--- a/pkg/cli/cert.go
+++ b/pkg/cli/cert.go
@@ -173,6 +173,7 @@ func runListCerts(cmd *cobra.Command, args []string) error {
 	fmt.Fprintf(os.Stdout, "Certificate directory: %s\n", baseCfg.SSLCertsDir)
 
 	certTableHeaders := []string{"Usage", "Certificate File", "Key File", "Expires", "Notes", "Error"}
+	alignment := "llllll"
 	var rows [][]string
 
 	addRow := func(ci *security.CertInfo, notes string) {
@@ -223,7 +224,7 @@ func runListCerts(cmd *cobra.Command, args []string) error {
 		addRow(cert, fmt.Sprintf("user: %s", user))
 	}
 
-	return printQueryOutput(os.Stdout, certTableHeaders, newRowSliceIter(rows))
+	return printQueryOutput(os.Stdout, certTableHeaders, newRowSliceIter(rows, alignment))
 }
 
 var certCmds = []*cobra.Command{
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -1873,3 +1873,38 @@ func Example_in_memory() {
 	// 1
 	// # 1 row
 }
+
+func Example_pretty_print_numerical_strings() {
+	c := newCLITest(cliTestParams{})
+	defer c.cleanup()
+
+	// All strings in pretty-print output should be aligned to left regardless of their contents
+	c.RunWithArgs([]string{"sql", "-e", "create database t; create table t.t (s string, d string);"})
+	c.RunWithArgs([]string{"sql", "-e", "insert into t.t values (e'0', 'positive numerical string')"})
+	c.RunWithArgs([]string{"sql", "-e", "insert into t.t values (e'-1', 'negative numerical string')"})
+	c.RunWithArgs([]string{"sql", "-e", "insert into t.t values (e'1.0', 'decimal numerical string')"})
+	c.RunWithArgs([]string{"sql", "-e", "insert into t.t values (e'aaaaa', 'non-numerical string')"})
+	c.RunWithArgs([]string{"sql", "--format=pretty", "-e", "select * from t.t"})
+
+	// Output:
+	// sql -e create database t; create table t.t (s string, d string);
+	// CREATE TABLE
+	// sql -e insert into t.t values (e'0', 'positive numerical string')
+	// INSERT 1
+	// sql -e insert into t.t values (e'-1', 'negative numerical string')
+	// INSERT 1
+	// sql -e insert into t.t values (e'1.0', 'decimal numerical string')
+	// INSERT 1
+	// sql -e insert into t.t values (e'aaaaa', 'non-numerical string')
+	// INSERT 1
+	// sql --format=pretty -e select * from t.t
+	// +-------+---------------------------+
+	// |   s   |             d             |
+	// +-------+---------------------------+
+	// | 0     | positive numerical string |
+	// | -1    | negative numerical string |
+	// | 1.0   | decimal numerical string  |
+	// | aaaaa | non-numerical string      |
+	// +-------+---------------------------+
+	// (4 rows)
+}
--- a/pkg/cli/format_table.go
+++ b/pkg/cli/format_table.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"html"
 	"io"
+	"reflect"
 	"strings"
 	"unicode/utf8"
 
@@ -34,6 +35,7 @@ import (
 type rowStrIter interface {
 	Next() (row []string, err error)
 	ToSlice() (allRows [][]string, err error)
+	Align() []int
 }
 
 // rowSliceIter is an implementation of the rowStrIter interface and it is used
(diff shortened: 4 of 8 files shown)


pkit:
Depends on upstream: https://github.com/olekukonko/tablewriter/pull/74

knz:
Many thanks for working on this. This problem had been bugging us for *ages*! cc @andreimatei who was regularly pushing me to look into this :smile_cat: 

I'll help with the upstream dependency a bit later. In the meantime, could you enhance this patch by supporting a string-based alignment specifier, instead of arrays of int constants from the tablewriter package? The idea would be as follows:

1) we want to restrict the dependency on the specific rendering package (`tablewriter`) to format_table.go. The rest of the source code should be agnostic wrt what rendering is being used.

2) the current code is excessively verbose.

The method I would like to use is the same as used in [LaTeX table alignment specifiers](https://en.wikibooks.org/wiki/LaTeX/Tables#The_tabular_environment), where each character in a string would specify the alignment for the corresponding result column. We'd support just "l", "c", "r", and "a" for tablewriter's "default" which really means "autodetected". 

Thank you again!

---

Reviewed 5 of 6 files at r1.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18491#-:-Ku48FN5fVYw5OY9Udx-:bg6t81v)*

knz:
Also please prefix your commit title with "`cli: `" Thanks!

knz:
Also please rebase your fix on top of #17816 - this significantly refactors the code and is likely to merge before this PR.

pkit:
Totally agree that dependency on tablewriter should be contained. Was not sure what's a better way to rewrite these alignment identifiers though. Using just one string seems good enough. Will use it.

---

Review status: 5 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18491#-:-Ku6N50JX5wA4X1ob6ew:bh0lshv)*

pkit:
Will rebase against #17816 when it lends in master.

knz:
It's ready.
-- 

pkit:
Rebased. Code fragility was strong in this one. Needed to refactor the strange `io.EOF` handling

knz:
Okay this is better. The first commit is fine.
Now I think  the plumbing you have achieved in `format_table.go` has too many conditionals. In general conditionals in a "hot path" (like  the `Next` methods) is undesirable, so I propose the following:

- change the return type of `Align` in the interface `rowStrIter` to `string`
- make `newRowSliceIter` take the alignment string as argument and store it in `rowSliceIter` without processing, and make `(*rowSliceIter).Align()` return it.
- add a new field `lastRow []driver.Value` in `rowIter`
- leave the methods `getAllRowStrings` and `getNextRowStrings` alone as they were initially, instead, for `(*rowIter).Next()` *inline* the code from `getNextRowString` into `(*rowIter).Next()` and make it store the current row being retrieved into the rowIter's `lastRow`
- make `(*rowIter).Align()` compute the alignment string based on the last row stored in `rowIter`
- extend the interface `rowReporter.beforeFirstRow` to take an additional `alignment string` parameter. 
- change the `render` function accordingly. It is possible to call `(rowStrIter).Align()` at the point `beforeFirstRow` is called, because by then there will have been at least one row read by the iterator.
- Ignore the new `beforeFirstRow` argument in all implementations of the interface except for `prettyReporter`, where you use it to call `p.table.SetTableAlignment`. No need to store a new field in `prettyReporter`. 

---

Reviewed 6 of 6 files at r2, 6 of 6 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18491#-:-KuKBvsd6V3KNTuhokNt:b48u5cy)*

pkit:
What about `nil` handling? `driver.Value` is just an `interface{}`, can I safely derive real type from `nil`s?
What if we really get an `interface{}(nil)` from driver up until "normal" value is materialized there?

knz:
Oh, how unfortunate.
It's rather important to not be computing something in the `Next` method especially since most formatters will not use this information.

So what can we do here? It's frustrating to look at this knowing that *the pq driver does know about the column types from the beginning*! 

So I suggest the following. It's a bit more intricate but I am confident it is worth it:
- patch the copy of `pq` in the `vendor` sub-directory, to add a new method `func (rs*rows).GetColTypes() []oid.Oid { return rs.colTyps }`
- extend the `sqlRowsI` interface in `sql_util.go` with `GetColTypes() []oid.Oid`
- (check that this runs fine: the rows object returned by Query can be casted to `sqlRowsI` without a panic, that is, we are indeed seeing pq's rows object)
- use this new method in the `rowsIter.Align` method.

You may need to create a PR for the change in the `vendor` directory as this is handled by a separate repository. See `build/README.md` for details.

pkit:
Yeah, now we're talking. Checked out and indeed got almost +50% of time spent in `Next()` which is obviously a bad idea. Was frustrated to see that no type info is easily accessible anywhere. Time to fix it!

pkit:
Added https://github.com/cockroachdb/vendored/pull/8

mjibson:
I've merged https://github.com/lib/pq/pull/647. You'll need to submit a PR to vendored with that change and verify it works for this.

pkit:
@mjibson @knz tests pass with current HEAD of https://github.com/lib/pq

knz:
All right. I think you can run `dep ensure --upgrade github.com/lib/pq` in the `vendor` sub-dir to just upgrade that dependency. Then submit a PR to our vendored repo.

---

Review status: 4 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18491#-:-KuLsc19aBcNLQZdzrhk:bdji8z1)*

pkit:
Did that already. The result is in https://github.com/cockroachdb/vendored/pull/9

---

Review status: 4 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18491#-:-KuNwadXjwq4SqQW-UQr:bku78mr)*

knz:
Ok you can bump the vendor hash now.

pkit:
I think https://github.com/olekukonko/tablewriter/pull/74 should land first. Then I can update all the hashes. And resubmit these commits.

knz:
Ack. Thanks

knz:
I think you have a green light now.

RaduBerinde:
Hi @pkit,

The vendor commit has been updated; you will have to rerun the update against the newer one.  You can then push the new commit to the `vendored` repo in a new feature branch (not `master`!). That branch will ensure the commit is accessible. Then, after this PR merges, we can `update vendored/master` to that new commit and delete the feature branch.

See this document for more details: https://github.com/cockroachdb/cockroach/tree/master/build#master-branch-pointer-in-vendored-repo

knz:
Technically, since you don't have permission to push to vendor directly, please create a new branch in your clone of the repo; I'll then copy your branch to the vendor repo so it becomes visible for CI. We'll merge the branch after this PR merges.

knz:
We have found a couple of bugs in our embedded copy of `dep`, which may (or may have not) caused issues when you tried to update the dependencies. These issues are fixed now, so you can probably safely retry now.

nvanbenschoten:
Ping! It would be nice to resolve the associated issue.

knz:
Hi everyone! I have upgraded everything that needed to be upgraded and updated this PR.

@pkit I took the liberty to update your branch with a rebase on top of a more recent version of CockroachDB and fixed the merge conflict. The commit is still all yours!

Merging this when CI lets me.

knz:
@pkit many thanks for your contribution. This definitely increases UX.

jseldess:
@knz, is there any user-facing change here? I can't tell if there are new options to go along with `pretty`.

knz:
@jseldess I reckon that example outputs where you have mixed text/numeric values in tables could benefit from a re-generation to show the new alignment. There is no functional change other than the way the tables appear on the screen.

jseldess:
Got it. Thanks, @knz.

PR #18475, opened 2017-09-13
sql: column equivalency groups

RaduBerinde:
This change adds equivalency groups to orderingInfo, replacing the multi-column
groups in ordering. This addresses a big limitation of the current structure - 
that we are not able to convey equivalency between columns that are not part of
the ordering. This will enable more optimizations, like setting equivalencies
between non-index columns according to filters.

Please ignore the first commit, it's a separate PR (#18461).

Because it's a non-trivial change, I don't plan on merging this for a couple of weeks but I wanted to get some early comments and give a heads-up to whomever might be working with `orderingInfo` code.

CC @arjunravinarayan in case you want to take a look at the changes related to merge-joins.

PR #18492, opened 2017-09-14
util: FastIntSet improvements

RaduBerinde:
#### util: use intsets.Sparse in FastIntSet

Improving the code for FastIntSet to use a `intsets.Sparse` when we have
elements outside of the [0,64) range. We don't just use `Sparse` directly
because it's larger (56 bytes) and must be on the heap (we can't have slices of
`Sparse`).

Also improving the interface to use `int`s and adding a `SubsetOf` method.

#### util: improve FastIntSet with math/bits

Use the new `math/bits` primitives in `FastIntSet`.

#

Please ignore the first commit (separate PR). Note that I will be out for two weeks and don't plan to merge these PRs until I come back, so there is no rush to review.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -635,9 +635,10 @@
   revision = "8be79e1e0910c292df4e79c241bb7e8f7e725959"
 
 [[projects]]
+  branch = "master"
   name = "golang.org/x/tools"
-  packages = ["cmd/goimports","cmd/goyacc","cmd/stringer","cover","go/ast/astutil","go/buildutil","go/gcexportdata","go/gcimporter15","go/loader","go/types/typeutil","imports"]
-  revision = "bf4b54dc687c73b6ef63de8b8abf0ad3951e3edc"
+  packages = ["cmd/goimports","cmd/goyacc","cmd/stringer","container/intsets","cover","go/ast/astutil","go/buildutil","go/gcexportdata","go/gcimporter15","go/loader","go/types/typeutil","imports"]
+  revision = "68e087e2a5786de2c035ed544b1c5a42e31f1933"
 
 [[projects]]
   name = "google.golang.org/api"
@@ -673,6 +674,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "0c4cdc41e60e9405fc2dab5b883635e3c5b6f15afef358d81d488091edea9e40"
+  inputs-digest = "6f894a326b8459636d3168b14fda3ab490b310ec4394b997402e0b437756a92d"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/pkg/sql/expand_plan.go
+++ b/pkg/sql/expand_plan.go
@@ -206,7 +206,7 @@ func doExpandPlan(
 			n.columnsInOrder = make([]bool, len(planColumns(n.plan)))
 			for i := range n.columnsInOrder {
 				group := ordering.eqGroups.Find(i)
-				if ordering.constantCols.Contains(uint32(group)) {
+				if ordering.constantCols.Contains(group) {
 					n.columnsInOrder[i] = true
 					continue
 				}
@@ -504,13 +504,13 @@ func simplifyOrderings(plan planNode, usefulOrdering sqlbase.ColumnOrdering) pla
 			// the useful ordering. Check for this, ignoring any constant columns.
 			sortOrder := make(sqlbase.ColumnOrdering, 0, len(n.ordering))
 			for _, c := range n.ordering {
-				if !constantCols.Contains(uint32(c.ColIdx)) {
+				if !constantCols.Contains(c.ColIdx) {
 					sortOrder = append(sortOrder, c)
 				}
 			}
 			givenOrder := make(sqlbase.ColumnOrdering, 0, len(usefulOrdering))
 			for _, c := range usefulOrdering {
-				if !constantCols.Contains(uint32(c.ColIdx)) {
+				if !constantCols.Contains(c.ColIdx) {
 					givenOrder = append(givenOrder, c)
 				}
 			}
--- a/pkg/sql/join.go
+++ b/pkg/sql/join.go
@@ -601,8 +601,8 @@ func (n *joinNode) joinOrdering() orderingInfo {
 	for i, leftIdx := range n.pred.leftEqualityIndices {
 		leftGroup := leftOrd.eqGroups.Find(leftIdx)
 		rightGroup := rightOrd.eqGroups.Find(n.pred.rightEqualityIndices[i])
-		if leftOrd.constantCols.Contains(uint32(leftGroup)) ||
-			rightOrd.constantCols.Contains(uint32(rightGroup)) {
+		if leftOrd.constantCols.Contains(leftGroup) ||
+			rightOrd.constantCols.Contains(rightGroup) {
 			info.addConstantColumn(leftCol(leftIdx))
 		}
 	}
--- a/pkg/sql/ordering.go
+++ b/pkg/sql/ordering.go
@@ -117,7 +117,7 @@ type orderingInfo struct {
 func (ord orderingInfo) check() {
 	// Only equivalency group representatives show up in constantCols.
 	for c, ok := ord.constantCols.Next(0); ok; c, ok = ord.constantCols.Next(c + 1) {
-		if repr := ord.eqGroups.Find(int(c)); repr != int(c) {
+		if repr := ord.eqGroups.Find(c); repr != c {
 			panic(fmt.Sprintf("non-representative const column %d (representative: %d)", c, repr))
 		}
 	}
@@ -128,13 +128,13 @@ func (ord orderingInfo) check() {
 			panic(fmt.Sprintf("non-representative order column %d (representative: %d)", o.ColIdx, repr))
 		}
 		// The ordering should not contain any constant or redundant columns.
-		if ord.constantCols.Contains(uint32(o.ColIdx)) {
+		if ord.constantCols.Contains(o.ColIdx) {
 			panic(fmt.Sprintf("const column %d appears in ordering", o.ColIdx))
 		}
-		if seen.Contains(uint32(o.ColIdx)) {
+		if seen.Contains(o.ColIdx) {
 			panic(fmt.Sprintf("duplicate column %d appears in ordering", o.ColIdx))
 		}
-		seen.Add(uint32(o.ColIdx))
+		seen.Add(o.ColIdx)
 	}
 	if ord.isKey && len(ord.ordering) == 0 {
 		panic("isKey with no ordering")
@@ -156,8 +156,8 @@ func (ord *orderingInfo) reduce(order sqlbase.ColumnOrdering) sqlbase.ColumnOrde
 	var groupsSeen util.FastIntSet
 	for i, o := range order {
 		group := ord.eqGroups.Find(o.ColIdx)
-		redundant := groupsSeen.Contains(uint32(group)) || ord.constantCols.Contains(uint32(group))
-		groupsSeen.Add(uint32(group))
+		redundant := groupsSeen.Contains(group) || ord.constantCols.Contains(group)
+		groupsSeen.Add(group)
 		if result == nil {
 			if !redundant && o.ColIdx == group {
 				// No modification necessary, continue.
@@ -206,7 +206,7 @@ func (ord *orderingInfo) Format(buf *bytes.Buffer, columns sqlbase.ResultColumns
 		representative := ord.eqGroups.Find(i)
 		if representative != i {
 			// We found a multi-column group.
-			groups.Add(uint32(representative))
+			groups.Add(representative)
 		}
 	}
 
@@ -220,9 +220,9 @@ func (ord *orderingInfo) Format(buf *bytes.Buffer, columns sqlbase.ResultColumns
 	for r, ok := groups.Next(0); ok; r, ok = groups.Next(r + 1) {
 		semiColon()
 		// The representative is always the first element in the group.
-		printCol(buf, columns, int(r))
-		for i := int(r) + 1; i < ord.eqGroups.Len(); i++ {
-			if ord.eqGroups.Find(i) == int(r) {
+		printCol(buf, columns, r)
+		for i := r + 1; i < ord.eqGroups.Len(); i++ {
+			if ord.eqGroups.Find(i) == r {
 				buf.WriteByte('=')
 				printCol(buf, columns, i)
 			}
@@ -273,7 +273,7 @@ func (ord *orderingInfo) isEmpty() bool {
 }
 
 func (ord *orderingInfo) addConstantColumn(colIdx int) {
-	ord.constantCols.Add(uint32(ord.eqGroups.Find(colIdx)))
+	ord.constantCols.Add(ord.eqGroups.Find(colIdx))
 	ord.ordering = ord.reduce(ord.ordering)
 	if len(ord.ordering) == 0 {
 		// TODO(radu): we can only keep track of "keys" for columns in ordering.
@@ -378,9 +378,9 @@ func (ord *orderingInfo) project(colMap []int) orderingInfo {
 	// Remap constant columns, ignoring column groups that have no projected
 	// columns.
 	for col, ok := ord.constantCols.Next(0); ok; col, ok = ord.constantCols.Next(col + 1) {
-		group := ord.eqGroups.Find(int(col))
+		group := ord.eqGroups.Find(col)
 		if r, ok := newRepr[group]; ok {
-			newOrd.constantCols.Add(uint32(newOrd.eqGroups.Find(r)))
+			newOrd.constantCols.Add(newOrd.eqGroups.Find(r))
 		}
 	}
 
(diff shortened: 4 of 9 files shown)


knz:
I did not review the first commit since I believe it comes from #18475. 2nd and 3rd commit LGTM.
I am curious about the performance difference on the ordering computations when switching from small to large int sets...

---

Reviewed 10 of 10 files at r2, 1 of 1 files at r3.
Review status: 10 of 28 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KuI8CuheMfRwcr-LqqW:b-5snu7k) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*
> ```Go
> // Add adds a value to the set. No-op if the value is already in the set.
> func (s *FastIntSet) Add(i int) {
> 	if i >= 0 && i < smallCutoff && s.large == nil {
> ```

Either document in the comment above that negative values are accepted and a no-op, or remove this condition here.

I think I'd rather remove it, and document that negative values are not allowed, and ensure that the callers don't provide one, because I think that where we use this API a negative value here would be a bug.

---

*[pkg/util/fast_int_set.go, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI993dIjxUwJoFZ35R:-KuI993dIjxUwJoFZ35S:bmweq14) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L90)):*
> ```Go
> 		return s.large.Has(i)
> 	}
> 	return i >= 0 && i < smallCutoff && (s.small&(1<<uint64(i))) != 0
> ```

ditto about i >= 0 - either comment above or remove this condition. Here I think probably the condition can stay, I can see why it is useful.

---

*[pkg/util/fast_int_set.go, line 106 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIA66b9faFO0RcdA0r:-KuIA66cekodENClXixm:b-12gbu4) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L106)):*
> ```Go
> 	}
> 	if startVal < smallCutoff {
> 		if startVal < 0 {
> ```

ditto, if a negative value is valid document it above.
I'm not sure why it is useful here, perhaps remind the reader why you did this.

---

*[pkg/util/fast_int_set.go, line 155 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIB4XjtnSjZrYIx6Kh:-KuIB4XjtnSjZrYIx6Ki:b-8k2m1m) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L155)):*
> ```Go
> 		return s.large.AppendTo([]int(nil))
> 	}
> 	// TODO(radu): when we switch to go1.9, use the new math/bits.OnesCount64 to
> ```

Regarding this: it's a bit dangerous, depending on the platform OneCount64 may actually iterate. What we'd really want is the bits package offering this function directly...

---

*[pkg/util/fast_int_set_test.go, line 27 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIBuSlm2_95BVV0bcm:-KuIBuSlm2_95BVV0bcn:bz5a14s) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set_test.go#L27)):*
> ```Go
> func TestFastIntSet(t *testing.T) {
> 	for _, m := range []int{1, 8, 30, smallCutoff, 2 * smallCutoff, 4 * smallCutoff} {
> 		t.Run(fmt.Sprintf("%d", m), func(t *testing.T) {
> ```

nit: t.Parallel()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-:-KuI8s2R4Ezqg-HcZYZu:bsck7yz)*

petermattis:
Review status: 10 of 28 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/util/fast_int_set.go, line 155 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIB4XjtnSjZrYIx6Kh:-KuKGjn2qxbvQpoQzHST:bbfqdn4) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L155)):*

Regarding this: it's a bit dangerous, depending on the platform OneCount64 may actually iterate. What we'd really want is the bits package offering this function directly...
</blockquote></details>

`math/bits.OnesCount64`does [something clever](https://golang.org/src/math/bits/bits.go?s=4586:4616#L124) in the portable implementation, but on the platforms we care about it is implemented by a compiler intrinsic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

andreimatei:
:lgtm: for 2nd and 3rd commits

---

Review status: 10 of 28 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/util/fast_int_set.go, line 50 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuMFZpg-MGkIrAcacag:-KuMFZpg-MGkIrAcacah:baaoagn) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L50)):*
> ```Go
> // Returns the bit encoded set from 0 to 63, and a flag that indicates whether
> // there are elements outside this range.
> func (s *FastIntSet) largeToSmall() (small uint64, otherValues bool) {
> ```

I wonder if this method is really necessary. When I saw it I wondered if we call it after every `Remove()`. Since we don't, the other uses don't seem too impressive to me; we could handle all the comparisons by only working on large sets.

---

*[pkg/util/fast_int_set.go, line 111 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuMHVESZ6rOza5B5bJS:-KuMHC2tUTi0YHPz6k9R:bb7ek88) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b8dae5071c4200b8f7891635cd4bddfbb87003d/pkg/util/fast_int_set.go#L111)):*
> ```Go
> 		}
> 
> 		if ntz := bits.TrailingZeros64(s.small >> uint64(startVal)); ntz < 64 {
> ```

cute

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-:-KuMIAyf21N4Zpkp04rn:b-fxqoee)*

RaduBerinde:
Review status: 10 of 28 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvTzx3bFKtjiiscbv3C:b320nur) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

Either document in the comment above that negative values are accepted and a no-op, or remove this condition here.

I think I'd rather remove it, and document that negative values are not allowed, and ensure that the callers don't provide one, because I think that where we use this API a negative value here would be a bug.
</blockquote></details>

Negative values are allowed, and they're not a no-op; they are treated just like large values (i.e. `large` is used).

---

*[pkg/util/fast_int_set.go, line 155 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIB4XjtnSjZrYIx6Kh:-KvU0UGnqFv8Gimp24tO:b9qd3b5) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L155)):*

`math/bits.OnesCount64`does [something clever](https://golang.org/src/math/bits/bits.go?s=4586:4616#L124) in the portable implementation, but on the platforms we care about it is implemented by a compiler intrinsic.
</blockquote></details>

Not sure what the concern is. It's well defined what`OneCount64` returns and it's not platform dependent.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

RaduBerinde:
Review status: 10 of 28 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/util/fast_int_set.go, line 50 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuMFZpg-MGkIrAcacag:-KvU65SvMvBrNTDlinJW:b-db2i0c) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L50)):*

I wonder if this method is really necessary. When I saw it I wondered if we call it after every `Remove()`. Since we don't, the other uses don't seem too impressive to me; we could handle all the comparisons by only working on large sets.
</blockquote></details>

To work on large sets, we either convert the small to large just for that operation, or we permanently change the representation of that set. I don't like either: the former can incur an allocation on every `Equals/SubsetOf` call, and the latter can cause the representation to change (unfavorably) simply because `Equals/SubsetOf` was called.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

petermattis:
Review status: 10 of 28 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/util/fast_int_set.go, line 155 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIB4XjtnSjZrYIx6Kh:-KvUWOrtdWyVH0tUrGKc:b-h0lt7r) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L155)):*

Not sure what the concern is. It's well defined what`OneCount64` returns and it's not platform dependent.
</blockquote></details>

I think @knz was concerned that `bits.OnesCount64` might be implemented using a naive iteration that is O(num-bits). That is, the concern was about the resulting performance, not correctness. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

RaduBerinde:
Review status: 0 of 28 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/util/fast_int_set_test.go, line 27 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIBuSlm2_95BVV0bcm:-KvXel0oRq71z7yQ7vPu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set_test.go#L27)):*

nit: t.Parallel()
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

knz:
Please merge #18475 and rebase this before I can complete the review.

RaduBerinde:
Rebased.

---

Review status: 0 of 10 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-:-KvXksJRP5nayL6W8HtO:bncvegj)*

knz:
Reviewed 8 of 10 files at r5, 1 of 2 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvXlNR16BM0cxlznpvY:bfri148) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

Negative values are allowed, and they're not a no-op; they are treated just like large values (i.e. `large` is used).
</blockquote></details>

Huh then where in the rest code do you assert that no invalid column index (-1) is added to a set?

---

*[pkg/util/fast_int_set.go, line 155 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuIB4XjtnSjZrYIx6Kh:-KvXlWO7ZPzodldsBktV:b-7aitaz) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L155)):*

I think @knz was concerned that `bits.OnesCount64` might be implemented using a naive iteration that is O(num-bits). That is, the concern was about the resulting performance, not correctness. 
</blockquote></details>

Thanks for clarifying.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

RaduBerinde:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvXlpTRJAJU987LrplM:b-ca1ow5) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

Huh then where in the rest code do you assert that no invalid column index (-1) is added to a set?
</blockquote></details>

We don't.. we rely on test cases to verify that we populate columns correctly.

I wanted to keep this structure general-purpose, and make it no less functional than `intsets.Sparse`. But if you feel strongly, I can require non-negative indices.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

knz:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvXmfzSxD6VBDqucKac:b-c5np0i) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

We don't.. we rely on test cases to verify that we populate columns correctly.

I wanted to keep this structure general-purpose, and make it no less functional than `intsets.Sparse`. But if you feel strongly, I can require non-negative indices.
</blockquote></details>

I can see the value of supporting negative values -- we could use that to implement other things later. 

However in the previous code it was easy to catch an invalid column index, because Go would refuse the indexing of a slice and issue a panic. In the new code I fear that negative indexes will be propagated longer, more silently. Perhaps it is worth asserting this in the caller of Add() instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

RaduBerinde:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvXoSh8TrxEHOFpxTO7:b-l6xzja) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

I can see the value of supporting negative values -- we could use that to implement other things later. 

However in the previous code it was easy to catch an invalid column index, because Go would refuse the indexing of a slice and issue a panic. In the new code I fear that negative indexes will be propagated longer, more silently. Perhaps it is worth asserting this in the caller of Add() instead.
</blockquote></details>

There are many callers of Add.. Are there any specific ones that are concerning? In most cases what we pass to `Add` is the result of `eqGroups.Find()` which does panic on -1.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

knz:

---

*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvXpbgUt6-XV5bjalzP:b-lfxhrq) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

There are many callers of Add.. Are there any specific ones that are concerning? In most cases what we pass to `Add` is the result of `eqGroups.Find()` which does panic on -1.
</blockquote></details>

Your response contains the answer: the caller of `Add` that do not use the result of `eqGroups.Find`, if any. Likewise you can exclude the callers that already assert the value is non-negative by other means (e.g. via `len` or others).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

RaduBerinde:
*[pkg/util/fast_int_set.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492#-KuI8CuheMfRwcr-LqqV:-KvXqTlPK4EA_rF1yW6y:btkdmh4) ([raw file](https://github.com/cockroachdb/cockroach/blob/850aa2e222f7afda39c05206fa8b658fdd47bab9/pkg/util/fast_int_set.go#L62)):*

Your response contains the answer: the caller of `Add` that do not use the result of `eqGroups.Find`, if any. Likewise you can exclude the callers that already assert the value is non-negative by other means (e.g. via `len` or others).
</blockquote></details>

Ok, I did an audit and it looks like all cases do `Find` first. There's a single exception in RowFetcher.Init where the value is a range loop variable which is always non-negative.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18492)*

knz:
Thanks for checking.

LGTM (modulo addressing the linter errors)

PR #18494, opened 2017-09-14
sql: fix TRUNCATE on a table in the middle of a schema change

vivekmenezes:
TRUNCATE drops the existing table and creates a new one with the
same schema. In doing so it currently copies over all the schema
change mutations and jobs associated with the schema change thus
associating the same job with two tables. This is broken!

This change associates all the mutations and jobs with the table being
dropped, and converts all outstanding schema changes on the new table
to the public state. The new table is empty and thus doesn't need
a backfill job.

### DIFF
--- a/pkg/sql/truncate.go
+++ b/pkg/sql/truncate.go
@@ -177,6 +177,19 @@ func (p *planner) truncateTable(ctx context.Context, id sqlbase.ID, traceKV bool
 	if err := newTableDesc.SetUpVersion(); err != nil {
 		return err
 	}
+	// Resolve all outstanding mutations. Make all new schema elements
+	// public because the table is empty and doesn't need to be backfilled.
+	for _, m := range newTableDesc.Mutations {
+		if m.Direction == sqlbase.DescriptorMutation_ADD {
+			if col := m.GetColumn(); col != nil {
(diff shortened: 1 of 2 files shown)


pkg/sql/truncate.go
```diff
@@ -177,6 +177,19 @@ func (p *planner) truncateTable(ctx context.Context, id sqlbase.ID, traceKV bool
 	if err := newTableDesc.SetUpVersion(); err != nil {
 		return err
 	}
+	// Resolve all outstanding mutations. Make all new schema elements
+	// public because the table is empty and doesn't need to be backfilled.
+	for _, m := range newTableDesc.Mutations {
+		if m.Direction == sqlbase.DescriptorMutation_ADD {
```
knz: What about mutations that drop a column? These columns must be removed from the new descriptor.

vivekmenezes: Those columns/indexes have already been removed from the descriptor and placed in the mutations. They are ignored here implying that they get dropped.

vivekmenezes: added a test for this case

jordanlewis (approved):

knz (approved):

jordanlewis:
Are you going to cherrypick this?

vivekmenezes:
yes will do

PR #18496, opened 2017-09-14
build: TestProtoMarshal correctly avoids Marshal methods

tamird:
Also use `git grep` exclusions rather than filtering its results.

### DIFF
--- a/build/style_test.go
+++ b/build/style_test.go
@@ -129,38 +129,72 @@ func TestStyle(t *testing.T) {
 
 	t.Run("TestEnvutil", func(t *testing.T) {
 		t.Parallel()
-		cmd, stderr, filter, err := dirCmd(pkg.Dir, "git", "grep", "-nE", `os\.(Getenv|LookupEnv)`, "--", "*.go")
-		if err != nil {
-			t.Fatal(err)
-		}
+		for _, tc := range []struct {
+			re       string
+			excludes []string
+		}{
+			{re: `\bos\.(Getenv|LookupEnv)\("COCKROACH`},
+			{
+				re: `\bos\.(Getenv|LookupEnv)\(`,
+				excludes: []string{
+					":!acceptance",
+					":!build/style_test.go",
+					":!ccl/acceptanceccl/backup_test.go",
+					":!ccl/sqlccl/backup_cloud_test.go",
+					":!ccl/storageccl/export_storage_test.go",
+					":!cmd",
+					":!util/envutil/env.go",
+					":!util/log/clog.go",
+					":!util/log/color.go",
+					":!util/sdnotify/sdnotify_unix.go",
+				},
+			},
+		} {
+			cmd, stderr, filter, err := dirCmd(
+				pkg.Dir,
+				"git",
+				append([]string{
+					"grep",
+					"-nE",
+					tc.re,
+					"--",
+					"*.go",
+				}, tc.excludes...)...,
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
 
-		if err := cmd.Start(); err != nil {
-			t.Fatal(err)
-		}
+			if err := cmd.Start(); err != nil {
+				t.Fatal(err)
+			}
 
-		if err := stream.ForEach(stream.Sequence(
-			filter,
-			stream.GrepNot(`^cmd(/.*)?/\w+\.go\b`),
-			stream.GrepNot(`^build/style_test\.go\b`),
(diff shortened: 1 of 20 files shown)


benesch:
This is on my radar! Just still catching up from vacation.

tamird:
Rebased to fix a merge conflict.

nvanbenschoten:
:lgtm:

@CLAassistant where you at?

---

Reviewed 20 of 20 files at r1.

---

*[build/style_test.go, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18496#-KuSbJWc9Fp3NCPphJuy:-KuSbJWc9Fp3NCPphJuz:bs474mn) ([raw file](https://github.com/cockroachdb/cockroach/blob/667ce2d2138bcc0c334d4bc99953b414c280f438/build/style_test.go#L139)):*
> ```Go
> 				re: `\bos\.(Getenv|LookupEnv)\(`,
> 				excludes: []string{
> 					":!acceptance",
> ```

Where does a layman find documentation on this `:!` exclusion syntax?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18496#-:-KuScng906XHgkeEU5YK:b-8al1yx)*

tamird:
Good question, I think I got it from some SO thread. Thanks!

tamird:
Rebased to avoid skew; looks like acceptance hung, and my teamcity privileges have been revoked :(

Can someone rebuild for me?

benesch:
Queued you a new build!

---

Review status: 19 of 20 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[build/style_test.go, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18496#-KuSbJWc9Fp3NCPphJuy:-KuUCrJX5demsx5pCyXr:b-fztly7) ([raw file](https://github.com/cockroachdb/cockroach/blob/667ce2d2138bcc0c334d4bc99953b414c280f438/build/style_test.go#L139)):*

Where does a layman find documentation on this `:!` exclusion syntax?
</blockquote></details>

https://git-scm.com/docs/gitglossary/2.13.1#gitglossary-aiddefpathspecapathspec

(Git pathspecs are _super_ powerful.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18496#-:-KuUCYifkqzT-cMsZnnT:b2qy5dn)*

build/style_test.go
```diff
@@ -129,38 +129,72 @@ func TestStyle(t *testing.T) {
...
+		for _, tc := range []struct {
+			re       string
+			excludes []string
+		}{
+			{re: `\bos\.(Getenv|LookupEnv)\("COCKROACH`},
+			{
+				re: `\bos\.(Getenv|LookupEnv)\(`,
+				excludes: []string{
+					":!acceptance",
+					":!build/style_test.go",
```
benesch: This isn't in `pkg`, is it? Why would it get picked up?

build/style_test.go
```diff
@@ -294,7 +355,17 @@ func TestStyle(t *testing.T) {
...
+			pkg.Dir,
+			"git",
+			"grep",
+			"-nE",
+			`\.Marshal\(`,
+			"--",
+			"*.go",
+			":!util/protoutil/marshal.go",
+			":!settings/settings_test.go",
+		)
```
benesch: This pattern really seems to deserve a `grepExcluding(`regexpattern`, "file", "file2", ...)` function, but this is an improvement as is so no need to hold up this PR.

build/style_test.go
```diff
@@ -649,6 +727,9 @@ func TestStyle(t *testing.T) {
 	// TODO(tamird): replace this with errcheck.NewChecker() when
 	// https://github.com/dominikh/go-tools/issues/57 is fixed.
 	t.Run("TestErrCheck", func(t *testing.T) {
+		if testing.Short() {
+			t.Skip("short flag")
+		}
```
benesch: Any particular reason you're sidecar-ing this bit?

benesch (commented):
(Sorry for the post-hoc review. If you don't mind responding to my comments with your thoughts I'm happy to pick off any changes myself.)

build/style_test.go
tamird: I think `git grep` doesn't care about the directory you're in, or at least it doesn't when you specify `'*.go'`.

build/style_test.go
tamird: Eh, I'm not sure I agree that there's much benefit to the indirection. Feel free to follow up if you feel strongly enough.

PR #18510, opened 2017-09-15
sql/sqlbase: perf fix to fks on interleave parents

jordanlewis:
The foreign key checker previously used an overly-wide span to scan to
check for foreign key existence that didn't take into account the fact
that there could tables interleaved into the foreign key table.

This patch tightens the span to include only those keys that are related
to the parent table's row.

For example, consider the following scenario:

```
CREATE TABLE t (a INT PRIMARY KEY);
CREATE TABLE u (a INT, b INT,
                FOREIGN KEY(a) REFERENCES parent(a),
                PRIMARY KEY (a, b))
               INTERLEAVE IN PARENT(t);
```

Inserts into `u` with, for example, `a = 2` need to check `t` to ensure
that a row with `a = 2` exists. Previously, this check was performed
with a scan over the span `.../2 -> .../3` where `...` represents the
index prefix for `t`. The size of this span increases as more rows are
added to `u`. This is not a problem for the actual scan performance in
this case, since the check will stop the scan after a key is found, but
it is a problem for the command queue, since it'll have dependencies
against other operations against keys in `u`.

This patch tightens the example span above to `.../2 -> .../2/<interleave sentinel>`.

This restores insert performance for TPCC to an acceptable level.

### DIFF
--- a/pkg/sql/sqlbase/fk.go
+++ b/pkg/sql/sqlbase/fk.go
@@ -374,12 +374,12 @@ func (f baseFKHelper) spanForValues(values parser.Datums) (roachpb.Span, error) 
 	} else {
 		key = roachpb.Key(f.searchPrefix)
 	}
-	return roachpb.Span{Key: key, EndKey: key.PrefixEnd()}, nil
+	return roachpb.Span{Key: key, EndKey: InterleaveEnd(key)}, nil
 }
 
 func (f baseFKHelper) span() roachpb.Span {
 	key := roachpb.Key(f.searchPrefix)
-	return roachpb.Span{Key: key, EndKey: key.PrefixEnd()}
+	return roachpb.Span{Key: key, EndKey: InterleaveEnd(key)}
 }
 
 // FkSpanCollector can collect the spans that foreign key validation will touch.
--- a/pkg/sql/sqlbase/table.go
+++ b/pkg/sql/sqlbase/table.go
@@ -331,6 +331,13 @@ func EncodePartialIndexKey(
 	return key, containsNull, err
 }
 
+// InterleaveEnd returns the key at the end of the current interleave segment.
+// This is useful for constructing the end of the span over all keys for a single
+// row, without including any rows from tables interleaved into that row.
+func InterleaveEnd(key []byte) []byte {
+	return encoding.EncodeNotNullDescending(key)
+}
+
 type directions []IndexDescriptor_Direction
 
 func (d directions) get(i int) (encoding.Direction, error) {
--- a/pkg/sql/logictest/testdata/logic_test/interleaved
+++ b/pkg/sql/logictest/testdata/logic_test/interleaved
@@ -279,6 +279,9 @@ INSERT INTO orders
   (1, 1002, 80.00),
   (2, 1003, 70.00)
 
+statement error foreign key violation: value \[3\] not found in customers@primary \[id\]
+INSERT INTO orders VALUES (3, 1000, 100.0)
+
 query IIR
 SELECT * FROM orders WHERE customer = 1 AND id = 1000
 ----
--- a/pkg/sql/sqlbase/structured.pb.go
+++ b/pkg/sql/sqlbase/structured.pb.go
@@ -485,7 +485,7 @@ func (*InterleaveDescriptor) ProtoMessage()               {}
 func (*InterleaveDescriptor) Descriptor() ([]byte, []int) { return fileDescriptorStructured, []int{4} }
 
 type InterleaveDescriptor_Ancestor struct {
-	// TableID the ID of the table being interleaved into.
+	// TableID is the ID of the table being interleaved into.
 	TableID ID `protobuf:"varint,1,opt,name=table_id,json=tableId,casttype=ID" json:"table_id"`
 	// IndexID is the ID of the parent index being interleaved into.
 	IndexID IndexID `protobuf:"varint,2,opt,name=index_id,json=indexId,casttype=IndexID" json:"index_id"`
--- a/pkg/sql/sqlbase/structured.proto
+++ b/pkg/sql/sqlbase/structured.proto
@@ -154,7 +154,7 @@ message ColumnFamilyDescriptor {
 // grandparent) with shared lengths 2 and 1.
 message InterleaveDescriptor {
   message Ancestor {
-    // TableID the ID of the table being interleaved into.
+    // TableID is the ID of the table being interleaved into.
     optional uint32 table_id = 1 [(gogoproto.nullable) = false,
         (gogoproto.customname) = "TableID", (gogoproto.casttype) = "ID"];
     // IndexID is the ID of the parent index being interleaved into.


pkg/sql/sqlbase/table.go
```diff
@@ -331,6 +331,13 @@ func EncodePartialIndexKey(
 	return key, containsNull, err
 }
 
+// InterleaveEnd returns the key at the end of the current interleave segment.
+// This is useful for constructing the end of the span over all keys for a single
+// row, without including any rows from tables interleaved into that row.
+func InterleaveEnd(key []byte) []byte {
+	return encoding.EncodeNotNullDescending(key)
+}
```
vivekmenezes: Mind adding a unittest for this function

pkg/sql/sqlbase/fk.go
```diff
@@ -374,12 +374,12 @@ func (f baseFKHelper) spanForValues(values parser.Datums) (roachpb.Span, error)
...
 	}
-	return roachpb.Span{Key: key, EndKey: key.PrefixEnd()}, nil
+	return roachpb.Span{Key: key, EndKey: InterleaveEnd(key)}, nil
 }
 
 func (f baseFKHelper) span() roachpb.Span {
 	key := roachpb.Key(f.searchPrefix)
-	return roachpb.Span{Key: key, EndKey: key.PrefixEnd()}
+	return roachpb.Span{Key: key, EndKey: InterleaveEnd(key)}
 }
```
vivekmenezes: I hope we have some logictests for FK over interleaved tables.

vivekmenezes (commented):
Really nice win for the size of this change. Nice line of thinking! It would be great if you could add a benchmark. Thanks!

knz:
cc @arjunravinarayan

danhhz:
:lgtm_strong: you found it!

style nit: our commit messages are in the imperative tone `fix perf of fks to interleaved tables`

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/table.go, line 337 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-Ku5iBHO9CU7mwXWM1i4:-Ku5iBHO9CU7mwXWM1i5:b-ao28z2) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a4ac7a6f43b63b3b3ffba554b909ae2b173f04d/pkg/sql/sqlbase/table.go#L337)):*
> ```Go
> // This is useful for constructing the end of the span over all keys for a single
> // row, without including any rows from tables interleaved into that row.
> func InterleaveEnd(key []byte) []byte {
> ```

The doc on this method could use some work. We don't use the term "interleave segment" anywhere else, so I'd nix that. It also doesn't define what parts of the key should be in the input. I think the answer is everything up to (but not including) the column family, which I usually call a "row prefix". I'd also consider making this return a roachpb.Span instead, it may help make the intention more clear. Maybe something like "RowWithoutInterleavesSpan when given an encoded row prefix (everything up to but not including the column family) returns a span that contains all data for the row but without any rows from interleaved children."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-:-Ku5hvEbYIYTrjPX2RGG:b-secd3z)*

RaduBerinde:
*[pkg/sql/sqlbase/fk.go, line 377 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-Ku5jivOTpdXLjnLGmZf:-Ku5jivOTpdXLjnLGmZg:b-bbo9y) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a4ac7a6f43b63b3b3ffba554b909ae2b173f04d/pkg/sql/sqlbase/fk.go#L377)):*
> ```Go
> 		key = roachpb.Key(f.searchPrefix)
> 	}
> 	return roachpb.Span{Key: key, EndKey: InterleaveEnd(key)}, nil
> ```

How does this work if we only have values for a prefix of the index cols? Say we have an index on `(a, b, c)` and we are creating a span for `/a/b`. I don't think appending the interleave sentinel in that case makes sense?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510)*

danhhz:
also, would you mind filing an issue mentioning that we should audit uses of PrefixEnd? I suspect many of them are wrong (either due to this or because they should be .Next which is also a tighter bound than PrefixEnd)

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-:-Ku5k0GhkSCiSZlO6wH6:b-508j3p)*

jordanlewis:
Yep, I've been looking over usages of `PrefixEnd` already. There's a large number of them and I agree that it's likely there are similar problem lurking around. I'll make an issue.

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/fk.go, line 377 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-Ku5jivOTpdXLjnLGmZf:-Ku5l34d0pa9p1oPh_Rs:b-ou20t0) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a4ac7a6f43b63b3b3ffba554b909ae2b173f04d/pkg/sql/sqlbase/fk.go#L377)):*

How does this work if we only have values for a prefix of the index cols? Say we have an index on `(a, b, c)` and we are creating a span for `/a/b`. I don't think appending the interleave sentinel in that case makes sense?
</blockquote></details>

In that case it's still okay to append the interleave sentinel (I think?) because it deliberately sorts next-to-last behind `encodedNullDesc`, which AFAIK won't appear in index keys. Is that correct? @danhhz does that make sense?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-:-Ku5lWwi4eIEqylnzioi:b6304wm)*

RaduBerinde:
*[pkg/sql/sqlbase/fk.go, line 377 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-Ku5jivOTpdXLjnLGmZf:-Ku5qdDuRQ4Jjkj6uD3l:bs4mjjh) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a4ac7a6f43b63b3b3ffba554b909ae2b173f04d/pkg/sql/sqlbase/fk.go#L377)):*

In that case it's still okay to append the interleave sentinel (I think?) because it deliberately sorts next-to-last behind `encodedNullDesc`, which AFAIK won't appear in index keys. Is that correct? @danhhz does that make sense?
</blockquote></details>

Perhaps but that's quite hacky. What's the difference between this and `PrefixEnd` in this case anyway?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510)*

jordanlewis:
Yeah, there are a lot of obvious similar problems. For example, `SELECT * from interleave_parent WHERE primary_key=something` is extremely slow when there are lot of rows in the child. It's clear that the span is too wide here as well. I think we should try to fix all of these at once, so I'll be amending this PR.

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/fk.go, line 377 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-Ku5jivOTpdXLjnLGmZf:-Ku5ujmpxcMUq4-yGpOD:bdhft6u) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a4ac7a6f43b63b3b3ffba554b909ae2b173f04d/pkg/sql/sqlbase/fk.go#L377)):*

Perhaps but that's quite hacky. What's the difference between this and `PrefixEnd` in this case anyway?
</blockquote></details>

Shouldn't be any difference. I'll improve this code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18510#-:-Ku5uvDwHyNhtmomaxaW:b-f16ijg)*

PR #18512, opened 2017-09-15
*: use production Raft settings in TestStoreConfig

### LINKED ISSUE #18200: storage: bring TestStoreConfig closer to the production config
From [a comment](https://github.com/cockroachdb/cockroach/pull/18187#issuecomment-326993288) on #18187:

> Many of the tests that would be sensitive to premature raft elections have their own config overrides. Maybe there aren't that many tests left that rely on a short election timeout and we can bring TestStoreConfig closer to the production config.

In particular, it would be great to use the production Raft election and tick settings. The low settings were necessary early in development in order to make tests run quickly. But we've been proactively campaigning for a while and it is possible that the production settings won't affect test times. Additionally the low settings might be obscuring badness that would otherwise be revealed by a test taking longer than expected.

petermattis:
Use the production Raft settings in most tests so that we can discover
and fix problems associated with them. Add storage.TestFastStoreConfig
and base.TestFastRaftConfig for tests which exercise Raft timeouts and
failover and thus want lower settings than used in production.

Fixes #18200

### DIFF
--- a/pkg/base/config.go
+++ b/pkg/base/config.go
@@ -55,6 +55,10 @@ const (
 	// defaultRaftTickInterval is the default resolution of the Raft timer.
 	defaultRaftTickInterval = 200 * time.Millisecond
 
+	// defaultRaftHeartbeatIntervalTicks is the default number of ticks between
+	// Raft heartbeats.
+	defaultRaftHeartbeatIntervalTicks = 5
+
 	// defaultRangeLeaseRaftElectionTimeoutMultiplier specifies what multiple the
 	// leader lease active duration should be of the raft election timeout.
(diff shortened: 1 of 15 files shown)


petermattis:
@bdarnell This uses the fast Raft config settings in too many tests. I need to investigate why some of the tests are slowing down even though they aren't exercising Raft timeouts or failover.

bdarnell:
:lgtm:

---

Reviewed 15 of 15 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18512#-:-KuKoAa_v45wiSfZmk7w:bnfp4nl)*

PR #18513, opened 2017-09-15
build: clean up after nightlies

benesch:
Add a script to be run from TeamCity's "all nightly tests" job that
deletes and recreates the `cockroach-nightly` resource group. This is
the only surefire way to clean up any VMs/virtual networks/IPs that are
leaked by the nightly tests; relying on each test to run `terraform
destroy` is not sufficient.

### DIFF
--- a/build/teamcity-reset-nightlies.sh
+++ b/build/teamcity-reset-nightlies.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Deletes and recreates the cockroach-nightly resource group, which contains the
+# Azure VMs, virtual networks, IPs, etc. that are automatically created by
+# nightly tests. This is a big hammer, but the nightlies leak resources
+# constantly as `terraform destroy` is not particularly reliable.
+
+set -euo pipefail
+
+docker run --rm --interactive azuresdk/azure-cli-python:2.0.17 bash <<EOF
+  set -euo pipefail
+
+  az login --service-principal --username="$ARM_CLIENT_ID" --password="$ARM_CLIENT_SECRET" --tenant="$ARM_TENANT_ID"
+  az account set --subscription="$ARM_SUBSCRIPTION_ID"
+
+  # Simulate delete-if-exists behavior.
+  az group delete --yes --no-wait --name=cockroach-nightly || true
+  az group wait --deleted --timeout=$((60 * 10)) --name=cockroach-nightly
+
+  az group create --location=eastus --name=cockroach-nightly
+EOF


jordanlewis:

---

*[build/teamcity-reset-nightlies.sh, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513#-Ku5x7nQFOCBIrED0q45:-Ku5x7nQFOCBIrED0q46:b9c4lsx) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f1b0296ee9e5d759951cb634737689b68763247/build/teamcity-reset-nightlies.sh#L20)):*
> ```Shell
>   az group wait --deleted --timeout=$((60 * 10)) --name=cockroach-nightly
> 
>   az group create --location=eastus --name=cockroach-nightly
> ```

Seems somewhat error prone to have this run with set -e - maybe add this create to an on exit trap with a `||true` to ensure that we at least always try to create this before the next run of nightlies?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513)*

benesch:

---

*[build/teamcity-reset-nightlies.sh, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513#-Ku5x7nQFOCBIrED0q45:-Ku601_8ypHFzi7wFDp3:b-q1y7k2) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f1b0296ee9e5d759951cb634737689b68763247/build/teamcity-reset-nightlies.sh#L20)):*

Seems somewhat error prone to have this run with set -e - maybe add this create to an on exit trap with a `||true` to ensure that we at least always try to create this before the next run of nightlies?
</blockquote></details>

Hmm, so my reasoning was that if `az group wait --deleted` fails, it's because the resource group still exists, in which case creating is pointless. I guess it can't hurt to try to create it anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513)*

jordanlewis:
Well either way this is strictly better. LGTM, your choice about whether to address that comment. Probably should merge this sooner than later 👍

cuongdo:
:lgtm:

In a different PR, could you also amend this script to recreate the `cockroachnightlyvhd` storage account? There are tons of leaked VHDs in there, and that's got to be costing us.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513#-:-KuKVT7ukvQK-DA4Ri1u:bhssz15)*

cuongdo:
*[build/teamcity-reset-nightlies.sh, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513#-KuKUuDU-2m4E7Ca9K0o:-KuKUuDU-2m4E7Ca9K0p:bdv8nst) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f1b0296ee9e5d759951cb634737689b68763247/build/teamcity-reset-nightlies.sh#L17)):*
> ```Shell
> 
>   # Simulate delete-if-exists behavior.
>   az group delete --yes --no-wait --name=cockroach-nightly || true
> ```

maybe extract `cockroach-nightly` into a variable somewhere above?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18513)*

benesch:
Whoops, dropped this. Merging, then fast following with another PR.

PR #16907, opened 2017-07-06
build: avoid spurious recompiles

benesch:
See individual commits for details. Only the first one is necessary to avoid spurious recompiles; the latter just makes a bare `go build` not print mountains of garbage.

PR #18514, opened 2017-09-15
build: run vet and stringer from source

benesch:
@RaduBerinde since you pulled the trigger on Go 1.9 figured I'd send this your way. Let me know if you'd rather I send this to someone else. (This is extracted from #16907.)

Prior to Go 1.9, Go static analysis tools, like `go vet` and `stringer`,
loaded packages using the `gcimporter`, which would extract type
information from installed artifacts. This was a colossally bad idea,
as type errors would appear or disappear after a `go install`.

Go 1.9 taught vet and stringer to instead use the `srcimporter`, which
ignores the installed artifacts and vets from the source code instead.
This means `make generate` and `make lint` no longer need to depend on
`gotestdashi`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,30 +284,21 @@ dupl: $(BOOTSTRAP_TARGET)
 	       -not -name '*.ir.go'     \
 	| dupl -files $(DUPLFLAGS)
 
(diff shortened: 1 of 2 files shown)


RaduBerinde:
Cool! LGTM

Issue #7881, opened 2016-07-18
stability: nil pointer panic in client.Txn.CleanupOnError


bdarnell:
On the register cluster running beta-20160629, two nodes have failed (ten minutes apart) with the following panic:

```
E160717 06:28:28.001279 internal/client/txn.go:364  failure aborting transaction: does not exist; abort caused by: does not exist
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x140 pc=0xb685a0]

goroutine 3094076 [running]:
panic(0x18ba060, 0xc82000e100)
        /usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/cockroachdb/cockroach/internal/client.(*Txn).sendEndTxnReq(0x0, 0xc830067400, 0x0, 0x0, 0x0)
        /go/src/github.com/cockroachdb/cockroach/internal/client/txn.go:440 +0x50
github.com/cockroachdb/cockroach/internal/client.(*Txn).Rollback(0x0, 0x0, 0x0)
        /go/src/github.com/cockroachdb/cockroach/internal/client/txn.go:433 +0x3b
github.com/cockroachdb/cockroach/internal/client.(*Txn).CleanupOnError(0x0, 0x7f09906bda70, 0xc835e66ba0)
        /go/src/github.com/cockroachdb/cockroach/internal/client/txn.go:363 +0x92
github.com/cockroachdb/cockroach/sql.(*Executor).execRequest(0xc8201a3860, 0x7f099136cac8, 0xc823552780, 0xc82351d000, 0xc820a519cb, 0x27, 0x0, 0x0, 0x0, 0xc820a51900)
        /go/src/github.com/cockroachdb/cockroach/sql/executor.go:502 +0xd07
github.com/cockroachdb/cockroach/sql.(*Executor).ExecuteStatements(0xc8201a3860, 0x7f099136cac8, 0xc823552780, 0xc82351d000, 0xc820a519cb, 0x27, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/github.com/cockroachdb/cockroach/sql/executor.go:361 +0xf6
github.com/cockroachdb/cockroach/sql/pgwire.(*v3Conn).executeStatements(0xc826b4c000, 0x7f099136cac8, 0xc823552780, 0xc820a519cb, 0x27, 0x0, 0x0, 0x0, 0x0, 0x7f09913b7201, ...)
[...]

PR #18515, opened 2017-09-15
cherrypick: don't auto-retry if we've already sent results to the client

### LINKED ISSUE #17592: sql: panic: cannot reset if we've already sent results for this txn
panic while executing "SHOW KV TRACE FOR DROP USER IF EXISTS TIME , REPEATABLE , VARCHAR , ident , FILTER , NAME": cannot reset if we've already sent results for this txn

https://teamcity.cockroachdb.com/viewLog.html?buildId=320932&buildTypeId=Cockroach_Nightlies_RandomSyntaxTests

Found with RSG.

andreimatei:
Cherrypicking #18268 for 1.1.

Fixes #17592

Ever since results streaming, the condition about whether we can
auto-retry or not has become more complicated: we can't auto-retry if
we've previously streamed any results to the client. However, we were
ignoring this fact and retrying regardless. This patch adds this
condition.

### DIFF
--- a/pkg/internal/client/db_test.go
+++ b/pkg/internal/client/db_test.go
@@ -407,6 +407,7 @@ func TestCommonMethods(t *testing.T) {
 		{txnType, "IsSerializableRestart"}:           {},
 		{txnType, "NewBatch"}:                        {},
 		{txnType, "Exec"}:                            {},
+		{txnType, "PrepareForRetry"}:                 {},
 		{txnType, "ResetDeadline"}:                   {},
 		{txnType, "Run"}:                             {},
 		{txnType, "Send"}:                            {},
@@ -427,6 +428,8 @@ func TestCommonMethods(t *testing.T) {
 		{txnType, "UserPriority"}:                    {},
 		{txnType, "AnchorKey"}:                       {},
 		{txnType, "ID"}:                              {},
+		{txnType, "IsAborted"}:                       {},
+		{txnType, "IsCommitted"}:                     {},
 	}
 
 	for b := range omittedChecks {
--- a/pkg/internal/client/txn.go
+++ b/pkg/internal/client/txn.go
@@ -154,10 +154,21 @@ func (txn *Txn) IsFinalized() bool {
 	return txn.mu.finalized
 }
 
+// status returns the txn proto status field.
 func (txn *Txn) status() roachpb.TransactionStatus {
 	txn.mu.Lock()
 	defer txn.mu.Unlock()
 	return txn.mu.Proto.Status
+}
+
+// IsCommitted returns true if the transaction has the committed status.
+func (txn *Txn) IsCommitted() bool {
+	return txn.status() == roachpb.COMMITTED
+}
+
+// IsAborted returns true if the transaction has the aborted status.
+func (txn *Txn) IsAborted() bool {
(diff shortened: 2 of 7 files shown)


andreimatei:
Cherry-picking had no conflicts.

knz (approved):

nvanbenschoten:
This is a large cherry-pick. I think we should run it by @bdarnell as well before merging.

andreimatei:
cc @cockroachdb/release 

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18515#-:-KuKfPgfjGcwnKY2BUpV:biv7wfu)*

petermattis:
I think we should go ahead with this cherry-pick. My rationale is that we'd certainly want to fix this bug in a patch release if a user encountered it. The only question I have is whether there is a smaller change that would accomplish the fix and if such a smaller change is less risky. 

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18515#-:-KuKjcC8kco3fmfimuwI:b-tw6rh6)*

andreimatei:
> The only question I have is whether there is a smaller change that would accomplish the fix and if such a smaller change is less risky.

Perhaps there's a change with fewer characters in it, but the way I see it, all the refactoring here reduces risk, not increases it. We already know that there was something broken with the interaction between the Executor and the txn interface (#7881) that we never tracked down; this patch removes a number of broken things that were happening there.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18515#-:-KuKls5J5z2nEchYb4v5:btj1k1r)*

bdarnell:
:lgtm: 

I think if we were cherry-picking for 1.1.1 I'd insist on a smaller change (can we just change the state from FirstBatch when we start streaming results), but I'm OK with this now.

Unless reviewable is confusing me, r5 contains only comment/test changes, and no changes to production code. 

---

Reviewed 1 of 1 files at r1, 3 of 6 files at r3.
Review status: 4 of 7 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18515#-:-KuKm_aj522NYop6iK5R:b-8pdmct)*

andreimatei:
> Unless reviewable is confusing me, r5 contains only comment/test changes, and no changes to production code.

Uf, this was me messing up a rebase. The actual fix moved to a previous commit.
The way that bug fix commit was intended can be seen at an older version (r10 in #18268):
https://github.com/cockroachdb/cockroach/commit/b896d3df2657e1cca43098b1a0dcaaf8235543a2

---

Review status: 4 of 7 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18515#-:-KuKrGG6KPPWQYvPLyKp:bmlbnta)*

PR #18521, opened 2017-09-16
server: support inter-node tunneling with logSpy

nvb:
This change adds the ability for `logSpy` requests to redirect to other
nodes and stream back log messages. This should be useful whenever a cluster
does not expose HTTP servers for all nodes.

In doing this, the change moves the logSpy functionality to the
`statusServer`. This means that it changes the URL from `/debug/logspy`
to `/_status/logspy/{node_id}`. I think this is a better URL because it
mirrors already existing URLs like `/_status/logs/{node_id}` and
`/_status/logfiles/{node_id}`.

Because logSpy is now accessed through grpc/grpc-gateway, a lot of the
old tests became unnecessary and could be removed.


pkg/server/status_logspy.go
```diff
@@ -0,0 +1,140 @@
...
+		case entries <- entry:
+		default:
+			// Consumer fell behind, just drop the message.
+			atomic.AddInt32(&countDropped, 1)
+		}
+	})
+	defer setInterceptOrErr(nil)
+
+	var count int
+	var done <-chan struct{} // set later; helps always send header message
```
tschottdorf: not your code, but grammar in comment is off. Might as well.

pkg/server/status_logspy.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2017 The Cockroach Authors.
```
tschottdorf: Any changes in this file or just a move?

tschottdorf:
btw, you could also fix

```
cockroach/pkg/server/serverpb/status.proto: warning: Import cockroach/pkg/storage/log.proto but not used.
cockroach/pkg/server/serverpb/status.proto: warning: Import cockroach/pkg/storage/log.proto but not used.
```

pkg/server/status_logspy.go
tschottdorf: Hard to review because of the change, could you post a diff in the comments?

pkg/server/status_logspy.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/cockroachdb/cockroach/pkg/util/caller"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
+)
+
+const (
+	logSpyDefaultDuration = 10 * time.Second
+	logSpyMaxDuration     = 1 * time.Minute
+	logSpyDefaultCount    = 1000
```
tschottdorf: Actually think that's too large. 1000 events is more than a human can typically consume. Let's default to something smaller.

pkg/server/status_logspy.go
```diff
@@ -0,0 +1,140 @@
...
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, opts.Duration)
+	defer cancel()
+
+	var countDropped int32
+	defer func() {
+		if dropped := atomic.LoadInt32(&countDropped); dropped > 0 {
+			log.Infof(ctx, "%d messages were dropped", dropped)
```
tschottdorf: This should be able to reach the caller.

tbg (commented):
The code changes LGTM, but I'm not sure I really want to look at this:

image

OTOH, this is the more likely format in which we'll want to consume this from "sugared" endpoints in the future.

How do you feel about looking at JSON? Maybe I'm making this worse than it needs to be, but it strikes me as inefficient.

bdarnell:
:lgtm: but I agree that reading raw json is unpleasant. 

---

Reviewed 1 of 2 files at r1, 17 of 17 files at r2.

---

*[pkg/server/server.go, line 1273 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-KuKrICTVEfGyLN3Jd4F:-KuKrICTVEfGyLN3Jd4G:bfsa6f0) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/server.go#L1273)):*
> ```Go
> 
> 	err := w.gz.Close()
> 	gzipWriterPool.Put(w.gz)
> ```

We probably only want to put the gzipWriter back in the pool if `err` is nil. 

---

*[pkg/server/status_logspy.go, line 34 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Ku7JtrXkEta9oUDqT61-r2-34:-KuKsAc1sZDTn74oC6T2:b-co1bta) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/status_logspy.go#L34)):*

Actually think that's too large. 1000 events is more than a human can typically consume. Let's default to something smaller.
</blockquote></details>

It's not unreasonable to load up a big chunk of logs and ⌘F to look around. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-:-KuKtaEugbBJnzmBdFe3:b-kl98nz)*

tschottdorf:

---

*[pkg/server/status_logspy.go, line 34 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Ku7JtrXkEta9oUDqT61-r2-34:-KuKv732aPBOtBTz90vf:b-i698g6) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/status_logspy.go#L34)):*

It's not unreasonable to load up a big chunk of logs and ⌘F to look around. 
</blockquote></details>

And you can, by asking for a big chunk.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521)*

nvanbenschoten:
Improved the pleasantness of the json enough to pass @tschottdorf's standards. We now remove the wrapping "result" json object, which removes most of the noise. If this is a problem in practice, we can reconsider this.

@tschottdorf where are those warnings coming from?

---

---

*[pkg/server/server.go, line 1273 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-KuKrICTVEfGyLN3Jd4F:-Kv4FdanTs1AR8L4-28Y:b-or9p6r) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/server.go#L1273)):*

We probably only want to put the gzipWriter back in the pool if `err` is nil. 
</blockquote></details>

Addressed in #18750.

---

*[pkg/server/status_logspy.go, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Ku7JtrXkEta9oUDqT61-r2-1:-Kv4JT6tRQOOW5w5rFgP:b-aiufi9) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/status_logspy.go#L1)):*

Hard to review because of the change, could you post a diff in the comments?
</blockquote></details>

I performed the file rename in a new commit so that things are easier to follow.

---

*[pkg/server/status_logspy.go, line 34 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Ku7JtrXkEta9oUDqT61-r2-34:-Kv4FifeCMZeneIVm-b3:bdhrqzh) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/status_logspy.go#L34)):*

And you can, by asking for a big chunk.
</blockquote></details>

Lowered to 100.

---

*[pkg/server/status_logspy.go, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Ku7JtrXkEta9oUDqT61-r2-86:-KuuNmrfBNnACbfIIxf3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/status_logspy.go#L86)):*

This should be able to reach the caller.
</blockquote></details>

Done.

---

*[pkg/server/status_logspy.go, line 124 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Ku7JtrXkEta9oUDqT61-r2-124:-KuuO-j_rSifdwjxhBC_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f17c4ee4312b3d2ae5e0a39f2105521b081606f6/pkg/server/status_logspy.go#L124)):*

not your code, but grammar in comment is off. Might as well.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-:-Kv4Fmr98vAmJBjJSSaC:b-cvjcwn)*

tschottdorf:
`make protobuf`, but probably already fixed.
Wasn't able to give this a solid read through, but should be able to do that tomorrow.

---

Reviewed 1 of 2 files at r1, 15 of 17 files at r3, 14 of 18 files at r4, 4 of 4 files at r5.

---

*[pkg/server/status.go, line 652 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Kv4OPdZmM3m_-UVc5np:-Kv4OPdZmM3m_-UVc5nq:b-bv91d1) ([raw file](https://github.com/cockroachdb/cockroach/blob/7798768af36f6d0ac4f3c9c4f09b7e851a1d7866/pkg/server/status.go#L652)):*
> ```Go
> 	ctx := s.AnnotateCtx(stream.Context())
> 	if remote, err := s.maybeRedirect(req.NodeId); err != nil {
> 		return err
> ```

Thought this would fail the linters.

---

*[pkg/server/status.go, line 682 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Kv4OkcZ-IrvxG1MGsZx:-Kv4OkcZ-IrvxG1MGsZy:bsrau68) ([raw file](https://github.com/cockroachdb/cockroach/blob/7798768af36f6d0ac4f3c9c4f09b7e851a1d7866/pkg/server/status.go#L682)):*
> ```Go
> 	go func() {
> 		if dropped := spy.run(ctx, msgC, opts); dropped > 0 {
> 			log.Infof(ctx, "%d messages were dropped", dropped)
> ```

You can return this to the main goroutine and feed it into the stream. I doubt we'll find it in the logs if it ever happens - it should hit our browser too.

---

*[pkg/server/status_logspy_test.go, line 41 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-Kv4PaQaRwFrsiAGF4-a:-Kv4PaQaRwFrsiAGF4-b:bhx3ypk) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9e322e279441ae8f8e4b82fb17231e3bcd4a46d/pkg/server/status_logspy_test.go#L41)):*
> ```Go
> 
> 	testCases := []struct {
> ```

Did all of this testing move somewhere else?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18521#-:-Kv4Puv-dMd0I0V2Wau7:b-50emxc)*

tbg:
>  Wasn't able to give this a solid read through, but should be able to do that tomorrow.

Oops.

I've recently wanted to logspy to a specific node more than once (dozens of times) and always had to do a silly dance of 1. note NodeID of interest 2. translate that to an IP address 3. load the UI there 4. do the logspy thing.

I think it's worth pushing this over the finish line.

PR #17767, opened 2017-08-19
distsqlrun: future-proof the consumer->producer communication

andreimatei:
This patch enhances the consumer->producer (gRPC server->client)
communication by adding a series of unidirectional "handshake" messages.
These are sent by the consumer node as soon as the producer connects to
inform it about the status of the consumer flow: has it already been
scheduled or not. The first message is sent immediately upon producer
connection with the current status. If the consumer flow was not
scheduled already, then a 2nd message is sent later once the consumer
flow is scheduled (if that happens within the scheduling deadline).

The idea here is to give some information to producers about what's
going on with the data they're sending: is it being consumed or not?
This should be helpful if we ever want to devise fallback schemes for
producers: if the consumer is never scheduled where it was supposed to
be scheduled, we might want the producer to send its data somewhere
else. This kind of schemes need a buffering component in the producer
which needs to not drop data it's streaming before knowing that it is in
fact being consumed - so it also needs to know when it's being consumed
so it stops buffering. If such schemes are to be introduced in say 1.2,
then 1.1 likely needs to start sending these messages.

Besides the more speculative uses of these signals, the patch
incidentally also has some more immediate benefits:
- the consumer status messages make it to the producer's trace spans,
which sounds like a good thing when you're trying to figure out the
timings of streaming events.
- the outbox code has been fixed to not consider any signal a drain
signal. This is a biggie because the current code was narrow minded and
didn't let any signals other that drain signals to be sent. This changed
is why this patch requires a DistSQL version bump and this is a good
time to do it since we've already bumped it lots since 1.0.
- related to the point above, other consumer->producer signals can now
be received by the producer after a drain signal, which sounds like a
sane thing.

PR #18522, opened 2017-09-16
distsql: increment DistSQL version

andreimatei:
As of #17767 I intended to update DistSQL Version and
MinAcceptedVersion, but I failed to do so. The code currently is lying
by claiming compatibility with a version it's not, in fact, compatible
with. This patch sets the correct version.

I'll cherry-pick this into 1.1.


RaduBerinde:
LGTM

PR #18524, opened 2017-09-16
cherrypick-1.1: sql: COUNT(*) optimizations

jordanlewis:
sql/sqlbase: don't read idx key if not necessary

During row fetching, if we don't need any of the columns contained within
the index key and the index is not interleaved, we can completely skip
reading and processing the index key.

This gives a nice speed boost for count(*) queries. I observe about 15%
improvement on a million row dataset on a local 1 node cluster.

sql: omit vars in replaced filter expressions

Previously, when a filter expression is completely replaced by an index
constraint, the expression tree was not properly reanalyzed to see which
columns were necessary to decode during row fetching. This led to
unnecessary extra column decodes.

This patch makes sure that the filter expression is properly reanalyzed.
Removing these extra decodes gives a 15-20% performance improvement for
filtered `SELECT COUNT(*)` queries on an index prefix. For example:

``` CREATE TABLE t (a INT, b INT, PRIMARY KEY (a, b));

SELECT COUNT(*) FROM t WHERE a = 0;
```

Performance on the above query on a single node with 1000000 rows in the
table where `a = 0` improved by 18% on my machine.

### DIFF
--- a/pkg/server/settingsworker.go
+++ b/pkg/server/settingsworker.go
@@ -46,7 +46,7 @@ func (s *Server) refreshSettings() {
 		// First we need to decode the setting name field from the index key.
 		{
 			nameRow := []sqlbase.EncDatum{{Type: tbl.Columns[0].Type}}
-			_, matches, err := sqlbase.DecodeIndexKey(a, &tbl, tbl.PrimaryIndex.ID, nameRow, nil, kv.Key)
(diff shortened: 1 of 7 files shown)


jordanlewis:
cc @cockroachdb/release

knz (approved):

PR #18531, opened 2017-09-17
cli/sql: use BSD's libedit for input.

### LINKED ISSUE #14938: cli stores all the duplicate commands
feature request: [referenced here](https://forum.cockroachlabs.com/t/cli-identical-command-question/560)

when using the CLI, i noticed that if i keep doing the same command (via up arrow) over and over again such as:
SELECT * FROM cockroachIsTotallyAsesome WHERE TRUE; that command seems to be saved in the command-line-recall buffer over and over again, even though each one is identical.

linux and mariadb don't remember all the duplicate commands.

the problem of course is that there are important commands such as a CREATE TABLE that get pushed pretty far down the stack.

in other words, if i do the following:

```
CREATE TABLE cockroachIsTotallyAsesome ( field1 blah blah blah);
INSERT INTO cockroachIsTotallyAsesome VALUES (blah blah blah);
SELECT * FROM cockroachIsTotallyAsesome WHERE TRUE;
SELECT * FROM cockroachIsTotallyAsesome WHERE TRUE;
-- repeated 20 times using up-arrow --
SELECT * FROM cockroachIsTotallyAsesome WHERE TRUE;
SELECT * FROM cockroachIsTotallyAsesome WHERE TRUE;
```

i should think my command-line-recall buffer would only contain:

```
SELECT * FROM cockroachIsTotallyAsesome WHERE TRUE;
INSERT INTO cockroachIsTotallyAsesome VALUES (blah blah blah);
CREATE TABLE cockroachIsTotallyAsesome ( field1 blah blah blah);
```

### LINKED ISSUE #15460: cli: cursor on back arrow causes reprint of the last command if it's multiline
inside `cockroach sql`, if the last command in history is multiline and I use back arrow to make a change the console would keep reprinting the command and cursor position is not visible (though it's at the right position).

`root@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'mroot@:26257/> SELECT IF((SELECT recep FROM posts WHERE sender = 'khoze') LIKE 'murino', (SELECT (INSERT INTO posts VALUES (2, 'a', 'b', 0)) LIKE 'INSERT 1'), 0);`

cli

### LINKED ISSUE #16324: cli: command history breaks after 500 line limit is exceeded
After entering 500 or more lines into the CLI, the history file will be broken by the readline rewrite mechanism.

This is a tracking issue for the bugfix in https://github.com/chzyer/readline/pull/121.

### LINKED ISSUE #18192: cli/sql: change the readline dependency
We have a number of hard readline-related issues for which the upstream library has shortcomings:
- cli: cursor on back arrow causes reprint of the last command if it's multiline #15460
- cli: command history breaks after 500 line limit is exceeded #16324
- cli stores all the duplicate commands #14938 

Also as noticed in the acceptance test outputs in `cli/interactive_tests` the upstream library is rather inefficient with refreshing the prompt on input, redrawing the entire line for every character entered (in turn causing input to become very slow as the line becomes longer).

This issue proposes to replace the upstream dependency by one that relies on `libedit`, the readline replacement bundled with BSD systems (including OSX).

A possible Go wrapper for this library is available here: https://github.com/knz/go-libedit

### LINKED ISSUE #18909: cli: \! clear retains typing state on scroll back
I observe the following using `\! clear` and scrolling back up on vanilla tmux with no `.tmux.conf` (doesn't appear to happen without using tmux for me).

```
~ cockroach sql --insecure
# Welcome to the cockroach SQL interface.
# All statements must be terminated by a semicolon.
# To exit: CTRL + D.
#
# Server version: CockroachDB CCL v1.1-alpha.20170817-1056-g57d142d15 (darwin amd64, built 2017/09/24 05:12:12, go1.9) (same version as client)
# Cluster ID: 0fa03c4d-b92a-46d8-978a-d1ab6f736b4e
#
# Enter \? for a brief introduction.
#
root@:26257/> \! clear
root@:26257/>
root@:26257/> S
root@:26257/> SE
root@:26257/> SEL
root@:26257/> SELE
root@:26257/> SELEC
root@:26257/> SELECT
root@:26257/> SELECT
root@:26257/> SELECT 1
root@:26257/> SELECT 1;
root@:26257/> SELECT 1;
root@:26257/> SELECT 1;
+---+
| 1 |
+---+
| 1 |
+---+
(1 row)

Time: 960.588µs

root@:26257/>
```

Build Tag:    v1.1-alpha.20170817-1056-g57d142d15
Build Time:   2017/09/24 05:12:12
Distribution: CCL
Platform:     darwin amd64
Go Version:   go1.9
C Compiler:   4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)
Build SHA-1:  57d142d15ad45359eac7521448fb7cbc9cbc487d
Build Type:   development

knz:
*Very many special thanks to @benesch and @bdarnell without whom this never could have been possible.*

Fixes #18192.
Fixes #16324.
Fixes #15460.
Fixes #14938.
Fixes #18909.

### cli/sql: use BSD's libedit for input.

This patch steps over to use http://github.com/knz/go-libedit,
a Go wrapper around BSD's
[editline](https://www.freebsd.org/cgi/man.cgi?query=editline&sektion=3)
library (`libedit`).

This supports:

- custom key bindings via `~/.editrc`
  See https://www.freebsd.org/cgi/man.cgi?query=editrc
  for details.
- proper entry across multiple lines.
- all terminal types supported by the system's terminfo library.

The name of the history file is changed from `~/.cockroachdb_history`
to `~/.cockroachsql_history` because the format is different. This
both avoids an error upon loading the previous format in the new
shell, and garbage history when trying to run an old shell over the
file using the new format.

### cli/sql: remove the option normalize_history

Prior to this patch, the CLI shell implemented logic to populate the
shell history using pretty-printed renderings of the AST of
parsed statements so far. The reason why this was done was a
combination of two factors:

- the readline input library was not able to handle multi-line
  entries.
- for UX it is desirable to have one "item" of SQL entry, that is,
  a full statement, as a single item of history so as to ease
  recollection of the history entry.

So to satisfy the 2nd point while dealing with the first, the shell
would concatenate the lines of a single statement, parse them,
then pretty-print the AST, because the pretty-print code guarantees
that the result contains a single line (significant newline characters
e.g. in string literals are escaped).

The second requirement still holds, but now the new readline library
*is* able to handle multi-line history entry.

So this patch simplifies the code to add multi-line statements
as one history entry, and removes the complexity associated with
normalization entirely.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,7 @@ expect in your collaboration with the Cockroach Labs team.
      - Optional: NodeJS 6.x and Yarn 0.22.0+. Required when compiling protocol
        buffers.
 
-    Note that at least 4GB of RAM is required to build from source and run tests.
+   Note that at least 4GB of RAM is required to build from source and run tests.
 
 2. Get the CockroachDB code:
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -121,13 +121,6 @@
   version = "2017.04.17"
 
 [[projects]]
-  branch = "master"
-  name = "github.com/chzyer/readline"
-  packages = ["."]
-  revision = "8ca065f6f7703ba76dfd61a8af7c3426b54e650e"
-  source = "https://github.com/cockroachdb/readline"
-
-[[projects]]
   name = "github.com/client9/misspell"
   packages = [".","cmd/misspell"]
   revision = "b5fc83876cd1bf9fbb4e39afa9e503c83edce3b5"
@@ -400,6 +393,12 @@
   name = "github.com/kisielk/gotool"
   packages = ["."]
   revision = "0de1eaf82fa3f583ce21fde859f1e7e0c5e9b220"
+
+[[projects]]
+  name = "github.com/knz/go-libedit"
+  packages = [".","common","other","unix"]
+  revision = "438167ad0c588fe6f2e87cb6ba8a4e9753e6d26e"
+  version = "v1.3"
 
 [[projects]]
   name = "github.com/knz/strtime"
@@ -695,6 +694,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "2ebf48b7717b1b5dfa8b42088bb4c1673d32004c4a69acbfadbb8a493fa9c990"
+  inputs-digest = "44d910075a6ddad082ceae7c050336ca8fbe6b4fabdd108edee8a9f192b08fde"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -45,11 +45,6 @@ ignored = [
   revision = "8dd1f3ff407c300cff0a4bfedd969111ca5a7903"
 
 [[constraint]]
-  name = "github.com/chzyer/readline"
-  source = "https://github.com/cockroachdb/readline"
-  branch = "master"
-
-[[constraint]]
   name = "github.com/coreos/etcd"
   branch = "master"
 
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -72,6 +72,10 @@ func (e *cliError) Error() string { return e.cause.Error() }
 // to be captured.
 var stderr = log.OrigStderr
 
+// stdin aliases os.Stdin; we use an alias here so that tests in this
+// package can redirect the input of the CLI shell.
+var stdin = os.Stdin
+
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "output version information",
--- a/pkg/cli/interactive_tests/common.tcl
+++ b/pkg/cli/interactive_tests/common.tcl
@@ -80,6 +80,15 @@ proc interrupt {} {
     sleep 0.4
 }
 
+# Convenience function that sends Ctrl+D to the monitored process.
+# Leaves some upfront delay to let the readline process the time
+# to initialize the key binding.
+proc send_eof {} {
+    report "EOF TO FOREGROUND PROCESS"
+    sleep 0.4
+    send "\004"
+}
+
 # Convenience functions to start/shutdown the server.
 # Preserves the invariant that the server's PID is saved
 # in `server_pid`.
--- a/pkg/cli/interactive_tests/test_history.tcl
+++ b/pkg/cli/interactive_tests/test_history.tcl
@@ -8,7 +8,7 @@ spawn $argv sql
 eexpect root@
 
 # Ensure the connection is up and working.
-send "select 1;\r"
+send "SELECT 1;\r"
 eexpect "1 row"
 eexpect root@
 
@@ -27,7 +27,7 @@ start_test "Test that we can recall a previous line with Ctrl+R"
 send "foo;\r"
 eexpect "syntax error"
 eexpect root@
-send "\022sel"
+send "\022SEL"
 eexpect "SELECT 1;"
 end_test
 
@@ -43,14 +43,14 @@ eexpect "SELECT 1;"
 end_test
 
 start_test "Test that client cannot terminate with Ctrl+D while cursor is on recalled line"
-send "\004"
+send_eof
 send "\r"
 eexpect "1 row"
 eexpect root@
 end_test
 
 start_test "Test that Ctrl+D does terminate client on empty line"
-send "\004"
+send_eof
 eexpect eof
 end_test
 
@@ -63,7 +63,7 @@ end_test
 
 start_test "Test that the client cannot terminate with Ctrl+C while cursor is on recalled line"
 interrupt
-send "\rselect 1;\r"
+send "\rSELECT 1;\r"
 eexpect "1 row"
 eexpect root@
 end_test
@@ -80,7 +80,7 @@ eexpect root@
 end_test
 
 start_test "Test that two statements on the same line can be recalled together."
-send "select 2; select 3;\r"
+send "SELECT 2; SELECT 3;\r"
 eexpect "1 row"
 eexpect "1 row"
 eexpect root@
--- a/pkg/cli/interactive_tests/test_history.tcl
+++ b/pkg/cli/interactive_tests/test_history.tcl
@@ -8,13 +8,13 @@ spawn $argv sql
 eexpect root@
 
 # Ensure the connection is up and working.
-send "SELECT 1;\r"
+send "select 1;\r"
 eexpect "1 row"
 eexpect root@
 
 start_test "Test that last line can be recalled with arrow-up"
 send "\033\[A"
-eexpect "SELECT 1;"
+eexpect "select 1;"
 end_test
 
 start_test "Test that recalled last line can be executed"
@@ -27,8 +27,8 @@ start_test "Test that we can recall a previous line with Ctrl+R"
 send "foo;\r"
 eexpect "syntax error"
 eexpect root@
-send "\022SEL"
-eexpect "SELECT 1;"
+send "\022sel"
+eexpect "select 1;"
 end_test
 
 start_test "Test that recalled previous line can be executed"
@@ -39,7 +39,7 @@ end_test
 
 start_test "Test that last recalled line becomes top of history"
 send "\033\[A"
-eexpect "SELECT 1;"
+eexpect "select 1;"
 end_test
 
 start_test "Test that client cannot terminate with Ctrl+D while cursor is on recalled line"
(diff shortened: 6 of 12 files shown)


knz:
@benesch I'll need your help to 

1) explain the builder how to generate linux binaries against a non-versioned libtinfo
2) vendor the dependency

knz:
I've done the vendoring but somehow teamcity does not agree with me. Not sure why.

benesch:
This seems to be the error:

```
[22:13:48][Step 2/3] In file included from vendor/github.com/knz/go-libedit/src/libedit-terminal.c:1:0,
[22:13:48][Step 2/3]                  from wrap-terminal.c:1:
[22:13:48][Step 2/3] vendor/github.com/knz/go-libedit/src/c-libedit/terminal.c:58:21: fatal error: termcap.h: No such file or directory
[22:13:48][Step 2/3]  #include <termcap.h>
[22:13:48][Step 2/3]                      ^
[22:13:48][Step 2/3] compilation terminated.
```

knz:
Okay I fixed the last issues with the Go code.
Blockers:
- #18687 - `dep` is making me unhappy for seemingly unrelated issues
- the builder image needs to be populated with the terminfo header files.

benesch:
Ah, so that indicates we've introduced a build-time dependency on `libterminfo-dev`. That'll need to be documented in the build from source instructions.

knz:
- the dep issue is solved
- the builder image has all the good files, no need to add more
- the remaining issue is that the acceptance tests do not use the builder image to build cockroachdb, and instead use our "acceptance test image". That image is missing some files (looking into it). I am somewhat confused as to why the acceptance tests do not use the builder to build though.

knz:
Okay so the builder has the terminfo lib and headers in /usr/include and /usr/lib, but the `make build` target uses the sysroot in /x-tools exclusively.

We need two things to happen, for the linux targets:
- symlink /usr/include/termcap.h into the corresponding /x-tools sysroots
- import the non-versioned libtinfo.so (e.g. from CentOS) and make the linker use that.

@benesch I'd like to sit down a few minutes with you to get this sorted.

knz:
Ok progress stalled once again due to `dep` breakage. Awaiting for update by @dt to rebase/regenerate.

knz:
Will rely on  #19007.

knz:
Builder updated; confirmed to generate working binaries with TYPE= `release-linux-gnu`, `release-linux-musl`, `portable`, `release-windows` and `release-darwin`.

Now waiting on `honnef.co` to resuscitate to update the vendor tree again.

bdarnell:
So the net result is that on mac/bsd, this links dynamically against libedit, on linux-glibc it embeds libedit and links dynamically against libtinfo, and on linux-musl it links everything statically, right? Does the linux-musl build need a terminfo database on disk or does it  have something embedded (for at least the most popular terminals)?

---

Reviewed 8 of 8 files at r1.

---

*[CONTRIBUTING.md, line 41 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-KvdKZ0ZMeDNxf1ZMR4l:-KvdKZ0ZMeDNxf1ZMR4m:bde5z6j) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a2b1fc8e9adfe22ee492cf7420ab5985d80cbf9/CONTRIBUTING.md#L41)):*
> ```Markdown
>        buffers.
>      - On GNU/Linux specifically, the terminfo/ncurses development
>        headers (note: not needed on BSD, Windows and macOS). This is
> ```

On bsd/mac, they're still needed, but they're present by default, right?

---

*[pkg/cli/interactive_tests/test_history.tcl, line 11 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-KvdL7IsQDPMCBp2z2vZ:-KvdL7IsQDPMCBp2z2v_:b-puqa01) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a2b1fc8e9adfe22ee492cf7420ab5985d80cbf9/pkg/cli/interactive_tests/test_history.tcl#L11)):*
> ```Tcl
> 
> # Ensure the connection is up and working.
> send "SELECT 1;\r"
> ```

Why are these now capitalized? Users will usually type in lowercase. If the history is now case-sensitive then the fact that we normalize to uppercase will be really annoying. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-:-KvdMe7N2PvspW0tsWwJ:b-f6acyn)*

knz:
your analysis is correct. Additionally on windows and other unknown systems it falls back to a dumb stdin readline.

The static build needs a terminfo database on disk, however I checked and Alpine does indeed have one as part of the base system. Also the user can populate entries in their home directory in ~/.terminfo. There's an option to embed some entries but I haven't felt the need to do so yet.

---

---

*[CONTRIBUTING.md, line 41 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-KvdKZ0ZMeDNxf1ZMR4l:-KvdPmyJmsKAYHjb1PJ2:b-r960f8) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a2b1fc8e9adfe22ee492cf7420ab5985d80cbf9/CONTRIBUTING.md#L41)):*

On bsd/mac, they're still needed, but they're present by default, right?
</blockquote></details>

This comment is stale. I changed the code to not need any dev headers at all. Will update.

---

*[pkg/cli/interactive_tests/test_history.tcl, line 11 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-KvdL7IsQDPMCBp2z2vZ:-KvdPqYMtylkXwziL_3R:b-tuq6z6) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a2b1fc8e9adfe22ee492cf7420ab5985d80cbf9/pkg/cli/interactive_tests/test_history.tcl#L11)):*

Why are these now capitalized? Users will usually type in lowercase. If the history is now case-sensitive then the fact that we normalize to uppercase will be really annoying. 
</blockquote></details>

Yes I just found this out. Will send out a separate PR in front of this one to disable normalization in the history. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-:-KvdPPBuluPlxzkVvsBN:blr3v8l)*

knz:
ready, PTAL

benesch:
This works impressively well! One small nit. On a machine that has a go-readline history file, this warning occurs:

```
cannot load the command-line history (file corrupted?): read_history: invalid argument
the history file will be cleared upon first entry
```

Perhaps we should use a new name for the history file? It's especially bad in the other direction, since go-readline mis-decodes libedit's history:

```
insert\040into\040db.t\040values\040(\012\040\040'a'\012);\012
```

On second thought, it might be pretty easy to migrate the history file. (Third thought: holy cow the escaping routines in libedit are _complicated_.)

---

Reviewed 6 of 8 files at r1, 2 of 2 files at r2.
Review status: 6 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-:-KvhSUHi59RUAUQvqgTw:bc1hwof)*

knz:
- Restored CONTRIBUTING.md (ncurses headers are not necessary).
- Changed the history file name to become `~/.cockroachsql_history` (instead of `~/.cockroachdb_history` as previously, to avoid running into the error message)

benesch:
:lgtm_strong: 

---

Reviewed 4 of 4 files at r3.
Review status: 6 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18531#-:-KvhclPjhCvWdY0uLGpl:b-uwkowd)*

knz:
Yay! it's green! Shall we press the big green button?

knz:
Updated to take the fix to https://github.com/knz/go-libedit/issues/1.

jseldess:
@knz, will this be cherrypicked into 1.1? If so, I'll remove the associated known limitation from the 1.1 docs.

bdarnell:
No, this is far too big a change to cherry-pick.

jseldess:
OK, thanks, @bdarnell. I'll wait until it shows up in a patch release (or 1.2) before removing the limitations then.

knz:
Agreed this will need a few months of experimental testing by the team before we are happy about its stability.

PR #18532, opened 2017-09-17
cherrypick: pgwire, cli/sql: support tables with no columns properly #17816

knz:
Note for docs:

this also changes the CSV and TSV formats to print the row count at the end (instead of at the beginning). This is a user-facing change that may impact users who use an automated tool to parse/analyze the output produced by `cockroach sql`.


knz:
cc @cockroachdb/release

jordanlewis (approved):

bdarnell:
> this also changes the CSV and TSV formats to print the row count at the end (instead of at the beginning). This is a user-facing change that may impact users who use an automated tool to parse/analyze the output produced by cockroach sql.

Why do CSV/TSV formats print the row count at all? That seems to just make it needlessly difficult to get this data into other programs.

knz:
When it was at the beginning the rationale was to allow for readers to know in advance how many rows to expect. At the end, like for every other format, the rationale is for the CLI shell to offer a reminder to the human reader of how many rows were printed.

bdarnell:
If it's for humans it should be conditioned on interactive mode. I should be able to pipe `cockroach sql --format=csv` into a program that expects CSV. (It looks like we also print the server version and cluster id unconditionally; these should also be interactive-only).

knz:
Ack. Will look into it.

PR #18546, opened 2017-09-18
cherrypick: distsql: increment DistSQL version

andreimatei:
Cherrypick #18522

As of #17767 I intended to update DistSQL Version and
MinAcceptedVersion, but I failed to do so. The code currently is lying
by claiming compatibility with a version it's not, in fact, compatible
with. This patch sets the correct version.

cc @cockroachdb/release

### DIFF
--- a/pkg/sql/distsqlrun/server.go
+++ b/pkg/sql/distsqlrun/server.go
@@ -71,11 +71,11 @@ type DistSQLVersion uint32
 //
 // ATTENTION: When updating these fields, add to version_history.txt explaining
 // what changed.
-const Version DistSQLVersion = 5
+const Version DistSQLVersion = 6
 
 // MinAcceptedVersion is the oldest version that the server is
 // compatible with; see above.
-const MinAcceptedVersion DistSQLVersion = 4
+const MinAcceptedVersion DistSQLVersion = 6
 
 var settingUseTempStorageSorts = settings.RegisterBoolSetting(
 	"sql.distsql.temp_storage.sorts",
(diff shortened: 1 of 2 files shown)


asubiotto:
:lgtm: with little nits.

---

---

*[pkg/sql/distsqlrun/version_history.txt, line 6 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18546#-KuKutLSr5MwdA55yPXk:-KuKutLSr5MwdA55yPXl:b-q087e3) ([raw file](https://github.com/cockroachdb/cockroach/blob/827eb06f8a574c719d6156bd620ed3f1cc102651/pkg/sql/distsqlrun/version_history.txt#L6)):*
> ```
>     sent by the server-side of FlowStream informing the producer of the
>     consumer's status. The messages would cause a server running the previous
>     version to erroneously think that the consumer has sent a drain signal,
> ```

s/has/had

---

*[pkg/sql/distsqlrun/version_history.txt, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18546#-KuKsF9QPGOyFlvzxGAD:-KuKsF9QPGOyFlvzxGAE:b8x07iv) ([raw file](https://github.com/cockroachdb/cockroach/blob/827eb06f8a574c719d6156bd620ed3f1cc102651/pkg/sql/distsqlrun/version_history.txt#L7)):*
> ```
>     consumer's status. The messages would cause a server running the previous
>     version to erroneously think that the consumer has sent a drain signal,
>     therefor MinAcceptedVersion was increased.
> ```

s/therefor/therefore

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18546#-:-KuKub0BAcCcj_XBWx1Z:b-lwzsb)*

andreimatei:
Too late for these nits, unfortunately. I'll fix them on the master branch though.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18546#-:-KuKx-EfwqKEFTy7trjH:bicgfgj)*

PR #18547, opened 2017-09-18
sqlccl: implement RESTORE ... AS OF SYSTEM TIME

danhhz:
This requires two pieces: an option to Export that saves every mvcc
value instead of just the latest and an option to Import that selects
the latest version of each key with an mvcc timestamp not greater than
some EndTime.

BACKUP is given an 'mvcc_all' option that triggers the export of all
mvcc versions of each key, instead of just the latest. The default is
left as latest since this is expected to be the common case.

RESTORE ... AS OF SYSTEM TIME then can restore to times in the middle of
BACKUPs made with the mvcc_all option. Not every BACKUP in the RESTORE
needs the option, only the one covering the relevant timestamp (e.g. a
full backup over (t0..t2] without mvcc_all and an incremental backup
over (t2, t4] with mvcc_all can restore to any time in [t2, t4].

If a node with this change sends an Import or Export request to an old
node, it will ignore the new field and do the old behavior, so using the
mvcc_all option to BACKUP or RESTORE ... AS OF SYSTEM TIME will require
all nodes to be upgraded.

A check to skip export of keys that are deleted before the selected
timestamp in a full backup is moved from MVCCIncrementalIterator to
Export where it fits more naturally.

Reviewers: I thought about the naming of the proto fields and enums but didn't come up with anything I loved. Definitely accepting suggestions here. Also note that we should gate these features on all nodes being upgraded. Pretty sure we can do that now, I'll look into how.


danhhz:
Note to self: just realized we need to note the gc threshold for each file in a full backup. This can then be used to disallow RESTORE ... AS OF SYSTEM TIME for long ago times.

I'll wait for initial comments on the approach before doing any followups though.

---

Review status: 0 of 18 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18547#-:-KuKrXPKxtS5DudQm1q2:brrift0)*

bdarnell:
The approach LGTM. A couple of syntactic bikeshed nits:
- What about `WITH multiversion` or `WITH mvcc` instead of `WITH mvcc_all`? Or `WITH mvcc=all` to allow for potential variations like `WITH mvcc=6h`
- I'm not sure about `AS OF SYSTEM TIME` on commands that are not read only. It feels weird to use that syntax for something other than setting the timestamp of the transaction, but maybe that's just me. 

---

Review status: 0 of 18 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18547#-:-KuLEbpNRA4nz79pEXmm:b-1r9l9m)*

dt:
:lgtm_strong: 

just to hop on option name bike shed: `WITH  revisions` or `WITH revision_history`might be more intuitive to someone who isn't familiar with the more niche terminology of `mvcc` ? 

and, just thinking aloud: if we have backups that contain all revisions, do we also want RESTORE to be able to that data restore all revisions, rather than just latest or up to some time, i.e. such that time travel on the restored table works?  I guess we'd have to do something I guess to protect them from immediate GC though. Honestly seems like read-only queries over backups that support the as-of time probably is more useful anyway.

---

Review status: 0 of 18 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/ccl/storageccl/engineccl/sst_iterator.go, line 141 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18547#-KuLK1SYMHyp_cYBzpyt:-KuLK1SYMHyp_cYBzpyu:bb7hqu0) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8a2ba3c3a906acdb91323317d19a4c10af5e8e4/pkg/ccl/storageccl/engineccl/sst_iterator.go#L141)):*
> ```Go
> 	}
> 	r.nextKeyStart = append(r.nextKeyStart[:0], r.mvccKey.Key...)
> 	for r.Next(); r.valid && r.err == nil && bytes.Equal(r.nextKeyStart, r.mvccKey.Key); r.Next() {
> ```

huh.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18547#-:-KuLHeZ46l8iobIlAs6h:b-6nod2j)*

bdarnell:
> do we also want RESTORE to be able to that data restore all revisions, rather than just latest or up to some time, i.e. such that time travel on the restored table works?

Yes, I think we probably want this at least as an option. People who run with an increased GC TTL would want to be able to preserve this data across restores (but since I'm not sure if anyone is doing that now it's not a major concern)

dianasaur323:
I noticed that you marked that the default behavior is still to store only the latest. based on an offline conversation with @spencerkimball, he believes that we should make the full MVCC backup the default. I'll leave it to you all to bikeshed, but happy to follow up with some customers if we can't come to an agreement.

dt:
my 2¢: current behavior (backups as a snapshot, not full revision history) makes sense as the default, but maybe the docs on time travel / gc threshold / how to configure it should include a link to backup docs and mention the fat backup option.

spencerkimball:
+1 on `WITH revision_history`.

The reason I'm arguing to change the default to store all MVCC values is that it's likely to be inexpensive and safer for the end user.

Here's an imaginary use case. A user sets the TTL on a table to 7 years. They do a full backup, their cluster goes AWOL, they try to do a restore and their history is missing. If a user changes the TTL to 7 years, it's a pretty good indication that they want all revisions backed up.

Instead of trying to guess whether they likely want the revisions, let's just keep them by default. In the user story above, that user would be piping mad.

dt:
In our prioritization yesterday, we said we wanted to get this feature in soon so we could start banging on it in practice and prod testing, so I went ahead and rebased it to get things going again. 

anyone want to throw opinions into the option name bike shed?

danhhz:
Haha, message received. I'll pick this back up again

We have 2 votes for `WITH revision_history` so I'll go with that unless someone speaks up. It's better to get testing going now and we have 3.5 months of alphas before we have to commit to it

I think we should file an issue to investigate switching the default and address it in a followup if necessary. @spencerkimball makes a very compelling point, but I think we should quantify how much extra space we think that'll take in the common case (in particular, there's all sorts of compressions involved so we should look at actual backup files sizes)

dt:
Heh, whoops, no message intended @danhhz, I just volunteered yesterday to pick up your branch since you had other things on your plate.

danhhz:
@dt no worries. It seems like I should have the bandwidth to land this, tho I'll happily punt the "note to self" above for you to do as a followup if you like

Anyway, I updated the option name to `WITH revision_history`

dt:
Also filed #19342 to capture considering adding full mvcc restore support, but AFAIK, that isn't on our roadmap as of yet.

danhhz:
I'm going to go ahead and merge this so @dianasaur323 can start testing it. I'll file issues for the concerns @bdarnell (is AS OF SYSTEM TIME appropriate for write queries) and @spencerkimball (should WITH revision_history be the default) raised, as well as my "note to self" to make sure they don't get dropped.

dt:
@danhhz the second of those is filed as https://github.com/cockroachdb/cockroach/issues/19341

danhhz:
Awesome. The other two are https://github.com/cockroachdb/cockroach/issues/19454 and https://github.com/cockroachdb/cockroach/issues/19455

dianasaur323:
Excellent! I'll start running this.. after I figure out what is wrong with our PM cluster =.=

PR #18549, opened 2017-09-18
cherrypick-1.1: sql: fix TRUNCATE on a table in the middle of a schema change

vivekmenezes:
TRUNCATE drops the existing table and creates a new one with the
same schema. In doing so it currently copies over all the schema
change mutations and jobs associated with the schema change thus
associating the same job with two tables. This is broken!

This change associates all the mutations and jobs with the table being
dropped, and converts all outstanding schema changes on the new table
to the public state. The new table is empty and thus doesn't need
a backfill job.

### DIFF
--- a/pkg/sql/schema_changer_test.go
+++ b/pkg/sql/schema_changer_test.go
@@ -2545,3 +2545,102 @@ CREATE TABLE t.test (k INT PRIMARY KEY, v INT, pi DECIMAL REFERENCES t.pi (d) DE
 		t.Fatalf("expected %d key value pairs, but got %d", e, len(kvs))
 	}
 }
+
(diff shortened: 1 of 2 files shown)


jordanlewis:
The title should be cherrypick-1.1

vivekmenezes:
ready to be merged

jordanlewis (approved):

jseldess:
@vivekmenezes, @jordanlewis, one of our 1.0 and 1.1 known limitations says:

> DDL statements cannot follow DML statements. As a workaround, arrange DML statements before DDL statements, or split the statements into separate transactions.

Have we resolved this limitation via this PR?

PR #18556, opened 2017-09-18
build,sql,cli: reveal the CockroachDB version number

knz:
Previous to this patch, the vtable `crdb_internal.node_build_info`,
the built-in SQL function `version()` and the command-line utilities
would only print the build tag as a version number.

This patch augments this display by prefixing it by the
`VersionPrefix()`, computed after the current branch.

Before:

```
CockroachDB CCL 9fcb2f35e (freebsd amd64, built 2017/09/18 18:27:19, go1.9)
```

After:

```
CockroachDB CCL v1.1-9fcb2f35e (freebsd amd64, built 2017/09/18 18:27:19, go1.9)
```

cc @jordanlewis @justinj

### DIFF
--- a/pkg/build/info.go
+++ b/pkg/build/info.go
@@ -76,9 +76,15 @@ func init() {
 	}
 }
 
+// VersionedTag prefixes the tag with VersionPrefix.
+func (b Info) VersionedTag() string {
+	return VersionPrefix() + "-" + b.Tag
+}
+
 // Short returns a pretty printed build and version summary.
 func (b Info) Short() string {
-	return fmt.Sprintf("CockroachDB %s %s (%s, built %s, %s)", b.Distribution, b.Tag, b.Platform, b.Time, b.GoVersion)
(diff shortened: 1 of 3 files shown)


benesch (approved):
LGTM, but is `version()` subject to backwards-compatibility requirements?

knz:
The overall format of version() doesn't change. From the client perspective where the VersionPrefix now appears would have been an opaque tag of variable length previously. clients that assume an opaque tag will continue to work.

benesch:
But clients that assumed that token was a SHA would break with this change,
right? I suspect this doesn't much matter, but wanted to flag it.

dt:
clients already couldn't assume it was a SHA, but they generally *could* assume it was a meaningful scm ref, which they can't with this... but I don't think that's a big deal. this output is more useful to humans anyway, and we can fix the machines if they care.

knz:
Thank you!

PR #18558, opened 2017-09-18
sql: fix SHOW [KV] TRACE on IPv6 systems

knz:
The way SHOW [KV] TRACE works is that it runs the query with tracing
enabled then collects the tracing messages and *splits* them across
the two columns "message" and "context".

The context column is meant to contain the logging tag, the part
between square brackets at the beginning of a log/trace message.
Unfortunately the entire logging/tracing infrastructure only collect a
single string that combines the tag (context) and the message (as
opposed to a structured format that would collect them separately), so
SHOW TRACE has to split them after the fact. Moreover, the tag can
contain spaces, so that is not a good delimiter, can be of variable
length, and can contain most other punctuation in variable amount, so
punctuation cannot be used as a delimiter either. Therefore, when
implementing SHOW TRACE in the last iteration I made it use a regular
expression to extract "everything enclosed in square brackets" at the
beginning of a message as a context.
That is, using the regular expression `'^\[[^]*\] '`.

However, this regular expression is incorrect *if the tag itself
contains data enclosed in square brackets*. This happens to be a
rather common case on macOS, and indeed any system where IPv6 is
enabled, because a CockroachDB node when started with `--insecure` by
default will bind to the local IPv6 interface, ::1, which happens to
render textually in Go as "`[::1]`".

This in turn caused:

1. the context to be incorrectly extracted/stripped from the trace
   message, causing a messed up context column in the output of SHOW
   TRACE.
2. SHOW KV TRACE to fail to match entirely, i.e. empty KV traces.

This patch improves the behavior by supporting *a single level of extra
bracketing in the log tag*.

This patch is not a panacea, given that any additional (ab)use of
brackets in the future may break this feature again, and we don't have
a clear way to protect the codebase against that yet. However, the
stronger solution in the shape of introducing a little bit more
structure to the whole tracing/logging infrastructure is not exactly
trivial either. This is a 80% solution.

### DIFF
--- a/pkg/sql/show.go
+++ b/pkg/sql/show.go
@@ -284,8 +284,8 @@ SELECT timestamp,
        operation,
        span
   FROM (SELECT timestamp,
-               regexp_replace(message, e'^.*\\[[^]]*\\] ', '') AS message,
-               regexp_extract(message, e'^.*\\[[^]]*\\]') AS context,
+               regexp_replace(message, e'^\\[(?:[^][]|\\[[^]]*\\])*\\] ', '') AS message,
+               regexp_extract(message, e'^\\[(?:[^][]|\\[[^]]*\\])*\\]') AS context,
                first_value(operation) OVER (PARTITION BY txn_idx, span_idx ORDER BY message_idx) as operation,
                (txn_idx, span_idx) AS span
           FROM crdb_internal.session_trace)
--- a/pkg/sql/trace_test.go
+++ b/pkg/sql/trace_test.go
@@ -19,12 +19,14 @@ import (
 	"fmt"
 	"net/url"
 	"sort"
+	"strings"
 	"testing"
 
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/security"
+	"github.com/cockroachdb/cockroach/pkg/sql"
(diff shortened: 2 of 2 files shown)


knz:
@bdarnell @benesch do you have any suggestion as to how to test that reliably, given that it would require a test to enforce the server to bind to an ipv6 address?

jordanlewis (approved):
Confirmed that this solves my problem. I'm okay with leaving this as untested as it is.

jordanlewis (changes requested):
Actually wait this doesn't solve the issue.

My context columns now all equal `1`.

bdarnell:
:lgtm: 

Can you get a tag containing brackets into the log context some other way? Maybe a user name containing brackets, or a test hook to modify the context arbitrarily? 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18558#-:-KuLP07PgJ-qz6qIXUF9:bexcutc)*

knz:
Fixed the remaining bug and added a test for it. PTAL.

bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18558#-:-KuLjfnuHb55bWoghg1P:bnfp4nl)*

jordanlewis (approved):

knz:
TFYR!

PR #18562, opened 2017-09-18
cherrypick: build,sql,cli: reveal the CockroachDB version number

knz:
Previous to this patch, the vtable `crdb_internal.node_build_info`,
the built-in SQL function `version()` and the command-line utilities
would only print the build tag as a version number.

This patch augments this display by prefixing it by the
`VersionPrefix()`, computed after the current branch.

- for the branch `release-1.0`, where the last commit in the branch is
  exactly a tag, the tag string is exactly `v1.0.6`.
- for the branch `release-1.1`, where there were some commits since the
  last tag, the tag string is something like
  `v1.1-beta.20170907-95-gdcc6ca4c3`.
- for the master branch, once we have at least one 1.2 alpha release,
  the tag string will be something like
  `v1.2-alpha.2017MMDD-NN-SHAAAAAAA`.

cc @cockroachdb/release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -361,8 +361,7 @@ $(ARCHIVE).tmp: .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
 # For details, see the "Possible timestamp problems with diff-files?" thread on
 # the Git mailing list (http://marc.info/?l=git&m=131687596307197).
 .buildinfo/tag: | .buildinfo
-	@{ git describe --tags --exact-match 2> /dev/null || git rev-parse --short HEAD; } | tr -d \\n > $@
-	@git diff --quiet HEAD || echo -dirty >> $@
+	@{ git describe --tags --dirty 2> /dev/null || git rev-parse --short HEAD; } | tr -d \\n > $@
 
 .buildinfo/basebranch: | .buildinfo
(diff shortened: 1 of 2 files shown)


bdarnell:

---

*[pkg/build/info.go, line 81 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18562#-KuLYzh8BKOENYXkVluy:-KuLYzh8BKOENYXkVluz:b-26g3dv) ([raw file](https://github.com/cockroachdb/cockroach/blob/b7e450ca5967e59be594fc0d73705623fe421a22/pkg/build/info.go#L81)):*
> ```Go
> // VersionedTag prefixes the tag with VersionPrefix.
> func (b Info) VersionedTag() string {
> 	return VersionPrefix() + "-" + b.Tag
> ```

I don't think this is going to the right thing for release builds. In those builds, `b.Tag` is a tag and looks like `v1.0.6`, not a commit hash. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18562)*

knz:
I'll rebase with a different commit.

knz:
Using the new fix, PTAL.

bdarnell (approved):

PR #18503, opened 2017-09-15
sqlccl: add tests for CSV create table errors

madelynnblue:
makeCSVTableDescriptor now checks the create statement for more
failures. This is needed because otherwise sql.MakeTableDesc can panic
due to the nil EvalContext (like when processing DEFAULT expressions).

PR #18435, opened 2017-09-11
sqlccl: populate CSV backup descriptor metadata

madelynnblue:

PR #18433, opened 2017-09-11
sqlccl: don't split CSV files mid column family

madelynnblue:
Previously, we would allow splitting between two KVs in the index of
different column families. This caused numerous odd errors, including
double-counting these rows during RESTORE. The select inconsistencies
mentioned in the fixed issues are also explained by this, I think,
because the system isn't quite sure what to do with these partial
rows and so does some weird stuff.

Fixes #18076

PR #18563, opened 2017-09-18
cherrypick-1.1: sqlccl: various CSV fixes

madelynnblue:
See: #18437, #18433, #18435, #18503, #18438

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -161,10 +161,21 @@ func doLocalCSVTransform(
 	// of pre-computed data improves overall performance.
 	const chanSize = 10000
 
-	group, gCtx := errgroup.WithContext(ctx)
 	recordCh := make(chan csvRecord, chanSize)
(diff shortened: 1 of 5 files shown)


danhhz:
:lgtm:

---

---

*[pkg/ccl/sqlccl/csv.go, line 269 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18563#-KuPl9VxUhNaExmKIFf5:-KuPke3vqshlhmWYs5Zm:bsjzkg8) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fc8f6db025ed3247f1bf961a91b6fb52462171/pkg/ccl/sqlccl/csv.go#L269)):*
> ```Go
> 	}
> 	for _, def := range create.Defs {
> 		switch def := def.(type) {
> ```

too late for this since it's a cherry-pick PR but in the future, I think it's clearer if this isn't shadowed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18563#-:-KuPl8bteVp0qVGUk8f7:bnfp4nl)*

PR #18564, opened 2017-09-18
sql: fix problem with autoCommit leaving txn in wrong state

andreimatei:
When executing an implicit transaction, a certain layer in the Executor
code (runTxnAttempt()) is supposed to always leave the session in the
NoTxn state. In other words, it's supposed to always clean up the
transaction.
Before this patch, it was failing to do so in a particular case: if it
looked like we were going to auto-retry but at the last moment we
decided that we can't do that because we've already streamed results to
the client, we were leaving the session (and the transaction) in the
NoTxn state.
This patch fixes this by adding the missing cleanup.

I will cherry-pick this into 1.1

### DIFF
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -950,7 +950,12 @@ func (e *Executor) execParsed(
 // returned as remainingStmts.
 // txnPrefix: Set if stmtsToExec corresponds to the start of the current
 // transaction. Used to trap nested BEGINs.
-// autoCommit: Set if the transction should be committed at the end of the
+// autoCommit: If set, the transaction will be committed after running the
+//   statement. If set, stmtsToExec can only contain a single statement.
+//   If set, the transaction state will always be NoTxn when this function
+//   returns, regardless of errors.
+//   Errors encountered when committing are reported to the caller and are
+//   indistinguishable from errors encountered while running the query.
 // protoTS: If not nil, the transaction proto sets its Orig and Max timestamps
 // to it each retry.
 //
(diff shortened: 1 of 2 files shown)


pkg/sql/executor.go
```diff
@@ -1104,6 +1109,10 @@ func runWithAutoRetry(
 				e.TxnAbortCount.Inc(1)
 				txnState.mu.txn.CleanupOnError(session.Ctx(), err)
 				txnState.resetStateAndTxn(Aborted)
+				// autoCommit always leaves the transaction in NoTxn.
+				if autoCommit {
+					txnState.resetStateAndTxn(NoTxn)
```
knz: Is it useful/desirable to call resetStateAndTxn two times in a row?  Can't you combine them and use a if/else here somehow?

andreimatei:

---

*[pkg/sql/executor.go, line 1114 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18564#-KuLdzM2ky9gD3DP7cro-r1-1114:-KuLrwUdBi40jsOhkzt0:bbqqeks) ([raw file](https://github.com/cockroachdb/cockroach/blob/c14477e7b3f8f8ff1d5b0139d238c74b6cf16f62/pkg/sql/executor.go#L1114)):*

Is it useful/desirable to call resetStateAndTxn two times in a row?  Can't you combine them and use a if/else here somehow?
</blockquote></details>

done. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18564)*

knz:
:lgtm_strong: 

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18564#-:-KuLsPJeH9gSUkfWugHe:b-uwkowd)*

PR #18565, opened 2017-09-18
build,sql,cli: reveal the CockroachDB version number (again)

knz:
Reverts/Supersedes #18556.

The previous patch was blindly concatenating the result of
`VersionPrefix()` with the build tag, to yield something like
`v1.1-abcdef`. However, as highlighted by @bdarnell, in the specific
case where the current tree is exactly a tagged version, then the
build tag becomes the (versioned) git tag, and the resulting visible
version number becomes `v1.1-v1.1`, which ... makes no sense.

This patch resolves this by removing the custom logic introduced by
the previous patch entirely, and using the output of `git describe`
instead of `git describe --exact-match` as buildinfo tag.

This gives us:

- for the branch `release-1.0`, where the last commit in the branch is
  exactly a tag, the tag string is exactly `v1.0.6`.
- for the branch `release-1.1`, where there were some commits since the
  last tag, the tag string is something like
  `v1.1-beta.20170907-95-gdcc6ca4c3`.
- for the master branch, once we have at least one 1.2 alpha release,
  the tag string will be something like
  `v1.2-alpha.2017MMDD-NN-SHAAAAAAA`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -361,8 +361,7 @@ $(ARCHIVE).tmp: .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
 # For details, see the "Possible timestamp problems with diff-files?" thread on
 # the Git mailing list (http://marc.info/?l=git&m=131687596307197).
 .buildinfo/tag: | .buildinfo
-	@{ git describe --tags --exact-match 2> /dev/null || git rev-parse --short HEAD; } | tr -d \\n > $@
-	@git diff --quiet HEAD || echo -dirty >> $@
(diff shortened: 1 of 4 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18565#-:-KuLqtwFW90ahHz2mVlN:bnfp4nl)*

knz:
TFYR!

PR #18571, opened 2017-09-18
sql: restore the logical order of UNION

knz:
In order to compute EXCEPT properly, the code for
UNION/INTERSECT/EXCEPT needs to read its right operand first to
populate the map of known values. However in the non-EXCEPT case the
resulting output is non-intuitive, as a savvy user with a reasonable
order expectation would see the right operand being consumed first,
even when not necessary.

This patch restores the "natural" order.

This is not a user-facing change worth noting in docs: the ordering of
UNION's result has never been guaranteed without an ORDER BY clause.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/explain_types
+++ b/pkg/sql/logictest/testdata/logic_test/explain_types
@@ -181,10 +181,10 @@ EXPLAIN (TYPES) VALUES (1) UNION VALUES (2)
 0  union   ·              ·                (column1 int)  ·
 1  values  ·              ·                (column1 int)  ·
 1  ·       size           1 column, 1 row  ·              ·
-1  ·       row 0, expr 0  (1)[int]         ·              ·
+1  ·       row 0, expr 0  (2)[int]         ·              ·
 1  values  ·              ·                (column1 int)  ·
 1  ·       size           1 column, 1 row  ·              ·
-1  ·       row 0, expr 0  (2)[int]         ·              ·
+1  ·       row 0, expr 0  (1)[int]         ·              ·
 
 query ITTTTT
 EXPLAIN (TYPES) SELECT DISTINCT k FROM t
--- a/pkg/sql/logictest/testdata/logic_test/show_trace
+++ b/pkg/sql/logictest/testdata/logic_test/show_trace
@@ -587,11 +587,11 @@ query T
 SELECT message FROM [SHOW KV TRACE FOR VALUES (1, 2), (1, 1), (1, 2), (2, 1), (2, 1) UNION VALUES (1, 3), (3, 4), (1, 1)]
  WHERE message LIKE 'fetched:%' OR message LIKE 'output row%'
 ----
+output row: [1 2]
+output row: [1 1]
+output row: [2 1]
 output row: [1 3]
 output row: [3 4]
-output row: [1 1]
-output row: [1 2]
-output row: [2 1]
 
 query T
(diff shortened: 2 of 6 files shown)


andreimatei:
:lgtm: thanks buddy!

Please take the joy of removing this TODO:
https://github.com/cockroachdb/cockroach/blob/b192b3701fe98d02a48b6e03cf0bf5a565d79e40/pkg/sql/txn_restart_test.go#L1801

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/plan_columns.go, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18571#-KuMD0kfRxunS92PYy4P:-KuMD0kfRxunS92PYy4Q:b-d4dkal) ([raw file](https://github.com/cockroachdb/cockroach/blob/727dde54946f6622502b0638a0cae01e82c79a6e/pkg/sql/plan_columns.go#L114)):*
> ```Go
> 		return getPlanColumns(n.plan, mut)
> 	case *unionNode:
> 		if n.inverted {
> ```

is this change necessary? The set of columns should be the same in the two sources, no?

---

*[pkg/sql/union.go, line 146 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18571#-KuMC1VgBSLf6PyMIggH:-KuMC1VgBSLf6PyMIggI:b-89vw6m) ([raw file](https://github.com/cockroachdb/cockroach/blob/727dde54946f6622502b0638a0cae01e82c79a6e/pkg/sql/union.go#L146)):*
> ```Go
> 	emit        unionNodeEmit
> 	scratch     []byte
> 	inverted    bool
> ```

please comment the field :)

And probably a note about it / about the ordering of results should be put in the node's description too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18571#-:-KuMCqTRabTi9qM7Jcdh:bmn4pwa)*

knz:
Done. TFYR!

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/plan_columns.go, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18571#-KuMD0kfRxunS92PYy4P:-KuMJphp9-_uwml3BPMJ:bsxfbvi) ([raw file](https://github.com/cockroachdb/cockroach/blob/727dde54946f6622502b0638a0cae01e82c79a6e/pkg/sql/plan_columns.go#L114)):*

is this change necessary? The set of columns should be the same in the two sources, no?
</blockquote></details>

No: SELECT * FROM kv UNION VALUES(...)
The VALUES clause has no useful column names. This is being tested.

---

*[pkg/sql/union.go, line 146 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18571#-KuMC1VgBSLf6PyMIggH:-KuMKj91JQlIAEKy1CIB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/727dde54946f6622502b0638a0cae01e82c79a6e/pkg/sql/union.go#L146)):*

please comment the field :)

And probably a note about it / about the ordering of results should be put in the node's description too.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18571#-:-KuMJoCIDuBZzPoijLtj:b-qqwx6r)*

PR #18573, opened 2017-09-18
cherrypick: sql: restore the logical order of UNION

knz:
In order to compute EXCEPT properly, the code for
UNION/INTERSECT/EXCEPT needs to read its right operand first to
populate the map of known values. However in the non-EXCEPT case the
resulting output is non-intuitive, as a savvy user with a reasonable
order expectation would see the right operand being consumed first,
even when not necessary.

This patch restores the "natural" order.

This is not a user-facing change worth noting in docs: the ordering of
UNION's result has never been guaranteed without an ORDER BY clause.

Cherry-pick of #18571

cc @cockroachdb/release

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/explain_types
+++ b/pkg/sql/logictest/testdata/logic_test/explain_types
@@ -181,10 +181,10 @@ EXPLAIN (TYPES) VALUES (1) UNION VALUES (2)
 0  union   ·              ·                (column1 int)  ·
 1  values  ·              ·                (column1 int)  ·
 1  ·       size           1 column, 1 row  ·              ·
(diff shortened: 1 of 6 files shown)


andreimatei:
I'm not convinced this should be cherry-picked... You said it's because it touched some tests, and so future cherry-picks would need it, but I think that's pretty weak; it touched very few tests.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18573#-:-KuPh-Gp0hoBRlk1oCFx:b-wqnwfr)*

knz:
Okay, let's let this sleep until needed.

PR #18576, opened 2017-09-19
cli: bug fix and new flag --escape-special-chars

### LINKED ISSUE #18507: sql/cli: pretty results force-close the k8s interactive client with "New state of 'nil' is invalid."
**BUG REPORT**

When running this query

```
select manifest from build_version_details;
```

from a container started with

```
kubectl run cockroachdb-client -it --image=cockroachdb/cockroach:v1.0.6 --rm --restart=Never -- sql
```

against the following table (source: https://github.com/estafette/estafette-ci-api/blob/master/cockroach/migrations/00006_create_build_version_details_table.sql)

```
CREATE TABLE build_version_details (
  id INT PRIMARY KEY DEFAULT unique_rowid(),
  build_version VARCHAR(256),
  repo_source VARCHAR(256),
  repo_full_name VARCHAR(256),
  repo_branch VARCHAR(256),
  repo_revision VARCHAR(256),
  manifest TEXT,
  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT now(),
  UNIQUE INDEX build_version_details_build_version_repo_source_repo_full_name_idx (build_version, repo_source, repo_full_name)
);
```

in a CockroachDB version 1.0.6 cluster running on Google Container Engine version 1.7.3 it fails and shows the following error or something similar, but always ending in `New state of 'nil' is invalid.`:

```
root@cockroachdb-public:26257/estafette_ci_api> select manifest from build_version_details;
E0915 16:54:22.973396    5620 v2.go:147] New state of 'nil' is invalid.
```

This seems to be caused by type TEXT, because querying any of the other fields works fine:

```
select id,build_version,repo_source,repo_full_name,repo_branch,repo_revision from build_version_details;
```

From my golang application using library github.com/lib/pq it works fine though, so the data seems to be okay.

**logs**

All nodes start with the following logs:

```
2017-09-14 19:28:20 | WARNING | RUNNING IN INSECURE MODE! | cli/start.go:587
-- | -- | -- | --
2017-09-14 19:28:20 | INFO | line format: [IWEF]yymmdd hh:mm:ss.uuuuuu goid file:line msg utf8=✓ | util/log/clog.go:1013
2017-09-14 19:28:20 | INFO | [config] arguments: [/cockroach/cockroach start --logtostderr --insecure --host cockroachdb-0.cockroachdb.estafette.svc.cluster.local --http-host 0.0.0.0 --join cockroachdb-public] | util/log/clog.go:1013
2017-09-14 19:28:20 | INFO | [config] binary: CockroachDB CCL v1.0.6 (linux amd64, built 2017/09/14 15:15:48, go1.8.3) | util/log/clog.go:1013
2017-09-14 19:28:20 | INFO | [config] running on machine: cockroachdb-0 | util/log/clog.go:1013
2017-09-14 19:28:20 | INFO | [config] file created at: 2017/09/14 19:28:20 | util/log/clog.go:1013
2017-09-14 19:28:20 | INFO | CockroachDB CCL v1.0.6 (linux amd64, built 2017/09/14 15:15:48, go1.8.3) | cli/start.go:593
2017-09-14 19:28:20 | INFO | system total memory: 2.0 GiB | server/config.go:375
2017-09-14 19:28:20 | INFO | server configuration: | server/config.go:377
2017-09-14 19:28:20 | INFO | starting cockroach node | cli/start.go:340
```

knz:
Fixes #18507. (Actually a workaround but I want this to close the other GH issue - the root cause is upstream.)

For details see the two commits.

@bdarnell ok to cherry-pick in 1.1?

### DIFF
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -996,6 +996,13 @@ func Example_sql_table() {
 	c.RunWithArgs([]string{"sql", "--format=html", "-e", "select * from t.t"})
 	c.RunWithArgs([]string{"sql", "--format=raw", "-e", "select * from t.t"})
 	c.RunWithArgs([]string{"sql", "--format=records", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=pretty", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=tsv", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=csv", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=sql", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=html", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=raw", "-e", "select * from t.t"})
+	c.RunWithArgs([]string{"sql", "--escape-non-ascii", "--format=records", "-e", "select * from t.t"})
 	c.RunWithArgs([]string{"sql", "--format=pretty", "-e", "select '  hai' as x"})
 	c.RunWithArgs([]string{"sql", "--format=pretty", "-e", "explain(indent) select s from t.t union all select s from t.t"})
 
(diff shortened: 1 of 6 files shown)


knz:
@JorritSalverda can you please try this patch (you'll probably need to build manually from this branch) and check whether you still encounter the problem in an interactive session with `--escape-special-chars`?

bdarnell:
So we have to give up on all non-ascii characters to work around this? That seems pretty extreme. The flag should be more clear than "special", maybe `--ascii-only`. 

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/cli/cli_test.go, line 838 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KuQvl_taYeOMZWHGsTm:-KuQvl_taYeOMZWHGsTn:b-n7l53u) ([raw file](https://github.com/cockroachdb/cockroach/blob/d86a239f2f3282b5fb738e42d9570a88f27cae10/pkg/cli/cli_test.go#L838)):*
> ```Go
> 	// # 1 row
> 	// sql --format=csv -e select * from t.u
> 	// """foo",\foo,foo\nbar,κόσμε,a|b,܈85
> ```

I don't know how many layers of unescaping this test goes through. Does this render as "foo backslash n bar" or is there an actual newline?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-:-KuQvFwx4qDCsOwpd5KA:bhzqn78)*

knz:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/cli/cli_test.go, line 838 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KuQvl_taYeOMZWHGsTm:-KuQxNnEff6YnKTmT06c:b-tussd2) ([raw file](https://github.com/cockroachdb/cockroach/blob/d86a239f2f3282b5fb738e42d9570a88f27cae10/pkg/cli/cli_test.go#L838)):*

I don't know how many layers of unescaping this test goes through. Does this render as "foo backslash n bar" or is there an actual newline?
</blockquote></details>

yes, there i no additional unescaping. The backslash appears in the output.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576)*

JorritSalverda:
@knz I'm still working on compiling it on my Windows machine. As soon as I manage to do that I'll give it a test run.

knz:
Renamed the flag to `--escape-non-ascii`. PTAL.

bdarnell:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/cli/cli_test.go, line 838 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KuQvl_taYeOMZWHGsTm:-KwVfApRZ56ktPv36gOC:bb19yfa) ([raw file](https://github.com/cockroachdb/cockroach/blob/d86a239f2f3282b5fb738e42d9570a88f27cae10/pkg/cli/cli_test.go#L838)):*

yes, there i no additional unescaping. The backslash appears in the output.
</blockquote></details>

Do CSV readers interpret `\n` in this way? My understanding is that the way to represent a cell containing a newline in a CSV is to enclose the value in quotes and then include the literal newline:

```
// """foo",\foo,"foo
// bar",κόσμε,a|b,܈85
```

---

*[pkg/cli/cli_test.go, line 1038 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KwVfyR3792E35MBOdi_:-KwVfyR3792E35MBOdia:by0xcf3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4920434fd00ba1352271566def56bb9a58c998d/pkg/cli/cli_test.go#L1038)):*
> ```Go
> 	// κόσμε	printable UTF8
> 	// ñ	printable UTF8 using escapes
> 	// \x01	non-printable UTF8 string
> ```

We need a test case containing both a non-printable character and a double quote.

---

*[pkg/cli/cli_test.go, line 1210 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KwVjAMOKTn6PwOHM-no:-KwVjAMOKTn6PwOHM-np:bs3i681) ([raw file](https://github.com/cockroachdb/cockroach/blob/9e42f66741fadb4a9a8315a379af5fcf2d4fbb28/pkg/cli/cli_test.go#L1210)):*
> ```Go
> 	// +--------------------------------+--------------------------------+
> 	// | foo                            | printable ASCII                |
> 	// | \"foo                          | printable ASCII with quotes    |
> ```

`"foo` only contains ascii characters so it should render the same way regardless of `--escape-non-ascii`. 

---

*[pkg/cli/sql_util.go, line 667 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KwVgY22nOkPIGCXg7kL:-KwVgY23jUpO0nYVZia_:b-cwtvit) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4920434fd00ba1352271566def56bb9a58c998d/pkg/cli/sql_util.go#L667)):*
> ```Go
> 		s := fmt.Sprintf("%+q", t)
> 		// Strip the start and final quotes. The surrounding display
> 		// format (e.g. CSV/TSV) will add its own quotes.
> ```

No test case shows quotes being added to a string that has passed through this path. 

What about other output formats? I assume HTML doesn't add quotes, and therefore it probably shouldn't transform existing quotes into `\"`. Proper escaping requires knowledge of the target output format. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576)*

knz:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/cli/cli_test.go, line 838 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576#-KuQvl_taYeOMZWHGsTm:-KwdqrogfPq_GZ_SVtcA:bjrr5od) ([raw file](https://github.com/cockroachdb/cockroach/blob/d86a239f2f3282b5fb738e42d9570a88f27cae10/pkg/cli/cli_test.go#L838)):*

Do CSV readers interpret `\n` in this way? My understanding is that the way to represent a cell containing a newline in a CSV is to enclose the value in quotes and then include the literal newline:

```
// """foo",\foo,"foo
// bar",κόσμε,a|b,܈85
```
</blockquote></details>

I just noticed your concern here is orthogonal to this entire PR (both patches). Filed separately as  #19307.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18576)*

knz:
Obsoleted by #22324.

PR #17565, opened 2017-08-10
storage: disallow splits in meta2 ranges

petermattis:
Truncate range lookup scans to the range.

See #16266

PR #18578, opened 2017-09-19
kv/storage: re-enable meta2 splits

### LINKED ISSUE #18032: storage: allow splitting meta2 ranges
See #16266 for the initial issue before it diverged into a bug hunt.

#17565 disallowed splits in meta2 ranges due to problems with `RangleLookups`. `RangeDescriptors` for user ranges whose keyspan straddled meta2 ranges would not always be seen if a `RangleLookup` went to the left meta2 range. This was because meta `RangeDescriptors` are stored at `RangeMetaKey(desc.EndKey)`, which will be on the right meta2 range for these "straddling" user ranges. We should adjust the handling of `RangleLookup` in the `RangeCache` and in `Replica` to address this case. After that, we should be able to turn meta2 splits back on.

nvb:
Fixes #18032.

After painstakingly discovering problems with meta2 splits and their interaction
with `RangeLookupRequests`, we disabled the ability for meta ranges to split
in #17565. During the investigation, we found that `RangeDescriptors` for user
ranges whose keyspan straddled meta2 ranges would not be visible to `RangleLookups`
that went to the left side of a meta2 split. This was because meta
`RangeDescriptors` are stored at `RangeMetaKey(desc.EndKey)`, which will be on
the right meta2 range for these "straddling" user ranges. This change addresses
this complication and re-enables meta2 splits.

To accommodate this, we introduce the notion of a `RangeLookup` "continuation".
When a `RangleLookup` finds no matching `RangeDescriptors`, the
`RangeDescriptorCache` will issue a second lookup request to the range directly
after a previous range that returned no matching descriptors. This allows the
cache to check for the `RangeDescriptors` on the next range, where it will live
in the case of straddling user ranges. The change also required a small amount
of additional logic at the `Replica` level, because in the case of these
`RangeLookup` continuations, the desired meta scan bounds need to be adjusted.

I've pulled the first four commits out into separate PRs (#18579 and #18580), so the
meat of this change is in the fifth commit.


tschottdorf:
Glad to see you're already on this! I have a few comments and one "big" clarifying question which
we've chatted about briefly, but I've forgotten the result of: Why aren't we making `RangeLookup`
more like a "dumb" scan? The idea would be that

* `DistSender` is tasked to find the range descriptor for key "d".
* currently, `DistSender` sends `RangeLookup` to `/meta2/d` which then internally does something like a `Scan(/meta2/d.Next(), ownDesc.EndKey, limit=8)`
* instead, `DistSender` could send `Scan(/meta2/d.Next(), desc.EndKey, limit=8) directly.
* if the result set is empty, send instead to nextDesc containing `/meta2/desc.EndKey` the RPC `Scan(nextDesc.StartKey, nextDesc.EndKey, limit=8)
* repeat as needed (if we allow empty meta2 ranges or splits which aren't also a real key, or have merges).

I realize that there's special logic with regards to intent handling (reverse scan; sometimes
return, sometimes ignore intents; etc), but even then it seems that clarity could be gained if we
added a command `RangeScan` which replaces `RangeLookup` and is implemented mostly as a `Scan` with
some adapted semantics on whether intents are returned. Plus, the span declarations would be
straightforward.

---

Reviewed 1 of 1 files at r1, 6 of 6 files at r2, 3 of 3 files at r3, 2 of 2 files at r4, 14 of 14 files at r5.
Review status: all files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

*[pkg/kv/dist_sender_test.go, line 81 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPN4bOlAykhD_blgPF:-KuPN4bOlAykhD_blgPG:b-1gwwqh) ([raw file](https://github.com/cockroachdb/cockroach/blob/a51165bf84f445cc657124f6c0c764749938a5d4/pkg/kv/dist_sender_test.go#L81)):*
> ```Go
> 		},
> 	}
> 	// single user-space descriptor with three replica.
> ```

nit: s/replica/replicas/

---

*[pkg/kv/dist_sender_test.go, line 395 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPNGDnSW-E_g7IMIRA:-KuPNGDnSW-E_g7IMIRB:b-i7sqxn) ([raw file](https://github.com/cockroachdb/cockroach/blob/a51165bf84f445cc657124f6c0c764749938a5d4/pkg/kv/dist_sender_test.go#L395)):*
> ```Go
> 	}
> 
> 	// filterDescs filters the provided slice to only descriptors who would be
> ```

`descriptors which`?

---

*[pkg/kv/dist_sender_test.go, line 1482 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPO2tz7UqhHo35BafS:-KuPO2tz7UqhHo35BafT:b-sk4lxi) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/dist_sender_test.go#L1482)):*
> ```Go
> 			TransportFactory: adaptLegacyTransport(stubRPCSendFn),
> 		},
> 		RangeDescriptorDB: splitMeta2MockRangeDescriptorDB,
> ```

It would be nice to track the actual lookups performed in each test case and verify that they are as expected.

---

*[pkg/kv/range_cache.go, line 431 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPMlhR2-Ain7CNs2hF:-KuPMlhR2-Ain7CNs2hG:b-g80akn) ([raw file](https://github.com/cockroachdb/cockroach/blob/cd493ea4f0cd4a6bac16771deb5b9d28bfe32252/pkg/kv/range_cache.go#L431)):*
> ```Go
> 	// Tag inner operations.
> 	ctx = log.WithLogTag(ctx, "range-lookup", nil)
> 	return rdc.db.RangeLookup(ctx, metadataKey, desc, useReverseScan)
> ```

what am I missing? Where did `descs` and `prefetched` go?

---

*[pkg/kv/range_cache.go, line 433 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPOBno9DMwEvdfcGCr:-KuPOBno9DMwEvdfcGCs:bqc250p) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L433)):*
> ```Go
> 	ctx = log.WithLogTag(ctx, "range-lookup", metadataKey)
> 	descs, prefetched, err := rdc.db.RangeLookup(
> 		ctx, metadataKey, desc, useReverseScan, false /* continuation */)
> ```

`/* !continuation */`

---

*[pkg/kv/range_cache.go, line 434 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPOW7D0R0pc3kpMWjd:-KuPOW7D0R0pc3kpMWje:b-iwg2hy) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L434)):*
> ```Go
> 	descs, prefetched, err := rdc.db.RangeLookup(
> 		ctx, metadataKey, desc, useReverseScan, false /* continuation */)
> 	if err == nil && len(descs) == 0 {
> ```

I think we should deal with empty meta2 splits (i.e. meta ranges for which the actual next meta key and the initially hit range are not neighbors). This can't happen today because a) we always split on an existing key and b) we don't merge, but neither of those are necessarily for eternity and it seems that you're already close to handling this (can't do better than a linear search, which in itself is a little bad, of course). Future-proofing that limitation in (in which case we'd somehow need to prevent a change in our split semantics plus make sure we realize to go back and fix things on merges) seems like more work to me, but maybe I'm missing something.

---

*[pkg/kv/range_cache.go, line 449 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPPCxpNMGAKUovGAHt:-KuPPCxpNMGAKUovGAHu:b-tr6aar) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L449)):*
> ```Go
> // scenario:
> //
> //   range 1 [a, d):
> ```

I think this would be clearer to the untrained eye as `range 1 [/meta2/a, /meta2/d)`, `/meta2/b -> [a,b)`, etc.

---

*[pkg/kv/range_cache.go, line 482 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPPyippFkpkUh3JOfn:-KuPPyippFkpkUh3JOfo:bsckgrv) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L482)):*
> ```Go
> 	ctx = log.WithLogTag(ctx, "range-lookup-continuation", nextKey)
> 
> 	// FOR REVIEW: is it worth worrying about empty meta ranges and scanning
> ```

Ah, hadn't seen this, but in my current opinion yes.

---

*[pkg/kv/range_cache_test.go, line 285 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPQusomJyiHrUke0T1:-KuPQusomJyiHrUke0T2:bn4rtum) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache_test.go#L285)):*
> ```Go
> 		continuation      bool
> 		metaDesc          roachpb.RangeDescriptor
> 		expectedRspansMap map[bool][]roachpb.RSpan
> ```

Comment on the `bool` here.

---

*[pkg/kv/range_cache_test.go, line 286 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPSRQk6ES0XqieedTF:-KuPSRQk6ES0XqieedTG:bkycj5w) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache_test.go#L286)):*
> ```Go
> 		metaDesc          roachpb.RangeDescriptor
> 		expectedRspansMap map[bool][]roachpb.RSpan
> 	}{
> ```

Add a comment explaining which of the descriptors are usually the prefetched ones. Can be on the first test case too, just nice to loop the reader into what's happening (I had to dig a little bit to find out).

---

*[pkg/kv/range_cache_test.go, line 362 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPSGeBcUtW_7AialKC:-KuPSGeBcUtW_7AialKD:b-unigsm) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache_test.go#L362)):*
> ```Go
> 			expectedRspansMap: map[bool][]roachpb.RSpan{
> 				false: {
> 					roachpb.RSpan{Key: roachpb.RKey("l"), EndKey: roachpb.RKey("m")}, // real
> ```

Add a comment that it's visible now.

---

*[pkg/roachpb/api.proto, line 443 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPSomH8gMJOdvqakNW:-KuPSomH8gMJOdvqakNX:b-gjtcxv) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/roachpb/api.proto#L443)):*
> ```protobuf
>   // standard meaning is appropriate here and that it's not confusing, but
>   // Lisp-enthusiasts may disagree. 
>   optional bool continuation = 5 [(gogoproto.nullable) = false];
> ```

I have no better suggestion.

---

*[pkg/storage/client_split_test.go, line 2336 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPTMVuqiexPUjtZygB:-KuPTMVuqiexPUjtZygC:b-w5va6w) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/storage/client_split_test.go#L2336)):*
> ```Go
> // TestRangeLookupAfterMeta2Split verifies that RangeLookups succeed even when
> // user ranges span the boundary of two split meta2 ranges.
> func TestRangeLookupAfterMeta2Split(t *testing.T) {
> ```

What happens with this test before your fixes?

---

*[pkg/storage/replica_command.go, line 1221 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPU0SPmsl_odRo5bYU:-KuPU0SPmsl_odRo5bYV:b-5x7tiz) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/storage/replica_command.go#L1221)):*
> ```Go
> 
> 	ts, txn, consistent := h.Timestamp, h.Txn, h.ReadConsistency != roachpb.INCONSISTENT
> 	reverse, continuation, rangeCount := args.Reverse, args.Continuation, int64(args.MaxRanges)
> ```

I wonder if (migration concerns aside) it wouldn't be better to just have three modes: `reverse`, `forward`, `forward-continued`.

---

*[pkg/storage/replica_gc_queue.go, line 211 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPUIWz3LsdFh12HLRq:-KuPUIWz3LsdFh12HLRr:b-ypagcc) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/storage/replica_gc_queue.go#L211)):*
> ```Go
> 	reply := br.Responses[0].GetInner().(*roachpb.RangeLookupResponse)
> 
> 	// TODO DURING REVIEW: Do we need to consider this case? Do we need to make
> ```

This code definitely needs to be able to get the right descriptor, if that's what you mean. Otherwise, we're never going to garbage-collect ranges if the lookup requires a continuation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-:-KuPUVxMURxomFkCmVbO:b-slq798)*

nvanbenschoten:
This also doesn't yet properly handle cluster versions or gating the ability to split meta2 until all nodes can handle the required `Replica`-side changes.

petermattis:

---

*[pkg/kv/range_cache.go, line 460 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPtWN66MnkyXvbNExc:-KuPtWN66MnkyXvbNExd:brlu5vs) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L460)):*
> ```Go
> // Now consider looking up the range containing key `d`. The RangeDescriptorDB
> // routing logic would send the RangeLookup request to range 1 since `d` lies
> // within the bounds of that range. But notice that the range descriptor containing
> ```

There is confusion here: `d` does not lie within the bounds of range 1. Per Toby's comment, I think you need to highlight the use of the `/meta` keys. I realize that part of this example arose from my commentary on the issue.

---

*[pkg/kv/range_cache.go, line 463 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPu5PFsdDfWR4cmymk:-KuPu5PFsdDfWR4cmyml:b-ojqo5) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L463)):*
> ```Go
> // `d` lies in range 2. This means that no matching RangeDescriptors will be
> // found on range 1 and returned from the first RangeLookup. In fact, a RangeLookup
> // for any key between ['c','d') will create this scenerio.
> ```

I'm not following this last sentence. A lookup for `cc` will be routed to range 1 where it will be found. 

---

*[pkg/kv/range_cache.go, line 468 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuPuXV-9_RVqwdl_jdW:-KuPuXV-9_RVqwdl_jdX:b-jp0j27) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/kv/range_cache.go#L468)):*
> ```Go
> // directly following the one that resulted in the lookup miss, in this case
> // range 2. This "continuation" lookup will begin scanning directly at the
> // range's StartKey and will be guaranteed to find the descriptor.
> ```

I need some more explanation of why the continuation RangeLookupRequest needs to be marked as such. Perhaps there is some commentary that I've missed explaining this. I'm not seeing why you need to adjust the scan bounds based on whether the request is a "continuation". 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578)*

bdarnell:
Reviewed 1 of 2 files at r4, 14 of 14 files at r5.
Review status: all files reviewed at latest revision, 21 unresolved discussions, all commit checks successful.

---

*[pkg/keys/spans.go, line 21 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuRGYsEY8AePxinfrJG:-KuRGYsEY8AePxinfrJH:b-8tvd46) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/keys/spans.go#L21)):*
> ```Go
> var (
> 	// Meta1Span holds all first level addressing.
> 	Meta1Span = roachpb.Span{Key: roachpb.KeyMin, EndKey: Meta2Prefix}
> ```

I think this needs a cluster version guard. We can't split meta2 until the upgrade to version 1.2 is finalized.

---

*[pkg/roachpb/api.proto, line 442 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuRIZctDPOvLkBWVV1v:-KuRIZctDPOvLkBWVV1w:b-olcyfg) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/roachpb/api.proto#L442)):*
> ```protobuf
>   // DURING REVIEW: "continuation" is an overloaded term in CS. I think its
>   // standard meaning is appropriate here and that it's not confusing, but
>   // Lisp-enthusiasts may disagree. 
> ```

`(fine-p (with 'me))`

I used "continuation" the same way in the google reader api. 

---

*[pkg/storage/replica_command.go, line 1168 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578#-KuRKSPO9V1TcvtWionD:-KuRKSPO9V1TcvtWionE:b-3m2l7m) ([raw file](https://github.com/cockroachdb/cockroach/blob/06eaa0a09329fd7d50fca5fc7c70e0d944b32685/pkg/storage/replica_command.go#L1168)):*
> ```Go
> 		// desc.StartKey == key because keys.MetaScanBounds internally performs
> 		// a key.Next call when constructing the span bounds.
> 		span.Key = desc.StartKey
> ```

Wait, so `RangeLookupRequest.Key` is outside of this replica's bounds? Why isn't this getting stopped by `Replica.requestCanProceed`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18578)*

nvanbenschoten:
Replaced by #18970.

PR #18579, opened 2017-09-19
storage: fix declareKeysRangeLookup to include all possible keys

nvb:
Previously, `declareKeysRangeLookup` only included the `RangeLookup's` lookup key,
even though the lookup would scan over a large span of keys in the meta range.
This change fixes this by declaring the correct upper bound on the span of keys
that the `RangeLookup` can touch.

This was not a serious issue in practice because `DistSender` only performs
inconsistent `RangeLookups` at the moment. `replicaGCQueue` does attempt to
perform consistent `RangeLookups`, but it's not clear if this has ever caused an
issue. Either way, we should be able to perform consistent lookups correctly.

### DIFF
--- a/pkg/storage/batcheval/cmd_range_lookup.go
+++ b/pkg/storage/batcheval/cmd_range_lookup.go
@@ -35,8 +35,60 @@ func init() {
 func declareKeysDeprecatedRangeLookup(
 	desc roachpb.RangeDescriptor, header roachpb.Header, req roachpb.Request, spans *spanset.SpanSet,
 ) {
-	DefaultDeclareKeys(desc, header, req, spans)
+	lookupReq := req.(*roachpb.DeprecatedRangeLookupRequest)
+
+	if keys.IsLocal(lookupReq.Key) {
+		// Error will be thrown during evaluation.
+		return
+	}
+	key := roachpb.RKey(lookupReq.Key)
+
+	// RangeLookupRequests depend on the range descriptor because they need
+	// to determine which range descriptors are within the local range.
 	spans.Add(spanset.SpanReadOnly, roachpb.Span{Key: keys.RangeDescriptorKey(desc.StartKey)})
+
+	// Both forward and reverse RangeLookupRequests scan forward initially. We are
+	// unable to bound this initial scan any further than from the lookup key to
+	// the end of the current descriptor.
+	scanBounds, err := rangeLookupScanBounds(&desc, key, false /* reverse */)
+	if err != nil {
+		// Errors will be caught during evaluation.
+		return
+	}
+	spans.Add(spanset.SpanReadOnly, roachpb.Span{
+		Key:    scanBounds.Key.AsRawKey(),
+		EndKey: scanBounds.EndKey.AsRawKey(),
+	})
+
+	if lookupReq.Reverse {
+		// A reverse RangeLookupRequest also scans backwards.
+		revScanBounds, err := rangeLookupScanBounds(&desc, key, true /* reverse */)
+		if err != nil {
+			// Errors will be caught during evaluation.
+			return
+		}
+		spans.Add(spanset.SpanReadOnly, roachpb.Span{
+			Key:    revScanBounds.Key.AsRawKey(),
+			EndKey: revScanBounds.EndKey.AsRawKey(),
+		})
+	}
+}
+
+// rangeLookupScanBounds returns the range [start,end) within the bounds of the
+// provided RangeDescriptor which the desired meta record can be found by means
+// of an engine scan.
+func rangeLookupScanBounds(
+	desc *roachpb.RangeDescriptor, key roachpb.RKey, reverse bool,
+) (roachpb.RSpan, error) {
+	boundsFn := keys.MetaScanBounds
+	if reverse {
+		boundsFn = keys.MetaReverseScanBounds
+	}
+	span, err := boundsFn(key)
+	if err != nil {
+		return roachpb.RSpan{}, err
+	}
+	return span.Intersect(desc)
 }
 
 // DeprecatedRangeLookup is used to look up RangeDescriptors - a RangeDescriptor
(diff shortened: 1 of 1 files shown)


tschottdorf:
Reviewed 1 of 1 files at r1.

---

*[pkg/storage/replica_command.go, line 1107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-KuPL3JIFxLQOgqq1yfH:-KuPL3JIFxLQOgqq1yfI:b94chzj) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1107)):*
> ```Go
> 
> 		// DURING REVIEW: should we change the signature of DeclareKeys to
> 		// return an error? Is this worth it?
> ```

Hmm, that does seem to make sense. @bdarnell WDYT?

---

*[pkg/storage/replica_command.go, line 1117 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-KuPLy8bklUgDjmV83Ex:-KuPLy8bklUgDjmV83Ey:bj8jycz) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1117)):*
> ```Go
> 
> 	// Both forward and reverse RangeLookupRequests scan forward initially. We are
> 	// unable to bound this initial scan and further than from the lookup key to
> ```

s/and/any

---

*[pkg/storage/replica_command.go, line 1119 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-KuPLG_ZFEFWGMKkuTVq:-KuPLG_ZFEFWGMKkuTVr:bi6e2nt) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1119)):*
> ```Go
> 	// unable to bound this initial scan and further than from the lookup key to
> 	// the end of the current descriptor.
> 	scanBounds, err := rangeLookupScanBounds(&desc, key, false /* reverse */)
> ```

nit: /* !reverse */

---

*[pkg/storage/replica_command.go, line 1130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-KuPLNAV4TTeIwhhLokM:-KuPLNAV4TTeIwhhLokN:bnr577d) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1130)):*
> ```Go
> 
> 	if lookupReq.Reverse {
> 		// A revese RangeLookupRequest also scans backwards.
> ```

`reverse`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579)*

tschottdorf:
I would expect that the `spanSetIterator` tests would've caught this, but apparently we don't exercise it well, or it's because we only use inconsistent lookups? Any idea why nothing ever tripped the span set tester?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-:-KuPM5v6luGjcmVgk1Jg:b-3muayp)*

bdarnell:
:lgtm:

The tests never caught this because the only times that we do consistent lookups are in replicaGCQueue, which also specifies a known range descriptor's key with MaxRanges:1. 

---

---

*[pkg/storage/replica_command.go, line 1107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-KuPL3JIFxLQOgqq1yfH:-KuRAZZVmnDo6ckDuUY0:bqmyis7) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1107)):*

Hmm, that does seem to make sense. @bdarnell WDYT?
</blockquote></details>

Probably. Returning without error is technically accurate (the command evaluation will finish without accessing any keys), although it might simplify things a bit if we could return errors here instead of relying on consistency at a distance between declareKeys and eval. 

We also have a few places that use Transaction.AssertInitialized in declareKeys; if we add an error return, those could be updated to return errors instead of panicking. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18579#-:-KuRCugU5tUH1z1F1tvq:b-log4s4)*

nvanbenschoten:
Even though RangeLookupRequests are now deprecated, I'm planning on reviving this PR to unblock #23997.

nvanbenschoten:
Rebased on master.

---

---

*[pkg/storage/replica_command.go, line 1107 at r1](https://beta.reviewable.io/reviews/cockroachdb/cockroach/18579#-KuPL3JIFxLQOgqq1yfH:-L7zSSyyDIstjQ4-GTSb:b-u77nxl) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1107)):*

Probably. Returning without error is technically accurate (the command evaluation will finish without accessing any keys), although it might simplify things a bit if we could return errors here instead of relying on consistency at a distance between declareKeys and eval. 

We also have a few places that use Transaction.AssertInitialized in declareKeys; if we add an error return, those could be updated to return errors instead of panicking. 
</blockquote></details>

I'm going to leave this as-is because the plumbing required to return an error is unfortunate and this entire file will be removed soon for 2.1 anyway, once `DeprecatedRangeLookup` is gone for good.

---

*[pkg/storage/replica_command.go, line 1117 at r1](https://beta.reviewable.io/reviews/cockroachdb/cockroach/18579#-KuPLy8bklUgDjmV83Ex:-L7zSHdM8c4GG6RS6I8K:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1117)):*

s/and/any
</blockquote></details>

Done.

---

*[pkg/storage/replica_command.go, line 1119 at r1](https://beta.reviewable.io/reviews/cockroachdb/cockroach/18579#-KuPLG_ZFEFWGMKkuTVq:-L7zSgr743_YAdzLTAN5:b-6yr30) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1119)):*

nit: /* !reverse */
</blockquote></details>

Our style guide from 5 months in the future disagrees.

---

*[pkg/storage/replica_command.go, line 1130 at r1](https://beta.reviewable.io/reviews/cockroachdb/cockroach/18579#-KuPLNAV4TTeIwhhLokM:-L7zSJ4m7-LPjc6K4SeC:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b66b60f10d51670c74d9531ec32f57aca72ba3e/pkg/storage/replica_command.go#L1130)):*

`reverse`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/cockroach/18579#-:-L7zUgynBq2g6tIUQ8ej:bc65pbk)*

bdarnell:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/cockroach/18579#-:-L7zXfzn4vRVygJoO5YO:bnfp4nl)*

PR #18581, opened 2017-09-19
kv: remove obnoxious RangeDescriptorCache stringLocked logs

nvb:
This logs are very loud in logspy and aren't helpful. They print the
entire contents of the `RangeDescriptorCache`, which can grow up to
`1e6` entries in size on large clusters. Because it prints every range
descriptor in the cache, it is also picked up by a grep for any rangeId,
even those unrelated to the lookup/eviction.


tschottdorf:
LGTM. Also cherry-pick to 1.1 please or we'll have this problem until 2019.

Issue #6981, opened 2016-06-01
sql: support inet data types


abh:
The inet and cidr data types in postgresql are very powerful for applications that deal with IP addresses; in particular IPv6 ones.

https://www.postgresql.org/docs/current/static/datatype-net-types.html
https://www.postgresql.org/docs/current/static/functions-net.html

jordanlewis:
This is required by a popular app that runs on Postgres - Mastodon.

nickjacques:
+1, [this is required for PowerDNS schema compatibility when using the postgresql backend.](https://github.com/PowerDNS/pdns/blob/master/modules/gpgsqlbackend/schema.pgsql.sql#L40)

dianasaur323:
adding @nstewart to the chain

aaomidi:
This is absolutely required for any modern day application. It being open for more than a year with barely any updates is worrisome.

nstewart:
This is on our radar and is scheduled to be part of the 1.2 release (mid-2018). We expect it to be in an alpha version of CockroachDB much earlier than that.

vivekmenezes:
We are not adding CIDR support now until we find a need for it. Thanks!

aaomidi:
What defines a need?

lego:
In the case of CIDR, if people bring up major projects that make use of that column specifically then I'd be more than happy to spend some time adding it. To my knowledge, CIDR is strictly less useful than INET as it seems to be a subset of INET. I'll be closing this as I've added the INET type and it's built-ins, but I created an issue to track the CIDR type, #18846.

I also didn't add many of the operators postgres supports yet so there's another issue to track that, #18847. I didn't find the operators were in common use so I just didn't get around to it yet.

KempWatson:
NetBox? Any modern cloud app (like, the apps that would want CRDB), would have good use for all three network datatypes.

knz:
@KempWatson you're lucky I was running a random script on the github database to notice your comment. We don't see comments on closed issues otherwise. Please file a new issue if you find this important.

PR #7069, opened 2016-06-06
sql/parser: Add FAMILY as reserved keyword to avoid ambiguity

nvb:
Fixes #7061.

This change adds `FAMILY` as a reserved keyword.

This is necessary to resolve ambiguity with our grammar. The ambiguity
is an artifact of `FAMILY` not being a reserved keyword. Without adding to
our reserved keywords or using multiple character lookahead, the
previous syntax was throwing a shift/reduce error. The ambiguity caused
by this error is shown with the following situation:

```
CREATE TABLE t (x type_t (opts));
RENAME t.x TO t.family;
SHOW CREATE TABLE t;

+-------+---------------------------------------+
| Table |                CreateTable            |
+-------+---------------------------------------+
| t     | CREATE TABLE t (family type_t (opts)) |
+-------+---------------------------------------+
```

Now that `FAMILY` is a reserved keyword, `family` will not be a valid
column name, and therefore the rename will not be possible.

@knz

---

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach/7069)

PR #18585, opened 2017-09-19
sql: Add INET type built-in functions

lgo:
Adds further support for #6981

This adds several built-in functions for the INET column. See
https://www.postgresql.org/docs/9.6/static/functions-net.html for
reference. The added built-ins are:
- abbrev(inet) text
- broadcast(inet) text
- family(inet) int (this is "ip_family" due to a YACC grammar conflict)
- host(inet) text
- hostmask(inet) inet
- masklen(inet) int
- netmask(inet) inet
- set_masklen(inet, int) inet
- text(inet) text
- inet_same_family(inet, inet) boolean

This does not include the built-ins
- network(inet) cidr
- inet_merge(inet, inet) cidr


knz:
Can you explain a little better what the problem is with having a function called "family". What makes this difficult?

---

Reviewed 6 of 6 files at r1.

---

*[pkg/sql/logictest/testdata/logic_test/inet, line 290 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWTJyCsUE6BgNQp-hQ:-KuWTJyCsUE6BgNQp-hR:bz2nv9l) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/logictest/testdata/logic_test/inet#L290)):*
> ```
> 
> query T
> SELECT abbrev('10.1.0.0/16'::INET)
> ```

huh, what's the difference between `abbrev` and `::string` then?

Please provide examples, if any, of where the input to `abbrev` will be different from the output.

---

*[pkg/sql/parser/builtins.go, line 407 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWSrWFBuUvPqa4GZru:-KuWSrWFBuUvPqa4GZrv:b-24yg01) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/parser/builtins.go#L407)):*
> ```Go
> 			Types:      ArgTypes{{"val", TypeINet}},
> 			ReturnType: fixedReturnType(TypeString),
> 			fn: func(_ *EvalContext, args Datums) (Datum, error) {
> ```

IT would help your reviewer(s) and future maintainers of this code to have a comment here (and for each function below) with

1) an example of the desired output
2) a link to the pg docs where these functions are documented/specified

---

*[pkg/sql/parser/builtins.go, line 411 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWSnNIe7WvfFKD4Mfl:-KuWSnNIe7WvfFKD4Mfm:b-jnp01o) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/parser/builtins.go#L411)):*
> ```Go
> 				return NewDString(dIPAddr.IPAddr.String()), nil
> 			},
> 			Info: "Abbreviated display format as text",
> ```

The info strings, here and below, must be composed of full sentences with final periods.

---

*[pkg/sql/parser/builtins_test.go, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWT49lz2QpyabwxO7X:-KuWT49lz2QpyabwxO7Y:b-y4441r) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/parser/builtins_test.go#L17)):*
> ```Go
> package parser
> 
> import (
> ```

Spurious change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KuWTyufyTDfzVsf2EDP:bm198eb)*

knz:
I just checked my networking book and what you have been calling "mask" all this time is really the "prefix length" [1](http://www.gadgetwiz.com/network/netmask.html) [2](https://www.cisco.com/c/en/us/about/press/internet-protocol-journal/back-issues/table-contents-12/ip-addresses.html). The mask is something else entirely (what's computed by the hostmask/netmask functions). It would do well to your patch to rename the fields and variables named "mask" to "prefixlen" where appropriate.

Otherwise we could sit together for a minute and check what's going on with that grammar.

---

Reviewed 2 of 2 files at r2.

---

*[pkg/sql/logictest/testdata/logic_test/inet, line 290 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWTJyCsUE6BgNQp-hQ:-KutTpsDNYQY09-uLKjD:bz1s0bc) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/logictest/testdata/logic_test/inet#L290)):*

For `inet` there is no difference between `abbrev` and `::string`. Where this ends up being different is when the input is a `cidr` type (see https://www.postgresql.org/docs/9.6/static/functions-net.html#CIDR-INET-FUNCTIONS-TABLE).

Although I'll also point out that the input and output differ in this case for `abbrev` (which is the same as `::string`). This is just because we infer that omitted octets are zeros. 

```
postgres=# select abbrev('10.0/16'::inet);
   abbrev
-------------
 10.0.0.0/16
```


Something I also want to call out is that there is a _very small_ behavioural difference between `abbrev`/`::string` and `text(x::inet)`. In postgres, for `text(...)` it _always_ has the mask length suffix, whereas `abbrev` and `::string` omit the mask length if it's a complete mask (32 for IPv4 or 128 for IPv6).
</blockquote></details>

These differences should be higlighted in the logic test, if only with a comment that explains what is going on.

---

*[pkg/sql/parser/builtins.go, line 411 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWSnNIe7WvfFKD4Mfl:-KutU9bHx4ntzhx_ZDQM:b9slr3a) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/parser/builtins.go#L411)):*

Done.

Something I'm a little uncertain about is the usage of the term "IP address". It seems more intuitive to me to use "IP address" to refer to the value as opposed to INET but this also makes using of "IP address" a bit ambiguous when referring to one of the two parts of the INET type: the address vs. the mask length.

Thoughts on the usage? For now, I changed most of the references to "IP value" to "the value" to refer to the input.
</blockquote></details>

I think it will help your reader to use the term "combined IP address and prefix length" here and below.

---

*[pkg/sql/parser/builtins.go, line 440 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUCTdX1IKahLkqMlo:-KutUCTdX1IKahLkqMlp:boyvjsk) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L440)):*
> ```Go
> 				return &DIPAddr{IPAddr: broadcastIPAddr}, nil
> 			},
> 			Info: "Gets the broadcast address for network represented by the value." +
> ```

...for the network address represented...

---

*[pkg/sql/parser/builtins.go, line 473 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUTn-pkvKYw9O534n:-KutUTn0x99msCy7f4QV:b2eann7) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L473)):*
> ```Go
> 				return NewDString(s), nil
> 			},
> 			Info: "Extracts the address part of value as text, omitting the mask length." +
> ```

So this becomes "Extracts the address part of the combined address/prefixlen value".

---

*[pkg/sql/parser/builtins.go, line 487 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUlBSO36O9JwJx3vH:-KutUlBTV5R8lR68ZrzM:bvd2oee) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L487)):*
> ```Go
> 				return &DIPAddr{ipAddr}, nil
> 			},
> 			Info: "Creates an IP address retaining only the host mask part of the value." +
> ```

"Creates an IP host mask corresponding to the prefix length in the value."

---

*[pkg/sql/parser/builtins.go, line 500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUqgbokue1uudT716:-KutUqgbokue1uudT717:b-xly7e1) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L500)):*
> ```Go
> 				return NewDInt(DInt(dIPAddr.Mask)), nil
> 			},
> 			Info: "Gets the mask length of the value." +
> ```

"Retrieves the prefix length stored in the value."

---

*[pkg/sql/parser/builtins.go, line 514 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUv4V9UkFvdYJ0u_o:-KutUv4V9UkFvdYJ0u_p:b-7ugbhk) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L514)):*
> ```Go
> 				return &DIPAddr{ipAddr}, nil
> 			},
> 			Info: "Creates an IP address retaining only the network mask part of the value." +
> ```

"Creates an IP network mask corresponding to the prefix length in the value."

---

*[pkg/sql/parser/builtins.go, line 536 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutV-oykukM88IP-Hkj:-KutV-ozri2xOwkA3-pP:b-e8ylcd) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L536)):*
> ```Go
> 				return &DIPAddr{IPAddr: ipaddr.IPAddr{Family: dIPAddr.Family, Addr: dIPAddr.Addr, Mask: byte(mask)}}, nil
> 			},
> 			Info: "Sets the mask length of `val` to `mask`.\n\n" +
> ```

Sets the prefix length.

---

*[pkg/sql/parser/builtins.go, line 554 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutV1Pad5kA2LW4Z2Nl:-KutV1Pad5kA2LW4Z2Nm:bb7eybu) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L554)):*
> ```Go
> 				return NewDString(s), nil
> 			},
> 			Info: "Converts the IP address and mask length to text.",
> ```

ditt

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KutV5KjCZyIHA0m3OK4:b-kywkiz)*

lego:
Good catch! It's quite funny how easily one's choices leak everywhere else (postgres in this case).

---

Review status: 3 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/sql/logictest/testdata/logic_test/inet, line 290 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWTJyCsUE6BgNQp-hQ:-KutgCcKGe5CQBkljl-o:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/logictest/testdata/logic_test/inet#L290)):*

These differences should be higlighted in the logic test, if only with a comment that explains what is going on.
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 411 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KuWSnNIe7WvfFKD4Mfl:-KutkW-Tr3WEyo_SAmFy:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4cbde88e6b416665aa35a320a024b5add8c07b45/pkg/sql/parser/builtins.go#L411)):*

I think it will help your reader to use the term "combined IP address and prefix length" here and below.
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 440 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUCTdX1IKahLkqMlo:-KutlG8dy8sQ-OHrK2uo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L440)):*

...for the network address represented...
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 473 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUTn-pkvKYw9O534n:-KutlFiGCwdZBKkh1rdx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L473)):*

So this becomes "Extracts the address part of the combined address/prefixlen value".
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 487 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUlBSO36O9JwJx3vH:-KutlE9H5w7kVfVRBgxQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L487)):*

"Creates an IP host mask corresponding to the prefix length in the value."
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUqgbokue1uudT716:-KutlDmO_mtDWIheptf0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L500)):*

"Retrieves the prefix length stored in the value."
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 514 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutUv4V9UkFvdYJ0u_o:-KutlDCqbcTK9lM8yAyq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L514)):*

"Creates an IP network mask corresponding to the prefix length in the value."
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 536 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutV-oykukM88IP-Hkj:-KutlCE24eWWUy90Z-O6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L536)):*

Sets the prefix length.
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 554 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-KutV1Pad5kA2LW4Z2Nl:-KutlBfdmd2uYHs9IOpk:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad0edea8ca2d3e2e91ffc4b926942ee512b75653/pkg/sql/parser/builtins.go#L554)):*

ditt
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KutlKMtubf1L3O2EMOL:b-1yla0l)*

lego:
So I ended up getting the `family` function working in the grammar by moving `FAMILY` from `reserved_keyword` to `type_func_name_keyword`.

It was added to `reserved_keyword` in https://github.com/cockroachdb/cockroach/pull/7069, which I don't think will be a regression as to my knowledge `type_func_name_keyword` keeps the keyword out of possible column names (without quotes). 

---

Review status: 3 of 7 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KutrwRsv5R1eCaJ9Zaa:bdi2oos)*

knz:
Excellent. The finish line is near.

---

Reviewed 4 of 4 files at r3.

---

*[pkg/sql/parser/builtins.go, line 440 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kuttq0Js8S2bKiuWsyu:-Kuttq0K6IVy4Km5dKCF:b-zgiatl) ([raw file](https://github.com/cockroachdb/cockroach/blob/7c845a3f75fdd78261ae517f6f333d3b6140b380/pkg/sql/parser/builtins.go#L440)):*
> ```Go
> 				return &DIPAddr{IPAddr: broadcastIPAddr}, nil
> 			},
> 			Info: "Gets the broadcast address for network address represented by the value." +
> ```

for the* network address

---

*[pkg/sql/parser/builtins.go, line 523 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kuttyc30SFdfeQoiBHa:-Kuttyc30SFdfeQoiBHb:b-i8bxxl) ([raw file](https://github.com/cockroachdb/cockroach/blob/7c845a3f75fdd78261ae517f6f333d3b6140b380/pkg/sql/parser/builtins.go#L523)):*
> ```Go
> 			Types: ArgTypes{
> 				{"val", TypeINet},
> 				{"mask", TypeInt},
> ```

s/mask/prefixlen here and below

---

*[pkg/sql/parser/sql.y, line 6503 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kutu2P6Gwcz79naxJc5:-Kutu2P6Gwcz79naxJc6:b-n3r360) ([raw file](https://github.com/cockroachdb/cockroach/blob/028bbd8116ac937d83b6b9f8424a7069c7cd64dd/pkg/sql/parser/sql.y#L6503)):*
> ```Yacc
> | EXCEPT
> | FALSE
> | FAMILY
> ```

check this in parse_test.go too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-Kutu5Hbo_lw1hjGAltI:b-3laepr)*

lego:

---

*[pkg/sql/parser/builtins.go, line 440 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kuttq0Js8S2bKiuWsyu:-Kuu9CGC2Jfy6m4kFb3B:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7c845a3f75fdd78261ae517f6f333d3b6140b380/pkg/sql/parser/builtins.go#L440)):*

for the* network address
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 523 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kuttyc30SFdfeQoiBHa:-Kuu9CxGLOn0VhM-vjSk:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7c845a3f75fdd78261ae517f6f333d3b6140b380/pkg/sql/parser/builtins.go#L523)):*

s/mask/prefixlen here and below
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 6503 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kutu2P6Gwcz79naxJc5:-Kuu9DTbhFFAEZawdjHf:bln26zu) ([raw file](https://github.com/cockroachdb/cockroach/blob/028bbd8116ac937d83b6b9f8424a7069c7cd64dd/pkg/sql/parser/sql.y#L6503)):*

check this in parse_test.go too.
</blockquote></details>

Hah, you know best.

I added the test case and it looks like it's getting transformed.

When trying to round trip
```
 expected SELECT FAMILY(x), but found SELECT "family"(x)
```

and from the cli
```
root@:26257/test> select FAMILY('192.168.0.1');
+-------------------------+
| "family"('192.168.0.1') |
+-------------------------+
|                       4 |
+-------------------------+
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585)*

lego:

---

*[pkg/sql/parser/sql.y, line 6503 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kutu2P6Gwcz79naxJc5:-Kuu9XZEpAABRD2q0LPH:b-5xl0ql) ([raw file](https://github.com/cockroachdb/cockroach/blob/028bbd8116ac937d83b6b9f8424a7069c7cd64dd/pkg/sql/parser/sql.y#L6503)):*

Hah, you know best.

I added the test case and it looks like it's getting transformed.

When trying to round trip
```
 expected SELECT FAMILY(x), but found SELECT "family"(x)
```

and from the cli
```
root@:26257/test> select FAMILY('192.168.0.1');
+-------------------------+
| "family"('192.168.0.1') |
+-------------------------+
|                       4 |
+-------------------------+
```
</blockquote></details>

Any idea why this is happening? It functionally works, but it's a funky behavior that the function name is getting wrapped in quotes. I tried moving the `FAMILY` into `col_name_keyword` also but that causes shift/reduce errors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585)*

knz:
You mean that you can enter `family(...)` but the pretty print renders as `"family"(...)`? If so, that's OK. Use a test in `TestParse2`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KuuFAnISypRzevJSzPh:b-urozlc)*

lego:
Yea that's the case. Then awesome, :raised_hands: done.

(also woops, I seemed to have not pushed when I published my last round of comments)

---

Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KuuQtrTATG8msRAzoHY:b-ojwg9s)*

knz:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---

*[pkg/sql/parser/sql.y, line 6503 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-Kutu2P6Gwcz79naxJc5:-KuuU_mIWcsuXujDhAUa:b-jiwb66) ([raw file](https://github.com/cockroachdb/cockroach/blob/028bbd8116ac937d83b6b9f8424a7069c7cd64dd/pkg/sql/parser/sql.y#L6503)):*

Any idea why this is happening? It functionally works, but it's a funky behavior that the function name is getting wrapped in quotes. I tried moving the `FAMILY` into `col_name_keyword` also but that causes shift/reduce errors.
</blockquote></details>

nothing to see here. No problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18585#-:-KuuUda14Pvys94F3z2C:bnfp4nl)*

Issue #15363, opened 2017-04-26
sql: EvalContext is insufficiently initialized in most places where it is used


knz:
Encountered while working on  #15357: contrary to `SemaContext`, which has `MakeSemaContext` constructor, `EvalContext` does not have a constructor and it is used in many places using the default values and a quick-and-dirty, minimal "set the timestamps" initialization at the point of use.

Setting the timestamps is only part of the story however

1) this puts the onus on the point of use to do the correct initialization, and a quick scan through the code reveals this is not done properly in several places

2) the evalcontext may contain other fields that need to be initialized properly for built-ins to make sense. For example NodeID, used by `unique_rowid`: if the default of 0 is used on separate nodes, we risk generating twice the same "unique rowid". This is especially a problem since #15344 has been merged.

3) This hampers things like the work on #15357 for example which want to see useful information about the current session in EvalContext.

So the EvalContext should really have a constructor that populates it with useful values.

knz:
cc @justinj not sure how this interacts with what you were doing.

knz:
cc @RaduBerinde I would suggest until this issue is fixed that we think at least about initializing NodeID properly in the evalCtx of distsql flows. Where should I look at to do this reliably?

justinj:
Agreed, I've run into this (having to deal with incomplete `EvalContext`s) in a bunch of places working on what I am now.

RaduBerinde:
@knz we don't currently allow context-dependent functions in distsql (they run non-distributed).
That said, I think [this](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/distsqlrun/server.go#L164) is the appropriate place to set NodeID and probably a bunch of other stuff.

arjunravinarayan:
@RaduBerinde but we plan to (see #13821, which is marked as 1.0)

andreimatei:
This has gotten better with #15628, but a ctor would still be nice.

arjunravinarayan:
@knz with #15628 in, I believe this milestone should be changed to 1.1 from 1.0 - is that correct?

knz:
1.0.1 actually...

petermattis:
Let's change the milestone to 1.1 and create a 1.0.1 label.

jordanlewis:
@knz to close and re-open with a more targeted action.

knz:
Subsumed by #15667.

Issue #15667, opened 2017-05-03
sql: we need a story about what EvalContext different operations need to be using


andreimatei:
The [`EvalContext`](https://github.com/cockroachdb/cockroach/blob/fce9109cc3e3f3c564de7334064f983b27b8f024/pkg/sql/parser/eval.go#L1687) is a bag of mostly session-dependent variables, to be used by various built-in functions. It contains stuff like a `time.Location` (i.e. timezone) used for string-timestamp casts and more, a `SearchPath` used to resolve unqualified identifiers, the statement and transaction timestamp to be reported by functions such as `now()`, etc.

- Generally, a query uses its session's `EvalContext`, which has been possibly populated by statements such as `SET TIME ZONE <x>`. 
- Most schema changes (on both the sync and async path) use the `EvalContext` of the session that ran the statement *for evaluating pure functions* in the new `DEFAULT` expressions. Everything that hasn't been evaluated by a constant by the upfront check added in #15345 is then evaluated during the backfill in a default `EvalContext` detached from any session. This context uses the UTC timestamp, a default `SearchPath`, and timestamps initialized each time the backfill is picked up by a new node (but consistent by all rows across all nodes processed during that backfill attempt).
- A `VALIDATE` schema change uses the session's ctx for all the checks (at the moment there's no upfront constant folding and there's no DistSQL involved that would require a more conscious choice when marshalling the `EvalCtx`).
- I'm not sure what the state is for foreign key checks.

I think we need to put some thought into what we want in every situation. At least the `VALIDATE` story is inconsistent with other schema change backfills.

Even for regular queries, it's not clear to me what we want in all cases - it seems questionable to me that, say, a cast from a string coming from the database into a timestamp uses the session's timezone (whereas an ALTER evaluating the same expressions would do something different and use UTC - which I suspect is better). Using the session's tz when operating on literals seems fine to me.

cc @dt @jordanlewis @jseldess @vivekmenezes

cuongdo:
how is this looking for 1.1?

andreimatei:
Nothing has been done here.

knz:
Note this is bug fix not new feature work so the time frame is slightly more flexible.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

cuongdo:
While the timing is slightly more flexible for this, let's try to get this done earlier if it involves significant code movement.

jordanlewis:
[...]

Issue #5583, opened 2016-03-25
sql: navicat compatibility


joel-solymosi:
To repro:
- Open navicat ( http://www.navicat.com/download/navicat-for-postgresql )
- Attempt to connect to a cockroachdb instance

Expected results:
- Displays cockroach server's databases

Observed results:
- ERROR: sql/select.go:354 (pmattis): unsupported FROM: FROM pg_database AS d LEFT JOIN pg_tablespace AS t ON d.dattablespace = t.oid

Adding support for database / table auto-discovery would allow the entire postgres toolchain / ecosystem to work against cockroach, which might significantly increase adoption.

petermattis:
@joel-custlabs CockroachDB != PostgreSQL. The proximate cause of the error you encountered is that CockroachDB currently doesn't support joins, though this is on the near-term roadmap to be fixed. More problematic is that CockroachDB does not support the postgres system tables (e.g. `pg_database` and `pg_tablespace`). There are no plans to attempt compatibility for these tables. I agree that such compatibility would ease/increase adoption, yet it isn't clear how much work is required there (the surface area appears huge).

digipigeon:
Hi, just wanted to know if there is any update on this, as I see that basic joins are now supported.

+1 on this feature from me.

jordanlewis:
We've also added support for `pg_database` to support common ORMs like Hibernate.

CockroachDB doesn't currently have a notion of tablespaces. We've implemented some dummy support for other `pg_*` tables, though, so it doesn't seem out of the question that we could do the same thing for `pg_tablespace`.

Can you determine whether a trivial implementation of `pg_tablespace` would be sufficient for these introspection queries? That would be helpful for us to figure out the prioritization and effort level of this issue.

digipigeon:
Currently the query which is failing is as follows


```
[2016-11-22 21:05:46.206] [000000] [Cockroach] [PGSQL]
SELECT d.datname, d.oid, pg_get_userbyid(d.datdba) AS owner, shobj_description(d.oid, 'pg_database') AS comment, t.spcname, d.datacl, d.datlastsysoid, d.encoding, pg_encoding_to_char(d.encoding) AS encodingname FROM pg_database d LEFT JOIN pg_tablespace t ON d.dattablespace=t.oid

[2016-11-22 21:05:46.240] [000000] [Cockroach] [PGSQL]
ERROR:  table "pg_tablespace" does not exist
```

[...]

PR #18588, opened 2017-09-19
sql: support casting strings to arrays

### LINKED ISSUE #18419: Feature request: cast strings to arrays
PostgreSQL can cast strings to arrays:

```
postgres=# SELECT '{}'::int[];
 int4 
------
 {}
(1 row)

postgres=# SELECT '{}'::oid[];
 oid 
-----
 {}
(1 row)
```

while CockroachDB fails with `invalid cast`:

```
root@:26257/> SELECT '{}'::int[];
pq: invalid cast: string -> INT[]
root@:26257/> SELECT '{}'::oid[];
pq: invalid cast: string -> OID[]
```

This came up while attempting to rewrite the [postgrex bootstrap query](https://github.com/elixir-ecto/postgrex/blob/f31048a8e9b21984a9e7b68eae069bd12a3691c7/lib/postgrex/types.ex#L73-L82) to not use a [correlated subquery](https://github.com/cockroachdb/cockroach/issues/3288); i.e. from

```sql
SELECT t.oid, t.typname, t.typsend, t.typreceive, t.typoutput, t.typinput,
       t.typelem, coalesce(r.rngsubtype, 0), ARRAY (
  SELECT a.atttypid
  FROM pg_attribute AS a
  WHERE a.attrelid = t.typrelid AND a.attnum > 0 AND NOT a.attisdropped
  ORDER BY a.attnum
)
FROM pg_type AS t
LEFT JOIN pg_range AS r ON r.rngtypid = t.oid;
```

to

```sql
SELECT t.oid, t.typname, t.typsend, t.typreceive, t.typoutput, t.typinput,
       t.typelem, coalesce(r.rngsubtype, 0), coalesce(atttypids, '{}'::oid[])
FROM pg_type AS t
LEFT JOIN pg_range AS r ON r.rngtypid = t.oid
LEFT JOIN (
  SELECT attrelid, array_agg(atttypid) AS atttypids
  FROM pg_attribute
  WHERE attnum > 0 AND NOT attisdropped
  GROUP BY 1
) a ON a.attrelid = t.typrelid;
```

That second `coalesce` fails because it can't cast `'{}'`.

As a workaround, I found the PostgreSQL-incompatible `array[]:::oid[];` (I am totally bewildered by the triple colon) and `ANNOTATE_TYPE(array[], oid[])`, also incompatible.

justinj:
Fixes #18419.

Support expressions like '{1,2,3}'::INT[], or '{abc,"xyz"}'::STRING[].

More testing, incl. fuzzing might be necessary.
I'm also iffy on the current ad-hoc construction of an EvalContext for
the purposes of having a location.

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -450,7 +450,7 @@ func convertRecord(
 	}
 
 	parse := parser.Parser{}
-	evalCtx := parser.EvalContext{}
+	evalCtx := parser.EvalContext{Location: &time.UTC}
 	// Although we don't yet support DEFAULT expressions on visible columns,
 	// we do on hidden columns (which is only the default _rowid one). This
 	// allows those expressions to run.
@@ -470,7 +470,7 @@ func convertRecord(
 				if nullif != nil && v == *nullif {
 					datums[i] = parser.DNull
 				} else {
-					datums[i], err = parser.ParseStringAs(col.Type.ToDatumType(), v, time.UTC)
+					datums[i], err = parser.ParseStringAs(col.Type.ToDatumType(), v, &evalCtx)
 					if err != nil {
 						return errors.Wrapf(err, "%s: row %d: parse %q as %s", batch.file, rowNum, col.Name, col.Type.SQLString())
 					}
--- a/pkg/sql/copy.go
+++ b/pkg/sql/copy.go
@@ -195,7 +195,8 @@ func (n *copyNode) addRow(ctx context.Context, line []byte) error {
 				return err
 			}
 		}
-		d, err := parser.ParseStringAs(n.resultColumns[i].Typ, s, n.session.Location)
+		evalCtx := n.session.evalCtx()
+		d, err := parser.ParseStringAs(n.resultColumns[i].Typ, s, &evalCtx)
 		if err != nil {
 			return err
 		}
--- a/pkg/sql/logictest/testdata/logic_test/array
+++ b/pkg/sql/logictest/testdata/logic_test/array
@@ -91,6 +91,33 @@ SELECT ARRAY[]:::int[]
 ----
 {}
 
+# casting strings to arrays
+
+query T
+SELECT '{1,2,3}'::INT[]
+----
+{1,2,3}
+
+query T
+SELECT '{hello,"hello"}'::STRING[]
+----
+{"hello","hello"}
+
+query T
+SELECT e'{he\\\\llo}'::STRING[]
+----
+{"he\\llo"}
+
+query T
+SELECT '{"abc\nxyz"}'::STRING[]
+----
+{"abcnxyz"}
+
+query T
+SELECT '{hello}'::VARCHAR(2)[]
+----
+{"he"}
+
 # array subscript access
 
 query T
--- a/pkg/sql/parser/eval.go
+++ b/pkg/sql/parser/eval.go
@@ -2232,8 +2232,11 @@ func (expr *CastExpr) Eval(ctx *EvalContext) (Datum, error) {
 		return d, nil
 	}
 	d = UnwrapDatum(d)
+	return performCast(ctx, d, expr.Type)
+}
 
-	switch typ := expr.Type.(type) {
+func performCast(ctx *EvalContext, d Datum, t CastTargetType) (Datum, error) {
+	switch typ := t.(type) {
 	case *BoolColType:
 		switch v := d.(type) {
 		case *DBool:
@@ -2426,7 +2429,7 @@ func (expr *CastExpr) Eval(ctx *EvalContext) (Datum, error) {
 		case *DOid:
 			s = t.name
 		}
-		switch c := expr.Type.(type) {
+		switch c := t.(type) {
 		case *StringColType:
 			// If the CHAR type specifies a limit we truncate to that limit:
 			//   'hello'::CHAR(2) -> 'he'
@@ -2545,6 +2548,10 @@ func (expr *CastExpr) Eval(ctx *EvalContext) (Datum, error) {
 		case *DInterval:
 			return d, nil
 		}
+	case *ArrayColType:
+		if s, ok := d.(*DString); ok {
+			return ParseDArrayFromString(ctx, string(*s), typ.ParamType)
+		}
 	case *OidColType:
 		switch v := d.(type) {
 		case *DOid:
@@ -2651,7 +2658,7 @@ func (expr *CastExpr) Eval(ctx *EvalContext) (Datum, error) {
 	}
 
 	return nil, pgerror.NewErrorf(
-		pgerror.CodeCannotCoerceError, "invalid cast: %s -> %s", d.ResolvedType(), expr.Type)
+		pgerror.CodeCannotCoerceError, "invalid cast: %s -> %s", d.ResolvedType(), t)
 }
 
 // Eval implements the TypedExpr interface.
--- a/pkg/sql/parser/expr.go
+++ b/pkg/sql/parser/expr.go
@@ -1086,6 +1086,7 @@ var (
 	oidCastTypes       = []Type{TypeNull, TypeString, TypeCollatedString, TypeInt, TypeOid}
 	uuidCastTypes      = []Type{TypeNull, TypeString, TypeCollatedString, TypeBytes, TypeUUID}
 	inetCastTypes      = []Type{TypeNull, TypeString, TypeCollatedString, TypeINet}
+	arrayCastTypes     = []Type{TypeNull, TypeString}
 )
 
 // validCastTypes returns a set of types that can be cast into the provided type.
@@ -1121,7 +1122,7 @@ func validCastTypes(t Type) []Type {
 		if t.FamilyEqual(TypeCollatedString) {
 			return stringCastTypes
 		} else if t.FamilyEqual(TypeArray) {
-			return []Type{TypeNull}
+			return arrayCastTypes
 		}
 		return nil
 	}
--- a/pkg/sql/parser/parse.go
+++ b/pkg/sql/parser/parse.go
@@ -210,7 +210,7 @@ func ParseType(sql string) (CastTargetType, error) {
 }
 
 // ParseStringAs parses s as type t.
-func ParseStringAs(t Type, s string, location *time.Location) (Datum, error) {
+func ParseStringAs(t Type, s string, evalCtx *EvalContext) (Datum, error) {
 	var d Datum
 	var err error
 	switch t {
@@ -219,7 +219,7 @@ func ParseStringAs(t Type, s string, location *time.Location) (Datum, error) {
 	case TypeBytes:
 		d = NewDBytes(DBytes(s))
 	case TypeDate:
-		d, err = ParseDDate(s, location)
+		d, err = ParseDDate(s, evalCtx.GetLocation())
 	case TypeDecimal:
 		d, err = ParseDDecimal(s)
 	case TypeFloat:
@@ -233,13 +233,24 @@ func ParseStringAs(t Type, s string, location *time.Location) (Datum, error) {
 	case TypeTimestamp:
 		d, err = ParseDTimestamp(s, time.Microsecond)
 	case TypeTimestampTZ:
-		d, err = ParseDTimestampTZ(s, location, time.Microsecond)
+		d, err = ParseDTimestampTZ(s, evalCtx.GetLocation(), time.Microsecond)
 	case TypeUUID:
 		d, err = ParseDUuidFromString(s)
 	case TypeINet:
 		d, err = ParseDIPAddrFromINetString(s)
 	default:
-		return nil, pgerror.NewErrorf(pgerror.CodeInternalError, "unknown type %s", t)
+		if a, ok := t.(TArray); ok {
+			typ, err := DatumTypeToColumnType(a.Typ)
+			if err != nil {
+				return nil, err
+			}
+			d, err = ParseDArrayFromString(evalCtx, s, typ)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, pgerror.NewErrorf(pgerror.CodeInternalError, "unknown type %s", t)
+		}
 	}
 	return d, err
 }
--- a/pkg/sql/parser/parse_array.go
+++ b/pkg/sql/parser/parse_array.go
@@ -1,0 +1,189 @@
+// Copyright 2016 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package parser
+
+import (
+	"bytes"
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
+)
+
+var enclosingError = pgerror.NewErrorf(pgerror.CodeInvalidTextRepresentationError, "array must be enclosed in { and }")
+var extraTextError = pgerror.NewErrorf(pgerror.CodeInvalidTextRepresentationError, "extra text after closing right brace")
+var nestedArraysNotSupportedError = pgerror.NewErrorf(pgerror.CodeFeatureNotSupportedError, "nested arrays not supported")
+var malformedError = pgerror.NewErrorf(pgerror.CodeInvalidTextRepresentationError, "malformed array")
+
+var isQuoteChar = func(ch byte) bool {
+	return ch == '"'
+}
+
+var isControlChar = func(ch byte) bool {
+	return ch == '{' || ch == '}' || ch == ',' || ch == '"'
+}
+
+var isElementChar = func(r rune) bool {
+	return r != '{' && r != '}' && r != ','
+}
+
+// gobbleString advances the parser for the remainder of the current string
+// until it sees a non-escaped termination character, as specified by
+// isTerminatingChar, returning the resulting string, not including the
+// termination character.
+func (p *parseState) gobbleString(isTerminatingChar func(ch byte) bool) (out string, err error) {
+	var result bytes.Buffer
+	start := 0
+	i := 0
+	for i < len(p.s) && !isTerminatingChar(p.s[i]) {
+		i++
+		// In these strings, we just encode directly the character following a
+		// '\', even if it would normally be an escape sequence.
+		if i < len(p.s) && p.s[i] == '\\' {
+			result.WriteString(p.s[start:i])
+			i++
+			if i < len(p.s) {
+				result.WriteByte(p.s[i])
+				i++
+			}
+			start = i
+		}
+	}
+	if i >= len(p.s) {
+		return "", malformedError
+	}
+	result.WriteString(p.s[start:i])
+	p.s = p.s[i:]
+	return result.String(), nil
+}
+
+type parseState struct {
+	s       string
+	evalCtx *EvalContext
+	result  *DArray
+	t       ColumnType
+}
+
+func (p *parseState) advance() {
+	_, l := utf8.DecodeRuneInString(p.s)
+	p.s = p.s[l:]
+}
+
+func (p *parseState) eatWhitespace() {
+	for unicode.IsSpace(p.peek()) {
+		p.advance()
+	}
+}
+
+func (p *parseState) peek() rune {
+	r, _ := utf8.DecodeRuneInString(p.s)
+	return r
+}
+
+func (p *parseState) eof() bool {
+	return len(p.s) == 0
+}
+
+func (p *parseState) parseQuotedString() (string, error) {
+	return p.gobbleString(isQuoteChar)
+}
+
+func (p *parseState) parseUnquotedString() (string, error) {
+	out, err := p.gobbleString(isControlChar)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(out), nil
+}
+
+func (p *parseState) parseElement() error {
+	var next string
+	var err error
+	r := p.peek()
+	switch r {
+	case '{':
+		return nestedArraysNotSupportedError
+	case '"':
+		p.advance()
+		next, err = p.parseQuotedString()
+		if err != nil {
+			return err
+		}
+		p.advance()
+	default:
+		if !isElementChar(r) {
+			return malformedError
+		}
+		next, err = p.parseUnquotedString()
+		if err != nil {
+			return err
+		}
+		if strings.EqualFold(next, "null") {
+			return p.result.Append(DNull)
+		}
+	}
+
+	d, err := performCast(p.evalCtx, NewDString(next), p.t)
+	if err != nil {
+		return err
+	}
+	return p.result.Append(d)
+}
+
+// ParseDArrayFromString parses the string-form of constructing arrays, handling
+// cases such as `'{1,2,3}'::INT[]`.
+func ParseDArrayFromString(evalCtx *EvalContext, s string, t ColumnType) (*DArray, error) {
+	parser := parseState{
+		s:       s,
+		evalCtx: evalCtx,
+		result:  NewDArray(CastTargetToDatumType(t)),
+		t:       t,
+	}
+
+	parser.eatWhitespace()
+	if parser.peek() != '{' {
+		return nil, enclosingError
+	}
+	parser.advance()
+	parser.eatWhitespace()
+	if parser.peek() != '}' {
+		if err := parser.parseElement(); err != nil {
+			return nil, err
+		}
+		parser.eatWhitespace()
+		for parser.peek() == ',' {
+			parser.advance()
+			parser.eatWhitespace()
+			if err := parser.parseElement(); err != nil {
+				return nil, err
+			}
+		}
+	}
+	parser.eatWhitespace()
+	if parser.eof() {
+		return nil, enclosingError
+	}
+	if parser.peek() != '}' {
+		return nil, malformedError
+	}
+	parser.advance()
+	parser.eatWhitespace()
+	if !parser.eof() {
+		return nil, extraTextError
+	}
+
+	return parser.result, nil
+}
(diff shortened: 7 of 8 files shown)


jordanlewis:
Updates #5583

richardwu:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/csv.go, line 447 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPe8YCiS_qQNUjB2FE:-KuPe8YCiS_qQNUjB2FF:b6ku4jn) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/ccl/sqlccl/csv.go#L447)):*
> ```Go
> 				datums[i] = parser.DNull
> 			} else {
> 				datums[i], err = parser.ParseStringAs(visibleCols[i].Type.ToDatumType(), r, &parser.EvalContext{Location: &time.UTC})
> ```

There's been an open issue on this ad-hoc-ness of `EvalContext` since `1.0` https://github.com/cockroachdb/cockroach/issues/15363. Perhaps it be best to consider introducing default constructor(s) `EvalContext`s at this point (I'm not too sure how `EvalContext` is plumbed throughout and what would be considered "default")?

---

*[pkg/sql/parser/datum.go, line 259 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPgBuwESsoHZo5HsF7:-KuPgBuwESsoHZo5HsF8:b-ewwn2n) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/sql/parser/datum.go#L259)):*
> ```Go
> 
> const (
> 	eofToken = iota
> ```

You should include your type alias on the first line here i.e. `eofToken arrayTokenType = iota`.

---

*[pkg/sql/parser/datum.go, line 283 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPnOSBDPXCeV8ZfODx:-KuPnOSBDPXCeV8ZfODy:b-a6auwa) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/sql/parser/datum.go#L283)):*
> ```Go
> 	result := bytes.NewBuffer(nil)
> 	start := i
> 	for i < len(in) && !isTerminatingChar(in[i]) {
> ```

I think you'll run into issues with some unicode characters here (and everywhere else where `byte` is used) since you're iterating over the bytes (`uint8`) of the string as opposed to the runes (semantically characters, `uint32`) of the string.

Here's a demo of how strings behave in golang: https://play.golang.org/p/LgKXp85RnI

So if someone happens to have a unicode character whose individual byte(s) correspond to any of the special tokens, then you'll end up mis-parsing the otherwise valid literal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

justinj:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/sql/parser/datum.go, line 259 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPgBuwESsoHZo5HsF7:-KuPqAKZjUoxGKBbGeCa:b-oz8tyz) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/sql/parser/datum.go#L259)):*

You should include your type alias on the first line here i.e. `eofToken arrayTokenType = iota`.
</blockquote></details>

Good call!

---

*[pkg/sql/parser/datum.go, line 283 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPnOSBDPXCeV8ZfODx:-KuPqDRr0Gg0ZxvrxLrB:b-w19gei) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/sql/parser/datum.go#L283)):*

I think you'll run into issues with some unicode characters here (and everywhere else where `byte` is used) since you're iterating over the bytes (`uint8`) of the string as opposed to the runes (semantically characters, `uint32`) of the string.

Here's a demo of how strings behave in golang: https://play.golang.org/p/LgKXp85RnI

So if someone happens to have a unicode character whose individual byte(s) correspond to any of the special tokens, then you'll end up mis-parsing the otherwise valid literal.
</blockquote></details>

I actually investigated this a bit, I don't think there's a problem, since UTF-8 is designed to avoid this - all the bytes in multibyte UTF-8 characters have high-bit `1` (which no single-byte character has) so none of them will masquerade as one of the special tokens.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

justinj:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/csv.go, line 447 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPe8YCiS_qQNUjB2FE:-KuPsRiYBwDkAXwy3_T7:bf7v804) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/ccl/sqlccl/csv.go#L447)):*

There's been an open issue on this ad-hoc-ness of `EvalContext` since `1.0` https://github.com/cockroachdb/cockroach/issues/15363. Perhaps it be best to consider introducing default constructor(s) `EvalContext`s at this point (I'm not too sure how `EvalContext` is plumbed throughout and what would be considered "default")?
</blockquote></details>

Yeah I'm not sure what to do here, I think the *correct* thing would probably be to plumb an `EvalContext` from up high all the way down here, but I think that might be a really wide-reaching change :/ I'll bite the bullet and see what would be involved, I guess. I don't really want to introduce a constructor, at least in this PR, since it's sort of a band-aid for the actual problem of not having the correct `EvalContext` for this situation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

knz:
Updates  #15667.

knz:
Reviewed 5 of 8 files at r1, 3 of 3 files at r2.

---

*[pkg/ccl/sqlccl/csv.go, line 447 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPe8YCiS_qQNUjB2FE:-KuPx3Uevgk92JCiw9YY:b-25z36y) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/ccl/sqlccl/csv.go#L447)):*

Yeah I'm not sure what to do here, I think the *correct* thing would probably be to plumb an `EvalContext` from up high all the way down here, but I think that might be a really wide-reaching change :/ I'll bite the bullet and see what would be involved, I guess. I don't really want to introduce a constructor, at least in this PR, since it's sort of a band-aid for the actual problem of not having the correct `EvalContext` for this situation.
</blockquote></details>

I think it would be a good investment of your time to prepare another PR to plumb the EvalContext (or create a constructor to make one using the appropriate information), that fixes #15363,  to serve as base commit on which you can rebase this one. And let this one open in the meantime. I don't think that kicking the can down the road is helping here.

---

*[pkg/sql/copy.go, line 198 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPxc3q_6V1o-KzRXVH:-KuPxc3q_6V1o-KzRXVI:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/copy.go#L198)):*
> ```Go
> 			}
> 		}
> 		d, err := parser.ParseStringAs(n.resultColumns[i].Typ, s, &parser.EvalContext{Location: &n.session.Location})
> ```

ditto

---

*[pkg/sql/parser/datum.go, line 251 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPxphmROw6aug19gD9:-KuPxphmROw6aug19gDA:b-rdumzl) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L251)):*
> ```Go
> }
> 
> var enclosingError = pgerror.NewErrorf(pgerror.CodeInvalidTextRepresentationError, "array must be enclosed in { and }")
> ```

Please split all this new code in a new file.

---

*[pkg/sql/parser/datum.go, line 422 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPy1okl_4E9Ndrf00B:-KuPy1olfd0rzF6-arBW:bmbm8ak) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L422)):*
> ```Go
> // ParseDArrayFromString parses the string-form of constructing arrays, handling
> // cases such as `'{1,2,3}'::INT[]`. The pattern here is adapted from
> // https://talks.golang.org/2011/lex.slide.
> ```

The code would be simpler to follow if you used a straight LL1 approach instead. (aka "recursive descent parsing")

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

richardwu:
Reviewed 8 of 8 files at r1.

---

*[pkg/sql/parser/datum.go, line 283 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPnOSBDPXCeV8ZfODx:-KuPrnRbwMF5Bol27jDf:bqbnv9g) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/sql/parser/datum.go#L283)):*

I actually investigated this a bit, I don't think there's a problem, since UTF-8 is designed to avoid this - all the bytes in multibyte UTF-8 characters have high-bit `1` (which no single-byte character has) so none of them will masquerade as one of the special tokens.
</blockquote></details>

Ah right, good point! Nevermind what I said then :)

---

*[pkg/sql/parser/datum.go, line 285 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPspqo_H-vTN-h-GUV:-KuPspqo_H-vTN-h-GUW:biohrlw) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L285)):*
> ```Go
> 	for i < len(in) && !isTerminatingChar(in[i]) {
> 		i++
> 		if i < len(in) && in[i] == '\\' {
> ```

I'm assuming this is for some special escaped characters? Perhaps add a comment since it might be non-trivial for someone like me :)

---

*[pkg/sql/parser/datum.go, line 302 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPteqJA7zNWTLVc_NE:-KuPteqJA7zNWTLVc_NF:b16ssck) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L302)):*
> ```Go
> 	i := 0
> 	return func() arrayToken {
> 		for i < len(s) && unicode.IsSpace(rune(s[i])) {
> ```

Looks like you might need to `range s` here since `unicode.IsSpace` may return true for 2-byte `NEL` and `NBSP` characters https://golang.org/pkg/unicode/#IsSpace.

---

*[pkg/sql/parser/datum.go, line 380 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPv3ZgACQfIJBwuCyf:-KuPv3ZgACQfIJBwuCyg:b-vvfr84) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L380)):*
> ```Go
> func elementState(ctx parsingContext, nextToken func() arrayToken) (parseStateFn, error) {
> 	next := nextToken()
> 	switch next.t {
> ```

Would you need to handle 
```
case openCurToken: 
    return nil, nestedArraysNotSupportedError
```
here?

Since you might want to return the nested array error with a string like `{42, {42}, 42}`.

---

*[pkg/sql/parser/datum_test.go, line 449 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPwqfSCsP4jJ3acsVX:-KuPwqfSCsP4jJ3acsVY:bi30f0j) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum_test.go#L449)):*
> ```Go
> }
> 
> func TestParseArray(t *testing.T) {
> ```

Good test coverage! Perhaps add a test with the special 2-byte wide space characters? Not sure what the best way to handle it is, but we can either disallow them or parse them as valid runes.

---

*[pkg/sql/parser/datum_test.go, line 483 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPyKkwDMaIV0wU-8wA:-KuPyKkwDMaIV0wU-8wB:b-8dw0n1) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum_test.go#L483)):*
> \\

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

andreimatei:

---

*[pkg/ccl/sqlccl/csv.go, line 447 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPe8YCiS_qQNUjB2FE:-KuSjKxnEqB6EWWo4PvR:bwqb8jl) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/ccl/sqlccl/csv.go#L447)):*

I think it would be a good investment of your time to prepare another PR to plumb the EvalContext (or create a constructor to make one using the appropriate information), that fixes #15363,  to serve as base commit on which you can rebase this one. And let this one open in the meantime. I don't think that kicking the can down the road is helping here.
</blockquote></details>

why is it hard to plumb the EvalCtx?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

justinj:
PTAL!

---

Review status: 3 of 8 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/csv.go, line 447 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPe8YCiS_qQNUjB2FE:-KuSo5JTKBkp8pmV3Ocu:b-ma7cqr) ([raw file](https://github.com/cockroachdb/cockroach/blob/58000fb0bb34e2638069c433be74ab3777e61780/pkg/ccl/sqlccl/csv.go#L447)):*

why is it hard to plumb the EvalCtx?
</blockquote></details>

upon further investigation, it actually wasn't! both cases had an evalCtx available (although one was just as ad-hoc as mine).

---

*[pkg/sql/copy.go, line 198 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPxc3q_6V1o-KzRXVH:-KuSoPey1rvkjldw3b_P:bp5h96m) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/copy.go#L198)):*

ditto
</blockquote></details>

I was wrong, there's an evalCtx accessible here.

---

*[pkg/sql/parser/datum.go, line 251 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPxphmROw6aug19gD9:-KuQ5sQPaRbOeHErTQzI:b-oz8tym) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L251)):*

Please split all this new code in a new file.
</blockquote></details>

Good call.

---

*[pkg/sql/parser/datum.go, line 285 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPspqo_H-vTN-h-GUV:-KuSnDIPDBT2NvTqQ9yw:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L285)):*

I'm assuming this is for some special escaped characters? Perhaps add a comment since it might be non-trivial for someone like me :)
</blockquote></details>

Done.

---

*[pkg/sql/parser/datum.go, line 302 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPteqJA7zNWTLVc_NE:-KuSmYJhATrKlP2gGft2:bf57jr2) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L302)):*

Looks like you might need to `range s` here since `unicode.IsSpace` may return true for 2-byte `NEL` and `NBSP` characters https://golang.org/pkg/unicode/#IsSpace.
</blockquote></details>

Handles whitespace runes properly now!

---

*[pkg/sql/parser/datum.go, line 380 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPv3ZgACQfIJBwuCyf:-KuSmVYVUckpfNxRD99z:bnosds8) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L380)):*

Would you need to handle 
```
case openCurToken: 
    return nil, nestedArraysNotSupportedError
```
here?

Since you might want to return the nested array error with a string like `{42, {42}, 42}`.
</blockquote></details>

This is handled now!

---

*[pkg/sql/parser/datum.go, line 422 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPy1okl_4E9Ndrf00B:-KuSknLKv7hAa3Tk4gJF:bjt72q3) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L422)):*

The code would be simpler to follow if you used a straight LL1 approach instead. (aka "recursive descent parsing")
</blockquote></details>

Ok, I rewrote it, I'm not 100% convinced I like it better (I liked the explicitness of the state machine in the previous one) but it is shorter!

---

*[pkg/sql/parser/datum_test.go, line 449 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPwqfSCsP4jJ3acsVX:-KuSmHfzEGaraoHd_RlD:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum_test.go#L449)):*

Good test coverage! Perhaps add a test with the special 2-byte wide space characters? Not sure what the best way to handle it is, but we can either disallow them or parse them as valid runes.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-:-KuUlVCKGaZHcc0RPjLF:b-83330z)*

justinj:
FWIW I ran this through go-fuzz (using the test cases as the initial corpus) and it didn't find any panics.

richardwu:
LGTM besides a few nits! :lgtm_strong: 

---

Review status: 3 of 8 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/parse.go, line 242 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-Kulnp79Q3SJAUIKQnpG:-Kulnp79Q3SJAUIKQnpH:b-pf8i4m) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse.go#L242)):*
> ```Go
> 		d, err = ParseDIPAddrFromINetString(s)
> 	default:
> 		if a, ok := t.(TArray); ok {
> ```

nit: couldn't this be a case statement?

---

*[pkg/sql/parser/parse_array.go, line 44 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuloIL31zD9A-_FgTda:-KuloIL4p3T9MjxeJMvk:b-d6mkww) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse_array.go#L44)):*
> ```Go
> 
> func (p *parseState) gobbleString(isTerminatingChar func(ch byte) bool) (out string, err error) {
> 	result := bytes.NewBuffer(nil)
> ```

nit: I think you can just do `var result bytes.Buffer` here. 

---

*[pkg/sql/parser/parse_array.go, line 77 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KulszNmz6vk2Gcnb8xk:-KulszNmz6vk2Gcnb8xl:b-8fx2u7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse_array.go#L77)):*
> ```Go
> 
> func (p *parseState) advance() {
> 	_, l := utf8.DecodeRuneInString(p.s)
> ```

:+1: 

---

*[pkg/sql/parser/parse_array.go, line 122 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KulqXIcLmMIdf_VQQ1x:-KulqXIcLmMIdf_VQQ1y:bg5c45v) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse_array.go#L122)):*
> ```Go
> 		p.advance()
> 	default:
> 		if !isElementChar(p.peek()) {
> ```

nit: you could assign the `p.peek()` in the `switch` clause and then re-use that here to save a function call. Unfortunately, I don't believe the compiler would be able to do constant propagation here with `p.peek()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-:-KultkMYVHdmf1TShbGe:b8sgn4s)*

justinj:
Review status: 3 of 8 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/parse.go, line 242 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-Kulnp79Q3SJAUIKQnpG:-KunwfoeEyoOhbINjCvG:b92mqh6) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse.go#L242)):*

nit: couldn't this be a case statement?
</blockquote></details>

No, because the *value* of the `TArray` differs depending on its contents, so we have to do a type assertion.

---

*[pkg/sql/parser/parse_array.go, line 44 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuloIL31zD9A-_FgTda:-KunwwEU52CfXOJ4eEiO:b-drur0q) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse_array.go#L44)):*

nit: I think you can just do `var result bytes.Buffer` here. 
</blockquote></details>

:thumbsup: nice, didn't know it didn't need to be initialized

---

*[pkg/sql/parser/parse_array.go, line 122 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KulqXIcLmMIdf_VQQ1x:-Kunx8KKWNHEfMyht9XB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d9488ef7d1817f4974e8db0b3e067d939f7a09f2/pkg/sql/parser/parse_array.go#L122)):*

nit: you could assign the `p.peek()` in the `switch` clause and then re-use that here to save a function call. Unfortunately, I don't believe the compiler would be able to do constant propagation here with `p.peek()`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588)*

jordanlewis:
:lgtm:

---

Reviewed 2 of 8 files at r1, 1 of 3 files at r2, 3 of 7 files at r3, 2 of 2 files at r4.

---

*[pkg/sql/parser/datum.go, line 280 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPx_mzIcZ61Lh22NtI:-KuPx_mzIcZ61Lh22NtJ:b-lqy0v9) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L280)):*
> ```Go
> }
> 
> func gobbleString(in string, i int, isTerminatingChar func(ch byte) bool) (end int, out string) {
> ```

Please add a docstring to this function

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-:-KuqkaR1Vvxm_K97G3Rc:bnfp4nl)*

justinj:
@knz now that this plumbs `EvalContext`s (despite one of them still being ad-hoc), are you ok with merging it?

---

Review status: 7 of 8 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/datum.go, line 280 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuPx_mzIcZ61Lh22NtI:-Kutjgsvt58XMWFsYkj5:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2e06590667dec2ec1fccff996b9de3162377a557/pkg/sql/parser/datum.go#L280)):*

Please add a docstring to this function
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-:-KutjhcedThi1HZnjbmq:b-xkesgx)*

knz:
LGTM with a test nit.

---

Reviewed 5 of 7 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.

---

*[pkg/sql/parser/parse_array_test.go, line 98 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-KuuzYBguDrwLLBwzTXC:-KuuzYBguDrwLLBwzTXD:b-i4iwki) ([raw file](https://github.com/cockroachdb/cockroach/blob/fea01b2a1bca119b1f63d2ca54e707a14582d0ce/pkg/sql/parser/parse_array_test.go#L98)):*
> ```Go
> 		{"{\"hello}", stringColTypeString, `malformed array`},
> 		// It might be unnecessary to disallow this, but Postgres does.
> 		{"{he\"lo}", stringColTypeString, "malformed array"},
> ```

Also provide tests that contains incomplete UTF-8 sequences in final positions of different syntactic elements.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18588#-:-KuuzoRwICJwOF024adw:b-atqklr)*

justinj:
TFTRs!

PR #18603, opened 2017-09-19
server,ui: fix the display of the number of indexes per table

### LINKED ISSUE #13888: ui: table inspector reports incorrect number of indices
```
root@:26257/test> show create table t2;
+-------+---------------------------------------------+
| Table |                 CreateTable                 |
+-------+---------------------------------------------+
| t2    | CREATE TABLE t2 (␤                          |
|       |     id INT NOT NULL,␤                       |
|       |     name STRING(100) NULL,␤                 |
|       |     CONSTRAINT "primary" PRIMARY KEY (id),␤ |
|       |     INDEX idx_name (name),␤                 |
|       |     FAMILY "primary" (id, name)␤            |
|       | )                                           |
+-------+---------------------------------------------+
(1 row)
root@:26257/test> show indexes from t2;
+-------+----------+--------+-----+--------+-----------+---------+----------+
| Table |   Name   | Unique | Seq | Column | Direction | Storing | Implicit |
+-------+----------+--------+-----+--------+-----------+---------+----------+
| t2    | primary  | true   |   1 | id     | ASC       | false   | false    |
| t2    | idx_name | false  |   1 | name   | ASC       | false   | false    |
| t2    | idx_name | false  |   2 | id     | ASC       | false   | true     |
+-------+----------+--------+-----+--------+-----------+---------+----------+
(3 rows)
```

As you can see the test.t2 table has 2 indexes: primary and idx_name.

However the Console lists a value of 3 for the "# OF INDICES" column:
cockroach-console-bug

I've tested this with various tables and the count shown in the Console seems to be the number of index parts rather than a count of the unique indexes--so not taking into account multi-column indexes (and all secondary indexes seem to be multi-column indexes, including a copy of the PK). 

I'm not sure if you want to change this in TableDescriptor, IndexDescriptor, or just in the Console... but it's at least confusing for the end user in the Console.

knz:
Fixes #13888.

### DIFF
--- a/pkg/ui/src/views/databases/data/tableInfo.tsx
+++ b/pkg/ui/src/views/databases/data/tableInfo.tsx
@@ -1,4 +1,5 @@
 import * as protos from "src/js/protos";
+import _ from "lodash";
 
 type TableDetailsResponse = protos.cockroach.server.serverpb.TableDetailsResponse;
 type TableStatsResponse = protos.cockroach.server.serverpb.TableStatsResponse;
@@ -18,7 +19,7 @@ export class TableInfo {
       this.name = name;
       this.id = details && details.descriptor_id.toNumber();
       this.numColumns = details && details.columns.length;
-      this.numIndices = details && details.indexes.length;
+      this.numIndices = details && _.uniqBy(details.indexes, idx => idx.name).length;
       this.rangeCount = stats && stats.range_count && stats.range_count.toNumber();
       this.createStatement = details && details.create_table_statement;
       this.grants = details && details.grants;


mrtracy:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/ui/src/views/databases/data/tableInfo.tsx, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18603#-KuQifuMERrB_oO4OvY7:-KuQifuMERrB_oO4OvY8:b-qb2erw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d528516e0966b40b661f928d9c0882a48db7e5bc/pkg/ui/src/views/databases/data/tableInfo.tsx#L21)):*
> ```XML
>       this.id = details && details.descriptor_id.toNumber();
>       this.numColumns = details && details.columns.length;
>       this.numIndices = details && details.index_count.toNumber();
> ```

Instead of the backend changes, could this be `_.chain(details.indexes).map(idx => idx.name).uniq().size()`?  It looks like all the necessary information is coming back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18603)*

mrtracy:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/ui/src/views/databases/data/tableInfo.tsx, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18603#-KuQifuMERrB_oO4OvY7:-KuQjXzSZAWejw0rwKut:b506wl3) ([raw file](https://github.com/cockroachdb/cockroach/blob/d528516e0966b40b661f928d9c0882a48db7e5bc/pkg/ui/src/views/databases/data/tableInfo.tsx#L21)):*

Instead of the backend changes, could this be `_.chain(details.indexes).map(idx => idx.name).uniq().size()`?  It looks like all the necessary information is coming back.
</blockquote></details>

Or actually, just `_.uniqBy(details.indexes, idx => idx.name).length`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18603)*

knz:
Thanks for the tip. This gives me:

```
ERROR in /go/src/github.com/cockroachdb/cockroach/pkg/ui/src/views/databases/data/tableInfo.tsx
(21,36): '_' refers to a UMD global, but the current file is a module. Consider adding an import instead.
```

knz:
`import _ from "lodash";
` I suppose?

knz:
Done, PTAL

mrtracy:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18603#-:-KuUIPsMYG3FkvbGfUtT:bnfp4nl)*

PR #18605, opened 2017-09-19
sqlccl: batch items in/out of convertRecord channels

dt:
```
ConvertRecord-8    10.4µs ± 6%     5.0µs ± 6%   -51.82%        (p=0.000 n=10+10)

name             old speed      new speed      delta
ConvertRecord-8  11.6MB/s ± 6%  24.0MB/s ± 6%  +107.50%        (p=0.000 n=10+10)

ConvertRecord-8    1.83kB ± 0%    3.63kB ± 0%   +97.93%        (p=0.000 n=10+10)

ConvertRecord-8      93.0 ± 0%      93.0 ± 0%      ~     (all samples are equal)
```

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -162,7 +162,7 @@ func doLocalCSVTransform(
 	const chanSize = 10000
 
 	recordCh := make(chan csvRecord, chanSize)
-	kvCh := make(chan roachpb.KeyValue, chanSize)
+	kvCh := make(chan []roachpb.KeyValue, chanSize)
 	contentCh := make(chan sstContent)
 	var backupDesc *BackupDescriptor
 	conf, err := storageccl.ExportStorageConfFromURI(dest)
@@ -402,11 +402,13 @@ type csvRecord struct {
 func convertRecord(
 	ctx context.Context,
 	recordCh <-chan csvRecord,
-	kvCh chan<- roachpb.KeyValue,
+	kvCh chan<- []roachpb.KeyValue,
 	nullif *string,
 	tableDesc *sqlbase.TableDescriptor,
 ) error {
 	done := ctx.Done()
+
+	const kvBatchSize = 500
 
 	visibleCols := tableDesc.VisibleColumns()
 	keyDatums := make(sqlbase.EncDatumRow, len(tableDesc.PrimaryIndex.ColumnIDs))
@@ -438,7 +440,8 @@ func convertRecord(
 	}
 
 	datums := make([]parser.Datum, len(visibleCols))
-	var insertErr error
+	kvBatch := make([]roachpb.KeyValue, 0, kvBatchSize)
+
 	for record := range recordCh {
 		for i, r := range record.r {
 			if nullif != nil && r == *nullif {
@@ -458,19 +461,24 @@ func convertRecord(
 		if err != nil {
 			return errors.Wrapf(err, "generate insert row: %s: row %d", record.file, record.row)
 		}
-		insertErr = nil
 		if err := ri.InsertRow(ctx, inserter(func(kv roachpb.KeyValue) {
-			select {
-			case kvCh <- kv:
-			case <-done:
-				insertErr = ctx.Err()
-			}
+			kvBatch = append(kvBatch, kv)
 		}), row, true /* ignoreConflicts */, false /* traceKV */); err != nil {
 			return errors.Wrapf(err, "insert row: %s: row %d", record.file, record.row)
 		}
-		if insertErr != nil {
-			return insertErr
+		if len(kvBatch) > kvBatchSize {
+			select {
+			case kvCh <- kvBatch:
+			case <-done:
+				return ctx.Err()
+			}
+			kvBatch = make([]roachpb.KeyValue, 0, kvBatchSize)
 		}
+	}
+	select {
+	case kvCh <- kvBatch:
+	case <-done:
+		return ctx.Err()
 	}
 	return nil
 }
@@ -488,7 +496,7 @@ const errSSTCreationMaybeDuplicateTemplate = "SST creation error at %s; this can
 // and sent on contents. It returns the number of KV pairs created.
 func writeRocksDB(
 	ctx context.Context,
-	kvCh <-chan roachpb.KeyValue,
+	kvCh <-chan []roachpb.KeyValue,
 	tempEngine engine.Engine,
 	sstMaxSize int64,
 	contentCh chan<- sstContent,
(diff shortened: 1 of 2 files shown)


dt:
no top-line change in `BenchmarkImport`, but the based on these microbenchmark changes, @mjibson suggests this is still a win -- at least we'll burn a little less CPU while we wait on the IO that is making it slow.

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 443 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-KuQssi5sargLHi5D43N:-KuQssi5sargLHi5D43O:b-5i66vr) ([raw file](https://github.com/cockroachdb/cockroach/blob/4149c2054ab57dd91fda312d688c176fb28bab7e/pkg/ccl/sqlccl/csv.go#L443)):*
> ```Go
> 
> 	datums := make([]parser.Datum, len(visibleCols))
> 	kvBatch := make([]roachpb.KeyValue, 0, kvBatchSize)
> ```

Assume a table with either some column families or secondary indexes. This means each CSV row will produce N > 1 KVs. Unless kvBatchSize % N == 0, this will force the append below to allocate a new kvBatch and copy over the old one (assuming it can't just realloc in place), just for the last couple of KVs in the CSV row. What if we did `make(..., 0, kvBatchSize + 10)` or something here to attempt to prevent that in the normal case? Or something really smart like `(len(tableDesc.Families) + len(tableDesc.Indexes)) * 500`?

---

*[pkg/ccl/sqlccl/csv.go, line 478 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-KuQtY6ax3Z4vYfYKyFY:-KuQtY6ax3Z4vYfYKyFZ:bslgwey) ([raw file](https://github.com/cockroachdb/cockroach/blob/4149c2054ab57dd91fda312d688c176fb28bab7e/pkg/ccl/sqlccl/csv.go#L478)):*
> ```Go
> 		}
> 	}
> 	select {
> ```

Should this be in a `if len(kvBatch) > 0` block? I realize it doesn't matter downstream since it'll unpack the 0-length slice and do nothing with it, but it might be nice to keep the  invariant that convertRecord won't ever send a 0-length slice on kvCh. I'm also totally ok leaving this as is since it's simpler and just a tiny optimization.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 443 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-KuQssi5sargLHi5D43N:-KuQyNhbJHeB4gvhxtyZ:b8mr17j) ([raw file](https://github.com/cockroachdb/cockroach/blob/4149c2054ab57dd91fda312d688c176fb28bab7e/pkg/ccl/sqlccl/csv.go#L443)):*

Assume a table with either some column families or secondary indexes. This means each CSV row will produce N > 1 KVs. Unless kvBatchSize % N == 0, this will force the append below to allocate a new kvBatch and copy over the old one (assuming it can't just realloc in place), just for the last couple of KVs in the CSV row. What if we did `make(..., 0, kvBatchSize + 10)` or something here to attempt to prevent that in the normal case? Or something really smart like `(len(tableDesc.Families) + len(tableDesc.Indexes)) * 500`?
</blockquote></details>

the size check/send and is in the same loop as the append, so I think we'll flush before we need to realloc (though we want  `>=`)? I'm not sure we care about flushing only full rows.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 478 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-KuQtY6ax3Z4vYfYKyFY:-KuQzGahSwa9esa0bHU3:bte800e) ([raw file](https://github.com/cockroachdb/cockroach/blob/4149c2054ab57dd91fda312d688c176fb28bab7e/pkg/ccl/sqlccl/csv.go#L478)):*

Should this be in a `if len(kvBatch) > 0` block? I realize it doesn't matter downstream since it'll unpack the 0-length slice and do nothing with it, but it might be nice to keep the  invariant that convertRecord won't ever send a 0-length slice on kvCh. I'm also totally ok leaving this as is since it's simpler and just a tiny optimization.
</blockquote></details>

I thought about it and decided I didn't care -- it'll be rare that we send an empty one and it'll be cheap anyway as you note, so I opted for simplicity.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 443 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-KuQssi5sargLHi5D43N:-KuSiCPWj19FooZSEar8:bpwk0ty) ([raw file](https://github.com/cockroachdb/cockroach/blob/4149c2054ab57dd91fda312d688c176fb28bab7e/pkg/ccl/sqlccl/csv.go#L443)):*

the size check/send and is in the same loop as the append, so I think we'll flush before we need to realloc (though we want  `>=`)? I'm not sure we care about flushing only full rows.
</blockquote></details>

The append happens inside of ri.InsertRow, which is called for as many KVs as the row will produce. Assume a table with 2 secondary indexes. The inserter func in ri.InsertRow will get called 3 times per CSV row, so kvBatch will have len that is a multiple of 3. After 333 rows, it will have 999 KVs. On the next row it will insert the first kv for 1000, then on 1001 it will have to realloc, which happens before the kvBatchSize check. Right? If so I think we should do the work to avoid that realloc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 443 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-KuQssi5sargLHi5D43N:-KuVCF57WWRoQotks0WL:baozwcp) ([raw file](https://github.com/cockroachdb/cockroach/blob/4149c2054ab57dd91fda312d688c176fb28bab7e/pkg/ccl/sqlccl/csv.go#L443)):*

The append happens inside of ri.InsertRow, which is called for as many KVs as the row will produce. Assume a table with 2 secondary indexes. The inserter func in ri.InsertRow will get called 3 times per CSV row, so kvBatch will have len that is a multiple of 3. After 333 rows, it will have 999 KVs. On the next row it will insert the first kv for 1000, then on 1001 it will have to realloc, which happens before the kvBatchSize check. Right? If so I think we should do the work to avoid that realloc.
</blockquote></details>

ah, I see, nice catch. padded the capacity by 2x len(indexes)+len(families). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605)*

mjibson:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18605#-:-KuVYUIHc8T59CCNscEy:bnfp4nl)*

PR #18606, opened 2017-09-19
sql: Refresh table leases asynchronously on access

### LINKED ISSUE #17227: sql: refresh table leases asynchronously

lgo:
Some parts of this PR are WIP. In particular, there are two FIXMEs for parts that I have FUD that I'd like a second opinion on. I also have yet to modify the tests to reflect the new behavior.
 
The first is in `func (t *tableState) acquire`. I had to add a continual check to ensure that a valid lease exists even after we've finished acquiring one. This happens in the following case:
```
1: acquire ENTER
1: acquire: no valid lease and no request, making new lease request
1: acquireWait ENTER
2: acquire ENTER
2: acquire: no valid lease but request already made
2: acquireWait ENTER
1: acquireAwait LOOP
1: acquireWait EXIT
1: acquire: already has non-expiring lease
1: acquire DONE ts=1505848342.977638576,0
releasing lease
2: acquireAwait LOOP
2: acquireWait EXIT
2: acquire: no valid lease and no request, no valid lease and no request
2: acquireWait ENTER
...
```

What this has is that routine 1 synchronously goes to acquire the lease, then routine 2 waits on it. When routine 1 returns and releases its lock it immediately releases the lease before routine 2 can be run again, meaning that even though routine 2 finished busy-waiting it needs to re-acquire a lease.

The busy-wait checking alleviates this and should only happen in situations where a lease is immediately released.

The second is in `func (t *tableState) acquireWait` where we sleep the routine by a fixed amount of time. This check is required when we asynchronously go to renew a lease, and another routine enters `acquire` before that routine has begun. That's when the flag `startedAcquiringNextLease` is up and we busy-wait until it's down, where either the `acquiring` channel has now been initialized or it has already completed.

---

Previously if we would only refresh a table lease if it would be expired for the
transaction timestamp. Now we will refresh the lease asynchronously on access if
it is about to expire in order to prevent transactions from blocking on lease
acquisition.

Fixes #17227.

### DIFF
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -40,14 +40,23 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
 )
 
-// TODO(pmattis): Periodically renew leases for tables that were used recently and
-// for which the lease will expire soon.
-
 var (
 	// LeaseDuration is the mean duration a lease will be acquired for. The
-	// actual duration is jittered in the range
-	// [0.75,1.25]*LeaseDuration. Exported for testing purposes only.
+	// actual duration is jittered using LeaseJitterMultiplier.
+	// Exported for testing purposes only.
 	LeaseDuration = 5 * time.Minute
+	// LeaseJitterMultiplier is the factor that we use to randomly jitter the lease
+	// duration when acquiring a new lease. The range of the actual lease duration will
+	// be [(1-LeaseJitterMultiplier) * LeaseDuration, (1-LeaseJitterMultiplier) * LeaseDuration]
+	// Exported for testing purposes only.
+	LeaseJitterMultiplier = 0.25
+	// LeaseRenewalTimeout is the duration for when we renew the lease early.
+	// This is specifically 30 seconds less than the minimum LeaseDuration, to be
+	// sure not to let the lease expire before renewing. Exported for testing
+	// purposes.
+	// TODO(joey): We can make the timeout an offset from the actual expiration,
+	// after jitter.
+	LeaseRenewalTimeout = LeaseDuration - time.Duration(float64(LeaseDuration)*LeaseJitterMultiplier) - 30*time.Second
 )
 
 // tableVersionState holds the state for a table version. This includes
@@ -122,9 +131,9 @@ type LeaseStore struct {
 }
 
 // jitteredLeaseDuration returns a randomly jittered duration from the interval
-// [0.75 * leaseDuration, 1.25 * leaseDuration].
+// [(1-LeaseJitterMultiplier) * leaseDuration, (1+LeaseJitterMultiplier) * leaseDuration].
 func jitteredLeaseDuration() time.Duration {
-	return time.Duration(float64(LeaseDuration) * (0.75 + 0.5*rand.Float64()))
+	return time.Duration(float64(LeaseDuration) * (1 - LeaseJitterMultiplier + 2*LeaseJitterMultiplier*rand.Float64()))
 }
 
 // acquire a lease on the most recent version of a table descriptor.
@@ -562,6 +571,10 @@ type tableState struct {
 		// If set, leases are released from the store as soon as their
 		// refcount drops to 0, as opposed to waiting until they expire.
 		dropped bool
+		// Timer for refreshing the lease before it expires. Once a lease has been
+		// acquired once, a timer continuously runs to refresh the lease on a
+		// separate routine after LeaseRenewalTimeout time.
+		timer *time.Timer
 	}
 }
 
@@ -573,11 +586,12 @@ func (t *tableState) acquire(
 ) (*tableVersionState, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
-	// Wait for any existing lease acquisition.
+
+	// Block and wait for the new lease to be acquired on another routine.
 	t.acquireWait()
 
-	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
+	// Ensure a lease is acquired. Even though we finished waiting, another
+	// routine which finished acquiring a lease could have released it by now.
 	if s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp) {
 		if err := t.acquireNodeLease(ctx, m, hlc.Timestamp{}); err != nil {
 			return nil, err
@@ -585,6 +599,43 @@ func (t *tableState) acquire(
 	}
 
 	return t.findForTimestamp(ctx, timestamp, m)
+}
+
+// acquireAlways ensures a new lease is acquired. This is used for making sure a
+// new lease moves forward when refreshing the lease.
+func (t *tableState) acquireAlways(ctx context.Context, m *LeaseManager) error {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	// Block and wait for the new lease to be acquired on another routine to
+	// prevent a race condition of entering acquireNodeLease
+	t.acquireWait()
+
+	// Ensure a lease is acquired.
+	return t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+}
+
+// startRefreshingRoutine begins the goroutine responsible for refreshing a
+// table lease.
+// TODO(joey): Currently this routine will run forever for each table
+// descriptor. Ideally, we stop the routine if the table descriptor hasn't been
+// used recently.
+//
+// t.mu needs to be locked.
+func (t *tableState) startRefreshingRoutine(m *LeaseManager) {
+	ctx := context.TODO()
+	t.stopper.RunWorker(ctx, func(ctx context.Context) {
+		for {
+			select {
+			case <-t.mu.timer.C:
+				if err := t.acquireAlways(ctx, m); err != nil {
+					log.Error(ctx, err)
+				}
+			case <-t.stopper.ShouldStop():
+				return
+			}
+		}
+	})
 }
 
 // ensureVersion ensures that the latest version >= minVersion. It will
@@ -800,6 +851,17 @@ func (t *tableState) acquireNodeLease(
 	}
 	t.upsertLocked(ctx, table, m)
 	t.tableNameCache.insert(table)
+
+	// Either begin a new routine to start refreshing the lease asynchronously or
+	// or reset the timer for it. After the first lease acquisition, we should
+	// always have a routine actively renewing the lease. If a descriptor is
+	// dropped LeaseStore.acquire() will fail and the timer will not be set.
+	if t.mu.timer == nil {
+		t.mu.timer = time.NewTimer(LeaseRenewalTimeout)
+		t.startRefreshingRoutine(m)
+	} else {
+		t.mu.timer.Reset(LeaseRenewalTimeout)
+	}
 	return nil
 }
 
--- a/pkg/sql/lease_internal_test.go
+++ b/pkg/sql/lease_internal_test.go
@@ -19,16 +19,20 @@ package sql
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/config"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
+	"github.com/pkg/errors"
 )
 
 func TestTableSet(t *testing.T) {
@@ -557,4 +561,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
 	}
 
 	wg.Wait()
+}
+
+// This test makes sure the lease gets renewed automatically in the background.
+func TestLeaseRefreshedAutomatically(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+	var testAcquiredCount int32
+	testingKnobs := base.TestingKnobs{
+		SQLLeaseManager: &LeaseManagerTestingKnobs{
+			LeaseStoreTestingKnobs: LeaseStoreTestingKnobs{
+				// We want to track what leases get acquired,
+				LeaseAcquiredEvent: func(table sqlbase.TableDescriptor, _ error) {
+					if table.Name == "test" {
+						atomic.AddInt32(&testAcquiredCount, 1)
+					}
+				},
+			},
+		},
+	}
+	s, sqlDB, kvDB := serverutils.StartServer(t, base.TestServerArgs{Knobs: testingKnobs})
+	defer s.Stopper().Stop(context.TODO())
+	leaseManager := s.LeaseManager().(*LeaseManager)
+
+	if _, err := sqlDB.Exec(`
+CREATE DATABASE t;
+CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
+`); err != nil {
+		t.Fatal(err)
+	}
+	now := s.Clock().Now()
+	tableDesc := sqlbase.GetTableDescriptor(kvDB, "t", "test")
+
+	// We set LeaseRenewalTimeout to be low so it will refresh frequently, and the
+	// LeaseJitterMultiplier to ensure any newer leases will have a higher
+	// expiration timestamp.
+	savedLeaseRenewalTimeout := LeaseRenewalTimeout
+	savedLeaseJitterMultiplier := LeaseJitterMultiplier
+	defer func() {
+		LeaseRenewalTimeout = savedLeaseRenewalTimeout
+		LeaseJitterMultiplier = savedLeaseJitterMultiplier
+	}()
+	LeaseRenewalTimeout = 5 * time.Millisecond
+	LeaseJitterMultiplier = 0
+
+	// Acquire the first lease. This begins the refreshing routine.
+	_, e1, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Keep checking for a new lease to be acquired by the refreshing routine.
+	testutils.SucceedsSoon(t, func() error {
+		// Acquire the newer lease.
+		ts, e2, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		defer func() {
+			err := leaseManager.Release(ts)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}()
+
+		if e2.WallTime <= e1.WallTime {
+			return errors.Errorf("expected new lease expiration (%s) to be after old lease expiration (%s)",
+				e2, e1)
+		} else if count := atomic.LoadInt32(&testAcquiredCount); count < 3 {
+			return errors.Errorf("expected at least 3 leases to be acquired, but only acquired %d times",
+				count)
+		}
+		return nil
+	})
 }
--- a/pkg/sql/lease_test.go
+++ b/pkg/sql/lease_test.go
@@ -223,8 +223,10 @@ func TestLeaseManager(testingT *testing.T) {
 	// table and expiration.
 	l1, e1 := t.mustAcquire(1, descID)
 	l2, e2 := t.mustAcquire(1, descID)
-	if l1.ID != l2.ID || e1 != e2 {
+	if l1.ID != l2.ID {
 		t.Fatalf("expected same lease, but found %v != %v", l1, l2)
+	} else if e1 != e2 {
+		t.Fatalf("expected same lease timestamps, but found %v != %v", e1, e2)
 	}
 	t.expectLeases(descID, "/1/1")
 	// Node 2 never acquired a lease on descID, so we should expect an error.


lego:

---

*[pkg/sql/lease.go, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-KuQnM7-aW_nlEBQa1dw:-KuQnM7-aW_nlEBQa1dx:b-jxusgw) ([raw file](https://github.com/cockroachdb/cockroach/blob/45910e0514ec3b449cb1b3c8738a764d9f14d455/pkg/sql/lease.go#L49)):*
> ```Go
> 	LeaseDuration = 5 * time.Minute
> 	// LeaseRenewalDelta is the duration where we will renew the lease early.
> 	LeaseRenewalDelta = 30 * time.Second
> ```

I just picked an arbitrary value for this, I'm not sure if this is appropriate.

---

*[pkg/sql/lease.go, line 800 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-KuQnV7W9zoLisBFU3Km:-KuQnV7W9zoLisBFU3Kn:b-j55f6x) ([raw file](https://github.com/cockroachdb/cockroach/blob/45910e0514ec3b449cb1b3c8738a764d9f14d455/pkg/sql/lease.go#L800)):*
> ```Go
> 	for t.mu.startedAcquiringNextLease {
> 		t.mu.Unlock()
> 		time.Sleep(time.Nanosecond * 5)
> ```

Same as above, this is a fairly arbitrary constant which definitely adds to the FUD here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

lego:

---

*[pkg/sql/lease.go, line 596 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-KuUqLRLeDyU4jDd8Wqr:-KuUqLRMcR0lzB-y1KBf:bc1yays) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c767a3c7e8f943be12b13a3b2fbd0be33d5c784/pkg/sql/lease.go#L596)):*
> ```Go
> 	// continues. We need to continue to check that we have acquired a lease and
> 	// check while still inside t.mu.
> 	for {
> ```

I feel very FUD about this, and I've come to realize the current structure of this fails liveliness for the current routine. If another routine continually enters `acquire`, and immediately `releases the lease this could cause the current routine to always block waiting on a new lease and never acquire one itself.

Although it also occurred to me how the code previously worked successfully: it would `acquireWait` and then while under the lock and not releasing it, it would check if a lease exists and ensure a lease is acquired no matter what.

The code I wrote could be rewritten to drop the loop and if a routine does `acquireWait` it could check if the lease exists and if not just `acquireNodeLease` immediately, as opposed to looping, checking if one exists or has already been requested.

I don't know what the practical implications would be but I definitely feel more confident if we prevent the liveliness problem. Do you know what the best way to stress test this would be?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

vivekmenezes:

---

*[pkg/sql/lease.go, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-KuQnM7-aW_nlEBQa1dw:-KuUwa4yzN6ndrLAlfXM:bjdoyn1) ([raw file](https://github.com/cockroachdb/cockroach/blob/45910e0514ec3b449cb1b3c8738a764d9f14d455/pkg/sql/lease.go#L49)):*

I just picked an arbitrary value for this, I'm not sure if this is appropriate.
</blockquote></details>

I think you want to pick a value that is like say a minute less than LeaseDuration. We could make it dynamic unless we need some hardcoded value in a test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-KuQnM7-aW_nlEBQa1dw:-Ku_eOFClB-KR1I30rqE:b-20q7pd) ([raw file](https://github.com/cockroachdb/cockroach/blob/45910e0514ec3b449cb1b3c8738a764d9f14d455/pkg/sql/lease.go#L49)):*

I think you want to pick a value that is like say a minute less than LeaseDuration. We could make it dynamic unless we need some hardcoded value in a test
</blockquote></details>

Out of curiosity, is there a reason we'd want the renewal timeout to be closer to the duration than the lease?

To me, it seems like we would cycle leases faster, which would increase the time to make a schema change? (unless I misunderstand this).

Also when you say dynamic do you mean similar to how we jitter the actual lease duration?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

vivekmenezes:
I see that you are acquiring a lease asynchronously through the acquire() call itself. While this is okay I can see a situation where a table is not used around the time (expire - 30seconds) and thus the lease is not reacquired and then when the table is used closer to its expiration the lease is acquired asynchronously.

An alternative approach is to add a timer to the tableState that keep reacquiring the lease after a timeout. That feels like a more natural way to periodically reacquire the lease and doesn't suffer from the above problem.

Would you like to try out that approach?

lego:
Ah, gotcha. So once we've made a `tableState`, which I'm guessing happens the first time we access a table, we'll just want to ensure we always have a lease available then?

vivekmenezes:
Well I think what you want to do is create a regular timed goroutine that is acquiring the latest lease. This lease acquisition will run along the regular path but will *not check* the lease expiration time, it will however do stuff like check whether another lease is being acquired etc. So it's literally just another lease acquisition, it just always moves forward . In the case where it finds an existing lease being acquired it need not acquire a new lease.

lego:
I'm not quite sure I followed everything you mentioned, but I think the changes I just made encompass most of that. It uses a timed goroutine which will conditionally use expiry just to prevent an edge case for an excess acquisition.

One problem with this approach is as you approach a larger number of tables, once you access them that table will always add the overhead of another goroutine. A simple solution that I think would work is if you kill the goroutine once the table hasn't been accessed in X minutes (say 5?).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-:-KufZ17Qo8fKHOWTE2Z9:b-4fqors)*

pkg/sql/lease.go
```diff
@@ -573,20 +584,57 @@ func (t *tableState) acquire(
...
 	defer t.mu.Unlock()
-	// Wait for any existing lease acquisition.
-	t.acquireWait()
 
-	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
+	if t.mu.acquiring != nil {
+		// Block and wait for the new lease to be acquired on another routine.
+		t.acquireWait()
+	}
```
vivekmenezes: You do not need to make this change because t.acquireWait() already checks if t.mu.acquiring != nil

pkg/sql/lease.go
```diff
@@ -573,20 +584,57 @@ func (t *tableState) acquire(
...
+
 	return t.findForTimestamp(ctx, timestamp, m)
 }
 
+// startRefreshingRoutine begins the goroutine responsible for refreshing a
+// table lease.
+//
+// t.mu needs to be locked.
+func (t *tableState) startRefreshingRoutine(m *LeaseManager) {
```
vivekmenezes: pass in the context

lego:
I just addressed the issue I mentioned above, where we would end up always refreshing leases once they've been accessed. I added a check to see if the TableDescriptor has been acquired recently as a longer-term cutoff for the continual refreshing.

Again, the duration values are ones I tossed in, and I'm not too sure if they're completely appropriate.

pkg/sql/lease.go
```diff
@@ -573,20 +584,57 @@ func (t *tableState) acquire(
...
+	t.stopper.RunWorker(ctx, func(ctx context.Context) {
+		for {
+			select {
+			case <-t.mu.timer.C:
+				t.mu.Lock()
+				// Get a new lease only if the currently newest one will expire soon.
+				// This lease refresh will be skipped if another acquision happens and
+				// the timer fires before it that routine resets the timer. Then we
+				// don't want to refresh the lease immediately after.
+				if s := t.mu.active.findNewest(); s == nil || s.willExpire(hlc.Timestamp{WallTime: hlc.UnixNano()}) {
```
vivekmenezes: We really don't need to run these checks because it's just not possible for an existing lease acquisition to exist. We should just go ahead and acquire the lease (even if a recent acquisition happened) there is no harm in making another acquisition. You can get rid of the willExpire() method

pkg/sql/lease.go
```diff
@@ -573,20 +584,57 @@ func (t *tableState) acquire(
...
+					t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+				}
+				t.mu.Unlock()
```
vivekmenezes: I'll just create a new function t.acquireNodeLeaseByTimer() and lock defer.Unlock() within that

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +560,78 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+
+	// Wait for the lease to get acquired. First we need to sleep to ensure the
+	// timer routine begins acquisition.
+	// FIXME(joey): This would be a test race condition between here and the
+	// lease refresh code. Is there a better way to do this to prevent a flake?
+	time.Sleep(50 * time.Nanosecond)
+	for acquiring := ts.mu.acquiring; acquiring != nil; acquiring = ts.mu.acquiring {
+		<-acquiring
+	}
+
```
vivekmenezes: You can use testutils.SucceedsSoon() and keep acquiring/release a lease until its expiration time changes

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +560,78 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+	// Acquire the newer lease.
+	_, e2, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if e2.WallTime <= e1.WallTime {
```
vivekmenezes: test this within the SucceedsSoon()

vivekmenezes (commented):
This is looking really great. I'd like us to explore disabling the timer once there is no traffic directed to the descriptor. But I also prefer keep this change small so you can make a follow up PR to do that. The followup can do two things 1. adjust the timeout to be actual-expiration - 1 minute, 2. delete the timer once there is no traffic on the descriptor for a while. Thanks!

pkg/sql/lease.go
lego: 👍

pkg/sql/lease.go
lego: 👍

pkg/sql/lease.go
```diff
@@ -40,14 +40,15 @@ import (
...
-
 var (
 	// LeaseDuration is the mean duration a lease will be acquired for. The
 	// actual duration is jittered in the range
 	// [0.75,1.25]*LeaseDuration. Exported for testing purposes only.
 	LeaseDuration = 5 * time.Minute
+	// LeaseRenewalTimeout is the duration for when we renew the lease early.
+	// This is specifically less than LeaseDuration * 0.75 (~3.75m), to be sure
+	// not to let the lease expire before renewing. Exported for testing purposes.
+	LeaseRenewalTimeout = 2 * time.Minute
```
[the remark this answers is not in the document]
lego: 👍

pkg/sql/lease.go
lego: Yea, I just realized because this will definitely be running before a routine synchronously tries to acquire a new lease as part of `acquire` so the only situation where this would renew immediately after would be if `acquireFreshestFromStoreLocked` executes exactly when the timer fires.

I'll remove this.

pkg/sql/lease_internal_test.go
lego: (took this test out, as it was for stopping the refresh routine)

pkg/sql/lease_internal_test.go
lego: 👍 that is really cool!

lgo (commented):
I didn't end up adding a test checking that no goroutines blocked while a new lease was getting acquired. I really wasn't sure what the best way to approach that would be beside flipping a bool in `acquireWait` and checking that in the test. Is that sane, or are there more idiomatic ways to do that?

Also are there any other things that can/should be done to test this, such as running a cluster for an extended period of time? (or any standard testing procedures I might not know about yet :) )

pkg/sql/lease.go
```diff
@@ -562,6 +565,14 @@ type tableState struct {
...
 		// refcount drops to 0, as opposed to waiting until they expire.
 		dropped bool
+		// Timer for refreshing the lease before it expires. Once a lease has been
+		// acquired once, a timer continuously runs to refresh the lease on a
+		// separate routine after LeaseRenewalTimeout time.
+		timer *time.Timer
+		// The error that was thrown on the last acquision.
+		// FIXME(joey): Should the first caller return the last async error or
+		// instead try to synchronously get the lease?
+		lastAcquisitionErr error
```
lego: I'm still not too sure what to do with the scenario when we get an error during the async acquire. I think the best action, for now, is to ignore the error. (although how do I get around the linter failing because of an unhandled err? D:)

Alternatively, the async refreshing routine could retry but that does get very complicated quickly (I would say out of scope for this PR).

Also, the other suggestion I have in the comment is to return the error on the next blocking acquisition call, but that doesn't make too much sense as that's a past error and doesn't reflect the present (where we may be able to block and succesfully acquire a lease).

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+	_, e1, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Set the lease duration to be higher so that the new refreshed lease will
+	// be guaranteed to have a higher expiry time, otherwise jitter may cause it
+	// to have a lower expiry time than the previous lease. We also need to set
+	// the LeaseRenewalTimeout to be high so we renew immediately.
+	savedLeaseDuration := LeaseDuration
```
vivekmenezes: since this is 5 minutes we don't need to set and reset it. You can take it out of this test.

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+	savedLeaseRenewalTimeout := LeaseRenewalTimeout
+
+	defer func() {
+		LeaseDuration = savedLeaseDuration
+		LeaseRenewalTimeout = savedLeaseRenewalTimeout
+	}()
+	LeaseDuration = 2 * LeaseDuration
+	LeaseRenewalTimeout = 2 * LeaseDuration
```
vivekmenezes: I would set this to 10ms so that it gets the timer to fire and renew the lease

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+
+	ts := leaseManager.mu.tables[tableDesc.ID]
+	// Force a lease refresh to occur the refreshing sequence
+	ts.mu.Lock()
+	ts.mu.timer.Reset(time.Nanosecond)
+	ts.mu.Unlock()
```
vivekmenezes: No need for all this. Just set LeaseRenewalTimeout to something small like 10ms

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+		// Acquire the newer lease.
+		_, e2, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if e2.WallTime <= e1.WallTime {
+			return errors.Errorf("expected new lease expiration (%s) to be after old lease expiration (%s)",
+				e2, e1)
+		} else if testAcquiredCount != 2 {
```
vivekmenezes: I'd change this to testAcquiredCount != 3 so that you are guaranteeing the timer fires twice. once for the time it is set for the first time and again. Also you need to read this counter using atomic.Get()

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+	ts.mu.Unlock()
+
+	// Keep checking for a new lease to be acquired by the refreshing routine.
+	testutils.SucceedsSoon(t, func() error {
+		// Acquire the newer lease.
+		_, e2, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+		if err != nil {
+			t.Error(err)
+		}
+
```
vivekmenezes: You want to add a defer leaseManager.Release() so that the lease you just acquired is released.

pkg/sql/lease.go
vivekmenezes: I think the right approach here is to catch the error and log it using log.Error() that way someone sees it in the logs

vivekmenezes (commented):
Looking really great. Let's make one last pass through it! Thanks

pkg/sql/lease_internal_test.go
lego: 👍

pkg/sql/lease_internal_test.go
lego: 👍

pkg/sql/lease_internal_test.go
lego: 👍, changed it to just be 3 or more leases.

pkg/sql/lease_internal_test.go
lego: 👍

pkg/sql/lease_internal_test.go
lego: It's being explicitly set to a higher value to make sure the new leases have higher expiration we can test with due to jitter. Otherwise, the tests have too much randomness. Depending on how high the original lease's jitter was and how low the new ones are, it could easily flake.

pkg/sql/lease.go
lego: 👍 great, even better!

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+	LeaseDuration = 2 * LeaseDuration
+
+	// Keep checking for a new lease to be acquired by the refreshing routine.
+	testutils.SucceedsSoon(t, func() error {
+		// Acquire the newer lease.
+		ts, e2, err := leaseManager.Acquire(context.TODO(), now, tableDesc.ID)
+		if err != nil {
+			t.Error(err)
+		}
+
```
vivekmenezes: move the defer here so that a successful Acquire() always results in a Release()

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+		if e2.WallTime <= e1.WallTime {
+			return errors.Errorf("expected new lease expiration (%s) to be after old lease expiration (%s)",
+				e2, e1)
+		}
+
+		defer leaseManager.Release(ts)
+
+		if atomic.LoadInt32(&testAcquiredCount) < 3 {
```
vivekmenezes: count := atomic.LoadInt32(&testAcquiredCount); count < 3

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,77 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+			return errors.Errorf("expected at least 3 leases to be acquired, but only acquired %d times",
+				atomic.LoadInt32(&testAcquiredCount))
```
vivekmenezes: use count here

pkg/sql/lease_internal_test.go
lego: This is done intentionally because if we release the first lease the call to acquire will block to get a new lease as opposed to the refresh routine acquiring it.

I'll add a comment to explicitly call this out.

pkg/sql/lease.go
```diff
@@ -573,20 +584,51 @@ func (t *tableState) acquire(
...
+//
+// t.mu needs to be locked.
+func (t *tableState) startRefreshingRoutine(ctx context.Context, m *LeaseManager) {
+	t.mu.timer = time.NewTimer(LeaseRenewalTimeout)
+	t.stopper.RunWorker(ctx, func(ctx context.Context) {
+		for {
+			select {
+			case <-t.mu.timer.C:
+				t.mu.Lock()
+				err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
```
lego: I think we need to add `acquireWait` here because I believe there's a race condition here. This routine can barge into `acquireNodeLease` while another routine in inside `acquireNodeLease` and just unlocked `t.mu` in order to call `m.LeaseStore.acquire`. This specifically happens because this routine doesn't respect the `acquiring` channel which acts as an implicit ordered lock.

(calling this now as I just thought of it and I'm fairly confident about it, albeit it's a rare condition, going to make the change)

pkg/sql/lease_internal_test.go
vivekmenezes: I don't think that's the case. Releasing a lease doesn't result in a new lease being acquired on the next acquire(). The refcount on a lease can go down to 0 without it resulting in it being released.

pkg/sql/lease.go
lego: I added `acquireWait`, but I didn't end up making the change due to running into unknown deadlock problems.

lego:
Just started testing with TPCC, and it hasn't yet gotten the problem (when it otherwise would have).

(from DM) I have a suspicion that the current LeaseRenewalTimeout isn’t low enough. If the lease has an expiration of 3.75m renewing at 3.5m might not be enough. Either we can lower it or we can make it always renew at 1 minute of expiry (would just need to weave it through, either it’s the jittered value or the minimum duration which we can get the delta of).

I also had to used `context.TODO()` in the new routine, because it can't use the passed in context -- the context gets canceled eventually, causing it to fail renewal.

pkg/sql/lease.go
```diff
@@ -587,6 +600,43 @@ func (t *tableState) acquire(
...
+
+// startRefreshingRoutine begins the goroutine responsible for refreshing a
+// table lease.
+// TODO(joey): Currently this routine will run forever for each table
+// descriptor. Ideally, we stop the routine if the table descriptor hasn't been
+// used recently.
+//
+// t.mu needs to be locked.
+func (t *tableState) startRefreshingRoutine(m *LeaseManager) {
+	ctx := context.TODO()
```
vivekmenezes: just pass in the context

pkg/sql/lease_internal_test.go
```diff
@@ -558,3 +562,76 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
...
+	// We set LeaseRenewalTimeout to be low so it will refresh frequently, and the
+	// LeaseJitterMultiplier to ensure any newer leases will have a higher
+	// expiration timestamp.
+	savedLeaseRenewalTimeout := LeaseRenewalTimeout
+	savedLeaseJitterMultiplier := LeaseJitterMultiplier
+	defer func() {
+		LeaseRenewalTimeout = savedLeaseRenewalTimeout
+		LeaseJitterMultiplier = savedLeaseJitterMultiplier
+	}()
+	LeaseRenewalTimeout = 10 * time.Millisecond
```
vivekmenezes: let's just make this 5 * time.Millisecond

vivekmenezes (approved):

vivekmenezes:
LGTM!

pkg/sql/lease.go
lego: I mentioned this in the last comment, but it turned out that that was failing (it wasn't evident until I was testing cockroach -- I'm not sure how this could be testable). It's because when the refreshing routine goes to use the context (several minutes later) the context has been canceled.

pkg/sql/lease.go
```diff
@@ -800,6 +850,16 @@ func (t *tableState) acquireNodeLease(
 	}
 	t.upsertLocked(ctx, table, m)
 	t.tableNameCache.insert(table)
+
+	// Either begin a new routine to start refreshing the lease asynchronously or
+	// or reset the timer for it. After the first lease acquisition, we should
+	// always have a routine actively renewing the lease.
```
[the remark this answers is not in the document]
lego: Good point didn't think about that. In addition, the routine may stop due to `tableState.stopper`, but I never concluded if the stopper was ever stopped.

pkg/sql/lease.go
vivekmenezes: Okay leave it as is

vivekmenezes (commented):
LGTM

andreimatei:
Unless I'm missing something, this patch makes it so that once a node has acquired a lease for a table, it will renew it forever. There's got to be something better we can do...

---

Review status: 0 of 3 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 59 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3KgXpOYjGP55dkPUj:-Kv3KgXpOYjGP55dkPUk:baw21w9) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L59)):*
> ```Go
> 	// TODO(joey): We can make the timeout an offset from the actual expiration,
> 	// after jitter.
> 	LeaseRenewalTimeout = LeaseDuration - time.Duration(float64(LeaseDuration)*LeaseJitterMultiplier) - 30*time.Second
> ```

the point of the the duration jitter is to make it so that, if a bunch of leases are acquired at once, they don't also expire at once, causing re-acquisitions to always be in sync.
You've defeated that purpose with the fixed renewal timeout. No?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-:-Kv3LBafGADzMqMKyNAD:b-ufwrcb)*

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 863 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3PAO6OcShc_UyV8A9:-Kv3PAO6OcShc_UyV8AA:b-545xsa) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L863)):*
> ```Go
> 		t.startRefreshingRoutine(m)
> 	} else {
> 		t.mu.timer.Reset(LeaseRenewalTimeout)
> ```

the code seems unnecessarily convoluted to me; I'd suggest moving the `Reset` in `startRefreshingRoutine`, after we've received the timer event. Better yet, you can use a `timer.Ticker` if you truly want the refresh to be periodic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 863 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3PAO6OcShc_UyV8A9:-Kv3Rry9bVFQDSU87NLS:b-tg3bep) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L863)):*

the code seems unnecessarily convoluted to me; I'd suggest moving the `Reset` in `startRefreshingRoutine`, after we've received the timer event. Better yet, you can use a `timer.Ticker` if you truly want the refresh to be periodic.
</blockquote></details>

Thoughts on a `timer.Ticker` using a jittered renewal timeout, in order to also address your other comment about the non-jittered renewal times? They'll be periodic and also alleviate that problem.

The `Reset` was here because if a renewal fails then we rely on the next synchronous call to `acquire` (due to expiry) or `ensureVersion` to kickstart the timer again. Alternatively, if the renewal routine fails it can clean up the timer and end the goroutine, waiting for the next acquire to start it again.

I'm in favor of both changes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 863 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3PAO6OcShc_UyV8A9:-Kv3Tjchbnf4yMosBHfx:b-gum5f7) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L863)):*

Thoughts on a `timer.Ticker` using a jittered renewal timeout, in order to also address your other comment about the non-jittered renewal times? They'll be periodic and also alleviate that problem.

The `Reset` was here because if a renewal fails then we rely on the next synchronous call to `acquire` (due to expiry) or `ensureVersion` to kickstart the timer again. Alternatively, if the renewal routine fails it can clean up the timer and end the goroutine, waiting for the next acquire to start it again.

I'm in favor of both changes.
</blockquote></details>

Ignore the Reset move if it's a ticker, but the routine ending behaviour applies for that ticker change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 605 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3UyGb-bV58LXmWnC2:-Kv3UyGb-bV58LXmWnC3:b-v6tigw) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L605)):*
> This is used for making sure a
> // new lease moves forward when refreshing the lease

I don't  know what this means. A lease "moving forward" is not a term we've used before. I can suggest a more precise explanation if this method stays.

---

*[pkg/sql/lease.go, line 610 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3VZPlwKJVHJDSoXcv:-Kv3VZPlwKJVHJDSoXcw:b-jayi8a) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L610)):*
> ```Go
> 	defer t.mu.Unlock()
> 
> 	// Block and wait for the new lease to be acquired on another routine to
> ```

I don't understand this method / don't see the need for it in relation with lease refresh.
If another lease acquisition is in progress, we probably shouldn't initiate a new one immediately after.

---

*[pkg/sql/lease.go, line 632 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3V7y7gFDaZn2FJYA9:-Kv3V7y8pPmXQfW2v9a9:b-jx2ci1) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L632)):*
> ```Go
> 			case <-t.mu.timer.C:
> 				if err := t.acquireAlways(ctx, m); err != nil {
> 					log.Error(ctx, err)
> ```

please wrap this error with more info. `errors.Wrapf()` or just put more info in the log message about the table in question and the fact that this is async refresh.

---

*[pkg/sql/lease.go, line 863 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-Kv3PAO6OcShc_UyV8A9:-Kv3WhXDc3O17o0k7q72:bz4js1p) ([raw file](https://github.com/cockroachdb/cockroach/blob/ec9f73b058b247de7b136a66a7a5a41641d9175c/pkg/sql/lease.go#L863)):*

Ignore the Reset move if it's a ticker, but the routine ending behaviour applies for that ticker change.
</blockquote></details>

I don't see the need for any period work. I think async refresh should be initiated on table access (as the commit says, but it lies AFAICT). For example, initiate the refresh if a lease is found to be more than half through its life when the table is accessed (and no other lease for a more recent version exists in the `tableState`).

Unrelated nit: in the future, please make the PR description the same as the commit message (if there's a single commit). If there's multiple commits, usually choose one of them for the PR description. In any case, don't put text in the PR description that's not reflected in any commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606)*

lego:
Initially, the approach I tried was as you said, we refreshed the lease when it was about to acquire and not periodically. Initially, it was pretty complex and so we explored doing a periodic refresh which is what this became.

I strongly agree with a lot of your FUD and after chatting with Vivek we figured out how to fix the complexity with the other approach, so PR incoming!

---

Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18606#-:-Kv4BC116F-4hBBQsdMI:bfbm4yz)*

PR #18611, opened 2017-09-19
kv: ForkCtxSpan in lookupRangeDescriptorInternal

nvb:
This change addresses a TODO in `lookupRangeDescriptorInternal` introduced
in #18197. We now simply fork the `ContextSpan` in `lookupRequests.DoChan`
as the leader of the coalesced `RangeLookupRequest`. This is permitted because
we no longer listen to context cancellation outside of `DoChan` as the lookup
leader. Instead, we guarantee that context cancellations are propagated
over the singleflight result channel, which assures that
`lookupRangeDescriptorInternal` never returns before `DoChan`. This
means that the parent span can never close before the child span is
forked, which was the issue that prompted the TODO.

We are able to do this because we now return whether a given call to
`DoChan` will lead a `Group` or not. In doing so, we also improve the
testing of `singleflight.Group`.

This, in turn, is all possible because the first two commits in this change
fork singleflight from `golang.org/x/sync/singleflight`.

cc. @petermattis @bdarnell who may remember why this TODO was introduced
in the first place.

### DIFF
--- a/build/style_test.go
+++ b/build/style_test.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"testing"
@@ -551,6 +552,26 @@ func TestStyle(t *testing.T) {
 
 	t.Run("TestForbiddenImports", func(t *testing.T) {
 		t.Parallel()
+
+		// forbiddenImportPkg -> permittedReplacementPkg
+		forbiddenImports := map[string]string{
+			"context": "golang.org/x/net/context",
+			"log":     "util/log",
+			"path":    "path/filepath",
+			"github.com/golang/protobuf/proto": "github.com/gogo/protobuf/proto",
+			"github.com/satori/go.uuid":        "util/uuid",
+			"golang.org/x/sync/singleflight":   "github.com/cockroachdb/cockroach/pkg/util/syncutil/singleflight",
+		}
+
+		// grepBuf creates a grep string that matches any forbidden import pkgs.
+		var grepBuf bytes.Buffer
+		grepBuf.WriteByte('(')
+		for forbiddenPkg := range forbiddenImports {
+			grepBuf.WriteByte('|')
+			grepBuf.WriteString(regexp.QuoteMeta(forbiddenPkg))
+		}
+		grepBuf.WriteString(")$")
+
 		filter := stream.FilterFunc(func(arg stream.Arg) error {
 			for _, useAllFiles := range []bool{false, true} {
 				buildContext := build.Default
@@ -587,8 +608,8 @@ func TestStyle(t *testing.T) {
 			filter,
 			stream.Sort(),
 			stream.Uniq(),
+			stream.Grep(`^`+settingsPkgPrefix+`: | `+grepBuf.String()),
 			stream.GrepNot(`cockroach/pkg/cmd/`),
-			stream.Grep(`^`+settingsPkgPrefix+`: | (github\.com/golang/protobuf/proto|github\.com/satori/go\.uuid|log|path|context|syscall)$`),
 			stream.GrepNot(`cockroach/pkg/(cli|security): syscall$`),
 			stream.GrepNot(`cockroach/pkg/(base|security|util/(log|randutil|stop)): log$`),
 			stream.GrepNot(`cockroach/pkg/(server/serverpb|ts/tspb): github\.com/golang/protobuf/proto$`),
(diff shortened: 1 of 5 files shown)


petermattis:
Reviewed 2 of 2 files at r1, 3 of 3 files at r2.
Review status: 1 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/kv/range_cache.go, line 351 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18611#-KuRk5hOxz0zZ74HW96y:-KuRk5hOxz0zZ74HW96z:b-9yng31) ([raw file](https://github.com/cockroachdb/cockroach/blob/cc249707f421ef4a49a62bbcbc3786ded121ea7f/pkg/kv/range_cache.go#L351)):*
> ```Go
> 	ctxDone := ctx.Done()
> 	if leader {
> 		ctxDone = nil
> ```

Is blocking the leader when the context has been cancelled problematic? Another approach would be to move the `tracing.ForkCtxSpan` outside the call to `DoChan` and then to call `tracing.FinishSpan()` here if `!leader`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18611)*

nvanbenschoten:

---

*[pkg/kv/range_cache.go, line 351 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18611#-KuRk5hOxz0zZ74HW96y:-KuRmqjVDDF2cvjGjQA4:b-qv9ar8) ([raw file](https://github.com/cockroachdb/cockroach/blob/cc249707f421ef4a49a62bbcbc3786ded121ea7f/pkg/kv/range_cache.go#L351)):*

Is blocking the leader when the context has been cancelled problematic? Another approach would be to move the `tracing.ForkCtxSpan` outside the call to `DoChan` and then to call `tracing.FinishSpan()` here if `!leader`.
</blockquote></details>

The leader still won't be blocked during context cancellations, we just won't detect the cancellation here. Instead, it will be detected in `performRangeLookup` (or somewhere down the stack, probably at the transport level). This assures that a leader context cancellation is propagated over `resC` instead, which gives us the nice behavior that for the leader, `lookupRangeDescriptorInternal` always outlives the closure. 

This is all tested in `TestRangeCacheContextCancellation`. Coincidentally, while writing that test I originally ran into an issue because the context cancellation could race and come from either `<-resC` or `<-ctx.Done()`. That will no longer be an issue.

There were other ways to address the TODO, like what @RaduBerinde suggested in https://github.com/cockroachdb/cockroach/pull/18197#issuecomment-327173669, but I chose this approach because of this extra behavior change. By limiting the possible concurrent states and making sure that the `DoChan` fn is always finished before `lookupRangeDescriptorInternal` returns, I predict that we'll have an easier time reasoning about this in the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18611)*

petermattis:
:lgtm:

---

---

*[pkg/kv/range_cache.go, line 351 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18611#-KuRk5hOxz0zZ74HW96y:-KuRryJzcvJmCIjsrqW8:b-8fkdvz) ([raw file](https://github.com/cockroachdb/cockroach/blob/cc249707f421ef4a49a62bbcbc3786ded121ea7f/pkg/kv/range_cache.go#L351)):*

The leader still won't be blocked during context cancellations, we just won't detect the cancellation here. Instead, it will be detected in `performRangeLookup` (or somewhere down the stack, probably at the transport level). This assures that a leader context cancellation is propagated over `resC` instead, which gives us the nice behavior that for the leader, `lookupRangeDescriptorInternal` always outlives the closure. 

This is all tested in `TestRangeCacheContextCancellation`. Coincidentally, while writing that test I originally ran into an issue because the context cancellation could race and come from either `<-resC` or `<-ctx.Done()`. That will no longer be an issue.

There were other ways to address the TODO, like what @RaduBerinde suggested in https://github.com/cockroachdb/cockroach/pull/18197#issuecomment-327173669, but I chose this approach because of this extra behavior change. By limiting the possible concurrent states and making sure that the `DoChan` fn is always finished before `lookupRangeDescriptorInternal` returns, I predict that we'll have an easier time reasoning about this in the future.
</blockquote></details>

Ah, got it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18611#-:-KuRsFGw9zPP5Hv8Sz34:bnfp4nl)*

PR #18612, opened 2017-09-19
sql: plumb traceKV into kvfetcher for scans

jordanlewis:
This commit plumbs traceKV into the kvfetcher and adds a new kv trace
output that describes a scan which is about to be performed, such as:

`Scan /Table/52/{1-2}`

This is useful to see because the row fetcher by itself currently only
shows the fetched rows, but not the span it is iterating over.

### DIFF
--- a/pkg/sql/distsqlrun/columnbackfiller.go
+++ b/pkg/sql/distsqlrun/columnbackfiller.go
@@ -186,7 +186,7 @@ func (cb *columnBackfiller) runChunk(
 		// populated and deleted by the OLTP commands but not otherwise
 		// read or used
 		if err := cb.fetcher.StartScan(
-			ctx, txn, []roachpb.Span{sp}, true /* limitBatches */, chunkSize,
+			ctx, txn, []roachpb.Span{sp}, true /* limitBatches */, chunkSize, false, /* traceKV */
 		); err != nil {
 			log.Errorf(ctx, "scan error: %s", err)
 			return err
--- a/pkg/sql/distsqlrun/indexbackfiller.go
+++ b/pkg/sql/distsqlrun/indexbackfiller.go
@@ -144,7 +144,7 @@ func (ib *indexBackfiller) runChunk(
 		// populated and deleted by the OLTP commands but not otherwise
 		// read or used
 		if err := ib.fetcher.StartScan(
-			ctx, txn, []roachpb.Span{sp}, true /* limitBatches */, chunkSize,
+			ctx, txn, []roachpb.Span{sp}, true /* limitBatches */, chunkSize, false, /* traceKV */
 		); err != nil {
 			log.Errorf(ctx, "scan error: %s", err)
 			return nil, err
--- a/pkg/sql/distsqlrun/joinreader.go
+++ b/pkg/sql/distsqlrun/joinreader.go
@@ -157,7 +157,7 @@ func (jr *joinReader) mainLoop(ctx context.Context) error {
 			})
 		}
 
-		err := jr.fetcher.StartScan(ctx, txn, spans, false /* no batch limits */, 0)
+		err := jr.fetcher.StartScan(ctx, txn, spans, false /* no batch limits */, 0, false /* traceKV */)
 		if err != nil {
 			log.Errorf(ctx, "scan error: %s", err)
 			return err
--- a/pkg/sql/distsqlrun/tablereader.go
+++ b/pkg/sql/distsqlrun/tablereader.go
@@ -193,7 +193,7 @@ func (tr *tableReader) Run(ctx context.Context, wg *sync.WaitGroup) {
 	}
 
 	if err := tr.fetcher.StartScan(
-		ctx, txn, tr.spans, true /* limit batches */, tr.limitHint,
+		ctx, txn, tr.spans, true /* limit batches */, tr.limitHint, false, /* traceKV */
 	); err != nil {
 		log.Errorf(ctx, "scan error: %s", err)
 		tr.out.output.Push(nil /* row */, ProducerMetadata{Err: err})
--- a/pkg/sql/logictest/testdata/logic_test/show_trace
+++ b/pkg/sql/logictest/testdata/logic_test/show_trace
@@ -173,6 +173,7 @@ query TTT
 SELECT span, operation, message FROM [SHOW KV TRACE FOR UPSERT INTO t.kv(k, v) VALUES (2,3)]
 ----
 (0,2)  consuming rows    output row: []
+(0,0)  sql txn implicit  Scan /Table/52/1/{2-3}
 (0,0)  sql txn implicit  querying next range at /Table/52/1/2
 (0,0)  sql txn implicit  r1: sending batch 1 Scan to (n1,s1):1
 (0,0)  sql txn implicit  CPut /Table/52/1/2/0 -> /TUPLE/2:2:Int/3
@@ -184,6 +185,7 @@ query TTT
 SELECT span, operation, message FROM [SHOW KV TRACE FOR UPSERT INTO t.kv(k, v) VALUES (1,2)]
 ----
 (0,2)  consuming rows    output row: []
+(0,0)  sql txn implicit  Scan /Table/52/1/{1-2}
 (0,0)  sql txn implicit  querying next range at /Table/52/1/1
 (0,0)  sql txn implicit  r1: sending batch 1 Scan to (n1,s1):1
 (0,0)  sql txn implicit  fetched: /kv/primary/1/v -> /2
@@ -195,6 +197,7 @@ query TTT
 SELECT span, operation, message FROM [SHOW KV TRACE FOR UPSERT INTO t.kv(k, v) VALUES (2,2)]
 ----
 (0,2)  consuming rows    output row: []
+(0,0)  sql txn implicit  Scan /Table/52/1/{2-3}
 (0,0)  sql txn implicit  querying next range at /Table/52/1/2
 (0,0)  sql txn implicit  r1: sending batch 1 Scan to (n1,s1):1
 (0,0)  sql txn implicit  fetched: /kv/primary/2/v -> /3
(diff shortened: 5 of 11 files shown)


andreimatei:
I think the new `Scan` messages should always be part of the trace. Each one probably already generates a ton of other messages, particularly if it's an RPC.
Also, can't all this `traceKV` stuff be done at a KV level? In the `DistSender`, for example. Then we wouldn't need to litter all clients with code about it.

---

Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/errors.go, line 278 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSk_Chtm8EgvRfSDLH:-KuSk_Chtm8EgvRfSDLI:b-96w0vb) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/errors.go#L278)):*
> ```Go
> 		}
> 		f := singleKVFetcher{kv: client.KeyValue{Key: key, Value: cErr.ActualValue}}
> 		if err := rf.StartScanFrom(ctx, &f, false /* traceKV */); err != nil {
> ```

why false?

---

*[pkg/sql/sqlbase/fk.go, line 356 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSlicJqm5cWyq-wHvo:-KuSlicJqm5cWyq-wHvp:beonqxj) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/fk.go#L356)):*
> ```Go
> // foreign key was found.
> // TODO(dt): Batch checks of many rows.
> func (f baseFKHelper) check(ctx context.Context, values parser.Datums, traceKV bool) (bool, error) {
> ```

can traceKV be a config option on the `baseFKHelper`?

---

*[pkg/sql/sqlbase/kvfetcher.go, line 208 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSlPqbfRVSFALBfUqW:-KuSlPqbfRVSFALBfUqX:b594ua2) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/kvfetcher.go#L208)):*
> ```Go
> 
> // fetch retrieves spans from the kv
> func (f *txnKVFetcher) fetch(ctx context.Context, traceKV bool) error {
> ```

can this be an configuration option on the `kvFetcher` and not an argument to every single `fetch` call? It's not intended to change between calls, is it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-:-KuSmFxDDmh3XlOeY935:b-cbdq1u)*

knz:
>
>can traceKV be a config option on the `baseFKHelper`?
>can this be an configuration option on the `kvFetcher` and not an
>argument to every single `fetch` call? It's not intended to change
>between calls, is it?

1) the flag is not known by the time the data structures are instantiated. It is only set after the logical plan has been built.

2) I think it is useful to be able to set it for a sub range of rows in a query in tests to limit the size of the trace.



-- 

jordanlewis:
Thanks for the suggestions @andreimatei! I'd like to get this merged without a big refactor to the whole traceKV stuff though, since I'd like to get this into 1.1. I think having the Scan line will help with debugging and I'd rather not hold up on a bigger reimagining of the system.

I'm happy to work with you to improve this system going forward though.

---

Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/errors.go, line 278 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSk_Chtm8EgvRfSDLH:-KuUhAnGn2HidIYjWx7p:bitugoh) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/errors.go#L278)):*

why false?
</blockquote></details>

This uses the singleKVFetcher (it's a mock that just returns the above key/value) so nothing would be output anyway.

---

*[pkg/sql/sqlbase/fk.go, line 356 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSlicJqm5cWyq-wHvo:-KuUhJJaqV2xWvXuQDWl:bw5wu54) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/fk.go#L356)):*

can traceKV be a config option on the `baseFKHelper`?
</blockquote></details>

I find this more straightforward because it follows the same threading pattern we use elsewhere.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 208 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSlPqbfRVSFALBfUqW:-KuUhRCNLBXw6inDzaeP:b-e1dmlo) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/kvfetcher.go#L208)):*

can this be an configuration option on the `kvFetcher` and not an argument to every single `fetch` call? It's not intended to change between calls, is it?
</blockquote></details>

Like Raphael says, it might change between calls so you can turn on tracing in the middle of a query? I suppose? I'm fine with not permitting that but I'd prefer that be done in a separate PR because it happens elsewhere as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-:-KuUhjl-oaHm4AgZk7da:b-xy2eb0)*

andreimatei:
Well if you don't want more discussion, why not always output that scan line with `VEventf()`, and stop mocking around with `traceKV`?

Anyway, LGTM.

---

Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/errors.go, line 278 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSk_Chtm8EgvRfSDLH:-KuUtoacIOwg4kH_x8ZT:b3o465) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/errors.go#L278)):*

This uses the singleKVFetcher (it's a mock that just returns the above key/value) so nothing would be output anyway.
</blockquote></details>

"yozer!" 
This explains why I couldn't figure out what the fell this code was doing... Why did someone think it was a good idea to use a rowFetcher with a mock kvFetcher just to do the decoding of data you already have...
Please take the opportunity to add whatever comment you can express coherently about what this code is generally doing and/or this mocking in particular.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 208 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSlPqbfRVSFALBfUqW:-KuV9NIvW8IM7hK1han6:bt8ox7m) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/kvfetcher.go#L208)):*

Like Raphael says, it might change between calls so you can turn on tracing in the middle of a query? I suppose? I'm fine with not permitting that but I'd prefer that be done in a separate PR because it happens elsewhere as well.
</blockquote></details>

It doesn't change in the middle of a query.
Perhaps it changes in between planning and execution, but then still it should be a single setting call (even if it's not the ctor).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-:-KuV9d1JV0F7lIlAjp2Y:b-5gv7l4)*

knz:
On 20-09-17 12:43, Andrei Matei wrote:
> why not always output that scan line with |VEventf()|

Because in benchmarks you'd be paying that with a 10% drop in scan
throughput.

-- 
Raphael 'kena' Poss

andreimatei:
ha? Drop in benchmark from *one log message per Scan* call? Citation needed.
Note that we're not talking about per-row operations here.

knz:
On 20-09-17 13:31, Andrei Matei wrote:
> Note that we're not talking about per-row operations here.

I was assuming you were suggesting replacing *all* uses of the traceKV
parameter by a naked VEvent call. In that case the overhead would be per
row.

-- 
Raphael 'kena' Poss

jordanlewis:
Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/errors.go, line 278 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSk_Chtm8EgvRfSDLH:-KuWJM3LXPoAdGzBbD5R:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/errors.go#L278)):*

"yozer!" 
This explains why I couldn't figure out what the fell this code was doing... Why did someone think it was a good idea to use a rowFetcher with a mock kvFetcher just to do the decoding of data you already have...
Please take the opportunity to add whatever comment you can express coherently about what this code is generally doing and/or this mocking in particular.



</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 208 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612#-KuSlPqbfRVSFALBfUqW:-KuWJNIKt7DiUT5ioKr4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf0de553f4de91b62859ba38d413d490dac542b/pkg/sql/sqlbase/kvfetcher.go#L208)):*

It doesn't change in the middle of a query.
Perhaps it changes in between planning and execution, but then still it should be a single setting call (even if it's not the ctor).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18612)*

Issue #18206, opened 2017-09-05
ui: Login Flow for Admin UI


mrtracy:
The Admin API now fully supports cookie-based authentication, which will restrict admin functions to authenticated users.  However, the UI does not currently provide a way for a user to authenticate, so that system has been disabled in 1.1.

A login page should be added to the frontend, and all other activities on the Admin UI should be protected by that login.  Details are available in the "web_session_login" RFC found under `/docs/RFCS`.

Works towards #6307

dianasaur323:
Draft Acceptance Criteria - Accepting Comments

**User Story**
This is an important open source and enterprise user request. Currently, you have to some manual workarounds to block access to the admin UI. 

**Feature Scope**
Given that we will be implementing RBAC in 1.2, we will need to update user login to match that use case. For now, we will release user login as a beta feature before rolling it out permanently. The beta feature will be composed of:
- [ ] Simple login page that allows a user to log into the admin UI
- [ ] Talks to the backend, which authorizes any user with a password to log into the admin UI
- [ ] We will need a cluster setting to turn this on. It should be off by default so as not to effect our existing users and their workflows

**Required States**
- [ ] Login page
- [ ] Loading state
- [ ] Some sign of logged in status in the admin UI

**PM Acceptance Testing**
- [ ] Create a user and password
- [ ] Login as that user
- [ ] Observe the different states that the admin UI gets into
- [ ] Try accessing the admin UI from a different node
- [ ] Try chrome, safari, mozilla, IE
- [ ] Observe what happens when you close out the window

couchand:
Closed in #28416

PR #18619, opened 2017-09-20
ui,server: Add show queries page

lgo:
A bit WIP, but I've added a basic show queries page which is largely a copy of the jobs page.

I'm not sure what the goal of this page in particular was, but it was something I did a week back when we were talking about some of the tracing work and this was an MVP for having more details on existing queries cc: @vivekmenezes. The goal of the page would dictate where it shows up (main page vs. debug page) and what we want to show and emphasize here.


lego:
What it currently looks like
screen shot 2017-09-20 at 11 08 27 am

lego:
Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/server/admin.go, line 1164 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18619#-KuUpp0LCQKS98SET_4q:-KuUpp0LCQKS98SET_4r:b-m42pvx) ([raw file](https://github.com/cockroachdb/cockroach/blob/41abfb1e592a06583dbf53e4b17170df5673ea6b/pkg/server/admin.go#L1164)):*
> ```Go
> 			&query.ClientAddress,
> 			&query.ApplicationName,
> 			// &query.Distributed,
> ```

I intentionally omitted the `Distributed` column for now as I wasn't sure what the column type was or what it represents. My guess is it denotes whether or not the query is using DistSQL?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18619)*

bdarnell:
Queries often contain sensitive information, so I think we have to defer this until we have authentication for the UI.

vivekmenezes:
Thanks for doing this!

blocked on https://github.com/cockroachdb/cockroach/issues/18206

instead of the query id, it would be cool to just have a cancel button to the right of each row. No need to do this before the blocking issue is closed and before receiving more input from the design folks.

@kuanluo this PR is a working PR for SHOW QUERIES on the admin UI. Let's follow up separately with matt on when he expects this to get unblocked

dianasaur323:
this definitely needs a UI pass, but thanks for doing it! It's unlikely we'll have login until probably late Q4, so I'll put this on the schedule to clean up, but unlikely for us to get to it anytime soon.

dianasaur323:
@lego I'm skimming through the frontend bugs we have right now, and the list is pretty crazy. I'd like for us to get this into 1.2, but for the sake of not being overly aggressive on that list, I put this into 1.3 for now. If we have extra time, we can get to this, but it might not make it in.

lego:
A-okay :). I just tossed this PR up since I did it while playing around and
had no expectation.

vilterp:
Closing out, as we'll probably take a different approach with a query perf page. I.e. showing workload over time, not just currently running queries.

PR #18624, opened 2017-09-20
sqlccl: set a good open file limit for load csv

madelynnblue:


bdarnell:

---

*[pkg/ccl/sqlccl/csv.go, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624#-KuVhoYuqCrba5GU0FY0:-KuVhoYuqCrba5GU0FY1:b-8c5ub3) ([raw file](https://github.com/cockroachdb/cockroach/blob/86c77c35fb91f9e190385fc4fd287cd3a1dc40d5/pkg/ccl/sqlccl/csv.go#L130)):*
> ```Go
> 	}()
> 
> 	fileLimit, err := server.SetOpenFileLimitForOneStore()
> ```

This is too high (i believe it's also guaranteed to be higher than the hard-coded 1024 that was here before). The logic in `server/config_unix.go` already allocates all the file descriptors to different purposes. You're claiming all the file descriptors that a store could use if it were the only store in the system; this only works because the other stores (and the "network"/misc bucket) don't eagerly consume all the file descriptors they've been allocated. 

We need to reserve some file descriptors here that won't be given to the regular stores, and also to ensure that only one LoadCSV operation can proceed at a time (unless we assign smaller limits to them). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624#-KuVhoYuqCrba5GU0FY0:-KuVjmDRLpOanbXJqdmS:b-gx4mes) ([raw file](https://github.com/cockroachdb/cockroach/blob/86c77c35fb91f9e190385fc4fd287cd3a1dc40d5/pkg/ccl/sqlccl/csv.go#L130)):*

This is too high (i believe it's also guaranteed to be higher than the hard-coded 1024 that was here before). The logic in `server/config_unix.go` already allocates all the file descriptors to different purposes. You're claiming all the file descriptors that a store could use if it were the only store in the system; this only works because the other stores (and the "network"/misc bucket) don't eagerly consume all the file descriptors they've been allocated. 

We need to reserve some file descriptors here that won't be given to the regular stores, and also to ensure that only one LoadCSV operation can proceed at a time (unless we assign smaller limits to them). 
</blockquote></details>

It's not obvious and I'm not sure if you were aware, but all of this is happening inside of a CLI invocation. This mode doesn't start or ever contact a cockroach server, and doesn't make any network requests. I was using this API because of its convenient checking and setting of the system limits. If multiple `load csv` invocations are done at the same time, they are being run from different shells. Is your comment still valid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624)*

bdarnell:

---

*[pkg/ccl/sqlccl/csv.go, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624#-KuVhoYuqCrba5GU0FY0:-KuVl5qRQnT-1v5_F9Ta:b7ch7ws) ([raw file](https://github.com/cockroachdb/cockroach/blob/86c77c35fb91f9e190385fc4fd287cd3a1dc40d5/pkg/ccl/sqlccl/csv.go#L130)):*

It's not obvious and I'm not sure if you were aware, but all of this is happening inside of a CLI invocation. This mode doesn't start or ever contact a cockroach server, and doesn't make any network requests. I was using this API because of its convenient checking and setting of the system limits. If multiple `load csv` invocations are done at the same time, they are being run from different shells. Is your comment still valid?
</blockquote></details>

Ah, in that case this is fine. I thought this was for the SQL `IMPORT` statement. (it's in package `sqlccl`; maybe it should be in `cliccl` instead?)

I think it's still true though that SetOpenFileLimitForOneStore will never return a value smaller than 1700 (if not there are some stale comments). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624#-KuVhoYuqCrba5GU0FY0:-KuVrYfJBodVVfPY4s5I:bcdjfz8) ([raw file](https://github.com/cockroachdb/cockroach/blob/86c77c35fb91f9e190385fc4fd287cd3a1dc40d5/pkg/ccl/sqlccl/csv.go#L130)):*

Ah, in that case this is fine. I thought this was for the SQL `IMPORT` statement. (it's in package `sqlccl`; maybe it should be in `cliccl` instead?)

I think it's still true though that SetOpenFileLimitForOneStore will never return a value smaller than 1700 (if not there are some stale comments). 
</blockquote></details>

Yes, but what Cuong and I think happened is that it hit the soft limit, not the hard limit, and this function call increases the soft limit. So I do think that even though the number is going up from what it was before, it is safer because the soft limit is increased, and and error is returned if it can't be increased. I think 1700 is below the default hard limits of most systems, so this change should result in less errors like this in the future.

I do agree that this should probably move to the cli package. That can be in another PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18624)*

bdarnell (approved):

PR #18627, opened 2017-09-20
storage: avoid trace use-after-finish

tbg:
action() closed over a context but was then called in a goroutine.
Use the context provided by the stopper instead, which inserts
a span with the proper lifetime.


nvb (approved):

bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18627#-:-KuVgYtCb_Nel1byhT7B:bnfp4nl)*

PR #18628, opened 2017-09-20
keys: highlight a panic in EnsureSafeSplitKey

tbg:
cc @petermattis

### DIFF
--- a/pkg/keys/keys.go
+++ b/pkg/keys/keys.go
@@ -640,7 +640,9 @@ func EnsureSafeSplitKey(key roachpb.Key) (roachpb.Key, error) {
 	if err != nil {
 		return nil, err
 	}
-	if int(colIDLen)+1 > n {
+	// Note how this next comparison (and by extension the code after it) is overflow-safe. There
+	// are more intuitive ways of writing this that aren't as safe. See #18628.
+	if colIDLen > uint64(n-1) {
 		// The column ID length was impossible. colIDLen is the length of
 		// the encoded column ID suffix. We add 1 to account for the byte
 		// holding the length of the encoded column ID and if that total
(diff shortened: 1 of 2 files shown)


petermattis:
Thanks for finding this.

---

---

*[pkg/keys/keys.go, line 655 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18628#-KuVcXOFz4z0vPXkd8lB:-KuVcXOFz4z0vPXkd8lC:byh3xq9) ([raw file](https://github.com/cockroachdb/cockroach/blob/d55d12912e9ee0892e57be93dd05a990b1a1711e/pkg/keys/keys.go#L655)):*
> ```Go
> 	}
> 	from := len(key) - int(colIDLen) - 1
> 	if from > len(key) {
> ```

The real problem is that `colIDLen` is being decoded as `18446744073709551610` and then cast to an `int` which gives a value of `-6`. I think we want to enhance the conditional above:

```
if int(colIDLen) < 0 || int(colIDLen)+1 > n {
  ...
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18628#-:-KuVdsTAJEjK8w4GRxh3:bkgq834)*

petermattis:

---

*[pkg/keys/keys.go, line 655 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18628#-KuVcXOFz4z0vPXkd8lB:-KuVyH6jdlrLMhIIcz1e:basro9g) ([raw file](https://github.com/cockroachdb/cockroach/blob/d55d12912e9ee0892e57be93dd05a990b1a1711e/pkg/keys/keys.go#L655)):*

The real problem is that `colIDLen` is being decoded as `18446744073709551610` and then cast to an `int` which gives a value of `-6`. I think we want to enhance the conditional above:

```
if int(colIDLen) < 0 || int(colIDLen)+1 > n {
  ...
}
```
</blockquote></details>

Did you want me to take over this PR?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18628)*

tschottdorf:
I've done a first pass, but if you think you can do better, feel free to take this over.

tschottdorf:
(but take note that I *just* pushed).

petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18628#-:-KuW6Bh8-7dVSul9MrmB:bnfp4nl)*

PR #18631, opened 2017-09-20
ui: remove Schema Change column from Database Tables page

### LINKED ISSUE #18594: ui: schema change column in database page seems to not serve any purpose
image
I am in the PM cluster running CREATE INDEX c_p ON customer(c_phone); and would expect something to show up in the schema changes column, but it's not. 

Tracked in https://github.com/cockroachdb/cockroach/issues/18593

couchand:
The Schema Change column never has any content.  This removes the
always-empty column.  Fixes #18594.

### DIFF
--- a/pkg/ui/src/views/databases/containers/databaseTables/index.tsx
+++ b/pkg/ui/src/views/databases/containers/databaseTables/index.tsx
@@ -99,10 +99,6 @@ class DatabaseSummaryTables extends DatabaseSummaryBase {
                         cell: (tableInfo) => tableInfo.numIndices,
                         sort: (tableInfo) => tableInfo.numIndices,
                       },
-                      {
-                        title: "Schema Change",
-                        cell: (_tableInfo) => "",
-                      },
                     ]} />
               }
             </div>


dianasaur323:
woot! lightspeed!

vivekmenezes:
I think it will be valuable for users to see if there are an existing schema changes scheduled for the job with a link to the jobs page perhaps? But I'm also happy for you to merge this and create another issue that creates a plan to do that correctly like provide a link to a filtered jobs table reflecting only jobs for schema changes associated with the table.

vivekmenezes:
perhaps this column can have a check mark or something indicating that a schema change exists

couchand:
It seems like a nice idea to have anything at all useful here @vivekmenezes.  Since that's unlikely to land before 1.1, I'm inclined to have that be a future issue, and go ahead and remove this empty column for now.

vivekmenezes:
I'm fine with blowing it away

vivekmenezes:
Wanna create another issue

mrtracy:
LGTM, lightning indeed

Issue #15931, opened 2017-05-15
ts: Make timeseries truncation threshold configurable


a-robinson:
Our timeseries data can take up a considerable amount of space in its default configuration, which samples metrics every 10 seconds and stores them for 30 days. We let users configure the sampling interval, but not the storage duration. We may want to let them configure both, although doing so could have consequences for the admin UI, which currently assumes that metrics are valid for a month.

Asked about in https://forum.cockroachlabs.com/t/live-bytes-steadily-increasing/629/4

Also, @mrtracy, is there an issue tracking downsampling of old metrics [as proposed in the RFC](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/time_series_culling.md)? All I see is #5910 about truncating them, which has been fixed.

dianasaur323:
This feature is also mentioned here: https://github.com/cockroachdb/cockroach/issues/16016

dianasaur323:
**PM Acceptance Testing**
- [ ] Set time series setting to something crazy short to see what happens in the admin UI

PR #18632, opened 2017-09-20
ts: Make timeseries truncation threshold configurable

pydima:
Closes: #15931

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -177,7 +177,7 @@ type Config struct {
 	// ReadWithinUncertaintyIntervalError.
 	MaxOffset MaxOffsetType
 
-	// MetricsSamplePeriod determines the time between records of
+	// MetricsSampleInterval determines the time between records of
 	// server internal metrics.
 	// Environment Variable: COCKROACH_METRICS_SAMPLE_INTERVAL
 	MetricsSampleInterval time.Duration
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -301,7 +301,7 @@ func NewServer(cfg Config, stopper *stop.Stopper) (*Server, error) {
 	s.adminMemMetrics = sql.MakeMemMetrics("admin", cfg.HistogramWindowInterval())
 	s.registry.AddMetricStruct(s.adminMemMetrics)
 
-	s.tsDB = ts.NewDB(s.db)
+	s.tsDB = ts.NewDB(s.db, s.cfg.Settings)
 	s.tsServer = ts.MakeServer(s.cfg.AmbientCtx, s.tsDB, s.cfg.TimeSeriesServerConfig, s.stopper)
 
 	sqlExecutor := sql.InternalExecutor{LeaseManager: s.leaseMgr}
--- a/pkg/sql/logictest/testdata/logic_test/show_source
+++ b/pkg/sql/logictest/testdata/logic_test/show_source
@@ -91,6 +91,7 @@ sql.metrics.statement_details.threshold            0s             d     minimum 
 sql.trace.log_statement_execute                    false          b     set to true to enable logging of executed statements
 sql.trace.session_eventlog.enabled                 false          b     set to true to enable session tracing
 sql.trace.txn.enable_threshold                     0s             d     duration beyond which all transactions are traced (set to 0 to disable)
+timeseries.resolution_10s.storage_duration         720h0m0s       d     the amount of time to store timeseries data
 trace.debug.enable                                 false          b     if set, traces for recent requests can be seen in the /debug page
 trace.lightstep.token                              ·              s     if set, traces go to Lightstep using this token
 trace.zipkin.collector                             ·              s     if set, traces go to the given Zipkin instance (example: '127.0.0.1:9411'); ignored if trace.lightstep.token is set.
--- a/pkg/storage/ts_maintenance_queue_test.go
+++ b/pkg/storage/ts_maintenance_queue_test.go
@@ -235,8 +235,8 @@ func TestTimeSeriesMaintenanceQueueServer(t *testing.T) {
 	seriesName := "test.metric"
 	sourceName := "source1"
 	now := tsrv.Clock().PhysicalNow()
-	nearPast := now - (ts.Resolution10s.PruneThreshold() * 2)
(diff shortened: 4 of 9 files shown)


bdarnell:
Thanks for the contribution! This looks good to me but @mrtracy should have the final say. Should we think ahead here to when we support downsampling and multiple resolutions, or is that far enough off that we should just do the simple thing for now?

---

Reviewed 9 of 9 files at r1.

---

*[pkg/server/config.go, line 595 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632#-KuWRIJI-JXiEIHkF8-G:-KuWRIJI-JXiEIHkF8-H:b-3n56xz) ([raw file](https://github.com/cockroachdb/cockroach/blob/cebdd5915ec8eda2e42af205503b4a1089449598/pkg/server/config.go#L595)):*
> ```Go
> 	cfg.Linearizable = envutil.EnvOrDefaultBool("COCKROACH_LINEARIZABLE", cfg.Linearizable)
> 	cfg.MetricsSampleInterval = envutil.EnvOrDefaultDuration("COCKROACH_METRICS_SAMPLE_INTERVAL", cfg.MetricsSampleInterval)
> 	cfg.MetricsStoreDurationDays = envutil.EnvOrDefaultInt64("COCKROACH_METRICS_STORE_DURATION_DAYS", cfg.MetricsStoreDurationDays)
> ```

I wish `EnvOrDefaultDuration` supported a `d` suffix so we didn't have to require this to be an integer number of days, but this is probably better than requiring this to be set in terms of hours.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632#-:-KuWTDsWOdobFOAUyT5T:bexi8pc)*

petermattis:

---

*[pkg/server/config.go, line 595 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632#-KuWRIJI-JXiEIHkF8-G:-KuWqDHRnThFNZLfOcjr:binoyoq) ([raw file](https://github.com/cockroachdb/cockroach/blob/cebdd5915ec8eda2e42af205503b4a1089449598/pkg/server/config.go#L595)):*

I wish `EnvOrDefaultDuration` supported a `d` suffix so we didn't have to require this to be an integer number of days, but this is probably better than requiring this to be set in terms of hours.
</blockquote></details>

Shouldn't this be a cluster setting (see `settings.RegisterDurationSetting`)? The usage of env vars is somewhat historical and are primarily useful for per-node settings. `COCKROACH_METRICS_SAMPLE_INTERVAL` should probably be a cluster setting as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632)*

bdarnell:

---

*[pkg/server/config.go, line 595 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632#-KuWRIJI-JXiEIHkF8-G:-KuX8DE-lLYUJF_iS0uu:b9kelza) ([raw file](https://github.com/cockroachdb/cockroach/blob/cebdd5915ec8eda2e42af205503b4a1089449598/pkg/server/config.go#L595)):*

Shouldn't this be a cluster setting (see `settings.RegisterDurationSetting`)? The usage of env vars is somewhat historical and are primarily useful for per-node settings. `COCKROACH_METRICS_SAMPLE_INTERVAL` should probably be a cluster setting as well.
</blockquote></details>

Oh, good point. Both of these should move to cluster settings. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632)*

mrtracy:
The environment-variable setting does need to be moved to a cluster setting.  Check out the file `pkg/settings/duration.go`, you should be able to grep on that file to find an example of duration settings.  They're a bit more straightforward to implement than command-line settings, the hardest part is choosing a name; I would name it "timeseries.resolution_10s.storage_duration", which leaves us flexibility if we implement multi-resolution storage in the future.

The only tedious part of adding a system setting is modifying the expected result of logic tests - there are tests the depend on the exact number of system variables being static, and you will need to modify those results.  Once you've added the setting, tests which have this dependency will fail until you have repaired them (running `make testlogic` is the fastest way to iterate on this.)

You will need to change to ts.DB to accept a *settings.Settings, from which you can read the time series setting.

The other modifications to ts.DB (moving the pruning intervals into the DB object) are implemented correctly, nice work.

pydima:
Thank you all for the review! I hope that I understood everything correctly, so I've changed the following:
- Created `Resolution10StoreDuration` as a cluster setting. it's datatype is duration, as suggested, but as @bdarnell wrote it feels a bit inconvenient to set it in terms of hours instead of days.
- `ts.NewDB` now accepts `*cluster.Settings`

And as for moving `COCKROACH_METRICS_SAMPLE_INTERVAL` to cluster settings, for me it's not trivial how to do it, since there are a bunch of tickers created from this variable. Any suggestions?

bdarnell:
LGTM. I think the COCKROACH_METRICS_SAMPLE_INTERVAL change is best left to a separate PR. It looks like it probably needs a lot of plumbing although nothing especially tricky. 

---

Reviewed 6 of 6 files at r2.

---

*[pkg/ts/db.go, line 51 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632#-KufNV-pgfQ_j1Hst_T_:-KufNV-pgfQ_j1Hst_Ta:b-his6o4) ([raw file](https://github.com/cockroachdb/cockroach/blob/3784a437f01f3045c70b58f4c53bcbaa2d19493e/pkg/ts/db.go#L51)):*
> ```Go
> 	// which is older than the given threshold for a resolution is considered
> 	// eligible for deletion. Thresholds are specified in nanoseconds.
> 	pruneThresholdByResolution map[Resolution]func() int64
> ```

In Go we generally use the `time.Duration` type instead of int64 as much as possible. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18632#-:-KufOj-bL7Vylo0A3UXX:bcpfnl5)*

pydima:
@mrtracy  do I need to do anything else here?

mrtracy:
No, this is LGTM and I am merging it.  Thanks for another contribution!

Issue #15997, opened 2017-05-17
storage: test GC queue in presence of many intents


tbg:
We have done little testing on this case.

cuongdo:
@tschottdorf should this move to 1.2? or later?

tschottdorf:
1.2 will do for now.

tschottdorf:
I wrote a little tool that fires up a local single node and writes expired txn entries into its liveness range. I haven't tested how the GC queue reacts yet, but the `cockroach debug estimate-gc` output looks promising (this is after running for ~5s)

```
$ ./cockroach debug estimate-gc ./cockroach-data-gcpressurizer/1/ 3
RangeID: 3 [/System/NodeLiveness, /System/NodeLivenessMax):
storage.GCInfo{
    Now:                        hlc.Timestamp{WallTime:1505719085523655593, Logical:0},
    Policy:                     config.GCPolicy{TTLSeconds:86400},
    NumKeysAffected:            0,
    IntentsConsidered:          0,
    IntentTxns:                 0,
    TransactionSpanTotal:       13000,
    TransactionSpanGCAborted:   0,
    TransactionSpanGCCommitted: 0,
    TransactionSpanGCPending:   13000,
    TxnSpanGCThreshold:         hlc.Timestamp{WallTime:1505715485523655593, Logical:0},
    AbortSpanTotal:             1,
    AbortSpanConsidered:        0,
    AbortSpanGCNum:             1,
    PushTxn:                    13000,
    ResolveTotal:               0,
    ResolveSuccess:             0,
    Threshold:                  hlc.Timestamp{WallTime:1505632685523655593, Logical:0},
```

This means that the GC queue will try to push 13k transactions the next time it runs. Each of the transactions holds 100 intents, so after that it will try to resolve 13k*100 = 1.3 million intents.

If a cluster can remain available with this going on on its liveness range, I think we're fairly safe. I'm sure it won't fare too well when I test this tomorrow.

tschottdorf:
cc @nvanbenschoten wrt #18199.

tschottdorf:
Yep, as expected this is pretty effective at overwhelming the GC queue on different levels.

![image](https://user-images.githubusercontent.com/5076964/30549022-a405e502-9c61-11e7-9bd1-5a923b38af28.png)

nvanbenschoten:
[...]

PR #18635, opened 2017-09-20
storage: add "noop" intent resolution poisoning option

tbg:
NB: I'm flushing this out of a slew of related work, and think the migration story and the
changes to when we clear the abort cache deserve a particularly scrutinous review.

----

Manual testing in https://github.com/cockroachdb/cockroach/issues/15997 surfaced that one limiting
factor in resolving many intents is contention on the transaction's abort cache entry. In one
extreme test, I wrote 10E6 abortable intents into a single range, in which case the GC queue sends
very large batches of intent resolution requests for the same transaction to the intent resolver.

These requests all overlapped on the transaction's abort cache key, causing very slow progress, and
ultimately preventing the GC queue from making a dent in the minute allotted to it. Generally this
appears to be a somewhat atypical case, but since @nvanbenschoten observed something similar in
https://github.com/cockroachdb/cockroach/issues/18199 it seemed well worth addressing, by means of

1. allow intent resolutions to not touch the abort span
2. correctly declare the keys for `ResolveIntent{,Range}` to only declare the abort cache key
   if it is actually going to be accessed.

With these changes, the gc queue was able to clear out a million intents comfortably on my older
13" MacBook (single node).

Also use this option in the intent resolver, where possible -- most transactions don't receive abort
cache entries, and intents are often "found" by multiple conflicting writers. We want to avoid
adding artificial contention there, though in many situations the same intent is resolved and so a
conflict still exists.

Migration: a new field number was added to the proto and the old one preserved. We continue to
populate it. Downstream of Raft, we use the new field but if it's unset, synthesize from the
deprecated field. I believe this is sufficient and we can just remove all traces of the old field in
v1.3. (v1.1 uses the old, v1.2 uses the new with compatibility for the old, v1.3 only the new field).


bdarnell:
Reviewed 3 of 8 files at r1.
Review status: 3 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/roachpb/api.proto, line 627 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_eImWYn4YNJ7AslS2:-Ku_eImWYn4YNJ7AslS3:bkqkvp7) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/roachpb/api.proto#L627)):*
> ```protobuf
>   // TODO(tschottdorf): this can be removed in CockroachDB v1.3.
>   optional bool deprecated_poison = 4 [(gogoproto.nullable) = false];
>   optional PoisonType poison = 5 [(gogoproto.nullable) = false];;
> ```

It's safe to change bools to enums in protobufs if the values line up (going from old to new, true and false are mapped to 1 and 0. going the other direction, zero is false and any non-zero is true). I'm not sure if that works out in this case or not. 

---

*[pkg/storage/gc_queue.go, line 423 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_ej7g-oEEHGSnK0d1:-Ku_ej7g-oEEHGSnK0d2:b-8cteuo) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/storage/gc_queue.go#L423)):*
> ```Go
> 				infoMu.Unlock() // intentional
> 				defer infoMu.Lock()
> 				// Contention between transactions is not a problem in this path, so use Clear.
> ```

The poison mode doesn't matter for a committed intent. 

---

*[pkg/storage/intent_resolver.go, line 109 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_d0hGj5AWcWqaYULV:-Ku_d0hGj5AWcWqaYULW:b4fn0ce) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e923a32b79464015ff5bf767d17dc6a1c192d50/pkg/storage/intent_resolver.go#L109)):*
> ```Go
> 
> 	if err := ir.resolveIntents(ctx, resolveIntents,
> 		ResolveOptions{Wait: false, Poison: pushType == roachpb.PUSH_ABORT}); err != nil {
> ```

Now that I'm paging all of this in, this (the old code) looks incorrect to me. Suppose two transactions call processWriteIntentError for the same intents, one with PUSH_TIMESTAMP and one with PUSH_ABORT.  The PUSH_ABORT's PushTxn completes first, then the PUSH_TIMESTAMP's PushTxn sees that the txn was already aborted. Both will then try to resolve the intent, but only one will try to poison. If the PUSH_TIMESTAMP resolve happens first, there will be a window during which the intent is missing but the abort cache has not been written. 

It's not clear to me when it's ever safe to resolve an intent from an aborted transaction without setting the abort cache (Except after EndTransaction, when we can clear it), since it appears to depend on knowledge from the operation responsible for the abort. 

I think even in the GC queue we have to set the abort cache for intents we resolve (and then we'll have to clean up those abort cache entries on a future GC queue iteration... ouch!). I might be wrong on this point, though. 

---

*[pkg/storage/replica_command.go, line 1853 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku__6bpYvVws_IL-Zrw:-Ku__6bpYvVws_IL-Zrx:b-72n7pv) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/storage/replica_command.go#L1853)):*
> ```Go
> ) (EvalResult, error) {
> 	args := cArgs.Args.(*roachpb.ResolveIntentRequest)
> 	args.Poison = fromDeprecatedPoison(args.Poison, args.DeprecatedPoison)
> ```

This mutates the input. I can't remember whether that's safe or not but it seems dirty; let's use a local variable instead. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635)*

tschottdorf:
Ready for another look.

---

Review status: 2 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/roachpb/api.proto, line 627 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_eImWYn4YNJ7AslS2:-KutqmPGkntwfy0EqIhU:bbhkjab) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/roachpb/api.proto#L627)):*

It's safe to change bools to enums in protobufs if the values line up (going from old to new, true and false are mapped to 1 and 0. going the other direction, zero is false and any non-zero is true). I'm not sure if that works out in this case or not. 
</blockquote></details>

Do you think it's worth worrying about? I'd hate to shoot myself in the foot here.

---

*[pkg/storage/gc_queue.go, line 423 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_ej7g-oEEHGSnK0d1:-KutqthHFKScoMnBgabO:b1noo3) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/storage/gc_queue.go#L423)):*

The poison mode doesn't matter for a committed intent. 
</blockquote></details>

Good call, updated.

---

*[pkg/storage/intent_resolver.go, line 109 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_d0hGj5AWcWqaYULV:-Kutra7ZNyXwfSXfWhRp:bwli0qc) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e923a32b79464015ff5bf767d17dc6a1c192d50/pkg/storage/intent_resolver.go#L109)):*

Now that I'm paging all of this in, this (the old code) looks incorrect to me. Suppose two transactions call processWriteIntentError for the same intents, one with PUSH_TIMESTAMP and one with PUSH_ABORT.  The PUSH_ABORT's PushTxn completes first, then the PUSH_TIMESTAMP's PushTxn sees that the txn was already aborted. Both will then try to resolve the intent, but only one will try to poison. If the PUSH_TIMESTAMP resolve happens first, there will be a window during which the intent is missing but the abort cache has not been written. 

It's not clear to me when it's ever safe to resolve an intent from an aborted transaction without setting the abort cache (Except after EndTransaction, when we can clear it), since it appears to depend on knowledge from the operation responsible for the abort. 

I think even in the GC queue we have to set the abort cache for intents we resolve (and then we'll have to clean up those abort cache entries on a future GC queue iteration... ouch!). I might be wrong on this point, though. 
</blockquote></details>

Yikes, good call on the first one. I expanded the comment.

Re: the GC queue, I think we're in a better position there because we only touch aborted transactions for which we know that their most recent heartbeat is a (large) [multiple of the heartbeat timeout](https://github.com/cockroachdb/cockroach/blob/fbcd8ca7350a79e91f635fe59f8c5f74bde79eba/pkg/storage/gc_queue.go#L57). That is, we assume that the client running the transaction is gone (and even if it isn't, it would not reach the KV store on its next request).

---

*[pkg/storage/replica_command.go, line 1853 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku__6bpYvVws_IL-Zrw:-Kutsjt1DC9bIK4Fzs0I:b-fip7of) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/storage/replica_command.go#L1853)):*

This mutates the input. I can't remember whether that's safe or not but it seems dirty; let's use a local variable instead. 
</blockquote></details>

Yeah, this is pretty dirty. Somehow didn't realize this was a pointer. Changed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-:-KuuoWe1-0fSVNrb4I29:b467oqh)*

tschottdorf:
@bdarnell friendly ping -- looks like this fell off your radar. I'd like to include this in a 1.1 release, though at this point it would probably be 1.1.1 instead of 1.1.

bdarnell:
Reviewed 7 of 7 files at r2.

---

*[pkg/roachpb/api.proto, line 627 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_eImWYn4YNJ7AslS2:-KvJmL8fkLi0C5DohfV_:b-dr53sl) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/roachpb/api.proto#L627)):*

Do you think it's worth worrying about? I'd hate to shoot myself in the foot here.
</blockquote></details>

Well, it would be a change we could make all at once instead of going through a deprecation cycle. But probably not worth figuring out the subtlety and making sure it's safe.

Nit: this line has two semicolons.

---

*[pkg/storage/gc_queue.go, line 423 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_ej7g-oEEHGSnK0d1:-KvNoFTTTc68C3dO9Nfv:b-1i4gcf) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3864c9ef9e30d41a78e7e13ac8a458f9d9aec66/pkg/storage/gc_queue.go#L423)):*

Good call, updated.
</blockquote></details>

I think it would be more clear to use PoisonType_Noop, since nothing will happen regardless.

---

*[pkg/storage/gc_queue.go, line 411 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-KvNoVdFaxE2CGxvLRa1:-KvNoVdFaxE2CGxvLRa2:b-mi9lll) ([raw file](https://github.com/cockroachdb/cockroach/blob/66507ae5daf424d4e340cb84c6e7b73c3a647386/pkg/storage/gc_queue.go#L411)):*
> ```Go
> 				infoMu.Unlock() // intentional
> 				defer infoMu.Lock()
> 				// This transaction has committed, so poisoning is not relevant.
> ```

I think you've gotten your comments mixed up. This is the `ABORTED` branch. 

---

*[pkg/storage/intent_resolver.go, line 109 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_d0hGj5AWcWqaYULV:-KvNuFOTL7nTmTjUzE2E:b-6u71pi) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e923a32b79464015ff5bf767d17dc6a1c192d50/pkg/storage/intent_resolver.go#L109)):*
> Yikes, good call on the first one. I expanded the comment.

But the new code is still incorrect. The old version was incorrect in either ordering for a short time if the TIMESTAMP resolve came first (leaving the abort cache temporarily cleared until the ABORT resolve set it), or for a long time if the ABORT resolve came first and then had its abort cache entry cleared by the TIMESTAMP resolve. The new version only has the former bug, but it's still a problem. The transaction could sneak in in between the TIMESTAMP resolve and the ABORT resolve and silently fail to see the results of its own writes. 

> Re: the GC queue, I think we're in a better position there because we only touch aborted transactions for which we know that their most recent heartbeat is a (large) multiple of the heartbeat timeout. 

That's for cleaning up transaction records, not resolving intents. The GC queue will resolve intents based on the separate intentAgeThreshold. That threshold is currently higher than the transaction threshold (!) but it doesn't have to be (because in any case it will validate the transaction's status before resolving the intent). 

I think we can almost get rid of the poison flag in ResolveIntentRequest completely: in most cases the correct poison behavior is determined from the TxnMeta contained in the request (even the GC queue's "very old transaction" optimization can be obtained from looking at TxnMeta.Timestamp) instead of from extra knowledge that the caller supplies. 

The exception is for EndTransaction: here, as an optimization, we avoid setting the abort cache for intents we resolve because we know that the coordinator is responsible for this abort (and no future requests will make it back to the client).

So here's my proposal for fixing the existing bug here and optimizing command queue contention as much as we safely can:
- Keep the existing poison boolean; we don't need three states
- All ResolveIntent callers except the EndTransaction branch of processIntents pass poison=true
- declareKeysResolveIntent gets smarter, and only declares a write to the abort cache if one will occur
- evalResolveIntent gains a "very old transaction" optimization and skips the write to the abort cache if TxnMeta.Timestamp is too old

---

*[pkg/storage/replica_command.go, line 1830 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-KvNpu-lUpo_sUMAPq7k:-KvNpu-lUpo_sUMAPq7l:b-x86ihm) ([raw file](https://github.com/cockroachdb/cockroach/blob/66507ae5daf424d4e340cb84c6e7b73c3a647386/pkg/storage/replica_command.go#L1830)):*
> ```Go
> 
> func writesAbortCache(intentStatus roachpb.TransactionStatus, poisonType roachpb.PoisonType) bool {
> 	return (intentStatus == roachpb.ABORTED && poisonType == roachpb.PoisonType_Do) ||
> ```

This doesn't match the implementation, which writes to the abort cache if `intentStatus == roachpb.ABORTED && poisonType != PoisonType_Noop`. 

---

*[pkg/storage/replica_command.go, line 1923 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-KvNqICkY450kGn0HsU0:-KvNqICkY450kGn0HsU1:bwwfys5) ([raw file](https://github.com/cockroachdb/cockroach/blob/66507ae5daf424d4e340cb84c6e7b73c3a647386/pkg/storage/replica_command.go#L1923)):*
> ```Go
> 		return EvalResult{}, err
> 	}
> 	if intent.Status == roachpb.ABORTED {
> ```

Making this `if wrtiesAbortCache(...)` would help centralize the knowledge of this condition. (and then maybe we would remove the `Noop` case from `setAbortCache`)

---

*[pkg/storage/replica_command_test.go, line 48 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-KvNpT2yDIYFEzxzYv6b:-KvNpT2yDIYFEzxzYv6c:btyf9gz) ([raw file](https://github.com/cockroachdb/cockroach/blob/66507ae5daf424d4e340cb84c6e7b73c3a647386/pkg/storage/replica_command_test.go#L48)):*
> ```Go
> 			status:   roachpb.ABORTED,
> 			poison:   roachpb.PoisonType_Clear,
> 			expSpans: []string{"1 0: {b-c}", `1 1: /{Local/RangeID/99/r/AbortCache/"00000000-0000-0000-0000-000000000000"-Min}`},
> ```

Set a non-zero transaction id in the request just so we can make sure it's getting plumbed through correctly. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635)*

tschottdorf:

---

*[pkg/storage/intent_resolver.go, line 109 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_d0hGj5AWcWqaYULV:-KvbooQYq3UsKz5U6ARl:blzv8xm) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e923a32b79464015ff5bf767d17dc6a1c192d50/pkg/storage/intent_resolver.go#L109)):*
> The new version only has the former bug, but it's still a problem. The transaction could sneak in in between the TIMESTAMP resolve and the ABORT resolve and silently fail to see the results of its own writes.

What am I missing? If a timestamp resolve comes first it won't poison but that's fine since the intent is still there, and then the abort would remove the intent and poison. The other way around you poison+remove first, and then leave it alone after. Your comment makes it sound like a timestamp resolve would remove the intent, too.

Your suggestions below sound interesting, I'll take a look but it sounds like it should work out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635)*

bdarnell:

---

*[pkg/storage/intent_resolver.go, line 109 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_d0hGj5AWcWqaYULV:-KvcBhF3bCR-Se1MOGq1:be88gay) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e923a32b79464015ff5bf767d17dc6a1c192d50/pkg/storage/intent_resolver.go#L109)):*

> The new version only has the former bug, but it's still a problem. The transaction could sneak in in between the TIMESTAMP resolve and the ABORT resolve and silently fail to see the results of its own writes.

What am I missing? If a timestamp resolve comes first it won't poison but that's fine since the intent is still there, and then the abort would remove the intent and poison. The other way around you poison+remove first, and then leave it alone after. Your comment makes it sound like a timestamp resolve would remove the intent, too.

Your suggestions below sound interesting, I'll take a look but it sounds like it should work out.
</blockquote></details>

Our terminology is getting confusing here. There's not actually any such thing as a timestamp resolve. There is a timestamp *push*, and if that push discovers that the transaction is aborted anyway, then we resolve the intent. So the buggy sequence is:

1. Transaction A (in snapshot isolation) writes an intent
1. Transaction B attempts a write which hits this intent. It pushes (aborts) transaction A
1. Transaction C attempts a read which hits txn A's intent. It tries a push (timestamp), but the push response tells it that txn A is aborted
1. Txn C's intent resolver resolves the intent. The original pushType no longer matters here; what matters is that the txn was aborted and the intent will be deleted. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635)*

tschottdorf:

---

*[pkg/storage/intent_resolver.go, line 109 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635#-Ku_d0hGj5AWcWqaYULV:-Kvh6z3Pl0p0wca_3pUt:b-km39hs) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e923a32b79464015ff5bf767d17dc6a1c192d50/pkg/storage/intent_resolver.go#L109)):*
> So here's my proposal for fixing the existing bug here and optimizing command queue contention as much as we safely can:

in `ResolveIntent`, we don't have enough of the proto to decide this right off the bat. I think we want something like

`clientKnowsOfAbort := txnProtoAsSeenDuringPush.LastActive() + 2*HeartbeatInterval < ResolveBatchTimestamp`

but we only have `txn.Timestamp` which is largely irrelevant to computing `LastActive()`.

What do you think about the following:

- replace `Poison` with `LastActive hlc.Timestamp` (or nullable, we'll see)
- `LastActive` is forwarded to `ResolveBatchTimestamp` if `LastActive` is zero/nil (for the migration)
- `ResolveIntent` uses `WillRemoveIntent && (LastActive + 2*HeartbeatInterval < ResolveBatchTimestamp)` to decide whether to populate the abort cache

Note that then, we never explicitly clear the abort cache (except during GC, where these entries are thrown away based on a similar criterion), but that should be OK because the only situation in which we can do that is when an `EndTransaction(commit=false)` happens, and there we only meaningfully clear out an abort cache on external ranges for which the intents were supplied. So if, say, we had 10 intents on range 2 (and the txn is on range 1), we currently end up sending a batch of 10 `ResolveIntent`s that all overlap on the abort cache key because they all want to clear the span, potentially blocking other activity on these keys for much longer than needed. Seems better to just let the GC queue deal with it instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18635)*

bdarnell:
SGTM

PR #18642, opened 2017-09-20
storage: short-circuit GC queue when context expires

tbg:
Fallout from https://github.com/cockroachdb/cockroach/issues/18199 and corresponding
testing in https://github.com/cockroachdb/cockroach/issues/15997.

When the context is expired, there is no point in shooting off another gazillion requests.

### DIFF
--- a/pkg/storage/gc_queue.go
+++ b/pkg/storage/gc_queue.go
@@ -405,14 +405,19 @@ func processLocalKeyRange(
 			// but instead by the coordinator - those will not have any intents
 			// persisted, though they still might exist in the system.
 			infoMu.TransactionSpanGCAborted++
-			func() {
+			if err := func() error {
 				infoMu.Unlock() // intentional
 				defer infoMu.Lock()
-				if err := resolveIntents(roachpb.AsIntents(txn.Intents, &txn),
-					ResolveOptions{Wait: true, Poison: false}); err != nil {
-					log.Warningf(ctx, "failed to resolve intents of aborted txn on gc: %s", err)
+				return resolveIntents(roachpb.AsIntents(txn.Intents, &txn),
+					ResolveOptions{Wait: true, Poison: false})
+			}(); err != nil {
+				// Ignore above error, but if context is expired, no point in keeping going.
+				if ctx.Err() != nil {
+					return errors.Wrap(err, "context timed out during local key range processing after")
 				}
-			}()
+				log.Warningf(ctx, "failed to resolve intents of aborted txn on gc (removing anyway): %s", err)
+				// Keep going.
+			}
 		case roachpb.COMMITTED:
 			// It's committed, so it doesn't need a push but we can only
 			// GC it after its intents are resolved.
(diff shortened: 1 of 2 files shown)


bdarnell:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[pkg/storage/gc_queue.go, line 859 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642#-KuWUWSbk58j4gNo39tn:-KuWUWSbk58j4gNo39to:bo1e9xz) ([raw file](https://github.com/cockroachdb/cockroach/blob/14eb792aa5dcaec2a7180d781e6e813aed3c7d54/pkg/storage/gc_queue.go#L859)):*
> ```Go
> 	// succeeded in pushing a lot of transactions, the next time around we should have
> 	// less work here and manage to get to the intents, but it would be better to let
> 	// another goroutine push intents for transactions we've handled so that work that
> ```

s/push/resolve/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642#-:-KuWUaNIwo-qG4ff0FnR:bnfp4nl)*

tschottdorf:

---

*[pkg/storage/gc_queue.go, line 859 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642#-KuWUWSbk58j4gNo39tn:-Ku_0Hh6YYg-cRzUrEDZ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/14eb792aa5dcaec2a7180d781e6e813aed3c7d54/pkg/storage/gc_queue.go#L859)):*

s/push/resolve/
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642)*

nvanbenschoten:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/storage/gc_queue.go, line 415 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642#-Ku_M06AC4hoEyvKK2LD:-Ku_M06AC4hoEyvKK2LE:b-7huegl) ([raw file](https://github.com/cockroachdb/cockroach/blob/71ccf1c3b94f6a54b99f096dc2d5850fe361b80c/pkg/storage/gc_queue.go#L415)):*
> ```Go
> 			}(); err != nil {
> 				// Ignore above error, but if context is expired, no point in keeping going.
> 				if ctx.Err() != nil {
> ```

`ctx.Err` can return different types of errors, depending on whether the context timed out or was canceled. Why don't we return `ctx.Err`? Is the idea that `err` probably already is this `ctx.Err`? Same question below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642)*

tschottdorf:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/storage/gc_queue.go, line 415 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642#-Ku_M06AC4hoEyvKK2LD:-Ku_R1RHUhrZ6QAOnVO9:bp32i2m) ([raw file](https://github.com/cockroachdb/cockroach/blob/71ccf1c3b94f6a54b99f096dc2d5850fe361b80c/pkg/storage/gc_queue.go#L415)):*

`ctx.Err` can return different types of errors, depending on whether the context timed out or was canceled. Why don't we return `ctx.Err`? Is the idea that `err` probably already is this `ctx.Err`? Same question below.
</blockquote></details>

Yes, and even if it isn't, I think the last error is probably more interesting than a context expired/canceled error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18642)*

PR #18643, opened 2017-09-20
storage: do not gather up too many intents/keys

tbg:
Fallout from #18199 and corresponding testing in #15997. I think it'll be nontrivial to max out
these budgets in practice, but I can definitely do it in intentionally evil tests, and it's good to
know that there is some rudimentary form of memory accounting in this queue.

### DIFF
--- a/pkg/storage/gc_queue.go
+++ b/pkg/storage/gc_queue.go
@@ -751,6 +751,13 @@ func RunGC(
 	txnMap := map[uuid.UUID]*roachpb.Transaction{}
 	intentSpanMap := map[uuid.UUID][]roachpb.Span{}
 
+	const (
+		intentBudgetBytes = 30 * 1 << 20 // 30mb
+		keyBudgetBytes    = 30 * 1 << 20 // 30mb
+	)
+
+	var intentBytes int
+	var keyBytes int
 	// processKeysAndValues is invoked with each key and its set of
 	// values. Intents older than the intent age threshold are sent for
 	// resolution and values after the MVCC metadata, and possible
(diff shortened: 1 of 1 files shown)


pkg/storage/gc_queue.go
```diff
@@ -767,30 +774,50 @@ func RunGC(
...
 					}
 					// With an active intent, GC ignores MVCC metadata & intent value.
 					startIdx = 2
 				}
+				if keyBytes > keyBudgetBytes {
+					// Stop considering keys for GC if we've already crashed the budget.
+					//
+					// TODO(tschottdorf): we already send multiple GCRequests if we have too many keys,
+					// but we should be dispatching them as we scan to avoid holding the whole chunk of
+					// keys in memory. Large Ranges should be the exception, but they can definitely be
```
nvanbenschoten: nit: `s/Ranges/ranges/`

pkg/storage/gc_queue.go
```diff
@@ -751,6 +751,13 @@ func RunGC(
 	txnMap := map[uuid.UUID]*roachpb.Transaction{}
 	intentSpanMap := map[uuid.UUID][]roachpb.Span{}
 
+	const (
+		intentBudgetBytes = 30 * 1 << 20 // 30mb
```
nvanbenschoten: `intentBytesBudget` and `keyBytesBudget` sound better to me. Take it or leave it.

pkg/storage/gc_queue.go
```diff
@@ -767,30 +774,50 @@ func RunGC(
...
-					// multiple requests in the event that more than X keys
-					// are added to the request.
 					gcKeys = append(gcKeys, roachpb.GCRequest_GCKey{Key: expBaseKey, Timestamp: gcTS})
+					keyBytes += len(expBaseKey)
 				}
 			}
 		}
 	}
 
+	if intentBytes > intentBudgetBytes {
```
nvanbenschoten: Will this... work? We haven't called `processKeysAndValues` yet.

pkg/storage/gc_queue.go
```diff
@@ -767,30 +774,50 @@ func RunGC(
...
 				if meta.Txn != nil {
 					// Keep track of intent to resolve if older than the intent
-					// expiration threshold.
-					if meta.Timestamp.Less(intentExp) {
+					// expiration threshold. Don't do this if we've already collected
+					// a lot of potential intents (note that what we collect here is
+					// extremely likely to be collectable unless the user is running
+					// very large numbers of very long transactions). But the same
+					// consideration as the TODO below holds: we should be handling
+					// the chunk and resetting the budget.
```
nvanbenschoten: `s/handling the chunks/handling the chunks as we scan`

pkg/storage/gc_queue.go
bdarnell: Yeah, I think this needs a test.

tschottdorf:
Obviated by @spencerkimball's improvements.

Issue #18646, opened 2017-09-20
storage: investigate busy loop in split queue


tschottdorf:
Both #18199 and #15997 observed the split queue showing up in profiles/participating in obvious contention.

In #15997 we were writing a range chock full of illegal split keys. The split operation should back off more and more (owing to the doubling of the in-memory split target size), but this perhaps doesn't happen. Also, reading an engine snapshot in a tight loop appears to have interfered pretty severely with Raft handling in my testing, leading to up to 50s instances of `handleRaftReady`.

petermattis:
We don't double the split target size if an error is returned from `EnsureSafeSplitKey`. Perhaps we should. 

Removing the engine snapshot from `MVCCFindSplitKey` is trivial. There is also some other low-hanging performance changes there which I'll send out a PR for. But it will be worthwhile to figure out what is causing the long `handleRaftReady`. If it is the engine snapshot, then I clearly don't understand something.

PR #18649, opened 2017-09-20
storage/engine: optimize MVCCFindSplitKey

petermattis:
Use an explicit iterator instead of Reader.Iterate() in order to reduce
allocations. Specifically, we can avoid 2 allocations per key/value.

Do not create a snapshot when calling MVCCFindSplitKey as internally it
uses a single iterator which has an implicit snapshot.

Remove the unused rangeID parameter to MVCCFindSplitKey. This parameter
was only being used to lookup the range stats to log them under V(2).

```
name                                     old time/op    new time/op    delta
MVCCFindSplitKey_RocksDB/valueSize=32-8     341ms ±23%     269ms ± 8%   -21.29%  (p=0.000 n=10+10)

name                                     old speed      new speed      delta
MVCCFindSplitKey_RocksDB/valueSize=32-8   198MB/s ±19%   250MB/s ± 8%   +26.34%  (p=0.000 n=10+10)

name                                     old alloc/op   new alloc/op   delta
MVCCFindSplitKey_RocksDB/valueSize=32-8    36.6MB ± 0%     0.0MB ± 0%  -100.00%  (p=0.000 n=9+10)

name                                     old allocs/op  new allocs/op  delta
MVCCFindSplitKey_RocksDB/valueSize=32-8     1.16M ± 0%     0.00M ± 0%  -100.00%  (p=0.000 n=9+10)
```

See #18646

### DIFF
--- a/pkg/storage/engine/bench_rocksdb_test.go
+++ b/pkg/storage/engine/bench_rocksdb_test.go
@@ -80,6 +80,14 @@ func BenchmarkMVCCComputeStats_RocksDB(b *testing.B) {
 	}
 }
 
+func BenchmarkMVCCFindSplitKey_RocksDB(b *testing.B) {
+	for _, valueSize := range []int{32} {
+		b.Run(fmt.Sprintf("valueSize=%d", valueSize), func(b *testing.B) {
+			runMVCCFindSplitKey(setupMVCCRocksDB, valueSize, b)
+		})
+	}
+}
+
 func BenchmarkIterOnBatch_RocksDB(b *testing.B) {
 	for _, writes := range []int{10, 100, 1000, 10000} {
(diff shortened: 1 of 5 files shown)


petermattis:
Note that there is another 30-50% performance boost to be had by moving `MVCCFindSplitKey` into C++ based on the performance of `MVCCComputeStats`:

```
name                                      time/op
MVCCComputeStats_RocksDB/valueSize=32-8      180ms ± 4%

name                                      speed
MVCCComputeStats_RocksDB/valueSize=32-8    373MB/s ± 4%
```

bdarnell:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18649#-:-KuWWeYpW7evi8MnyBdC:bnfp4nl)*

PR #18663, opened 2017-09-21
cli: Fix an IPAddr error in a flaky test

lgo:
There was an error when in a skipped test with how the IPAddr is encoded when randomly generated. This is fixed but is not the original reason for why the test is skipped.

cc: @justinj


knz (approved):

justinj:
:lgtm: assuming you removed the `t.Skip` above and made sure that the test passes at least some of the time now!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18663#-:-KuZztwMwvxqBLh2Cldp:b-rh76as)*

Issue #17289, opened 2017-07-28
sql: perf difference between implicit and explicit transactions


petermattis:
In writing a new load generator I noticed a significant performance difference between using implicit and explicit transactions. I'm guessing that the explicit transaction is not hitting the 1PC code path while I think it should.

In both case the transaction contains a single `UPDATE` statement. The implicit version:

```
UPDATE test.rmw SET data='%s' WHERE id='%s'
```

And the explicit version:

```
BEGIN; UPDATE test.rmw SET data='%s' WHERE id='%s'; COMMIT;
```

The schema is:

```
CREATE TABLE IF NOT EXISTS test.rmw (
  id STRING PRIMARY KEY,
  data STRING
)
```

With the implicit transactions, a single worker is able to perform 850 ops/sec. With a the explicit transaction I see only 450 ops/sec. 

@andreimatei You tend to play around in this area of the code. Please reassign if there is someone better to look at this.

andreimatei:
Hmm well we have no COMMIT elision at the sql level, so there's can't be a 1PC here. I guess you're saying there should be.

petermattis:
I thought there was equivalent performance for these two queries at one point. Perhaps I'm misremembering. Eliding the COMMIT at the sql level isn't necessary, just looking ahead past the currently executing statement to see if the next is a COMMIT.

arjunravinarayan:
Assigned to @richardwu for investigation.

richardwu:
cc: @andreimatei who is refactoring the executor.

jordanlewis:
Closing due to staleness.

andreimatei:
Well but this issue is very valid, and actionable, and everything. If anything, the fact that it's old makes it more import, not less... Why close?

jordanlewis:
Nobody's encountered or tested this scenario since 2017. Do you think it's still valid? I don't see the value in keeping something like this open if there's no movement or planned work on it. That being said, happy to keep open if you find it useful.

andreimatei:
I think it's still valid, yeah. However it looks like we tried to do something about it in #18664 and it proved to not be entirely trivial.

andreimatei:
Folded into #4703

PR #18664, opened 2017-09-21
sql: elide 1-statement inserts/updates in transaction blocks contained in single batch

richardwu:
Fixes https://github.com/cockroachdb/cockroach/issues/17289

## Context

We have the notion of implicit and explicit transactions when the executor runs a batch of statements. Implicit transactions are singular statements that are not within a transaction block thus they are run as a transaction by itself. For example:
```
UPDATE foo SET data = 'bar' WHERE id = 1
```

Explicit transactions are transactions that have a `BEGIN (TRANSACTION)` and `END/COMMIT/ROLLBACK (TRANSACTION)`. For example, the equivalent explicit transaction for the above implicit transaction is:
```
BEGIN; UPDATE foo SET data = 'bar' WHERE id = 1; COMMIT;
```

## Problem

Implicit transactions are allowed to "auto commit" or 1-Phase-Commit (1PC); that is, instead of creating transaction intents when the write statement then sending an second commit statement (this has 2 phases or 2PC) it simply commits the statement (transaction) in 1-phase.

We do not currently recognize 1-statement transaction blocks as a transaction that can 1PC. This incurs a performance overhead.

## Fix

For a given "batch" of statements (in the perspective of the executor), we check whether or not a `BEGIN TRANSACTION` block contains one and only one write statement (currently `INSERT` or `UPDATE`) followed by a `COMMIT TRANSACTION` statement. If this is satisfied, the `BEGIN` and `COMMIT` are elided and `autoCommit` is set to true. In essence, 1-write-statement transaction blocks are converted into implicit transactions.

## Caveats

1. We whitelist the single-statements to either `INSERT`s or `UPDATE`s. This is because there exists transaction dependent statements (e.g. `SET TRANSACTION LEVEL xxx`) that most likely depend on transaction state set by `BEGIN` and `END`. Non-write statements like `SELECT` are more difficult to deal with since they may have sub-queries tht are dependent on `BEGIN` and `END`.
2. We cannot currently detect 1-statement transaction blocks between batches. For example, the following will not be elided since they are executed as separately 3 separate batches:
   ```
   BEGIN;

   UPDATE foo SET data = 'bar' WHERE id = 1;

   COMMIT;
   ```
   This brings up the point of buffering and combining batches whenever explicit transaction blocks 
   are involved, which is a separate issue.
3. We can probably do more intricate eliding of 2PC. Within an explicit transaction block, if the last statement in the block (before `COMMIT`) is to be committed and there has been no write statements  since `BEGIN`, then we can set `autoCommit` to true for the given statement. For example:
```
BEGIN; SELECT 1; SELECT 2; UPDATE foo SET data = 'bar' WHERE id = 1; COMMIT;
```
`SELECT 1;` and `SELECT 2` are non-write statements, thus the final `COMMIT` should be elided. This will obviously require some non-trivial refactor to maintain atomicity of transaction blocks while permitting auto-commit/1PC.

cc: @petermattis

### DIFF
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -799,8 +799,60 @@ func (e *Executor) execParsed(
 		// transaction (implicit txn or explicit txn). We do the corresponding state
 		// reset.
 		if !inTxn {
-			// Detect implicit transactions - they need to be autocommitted.
-			if _, isBegin := stmts[0].AST.(*parser.BeginTransaction); !isBegin {
+			_, isBegin := stmts[0].AST.(*parser.BeginTransaction)
+
+			elide := false
+			if isBegin {
+				// Explicit transaction (BEGIN statement) - if
+				// the next two statements are (in order):
+				// 1. UPDATE/INSERT/UPSERT/whitelist statement
+				// 2. COMMIT statement
+				// Then we elide the BEGIN and COMMIT
+				// statements and execute #1 as an implicit
+				// transaction.
+				// #1 is whitelisted since there exists
+				// transaction-dependent statements (e.g.
+				// SHOW TRANSACTION STATUS) that may depend on
+				// our explicit BEGIN/END block.
+				// Whitelisting is chosen over blacklisting
+				// since the introduction of new transaction-
+				// dependent statements may crash the system.
+
+				// There must be at least 3 statements in our
+				// current batch for this elision.
+				if len(stmts) >= 3 {
+					if _, isCommit := stmts[2].AST.(*parser.CommitTransaction); isCommit {
+						// Permit only certain types
+						// of statements to be converted
+						// to an implicit statement.
+						// Note non-commit statements
+						// are irrelvant since they
+						// do not take advantage of
+						// auto-commit.
+						ast := stmts[1].AST
+						if traceStmt, ok := ast.(*parser.ShowTrace); ok {
+							ast = traceStmt.Statement
+						}
+						switch ast.(type) {
+						case *parser.Update:
+							elide = true
+						case *parser.Insert:
+							elide = true
+						}
+
+						// Do not execute BEGIN and
+						// COMMIT statement.
+						if elide {
+							stmts = append(stmts[1:2], stmts[3:]...)
+							stmtsToExec = stmts
+						}
+					}
+				}
+			}
+
+			if !isBegin || elide {
+				// Implicit or elided explicit transactions -
+				// they need to be autocommitted.
 				autoCommit = true
 				stmtsToExec = stmtsToExec[:1]
 				// Check for AS OF SYSTEM TIME. If it is present but not detected here,
--- a/pkg/sql/logictest/testdata/logic_test/insert
+++ b/pkg/sql/logictest/testdata/logic_test/insert
@@ -676,3 +676,37 @@ INSERT INTO bytes_t SELECT * FROM string_t
 
 query error value type bytes doesn't match type STRING of column "s"
 INSERT INTO string_t SELECT * FROM bytes_t
+
+# Test that only single-insert-statement, explicit transaction batches are
+# elided and executed as implicit transactions with 1PC.
+
+statement ok
+CREATE TABLE ab (
+  a STRING PRIMARY KEY,
+  b STRING
+)
+
+# 1 PC with one insert statement (final commit elided).
+query T
+SET TRACING=on;
+BEGIN; INSERT INTO ab VALUES ('foo1', 'bar1'); COMMIT;
+SET TRACING=off;
+SELECT message FROM [SHOW TRACE FOR SESSION]
+  WHERE MESSAGE LIKE 'executing _/_%';
+----
+executing 1/1: INSERT INTO ab VALUES ('foo1', 'bar1')
+executing 1/1: SET tracing = off
+
+# 2PC since there exists more than one statement.
+query T
+SET TRACING=on;
+BEGIN; INSERT INTO ab VALUES ('foo2', 'bar2'); INSERT INTO ab VALUES ('foo3', 'bar3'); COMMIT;
+SET TRACING=off;
+SELECT message FROM [SHOW TRACE FOR SESSION]
+  WHERE MESSAGE LIKE 'executing _/_%';
+----
+executing 1/6: BEGIN TRANSACTION
+executing 2/6: INSERT INTO ab VALUES ('foo2', 'bar2')
+executing 3/6: INSERT INTO ab VALUES ('foo3', 'bar3')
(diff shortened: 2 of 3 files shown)


richardwu:

---

*[pkg/sql/logictest/testdata/logic_test/insert, line 711 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18664#-Ku_06w6DlyUSM2BK3-Y:-Ku_06w6DlyUSM2BK3-Z:bufbftf) ([raw file](https://github.com/cockroachdb/cockroach/blob/b959ffecf1d4c6949fe2c116e895e0f8e7f08a6a/pkg/sql/logictest/testdata/logic_test/insert#L711)):*
> ```
> executing 2/6: INSERT INTO ab VALUES ('foo2', 'bar2')
> executing 3/6: INSERT INTO ab VALUES ('foo3', 'bar3')
> executing 4/6: COMMIT TRANSACTION
> ```

I'm not too sure why the denominator is 6 here (even though there are only 4 statements). I've tested this in the `./cockroach sql` shell and it behaved correctly. Perhaps I'm doing something incorrectly here with `SET TRACING` and `SHOW TRACE`? Would appreciate your insight here @andreimatei 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18664)*

knz:
You should also whitelist `set transaction isolation level xxx` immediately after `begin`. This is often used by e.g. gosql and other drivers instead of `begin transaction isolation level xxx`

richardwu:
I can't seem to reproduce the non-trivial benchmark disparity between the two types of transactions described in #17289. My benchmark does show that explicit transactions are slightly slower (3783.865597094 ops/sec for implicit vs 3632.796890326 ops/sec), which is probably due to parsing two extra statements, but I get similar numbers on master vs this branch.

That being said, would be great to see if the disparity still exists @petermattis with your benchmark.

bdarnell:

---

*[pkg/sql/executor.go, line 839 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18664#-Ku_rwH_c2ShaDoy7XuN:-Ku_rwH_c2ShaDoy7XuO:b-sfj9cr) ([raw file](https://github.com/cockroachdb/cockroach/blob/b959ffecf1d4c6949fe2c116e895e0f8e7f08a6a/pkg/sql/executor.go#L839)):*
> ```Go
> 						case *parser.Update:
> 							elide = true
> 						case *parser.Insert:
> ```

DELETE should work here too, and we might as well whitelist SELECT even though it won't benefit as much. (should this be a blacklist instead of a whitelist? statements like SHOW TRANSACTION STATUS seem uncommon)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18664)*

richardwu:
So after digging around in the rabbit hole and almost getting something working, I realized that eliding statements can be an issue for ensuring that the number of results within `StatementResult` matches the number of actual statements executed.

Since in the current codepath we'll end up executing singular implicit statements (when there could be `n` explicit statements), we'll end up sending fewer results in our `StatementResult` than a given client expects.

petermattis:
Rather than eliding statements, the way I thought this would work is that `sql.Executor` would look-ahead in the list of statements being executed to see if the next statement was a `COMMIT` and then set the `autoCommit` flag. Specifically, for a query that contains the statements `BEGIN; INSERT INTO ...; COMMIT`, when executing the `INSERT` we'd notice that the next statement is a `COMMIT`, set `autoCommit = true` and then skip the `COMMIT`.

Cc @andreimatei who knows much more about the `sql.Executor` state machine and can correct all the errors I'm making.

andreimatei:
I was thinking similar things on my ride this morning, given the difficulties with elision that we've  been discussing. However, I think just setting some auto-commit flag somewhere will not do - at the minimum you need to do something to ensure that the next `COMMIT` does bork because there's nothing for it to `COMMIT` any more. So I think we should configure the statement before the commit to also transition to state `CommitWait` somehow. I'll read some code.

andreimatei:
Another problem is that `autoCommit` is currently computed at the level of a transaction, not at the level of a statement.
The line that tells a plan that it can auto-commit if it wants is this one, in `execStmtInOpenTxn()`:
https://github.com/cockroachdb/cockroach/blob/1d5c72d03ced1cb44c3f00dc8a656a5fb347f18b/pkg/sql/executor.go#L1744
Notice that the value comes from a `txnState.implicitTxn`, which is currently set at the level of a txn. I think we should plumb an explicit `autoCommit` flag to `execStmtInOpenTxn()` and teach the higher-level `runTxnAttempt()` to do the look-ahead for the `COMMIT` stmt. I think this means that the handling for autoCommit that we currently have in `runWithAutoRetry` should move one layer down to `runTxnAttempt`.

andreimatei:
Some more notes: this PR/issue is fundamentally about moving an `EndTransaction` from a statement to another: from a `COMMIT` to the previous statement. Given that evaluating the `EndTransaction` can encounter various errors, doing this move has consequences for the session state machine: 
Errors encountered by statements generally stop the execution of subsequent statements in the (statement) batch and leave the session in the `Aborted` state, requiring a `ROLLBACK` to get out of the error state (i.e. move to `NoTxn`). Except:
- Implicit transactions never leave the session in `Aborted`.
- A `COMMIT` statement never leaves the session in `Aborted`.

You've noticed how the move changes the behavior for statement batches because of the 2nd point above. Moving the error out of `COMMIT` will cause some batches to leave the session in `Aborted` whereas before they didn't. I'm not saying that this is necessarily illegal... but I wonder if we should do work in order to avoid it when possible. Currently a `BatchResponse` doesn't have info about which request in the batch of requests caused the error, but in theory I guess it could. And then it'd the responsibility of each `planNode` capable of doing autoCommit/1PC to recognize errors from the `EndTransaction` and report them separately to the `Executor`, which could then behave as if the error came from the execution of the `COMMIT`.
Does anyone have any thoughts on this? cc @bdarnell @jordanlewis @knz @mjibson 

I generally wonder how libraries/drivers/people deal with these different states that a connection can be in after an error encountered by a batch of statements depending on which particular statement triggered the error.

andreimatei:
I had a thought that perhaps moving all the errors from COMMIT to the preceding statement is what postgres and nature intended, as it'd remove the complications about what state a batch of statements left the session in (an error would always require the client to issue a ROLLBACK).
I've emailed a postgres mailing list asking for guidance on this, but didn't get any answers yet. I did however find an old message from another postgres list with a somewhat similar question where a pg dev remarked that the ANSI specification about these states is hard to parse...

richardwu:
Closing this PR since the approach is not quite admissible (see @andreimatei's comments regarding pushing commit responsibilities to preceding statement).

PR #18676, opened 2017-09-21
sql/pgwire,parser: render non-tz timestamp in its loc, set loc when c…

### LINKED ISSUE #16228: sql: casting timestamptz to timestamp not respecting timezone
**BUG REPORT**
Version:
```
Build Tag:    v1.0.1
Build Time:   2017/05/25 15:17:49
Distribution: CCL
Platform:     linux amd64
Go Version:   go1.8.1
C Compiler:   gcc 6.3.0
Build SHA-1:  50fd18dcf8da75a0f4980344a885fe6105d0cf23
Build Type:   release
```
Code:
```
SET TIME ZONE 'Asia/Kolkata';
CREATE TABLE a (
	t TIMESTAMP WITH TIME ZONE NOT NULL, 
	CONSTRAINT "primary" PRIMARY KEY (id ASC));
INSERT INTO a(t) VALUES (TIMESTAMPTZ '2017-01-01 00:00:00');
SELECT extract('year', t::TIMESTAMP) FROM a; -- Returns 2016
SELECT extract('year', t::DATE) FROM a; -- Returns 2017
```
Expected behaviour: Conversion of timestamptz to timestamp is expected to respect time zone as Postgres does that.

dt:
…asting

Updated logic test behaviors verified from postgres: when a timestamptz is cast to timestamp, the result renders a localized time, with no offset.

Fixes #16228.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/datetime
+++ b/pkg/sql/logictest/testdata/logic_test/datetime
@@ -471,7 +471,17 @@ SELECT '2015-08-25 05:45:45.53453'::timestamptz
 query T
 SELECT '2015-08-25 05:45:45-01:00'::timestamp
 ----
-2015-08-25 06:45:45 +0000 +0000
+2015-08-25 05:45:45 +0000 +0000
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamp::timestamptz
+----
+2015-08-25 05:45:45 +0100 +0100
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamptz::timestamp
+----
+2015-08-25 07:45:45 +0000 +0000
 
 query T
 SELECT '2015-08-25 05:45:45-01:00'::timestamptz
@@ -504,7 +514,7 @@ SELECT '2015-08-24 23:45:45.53453 UTC'::timestamptz
 query T
 SELECT '2015-08-24 23:45:45.53453-02:00'::timestamp
 ----
-2015-08-25 01:45:45.53453 +0000 +0000
+2015-08-24 23:45:45.53453 +0000 +0000
 
 query T
 SELECT '2015-08-24 23:45:45.53453-02:00'::timestamptz
@@ -519,12 +529,38 @@ SELECT '2015-08-24 23:45:45.53453-05:00'::timestamptz
 query T
 SELECT '2015-08-24 23:45:45.534 -02:00'::timestamp
 ----
-2015-08-25 01:45:45.534 +0000 +0000
+2015-08-24 23:45:45.534 +0000 +0000
 
 query T
 SELECT '2015-08-24 23:45:45.534 -02:00'::timestamptz
 ----
 2015-08-24 20:45:45.534 -0500 -0500
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamp::timestamptz
+----
+2015-08-25 05:45:45 -0500 -0500
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamptz::timestamp
+----
+2015-08-25 01:45:45 +0000 +0000
+
+# using Eastern instead of fixed -5 should handle DST.
+statement ok
+SET TIME ZONE 'America/New_York'
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamp::timestamptz
+----
+2015-08-25 05:45:45 -0400 -0400
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamptz::timestamp
+----
+2015-08-25 02:45:45 +0000 +0000
+
+
 
 statement error cannot find time zone "foobar": timezone data cannot be found
 SET TIME ZONE 'foobar'
--- a/pkg/sql/parser/eval.go
+++ b/pkg/sql/parser/eval.go
@@ -2510,7 +2510,7 @@ func (expr *CastExpr) Eval(ctx *EvalContext) (Datum, error) {
 		case *DTimestamp:
 			return d, nil
 		case *DTimestampTZ:
-			return MakeDTimestamp(d.Time, time.Microsecond), nil
+			return MakeDTimestamp(d.Time.In(ctx.GetLocation()), time.Microsecond), nil
 		}
 
 	case *TimestampTZColType:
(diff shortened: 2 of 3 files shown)


pkg/sql/logictest/testdata/logic_test/datetime
```diff
@@ -471,7 +471,12 @@ SELECT '2015-08-25 05:45:45.53453'::timestamptz
 query T
 SELECT '2015-08-25 05:45:45-01:00'::timestamp
 ----
-2015-08-25 06:45:45 +0000 +0000
+2015-08-25 05:45:45 +0000 +0000
+
+query T
+SELECT '2015-08-25 05:45:45-01:00'::timestamp::timestamptz
+----
+2015-08-25 07:45:45 +0100 +0100
```
dt: huh, this one does not match postgres:

```
david=# set time zone +1;
SET
david=# SELECT '2015-08-25 05:45:45-01:00'::timestamp::timestamptz;
      timestamptz
------------------------
 2015-08-25 05:45:45+01
(1 row)
```

justinj: what do we think is the correct behaviour? granted I am not quite the timestamp-knower that dan seems to have me pegged as now, but I find cockroach's behaviour in this logic test surprising, given the one above it. it seems like the only way this behaviour should happen is if the intermediate `timestamp` knows which time zone it was rendered from (despite not presenting that info) is that how `timestamp`s work?

dt: so one big difference in behavior from pg is this:
```
In a literal that has been determined to be timestamp without time zone, PostgreSQL will silently ignore any time zone indication. That is, the resulting value is derived from the date/time fields in the input value, and is not adjusted for time zone.
```

I'm very wary of that, but in an example like this, i'm not sure we'll be able to match their result without doing so.

The other is that when pg casts a timestamp to a timestamptz, it appears to just assume it was already in the session offset, i.e. the rendered hour doesn't change, but now has an offset... if you'd assumed the original `timestamp` had a 0 offset, that would mean the utc/unix time of the underlying time.Time has changed.

Huh. I don't like either of those.

dt: indeed, pg docs say:
> Conversions between timestamp without time zone and timestamp with time zone normally assume that the timestamp without time zone value should be taken or given as timezone local time. A different time zone can be specified for the conversion using AT TIME ZONE.

dt:

---

*[pkg/sql/logictest/testdata/logic_test/datetime, line 479 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-Ku_qL1HELQe0tkrA4v1-r1-479:-KuetOQfyKNPZsZPEXz-:b-fzqhed) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a652b9dcb7048f0c07fb131a479428489c184a1/pkg/sql/logictest/testdata/logic_test/datetime#L479)):*

indeed, pg docs say:
> Conversions between timestamp without time zone and timestamp with time zone normally assume that the timestamp without time zone value should be taken or given as timezone local time. A different time zone can be specified for the conversion using AT TIME ZONE.

</blockquote></details>

Okay, with change below to timestamp -> timestamptz casts, where we add the session offset to the time.Time but update the time.Time itself as well so that the clock values (hour, minute, second) stay the same, we now appear to match pg behavior.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676)*

justinj:
:lgtm: with a couple questions

---

---

*[pkg/sql/logictest/testdata/logic_test/datetime, line 479 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-Ku_qL1HELQe0tkrA4v1-r1-479:-KugS-Lnx8vEQ63lVIuc:b-n7nf31) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a652b9dcb7048f0c07fb131a479428489c184a1/pkg/sql/logictest/testdata/logic_test/datetime#L479)):*

Okay, with change below to timestamp -> timestamptz casts, where we add the session offset to the time.Time but update the time.Time itself as well so that the clock values (hour, minute, second) stay the same, we now appear to match pg behavior.
</blockquote></details>

Is there any concern about having a timestamp that lands in vs. outside of DST? Is it worth having a test case for that? maybe also a test for string->timestamptz->timestamp would be good just for completeness

---

*[pkg/sql/parser/eval.go, line 2527 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-KugWncvvJX2Hzs2Di2u:-KugWncvvJX2Hzs2Di2v:b-3e9142) ([raw file](https://github.com/cockroachdb/cockroach/blob/308caaae90caa989d83d9e21a3038d5f2ae1f53c/pkg/sql/parser/eval.go#L2527)):*
> ```Go
> 		case *DTimestamp:
> 			_, before := d.Time.Zone()
> 			d.Time = d.Time.In(ctx.GetLocation())
> ```

Is it safe to mutate datums returned from `Eval`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-:-KugW24_ZJolDTlnqniA:b-dh5gix)*

dt:
TFTR!

---

---

*[pkg/sql/logictest/testdata/logic_test/datetime, line 479 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-Ku_qL1HELQe0tkrA4v1-r1-479:-KutHekgv7Z2DIzdoqtP:bxrxq1q) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a652b9dcb7048f0c07fb131a479428489c184a1/pkg/sql/logictest/testdata/logic_test/datetime#L479)):*

Is there any concern about having a timestamp that lands in vs. outside of DST? Is it worth having a test case for that? maybe also a test for string->timestamptz->timestamp would be good just for completeness
</blockquote></details>

Good point.
Done.

---

*[pkg/sql/parser/eval.go, line 2527 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-KugWncvvJX2Hzs2Di2u:-KutHhMWgF2MSZg_Y1Ep:b8nivt6) ([raw file](https://github.com/cockroachdb/cockroach/blob/308caaae90caa989d83d9e21a3038d5f2ae1f53c/pkg/sql/parser/eval.go#L2527)):*

Is it safe to mutate datums returned from `Eval`?
</blockquote></details>

Hm, not sure, but either way easy to avoid. Nice catch!

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-:-KutHuWfmjeQwY32FZt0:b-814g8k)*

knz:
:lgtm:

---

Reviewed 1 of 3 files at r1, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18676#-:-Kuv1tuA3iRCjVIhqZHL:bnfp4nl)*

knz:
this needs a cherrypick too

Issue #18630, opened 2017-09-20
sqlccl: LOAD CSV needs progress


cuongdo:
When performing a large `LOAD CSV` operation, there's no indication of progress. That's a significant irritant when loading large amounts of data, such as `LOAD CSV` with TPC-H scale factor 100 data. Ideally, we'd report similar information that we do for other long-running jobs.

dt:
for the record, our current job progress is updated after the sampling to 1/3 done, then to 1 when we finish the SSTs.

dianasaur323:
also related to this larger meta issue: https://github.com/cockroachdb/cockroach/issues/17946, although that refers to more than just load csv.

PR #18677, opened 2017-09-21
sqlccl: add incremental progress to local csv import

madelynnblue:
This required adding a new Size method to ExportStorage, which works
for all storages except S3 (due to a client library limitation,
not S3).

Progress is divided into two 50% chunks. The first is for reading
CSVs into KVs. We use the new Size method to find the total size of
the CSV files, and track how many bytes we've read. If Size fails,
then we just report after reading each file. The second phase reports
the number of KVs written against the total number of KVs, which are
known by this time.

This means the before- and after-50% rates may differ, but that's ok
and may be as good as we can get, since it is unknown how long the
write SST phase will take, since it may depend on network speed.

Distributed will be in a future PR.

See #18630

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -166,8 +166,10 @@ func doLocalCSVTransform(
 	execCfg *sql.ExecutorConfig,
 ) (csvCount, kvCount, sstCount int64, err error) {
 	// Some channels are buffered because reads happen in bursts, so having lots
-	// of pre-computed data improves overall performance.
-	const chanSize = 10000
+	// of pre-computed data improves overall performance. If this value is too
+	// high, it will decrease the accuracy of the progress estimation because
+	// of the hidden buffered work to be done.
+	const chanSize = 1000
 
 	recordCh := make(chan csvRecord, chanSize)
 	kvCh := make(chan []roachpb.KeyValue, chanSize)
@@ -183,16 +185,36 @@ func doLocalCSVTransform(
 	}
 	defer es.Close()
 
-	group, gCtx := errgroup.WithContext(ctx)
-	group.Go(func() error {
-		defer close(recordCh)
-		var err error
-		csvCount, err = readCSV(gCtx, comma, comment, len(tableDesc.VisibleColumns()), dataFiles, recordCh)
-		if job != nil {
-			if err := job.Progressed(ctx, 1.0/3.0, jobs.Noop); err != nil {
+	var readProgressFn, writeProgressFn func(float32)
+	if job != nil {
+		// These consts determine how much of the total progress the read csv and
+		// write sst groups take overall. 50% each is an approximation but kind of
+		// accurate based on my testing.
+		const (
+			readPct  = 0.5
+			writePct = 1.0 - readPct
+		)
+		// Both read and write progress funcs register their progress as 50% of total progress.
+		readProgressFn = func(pct float32) {
+			if err := job.Progressed(ctx, pct*readPct, jobs.Noop); err != nil {
 				log.Warningf(ctx, "failed to update job progress: %s", err)
 			}
 		}
+		writeProgressFn = func(pct float32) {
+			if err := job.Progressed(ctx, readPct+pct*writePct, jobs.Noop); err != nil {
+				log.Warningf(ctx, "failed to update job progress: %s", err)
+			}
+		}
+	}
+
+	// The first group reads the CSVs, converts them into KVs, and writes all
+	// KVs into a single RocksDB instance.
+	group, gCtx := errgroup.WithContext(ctx)
+	store := engine.NewRocksDBMultiMap(tempEngine)
+	group.Go(func() error {
+		defer close(recordCh)
+		var err error
+		csvCount, err = readCSV(gCtx, comma, comment, len(tableDesc.VisibleColumns()), dataFiles, recordCh, readProgressFn)
 		return err
 	})
 	group.Go(func() error {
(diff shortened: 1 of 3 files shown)


dt (approved):

pkg/ccl/sqlccl/csv.go
```diff
@@ -338,7 +377,30 @@ func readCSV(
...
 	for _, dataFile := range dataFiles {
+		conf, err := storageccl.ExportStorageConfFromURI(dataFile)
+		if err != nil {
+			return 0, err
+		}
+		es, err := storageccl.MakeExportStorage(ctx, conf)
+		if err != nil {
+			return 0, err
+		}
+		sz, _ := es.Size(ctx, "")
```
dt: do we want to log the error? or potentially also log a warning explaining that we'll fall back in the size == 0 block below?

pkg/ccl/storageccl/export_storage.go
```diff
@@ -268,11 +283,22 @@ func (h *httpStorage) Delete(_ context.Context, basename string) error {
 	return err
 }
 
+func (h *httpStorage) Size(_ context.Context, basename string) (int64, error) {
+	resp, err := h.req("HEAD", basename, nil)
```
dt: Some servers might not implement this -- at least it isn't part of our current required API.
We potentially want to add a test where our mock http server errors or otherwise returns an unusable response, just to make sure we keep handling it gracefully.

dt:
Some HTTP servers may or may not support the HEAD verb as this api expects, so we should potentially include a test where our mock server errors or returns garbage results (unless we add the HEAD verb to our required semantics) to ensure we handle it reasonably gracefully when we don't have usable size info.

---

Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/ccl/sqlccl/csv.go, line 449 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-KuwSA6Y7HPL3pF49IDS:-KuwSA6Y7HPL3pF49IDT:boh4ez0) ([raw file](https://github.com/cockroachdb/cockroach/blob/9e625f3a5fe66ec0acca1650aa39f5e201980686/pkg/ccl/sqlccl/csv.go#L449)):*
> ```Go
> 						}
> 					}
> 					const FiftyMiB = 50 << 20
> ```

capitalized?

---

*[pkg/ccl/storageccl/export_storage.go, line 379 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-KuwRD4eASUMp59KpWo5:-KuwRD4eASUMp59KpWo6:bd06r8t) ([raw file](https://github.com/cockroachdb/cockroach/blob/9e625f3a5fe66ec0acca1650aa39f5e201980686/pkg/ccl/storageccl/export_storage.go#L379)):*
> ```Go
> 
> func (s *s3Storage) Size(_ context.Context, basename string) (int64, error) {
> 	// TODO(mjibson): https://github.com/rlmcpherson/s3gof3r/issues/124
> ```

yeah, we should potentially just switch to the official aws sdk in a follow up, if it now supports the features we need (streaming).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-:-KuwQdcXJqJuU1_CR_l_:b-ke8szh)*

mjibson:
Added a test for 500 responses on HEAD requests.

---

---

*[pkg/ccl/sqlccl/csv.go, line 392 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-Ku_ugb_G0In8WloAAUz-r1-392:-KuzKYtzG-LiHrgZbXR7:b242t0c) ([raw file](https://github.com/cockroachdb/cockroach/blob/9e625f3a5fe66ec0acca1650aa39f5e201980686/pkg/ccl/sqlccl/csv.go#L392)):*

do we want to log the error? or potentially also log a warning explaining that we'll fall back in the size == 0 block below?
</blockquote></details>

Added an info when we fallback.

---

*[pkg/ccl/sqlccl/csv.go, line 449 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-KuwSA6Y7HPL3pF49IDS:-KuzKaCRLd0JrmMLbkHG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9e625f3a5fe66ec0acca1650aa39f5e201980686/pkg/ccl/sqlccl/csv.go#L449)):*

capitalized?
</blockquote></details>

Done.

---

*[pkg/ccl/storageccl/export_storage.go, line 287 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-Ku_ugb_G0In8WloAAV--r1-287:-KuzKb7IWVVtBhwaNL_T:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9e625f3a5fe66ec0acca1650aa39f5e201980686/pkg/ccl/storageccl/export_storage.go#L287)):*

Some servers might not implement this -- at least it isn't part of our current required API.
We potentially want to add a test where our mock http server errors or otherwise returns an unusable response, just to make sure we keep handling it gracefully.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-:-KuzKVlqpyjGR8bMQ9Pt:bit7041)*

mjibson:
@dt PTAL at the new HTTP tests.

dt:
:lgtm:

---

Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18677#-:-KuzdqPwFdUVi_ZXE6ZP:bnfp4nl)*

PR #18680, opened 2017-09-21
log: allow log.Safe arguments in Fatal{f,}

tbg:
Principally allow the following idioms:

```
panic(err) // reported as %T unless it's a known whitelisted error
panic(log.Safe(err)) // reported as %+v
// String is reported
panic(log.Safe("I am just a const string so you can probably display me on sentry.io"))
// Reports roughly as "God knows what's in here %+v, %s | <*potentiallySensitive>"; OK
panic(fmt.Sprintf("God knows what's in here: %+v, %s", &potentiallySensitive, log.Safe("OK")))
log.Fatal(ctx, runtimeErr) // reported as a runtime error if it is one
log.Fatalf(ctx, "%s %s", log.Safe("x"), "y") // %s %s | x; <string>
```

See #18680 for first use case.


bdarnell:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18680#-:-KuaCGSIVLm89ZFZJXcE:bnfp4nl)*

Issue #17859, opened 2017-08-23
distsql: add a cli flag for overriding the temp store location


arjunravinarayan:
We default for using a subdirectory of store 1 for the temp store, but we want to allow users to override this, for instance, when they have a VM with a persistent disk and an ephemeral disk.

Add a CLI flag to set the temp store directory.

dianasaur323:
We probably need documentation for this. cc @jseldess

arjunravinarayan:
There's an open PR (#18544), so let's wait until that is resolved and cherry-picked before we document it. Thank you!

jseldess:
Thanks, @dianasaur323. We have an issue for documenting external storage. Not sure we'll get it done for the release, but in any case, I've link these issues: https://github.com/cockroachdb/docs/issues/1797.

PR #18686, opened 2017-09-21
cli,distsqlrun: Turn disk usage limits from cluster settings to flags

andreimatei:
Before this patch, the cluster setting sql.distsql.temp_storage.max_bytes and
sql.distsql.temp_storage.max_percentage were controlling how much
storage capacity could be used by distsql temp storage.
This didn't make sense: the settings should be per node, not per
cluster.

This patch replaces both settings with --max-disk-temp-storage flag. It
can take either a value or a percentage. One option briefly considered
was allowing the control over this temp storage through the --store
flag which already has an expressive language for specifying limits.
This option has the upside that it'd make it clear that the temp storage
is collocated with the first store and they share resources - in fact we
could easily prevent broken configuration where both the store and the
temp storage are configured to take the whole disk. However it's unclear
if tying temp storage to stores is a great idea thinking forward, and
also it's too late for me to explore that option more for 1.1.

This patch also changes the interpretation of the flag/cluster setting
when it's expressed as a percentage and the first store is type=mem.
Before, it used to be interpreted at p% of 512MB. Now, it will be
interpreted as p% of the internal memory.

The patch also has the effect that the dir for the first store will be
created earlier in the code path than it used to be, with permissions
0755. As far as I can tell, before the patch, the dir would have been
created later with the same permissions either when when initialized the
log dir or, if (--log_dir was specified) when we created the engine.

I'll cherry-pick this into 1.1 to remove the cluster settings before the
world gets wind of them.

### DIFF
--- a/pkg/base/store_spec.go
+++ b/pkg/base/store_spec.go
@@ -43,7 +43,10 @@ const MinimumStoreSize = 10 * 64 << 20
 // StoreSpec contains the details that can be specified in the cli pertaining
 // to the --store flag.
 type StoreSpec struct {
-	Path        string
+	Path string
+	// SizeInBytes is used for calculating free space and making rebalancing
+	// decisions. Zero indicates that there is no maximum size. This value is not
+	// actually used by the engine and thus not enforced.
 	SizeInBytes int64
 	SizePercent float64
 	InMemory    bool
--- a/pkg/base/test_server_args.go
+++ b/pkg/base/test_server_args.go
@@ -114,8 +114,7 @@ type TestClusterArgs struct {
 // DefaultTestStoreSpec is just a single in memory store of 100 MiB with no
 // special attributes.
 var DefaultTestStoreSpec = StoreSpec{
-	SizeInBytes: 100 << 20,
-	InMemory:    true,
+	InMemory: true,
 }
 
 // TestClusterReplicationMode represents the replication settings for a TestCluster.
--- a/pkg/cli/cliflags/flags.go
+++ b/pkg/cli/cliflags/flags.go
@@ -92,11 +92,36 @@ accept requests.`,
 	SQLMem = FlagInfo{
 		Name: "max-sql-memory",
 		Description: `
-Total size in bytes available for use to store temporary data for SQL
-clients, including prepared queries and intermediate data rows during
-query execution. Size suffixes are supported (e.g. 1GB and 1GiB). If
-left unspecified, defaults to 128MiB. A percentage of physical memory
-can also be specified (e.g. 25%).`,
+Maximum memory capacity available to store temporary data for SQL clients,
+including prepared queries and intermediate data rows during query execution.
+Accepts numbers interpreted as bytes, size suffixes (e.g. 1GB and 1GiB) or a
+percentage of physical memory (e.g. 25%).
+If left unspecified, defaults to 128MiB.
+`,
+	}
+
+	SQLTempStorage = FlagInfo{
+		Name: "max-disk-temp-storage",
+		Description: `
+Maximum storage capacity available to store temporary disk-based data for SQL
+queries that exceed the memory budget (e.g. join, sorts, etc are sometimes able
+to spill intermediate results to disk).
+Accepts numbers interpreted as bytes, size suffixes (e.g. 32GB and 32GiB) or a
+percentage of disk size (e.g. 10%).
+If left unspecified, defaults to 32GiB.
+
+The location of the temporary files is within the first store dir (see --store).
+If expressed as a percentage, --max-disk-temp-storage is interpreted relative to
+the size of the storage device on which the first store is placed. The temp
+space usage is never counted towards any store usage (although it does share the
+device with the first store) so, when configuring this, make sure that the size
+of this temp storage plus the size of the first store don't exceed the capacity
+of the storage device.
+If the first store is an in-memory one (i.e. type=mem), then this temporary "disk"
+data is also kept in-memory. A percentage value is interpreted as a percentage
+of the available internal memory. If not specified, the default shifts to 100MiB
+when the first store is in-memory.
+`,
 	}
 
 	Cache = FlagInfo{
--- a/pkg/cli/flags.go
+++ b/pkg/cli/flags.go
@@ -249,6 +249,10 @@ func init() {
 		// Engine flags.
 		varFlag(f, cacheSizeValue, cliflags.Cache)
 		varFlag(f, sqlSizeValue, cliflags.SQLMem)
+		// N.B. diskTempStorageSizeValue.ResolvePercentage() will be called after
+		// the stores flag has been parsed and the storage device that a percentage
+		// refers to becomes known.
+		varFlag(f, diskTempStorageSizeValue, cliflags.SQLTempStorage)
 	}
 
 	for _, cmd := range certCmds {
--- a/pkg/cli/interactive_tests/test_flags.tcl
+++ b/pkg/cli/interactive_tests/test_flags.tcl
@@ -1,0 +1,32 @@
+#! /usr/bin/env expect -f
+
+source [file join [file dirname $argv0] common.tcl]
+
+spawn /bin/bash
+send "PS1=':''/# '\r"
+eexpect ":/# "
+
+start_test "Check that --max-disk-temp-storage works."
+send "$argv start --insecure --store=path=mystore --max-disk-temp-storage=10GiB\r"
+eexpect "node starting"
+interrupt
+eexpect ":/# "
+end_test
+
+start_test "Check that --max-disk-temp-storage can be expressed as a percentage."
+send "$argv start --insecure --store=path=mystore --max-disk-temp-storage=10%\r"
+eexpect "node starting"
+interrupt
+eexpect ":/# "
+end_test
+
+start_test "Check that --max-disk-temp-storage percentage works when the store is in-memory."
+send "$argv start --insecure --store=type=mem,size=1GB --max-disk-temp-storage=10%\r"
+eexpect "node starting"
+interrupt
+eexpect ":/# "
+end_test
+
+send "exit 0\r"
+eexpect eof
+
--- a/pkg/cli/interactive_tests/test_missing_log_output.tcl
+++ b/pkg/cli/interactive_tests/test_missing_log_output.tcl
@@ -27,7 +27,10 @@ system "grep -F 'log: exiting because of error: write /dev/stderr: broken pipe' 
 end_test
 
 start_test "Check that a broken log file prints a message to stderr."
-send "$argv start -s=path=/dev/null --insecure --logtostderr\r"
+# The path that we pass to the --log-dir will already exist as a file.
+system "mkdir -p logs"
+system "touch logs/broken"
+send "$argv start -s=path=logs/db --log-dir=logs/broken --insecure --logtostderr\r"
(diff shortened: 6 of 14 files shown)


bdarnell:
:lgtm: but this needs some tests (probably in the tcl CLI tests) that ensure that both absolute numbers and percentages are parsed correctly (the percentage test is probably tricky to do exactly, but make sure that it at least doesn't crash)

---

Reviewed 1 of 4 files at r1, 3 of 3 files at r2, 7 of 7 files at r3.

---

*[pkg/cli/start.go, line 309 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KufJJQvp82AE7ueTn3h:-KufJJQvp82AE7ueTn3i:bsuzbxm) ([raw file](https://github.com/cockroachdb/cockroach/blob/a2a597ed1f404eb2c2a05099c10d0889f8ca320e/pkg/cli/start.go#L309)):*
> ```Go
> }
> 
> // distPercentResolverFactory takes in a path and produces a percentResolverFunc
> ```

s/dist/disk/

---

*[pkg/cli/cliflags/flags.go, line 113 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KufHMiGJDb4b-j65w-H:-KufHMiGJDb4b-j65w-I:bfta5ie) ([raw file](https://github.com/cockroachdb/cockroach/blob/a2a597ed1f404eb2c2a05099c10d0889f8ca320e/pkg/cli/cliflags/flags.go#L113)):*
> ```Go
> If left unspecified, defaults to 32GiB.
> 
> The location of the temporary files is within the first store dir (see --store).
> ```

cc @richardwu who is changing this (but not until 1.2: #18544)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-:-KufJpZ-t7vnM_qnI6w2:b9gbjlg)*

richardwu:
So I'm a little wary about introducing this flag (that is to my understanding a NOOP at this point).

Since we do plan on introducing a `--temp-store` or `--temp-dir` flag, it might be best to model the flag for temporary stores after what we currently do for `--store`, which allows keyed arguments like `path` and `size`.

With https://github.com/cockroachdb/cockroach/pull/18544, we'll separate temp store from main store.

That being said, this will probably be in 1.2 unless we want to cherry-pick https://github.com/cockroachdb/cockroach/pull/18544 (it does correct a 1.1 bug so I'm uncertain to what the qualification for cherry-picking is). I should have the PR ready with comments addressed by early next week. Maybe removing the cluster setting and having a temporary note somewhere will suffice?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-:-KufO5svmVLph2jZn7_s:bmi44up)*

andreimatei:
@richardwu I'm not sure I follow your point.
This present PR is not a no-op; it removes some cluster settings. I think it's not controversial that these cluster settings don't make much sense and should be removed before 1.1 ships (so this PR has to be cherry-picked).
This PR introduces a size limit flag. Whether or not this flag should be conflated with a path into a spec models after a store spec (or, in fact, whether the temp storage should be configured by a `--store` flag which was an option noted above) can be discussed afterwards. These are questions #18544 should answer (so, #18544 should be based on this). Whatever the answer is, I think this PR is clearly a step forward.
I haven't looked at #18544 carefully yet, but I think it's unlikely we want to cherry-pick it. Even if we were to cherry-pick it, I think it should still be based on this one. You said it "corrects a 1.1 bug". Are you referring to #17859? Technically that's not a bug but a feature request. 

Does this address your concerns?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-:-KupQuCBWmYUALVOu1Mw:b-xao1b2)*

andreimatei:
I've added a tcl test. @knz, PTAL and tell me if you think any of the comments from `test_log_flags.tcl` should be copied over.

---

Review status: 1 of 13 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-:-Kuq16GoUtN-y_iSJmcS:bblfegc)*

richardwu:
Ah my mistake, for some reason I conflated `MaxSizeBytes` used in the RocksDB config for temp storage (which we don't currently set for anything) with `TempStorageMaxSizeBytes` that I now see is being passed to our memory monitor.

Concerns addressed!

asubiotto:

---

*[pkg/cli/start.go, line 286 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KutlPgoPMF_y_th4IZ4:-KutlPgoPMF_y_th4IZ5:b-xfsoyj) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/start.go#L286)):*
> ```Go
> // For example, we have flags that can be expressed as percentages of the
> // capacity of storage device. Which storage device is in question might only be
> // known once other flags are parsed (e.g. --max-dist-temp-storage=10% depends
> ```

s/dist/disk

---

*[pkg/cli/start.go, line 373 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KutlanPMCe1DAx1mSE0:-KutlanPMCe1DAx1mSE1:b-ydxqr1) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/start.go#L373)):*
> ```Go
> }
> 
> // Resolve can be called to get the flag's value (if any). If the flag had been
> ```

I think this comment should be clear that the function doesn't return the value but sets the `bytesOrPercentageValue` to the value.

---

*[pkg/cli/start.go, line 443 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-Kutj2pGhkA86tbSH63t:-Kutj2pGhkA86tbSH63u:bktqspi) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/start.go#L443)):*
> ```Go
> 	if serverCfg.Stores.Specs[0].InMemory && !diskTempStorageSizeValue.IsSet() {
> 		// The default temp store size is 100MB when the first store is in memory.
> 		serverCfg.TempStoreMaxSizeBytes = 100 << 20 // 100MB
> ```

Consider extracting this value into a const.

---

*[pkg/cli/cliflags/flags.go, line 108 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-Kutji1v28VqWgzWCywU:-Kutji1v28VqWgzWCywV:bxe3jab) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/cliflags/flags.go#L108)):*
> ```Go
> Maximum storage capacity available to store temporary disk-based data for SQL
> queries that exceed the memory budget (e.g. join, sorts, etc are sometimes able
> to spill to intermediate results to disk).
> ```

s/to spill to/to spill

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686)*

andreimatei:
Review status: 1 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/cli/start.go, line 309 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KufJJQvp82AE7ueTn3h:-Kuuiy9nx9DUxqTBceqw:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a2a597ed1f404eb2c2a05099c10d0889f8ca320e/pkg/cli/start.go#L309)):*

s/dist/disk/
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 286 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KutlPgoPMF_y_th4IZ4:-KutssBeqxGejzvgM3gH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/start.go#L286)):*

s/dist/disk
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 373 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KutlanPMCe1DAx1mSE0:-KuttOm_jusx-3gzhyrC:btxprux) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/start.go#L373)):*

I think this comment should be clear that the function doesn't return the value but sets the `bytesOrPercentageValue` to the value.
</blockquote></details>

the comment talks about `*v` and the method doesn't return anything. How can I make it clearer?

---

*[pkg/cli/start.go, line 443 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-Kutj2pGhkA86tbSH63t:-Kuuj4bjQ13cajw6-JXR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/start.go#L443)):*

Consider extracting this value into a const.
</blockquote></details>

Done.

---

*[pkg/cli/cliflags/flags.go, line 108 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-Kutji1v28VqWgzWCywU:-KuttxezRdvn9JTOWtG5:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/cliflags/flags.go#L108)):*

s/to spill to/to spill
</blockquote></details>

Done.

---

*[pkg/cli/interactive_tests/test_flags.tcl, line 9 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-Kut5WoxWkmCF-uMVLyb:-Kuuj67TMBqjJE229iiq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fe1ab33ec52fdf22fcb69a4a2ce15bd61ba24257/pkg/cli/interactive_tests/test_flags.tcl#L9)):*

s/dist/disk/g (3 times in this file)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686)*

knz:
Reviewed 2 of 10 files at r9.
Review status: 3 of 14 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[pkg/cli/interactive_tests/test_flags.tcl, line 9 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686#-KuuxsarCRm1ceRsa6-R:-KuuxsarCRm1ceRsa6-S:b-94eaf0) ([raw file](https://github.com/cockroachdb/cockroach/blob/9deca092f9fab0eabacb122a1a2f79b317addf60/pkg/cli/interactive_tests/test_flags.tcl#L9)):*
> ```Tcl
> eexpect ":/# "
> 
> start_test "Check that --max-disk-temp-storage works."
> ```

s/works/is recognized without error/

(later we'll have other tests to check it works)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18686)*

PR #18687, opened 2017-09-21
vendor: fix dep breakage

knz:
The commit from vendored#7 didn't make it properly to the main repo.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -143,7 +143,7 @@
   branch = "master"
   name = "github.com/cockroachdb/stress"
   packages = ["."]
-  revision = "60f5b5fa410fdfd786059ccb56d468691f2305aa"
+  revision = "29b5d31b4c3a949cb3a726750bc34c4d58ec15e8"
 
 [[projects]]
   name = "github.com/codahale/hdrhistogram"
@@ -616,7 +616,7 @@
 
 [[projects]]
   name = "golang.org/x/sync"
-  packages = ["errgroup","singleflight","syncmap"]
+  packages = ["errgroup","syncmap"]
   revision = "f52d1811a62927559de87708c8913c1650ce4f26"
 
 [[projects]]
@@ -673,6 +673,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "b91ad64e984d1dc807522ec2a35cf1b198abb4cc4f2ffa3e5fdc9fe91c20d87a"
+  inputs-digest = "0c4cdc41e60e9405fc2dab5b883635e3c5b6f15afef358d81d488091edea9e40"
   solver-name = "gps-cdcl"
   solver-version = 1


tschottdorf:
Are you sure? in my local up to date `cockroachdb/stress` I can run `diff main.go ../cockroach/vendor/github.com/cockroachdb/stress/main.go` against current master's vendored dir (029c47e016ef0dfe34a346eac9e67987ab1de4ef) and it comes back empty. Are perhaps accidentally going the wrong way?

knz:
if you run `dep` it will update the file. This means the version you checked in in the vendor repo and linked to the cockroach master is not the one generated by `dep`. In my next PR that updates Gopkg.toml, the linter complains that the file in stress is not up to date.

tschottdorf:
```
[[constraint]]
  name = "github.com/cockroachdb/stress"
  branch = "master"
```

I think you just need to update your `$GOPATH/src/github.com/cockroachdb/stress`.

knz:
Why is `dep` fetching files from my GOPATH? Isn't supposed to fetch over the network?

tschottdorf:
I'm not sure what it does, but if it did use the network it seems that it would not produce a diff, for the top commit is the one we currently use: https://github.com/cockroachdb/stress/commits/master

knz:
Interesting. Ok I'll drop this one and investigate what's going on on my side.

knz:
Welp, I really cannot reproduce *your* commit.

knz:
Steps to reproduce:
1. `git checkout master`
2. `go install ./vendor/github.com/golang/dep/cmd/dep` (As per `build/README.md`)
3. `dep ensure`

(let it run)

When this completes I get this PR as well as the diff in https://github.com/cockroachdb/vendored/pull/10.

knz:
@dt @tschottdorf can you really double check what you get when you follow the 3 steps above?

knz:
(for reference I have none of the affected packages in my GOPATH - I checked)

dt:
huh, bizarre. When I run ensure, I don't see the revision in the lock change (still `60f5b5fa410fdfd786059ccb56d468691f2305aa`) but I see the content in vendor change.

tschottdorf:
@knz I was able to reproduce the problem. Unfortunately, even after 10 minutes of digging it's completely unclear to me why it would chose a nonrecent `master` revision.

Workaround:

```diff
diff --git a/Gopkg.toml b/Gopkg.toml
index eb842870f..e4251a7a6 100644
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -55,4 +55,4 @@ ignored = [

 [[constraint]]
   name = "github.com/cockroachdb/stress"
-  branch = "master"
+  revision = "29b5d31b4c3a949cb3a726750bc34c4d58ec15e8"
```

but I'd really *love* to know what's going on there. I'll file an upstream issue. In the meantime, would you like to just apply that patch in this PR?

tschottdorf:
Maybe the problem is that unless `-update` is specified to `dep`, it won't attempt to bump any versions?  I'll try `dep ensure -v -update github.com/cockroachdb/stress` now.

tschottdorf:
Yep -- that leaves `vendor` intact, but produces the following lockfile diff:

```diff
diff --git a/Gopkg.lock b/Gopkg.lock
index 647b188b6..ca090d8f8 100644
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -143,7 +143,7 @@
   branch = "master"
   name = "github.com/cockroachdb/stress"
   packages = ["."]
-  revision = "60f5b5fa410fdfd786059ccb56d468691f2305aa"
+  revision = "29b5d31b4c3a949cb3a726750bc34c4d58ec15e8"

 [[projects]]
   name = "github.com/codahale/hdrhistogram"
@@ -616,7 +616,7 @@

 [[projects]]
   name = "golang.org/x/sync"
-  packages = ["errgroup","singleflight","syncmap"]
+  packages = ["errgroup","syncmap"]
   revision = "f52d1811a62927559de87708c8913c1650ce4f26"

 [[projects]]
@@ -673,6 +673,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "b91ad64e984d1dc807522ec2a35cf1b198abb4cc4f2ffa3e5fdc9fe91c20d87a"
+  inputs-digest = "0c4cdc41e60e9405fc2dab5b883635e3c5b6f15afef358d81d488091edea9e40"
   solver-name = "gps-cdcl"
   solver-version = 1
```

Pretty sure that's how I messed up your flow -- I probably ran `dep ensure` and didn't commit the lockfile bump somehow? 60f5b5fa410fdfd786059ccb56d468691f2305aa is the commit it winds back to in your PR here and 29b5d31b4c3a949cb3a726750bc34c4d58ec15e8 is where it is (and should be).

tschottdorf:
So, for this PR, please run:

- `dep ensure -v -update github.com/cockroachdb/stress`
- you should see the above diff in `Gopkg.lock`
- commit that
- run `dep ensure -v` just to see that it's a no-op

dt:
we could also try adding a semvar-formed tag in `stress` since that's what dep prefers to find

tschottdorf:
@dt would that have changed anything? My understanding is that I simply confused `dep` by not updating the lock file. I like semvars principally, but I worry that it's not worth here because it'd be nice if we could just bump the fork, run `dep ensure -update` and avoid having another intermediate step which makes a new semvar release.

benesch:
Agreed, I don't think it's worth semver'ing stress. (What stress really
needs are some tests.)

knz:
Done, PTAL

tbg (approved):

Issue #18659, opened 2017-09-21
perf: quantify space/write amplification


petermattis:
Forked from #18657:

> > The disk io though is quite worrying at around 26MiB writes/s and 8% CPU spent in iowait as indicated by dstat. The data being updated is very small (one integer). Granted, CockroachDB keeps all past values so let's assume each update is like an insert. The string has 3 bytes plus 4 byte integer plus overhead for metadata and encoding. Let's assume a generous 64 bytes per entry. At 2500 qps, that would be around 256KiB/s. LSM storage engines have write amplification. Not sure how many levels were generated in this test but I'd assume not too many. So let's assume each row is actually written 4 times as time goes by. That's 1MiB/s. Still off by a factor of 26. Not sure where all this disk io comes from but it seems excessive.
>
> Unless you were running for a long period of time, the LSM write amplification should be a non-factor. Note that every write involves a 4x write amplification outside of the LSM due to writing the Raft log and then subsequently committing the command. Both the Raft log write and the committing of the command involve a 2x write amplification due to RocksDB writing to its internal WAL and then subsequently flushing the memtable. There is some a possibility this will be reduced in the future. See #16948.
>
> But that write amplification still does explain the 26MiB/sec that you're seeing. The 64 bytes per entry might be a bit pessimistic given some of the other Raft and internal state that is written on each write. I'm going to file an issue to investigate in detail where the space overhead for each write is coming from.

The schema in question is:

```
CREATE TABLE bench (s string primary key, n int);
```

And the inserts have the form:

```
INSERT INTO bench (s, n) VALUES($1, 1) ON CONFLICT (s) DO UPDATE SET n = bench.n + 1
```

The [load generator script](https://gist.github.com/anonymous/c2d43573f40c60f7f5be0d594bdf4b18) is generating random 3 letter keys using upper-case characters giving a total of 17576 rows. 

This issue is about understanding where the space and write amplification is coming from. Some places to investigate are the overhead of `RaftCommand`, `raftpb.HardState` and `MVCCStats` which are all written on every write.

tschottdorf:
At a recent talk, a user asked about space amplification due to having to sync to disk on every write. If we're doing only sequential small writes, is there a detrimental effect where each WAL write rewrites a whole block? This shouldn't be happening in the above example due to batching, but I'm curious.

petermattis:
> If we're doing only sequential small writes, is there a detrimental effect where each WAL write rewrites a whole block? 

[...]

PR #18689, opened 2017-09-22
storage: reduce Raft per-op overhead

petermattis:
For simple Raft commands, this reduces size of Raft proposals by 45%
which directly reduces the majority of Raft network overhead by 45%. The
proposals are also written to the Raft log and results in a 34%
reduction in Raft log overhead.

These changes are backward compatible because they involve not encoding
zero-valued fields.

See #18659


petermattis:
I still need to double-check the network and disk bandwidth savings on a real cluster.

petermattis:
Running a write-only workload (`kv --read-percent 0`) shows a modest reduction in network bandwidth:

screen shot 2017-09-22 at 9 15 39 am

The period on the left is for current master and the period on the right is with this PR. Note that I specified `--max-rate 10000` so we're comparing apples-to-apples numbers of operations.

This is smaller than the number calculated above. Probably because I wasn't accounting for other per-op overhead or figuring in the pgwire traffic. 

screen shot 2017-09-22 at 9 16 22 am

Disk bandwidth seems only very slightly lower. Even more surprising is that there is a 15% drop in throughput when the max-rate is not capped. I have no explanation for that. Is there more memory pressure somehow? Are the additional pointers affecting GC?

petermattis:
Figured out the cause of the throughput drop. It was due to allocating `ReplicaState` in `ReplicatedEvalResult` if that field was nil when applying a Raft command in order to pass through `{Raft,Lease}AppliedIndex`. Somewhat surprising that one allocation could cause such a drop and yet it did. I checked and double-check this by removing and adding that allocation.

petermattis:
Added a few more commits which shave down the proposal overhead. Happy to break some of these out into separate PRs.

bdarnell:
:lgtm:

---

Review status: 0 of 29 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/storage/replica.go, line 3042 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-Kujmi7g6cKQVnMdfFgB:-Kujmi7g6cKQVnMdfFgC:b-11gn5x) ([raw file](https://github.com/cockroachdb/cockroach/blob/87246c43f86f0af25ec3744c1dcd0499eb95a838/pkg/storage/replica.go#L3042)):*
> ```Go
> 	defer r.store.enqueueRaftUpdateCheck(r.RangeID)
> 
> 	if log.V(5) {
> ```

This will probably be annoying with `/debug/logspy`

---

*[pkg/storage/replica_state.go, line 306 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KujoHkG60EJA_Nk3Gyb:-KujoHkG60EJA_Nk3Gyc:bjy46tb) ([raw file](https://github.com/cockroachdb/cockroach/blob/486050d271a478ed72302e3089db66e454609535/pkg/storage/replica_state.go#L306)):*
> ```Go
> func (rsl replicaStateLoader) loadGCThreshold(
> 	ctx context.Context, reader engine.Reader,
> ) (*hlc.Timestamp, error) {
> ```

It seems weird to return a pointer here when most callers immediately dereference it. I'd rather keep this method (and loadTxnSpanGCThreshold) as it was, and use a local variable in the callers that do need a pointer (only replicaStateLoader.load, right?)

For MVCCStats it makes more sense to return a pointer because that's a big struct to copy around. 

---

*[pkg/storage/replica_test.go, line 5565 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KujrOSUqq7LgUWvFPU5:-KujrOSUqq7LgUWvFPU6:bdcotjb) ([raw file](https://github.com/cockroachdb/cockroach/blob/f30c8d3b0553887e924dab3bc62a6e63a6c93432/pkg/storage/replica_test.go#L5565)):*
> ```Go
> 	// nontrivial lease requested with the first write below.
> 	baseStats.Add(enginepb.MVCCStats{
> 		SysBytes: 26,
> ```

Why did this increase? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-:-KujsqnI3gR3LiYyaHs5:bnfp4nl)*

petermattis:
Review status: 0 of 29 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/storage/replica.go, line 3042 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-Kujmi7g6cKQVnMdfFgB:-Kuk7pnYyEEKBmfQEQM6:b-oqw0rg) ([raw file](https://github.com/cockroachdb/cockroach/blob/87246c43f86f0af25ec3744c1dcd0499eb95a838/pkg/storage/replica.go#L3042)):*

This will probably be annoying with `/debug/logspy`
</blockquote></details>

Yeah, I'll change this to `if false` before merging. It has been pretty handy it determining where the bytes are going.

---

*[pkg/storage/replica_state.go, line 306 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KujoHkG60EJA_Nk3Gyb:-Kuk88IpyseWSkTrR0xU:b-68uhxc) ([raw file](https://github.com/cockroachdb/cockroach/blob/486050d271a478ed72302e3089db66e454609535/pkg/storage/replica_state.go#L306)):*

It seems weird to return a pointer here when most callers immediately dereference it. I'd rather keep this method (and loadTxnSpanGCThreshold) as it was, and use a local variable in the callers that do need a pointer (only replicaStateLoader.load, right?)

For MVCCStats it makes more sense to return a pointer because that's a big struct to copy around. 
</blockquote></details>

While the other callers don't need a pointer, the pointer is already present (we're passing `&t` to `MVCCGetProto`). That was my rationale for changing the return type. 

---

*[pkg/storage/replica_test.go, line 5565 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KujrOSUqq7LgUWvFPU5:-Kuk8PxDyiWl-32NtZ9Z:bwrj3q6) ([raw file](https://github.com/cockroachdb/cockroach/blob/f30c8d3b0553887e924dab3bc62a6e63a6c93432/pkg/storage/replica_test.go#L5565)):*

Why did this increase? 
</blockquote></details>

`initialState().SysBytes` decreased by 12 bytes for some reason which was affecting another test but not this one. I've forgotten the exact reason, let me get back to you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689)*

petermattis:
Review status: 0 of 29 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/storage/replica_test.go, line 5565 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KujrOSUqq7LgUWvFPU5:-KukFEzUgPHk36k6PgGI:b2pqnx1) ([raw file](https://github.com/cockroachdb/cockroach/blob/f30c8d3b0553887e924dab3bc62a6e63a6c93432/pkg/storage/replica_test.go#L5565)):*

`initialState().SysBytes` decreased by 12 bytes for some reason which was affecting another test but not this one. I've forgotten the exact reason, let me get back to you.
</blockquote></details>

When the first range is created, `replicaStateLoader.save` is called on an empty lease. The change to make some fields in `roachpb.Lease` nullable reduced encoded size of an empty lease from 26 bytes to 14 bytes which necessitated the change to `initalState().SysBytes`. But this test forces a valid lease to be written and doing so bumps up `SysBytes` again. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689)*

tschottdorf:
:lgtm:

The change makes me nervous but I think it's because I just don't know off the top of my head all the places in which we are serializing these. However, `below_raft_protos_test.go` has us covered, right?

Do you think we adequately protect against accidental re-blowup here in future changes? I assume those will be rare enough and scrutinized anyway, but just curious if there would be enough test breakage for the developer to realize that it's a really bad idea to make these structs any bigger than they need to be.

---

Reviewed 1 of 7 files at r1, 9 of 12 files at r6, 4 of 7 files at r7, 11 of 12 files at r8, 4 of 4 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-:-KutW5gb_alFAGj7P8Wn:b-ln7s3y)*

petermattis:
Yes, `below_raft_protos_test.go` covers us. That test verifies both that the specified protos are seen below Raft and that their encodings haven't changed. I accidentally triggered breakage there when constructing this PR.

> Do you think we adequately protect against accidental re-blowup here in future changes? 

That's a good point and the answer is definitely no. I'll file an issue as I'm not quite sure what to do, though perhaps something as simple as encoding the zero value of `storagebase.RaftCommand` would be sufficient.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-:-KutYzl_GOs5t9MbQHDe:bqnlz52)*

petermattis:
Making `ReplicaState.Stats` nullable is causing `TestVersionUpgrade` to hang forever. Not at all sure why is that part of these changes looks innocuous. I'm investigating.

petermattis:
The hang was caused by a node in the local cluster used by the acceptance test crashing with an assertion due to an assertion. Old nodes encode a zero-value `ReplicaState.Stats` which the new node happily decodes but was asserting that stats weren't being sent. The fix is a few lines. Thanks for the test @tschottdorf.

tschottdorf:
:thumbsup: We really need to run these mixed-version clusters for longer periods of time. Is that on our priority TODO list somewhere?

---

Reviewed 28 of 28 files at r10, 12 of 12 files at r11, 7 of 7 files at r12, 12 of 12 files at r13, 4 of 4 files at r14.

---

*[pkg/storage/replica_proposal.go, line 850 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KuyTgWj2FL-P3gdW_A4:-KuyTgWj2FL-P3gdW_A5:b97cai7) ([raw file](https://github.com/cockroachdb/cockroach/blob/589d16474fe94048d36dbe7a3d86fbfd4e2af22e/pkg/storage/replica_proposal.go#L850)):*
> ```Go
> 		}
> 
> 		if rResult.State.Stats != nil && (*rResult.State.Stats == enginepb.MVCCStats{}) {
> ```

Comment here wouldn't hurt.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-:-KuyU-y_ikLZCQvpMw8W:b-3b8y1q)*

petermattis:
> We really need to run these mixed-version clusters for longer periods of time. Is that on our priority TODO list somewhere?

Let me check. I'll file an issue if we don't already have one.

---

Review status: 10 of 29 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/storage/replica_proposal.go, line 850 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-KuyTgWj2FL-P3gdW_A4:-KuyVHqrYcI-cIhL2Uyj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/589d16474fe94048d36dbe7a3d86fbfd4e2af22e/pkg/storage/replica_proposal.go#L850)):*

Comment here wouldn't hurt.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689#-:-KuyV7lA6c6C85W-o2AE:b-et7oca)*

tschottdorf:
Reviewed 19 of 19 files at r15, 7 of 7 files at r16, 12 of 12 files at r17, 4 of 4 files at r18.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18689)*

PR #18520, opened 2017-09-16
distsql: added local and final stage aggregation for STDDEV/VARIANCE

richardwu:
Fixes #14351

Local/final aggregation for STDDEV/VARIANCE is slightly non-trivial because of how standard 
deviation and variance is calculated. 

The [naive two-pass approach using the sum of squared differences formula](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Two-pass_algorithm) requires the global average to be known, causing local/final aggregation to be irrelevant.

The [second naive approach using the sum of squared values and sum of values](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Na.C3.AFve_algorithm) that can employ local/final aggregation is numerically unstable and can lead to catastrophic cancellation.

Ultimately, the [Knuth/Welford algorithm for variance](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm) allows us to perform local/final aggregation for STDDEV/VARIANCE under distsql. This works for local aggregation (which we currently use for non-distsql calculation of STDDEV/VARIANCE). For final aggregation, we must take the sum of Knuth/Welford squared differences. Refer to https://www.johndcook.com/blog/skewness_kurtosis and our adaptation for `NumericStats` (https://github.com/cockroachdb/cockroach/pull/17728).

Issue #13689, opened 2017-02-21
sql,distsql: inconsistent aggregation results with DistSQL


RaduBerinde:
I am seeing a few aggregation logic tests occasionally fail when DistSQL is enabled on a 3 node cluster (with fake span resolver). This happens both with and without the apd change #13551.

```
query RRR
SELECT VARIANCE(x), VARIANCE(y::decimal), VARIANCE(z) FROM xyz
----
9.00 4.5 6.333333333333333
We get:
9.00 4.5 6.333333333333334  
```
This comes just from STDDEV/VARIANCE processing arguments in a different order. Indeed, we can reproduce on regular sql using:
```
SELECT VARIANCE(x), VARIANCE(y::decimal), VARIANCE(z) FROM (SELECT x,y,z FROM xyz ORDER BY random())
```

This is an incorrect<sup>1</sup> result. I think that whatever output precision `p` we want for the result, `decimalVarianceAggregate` must use higher precision for its internal calculations (e.g. `p + 3`). Technically we may need something like `p + O(log count)` - we could increase the precision of the internal calculations as the count increases.

Another one is:
```
query RRR
SELECT STDDEV(x), STDDEV(y::decimal), STDDEV(z) FROM xyz
----
3  2.121320343559643  2.516611478423583
We get:
3  2.121320343559643  2.5166114784235836
```

Here we have an inconsistency in the number of decimals we output, and the result is also incorrect, the following decimal should be `2` not `6` (the correct result is: `2.51661147842358323241...`).

CC @mjibson @nvanbenschoten @knz

[1]: It can be argued that this is not incorrect, just "imprecise". My point of view is that we are choosing how many decimals to output; if we only compute to 14 accurate decimals, we should output just 14..  In this sense if the true value is `1.242`, it's imprecise but not incorrect to output `1.24`, but it's incorrect to output `1.23` or `1.25`.  Of course things that are on the "edge" like `1.25499xy..` or `1.25000xy..` could go either way - both `1.25` and `1.26` are correct.. I don't quite know what the formal guarantee should be.

knz:
It's a bug, and it's bad :)

mjibson:
I'll work on this after the apd branch gets in.

RaduBerinde:
Note that PR #13706 temporarily comments out these tests; they should be re-enabled when the issue is fixed.

mjibson:
[...]

PR #18701, opened 2017-09-22
distsql: increase intermediate precision of decimal calculations in square difference calculations

### LINKED ISSUE #18691: sql/distsqlplan: TestDistAggregationTable/VARIANCE/int1/50 failed under stress
SHA: https://github.com/cockroachdb/cockroach/commits/1d5c72d03ced1cb44c3f00dc8a656a5fb347f18b

Parameters:
```
TAGS=
GOFLAGS=
```

Stress build found a failed test: https://teamcity.cockroachdb.com/viewLog.html?buildId=358086&tab=buildLog

```

```

richardwu:
This PR increases the intermediate precision of our apd contexts for
intermediate calculations to address off by last-digit issues in our
final decimal results for STDDEV/VARIANCE.

The imprecision was introduced during
https://github.com/cockroachdb/cockroach/pull/18520 when STDDEV/VARIANCE
aggregations can now take advantage of local and final stage
aggregating. Since this architecture involves non-deterministic ordering
(which implies non-deterministic order of operations for our running
squared difference algorithm), there is a trivial discrepancy between
local and non-local results.

Fixes #18691, #18692, #18693, #18694.

### DIFF
--- a/pkg/sql/parser/aggregate_builtins.go
+++ b/pkg/sql/parser/aggregate_builtins.go
@@ -873,11 +873,6 @@ func (a *intSqrDiffAggregate) Count() *apd.Decimal {
 	return a.agg.Count()
 }
 
-// Ed is part of the decimalSqrDiff interface.
-func (a *intSqrDiffAggregate) Ed() *apd.ErrDecimal {
-	return a.agg.Ed()
-}
-
 // Tmp is part of the decimalSqrDiff interface.
 func (a *intSqrDiffAggregate) Tmp() *apd.Decimal {
 	return a.agg.Tmp()
@@ -964,15 +959,8 @@ type decimalSqrDiffAggregate struct {
 }
 
 func newDecimalSqrDiff() decimalSqrDiff {
-	// Use extra internal precision during squared difference to protect against
-	// order changes that can happen in dist SQL. The additional 3 here should
-	// allow for correctness up to 1000 more worst case inputs than non-worst
-	// case inputs. See #13689 for more analysis and other algorithms.
-	c := DecimalCtx.WithPrecision(DecimalCtx.Precision + 3)
-	ed := apd.MakeErrDecimal(c)
-	return &decimalSqrDiffAggregate{
-		ed: &ed,
-	}
+	ed := apd.MakeErrDecimal(IntermediateCtx)
+	return &decimalSqrDiffAggregate{ed: &ed}
 }
 
 func newDecimalSqrDiffAggregate(_ []Type, _ *EvalContext) AggregateFunc {
@@ -982,11 +970,6 @@ func newDecimalSqrDiffAggregate(_ []Type, _ *EvalContext) AggregateFunc {
 // Count is part of the decimalSqrDiff interface.
 func (a *decimalSqrDiffAggregate) Count() *apd.Decimal {
 	return &a.count
-}
-
-// Ed is part of the decimalSqrDiff interface.
-func (a *decimalSqrDiffAggregate) Ed() *apd.ErrDecimal {
-	return a.ed
 }
 
 // Tmp is part of the decimalSqrDiff interface.
@@ -1107,25 +1090,13 @@ type decimalSumSqrDiffsAggregate struct {
 }
 
 func newDecimalSumSqrDiffs() decimalSqrDiff {
-	// Use extra internal precision during squared difference to protect against
-	// order changes that can happen in dist SQL. The additional 3 here should
-	// allow for correctness up to 1000 more worst case inputs than non-worst
-	// case inputs. See #13689 for more analysis and other algorithms.
-	c := DecimalCtx.WithPrecision(DecimalCtx.Precision + 3)
-	ed := apd.MakeErrDecimal(c)
-	return &decimalSumSqrDiffsAggregate{
-		ed: &ed,
-	}
+	ed := apd.MakeErrDecimal(IntermediateCtx)
+	return &decimalSumSqrDiffsAggregate{ed: &ed}
 }
 
 // Count is part of the decimalSqrDiff interface.
 func (a *decimalSumSqrDiffsAggregate) Count() *apd.Decimal {
 	return &a.count
-}
-
-// Ed is part of the decimalSqrDiff interface.
-func (a *decimalSumSqrDiffsAggregate) Ed() *apd.ErrDecimal {
-	return a.ed
 }
 
 // Tmp is part of the decimalSumSqrDiffs interface.
@@ -1186,10 +1157,6 @@ func (a *decimalSumSqrDiffsAggregate) Result() (Datum, error) {
 		return DNull, nil
 	}
 	dd := &DDecimal{a.sqrDiff}
-	// Remove trailing zeros. Depending on the order in which the input
-	// is processed, some number of trailing zeros could be added to the
-	// output. Remove them so that the results are the same regardless of order.
-	dd.Decimal.Reduce(&dd.Decimal)
 	return dd, nil
 }
 
@@ -1204,7 +1171,6 @@ type floatSqrDiff interface {
 type decimalSqrDiff interface {
 	AggregateFunc
 	Count() *apd.Decimal
-	Ed() *apd.ErrDecimal
 	Tmp() *apd.Decimal
 }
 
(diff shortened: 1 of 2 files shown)


richardwu:
I ran `make stress` on the above failed tests for a few minutes locally and they seem to be fixed.

mjibson:

---

*[pkg/sql/parser/aggregate_builtins.go, line 951 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu1tWOF4G8N8HBhHOu:-Kuu1tWOF4G8N8HBhHOv:b-hqu9d4) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L951)):*
> ```Go
> // intermediateApdCtx returns the apd context that should be used for all
> // intermediate decimal calculations.
> func intermediateApdCtx() *apd.Context {
> ```

This function only exists because it's called in a few places. If it's generally useful like this, we should just add another permanent decimal context alongside DecimalCtx.

---

*[pkg/sql/parser/aggregate_builtins.go, line 952 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu0Lyhk3Q2fDUgk4aI:-Kuu0Lyhk3Q2fDUgk4aJ:b-nku216) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L952)):*
This comment is not accurate anymore because the algorithm here has changed. I would change it to:

> Use extra internal precision during intermediate calculations to protect against order changes that can happen in dist SQL. The additional 5 allows the stress tests to pass, exercise this problem. See #13689 for more analysis (of the previous algorithm) and other methods to increase precision.

And remove the 2 line comment you added.

---

*[pkg/sql/parser/aggregate_builtins.go, line 964 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu1Rg_Gi2-rkhOWv8k:-Kuu1Rg_Gi2-rkhOWv8l:bdcp1vs) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L964)):*
> ```Go
> // finalApdCtx returns the apd context that should be used for the final
> // decimal calculation that will be propagated to the user.
> func finalApdCtx() *apd.Context {
> ```

I don't understand the point of finalApdCtx. DecimalCtx's whole job is to be the decimal context at which to do lossy operations before returning to users. This is an unneeded indirection.

---

*[pkg/sql/parser/aggregate_builtins.go, line 1269 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu2KSoEvM8b5m-Mt3D:-Kuu2KSoEvM8b5m-Mt3E:bh266x2) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L1269)):*
> ```Go
> 		return nil, err
> 	}
> 	if _, err = intermediateApdCtx().Sub(a.agg.Tmp(), a.agg.Count(), decimalOne); err != nil {
> ```

Why not `a.agg.ed`? Then we could remove intermediateApdCtx().

---

*[pkg/sql/parser/aggregate_builtins.go, line 1357 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu1DbMsERo5rHzUZYW:-Kuu1DbMsERo5rHzUZYX:bfpkp6l) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L1357)):*
> ```Go
> func (a *decimalStdDevAggregate) Result() (Datum, error) {
> 	// TODO(richardwu): both decimalVarianceAggregate and
> 	// finalDecimalVarianceAggregate returns a decimal result with
> ```

returns -> return

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701)*

richardwu:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/aggregate_builtins.go, line 951 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu1tWOF4G8N8HBhHOu:-KuuRd_iysHtfTwsS2gi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L951)):*

This function only exists because it's called in a few places. If it's generally useful like this, we should just add another permanent decimal context alongside DecimalCtx.
</blockquote></details>

Done.

---

*[pkg/sql/parser/aggregate_builtins.go, line 952 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu0Lyhk3Q2fDUgk4aI:-KuuS8S58ASnll63yw2G:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L952)):*

This comment is not accurate anymore because the algorithm here has changed. I would change it to:

> Use extra internal precision during intermediate calculations to protect against order changes that can happen in dist SQL. The additional 5 allows the stress tests to pass, exercise this problem. See #13689 for more analysis (of the previous algorithm) and other methods to increase precision.

And remove the 2 line comment you added.
</blockquote></details>

Done.

---

*[pkg/sql/parser/aggregate_builtins.go, line 964 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu1Rg_Gi2-rkhOWv8k:-KuuS8oreNW7hxT74hAj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L964)):*

I don't understand the point of finalApdCtx. DecimalCtx's whole job is to be the decimal context at which to do lossy operations before returning to users. This is an unneeded indirection.
</blockquote></details>

Done.

---

*[pkg/sql/parser/aggregate_builtins.go, line 1269 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu2KSoEvM8b5m-Mt3D:-KuuSN9GRCwxFWwimiIc:b-yzajit) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L1269)):*

Why not `a.agg.ed`? Then we could remove intermediateApdCtx().
</blockquote></details>

Now that I've moved everything to a global variable `IntermediateCtx`, I've avoided using `a.agg.ed` since that would require introducing a getter function to the `decimalSqrDiff` interface (which `a.agg` is defined as). Furthermore, we only need to do one intermediate calculation here so I believe it's unnecessary to use an `ErrDecimal` here.

---

*[pkg/sql/parser/aggregate_builtins.go, line 1357 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-Kuu1DbMsERo5rHzUZYW:-KuuT4EljVohgpPvaRdD:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6cad74eda72a5a06599dfa33d1d58e24897c9b58/pkg/sql/parser/aggregate_builtins.go#L1357)):*

returns -> return
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701)*

mjibson:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/aggregate_builtins.go, line 1340 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-KuueKWatfNnKleSz_y1:-KuueKWbCdWqflY1dmZJ:brsh6du) ([raw file](https://github.com/cockroachdb/cockroach/blob/79e6f7a91e879070ea6a3ffb1255a99caa735379/pkg/sql/parser/aggregate_builtins.go#L1340)):*
> ```Go
> 	// finalDecimalVarianceAggregate return a decimal result with
> 	// default DecimalCtx precision. We want to be able to specify that the
> 	// varianceAggregate use intermediateApdCtx (with the extra precision)
> ```

use intermediateApdCtx -> uses IntermediateCtx

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-:-Kuue_afTOCO2fpfb_Od:bnfp4nl)*

richardwu:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/parser/aggregate_builtins.go, line 1340 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701#-KuueKWatfNnKleSz_y1:-KuumJI-cHoX3HBAytRR:b-ubb87t) ([raw file](https://github.com/cockroachdb/cockroach/blob/79e6f7a91e879070ea6a3ffb1255a99caa735379/pkg/sql/parser/aggregate_builtins.go#L1340)):*

use intermediateApdCtx -> uses IntermediateCtx
</blockquote></details>

Good catch, thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18701)*

PR #18702, opened 2017-09-22
distsql: use ASCENDING encoding for comparisons

### LINKED ISSUE #18586: Bug: Select DISTINCT (col) does not return correct result ...
Is this a question, feature request, or bug report?

bug

Have you checked our documentation at https://cockroachlabs.com/docs/stable/? If you could not find an answer there, please consider asking your question in our community forum at https://forum.cockroachlabs.com/, as it would benefit other members of our community.

Prefer live chat? Message our engineers on our Gitter channel at https://gitter.im/cockroachdb/cockroach.

**FEATURE REQUEST**

1. Does an issue already exist addressing this request? If yes, please add a :+1: reaction to the existing issue. If not, move on to step 2.

2. Please describe the feature you are requesting, as well as your proposed use case for this feature.

3. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

**BUG REPORT**

     grep -F '[config]' cockroach-data/logs/cockroach.log

   When log files are not available, supply the output of `cockroach version`
   and all flags/environment variables passed to `cockroach start` instead.

- What did you do?

- What did you expect to see?

- What did you see instead?

W170915 11:12:23.488359 1 cli/start.go:650  RUNNING IN INSECURE MODE!

- Your cluster is open for any client that can access <all your IP addresses>.
- Any user, even root, can log in without providing a password.
- Any user, connecting as root, can read or write any data in your cluster.
- There is no network encryption nor authentication, and thus no confidentiality.

Check out how to secure your cluster: https://www.cockroachlabs.com/docs/stable/secure-a-cluster.html

```
I170915 11:12:23.489434 1 server/config.go:303  available memory from cgroups (8.0 EiB) exceeds system memory 3.9 GiB, using system memory
W170915 11:12:23.489487 1 cli/start.go:570  Using the default setting for --cache (128 MiB).
  A significantly larger value is usually needed for good performance.
  If you have a dedicated server a reasonable setting is --cache=25% (989 MiB).
I170915 11:12:23.489513 1 cli/start.go:658  CockroachDB CCL be60cc5-dirty (linux amd64, built 2017/09/15 11:04:19, go1.8.3)
I170915 11:12:23.590373 1 server/config.go:303  available memory from cgroups (8.0 EiB) exceeds system memory 3.9 GiB, using system memory
I170915 11:12:23.590420 1 server/config.go:417  system total memory: 3.9 GiB
I170915 11:12:23.590633 1 server/config.go:419  server configuration:
max offset                500000000
cache size                128 MiB
SQL memory pool size      128 MiB
scan interval             10m0s
scan max idle time        200ms
metrics sample interval   10s
event log enabled         true
linearizable              false
```

have some fresh created 9 node cluster across 3 dc with custom build from 1.1 branch.
Looks like SELECT DISTINCT .. has some issues : 

have some table requests with below structure;

```
| Field | Type | Null | Default | Indices |
| requestkey | INT | false | unique_rowid() | {"primary","requests_requeststate_dcname_id_idx"} |
| requesttype | STRING | false | NULL | {} |
| requesttime | TIMESTAMP WITH TIME ZONE | true | NULL | {} |
| requeststate | STRING | true | NULL | {"requests_requeststate_dcname_id_idx"} |
| trace | STRING | true | NULL | {} |
| laststep | STRING | true | NULL | {} |
| lasterror | STRING | true | NULL | {} |
| resourcetype | STRING | false | NULL | {} |
| dcname | STRING | true | NULL | {"requests_requeststate_dcname_id_idx"} |
| id | BYTES | false | NULL | {"requests_requeststate_dcname_id_idx"} |
| body | STRING | true | NULL | {} |
```

Data distribution looks like this

```
select resourceType,count(1) from requests group by resourceType;
+--------------+----------+
| resourcetype | count(1) |
+--------------+----------+
| INTERNAL | 315 |
| DB | 64 |
| JOB | 35 |
```

execute

```
select distinct resourceType from requests;
+--------------+
| resourcetype |
+--------------+
| INTERNAL |
| DB |
| DB |
| DB |
| JOB |
| JOB |
| INTERNAL |
+--------------+
(7 rows)
```

Got 7 rows, expected 3 

Executing 

```
select distinct 'X'||resourceType||'X' from requests;
+------------------------------+
| ('X' || resourcetype) || 'X' |
+------------------------------+
| XINTERNALX |
| XDBX |
| XJOBX |
+------------------------------+
(3 rows)
```

Was not yet able to create a reproducible testcase.
A short test works generally fine for same type of query.

```
apiuserftx@10.96.214.109:26257/restapiftx> create table test (a string);
CREATE TABLE
apiuserftx@10.96.214.109:26257/restapiftx> INSERT INTO test VALUES ('A');
INSERT 1
apiuserftx@10.96.214.109:26257/restapiftx> INSERT INTO test VALUES ('A');
INSERT 1
apiuserftx@10.96.214.109:26257/restapiftx> INSERT INTO test VALUES ('B');
INSERT 1
apiuserftx@10.96.214.109:26257/restapiftx> INSERT INTO test VALUES ('B');
INSERT 1
apiuserftx@10.96.214.109:26257/restapiftx> SELECT * FROM test;
+---+
| a |
+---+
| A |
| A |
| B |
| B |
+---+
(4 rows)
apiuserftx@10.96.214.109:26257/restapiftx> SELECT DISTINCT a FROM test;
+---+
| a |
+---+
| A |
| B |
+---+
(2 rows)
```

rjnn:
As the comment on` EncDatum.Encode` indicates, using the `DatumEncoding_VALUE`
encoding is not valid for comparisons, as the encoded datum can
contain a ColumnID prefix. Added a brief test case that minimally reproduces this bug.

Fixes #18586.

There is an open TODO when I tried to audit other uses of `DatumEncoding_VALUE` when encoding, and I'm not sure if the bug will occur in the use in `disk_row_container.go`. cc @asubiotto to make that call. 

I will create a cherrypick commit after this is merged.

A longer investigation log is attached in a comment below, if anyone is interested in my logbook notes while I was tracking down the source of this bug (cc @asubiotto, @richardwu who previously expressed interest).

### DIFF
--- a/pkg/sql/distsqlrun/aggregator.go
+++ b/pkg/sql/distsqlrun/aggregator.go
@@ -409,7 +409,7 @@ func (ag *aggregator) encode(
 	appendTo []byte, row sqlbase.EncDatumRow,
 ) (encoding []byte, err error) {
 	for _, colIdx := range ag.groupCols {
-		appendTo, err = row[colIdx].Encode(&ag.datumAlloc, sqlbase.DatumEncoding_VALUE, appendTo)
+		appendTo, err = row[colIdx].Encode(&ag.datumAlloc, sqlbase.DatumEncoding_ASCENDING_KEY, appendTo)
 		if err != nil {
 			return appendTo, err
 		}
--- a/pkg/sql/distsqlrun/distinct.go
+++ b/pkg/sql/distsqlrun/distinct.go
@@ -175,10 +175,10 @@ func (d *distinct) encode(appendTo []byte, row sqlbase.EncDatumRow) ([]byte, err
 		// TODO(irfansharif): Different rows may come with different encodings,
 		// e.g. if they come from different streams that were merged, in which
 		// case the encodings don't match (despite having the same underlying
-		// datums). We instead opt to always choose sqlbase.DatumEncoding_VALUE
+		// datums). We instead opt to always choose sqlbase.DatumEncoding_ASCENDING_KEY
 		// but we may want to check the first row for what encodings are already
 		// available.
-		appendTo, err = datum.Encode(&d.datumAlloc, sqlbase.DatumEncoding_VALUE, appendTo)
+		appendTo, err = datum.Encode(&d.datumAlloc, sqlbase.DatumEncoding_ASCENDING_KEY, appendTo)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/sql/logictest/testdata/logic_test/distsql_agg
+++ b/pkg/sql/logictest/testdata/logic_test/distsql_agg
@@ -543,3 +543,17 @@ query T
 SELECT "URL" FROM [EXPLAIN (DISTSQL) SELECT * FROM (VALUES (1, '222'), (2, '444')) t1(a,b) JOIN (VALUES (1, 100.0), (3, 32.0)) t2(a,b) ON t1.a = t2.a]
 ----
 https://cockroachdb.github.io/distsqlplan/decode.html?eJykkDFrwzAQhff-CnNTAjfYTuggKJhOTYemdOhSNBjrapu6OqOToBD834ulIXVoWkI26e6-e-_eASwbeqo_SUC9QQEaYXTckAi7uZQGduYLVI7Q2zH4uawRGnYE6gC-9wOBgtd6CCSAYMjX_ZD2bbP7bFVmTRfsh6xBTwgc_HGJ-LolUPmEVwrdXit03B8sO0OOzEJCz-R_I7-4faile-Tekls6Hujdr6pifef6tosvQNgHr7KqwKrEaoPV9uwdxSWBvZCMbIVO7zmTkEYg01IKRTi4hp4dN1EmffeRiwVD4lO3TJ-dja1o8CdcXACXp3D5J7xZwPmkp5vvAAAA__-HZe5E
+
+statement ok
+CREATE TABLE nullables (a INT, b INT, c INT, PRIMARY KEY (a))
+
+statement ok
+INSERT INTO nullables VALUES (1,1,1)
+
+statement ok
+INSERT INTO nullables VALUES (2,NULL,1)
+
+query II
+SELECT c, COUNT(*) FROM nullables GROUP BY c;
+----
+1 2
--- a/pkg/sql/sqlbase/encoded_datum.go
+++ b/pkg/sql/sqlbase/encoded_datum.go
@@ -212,7 +212,7 @@ func (ed *EncDatum) Encoding() (DatumEncoding, bool) {
 // a column ID so they should not be used to test for equality.
 func (ed *EncDatum) Encode(a *DatumAlloc, enc DatumEncoding, appendTo []byte) ([]byte, error) {
 	if ed.encoded != nil && enc == ed.encoding {
-		// We already have an encoding that matches
+		// We already have an encoding that matches that we can use.
 		return append(appendTo, ed.encoded...), nil
 	}
 	if err := ed.EnsureDecoded(a); err != nil {


arjunravinarayan:
This is a verbose log of my investigation into this bug, with
dead-ends and unproductive rabbitholes included, for anyone interested
in the spelunking expedition.

The original bug report is at #18586. Reproduced locally with a
dump of the data. Additional observations: this occurs not just
with `DISTINCT`, but also `GROUP BY`:

```
root@localhost:26257/heiko> SELECT resourcetype, count(1) FROM requests GROUP BY resourcetype;
+--------------+----------+
| resourcetype | count(1) |
+--------------+----------+
| INTERNAL     |     1681 |
| JOB          |       33 |
| DB           |       60 |
+--------------+----------+
(3 rows)

Time: 5.639517ms

root@localhost:26257/heiko> SELECT resourcetype, count(*) FROM requests GROUP BY resourcetype;
+--------------+----------+
| resourcetype | count(*) |
+--------------+----------+
| JOB          |        1 |
| INTERNAL     |       21 |
| INTERNAL     |     1659 |
| JOB          |       32 |
| DB           |       27 |
| DB           |       20 |
| INTERNAL     |        1 |
| DB           |       13 |
+--------------+----------+
(8 rows)

Time: 9.25811ms
```

The fact that this reproduces with `GROUP BY` finally pointed me in
the right direction. My first guesses were that the DISTINCT code was
using a spurious ordering. This is not the case: Group does not use
any orderings for speedups, since it has to put all the values in
buckets anyway. So it could not be ordering-related. (As an aside, the
`COUNT(*)` and `COUNT(1)` queries have different
plans:
[`COUNT(*)`](https://cockroachdb.github.io/distsqlplan/decode.html?eJyUkEFLxDAQhe_-CnmnXchhsyfJqaIeemmlVjxIkdgMpdBtamYCytL_Lm0O6sKKHudNvveFOWL0jgp7IIZ5hkajMAXfErMPS5Qe5O4dZqfQj1OUJW4UWh8I5gjpZSAY1PZ1oIqsowAFR2L7YS2dQn-w4SML9BaJhaFQRjGX2RWaWcFH-SplsR3B6Fn9XXzddYE6K_7Em2ko5Ld3Rb3J9BYKN-VjUb9U5dPDZntWvf-PuiKe_Mj0Q3yueTc3CuQ6SndlH0NL98G3qyaN5cqtgSOWtNVpyMe0Wj74Hda_wvsTuJkvPgMAAP__zq6kbQ==) and
[`COUNT(1)`](https://cockroachdb.github.io/distsqlplan/decode.html?eJyUkLFq9DAQhPv_KY6p_gMVJ1dBlUOSwo0vGKcKLhRrMAaf5axkSDj87sFWkeTgQtJpZzXzDXvG6B1Le2KAeYZGozCJbxmCl1VKHwr3BnNQ6MdpjqvcKLReCHNG7ONAGNT2ZWBF6yhQcIy2H7bQSfqTlfdc-DozxACFiqOjmF1-o3baGFOUNZpFwc_xExCi7QijF_X7ErddJ-xs9Bcdcg2F4v6hrP_neg-Fu-PT-s72V7nZX7gVw-THwG_Ua8mHpVGg65gOHPwsLR_FtxsmjcfNtwmOIaatTkMxptVa8KtZ_2jOLszN8u8jAAD__3N_pnM=) as
the `COUNT(*)` path is optimized, although this isn't directly related
to the source of the bug, it gives us a neat side-by-side comparison
of results).

I logged the actual encoding bytes being used by the `aggregator` in
[`distsqlrun/aggregator.go:280`](https://github.com/cockroachdb/cockroach/blob/3e15c99c030a9fa946027d59fe7ceac397b897fc/pkg/sql/distsqlrun/aggregator.go#L280), which reveals the source of the spurious buckets:

```
COUNT(*):
%X[22 2 68 66]
%X[22 8 73 78 84 69 82 78 65 76]
%X[38 2 68 66]
%X[38 3 74 79 66]
%X[54 2 68 66]
%X[54 3 74 79 66]
%X[54 8 73 78 84 69 82 78 65 76]
%X[70 8 73 78 84 69 82 78 65 76]
COUNT(1):
%X[6 2 68 66]
%X[6 3 74 79 66]
%X[6 8 73 78 84 69 82 78 65 76]
```

This reveals that it's the first byte of the encoding that is off
in the `COUNT(*)` path: examination revelas the encoding is `TAG
|| SIZE || RAW BYTES` (where `||` represents concatenation). The
tag is wrong in the `COUNT(*)` case: the sizes and raw bytes are
correct (`68 66` is ascii for `DB`, and so on). What are `22`,
`38`, `54` and `70`? The `6` in the `COUNT(1)` case is expected:
if you follow the chain of calls from
[`aggregator.go:412`](https://github.com/cockroachdb/cockroach/blob/3e15c99c030a9fa946027d59fe7ceac397b897fc/pkg/sql/distsqlrun/aggregator.go#L412),
you eventually get to
[`encoding.go:1165`](https://github.com/cockroachdb/cockroach/blob/3e15c99c030a9fa946027d59fe7ceac397b897fc/pkg/util/encoding/encoding.go#L1164),
which prepends `uint32(typ)`, which has 15 possible values, the
6th of which is `Bytes`. So how do we get 22, 38, 54, and 70 in
the `COUNT(*)` case?

So where do the incorrect TAGs come from? Let's walk back the
call path: when we call `Encode`,
[`sqlbase/encoded_datum:215`](https://github.com/cockroachdb/cockroach/blob/3e15c99c030a9fa946027d59fe7ceac397b897fc/pkg/sql/sqlbase/encoded_datum.go#L215)
decides that if there is already an encoding, it's going to use
that one, rather than wastefully re-encoding. That's definitely
how this gets there.

But the previous line ensures that we only re-use the encoding as long
as it's the same as our desired encoding type (in this case, a
`DatumEncoding_VALUE`). The comment at the top of `EncDatum.Encode` is
suspicious: it says not to use `DatumEncoding_VALUE` for comparisons,
but we do. This is almost certainly the source of the bug, but let us
continue to see where it first gets encoded.

In the `COUNT(*)` case, somebody else has already encoded the
datum, using differing prefixes, and we've used them. When
someone encodes with a `ColumnID` that's not set to `NoColumnID`,
we smush together both the `ColumnID` and the
`Type` ([here](https://github.com/cockroachdb/cockroach/blob/3e15c99c030a9fa946027d59fe7ceac397b897fc/pkg/util/encoding/encoding.go#L1170)). This
is the source of the different values: the numbers `22`, `38`,
`54`, and `70` are consistent with column IDs `[1,2,3,4] << 4 +
6`. We're reading from a single
column (`requests.resourceType`). Why are there four different
column IDs?

Adding some debug log statements to dump the `columnID` when
encoding did not result in any output when rerunning the
queries. This means that the rowfetcher is fetching these encoded
values directly from the tables, so this encoding is
happening on *INSERTt*. Dropping the database and rerunning the
inserts verifies this claim - the log statements print out when
inserting. We also see that the different column IDs are exactly
the count that we see in the differing groups:

```
> less cockroach.log| grep "encoded bytes" | grep "73 78 84 69 82" | grep -E "1$" | wc -l
1
> less cockroach.log| grep "encoded bytes" | grep "73 78 84 69 82" | grep -E "4$" | wc -l
21
less cockroach.log| grep "encoded bytes" | grep "73 78 84 69 82" | grep -E "3$" | wc -l
1659
```

More importantly, now we have the ordering of which inserts resulted in the differing column ID:

```
less cockroach.log| grep "encoded bytes" | grep "73 78 84 69 82" | grep -E "4$" -n | cut -d':' -f1
1188
1192
1193
1273
1274
1275
1354
1356
1357
1435
1436
1438
1528
1529
1532
1533
1534
1535
1625
1627
1629
```

Now to look at the 1188th insert statement: The first thing that
stands out about this statement is that it has a lot of
NULLs. Careful examination shows that this pattern is repeated
across the other rows in that list. In effect, the 1,2,3, or 4
tracks the number of NULL columns before the column being
selected. I cross checked this with a bunch of rows in the entire
dataset.

It's pretty clear now that there's nothing that can be done about
this on the INSERT side: what needs to happen is that the value
should be re-encoded with the ColumnID removed by the DistSQL
processor. The comment on `EncDatum.Encode` should be respected, and
we should change the encoding used.

Now this gives us a straightforward minimal reproduction of the bug
(added as a test case to `distsql_agg`). After that, we can audit
other uses of `Encode` in DistSQL, to see if there are any other bad
uses of `DatumEncoding_VALUE` in comparisons.

arjunravinarayan:
*[pkg/sql/distsqlrun/disk_row_container.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-KufHGr-ibK3j7HxDFWI:-KufHGr-ibK3j7HxDFWJ:b2fi1qf) ([raw file](https://github.com/cockroachdb/cockroach/blob/5bfa9d6e2554960afbf4c4648347df1a6a6e09eb/pkg/sql/distsqlrun/disk_row_container.go#L133)):*
> ```Go
> 	for _, i := range d.valueIdxs {
> 		var err error
> 		// TODO(asubiotto): Should we use DatumEncoding_ASCENDING_KEY here instead?
> ```

@asubiotto would appreciate your input here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702)*

richardwu:
Very cool bug digging and log book!

I was wondering if it might be better to declare a global `sqlbase.preferredCompEncoding = EncodeDatum_ASCENDING_KEY` in `sqlbase/encoded_datum.go` and replace all these instances where we use an encoding for comparison via distsql?

This was sort of hinted at in `outbox.go` https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/distsqlrun/outbox.go so perhaps it's time we put emphasis on this fact. 

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-:-KufK25H-1WPdUuR4egZ:bc1urme)*

asubiotto:
LGTM. Great work!

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/distsqlrun/disk_row_container.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-KufHGr-ibK3j7HxDFWI:-KufO7kzB943HwKflOe4:b-xv2en6) ([raw file](https://github.com/cockroachdb/cockroach/blob/5bfa9d6e2554960afbf4c4648347df1a6a6e09eb/pkg/sql/distsqlrun/disk_row_container.go#L133)):*

@asubiotto would appreciate your input here.
</blockquote></details>

No, we only encode Datums as values if they are not an ordering column. Therefore they are not being compared and can stay as `DatumEncoding_VALUE`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-:-KufO-36VoBYDdck391U:b-leecem)*

arjunravinarayan:
*[pkg/sql/distsqlrun/disk_row_container.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-KufHGr-ibK3j7HxDFWI:-KufQUhR4nzEC--_83Km:brkhiir) ([raw file](https://github.com/cockroachdb/cockroach/blob/5bfa9d6e2554960afbf4c4648347df1a6a6e09eb/pkg/sql/distsqlrun/disk_row_container.go#L133)):*

No, we only encode Datums as values if they are not an ordering column. Therefore they are not being compared and can stay as `DatumEncoding_VALUE`.
</blockquote></details>

Thank you. Removed the TODO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702)*

arjunravinarayan:
TFTRs. 

@richardwu: I would prefer we have a single `EncodeForCompare` in DistSQL which encapsulates all these decisions. We can then audit all calls in DistSQL easily.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-:-KufQlqajmqU02P_0uyh:b-h6llg1)*

andreimatei:
Great find!

Should we make `EncDatum.Compare` re-encode on the fly if anything has the value encoding?

I also think we should have that `EncodeForCompare` defined as a constant (and thus hinting at comparisons), otherwise some comments would be in order at the lines where you've changed one encoding to the other.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-:-Kul1dpNNo6vE91uJLAs:bnko5xk)*

RaduBerinde (commented):
Great debugging, respect!  LGTM

arjunravinarayan:
In the interest of merging this fix pronto, I've filed another issue to track follow-on work: #18726.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18702#-:-KuoJtMawyjrKMrJ1N8K:b-p8go84)*

PR #18706, opened 2017-09-23
cache: introduce OrderedCache Entry API

nvb:
In 1a6ccc3 we introduced the `baseCache.DelEntry` method. This allows a
cache removal without requiring a lookup if the associated `cache.Entry`
has already been retrieved. This change exposes `...Entry` methods on
`OrderedCache` so that clients can use `DelEntry`.

The change then uses this new API in the `RangedescriptorCache` and
the `raftEntryCache`.

### DIFF
--- a/pkg/kv/range_cache.go
+++ b/pkg/kv/range_cache.go
@@ -162,8 +162,9 @@ func (rdc *RangeDescriptorCache) String() string {
 
 func (rdc *RangeDescriptorCache) stringLocked() string {
 	var buf bytes.Buffer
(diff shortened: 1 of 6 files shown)


bdarnell:
:lgtm:

---

Reviewed 1 of 3 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18706#-:-Kuk7hpxf1G86NL0U_3o:bnfp4nl)*

PR #1206, opened 2015-06-01
Meta2KeyMax should not be valid range split key

es-chow:
when try to make range split at key Meta2KeyMax (\x00\x00meta2\xff\xff), the call to AdminSplit will succeed, but the server will fall into a infinite loop:

``` log
W0601 22:56:18.657173    5953 dist_sender.go:588] failed to invoke InternalResolveIntent: key range "\x00\x00\x00k\x00\xff\x00\xffmeta2\xff\xff\x00\x01rtn-"-"" outside of boun\
ds of range ""-"\x00\x00meta2\xff\xff"
I0601 22:56:18.657203    5953 retry.go:95] routing InternalResolveIntent rpc failed; retrying immediately
W0601 22:56:18.657240    5953 dist_sender.go:588] failed to invoke InternalResolveIntent: key range "\x00\x00meta2\xff\xff"-"" outside of bounds of range ""-"\x00\x00meta2\xff\
\xff"
I0601 22:56:18.657260    5953 retry.go:95] routing InternalResolveIntent rpc failed; retrying immediately
W0601 22:56:18.657280    5953 dist_sender.go:588] failed to invoke InternalResolveIntent: key range "\x00range-tree-root"-"" outside of bounds of range ""-"\x00\x00meta2\xff\x\
ff"
I0601 22:56:18.657287    5953 retry.go:95] routing InternalResolveIntent rpc failed; retrying immediately
```

The reason for this is Meta1KeyMax (\x00\x00meta1\xff\xff) server as dual purpose:
1. boundary check of Meta1 addressing range (\x00~\x00\x00meta1\xff\xff)
2. metadata key for key Meta2KeyMax (\x00\x00meta2\xff\xff),
so when a split happens at Meta2KeyMax, initial range will be split into 2 ranges:
1. range1 (0x00~Meta2KeyMax)
2. range2 [Meta2KeyMax~KeyMax)
but Meta1KeyMax will fall into a dilemma: 
1. As boundary key for Meta1 addressing range, it's value should be range1
2. but as a metadata key for Meta2KeyMax, it's value should be range2.
In current implementation after splitRangeAddressing function, the value stored is range1, so the InternalResolveIntent(Key:Meta2KeyMax) request will be route to range1, but this key is outside of bounds of range1. so the InternalResolveIntent will retry infinitely but no reason to be succeed.
So the solution may be not allowed Meta2KeyMax be a split key.
Two issues for this fix are not resolved:
[...]

PR #18709, opened 2017-09-23
kv: fix RangeDescriptorCache eviction with meta2 splits

nvb:
Related to #18578.

Previously, it was possible for meta2 splits to cause incorrect meta2
range descriptors to be evicted from the `RangeDescriptorCache`. The
reason for this is that `evictCachedRangeDescriptor` would evict the
`RangeMetaKey` of the lookup key, instead of the `EndKey` of the
user-space descriptor. The comment below explains this in more detail:

> Evicting 'meta(descKey)' instead of 'meta(cachedDesc.EndKey)' could be
> incorrect here because it could result in us evicting the wrong meta2
> descriptor. Imagine that descriptors are in the configuration:
>
> meta2 descs:  [/meta2/a,/meta2/m),  [/meta2/m,/meta2/z)
> user  descs:  [a,f),   [f,k),   [k,p),   [p,u),   [u,z)
>
> A lookup for the key 'l' would return the third user-space descriptor
> [k,p), after a continuation lookup. This descriptor is stored on the
> second meta2 range because its end key is p, which maps to /meta2/p.
> If we later want to evict this user-space descriptor, we also want to
> evict the meta2 descriptor for the range that contains it. However,
> if we naively evicted meta('l'), we'd end up evicting the first meta2
> descriptor. Instead, we want to evict meta('p'), since 'p' is the end
> key of the user-space descriptor.

First four commits are from #18706. Can be ignored.

### DIFF
--- a/pkg/kv/dist_sender_test.go
+++ b/pkg/kv/dist_sender_test.go
@@ -858,7 +858,7 @@ func TestEvictCacheOnError(t *testing.T) {
 		if _, ok := ds.leaseHolderCache.Lookup(context.TODO(), 1); ok != !tc.shouldClearLeaseHolder {
 			t.Errorf("%d: lease holder cache eviction: shouldClearLeaseHolder=%t, but value is %t", i, tc.shouldClearLeaseHolder, ok)
 		}
-		if cachedDesc, err := ds.rangeCache.GetCachedRangeDescriptor(roachpb.RKey(key), false /* !inclusive */); err != nil {
+		if cachedDesc, err := ds.rangeCache.GetCachedRangeDescriptor(roachpb.RKey(key), false /* !inverted */); err != nil {
 			t.Error(err)
 		} else if cachedDesc == nil != tc.shouldClearReplica {
 			t.Errorf("%d: unexpected second replica lookup behaviour: wanted=%t", i, tc.shouldClearReplica)
@@ -1279,8 +1279,8 @@ func TestMultiRangeGapReverse(t *testing.T) {
 				if !reverse {
 					return key.Less(descs[i].EndKey)
 				}
-				// In reverse mode, the key is "inclusive". If we scan
-				// [a,z) in reverse mode, we'd look up key z.
+				// In reverse mode, the range boundary behavior is "inverted".
+				// If we scan [a,z) in reverse mode, we'd look up key z.
 				return !descs[i].EndKey.Less(key) // key <= EndKey
 			})
 		}
--- a/pkg/kv/range_cache.go
+++ b/pkg/kv/range_cache.go
@@ -319,8 +319,8 @@ func (rdc *RangeDescriptorCache) lookupRangeDescriptorInternal(
 		defer rdc.rangeCache.Unlock()
 
 		// These need to be separate because we need to preserve the pointer to rs[0]
-		// so that the seenDesc logic works correctly in EvictCachedRangeDescriptor. An
-		// append could cause a copy, which would change the address of rs[0]. We insert
+		// so that the compare-and-evict logic works correctly in EvictCachedRangeDescriptor.
+		// An append could cause a copy, which would change the address of rs[0]. We insert
 		// the prefetched descriptors first to avoid any unintended overwriting. We then
 		// only insert the first desired descriptor, since any other descriptor in rs would
 		// overwrite rs[0]. Instead, these are handled with the evictToken.
@@ -425,13 +425,15 @@ func (rdc *RangeDescriptorCache) performRangeLookup(
 	return rdc.db.RangeLookup(ctx, metadataKey, desc, useReverseScan)
 }
 
-// EvictCachedRangeDescriptor will evict any cached range descriptors
-// for the given key. It is intended that this method be called from a
-// consumer of rangeDescriptorCache if the returned range descriptor is
-// discovered to be stale.
-// seenDesc should always be passed in and is used as the basis of a
-// compare-and-evict (as pointers); if it is nil, eviction is unconditional
-// but a warning will be logged.
+// EvictCachedRangeDescriptor will evict any cached user-space and meta range
+// descriptors for the given key. It is intended that this method be called from
+// a consumer of rangeDescriptorCache through the EvictionToken abstraction if
+// the returned range descriptor is discovered to be stale.
+//
+// seenDesc should always be passed in if available, and is used as the basis of
+// a pointer-based compare-and-evict strategy. This means that if the cache does
+// not contain the provided descriptor, no descriptor will be evicted. If
+// seenDesc is nil, eviction is unconditional.
 func (rdc *RangeDescriptorCache) EvictCachedRangeDescriptor(
 	ctx context.Context, descKey roachpb.RKey, seenDesc *roachpb.RangeDescriptor, inclusive bool,
 ) error {
@@ -440,13 +442,16 @@ func (rdc *RangeDescriptorCache) EvictCachedRangeDescriptor(
 	return rdc.evictCachedRangeDescriptorLocked(ctx, descKey, seenDesc, inclusive)
 }
 
+// evictCachedRangeDescriptorLocked is like evictCachedRangeDescriptor, but it
+// assumes that the caller holds a write lock on rdc.rangeCache.
 func (rdc *RangeDescriptorCache) evictCachedRangeDescriptorLocked(
 	ctx context.Context, descKey roachpb.RKey, seenDesc *roachpb.RangeDescriptor, inclusive bool,
 ) error {
 	cachedDesc, entry, err := rdc.getCachedRangeDescriptorLocked(descKey, inclusive)
-	if err != nil {
+	if err != nil || cachedDesc == nil {
 		return err
 	}
+
 	// Note that we're doing a "compare-and-erase": If seenDesc is not nil,
 	// we want to clean the cache only if it equals the cached range
 	// descriptor as a pointer. If not, then likely some other caller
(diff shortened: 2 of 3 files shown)


nvanbenschoten:
While making this change I had a realization that I hope is wrong. I think that
we'll be able to have 3 levels of descriptor indirection for the last range in the
system once we allow meta2 splits again. To make that concrete, take for
example:
```
 r1: [/min,/meta2/)
 r2: [/meta2/,/meta2/m)
 r3: [/meta2/m,/metaMax) <- can't split at meta2max
 r4: [/metaMax,keymax)
```
r4's desc lives on r3, r3's desc lives on r2, r2's desc
lives on r1. Remember `meta(MetaMax) -> /meta2/metaMaxByte`.
The problem here is that the `EndKey` of a range that contains meta2
descriptors doesn't need to be a meta2 key itself, meaning that `RangeMetaKey`
will give a meta2 key address. Consequences of this include that evicting
`keymax` will evict two meta2 descriptors, but there are a lot more.

Is this ok? Do we expect to have meta2 descriptors stored in meta2 ranges?
For reference, I think the fundamental issue here was also the cause of https://github.com/cockroachdb/cockroach/pull/1206.
Life would be so much easier if we stored meta keys at `meta(desc.StartKey)`.

petermattis:
I was recently trying to wrap my head around the cause of #1206. I'm not seeing anything wrong with your logic, but I've just spent the day outside so perhaps my brain in sun baked.

bdarnell:
> Do we expect to have meta2 descriptors stored in meta2 ranges?

No, we don't. Meta1 lookups are supposed to be sufficient for finding all meta2 entries. But while it's surprising, I'm not immediately seeing any severe consequences here (I'm probably just missing things). It helps that with our current automatic split policies, the last range in the system is always empty. 

Is this unique to splits at exactly `/metaMax`? Could we avoid this by disallowing splits at that point too?

nvanbenschoten:
> Is this unique to splits at exactly /metaMax

No, this will happen if there is a split anywhere between (`/meta2max`,`/metaMax`], which there always has to be. This is because the last range descriptor in the system will always be stored at exactly `/meta2max` (`RangeMetaKey(KeyMax)`), which means that the smallest key that the last meta2 range can end at while still containing this descriptor is `/meta2max.Next()`.

nvanbenschoten:
Actually, I might be wrong about that. Ending meta2 ranges at `Meta2KeyMax.Next()` would also avoid this issue.

bdarnell:
:lgtm:

---

---

*[pkg/kv/range_cache.go, line 471 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-KuorOyJrgqienDQZB4I:-KuorOyJrgqienDQZB4J:b-80svrr) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L471)):*
> ```Go
> 	}
> 
> 	for {
> ```

It's been a long time since I've looked at this code, but this seems nuts. Evicting the meta range descriptors is expensive and we shouldn't do it until we have evidence that they're out of date. 

Evicting the meta descriptors is currently tricky because we do inconsistent reads for range lookups, which could go to arbitrarily stale replicas. But we don't need the read-from-any-replica semantics of inconsistent reads (do we?), we just need to be able to look behind pending intents. I think we should probably stop using the inconsistent mode for range lookups and instead use a flag specific to RangeLookupRequest that controls how it handles intents. Then I think we'd be able to handle evictions one meta level at a time instead of eagerly evicting multiple levels because we're not sure where we got bad information. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-:-KuowFLaIDTx8rwL0zla:bnfp4nl)*

tschottdorf:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[pkg/kv/range_cache.go, line 452 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-KutaA1lxDFeL5pBgJWj:-KutaA1lxDFeL5pBgJWk:b-yexhlh) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L452)):*
> ```Go
> }
> 
> // evictCachedRangeDescriptorLocked is like evictCachedRangeDescriptor, but it
> ```

Since you're polishing up comments, I think it'd be useful if you gave a little example for `inclusive` here (assuming you eat another CI anyways).

---

*[pkg/kv/range_cache.go, line 458 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-Kut_n11Lf_RIhxYUEn8:-Kut_n11Lf_RIhxYUEn9:b-ydaei4) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L458)):*
> ```Go
> ) error {
> 	cachedDesc, entry, err := rdc.getCachedRangeDescriptorLocked(descKey, inclusive)
> 	if err != nil || cachedDesc == nil {
> ```

Didn't think we'd ever cache a `nil` `cachedDesc`. How does that happen?

---

*[pkg/kv/range_cache.go, line 471 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-KuorOyJrgqienDQZB4I:-KutaWUCGO4aKq4GTBSI:b-baw0o1) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L471)):*

It's been a long time since I've looked at this code, but this seems nuts. Evicting the meta range descriptors is expensive and we shouldn't do it until we have evidence that they're out of date. 

Evicting the meta descriptors is currently tricky because we do inconsistent reads for range lookups, which could go to arbitrarily stale replicas. But we don't need the read-from-any-replica semantics of inconsistent reads (do we?), we just need to be able to look behind pending intents. I think we should probably stop using the inconsistent mode for range lookups and instead use a flag specific to RangeLookupRequest that controls how it handles intents. Then I think we'd be able to handle evictions one meta level at a time instead of eagerly evicting multiple levels because we're not sure where we got bad information. 
</blockquote></details>

I agree, and I can't remember there being a strong reason to evict unconditionally.

The inconsistent reads might've been premature optimization. Would be nice to see them go, especially given how unpredictable they are. They could be a good fallback for the case of unavailable meta ranges (to lessen the impact on the cluster), but let's see if that's ever necessary.

---

*[pkg/kv/range_cache.go, line 510 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-Kuta3P9Wn_cRrR8iqmm:-Kuta3P9Wn_cRrR8iqmn:bjxbegk) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L510)):*
> ```Go
> 		}
> 
> 		cachedDesc, entry, err = rdc.getCachedRangeDescriptorLocked(descKey, false)
> ```

optio-nit: `false /* !inclusive */`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-:-Kutb3fJUqGxtCl0sFni:bnfp4nl)*

nvanbenschoten:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/kv/range_cache.go, line 452 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-KutaA1lxDFeL5pBgJWj:-KvYJ0wvY2_ITDZioI0q:bh7o5gu) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L452)):*

Since you're polishing up comments, I think it'd be useful if you gave a little example for `inclusive` here (assuming you eat another CI anyways).
</blockquote></details>

Added a second commit that switches from the term `inclusive` in this package to `inverted`. Added a comment.

---

*[pkg/kv/range_cache.go, line 458 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-Kut_n11Lf_RIhxYUEn8:-KvYIDYyGfslglJpKy9h:b-97ogdw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L458)):*

Didn't think we'd ever cache a `nil` `cachedDesc`. How does that happen?
</blockquote></details>

We won't cache a `nil` descriptor, but `getCachedRangeDescriptorLocked` [returns `nil`](https://github.com/cockroachdb/cockroach/blob/a07fe20c2bb29210a446bafd6a942b258ad9e4db/pkg/kv/range_cache.go#L527) if it doesn't find a descriptor for the desired key (for instance, because of an eviction).

---

*[pkg/kv/range_cache.go, line 471 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-KuorOyJrgqienDQZB4I:-KvYIdZir-kd3ikeMvjM:b-88z01c) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L471)):*

I agree, and I can't remember there being a strong reason to evict unconditionally.

The inconsistent reads might've been premature optimization. Would be nice to see them go, especially given how unpredictable they are. They could be a good fallback for the case of unavailable meta ranges (to lessen the impact on the cluster), but let's see if that's ever necessary.
</blockquote></details>

I opened #18988 to follow this.

---

*[pkg/kv/range_cache.go, line 510 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709#-Kuta3P9Wn_cRrR8iqmm:-KvYIwiKLaeKJ6w-MTH4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5cebbf3a2c7e5c66329562de956295ceca6e7da0/pkg/kv/range_cache.go#L510)):*

optio-nit: `false /* !inclusive */`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709)*

tschottdorf:
Reviewed 2 of 2 files at r2, 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18709)*

Issue #17500, opened 2017-08-07
kv: make disk I/O asynchronous with respect to Raft state machine


nvanbenschoten:
⚠️ **EDIT: The original optimization proposed here was implemented in https://github.com/cockroachdb/cockroach/pull/38954. See https://github.com/cockroachdb/cockroach/issues/17500#issuecomment-425689048 for the remainder of this issue.**

Raft, along with most consensus protocols in the Paxos family, distinguishes committed entries from executed entries. An entry is committed when the Raft protocol has completed and the entry has been sufficiently replicated and persisted in the Raft log. The entry is then executed when it is applied to a given replica's state machine. This distinction is important because in cases where the execution of an entry onto the state machine does not produce a result, it is not necessary to wait for execution before sending a commit notification to clients.

Currently in Cockroach, the proposing replica [waits until command execution](https://github.com/cockroachdb/cockroach/blob/2e76b6e567311330041c97e2f564ece8485feefd/pkg/storage/replica.go#L4220) before responding to clients. However, changes made for PropEval KV assured that all MVCC related logic is made upstream of Raft and that by the time we reach entry execution, we're simply applying a `WriteBatch` to RockDB. While some of these execution steps can create `ReplicaCorruptionError`s, I don't think it's necessary or possibly even correct that we attach these errors to the proposal response itself. This is because the entry has already been replicated through Raft, so a local `ReplicaCorruptionError` doesn't mean that all replicas are corrupted or that the command failed. In fact, after looking at the code I don't think that `proposalResult` needs anything populated at execution time for correctness. Because of this, I have a suspicion that all of the time spent in `applyRaftCommand`, including writing to RocksDB and performing stats computations, is unnecessary latency.

Preliminary results from a (very) [rough draft of the change](https://github.com/nvanbenschoten/cockroach/commit/06b4cba909e5a79900833f5443580b1b85e47694) show a **2-3%** improvement on average latency for the [KV](https://github.com/cockroachdb/loadgen/tree/master/kv) benchmark with a `--batch` size of 4:

_Without change (batch=4):_
```
_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
  600.0s        0         597164          995.3     32.2     32.5     41.9     65.0    939.5

BenchmarkBlocks	  597164	   1004753.2 ns/op
```

_With change (batch=4):_
```
[...]

PR #18710, opened 2017-09-23
[DNM] storage: respond to Raft proposals after entries commit, not execute

nvb:
This change addresses the first optimization discussed in #17500.

The change seems to work and provides a modest performance boost.
Unfortunately, I don't think we'll want to consider merging it at
the moment. The problem is that while it is technically safe to
respond to clients before before performing the Raft command
application, doing so is a nightmare for testing. Pretty much every
test in the `storage` package expects to be able to perform an
operation and then "reach beneath raft" immediately to operate
on the result. This can range from inspecting Raft entries to
working on the most up-to-date `Replica` state.

To support this change, all of these tests would need to be
updated to handle the now asynchronous operations performed
in `handleEvalResultRaftMuLocked`. I addressed this by adding
a testing knob called `DisableRaftRespBeforeApplication` in
this change. The problem is that I don't feel very comfortable
with it because we basically need to use it for all tests
(indirectly through `multiTestContext` and `LocalTestCluster`)
which means that we probably aren't testing this optimization
thoroughly. We could disable the optimization on a finer
granularity but this would become a serious issue for
maintainability and I'm not sure it would be worth it.

Perhaps there's some middle ground between returning to the
client after performing in-memory state updates but before
performing persistent state updates? Something like calling:
1. `handleEvalResultRaftMuLocked`
2. `maybeRespondToClient`
3. `applyRaftCommand`

This would solve a lot of the testing issues present here without
the need to use the `DisableRaftRespBeforeApplication` knob, but
I'm almost certain that wouldn't be safe to do.

I think #15648 will run into a similar issue to this. We'll either
need to block clients while we combine Raft batches or we'll need
to update tests which expect a client response to be an indication
that the command has already been applied in all cases. Things
might not be as bad in that case though because less is being
done asynchronously.

### DIFF
--- a/pkg/ccl/storageccl/add_sstable_test.go
+++ b/pkg/ccl/storageccl/add_sstable_test.go
@@ -102,7 +102,7 @@ func runTestDBAddSSTable(ctx context.Context, t *testing.T, db *client.DB) {
 		if err := testutils.MatchInOrder(tracing.FormatRecordedSpans(collect()),
 			"evaluating AddSSTable",
 			"sideloadable proposal detected",
-			"ingested SSTable at index",
+			// "ingested SSTable at index",
 		); err != nil {
 			t.Fatal(err)
 		}
@@ -153,7 +153,7 @@ func runTestDBAddSSTable(ctx context.Context, t *testing.T, db *client.DB) {
 				"evaluating AddSSTable",
 				"target key range not empty, will merge existing data with sstable",
 				"sideloadable proposal detected",
-				"ingested SSTable at index",
+				// "ingested SSTable at index",
 			); err != nil {
 				t.Fatal(err)
 			}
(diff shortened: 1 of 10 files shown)


petermattis:
I ran a few quick tests using this PR and it showed a 1-2% boost in throughput. What are you seeing in your testing?

nvanbenschoten:
Depending on the workload, anywhere from a 1-10% decrease in average latency. I was primarily benchmarking with KV where the larger the block size, the larger the latency boost became. This is in-line with my expectation because the larger the block size is, the more work we're avoiding on each command before returning to clients. 

I'm actually a little surprised you saw any movement in throughput though as we're still doing the same amount of work overall. That must mean that write-batch application is not the bottleneck of whatever benchmark you were running.

petermattis:
I was running `kv --read-percent 0` with the default value size (1 byte). The throughput change might have just been noise. I didn't notice any change in avg, p50, p95 or p99 latency.

PR #18717, opened 2017-09-24
storage: remove /Local/RangeID/X/u/RaftLastIndex

petermattis:
Replace /Local/RangeID/X/u/RaftLastIndex with a reverse scan of the Raft
log for the range to find the last index. This removes the writing of 1
key/value pair every time Raft tells us to append to the Raft log.

No attempt is to remove this key if it already exists on a replica, we
just stop using it. Since the key is unreplicated, as normal replica
rebalancing takes place the key will eventually get deleted.

See #18659

### DIFF
--- a/pkg/server/node.go
+++ b/pkg/server/node.go
@@ -194,6 +194,9 @@ func bootstrapCluster(
 	sender := kv.NewTxnCoordSender(cfg.AmbientCtx, cfg.Settings, stores, cfg.Clock, false, stopper, txnMetrics)
 	cfg.DB = client.NewDB(sender, cfg.Clock)
 	cfg.Transport = storage.NewDummyRaftTransport(cfg.Settings)
+	if err := cfg.Settings.InitializeVersion(bootstrapVersion); err != nil {
+		return uuid.UUID{}, errors.Wrap(err, "while initializing cluster version")
+	}
 	for i, eng := range engines {
 		sIdent := roachpb.StoreIdent{
 			ClusterID: clusterID,
--- a/pkg/server/version_cluster_test.go
+++ b/pkg/server/version_cluster_test.go
@@ -108,13 +108,13 @@ func setupMixedCluster(
 	return testClusterWithHelpers{t, tc}
 }
 
-func TestClusterVersionUpgrade1_0To1_1(t *testing.T) {
+func TestClusterVersionUpgrade1_0To1_2(t *testing.T) {
 	defer leaktest.AfterTest(t)()
 	ctx := context.Background()
 
-	// Four nodes that are all compatible with 1.0, but are really 1.1. This is
-	// what the official v1.1 binary will look like.
-	versions := [][2]string{{"1.0", "1.1"}, {"1.0", "1.1"}, {"1.0", "1.1"}, {"1.0", "1.1"}}
+	// Four nodes that are all compatible with 1.0, but are really 1.2. This is
+	// what the official v1.2 binary will look like.
+	versions := [][2]string{{"1.0", "1.2"}, {"1.0", "1.2"}, {"1.0", "1.2"}, {"1.0", "1.2"}}
 
 	// Start by running 1.0.
 	bootstrapVersion := cluster.ClusterVersion{
@@ -135,6 +135,7 @@ func TestClusterVersionUpgrade1_0To1_1(t *testing.T) {
 		bootstrapVersion.MinimumVersion, // v1.0
 		{Major: 1, Unstable: 500},
 		{Major: 1, Minor: 1},
+		{Major: 1, Minor: 2},
 	} {
 		curVersion := tc.getVersionFromSelect(0)
 		isNoopUpdate := curVersion == newVersion.String()
--- a/pkg/settings/cluster/cluster_version.go
+++ b/pkg/settings/cluster/cluster_version.go
@@ -24,7 +24,7 @@ var (
 	BinaryMinimumSupportedVersion = VersionBase
 
 	// BinaryServerVersion is the version of this binary.
-	BinaryServerVersion = VersionStatsBasedRebalancing
+	BinaryServerVersion = VersionRaftLastIndex
 )
 
 // List all historical versions here in reverse chronological order, with
@@ -33,6 +33,9 @@ var (
 // NB: when adding a version, don't forget to bump ServerVersion above (and
 // perhaps MinimumSupportedVersion, if necessary).
 var (
+	// VersionRaftLastIndex is https://github.com/cockroachdb/cockroach/pull/18717.
+	VersionRaftLastIndex = roachpb.Version{Major: 1, Minor: 1, Unstable: 1}
+
 	// VersionStatsBasedRebalancing is https://github.com/cockroachdb/cockroach/pull/16878.
 	VersionStatsBasedRebalancing = roachpb.Version{Major: 1, Minor: 0, Unstable: 3}
 
@@ -47,3 +50,9 @@ var (
 	// this version is used.
 	VersionBase = roachpb.Version{Major: 1}
 )
+
+// IsActive returns true if the features of the supplied version are active at
+// the running version.
+func (cv ClusterVersion) IsActive(v roachpb.Version) bool {
+	return !cv.UseVersion.Less(v)
+}
--- a/pkg/settings/cluster/settings.go
+++ b/pkg/settings/cluster/settings.go
@@ -153,7 +153,7 @@ func (ecv *ExposedClusterVersion) BootstrapVersion() ClusterVersion {
 // IsActive returns true if the features of the supplied version are active at
 // the running version.
 func (ecv *ExposedClusterVersion) IsActive(v roachpb.Version) bool {
-	return !ecv.Version().UseVersion.Less(v)
+	return ecv.Version().IsActive(v)
 }
 
 // MakeTestingClusterSettings returns a Settings object that has had its version
--- a/pkg/sql/logictest/testdata/logic_test/show_source
+++ b/pkg/sql/logictest/testdata/logic_test/show_source
@@ -94,7 +94,7 @@ sql.trace.txn.enable_threshold                     0s             d     duration
 trace.debug.enable                                 false          b     if set, traces for recent requests can be seen in the /debug page
 trace.lightstep.token                              ·              s     if set, traces go to Lightstep using this token
 trace.zipkin.collector                             ·              s     if set, traces go to the given Zipkin instance (example: '127.0.0.1:9411'); ignored if trace.lightstep.token is set.
-version                                            1.0-3          m     set the active cluster version in the format '<major>.<minor>'.
+version                                            1.1-1          m     set the active cluster version in the format '<major>.<minor>'.
 
 query T colnames
 SELECT * FROM [SHOW SESSION_USER]
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -584,7 +584,7 @@ func newReplica(rangeID roachpb.RangeID, store *Store) *Replica {
 		abortCache:     NewAbortCache(rangeID),
 		pushTxnQueue:   newPushTxnQueue(store),
 	}
-	r.mu.stateLoader = makeReplicaStateLoader(rangeID)
+	r.mu.stateLoader = makeReplicaStateLoader(r.store.cfg.Settings, rangeID)
 	if leaseHistoryMaxEntries > 0 {
 		r.leaseHistory = newLeaseHistory()
 	}
@@ -609,7 +609,7 @@ func newReplica(rangeID roachpb.RangeID, store *Store) *Replica {
 		},
 	)
 	r.raftMu.timedMutex = makeTimedMutex(raftMuLogger)
-	r.raftMu.stateLoader = makeReplicaStateLoader(rangeID)
+	r.raftMu.stateLoader = makeReplicaStateLoader(r.store.cfg.Settings, rangeID)
 	return r
 }
 
@@ -4587,7 +4587,8 @@ func (r *Replica) applyRaftCommand(
 
 	var assertHS *raftpb.HardState
 	if util.RaceEnabled && rResult.Split != nil && r.store.cfg.Settings.Version.IsActive(cluster.VersionSplitHardStateBelowRaft) {
-		oldHS, err := loadHardState(ctx, r.store.Engine(), rResult.Split.RightDesc.RangeID)
+		rsl := makeReplicaStateLoader(r.store.cfg.Settings, rResult.Split.RightDesc.RangeID)
+		oldHS, err := rsl.loadHardState(ctx, r.store.Engine())
 		if err != nil {
 			log.Fatalf(ctx, "unable to load HardState: %s", err)
 		}
@@ -4600,7 +4601,8 @@ func (r *Replica) applyRaftCommand(
 
 	if assertHS != nil {
 		// Load the HardState that was just committed (if any).
-		newHS, err := loadHardState(ctx, r.store.Engine(), rResult.Split.RightDesc.RangeID)
+		rsl := makeReplicaStateLoader(r.store.cfg.Settings, rResult.Split.RightDesc.RangeID)
+		newHS, err := rsl.loadHardState(ctx, r.store.Engine())
 		if err != nil {
 			log.Fatalf(ctx, "unable to load HardState: %s", err)
 		}
--- a/pkg/storage/replica_command.go
+++ b/pkg/storage/replica_command.go
@@ -611,13 +611,11 @@ func declareKeysEndTransaction(
 				EndKey: rightRangeIDUnreplicatedPrefix.PrefixEnd(),
 			})
 
-			leftStateLoader := makeReplicaStateLoader(st.LeftDesc.RangeID)
 			spans.Add(SpanReadOnly, roachpb.Span{
-				Key: leftStateLoader.RangeLastReplicaGCTimestampKey(),
+				Key: keys.RangeLastReplicaGCTimestampKey(st.LeftDesc.RangeID),
 			})
-			rightStateLoader := makeReplicaStateLoader(st.RightDesc.RangeID)
 			spans.Add(SpanReadWrite, roachpb.Span{
-				Key: rightStateLoader.RangeLastReplicaGCTimestampKey(),
+				Key: keys.RangeLastReplicaGCTimestampKey(st.RightDesc.RangeID),
 			})
 
 			spans.Add(SpanReadOnly, roachpb.Span{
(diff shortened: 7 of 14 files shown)


petermattis:
Am I missing anything subtlety here? This was too easy.

petermattis:
Ah, the test failures indicate I should gate not storing `RaftLastIndex` on a version upgrade.

bdarnell:
Another thing that would have been simpler if we had had reverse scans from the beginning. We should probably do a quick audit to see if there are other opportunities like this. I think this should be fine once you add the version gate.

petermattis:
Plumbed the version gate, though ran into a bit of a problem with a `log.Fatal` in `ExposedClusterVersion.IsActive`.

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/settings/cluster/settings.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutKQrdUfg1LSIauyak:-KutKQrdUfg1LSIauyal:b-x1tlwt) ([raw file](https://github.com/cockroachdb/cockroach/blob/18785333343254eb2d285f053a051ce330f30007/pkg/settings/cluster/settings.go#L144)):*
> ```Go
> 	// if (v == ClusterVersion{}) {
> 	// 	log.Fatal(context.Background(), "Version() was called before having been initialized")
> 	// }
> ```

@tschottdorf Need your advice here. This fires when calling `replicaStateLoader.setLastIndex` from `Store.BootstrapRange` (which seems fixable by using `BootstrapVersion`), but also from `Replica.append` which I don't see how to fix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-:-KutJxjbJNOswJN7aD9A:b-log155)*

tschottdorf:
> Ah, the test failures indicate I should gate not storing RaftLastIndex on a version upgrade.

Just curious, which test failed and how? I assume it was one that ran the new binary and rolled back to the old one, but I'd be interested which one.


Code looks good so far, `PR.Lock()` so I can figure out your Fatalf.

---

Reviewed 6 of 12 files at r1, 6 of 6 files at r2.

---

*[pkg/server/version_cluster_test.go, line 138 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutbpLwb3KzTjIHEmOd:-KutbpLwb3KzTjIHEmOe:b-j4b05w) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/server/version_cluster_test.go#L138)):*
> ```Go
> 		{Major: 1, Unstable: 500},
> 		{Major: 1, Minor: 1},
> 		{Major: 1, Minor: 2},
> ```

Hmm, would be nice if this test extended itself automagically. I think we can do that if we make a slice of versions instead of the constants. Not for this PR, but I've made a note to myself at https://github.com/cockroachdb/cockroach/issues/18740.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-:-Kutd4UatdZcD1nOtNE8:b-okvt0n)*

tschottdorf:
`PR.Unlock()`.

---

---

*[pkg/settings/cluster/settings.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutKQrdUfg1LSIauyak:-KutgewHZzlf8l6-fX8P:bldvkjc) ([raw file](https://github.com/cockroachdb/cockroach/blob/18785333343254eb2d285f053a051ce330f30007/pkg/settings/cluster/settings.go#L144)):*

@tschottdorf Need your advice here. This fires when calling `replicaStateLoader.setLastIndex` from `Store.BootstrapRange` (which seems fixable by using `BootstrapVersion`), but also from `Replica.append` which I don't see how to fix.
</blockquote></details>

```diff
diff --git a/pkg/server/node.go b/pkg/server/node.go
index 6d07bc0b2..899e70926 100644
--- a/pkg/server/node.go
+++ b/pkg/server/node.go
@@ -194,6 +194,9 @@ func bootstrapCluster(
        sender := kv.NewTxnCoordSender(cfg.AmbientCtx, cfg.Settings, stores, cfg.Clock, false, stopper, txnMetrics)
        cfg.DB = client.NewDB(sender, cfg.Clock)
        cfg.Transport = storage.NewDummyRaftTransport(cfg.Settings)
+       if err := cfg.Settings.InitializeVersion(bootstrapVersion); err != nil {
+               return uuid.UUID{}, errors.Wrap(err, "while initializing cluster version")
+       }
        for i, eng := range engines {
                sIdent := roachpb.StoreIdent{
                        ClusterID: clusterID,
```

and `make test` passes. The problem was that we weren't at all initializing the cluster version during bootstrap, but it's trivial there because you have one node which you want to run at exactly the bootstrap version (typically the real version of the binary).

---

*[pkg/storage/replica_command.go, line 614 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KuthU1tz1CvXspXPE4N:-KuthU1tz1CvXspXPE4O:batq3w4) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_command.go#L614)):*
> ```Go
> 			})
> 
> 			leftStateLoader := makeReplicaStateLoader(nil, st.LeftDesc.RangeID)
> ```

I'd much prefer if we didn't do this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-:-KuthKsh7ox2VDFSZhtJ:bxgwn5w)*

nvanbenschoten:
Reviewed 6 of 12 files at r1, 3 of 6 files at r2.

---

*[pkg/settings/cluster/settings.go, line 148 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutdBgEbEBcXAgEpnAV:-KutdBgEbEBcXAgEpnAW:b-l2jqgb) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/settings/cluster/settings.go#L148)):*
> ```Go
> }
> 
> // VersionOrMinimumSupported returns the minimum clusger version the caller may
> ```

cluster

---

*[pkg/settings/cluster/settings.go, line 149 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutdEjlzAwvox694Lu_:-KutdEjlzAwvox694Lua:b-i2c5b4) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/settings/cluster/settings.go#L149)):*
> Unless Version

I don't follow. Unless Version what?

---

*[pkg/settings/cluster/settings.go, line 170 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-Kutd7LFsGaNCA-WtMjL:-Kutd7LFsGaNCA-WtMjM:b-n2huba) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/settings/cluster/settings.go#L170)):*
> ```Go
> }
> 
> // IsActive returns true if the features of the supplied version is active at
> ```

I think this was correct before.

---

*[pkg/storage/replica_command.go, line 614 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutebCRYST55MrOs452:-KutebCSf_J9FaOvAxB-:b-4e5ho3) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_command.go#L614)):*
> ```Go
> 			})
> 
> 			leftStateLoader := makeReplicaStateLoader(nil, st.LeftDesc.RangeID)
> ```

This seems fragile. It's relying on the fact that `raplicaStateLoader.RangeLastReplicaGCTimestampKey` doesn't actually need the `cluster.ClusterVersion`, right? 

---

*[pkg/storage/replica_state.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutgD44b-wPHZP1IvhQ:-KutgD44b-wPHZP1IvhR:b56flyy) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_state.go#L62)):*
> ```Go
> 	}
> 	if st == nil {
> 		rsl.version = cluster.ClusterVersion{
> ```

Do we ever actually want to use this version? If not, this seems like an easy place for a bug to slip in.

---

*[pkg/storage/replica_state.go, line 420 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-Kutf09UgfEZL1EzVGgE:-Kutf09UgfEZL1EzVGgF:b-8ermke) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_state.go#L420)):*
> ```Go
> 	ctx context.Context, eng engine.ReadWriter, lastIndex uint64,
> ) error {
> 	if rsl.version.IsActive(cluster.VersionRaftLastIndex) {
> ```

Is it safe for this to be dependent on the `ClusterVersion.use_version`? What if the cluster gets downgraded? Won't it default back to the previous `LastIndex`, which could be out of date since it's no longer updated at the newer version?

I think that we either need to clear `RaftLastIndexKey` to force a recomputation on a downgrade or we'll need to use the `ClusterVersion.minimum_version`. Actually, the first option won't work if `loadTruncatedState` doesn't take entries past the truncated state into account.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717)*

petermattis:

---

*[pkg/storage/replica_command.go, line 614 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KuthU1tz1CvXspXPE4N:-KutjuEz3jXHdv480dWh:b1p16m8) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_command.go#L614)):*

I'd much prefer if we didn't do this.
</blockquote></details>

Pass in `nil`? Yeah, I'm trying to figure out a way to plumb the settings. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717)*

petermattis:

---

*[pkg/settings/cluster/settings.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutKQrdUfg1LSIauyak:-Kuto4FbKRNlE2q4bgZ9:b-ucfknh) ([raw file](https://github.com/cockroachdb/cockroach/blob/18785333343254eb2d285f053a051ce330f30007/pkg/settings/cluster/settings.go#L144)):*

```diff
diff --git a/pkg/server/node.go b/pkg/server/node.go
index 6d07bc0b2..899e70926 100644
--- a/pkg/server/node.go
+++ b/pkg/server/node.go
@@ -194,6 +194,9 @@ func bootstrapCluster(
        sender := kv.NewTxnCoordSender(cfg.AmbientCtx, cfg.Settings, stores, cfg.Clock, false, stopper, txnMetrics)
        cfg.DB = client.NewDB(sender, cfg.Clock)
        cfg.Transport = storage.NewDummyRaftTransport(cfg.Settings)
+       if err := cfg.Settings.InitializeVersion(bootstrapVersion); err != nil {
+               return uuid.UUID{}, errors.Wrap(err, "while initializing cluster version")
+       }
        for i, eng := range engines {
                sIdent := roachpb.StoreIdent{
                        ClusterID: clusterID,
```

and `make test` passes. The problem was that we weren't at all initializing the cluster version during bootstrap, but it's trivial there because you have one node which you want to run at exactly the bootstrap version (typically the real version of the binary).
</blockquote></details>

Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717)*

petermattis:
Review status: 9 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/settings.go, line 148 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutdBgEbEBcXAgEpnAV:-KutpLD7dXtCWBTx7lBM:b5l4bme) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/settings/cluster/settings.go#L148)):*

cluster
</blockquote></details>

This was no longer necessary and has been removed.

---

*[pkg/settings/cluster/settings.go, line 149 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutdEjlzAwvox694Lu_:-KutpNjZTWuXu83gOFuG:b5l4bme) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/settings/cluster/settings.go#L149)):*

> Unless Version

I don't follow. Unless Version what?
</blockquote></details>

This was no longer necessary and has been removed.

---

*[pkg/storage/replica_command.go, line 614 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutebCRYST55MrOs452:-KutpRlgM6J_L6TwJqKo:b-ktp8oq) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_command.go#L614)):*

This seems fragile. It's relying on the fact that `raplicaStateLoader.RangeLastReplicaGCTimestampKey` doesn't actually need the `cluster.ClusterVersion`, right? 
</blockquote></details>

Yes, it is fragile. I can fix a handful of these cases, but fixing them all sure is irritating. 

---

*[pkg/storage/replica_state.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KutgD44b-wPHZP1IvhQ:-KutpdiT8nFGpHZ0QnVR:b9lcbb7) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_state.go#L62)):*

Do we ever actually want to use this version? If not, this seems like an easy place for a bug to slip in.
</blockquote></details>

Probably not. We should probably be plumbing a `cluster.Settings` everywhere. Working on it.

---

*[pkg/storage/replica_state.go, line 420 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-Kutf09UgfEZL1EzVGgE:-KutpqdQuULfatZ-Ebvd:bx97mha) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_state.go#L420)):*

Is it safe for this to be dependent on the `ClusterVersion.use_version`? What if the cluster gets downgraded? Won't it default back to the previous `LastIndex`, which could be out of date since it's no longer updated at the newer version?

I think that we either need to clear `RaftLastIndexKey` to force a recomputation on a downgrade or we'll need to use the `ClusterVersion.minimum_version`. Actually, the first option won't work if `loadTruncatedState` doesn't take entries past the truncated state into account.
</blockquote></details>

Good point, once you upgrade to 1.2 you won't be able to downgrade to 1.1 (unless we put in code to resuscitate `RaftLastIndex` keys). @tschottdorf is there anything I have to be doing in order to prevent the cluster from downgrading?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717)*

tschottdorf:
Reviewed 4 of 4 files at r3.

---

*[pkg/storage/replica_state.go, line 420 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-Kutf09UgfEZL1EzVGgE:-KuttL8YMx17cKiy_02H:b-8zvz2w) ([raw file](https://github.com/cockroachdb/cockroach/blob/e29a660edc999496880d5e958c46d1eb5aa005ba/pkg/storage/replica_state.go#L420)):*

Good point, once you upgrade to 1.2 you won't be able to downgrade to 1.1 (unless we put in code to resuscitate `RaftLastIndex` keys). @tschottdorf is there anything I have to be doing in order to prevent the cluster from downgrading?
</blockquote></details>

The way `UseVersion` will work (well, if we ever actually implement it - right now, `UseVersion == MinimumVersion`) is that the `UseVersion` can trail `MinimumVersion` and features which can be downgraded use `UseVersion` and others (like this one) `MinimumVersion`.

We could introduce a second API now (`IsActiveDowngradeable`?). However, if we introduce `UseVersion` in, say, 1.3, , it doesn't matter for any features from previous cycles, so I'm tempted to say "no".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717)*

petermattis:
Ok, this is ready for another look.

---

Review status: 8 of 14 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-:-Kuu5VhvMgAUZgG08Tu2:bp9rmyd)*

tschottdorf:
:lgtm:

You expect the reverse scan to be fast because this is a hot key, right? I forget the details, but I seem to remember that revscans are "generally" slower because the LSM is built for forward scans so a little more seeking is involved (?).

---

Reviewed 6 of 6 files at r4.

---

*[pkg/storage/replica_raftstorage.go, line 83 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KuuLdYgBWBoSkMjN4WD:-KuuLdYgBWBoSkMjN4WE:b-iukddt) ([raw file](https://github.com/cockroachdb/cockroach/blob/5e8c98f8f7dca3be7a8a9cfd49cdb95f0ec6dea9/pkg/storage/replica_raftstorage.go#L83)):*
> ```Go
> 	defer readonly.Close()
> 	ctx := r.AnnotateCtx(context.TODO())
> 	return entries(ctx, r.mu.stateLoader, readonly, r.RangeID, r.store.raftEntryCache,
> ```

`mu` is held, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-:-KuuLv9qwfr9xGN04XwB:b-kz2iqd)*

petermattis:
This isn't really a reverse scan: it is a `Seek` followed by `Prev`. This is somewhat slower than looking up a specific key, but `loadLastIndex` is called very infrequently as usually we're using `Replica.mu.lastIndex`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-:-KuuMUikIV5o79htvVec:b14j99x)*

petermattis:

---

*[pkg/storage/replica_raftstorage.go, line 83 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717#-KuuLdYgBWBoSkMjN4WD:-KuuNCW4rVXrESGNmZ-E:b-jblsa9) ([raw file](https://github.com/cockroachdb/cockroach/blob/5e8c98f8f7dca3be7a8a9cfd49cdb95f0ec6dea9/pkg/storage/replica_raftstorage.go#L83)):*

`mu` is held, right?
</blockquote></details>

Yes, `mu` is held for all of the `replicaRaftStorage` methods. See `Replica.raftEntriesLocked` for the same method on `Replica`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18717)*

PR #18718, opened 2017-09-24
libroach,storage/engine: implement MVCCFindSplitKey in C++

petermattis:
Added a facility to automatically generate C++ versions of key and span constants from the Go versions.

```
name                                     old time/op   new time/op   delta
MVCCFindSplitKey_RocksDB/valueSize=32-8    269ms ± 4%    133ms ± 2%   -50.47%  (p=0.000 n=10+10)

name                                     old speed     new speed     delta
MVCCFindSplitKey_RocksDB/valueSize=32-8  249MB/s ± 4%  503MB/s ± 2%  +101.87%  (p=0.000 n=10+10)
```


petermattis:
The implementation is incomplete (hence the WIP) as it doesn't check that the split key is valid, but I doubt that is affecting the speed. `MVCCFindSplitKey` isn't performance critical, but I again find it interesting the speedup we can get by moving a Go implementation into C++. I would bet that we could see a significant speedup by moving `MVCCScan` into C++, though we probably need [cppgo](https://github.com/petermattis/cppgo) in order to avoid awkward contortions in the interface.

bdarnell:
Nice speedup. Can you identify a particular reason it's faster? Are there allocations or copies on the go side that you avoid here? Or is it about minimizing crossings of the cgo boundary?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-:-Kut9mumr_9ubdGYuA9U:boujwgt)*

petermattis:
I think it is the Cgo crossing. There are no significant allocations in the Go version after my previous change to `MVCCFindSplitKey`. Specifically, the Go version performs 2 allocations per call which is insignificant. The structure of the implementation between the Go and C++ versions is identical.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-:-KutBQTLHJd_sjv4-FLV:bsmcfq)*

petermattis:
This is ready for another look. In particular, take a look at `keys/gen_cpp_keys.go` which generates C++ constants for keys and spans.

bdarnell:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---

*[c-deps/libroach/db.cc, line 2414 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO6Ja5uYFwt8NWxE_4:-KvO6Ja5uYFwt8NWxE_5:b-jknbn3) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/c-deps/libroach/db.cc#L2414)):*
> ```cpp
>     DBIterator* iter, DBKey start, DBKey end, int64_t target_size, DBString* split_key) {
>   auto iter_rep = iter->rep.get();
>   std::string start_key = EncodeKey(start);
> ```

Can't this be `const` too?

---

*[c-deps/libroach/keys.h, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO5lVHgUHHnxsr0iRw:-KvO5lVHgUHHnxsr0iRx:b-mtz83e) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/c-deps/libroach/keys.h#L7)):*
> ```objc
> const rocksdb::Slice kMeta2KeyMax("\x03\xff\xff", 3);
> 
> const std::vector<std::pair<rocksdb::Slice, rocksdb::Slice> > kSortedNoSplitSpans = {
> ```

This looks like it would be a problem if this header were included from more than one cc file. 

---

*[pkg/keys/gen_cpp_keys.go, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO4HFkKMd0dr_gtCtc:-KvO4HFkKMd0dr_gtCtd:bambsqi) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/keys/gen_cpp_keys.go#L55)):*
> ```Go
> 	// https://github.com/golang/go/issues/13560#issuecomment-277804473
> 	// https://github.com/Reviewable/Reviewable/wiki/FAQ#how-do-i-tell-reviewable-that-a-file-is-generated-and-should-not-be-reviewed
> 	fmt.Fprintf(f, `// Code generated by gen_cpp_keys.go; DO NOT EDIT.
> ```

Reviewable is flagging this file as generated, presumably because of this string literal. Do we have to do silly things like splitting it up into separate strings that get concatenated?

---

*[pkg/storage/engine/engine.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO7PvLWG_NRN69t9yG:-KvO7PvLWG_NRN69t9yH:b-pc1nvv) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/storage/engine/engine.go#L98)):*
> ```Go
> 	ComputeStats(start, end MVCCKey, nowNanos int64) (enginepb.MVCCStats, error)
> 	// FindSplitKey finds a split key from the given key range that aims to
> 	// roughly cut into targetSize bytes the left-hand side of the split. The
> ```

s/into targetSize bytes/targetSize bytes from/

---

*[pkg/storage/engine/mvcc.go, line 2272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO8Wm5-Jt0JtRQlH-t:-KvO8Wm5-Jt0JtRQlH-u:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/storage/engine/mvcc.go#L2272)):*
> ```Go
> 
> // MVCCFindSplitKey finds a split key from the given key range that aims to
> // roughly cut into targetSize bytes the left-hand side of the split. The
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-:-KvO8lHEmnIa1BidWRkA:bnfp4nl)*

petermattis:
Review status: 7 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[c-deps/libroach/db.cc, line 2414 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO6Ja5uYFwt8NWxE_4:-KvOGovkm-RgILgZ7uHG:bjg6x97) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/c-deps/libroach/db.cc#L2414)):*

Can't this be `const` too?
</blockquote></details>

Yep. Done.

---

*[c-deps/libroach/keys.h, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO5lVHgUHHnxsr0iRw:-KvOGsJj6cOpvCNc3D6V:b-b9hrya) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/c-deps/libroach/keys.h#L7)):*

This looks like it would be a problem if this header were included from more than one cc file. 
</blockquote></details>

Yes, we'd need to convert these to `extern const` and add a corresponding `.cc` file. I'd like to punt on doing that until it is necessary.

---

*[pkg/keys/gen_cpp_keys.go, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO4HFkKMd0dr_gtCtc:-KvOHD3KZ3udxws6_29c:b-fg0o0u) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/keys/gen_cpp_keys.go#L55)):*

Reviewable is flagging this file as generated, presumably because of this string literal. Do we have to do silly things like splitting it up into separate strings that get concatenated?
</blockquote></details>

Ok, string concatenation appeared to work.

---

*[pkg/storage/engine/engine.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO7PvLWG_NRN69t9yG:-KvOICAqLadzxoNWvDzV:b-7wp3pb) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/storage/engine/engine.go#L98)):*

s/into targetSize bytes/targetSize bytes from/
</blockquote></details>

Hmm, it reads better to me without the `from`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718)*

bdarnell:
Reviewed 2 of 2 files at r2.

---

*[c-deps/libroach/keys.h, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO5lVHgUHHnxsr0iRw:-KvOgcLeBvptAfOnpB5o:b-r9sz8) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/c-deps/libroach/keys.h#L7)):*

Yes, we'd need to convert these to `extern const` and add a corresponding `.cc` file. I'd like to punt on doing that until it is necessary.
</blockquote></details>

That's fine.

---

*[pkg/keys/gen_cpp_keys.go, line 56 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvOhCwQ6UIRr9efd7YY:-KvOhCwQ6UIRr9efd7YZ:b-mvv77g) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6116d58594908c81a2b4cc0b0ed0dc7004cb053/pkg/keys/gen_cpp_keys.go#L56)):*
> ```Go
> 	// https://github.com/Reviewable/Reviewable/wiki/FAQ#how-do-i-tell-reviewable-that-a-file-is-generated-and-should-not-be-reviewed
> 	//
> 	// The funk string concatenation is to foil reviewable so that it doesn't
> ```

s/funk/funky/

---

*[pkg/storage/engine/engine.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO7PvLWG_NRN69t9yG:-KvOfnXhPLeA2tVEtY5S:br72lb0) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/storage/engine/engine.go#L98)):*

Hmm, it reads better to me without the `from`. 
</blockquote></details>

It's the "into" that bothers me - it was "into half", but now "into targetSize" sounds weird to me. "FindSplitKey finds a key from the given span such that the left side of the split is roughly targetSize bytes"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718)*

petermattis:

---

*[pkg/keys/gen_cpp_keys.go, line 56 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvOhCwQ6UIRr9efd7YY:-KvOqs_BSr_LAHgMSTSp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6116d58594908c81a2b4cc0b0ed0dc7004cb053/pkg/keys/gen_cpp_keys.go#L56)):*

s/funk/funky/
</blockquote></details>

Done.

---

*[pkg/storage/engine/engine.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO7PvLWG_NRN69t9yG:-KvOr6ZUzTW9yCaviDJ-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/storage/engine/engine.go#L98)):*

It's the "into" that bothers me - it was "into half", but now "into targetSize" sounds weird to me. "FindSplitKey finds a key from the given span such that the left side of the split is roughly targetSize bytes"
</blockquote></details>

Done.

---

*[pkg/storage/engine/mvcc.go, line 2272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718#-KvO8Wm5-Jt0JtRQlH-t:-KvOr7Bi6Z5cDcRMg9o3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8bf67e8ffbc8152e1141c87f5cd3e60a91674120/pkg/storage/engine/mvcc.go#L2272)):*

ditto
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18718)*

PR #18727, opened 2017-09-24
cherrypick 1.1: distsql: use ASCENDING encoding for comparisons

rjnn:
Cherry-pick of #18702. 

As the comment on EncDatum.Encode indicates, using DatumEncoding_VALUE
encoding is not valid for comparisons, as the encoded datum can
contain a ColumnID prefix.


andreimatei:
Lgtm

PR #18732, opened 2017-09-25
rfcs: integrating PM concerns earlier on in RFC process

dianasaur323:
As Raphael suggested, there have been some top of mind issues on the product side that I think would be great to address in the RFC process. Namely, in addition to focusing on implementation, to also discuss the user experience. Otherwise, these discussions happen after implementation, when changes are more expensive. Also, introducing a more holistic view of how the RFC'd feature fits into the rest of the product -> again, this has happened in our current release cycle, where additional features like adding metrics to the admin UI were squeezed in after feature freeze. We should only consider a feature to be complete when it fits into the rest of the product cohesively. Finally, given our current focus on performance, we should identify some things we want to measure so that we don't introduce features that negatively impact the overall performance of the database.

### DIFF
--- a/docs/RFCS/00000000_template.md
+++ b/docs/RFCS/00000000_template.md
@@ -30,7 +30,7 @@ Explain the proposal as if it was already included in the project and
 you were teaching it to another CockroachDB programmer. That generally means:
 
 - Introducing new named concepts.
-- Explaining the feature largely in terms of examples.
+- Explaining the feature largely in terms of examples. Take into account that a product manager (PM) will want to connect back the work introduced by the RFC with user stories. Whenever practical, do ask PMs if they already have user stories that relate to the proposed work, and do approach PMs to attract user buy-in and mindshare if applicable.
 - Explaining how CockroachDB contributors and users should think about
   the feature, and how it should impact the way they use
   CockroachDB. It should explain the impact as concretely as possible.
@@ -51,6 +51,11 @@ This is the technical portion of the RFC. Explain the design in sufficient detai
 (You may replace the section title if the intent stays clear.)
 
 - Its interaction with other features is clear.
+- It covers where this feature may be surfaced in other areas of the product
+   - If the change influences a user-facing interface, make sure to preserve consistent user experience (UX). Prefer to avoid UX changes altogether unless the RFC also argues for a clear UX benefit to users. If UX has to change, then prefer changes that match the UX for related features, to give a clear impression to users of homogeneous CLI / GUI elements. Avoid UX surprises at all costs. If in doubt, ask for input from other engineers with past UX design experience and from your design department.
+- It considers how to monitor the success and quality of the feature.
+   - Your RFC must consider and propose a set of metrics to be collected, if applicable, and suggest which metrics would be useful to users and which need to be exposed in a public interface.
+   - Your RFC should outline how you propose to investigate when users run into related issues in production. If you propose new data structures, suggest how they should be checked for consistency. If you propose new asynchronous subsystems, suggest how a user can observe their state via tracing. In general, think about how your coworkers and users will gain access to the internals of the change after it has happened to either gain understanding during execution or troubleshoot problems.
 - It is reasonably clear how the feature would be implemented.
 - Corner cases are dissected by example.
 
@@ -74,6 +79,10 @@ other active RFCs.
 Why should we *not* do this?
 
 If applicable, list mitigating factors that may make each drawback acceptable.
+
+Investigate the consequences of the proposed change onto other areas of CockroachDB. If other features are impacted, especially UX, list this impact as a reason not to do the change. If possible, also investigate and suggest mitigating actions that would reduce the impact. You can for example consider additional validation testing, additional documentation or doc changes, new user research, etc.
+
+Also investigate the consequences of the proposed change on performance. Pay especially attention to the risk that introducing a possible performance improvement in one area can slow down another area in an unexpected way. Examine all the current "consumers" of the code path you are proposing to change and consider whether the performance of any of them may be negatively impacted by the proposed change. List all these consequences as possible drawbacks.
 
 ## Rationale and Alternatives
 


docs/RFCS/00000000_template.md
```diff
@@ -30,7 +30,8 @@ Explain the proposal as if it was already included in the project and
 you were teaching it to another CockroachDB programmer. That generally means:
 
 - Introducing new named concepts.
-- Explaining the feature largely in terms of examples.
+- Explaining the feature largely in terms of examples. If applicable, stick to PM provided user stories
```
nstewart: I'm wary about saying "stick to PM provided user stories" since right now we haven't had the bandwidth to fully flesh out features. maybe "if applicable, tie out (or maybe align) the RFC with PM provided user stories"?

docs/RFCS/00000000_template.md
```diff
@@ -51,6 +52,11 @@ This is the technical portion of the RFC. Explain the design in sufficient detai
 (You may replace the section title if the intent stays clear.)
 
 - Its interaction with other features is clear.
+- It covers where this feature may be surfaced in other areas of the product
+   - What metrics should be measured? Should they be surfaced in the admin UI?
+   - How will we debug issues? Do these metrics need to be surfaced anywhere?
```
nstewart: What do we mean by "metrics" here and are you distinguishing this from telemetry?

docs/RFCS/00000000_template.md
```diff
@@ -51,6 +52,11 @@ This is the technical portion of the RFC. Explain the design in sufficient detai
...
+   - Does the user interaction with this feature map to our existing design
+     paradigms?
```
nstewart: What are our "existing design paradigms"? @kuanluo do we have a style guide (not just for the brand, but for the UX) anywhere?

docs/RFCS/00000000_template.md
```diff
@@ -75,6 +81,10 @@ Why should we *not* do this?
 
 If applicable, list mitigating factors that may make each drawback acceptable.
 
+If drawbacks include potential impacts on other areas of the product, identify tests
+to run before marking the feature as complete to make sure we don't introduce incomplete
```
nstewart: This may be worth talking in person; I don't think running tests solves the problem here. This is getting out and actually talking to folks too. Potentially even comms plans if this changes customer behavior. Materially changing other areas of the product is a big deal.

nstewart (commented):
Left comments.

awoods187:

---

*[docs/RFCS/00000000_template.md, line 105 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-KutfOO3lCGugH5zfBKp:-KutfOO3lCGugH5zfBKq:b-cpl6lz) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L105)):*
> ```Markdown
> - What related issues do you consider out of scope for this RFC that
>   could be addressed in the future independently of the solution that
>   comes out of this RFC?
> ```

Is it worth having a section for specific signoff from clients? For example, we committed to providing partitioning by EOY to a particular client, would we want to have them "sign off" within an RFC? Or is that better served outside of the RFC?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732)*

docs/RFCS/00000000_template.md
knz: :+1: on nate's suggestion. However, let's not completely forget what's driving what in an open source project. May I suggest instead:

"""

- Explaining the feature largely in terms of examples. Take into account that a product manager (PM) will want to connect back the work introduced by the RFC with user stories. Whenever practical, do ask PMs if they already have user stories that relate to the proposed work, and do approach PMs to attract user buy-in and mindshare if applicable.

"""

docs/RFCS/00000000_template.md
knz: Nate: The word "metric" is used throughout the project (on the engineering side) with an unambiguous meaning, and yes this is separate from telemetry.

FYI / reminder: the "metrics" are samples of instantaneous measurements in a CockroachDB cluster, collected over time and stored in our internal time series database. Telemetry usually performs aggregate queries over the stored metrics and reports a summary to our reg server.

knz: Diana: do not phrase what you want as questions in here. So instead of:

"What metrics should be measured? Should they be surfaced in the admin UI?"

Write:

"Your RFC must consider and propose a set of metrics to be collected, if applicable, and suggest which metrics would be useful to users and need to be exposed in a public interface."

(Also avoid using the phrase "the admin UI" when you really mean "a public interface". The Admin UI is a specific sub-product and CockroachDB provides many more ways to expose data to users than just the admin UI.)

knz: ditto for the other thing about debugging:

"Your RFC should outline how you propose to investigate when users run into related issues in production. If you propose new data structures, suggest how they should be checked for consistency. If you propose new asynchronous subsystems, suggest how a user can observe their state via tracing. In general, think about how your coworkers and users will gain access to the internals of the change after it has happened to either gain understanding during execution or troubleshoot problems."

docs/RFCS/00000000_template.md
knz: Using the word "design" here is so ambiguous as to be unproductive / destructive about what needs to happen. Also, no questions!

May I suggest instead:

"
If the change influences a user-facing interface, make sure to preserve consistent user experience (UX). Prefer to avoid UX changes altogether unless the RFC also argues for a clear UX benefit to users. If UX has to change, then prefer changes that match tbe UX for related features, to give a clear impression to users of homogeneous CLI / GUI elements. Avoid UX surprises at all costs. If in doubt, ask for input from other engineers with past UX design experience and from your design department.
"

docs/RFCS/00000000_template.md
knz: Yeah the phrasing is also dangerous, because if a feature does not have natural "tests" this requirement can completely stall progress on the RFC.

I would suggest instead:

"
Investigate the consequences of the proposed change onto other areas of CockroachDB. If other features are impacted, especially UX, list this impact as a reason not to do the change. If possible, also investigate and suggest mitigating actions that would reduce the impact. You can for example consider additional validation testing, additional documentation or doc changes, new user research, etc.
"

knz (changes requested):

docs/RFCS/00000000_template.md
knz: Also I just noticed your initial remark about performance regression. This needs a separate paragraph altogether:

"
Also investigate the consequences of the proposed change on performance. Pay especially attention to the risk that introducing a possible performance improvement in one area can slow down another area in unexpected way. Examine all the current "consumers" of the code path you are proposing to change and consider whether any of them may be negatively impacted by the proposed change. List all these consequences as possible drawbacks.
"

dianasaur323:
*[docs/RFCS/00000000_template.md, line 33 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-Kuq_Yv-D_NFpy4mejTi-r1-33:-KwlMV4x2d-Hghgud_wp:b2k4i52) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L33)):*

:+1: on nate's suggestion. However, let's not completely forget what's driving what in an open source project. May I suggest instead:

"""

- Explaining the feature largely in terms of examples. Take into account that a product manager (PM) will want to connect back the work introduced by the RFC with user stories. Whenever practical, do ask PMs if they already have user stories that relate to the proposed work, and do approach PMs to attract user buy-in and mindshare if applicable.

"""
</blockquote></details>

Ok, I think this makes sense!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732)*

dianasaur323:

---

*[docs/RFCS/00000000_template.md, line 57 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-Kuq_Yv-D_NFpy4mejTi-r1-57:-KwlNWN5LGrDvGXsU3Ho:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L57)):*

ditto for the other thing about debugging:

"Your RFC should outline how you propose to investigate when users run into related issues in production. If you propose new data structures, suggest how they should be checked for consistency. If you propose new asynchronous subsystems, suggest how a user can observe their state via tracing. In general, think about how your coworkers and users will gain access to the internals of the change after it has happened to either gain understanding during execution or troubleshoot problems."
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_template.md, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-Kuq_Yv-D_NFpy4mejTi-r1-59:-KwlNVzbf8KqqTJDH0V4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L59)):*

Using the word "design" here is so ambiguous as to be unproductive / destructive about what needs to happen. Also, no questions!

May I suggest instead:

"
If the change influences a user-facing interface, make sure to preserve consistent user experience (UX). Prefer to avoid UX changes altogether unless the RFC also argues for a clear UX benefit to users. If UX has to change, then prefer changes that match tbe UX for related features, to give a clear impression to users of homogeneous CLI / GUI elements. Avoid UX surprises at all costs. If in doubt, ask for input from other engineers with past UX design experience and from your design department.
"
 
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_template.md, line 85 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-Kuq_Yv-D_NFpy4mejTi-r1-85:-KwlNgREKi41LRvfRkIo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L85)):*

Also I just noticed your initial remark about performance regression. This needs a separate paragraph altogether:

"
Also investigate the consequences of the proposed change on performance. Pay especially attention to the risk that introducing a possible performance improvement in one area can slow down another area in an unexpected way. Examine all the current "consumers" of the code path you are proposing to change and consider whether the performance of any of them may be negatively impacted by the proposed change. List all these consequences as possible drawbacks.
"
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_template.md, line 105 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-KutfOO3lCGugH5zfBKp:-KwlNiCgGM0vRerHQNJ4:bawr2fm) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L105)):*

Is it worth having a section for specific signoff from clients? For example, we committed to providing partitioning by EOY to a particular client, would we want to have them "sign off" within an RFC? Or is that better served outside of the RFC?
</blockquote></details>

Hmm.. I think probably better reserved for outside the RFC for now, and maybe implemented as part of our PM process?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732)*

knz:
All right, I'm good with this. Now on to get buy-in from the rest of the team...

---

---

*[docs/RFCS/00000000_template.md, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-Kuq_Yv-D_NFpy4mejTi-r1-59:-KwngBDqMNvSjAQFfm0n:ba3aqyi) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L59)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

typo: tbe -> the

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-:-KwngQGi1T93DWuiUY4C:b-7kxy20)*

knz (approved):

bdarnell:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/00000000_template.md, line 55 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-KwqnKjAGub_p6m03_Zy:-KwqnKjAGub_p6m03_Zz:bwhize4) ([raw file](https://github.com/cockroachdb/cockroach/blob/efb52b28acb3a1c7283998d5f2530bb823fcea33/docs/RFCS/00000000_template.md#L55)):*
> ```Markdown
> - Its interaction with other features is clear.
> - It covers where this feature may be surfaced in other areas of the product
>    - Your RFC must consider and propose a set of metrics to be collected, if applicable, and suggest which metrics would be useful to users and need to be exposed in a public interface.
> ```

I might pull this bullet and the next into a separate top-level section ("Monitoring"?). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-:-KwuACSWdRrcfUpYOr3x:bnfp4nl)*

dianasaur323:
*[docs/RFCS/00000000_template.md, line 55 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-KwqnKjAGub_p6m03_Zy:-KwuQTJxjNa79hPIbNp3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/efb52b28acb3a1c7283998d5f2530bb823fcea33/docs/RFCS/00000000_template.md#L55)):*

I might pull this bullet and the next into a separate top-level section ("Monitoring"?). 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732)*

dianasaur323:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/00000000_template.md, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-Kuq_Yv-D_NFpy4mejTi-r1-59:-KwuQ_zjVVXnqEdFMYRX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a9caee26686d7649355724900da61e82dc1dd2d4/docs/RFCS/00000000_template.md#L59)):*

typo: tbe -> the
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732)*

knz:
You probably need to rebase this to make teamcity happy. Then you can merge.

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18732#-:-KwvjaWnLQNxfLlKhtGG:bgw3vqj)*

a-robinson:
@dianasaur323 How did you merge this to master without a corresponding merge commit?

dianasaur323:
@a-robinson I have no idea! I'm pretty sure I had a commit message when I clicked merge. Is that not something I should be able to do?

arjunravinarayan:
@a-robinson Looks like a squash merge, which is allowed according to our permissions.

dianasaur323:
@arjunravinarayan aren't I supposed to squash merge? I was told to do so before.

jordanlewis:
image
Did you click one of those other options besides the first one?

dianasaur323:
Yes! My default green button is squash and merge. So much scrutiny! Shrivel in a cave.

andreimatei:
Perhaps you were told to squash, but I doubt you were told to squash-merge
since we always do the "create a merge commit" one.

jordanlewis:
Ah, gotcha! I wonder how that default got swapped. Anyway, generally we like to use the `Create a merge commit` option - that's laid out [here](https://github.com/cockroachdb/cockroach/blob/master/docs/first-pr.md#merging). We should add it to `CONTRIBUTING.md` as well - and probably also disable squash + merge since we prefer to avoid it.

mjibson:
FWIW this is news to me. I thought squash and merge would produce a merge commit. I've done this before.

dianasaur323:
Guess what, I learned how to squash commits locally today from Andrew. #professionalwin

jordanlewis:
"Squash and merge" is now disabled - thanks for helping us ensmoothen our process :)

dianasaur323:
Always.

PR #14968, opened 2017-04-16
rfc: add postgres syntax compat RFC

madelynnblue:

PR #18739, opened 2017-09-25
rfc: JSONB Scope

justinj:


docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,98 @@
...
+* Storing data that the schema author has little-to-no control over in a
+  structured, queryable way
+* Having access to CockroachDB's transactions and strong consistency while
+  working with the ergonomics of a Mongo-style loose-schema database
+
+## Design
+
+Postgres supports two data types - `JSON` and `JSONB`.
+Our datatype will be more similar to `JSONB` in its behaviour, and as a result we
+will refer to it as `JSONB`, rather than `JSON`.
```
knz: Add a little footnote below to remind the reader what is the difference between JSON and JSONB in pg.

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,98 @@
...
+towards keeping the door open to split them up if this becomes a bottleneck for
+users.
+
+## Details
+
+### Operators
+
+We plan to support all of Postgres’s JSON manipulation and access operators
+(`->`, `->>`, `||`, `-`, `#-`), in addition to the ones which can be accelerated by
+inverted indexes (`@>`, `?`, `?|`, `?&`).
```
knz: Make a table to explain here what each operator is for, with a small illustrative example.

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,98 @@
...
+don't have much control over, and we should support it.
+There will be a dedicated RFC for inverted indexes in Cockroach.
+
+### Lower Priority Features
+
+There are some nice-to-haves that are less essential for the initial version of
+JSONB.
+
+These include
+* Functions which can be implemented on the client side
```
knz: Provide examples.

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,98 @@
...
+* DistSQL support for JSON aggregates
```
knz: Examples.

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,98 @@
...
+can rewrite JSONPath queries as SQL operators so our existing SQL machinery can
+deal with them efficiently.
+
+### Indexing
+
+Postgres GIN indexing is supported on JSONB columns to speed up queries
+involving the `?`, `?|`, `?&`, and `@>` operators.
+This is an important feature for users dealing with unstructured data they
+don't have much control over, and we should support it.
+There will be a dedicated RFC for inverted indexes in Cockroach.
```
knz: "CockroachDB"

knz (commented):
I am very excited about this proposal. I am fully on board with the proposed scope.

awoods187:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuu476hVXM2LiqaBta_:-Kuu476hVXM2LiqaBtaa:buvsidp) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L74)):*
> ```Markdown
> specifies a sub-language (“the SQL/JSON path language”, or JSONPath) for
> querying data from JSON objects within SQL.
> JSONPath support is not in any current release of Postgres, but is slated for Postgres 11.
> ```

When is 11 slated for release? Would it be valuable to beat them to market? From a marketing perspective, it would likely be nice to be able to show a clear winner of postgres (or at least not something that lags, however slightly, behind. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

justinj:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-69:-KuuE3czf1ciZqxIPTWb:bxcpe6h) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L69)):*

Make a table to explain here what each operator is for, with a small illustrative example.
</blockquote></details>

Done, great idea.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuu476hVXM2LiqaBta_:-Kuu9dljUSgVGeumvof7:b61hz3n) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L74)):*

When is 11 slated for release? Would it be valuable to beat them to market? From a marketing perspective, it would likely be nice to be able to show a clear winner of postgres (or at least not something that lags, however slightly, behind. 
</blockquote></details>

I'm not sure, and the only source I've seen that it's slated for 11 is [this slide deck](https://postgrespro.ru/media/2017/04/04/jsonb-pgconf.us-2017.pdf). I would also love to get this into 1.2, but I think that might be a bit ambitious given all of the other considerations we have for JSONB. I'm happy to not completely write off the possibility, though, if we want to discuss.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 86 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-86:-KuuE5uSRxDouE1tcsvr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L86)):*

"CockroachDB"
</blockquote></details>

Done.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-94:-KuuE6JcxybaHT9tuDif:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L94)):*

Provide examples.
</blockquote></details>

Done.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-95:-KuuE6c63WtxtAg019aY:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L95)):*

Examples.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

justinj:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 52 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-52:-KuuF8G6Il6BQjEE5M_G:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L52)):*

Add a little footnote below to remind the reader what is the difference between JSON and JSONB in pg.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,123 @@
...
+# Summary
+
+Support JSONB in CockroachDB.
+JSONB is a column type supported by Postgres that allows users to store and
+manipulate JSON documents directly in their database.
+It's distinguished from Postgres's plain `JSON` column type in that it stores
+a binary representation of the documents, rather than the raw JSON string.
+The high-level overview of the features we intend to include is:
+* JSON Encoding/Decoding
+* Key Encoding for JSON values
```
benesch: highly optional nit: these first two bullets should use sentence case like the following bullets

tschottdorf:
Good read.

---

Reviewed 1 of 1 files at r2.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-69:-KuuJt_S8O1sBEmJoaHK:b-5z4sp8) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L69)):*

Done, great idea.
</blockquote></details>

So good :thumbsup: 

---

*[docs/RFCS/20170925_jsonb_scope.md, line 66 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuLI39PT0mAa97OBGa:-KuuLI39PT0mAa97OBGb:b-ew3gbw) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L66)):*
> ```Markdown
> with JSONB columns.
> We opt to store JSONB blobs entirely within a single kv entry, with an eye
> towards keeping the door open to split them up if this becomes a bottleneck for
> ```

I don't see the eye below.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuK8PWCbHwb9LAMiGp:-KuuK8PWCbHwb9LAMiGq:b-mow141) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L85)):*
> ```Markdown
> | `||`         | Concatenate JSON arrays, or append to a JSON array.   | `'[1, 2]'::JSONB || '[3]'::JSONB` = `'[1, 2, 3]'::JSONB`                   |
> | `-`          | Removes a key from a JSON object or array.            | `'{"foo": "bar"}'::JSONB - 'foo'` = `'{}'::JSONB`                          |
> | `#-`         | Removes a path from a JSON object or array.           | `'{"foo": {"bar":"baz"}}'::JSONB #- '{foo,bar}'` = `'{"foo": {}}'::JSONB`  |
> ```

`{foo,bar}` is interesting, I would've expected `["foo", "bar"]`. (Just thinking out loud, probably that's just how paths are specified?).

---

*[docs/RFCS/20170925_jsonb_scope.md, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKTbh5HKL1HXmlEmb:-KuuKTbh5HKL1HXmlEmc:b-4bfn4t) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L86)):*
> '{"foo": 4}'

What would this example look like with an actual path? `{{foo, bar}: 4}`? Something else?

---

*[docs/RFCS/20170925_jsonb_scope.md, line 87 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKgoxTM2Koyq1fyo_:-KuuKgoxTM2Koyq1fyoa:b-qrlpi9) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L87)):*
> ```Markdown
> | `#-`         | Removes a path from a JSON object or array.           | `'{"foo": {"bar":"baz"}}'::JSONB #- '{foo,bar}'` = `'{"foo": {}}'::JSONB`  |
> | `@>`         | Path-value check                                      | `'{"foo": 4, "bar": 2}'::JSONB @> '{"foo": 4}'::JSONB"` = `true`           |
> | `?`          | Key-contains check                                    | `'{"foo": 4, "bar": 2}'::JSONB ? 'foo'` = `true`                           |
> ```

Doesn't work with paths?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-:-KuuKyaaLe1WhDFDTpeI:b-oqxlgm)*

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,123 @@
...
+(`->`, `->>`, `||`, `-`, `#-`), in addition to the ones which can be accelerated by
+inverted indexes (`@>`, `?`, `?|`, `?&`).
+
+The behaviour of each of these operators is described below:
+
+| Operator     | Description                                           | Example                                                                    |
+| ------------ | ----------------------------------------------------- | -------------------------------------------------------------------------- |
+| `->`         | Access a JSON field, returning a JSON value.          | `'{"foo":"bar"}'::JSONB->'foo'` = `'bar'::JSONB`                           |
+| `->>`        | Access a JSON field, returning a string.              | `'{"foo":"bar"}'::JSONB->>'foo'` = `'bar'::STRING`                         |
+| `||`         | Concatenate JSON arrays, or append to a JSON array.   | `'[1, 2]'::JSONB || '[3]'::JSONB` = `'[1, 2, 3]'::JSONB`                   |
```
benesch: This row is broken in the rendered output because `|` is interpreted as a Markdown table column separator. If you care, I think you can fix it by using `<code>&#124;&#124;</code>` instead at the cost of making the plain-text version unreadable.

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,123 @@
...
+| `-`          | Removes a key from a JSON object or array.            | `'{"foo": "bar"}'::JSONB - 'foo'` = `'{}'::JSONB`                          |
+| `#-`         | Removes a path from a JSON object or array.           | `'{"foo": {"bar":"baz"}}'::JSONB #- '{foo,bar}'` = `'{"foo": {}}'::JSONB`  |
+| `@>`         | Path-value check                                      | `'{"foo": 4, "bar": 2}'::JSONB @> '{"foo": 4}'::JSONB"` = `true`           |
+| `?`          | Key-contains check                                    | `'{"foo": 4, "bar": 2}'::JSONB ? 'foo'` = `true`                           |
+| `?|`         | Some key-contains check                               | `'{"foo": 4, "bar": 2}'::JSONB ?| '{foo,baz}'` = `true`                    |
```
benesch: (This operator also breaks the table rendering.)

docs/RFCS/20170925_jsonb_scope.md
benesch: That's covered by the next sentence, isn't it? (Unless that sentence was added to address your comment?)

benesch (commented):
:tada: no substantive feedback!

knz:
Reviewed 1 of 1 files at r2.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 42 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuVJQyaCnk4OZ7-mUw:-KuuVJQyaCnk4OZ7-mUx:b-h07c1v) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L42)):*
> ```Markdown
> Reasons one might want to use JSON with CockroachDB include:
> * Ease of storing and manipulating existing JSON data, without having to do a
>   manual conversion to a relational schema
> ```

same nit as nikhil: sentences in bullet points means a punctuation at the end. I am personally fan of having a single running sentence from the start of the paragraph above to the last bullet, with no capitals, semicolons after each bullet point and a period after the last. Jesse OTOH likes separate sentences for each point, with a period at the end of each. At any rate, punctuation is missing here.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 88 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r2-88:-KuuW0_Z-3CzGA-A_2iA:b-8qo7g2) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L88)):*

(This operator also breaks the table rendering.)
</blockquote></details>

Markdown is such a piece of trash. I personally wouldn't mind making an example of the benefits of rST with this RFC.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

justinj:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 52 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r1-52:-KuuWR0itYSGFKA1wCwi:bvsbipa) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8cbe4a7981090f403a0fdd03585e323fc2fbade/docs/RFCS/20170925_jsonb_scope.md#L52)):*

That's covered by the next sentence, isn't it? (Unless that sentence was added to address your comment?)
</blockquote></details>

It was, yeah.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 17 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r2-17:-KuuX4MvwQIdE1GdndIp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L17)):*

highly optional nit: these first two bullets should use sentence case like the following bullets
</blockquote></details>

Done.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 42 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuVJQyaCnk4OZ7-mUw:-KuuWqVUmhEo96agQ6LJ:b-phux8u) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L42)):*

same nit as nikhil: sentences in bullet points means a punctuation at the end. I am personally fan of having a single running sentence from the start of the paragraph above to the last bullet, with no capitals, semicolons after each bullet point and a period after the last. Jesse OTOH likes separate sentences for each point, with a period at the end of each. At any rate, punctuation is missing here.
</blockquote></details>

I agree, will leave the formatting decision here up to @m-schneider.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 83 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r2-83:-KuuVgPJW6Q5oMpxtbt1:b-7wyanm) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L83)):*

This row is broken in the rendered output because `|` is interpreted as a Markdown table column separator. If you care, I think you can fix it by using `<code>&#124;&#124;</code>` instead at the cost of making the plain-text version unreadable.
</blockquote></details>

😩 (fixed)

---

*[docs/RFCS/20170925_jsonb_scope.md, line 88 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r2-88:-KuuW87JeJq8jV-kTmYJ:b-7wyanm) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L88)):*

Markdown is such a piece of trash. I personally wouldn't mind making an example of the benefits of rST with this RFC.
</blockquote></details>

😩 (fixed)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

m-schneider:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuK8PWCbHwb9LAMiGp:-KuuO5KDaT7nZWJejaz-:b-u37cbn) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L85)):*

`{foo,bar}` is interesting, I would've expected `["foo", "bar"]`. (Just thinking out loud, probably that's just how paths are specified?).
</blockquote></details>

Your example is an array. What this is syntax is saying is remove remove field bar from field foo.  If the example was an actual array so:
["foo", {"bar":"baz"}] then it would look like '["foo", {"bar":"baz"}]':JSONB  #- ' {1, "bar"}', which would remove field bar from index 1.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKTbh5HKL1HXmlEmb:-KuuYHi8rjeaiU5sTTX2:b-5l4p3x) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L86)):*

> '{"foo": 4}'

What would this example look like with an actual path? `{{foo, bar}: 4}`? Something else?
</blockquote></details>

SELECT '[1, 2, [1, 3]]'::jsonb @> '[[1, 3]]'::jsonb; 

---

*[docs/RFCS/20170925_jsonb_scope.md, line 87 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKgoxTM2Koyq1fyo_:-KuuTaJ6ccn1UBri2a1u:b-st6rj9) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L87)):*

Doesn't work with paths?
</blockquote></details>

? doesn't, you would use @>.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

m-schneider:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 42 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuVJQyaCnk4OZ7-mUw:-KuuYihUKC3e7wGW-o6g:b-3jd22h) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L42)):*

I agree, will leave the formatting decision here up to @m-schneider.
</blockquote></details>

I agree with Jesse, I like separate sentences with periods.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

justinj:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 66 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuLI39PT0mAa97OBGa:-Kuu_bQQ8J7CcrlxgfRO:b-2nok0v) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L66)):*

I don't see the eye below.
</blockquote></details>

This comment refers to making the binary encoding able to be forwards-compatible with some future multi-kv format, I'll make that explicit.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuK8PWCbHwb9LAMiGp:-KuuZ0sw_Sr6slxyWld2:b-tnb8qt) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L85)):*

Your example is an array. What this is syntax is saying is remove remove field bar from field foo.  If the example was an actual array so:
["foo", {"bar":"baz"}] then it would look like '["foo", {"bar":"baz"}]':JSONB  #- ' {1, "bar"}', which would remove field bar from index 1.
</blockquote></details>

The syntax is because the right argument to `#-`, `'{foo,bar}'`, is a postgres `ARRAY`, not a postgres JSON array. `ARRAY['foo', 'bar']` would also work.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKTbh5HKL1HXmlEmb:-KuuZNkSLQJMU7sM0Gg5:blputfp) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L86)):*

SELECT '[1, 2, [1, 3]]'::jsonb @> '[[1, 3]]'::jsonb; 
</blockquote></details>

re: tobi's example, `'{"foo":{"bar": 4},"baz":2}' @> '{"foo":{"bar":4}}'` is true.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 87 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKgoxTM2Koyq1fyo_:-KuuZvZHn6ZsMC6TyHRV:b-l3dpma) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L87)):*

? doesn't, you would use @>.
</blockquote></details>

You could also do `obj->'foo' ? 'bar'`, which would only be optimized if you had a (computed) GIN index on `obj->'foo'`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

tschottdorf:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

benesch:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

knz:

---

*[docs/RFCS/20170925_jsonb_scope.md, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuuKTbh5HKL1HXmlEmb:-KuumQiuU5f4phXhRHST:b-gymc4k) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L86)):*

re: tobi's example, `'{"foo":{"bar": 4},"baz":2}' @> '{"foo":{"bar":4}}'` is true.
</blockquote></details>

add the example too then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

docs/RFCS/20170925_jsonb_scope.md
```diff
@@ -0,0 +1,123 @@
...
+JSONPath support is not in any current release of Postgres, but is slated for Postgres 11.
+This is out of scope for 1.2, but may be in scope for 1.3.
+The challenges of supporting JSONPath are primarily implementing it in a way that
+can rewrite JSONPath queries as SQL operators so our existing SQL machinery can
+deal with them efficiently.
+
+### Indexing
+
+Postgres GIN indexing is supported on JSONB columns to speed up queries
+involving the `?`, `?|`, `?&`, and `@>` operators.
```
vivekmenezes: It's valuable to explain these operators

vivekmenezes (commented):
Thanks for putting this together!

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-Kuuv24bqtqWHagLhnjG:bslqs56) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*
> ```Markdown
> We plan to support all of Postgres’s JSON manipulation and access operators
> (`->`, `->>`, `||`, `-`, `#-`), in addition to the ones which can be accelerated by
> inverted indexes (`@>`, `?`, `?|`, `?&`).
> ```

Am I the only one who finds these operators ugly? I realize there is a strong argument for Postgres compatibility, I'm just having trouble accepting these operators.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

vilterp:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuuxP0f1mQwJRnLsIFi:bthtgi3) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

Am I the only one who finds these operators ugly? I realize there is a strong argument for Postgres compatibility, I'm just having trouble accepting these operators.
</blockquote></details>

Yeah, they're not pretty… In my experience of using JSON in Postgres, `->>` is the most commonly used, which isn't that bad.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

benesch:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-Kuv1C8yxSIpKdPnezqi:bxchni8) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

Yeah, they're not pretty… In my experience of using JSON in Postgres, `->>` is the most commonly used, which isn't that bad.
</blockquote></details>

Perhaps you'd prefer OCaml's build-your-own-operator approach? https://stackoverflow.com/a/6150853/1122351

:trollface: 

---

*[docs/RFCS/20170925_jsonb_scope.md, line 107 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuta7aTX7jAxccWcuoI-r4-107:-Kuv1NFV5qlQN4IthToR:bg4f83i) ([raw file](https://github.com/cockroachdb/cockroach/blob/b150cb4fb0af62e9a4edc67c1a97e5c1309a7bec/docs/RFCS/20170925_jsonb_scope.md#L107)):*

It's valuable to explain these operators
</blockquote></details>

I believe they're explained in the table above?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-Kuuz6mBk5avRN49n28p:bahxp8r) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

Perhaps you'd prefer OCaml's build-your-own-operator approach? https://stackoverflow.com/a/6150853/1122351

:trollface: 
</blockquote></details>

@benesch That is horrifying.

@vilterp Interesting to know. Makes me sad to have both `->` and `->>` which are so visually similar. I'm curious if there is a technical reason why they didn't use `.` for field access. Might have something to do with the extensible way operators are implemented in postgres.

Can we make `.` work for field access using something that looks more like name resolution. Consider this example which works in Postgres:

```
drop table test;
create table test (
  v jsonb
);

insert into test (v) values ('{"foo": {"bar": {"blah": 1}}}');
select v->'foo'->'bar'->'blah' from test;
```

That `select` is quite awkward. It would be great if it could be written as:

```
select v.foo.bar.blah from test;
```

Getting back to the usage of `->` and `->>`, if we had dot-notation and didn't have to use strings for the fields, we wouldn't need 2 operators: `->>` could be written as `select v.foo.bar::string from test`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuvUTbckL-ooFdejIwp:bvpdkio) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

@benesch That is horrifying.

@vilterp Interesting to know. Makes me sad to have both `->` and `->>` which are so visually similar. I'm curious if there is a technical reason why they didn't use `.` for field access. Might have something to do with the extensible way operators are implemented in postgres.

Can we make `.` work for field access using something that looks more like name resolution. Consider this example which works in Postgres:

```
drop table test;
create table test (
  v jsonb
);

insert into test (v) values ('{"foo": {"bar": {"blah": 1}}}');
select v->'foo'->'bar'->'blah' from test;
```

That `select` is quite awkward. It would be great if it could be written as:

```
select v.foo.bar.blah from test;
```

Getting back to the usage of `->` and `->>`, if we had dot-notation and didn't have to use strings for the fields, we wouldn't need 2 operators: `->>` could be written as `select v.foo.bar::string from test`.
</blockquote></details>

Peter the dot syntax is not an operator and it's not handled in the same phase of semantic analysis (dots are about name resolution; the other things are about expression operators). I am not keen on having a syntax for something that is either a scoping operator or an access operator, and have to disambiguate that.

OTOH I would be happy using a colon. `v.foo:bar:blah` feels nice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuvX9RHo5PyH4qWTGS2:b-co7fxg) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

Peter the dot syntax is not an operator and it's not handled in the same phase of semantic analysis (dots are about name resolution; the other things are about expression operators). I am not keen on having a syntax for something that is either a scoping operator or an access operator, and have to disambiguate that.

OTOH I would be happy using a colon. `v.foo:bar:blah` feels nice.
</blockquote></details>

Isn't `v.foo.bar` about name resolution too? You're resolving the name of the json field. Certainly, column name resolution and json field name resolution would happen in different parts of the code, but that doesn't imply they need different syntax.

FYI, Spanner uses dot-notation for struct field access (https://cloud.google.com/spanner/docs/query-syntax).

If we're to use any syntax other than Postgres, I think we'd want to use something that appears natural in other mainstream programming languages. Dot-notation meets that criteria, colon doesn't (for me). But before we bikeshed this, we should figure out if diverging from Postgres is feasible or if users care about the specific Postgres syntax.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuvZWnXTwHwaK-i1K8c:b-wwl5ct) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

Isn't `v.foo.bar` about name resolution too? You're resolving the name of the json field. Certainly, column name resolution and json field name resolution would happen in different parts of the code, but that doesn't imply they need different syntax.

FYI, Spanner uses dot-notation for struct field access (https://cloud.google.com/spanner/docs/query-syntax).

If we're to use any syntax other than Postgres, I think we'd want to use something that appears natural in other mainstream programming languages. Dot-notation meets that criteria, colon doesn't (for me). But before we bikeshed this, we should figure out if diverging from Postgres is feasible or if users care about the specific Postgres syntax.
</blockquote></details>

TBH I do not care much about the end result, but I currently take issue with the methodology.
If we are introducing support for the pg operators, then the change is orthogonal to the current code and can use the same infrastructure.

If we want to overload the dot operator, then the RFC must thoroughly investigate what needs to be changed in the current code, research alternatives, and highlight the possible consequences (e.g. performance) of altering the code path for queries that do not use JSON. As long as the RFC does not investigate these aspects, I'm :-1: overall on the overload.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuyynxF5gX8WF-eYHnF:beo7xrj) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

TBH I do not care much about the end result, but I currently take issue with the methodology.
If we are introducing support for the pg operators, then the change is orthogonal to the current code and can use the same infrastructure.

If we want to overload the dot operator, then the RFC must thoroughly investigate what needs to be changed in the current code, research alternatives, and highlight the possible consequences (e.g. performance) of altering the code path for queries that do not use JSON. As long as the RFC does not investigate these aspects, I'm :-1: overall on the overload.
</blockquote></details>

If we invent our own syntax for json operations, it will be the *third* we intend to support (after the postgres-compatible operators and the SQL standard JSONPath stuff). Even though I bet we can come up with a nicer syntax than either of those, I don't think we can improve on it by enough to justify introducing yet another variant. 

Even if we can co-opt some other syntax for this today, we'd potentially be diverging in a way that may make future postgres/standard compatibility more difficult. 

Introducing named functions equivalent to these operators might be worthwhile, since it makes things less opaque without opening up any tricky issues of syntax, but I wouldn't go any further than that. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuzNX_i3TbuYL990FMr:b-wwwbor) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

If we invent our own syntax for json operations, it will be the *third* we intend to support (after the postgres-compatible operators and the SQL standard JSONPath stuff). Even though I bet we can come up with a nicer syntax than either of those, I don't think we can improve on it by enough to justify introducing yet another variant. 

Even if we can co-opt some other syntax for this today, we'd potentially be diverging in a way that may make future postgres/standard compatibility more difficult. 

Introducing named functions equivalent to these operators might be worthwhile, since it makes things less opaque without opening up any tricky issues of syntax, but I wouldn't go any further than that. 
</blockquote></details>

I'll reiterate my earlier question of whether we need to support the postgres-compatible operators. I'd much prefer a set of named functions that are equivalent to the operators.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

justinj:
Thanks for the comments so far everyone! I just pushed a change to a discussion section at the end including my reasoning for why I think computed indexes should be upgraded from a low-priority item and why I think they're an important feature for us to have alongside JSONB columns - I would appreciate another look through!

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuzOpTMTa70B7yE-613:b-2jqbvh) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

I'll reiterate my earlier question of whether we need to support the postgres-compatible operators. I'd much prefer a set of named functions that are equivalent to the operators.
</blockquote></details>

I think it's important to support the postgres operators; this is one of the bigger things keeping us from being drop-in compatible with existing applications. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

knz:
Reviewed 1 of 1 files at r5.

---

*[docs/RFCS/20170925_jsonb_scope.md, line 182 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuzR1wpPo0trHsPvhAk:-KuzR1wq0RGZhhkni2_f:b4fnv5i) ([raw file](https://github.com/cockroachdb/cockroach/blob/912ed52841b1c654de2c62e7c3ea99c0b0560dd5/docs/RFCS/20170925_jsonb_scope.md#L182)):*
> ```Markdown
> blob-style field, computed indexes allow one to place a GIN index *only on that
> field*, rather than having to manually lift it into its own column which is
> then GIN-indexed.
> ```

Good arguments. I'm convinced! (Disclaimer: I was already convinced.)

@danhhz @benesch I recall you had other arguments in favor of prioritizing computed columns/indexes too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

petermattis:

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuzSIAnYRap0DMIttQZ:bs4j6d3) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

I think it's important to support the postgres operators; this is one of the bigger things keeping us from being drop-in compatible with existing applications. 
</blockquote></details>

Huh, I missed drop-in compatibility with Postgres JSONB as a requirement. Is that drive by some user story?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

bdarnell:

---

*[docs/RFCS/20170925_jsonb_scope.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-Kuuv24bqtqWHagLhnjF:-KuzUDmRNxWNUhTmhYyP:b-g6efu0) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c278181e24268a6320055fd3cafd40c0e034869/docs/RFCS/20170925_jsonb_scope.md#L75)):*

Huh, I missed drop-in compatibility with Postgres JSONB as a requirement. Is that drive by some user story?
</blockquote></details>

It's something I've run into when trying to get open source applications that support postgres to run on CRDB. And we have an RFC where we agreed to aim for a high level of compatibility with postgres (#14968). Drop-in compatibility is not an absolute requirement, but we need a better reason to deviate from postgres than "those operators are ugly"

---

*[docs/RFCS/20170925_jsonb_scope.md, line 182 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739#-KuzR1wpPo0trHsPvhAk:-KuzTCeSigyoup1_F7Ms:bfplztw) ([raw file](https://github.com/cockroachdb/cockroach/blob/912ed52841b1c654de2c62e7c3ea99c0b0560dd5/docs/RFCS/20170925_jsonb_scope.md#L182)):*

Good arguments. I'm convinced! (Disclaimer: I was already convinced.)

@danhhz @benesch I recall you had other arguments in favor of prioritizing computed columns/indexes too.
</blockquote></details>

+1 to computed indexes (they sound like a higher priority than GIN indexes for JSON columns). Partial indexes would also be good to have and the implementation might be related.

Computed columns can also make it possible to use foreign keys that come from json. However, I'm not convinced about the need for computed columns. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18739)*

justinj:
Final comment period!

a-robinson (approved):
👍

justinj:
Thanks for the reviews everyone!

PR #18743, opened 2017-09-25
sql,keys: bugfix to rowfetcher for COUNT(*)

### LINKED ISSUE #18705: sql: COUNT(*) returns odd results
Version `CockroachDB CCL v1.1-beta.20170907-120-g6f7e3b0`

I backed up some data from reg and restored it on sapphire and did a quick sanity check count(*) and was surprised to see 4m rows. I tried to narrow the date range where the new rows appeared and instead found pretty much any WHERE clause got rid of them, even a `NULL OR NOT NULL`:
```
> SELECT count(*) FROM updates WHERE timestamp IS NULL OR timestamp IS NOT NULL;
+----------+
| count(*) |
+----------+
|  1436341 |
+----------+
(1 row)

Time: 2.381654597s

> SELECT count(*) FROM updates;
+----------+
| count(*) |
+----------+
|  4308600 |
+----------+
(1 row)

Time: 8.987759834s

root@localhost:26257/registration>
```

jordanlewis:
d5c9932 introduced a critical bug to rowfetcher that prevented it from
correctly understanding tables with multiple column families when no
columns from the table were marked as needed, such as in the case of
`COUNT(*)`.

This commit fixes the bug by reusing the logic from EnsureSafeSplitKey
to advance the rowfetcher's key to the family id section as it must,
without needing to decode any of the prefix of the key.

Also add a test to ensure COUNT(*) and multiple column families interact
properly.

Fixes #18705.

### DIFF
--- a/pkg/sql/sqlbase/rowfetcher.go
+++ b/pkg/sql/sqlbase/rowfetcher.go
@@ -22,6 +22,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/internal/client"
+	"github.com/cockroachdb/cockroach/pkg/keys"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/sql/parser"
 	"github.com/cockroachdb/cockroach/pkg/util"
@@ -266,6 +267,14 @@ func (rf *RowFetcher) NextKey(ctx context.Context) (rowDone bool, err error) {
 				// interleaved data from some other table or index.
 				continue
 			}
+		} else {
+			// We still need to consume the key until the family id, so processKV can
+			// know whether we've finished a row or not.
+			prefixLen, err := keys.GetRowPrefixLength(rf.kv.Key)
+			if err != nil {
+				return false, err
+			}
+			rf.keyRemainingBytes = rf.kv.Key[prefixLen:]
 		}
 
 		// For unique secondary indexes, the index-key does not distinguish one row
--- a/pkg/keys/keys.go
+++ b/pkg/keys/keys.go
@@ -615,22 +615,24 @@ func MakeFamilyKey(key []byte, famID uint32) []byte {
 	return encoding.EncodeUvarintAscending(key, uint64(len(key)-size))
 }
 
-// EnsureSafeSplitKey transforms an SQL table key such that it is a valid split key
-// (i.e. does not occur in the middle of a row).
-func EnsureSafeSplitKey(key roachpb.Key) (roachpb.Key, error) {
-	if encoding.PeekType(key) != encoding.Int {
-		// Not a table key, so already a split key.
-		return key, nil
-	}
-
+// GetRowPrefixLength returns the length of the row prefix of the key. A table
+// key's row prefix is defined as the maximal prefix of the key that is also a
+// prefix of every key for the same row. (Any key with this maximal prefix is
+// also guaranteed to be part of the input key's row.)
+// For secondary index keys, the row prefix is defined as the entire key.
+func GetRowPrefixLength(key roachpb.Key) (int, error) {
 	n := len(key)
+	if encoding.PeekType(key) != encoding.Int {
+		// Not a table key, so the row prefix is the entire key.
+		return n, nil
+	}
 	// The column ID length is encoded as a varint and we take advantage of the
 	// fact that the column ID itself will be encoded in 0-9 bytes and thus the
 	// length of the column ID data will fit in a single byte.
 	buf := key[n-1:]
 	if encoding.PeekType(buf) != encoding.Int {
 		// The last byte is not a valid column ID suffix.
-		return nil, errors.Errorf("%s: not a valid table key", key)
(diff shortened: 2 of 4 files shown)


knz:
Wow, *very* nice work.

pkg/sql/sqlbase/rowfetcher.go
```diff
@@ -266,6 +267,13 @@ func (rf *RowFetcher) NextKey(ctx context.Context) (rowDone bool, err error) {
 				// interleaved data from some other table or index.
 				continue
 			}
+		} else {
+			// We still need to consume the key until the family id, so processKV can
+			// know whether we've finished a row or not.
+			rf.keyRemainingBytes, err = keys.GetFamilySuffix(rf.kv.Key)
+			if err != nil {
+				return false, err
+			}
```
vivekmenezes: Don't quite follow the need for this. I wonder if we need to update a test in sqlbase as well

vivekmenezes (commented):
Nice find!

pkg/sql/sqlbase/rowfetcher.go
jordanlewis: The reason is that `rf.keyRemainingBytes` is used by `processKV` to calculate whether a row has been finished or not. That's required since each row can be split up into multiple column families, each of which lives in its own key.

I would love to augment some rowfetcher tests. However, they do not exist. Adding a unit test suite for rowfetcher would be a lot of work and is probably out of scope for this bug fix. On the plus side, rowfetcher is extensively tested, since it is used by every single sql feature that exercises any table or index reads.

vivekmenezes: Gotcha. Didn't know you introduced mustDecodeIndexKey in the prior change; I was surprised this bug was around all along, and it wasn't. Thanks

dt (approved):

danhhz:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/keys/keys.go, line 621 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18743#-KuzF5zFHeEKr9g-uP5r:-KuzF5zG6T1p2DPwsQFV:b-h8h3d8) ([raw file](https://github.com/cockroachdb/cockroach/blob/91531a4459e40961f03a81bcb5f4a3722cfbd462/pkg/keys/keys.go#L621)):*
> ```Go
> // the key is not a table key, the index returned is the end of the key.
> func findFamilyIndex(key roachpb.Key) (int, error) {
> 	if encoding.PeekType(key) != encoding.Int {
> ```

This logic is specific to EnsureSafeSplitKey, so it should be moved there. findFamilyIndex may then want to return an error if it gets a non-table key

---

*[pkg/keys/keys.go, line 659 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18743#-KuzFeC9CKUfOMk8M--y:-KuzFeC9CKUfOMk8M--z:bh3xc2l) ([raw file](https://github.com/cockroachdb/cockroach/blob/91531a4459e40961f03a81bcb5f4a3722cfbd462/pkg/keys/keys.go#L659)):*
> ```Go
> 
> // GetFamilySuffix returns a byte slice of the input table key containing just
> // the family id/column id length suffix.
> ```

for families other than 0, this also includes the family id length

I also think this would be more obvious as something like GetRowPrefixLen where "prefix" is defined as "anything with this prefix is in the same row and everything in the same row has this prefix".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18743#-:-KuzH1lxhQ3evS4q0uvg:bnfp4nl)*

jordanlewis:

---

*[pkg/keys/keys.go, line 621 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18743#-KuzF5zFHeEKr9g-uP5r:-KuzqSRApjCk7UNcH8wi:b29z85s) ([raw file](https://github.com/cockroachdb/cockroach/blob/91531a4459e40961f03a81bcb5f4a3722cfbd462/pkg/keys/keys.go#L621)):*

This logic is specific to EnsureSafeSplitKey, so it should be moved there. findFamilyIndex may then want to return an error if it gets a non-table key
</blockquote></details>

Improved the comments here.

---

*[pkg/keys/keys.go, line 659 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18743#-KuzFeC9CKUfOMk8M--y:-KuzdrZCos3qKjUKUYpz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/91531a4459e40961f03a81bcb5f4a3722cfbd462/pkg/keys/keys.go#L659)):*

for families other than 0, this also includes the family id length

I also think this would be more obvious as something like GetRowPrefixLen where "prefix" is defined as "anything with this prefix is in the same row and everything in the same row has this prefix".
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18743)*

jordanlewis:
TFTRs!

PR #18746, opened 2017-09-25
sqlccl: gate IMPORT behind a cluster setting

### LINKED ISSUE #18608: backup/restore: bump import csv to beta status with cluster setting
Based on our discussion, we are moving import csv to beta so that users have a experience that matches their expectations. We will work on adding usability polish to import csv in 1.2

@jseldess this needs to be documented. Also, we need to mention that import csv will be free while in beta period. also cc @sploiselle 

@cuongdo @danhhz for triage.

madelynnblue:
Fixes #18608

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -31,6 +31,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/server"
 	"github.com/cockroachdb/cockroach/pkg/server/serverpb"
+	"github.com/cockroachdb/cockroach/pkg/settings"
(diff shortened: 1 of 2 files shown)


pkg/ccl/sqlccl/csv.go
```diff
@@ -738,14 +739,30 @@ func importJobDescription(
 	return parser.AsStringWithFlags(&stmt, parser.FmtSimpleQualified), nil
 }
 
+const importCSVEnabledSetting = "experimental.importcsv.enabled"
```
danhhz: are we calling this "experimental" or "beta"?

danhhz (approved):

pkg/ccl/sqlccl/csv.go
mjibson: experimental. See https://github.com/cockroachdb/cockroach/issues/18608#issuecomment-331277940

PR #18750, opened 2017-09-25
server: fix gzipResponseWriter Flush, avoid all allocations

nvb:
This contains three small changes, some of which are being pulled out of
#18521:
- `gzipResponseWriter.Flush` was a no-op. We now properly
flush the `gzip.Writer` and underlying flusher if possible.
- properly return an error from the writer's `Close`
method.
- before we used a `sync.Pool` to avoid allocations for
`*gzip.Writers`. However, we always wrapped these writers in
a `gzipResponseWriter`, which we allocated for each connection.
We now embed the `gzip.Writer` in the `gzipResponseWriter` and
put `*gzipResponseWriter` objects in the `sync.Pool`.

@bdarnell since you already had some of this paged in from before. The answer to the question you had in https://github.com/cockroachdb/cockroach/pull/18521#issuecomment-330284385 is that `gzip.Writer.Reset` discards all of the writer's state, so it's safe to put back in the pool even if `Close` fails.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1223,21 +1223,21 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case strings.Contains(ae, httputil.GzipEncoding):
 		w.Header().Set(httputil.ContentEncodingHeader, httputil.GzipEncoding)
 		gzw := newGzipResponseWriter(w)
-		defer gzw.Close()
+		defer func() {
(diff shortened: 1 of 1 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18750#-:-KuzJkoL7WPe3IAfExB9:bnfp4nl)*

PR #18761, opened 2017-09-25
cherrypick 1.1: cli,distsqlrun: Turn disk usage limits from cluster settings to flags

andreimatei:
cherry-pick of #18686
cc @cockroachdb/release 

Before this patch, the cluster setting sql.distsql.temp_storage.max_bytes and
sql.distsql.temp_storage.max_percentage were controlling how much
storage capacity could be used by distsql temp storage.
This didn't make sense: the settings should be per node, not per
cluster.

This patch replaces both settings with --max-disk-temp-storage flag. It
can take either a value or a percentage. One option briefly considered
was allowing the control over this temp storage through the --store
flag which already has an expressive language for specifying limits.
This option has the upside that it'd make it clear that the temp storage
is collocated with the first store and they share resources - in fact we
could easily prevent broken configuration where both the store and the
temp storage are configured to take the whole disk. However it's unclear
if tying temp storage to stores is a great idea thinking forward, and
also it's too late for me to explore that option more for 1.1.

This patch also changes the interpretation of the flag/cluster setting
when it's expressed as a percentage and the first store is type=mem.
Before, it used to be interpreted at p% of 512MB. Now, it will be
interpreted as p% of the internal memory.

The patch also has the effect that the dir for the first store will be
created earlier in the code path than it used to be, with permissions
0755. As far as I can tell, before the patch, the dir would have been
created later with the same permissions either when when initialized the
log dir or, if (--log_dir was specified) when we created the engine.

### DIFF
--- a/pkg/base/store_spec.go
+++ b/pkg/base/store_spec.go
@@ -43,7 +43,10 @@ const MinimumStoreSize = 10 * 64 << 20
 // StoreSpec contains the details that can be specified in the cli pertaining
 // to the --store flag.
 type StoreSpec struct {
-	Path        string
+	Path string
+	// SizeInBytes is used for calculating free space and making rebalancing
+	// decisions. Zero indicates that there is no maximum size. This value is not
+	// actually used by the engine and thus not enforced.
 	SizeInBytes int64
 	SizePercent float64
 	InMemory    bool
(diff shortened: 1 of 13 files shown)


asubiotto:
cc @sploiselle re temp storage cluster settings becoming flags

bdarnell:
:lgtm:

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18761#-:-KuzK_0k6tpJQotnr77F:bnfp4nl)*

PR #18762, opened 2017-09-25
cherrypick-1.1: sql: note unimplemented errors raised after parsing as well

dt:
previously we were only recording unimplemented errors encountered during parsing, but we return them in some branches of evaluation as well.
I briefly started trying to add logging to execParsed, execRequest and looking for other possible paths.
That felt brittle and error prone though, as a new path could easily be missed, whereas the common pgwire helper handles all error responses.

### DIFF
--- a/pkg/server/updates_test.go
+++ b/pkg/server/updates_test.go
@@ -124,8 +124,18 @@ func TestReportUsage(t *testing.T) {
 				t.Fatal(err)
 			}
 		}
+		if _, err := db.Exec(`SELECT 1::INTERVAL(1)`); !testutils.IsError(
+			err, "unimplemented",
+		) {
+			t.Fatal(err)
+		}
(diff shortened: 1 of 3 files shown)


knz:
cc @cockroachdb/release

knz (approved):

bdarnell:
I don't think this meets the "would we patch this into 1.1.1" test for cherry-picks at this stage.

dt:
@bdarnell hm... *should* we include this in 1.1.1? to start getting numbers on unimplemented feature attempts (other than from parse) before 1.2?

bdarnell:
Will this be a significant driver of prioritization for 1.3? Even with this, our information will be incomplete, since not all unimplemented features are tagged as such. Expanding our collection of this class of information doesn't seem like a priority to me.

dt:
👍

PR #18553, opened 2017-09-18
util/timeutil: remove fake time support 

petermattis:

PR #18764, opened 2017-09-25
cherrypick-1.1: sql/pgwire,parser: render non-tz timestamp in its loc, set loc when c…

### LINKED ISSUE #16228: sql: casting timestamptz to timestamp not respecting timezone

dt:
…asting

Updated logic test behaviors verified from postgres: when a timestamptz is cast to timestamp, the result renders a localized time, with no offset.

Fixes #16228.


knz (approved):

justinj (approved):

dt:
This also needs #18783 or another fix for #18774, so I'll wait to see how we resolve that on master before moving forward here.

dt:
so #18783 feels a bit big, just in terms of files changed, for a late cherry-pick, though the actual change (`time.Unix()` vs `time.Unix().UTC()`) isn't that scary. We could potentially selectively `.UTC()`  just a couple callsites here instead. That said, this change, as-is, without ensuring we've `.UTC()`ed all the applicable paths, is potentially worse than just leaving the timestamp casts knonwn-broken?

knz:
I think your patch in #18783 is acceptable for a cherry-pick. Although perhaps we need a second set of eyes to check.

dt:
@knz okay, this now has #18676, #18553,  #18590 and #18783 so it's ready for another look. 
Also cc @bdarnell for thoughts on the scope here, as far as a cherrypick goes.

bdarnell:
LGTM

knz:
:lgtm: thank you!

---

Reviewed 3 of 3 files at r1, 5 of 5 files at r2, 3 of 3 files at r3, 39 of 39 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18764#-:-Kv4KBGqP2Xiid2jYT8i:bro3093)*

Issue #17369, opened 2017-08-01
perf: provide per-node performance self-test report


petermattis:
Provide a cluster performance self-test report that shows inter-node latencies, CPU capacity, disk throughput and disk latencies (reads, writes and syncs). This would have been useful during a recent forum debugging effort to highlight the usage of spinning disks and the resulting effect on disk syncs. 

Jira issue: CRDB-6040

petermattis:
Some Azure VMs throttle disk writes in pretty horrible fashion. The performance self-test should detect this throttling.

awoods187:
This will likely by ops/tools

tschottdorf:
On the face of it, we should be able to use `cli/synctest` for a part of this.

PR #18765, opened 2017-09-25
engine: io limit RocksDBMap batch commits

### LINKED ISSUE #15332: distsql: write throttling on azure leads to liveness failures and canceled context
Running
```
SELECT * FROM tpch.lineitem ORDER BY l_extendedprice
```
with 1227734 on `navy` fails with
```
Error: pq: [n1] communication error: rpc error: code = Canceled desc = context canceled
```
As far as I can see, the logs don't really shed light on the reason for the cancellation and I'm not sure where to start looking.

Subsequent runs sometimes end with the same failure or with an OOM. I think this is caused by aggregating the result set before outputting it and seems to be the same sort of memory issue we're seeing with other queries. Still, it's unfortunate that this crash happens sorting 2.5GiB of data.

screen shot 2017-04-25 at 2 23 24 pm

madelynnblue:
Similar to raft sideloading, limit the IO of tempstore writes. These
happen during DistSQL work and CSV importing, both of which have caused
node liveness errors. Until those are solved, we need a way to limit
these writes for DistSQL and CSV import so that use of those features
on Azure (and possibly others) doesn't cause cluster instability.

To do that, we add a limiter to RocksDBMapBatchWriter that is
auto-populated from the underlying RocksDB instance. (Or, in the case
where it's something else, does not set a limiter at all.)

Fixes #15332

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -530,7 +530,7 @@ func writeRocksDB(
 	writer := store.NewBatchWriter()
 	for kvBatch := range kvCh {
 		for _, kv := range kvBatch {
-			if err := writer.Put(kv.Key, kv.Value.RawBytes); err != nil {
+			if err := writer.Put(ctx, kv.Key, kv.Value.RawBytes); err != nil {
 				return 0, err
 			}
 		}
@@ -1291,7 +1291,7 @@ func (sp *sstWriter) Run(ctx context.Context, wg *sync.WaitGroup) {
 					val = b
 				}
 			}
-			if err := batch.Put(key, val); err != nil {
+			if err := batch.Put(ctx, key, val); err != nil {
 				return err
 			}
 		}
--- a/pkg/sql/distsqlrun/disk_row_container.go
+++ b/pkg/sql/distsqlrun/disk_row_container.go
@@ -142,7 +142,7 @@ func (d *diskRowContainer) AddRow(ctx context.Context, row sqlbase.EncDatumRow) 
 	if err := d.diskAcc.Grow(ctx, int64(len(d.scratchKey)+len(d.scratchVal))); err != nil {
 		return err
 	}
-	if err := d.bufferedRows.Put(d.scratchKey, d.scratchVal); err != nil {
+	if err := d.bufferedRows.Put(ctx, d.scratchKey, d.scratchVal); err != nil {
 		return err
 	}
 	d.scratchKey = d.scratchKey[:0]
@@ -203,7 +203,7 @@ type diskRowIterator struct {
 var _ rowIterator = diskRowIterator{}
 
 func (d *diskRowContainer) NewIterator(ctx context.Context) rowIterator {
-	if err := d.bufferedRows.Flush(); err != nil {
+	if err := d.bufferedRows.Flush(ctx); err != nil {
 		log.Fatal(ctx, err)
 	}
 	return diskRowIterator{rowContainer: d, SortedDiskMapIterator: d.diskMap.NewIterator()}
--- a/pkg/sql/distsqlrun/hash_row_container.go
+++ b/pkg/sql/distsqlrun/hash_row_container.go
@@ -490,7 +490,7 @@ func (i hashDiskRowBucketIterator) Mark(ctx context.Context, mark bool) error {
 	// These marks only matter when using a hashDiskRowIterator to iterate over
 	// unmarked rows. The writes are flushed when creating a NewIterator() in
 	// NewUnmarkedIterator().
-	return i.hashDiskRowContainer.bufferedRows.Put(i.Key(), rowVal)
+	return i.hashDiskRowContainer.bufferedRows.Put(ctx, i.Key(), rowVal)
 }
 
 // hashDiskRowIterator iterates over all unmarked rows in a
--- a/pkg/storage/engine/batch.go
+++ b/pkg/storage/engine/batch.go
@@ -19,9 +19,16 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"runtime/debug"
+	"time"
 
+	"golang.org/x/net/context"
+
+	"github.com/cockroachdb/cockroach/pkg/settings/cluster"
 	"github.com/cockroachdb/cockroach/pkg/util/bufalloc"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 	"github.com/pkg/errors"
 )
 
@@ -497,4 +504,33 @@ func (r *RocksDBBatchReader) varstring() ([]byte, error) {
 		return nil, fmt.Errorf("expected %d bytes, but found %d", n, c)
 	}
 	return s, nil
+}
+
+const bulkIOWriteLimiterLongWait = 500 * time.Millisecond
+
+// LimitBulkIOWrite limits writes of size cost to the cluster
+// BulkIOWRiteLimiter settings.
+func LimitBulkIOWrite(ctx context.Context, st *cluster.Settings, cost int) {
+	// The limiter disallows anything greater than its burst (set to
+	// BulkIOWriteLimiterBurst), so cap the batch size if it would overflow.
+	//
+	// TODO(dan): This obviously means the limiter is no longer accounting for
+	// the full cost. I've tried calling WaitN in a loop to fully cover the
+	// cost, but that doesn't seem to be as smooth in practice (TPCH-10 restores
+	// on azure local disks), I think because the file is written all at once at
+	// the end. This could be fixed by writing the file in chunks, which also
+	// would likely help the overall smoothness, too.
+	if cost > cluster.BulkIOWriteLimiterBurst {
+		cost = cluster.BulkIOWriteLimiterBurst
+	}
+
+	begin := timeutil.Now()
+	if err := st.BulkIOWriteLimiter.WaitN(ctx, cost); err != nil {
+		log.Errorf(ctx, "error rate limiting bulk io write: %+v", err)
+	}
+
+	if d := timeutil.Since(begin); d > bulkIOWriteLimiterLongWait {
+		log.Warningf(ctx, "bulk io write limiter took %s (>%s):\n%s",
+			d, bulkIOWriteLimiterLongWait, debug.Stack())
+	}
 }
--- a/pkg/storage/engine/disk_map.go
+++ b/pkg/storage/engine/disk_map.go
@@ -75,10 +75,10 @@ type SortedDiskMapBatchWriter interface {
 	// Put writes the given key/value pair to the batch. The write to the
 	// underlying store happens on Flush(), Close(), or when the batch writer
 	// reaches its capacity.
-	Put(k []byte, v []byte) error
+	Put(ctx context.Context, k []byte, v []byte) error
 	// Flush flushes all writes to the underlying store. The batch can be reused
 	// after a call to Flush().
-	Flush() error
+	Flush(context.Context) error
 
 	// Close flushes all writes to the underlying store and frees up resources
 	// held by the batch writer.
@@ -120,6 +120,7 @@ type RocksDBMapBatchWriter struct {
 	makeKey func(k []byte) MVCCKey
 	batch   Batch
 	store   Engine
+	limiter func(context.Context, int)
 }
 
 // RocksDBMapIterator iterates over the keys of a RocksDBMap in sorted order.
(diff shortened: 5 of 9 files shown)


andreimatei:
You want a limit per writer or per category of writes to temp storage or globally for all writes to temporary storage?
If I read this correctly, you've implemented a global limit, although it's linked to every writer. So I'd ask why to writers need to be aware of it?
More importantly, you seem to have given the shared storage a limit that says something about "bulk IO", although the storage is also used for things that are not "bulk IO". What gives? For queries using this storage, if we need to limit them, there should be some per-query minimum quota (similar to memory reservations).

I remember that I had some comments about sharing the temp storage between queries and whatever backup/restore does with it, and you've said that they'll be addressed in the 1.2 timeframe. The timeframe is upon us, so perhaps this is the time :)

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/storage/engine/disk_map.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-Kuv4j1cdw-DVqxYJBWC:-Kuv4j1cdw-DVqxYJBWD:b-lt1q19) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ead1ce3060a99cf698cc32f1b1049a8746c85e8/pkg/storage/engine/disk_map.go#L123)):*
> ```Go
> 	batch   Batch
> 	store   Engine
> 	limiter func(context.Context, int)
> ```

please add a comment and a name for the arg

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-Kuv4y_1xaG2XwBv24_T:b3tam3v)*

mjibson:
Yes, this is quite specifically a global limit. The problem is when we write to Azure disks too fast, we get node liveness errors. It is indeed linked to every writer, because that appeared to be a good way to implement it. (I thought it would be a bad choice to do the type casting in each call to .Flush because the temp store default capacity is 4096, which isn't that large, so the cast overhead would be not tiny.)

It does indeed use the named BulkIO variable, and that could change, but the limit needs to be a global limit that is shared among anything that does lots of writes to disk, like restore, csv import, and large dist sql disk overflow queries. It was just named bulkio because that's the first thing that needed it.

This change has nothing to do with the temp storage sharing between csv and distsql. Even when/if we do split up those temp stores, BOTH of those new stores would need to limit themselves to this same global limit.

The problem is Azure disk use is too high, and we need a store-wide limiter for it, and this is what we implemented already to fix this exact problem that restore encountered some time ago.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-Kuv7f3fFiejjUIv5EYY:b-rosfmm)*

andreimatei:
I understand that a global limit is needed since it's protecting a shared resource, but I'm saying that the implementation needs to give quotas to the different actors and not let one starve the others.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-Kuv9Sv-cr8zDo-SoJEp:b6ezo5d)*

andreimatei:
Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/storage/engine/batch.go, line 524 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-KuvDZt2LkdeEmDj0Qp9:-KuvDZt2LkdeEmDj0QpA:bnz7tny) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ead1ce3060a99cf698cc32f1b1049a8746c85e8/pkg/storage/engine/batch.go#L524)):*
> ```Go
> 	// would likely help the overall smoothness, too.
> 	if cost > cluster.BulkIOWriteLimiterBurst {
> 		cost = cluster.BulkIOWriteLimiterBurst
> ```

what's the point of setting a (small) burst limit if we're doing this? If there's no point, I think we should remove the burst limit (or, rather, set it to the refill rate). This code here would stay.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765)*

mjibson:
Ah yes, well, that's significantly more complicated and I'd like this to get into 1.1, but I agree that'd be real nice.

---

Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-KuvY9Q-D1eFde-lr07r:bs33gqf)*

petermattis:
Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/storage/engine/batch.go, line 512 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-KuvYeUT2_CY7rvFIbZC:-KuvYeUT2_CY7rvFIbZD:b-asu0sw) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ead1ce3060a99cf698cc32f1b1049a8746c85e8/pkg/storage/engine/batch.go#L512)):*
> ```Go
> 
> // LimitBulkIOWrite limits writes of size cost to the cluster
> // BulkIOWRiteLimiter settings.
> ```

`s/WRite/Write/g`

---

*[pkg/storage/engine/disk_map.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-KuvZ7pOTc5vLw_iHtcs:-KuvZ7pOTc5vLw_iHtct:b-j2ptb8) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ead1ce3060a99cf698cc32f1b1049a8746c85e8/pkg/storage/engine/disk_map.go#L244)):*
> ```Go
> 	}
> 	var limiter func(context.Context, int)
> 	if rocksdb, ok := r.store.(*RocksDB); ok {
> ```

Can this type assertion every return `false`? Seems better to remove the `ok` instead of stumbling ahead and not performing rate limiting (i.e. `rocksdb := r.store.(*RocksDB)`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765)*

andreimatei:
Here's how I see things:
This patch introduces a certain kind of throttling for unrelated classes of traffic, with no control over the policy (in fact, there's no policy versus mechanisms planes). The patch arbitrarily choses to put some I/O workloads under the control of this limit, but not others (Raft and logging I/O don't go through the limit). It also doesn't look to me like thought has been put into the mechanism to make integrating other workloads easy.
The patch claims to do what it's doing in the name of improving the situation with some particular VM storages devices. I don't think we need to particularly care about those devices; the patch should discuss whether the problem is more general and thus worth addressing. Also the commit does not discuss the particular limits of the particular device it claims to address - what's the time unit on which Azure measures IO, what's the allowed burst rate, what's the time scale of token recharge, etc. If we're discussing solutions for this particular Azure device, more information is needed to ascertain whether the particular type of throttling the patch is doing meshes well with the Azure policy (e.g. does the device force us to smooth out our IO rate over the course of some time window, or does it allow us to do IO as fast as the device supports it until some quota expires - and thus we'd be allowed to front-load requests). Do reads matter (the patch doesn't seem to say anything about them). The patch implements one type of throttling without discussing whether it meshes with the problem it claims to address.

In my opinion, short of implementing a real I/O scheduler (or at least laying some foundation for one) that allows different policies for different classes of traffic and understands the notion of different writers, we should not introduce this shared mechanism and subject unsuspecting workloads to it. FWIW, if the patch were to do something limited to backup/restore, the patch could be phrased as "backup/restore supports a type of throttling" and I'd have less of a problem with it if the author considered it useful for whatever reason. I already think that the sharing of temp storage between classes of use was done without sufficient policy support; this patch is advancing in a direction that's, IMHO, wrong.

Moreover, the patch introduces an implementation of throttling that does not strike me as high-quality enough for use as a shared and general mechanism. For one, the definition of "burst" that the Limiter uses seems to me to not be useful - it just refers to "one request", not to requests coming in quick succession.

If you care about I/O scheduling and want to advance this agenda, I'd suggest doing that through an RFC; I think it'd be a significant change.


Side note: we're limiting based on a cluster setting that should be a per-node flag.

---

Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-Kuywh8FMq7ojuZ8SGWa:b-kr3nxn)*

benesch:
Agreed in principle on all counts, @andreimatei, but the fact remains that without throttling restore (and now presumably load CSV) can wedge a cluster running on any Azure VM class, save for the Ls-series. As much as I wish that wasn't our problem, it's hard to claim we're a cloud-native database if we can't run on most VMs on the second-largest cloud provider. FWIW, I have a more detailed explanation of why Azure's I/O throttling is so nasty here: https://github.com/cockroachlabs/production/wiki/Azure-Disks.

petermattis:
Should we be investing in running smoothly on disks that have such a strange performance characteristic. For a normal disk, as you increase throughput latencies steadily increase. With these Azure disks we hit a cliff. I'm concerned that trying to work on such disks will be a game of whack-a-mole. An alternative would be to try and detect such throttling and shout a warning or refuse to run. See also https://github.com/cockroachdb/cockroach/issues/17369#issuecomment-330609631.

andreimatei:
@benesch as far as I see, this patch does not generally get us closer to "running on any Azure VM class". The patch is throttling certain classes of traffic, but not the most important one - regular Raft writes. Of course, that's the most difficult to do, because within it there are very different sub-classes - some (meta)requests are critical and should probably never be throttled, others aren't.
So the patch is both too narrow and too broad: it doesn't address the biggest problem so it's too narrow. It does address backup/restore's problem but in a too broad of a manner.

---

Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-KuzHezvevebIuXvafP0:b-sbmda1)*

andreimatei:
^ also all types of reads, which are conspicuously absent from the discussion

---

Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18765#-:-KuzIzPRA0wwQJvp__Ls:bmzs1zr)*

danhhz:
> Should we be investing in running smoothly on disks that have such a strange performance characteristic. For a normal disk, as you increase throughput latencies steadily increase. With these Azure disks we hit a cliff. I'm concerned that trying to work on such disks will be a game of whack-a-mole. An alternative would be to try and detect such throttling and shout a warning or refuse to run. See also #17369 (comment).

I'm starting to lean in favor of this. We spent a lot of time in the 1.1 cycle making RESTORE work on disks with bizarre throttling characteristics and I can't say it felt like we were spending time on the most important things we could be working on

petermattis:
It doesn't seem difficult to take the code in Nikhil's Azure-Disks wiki page and convert it into some benchmark that is run when a store is bootstrapped. Until we have customer demand for running on this type of storage, I'd rather not take on the burden of supporting it.

mjibson:
I'm only interested in merging this if we think we need it to be in 1.1. I think we need to decide (probably today or this week so we can figure out if this should go into the release) exactly how we will handle cases like this for users during the next 6 months (after which maybe we have a real solution to this). I think we can either:

1. Do stuff like this patch, and whack-a-mole the problems away until they are good enough. This is not a horrible choice since some of our flagship features (fast bulk import, distributed sql) can both hit this bug and take down a cluster, and at least we can give users a knob.
1. Clearly document what will happen if you run on non-L series Azure machines, and possibly other cloud provider machines, too. I think @benesch's point about us advertising ourselves as a cloud-native database, but one that can't actually run in general on Azure, is really good.

mjibson:
@petermattis SGTM. Are we then OK removing the existing BulkIO limiter, because it only exists for this reason?

petermattis:
> Are we then OK removing the existing BulkIO limiter, because it only exists for this reason?

Yes, assuming that it isn't providing some side benefit. I suppose I'd remove it and then test on better disks.

danhhz:
Last I checked most of our prod clusters had the bad disks. So let's fix that before we remove it, please, otherwise I'll be fielding questions about why RESTORE doesn't work on someone's cluster

petermattis:
Agreed. I would go further and state we shouldn't remove the existing throttling until we can adequately detect the bad disks at store bootstrap.

PR #18767, opened 2017-09-25
distsql,client: have DistSQL reads fill in the gateway field in BatchRequests

### LINKED ISSUE #18700: distsql/kv: queries processed by distsql can propagate incorrect GatewayNodeID
The [`DistSender` attaches a `GatewayNodeID` field to all batches that it sends](https://github.com/cockroachdb/cockroach/blob/9f0bb3691c10beb74911b22baa7839c21bf85dac/pkg/kv/dist_sender.go#L471) as a way of representing which node served as the gateway for the request. This enables us to [rebalance leaseholders towards where requests are coming from to reduce latency](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20170125_leaseholder_locality.md).

However, this method of attaching a `GatewayNodeID` with the local node's ID doesn't work so well when a `DistSQL` processor on a node that isn't the true gateway is the one sending the batch request through the `DistSender`.

@jseldess ran into this when running tpcc on a cluster distributed across the US, and it made leases not get moved as well as they should have because as many as half of the requests to some ranges were getting counted as coming from the node that their leaseholder was on rather than the node that was actually receiving all the SQL connections.

Tracing `Send(` calls up the stack is pretty frustrating due to how widely used and overloaded it is, so I'll start by asking. @andreimatei, do distsql nodes know the node ID of the SQL coordinator and how far separated is that knowledge from where distsql is constructing its `BatchRequest`s?

andreimatei:
Fixes #18700

Before this patch, the DistSender assumed that the gateway node was the
current node for all KV requests that it sent out. This isn't true for
requests produced by DistSQL flows.
This patch asks people creating client.Txns to be explicit about what
the gateway is, and DistSQL correctly identifies the true gateway.
client.Txn will then use that information to fill out all the requests
it sends.

I will cherry-pick this into 1.1

### DIFF
--- a/pkg/ccl/sqlccl/backup.go
+++ b/pkg/ccl/sqlccl/backup.go
@@ -340,7 +340,10 @@ func makeBackupDescriptor(
 	db := p.ExecCfg().DB
 
 	{
-		txn := client.NewTxn(db)
+		// TODO(andrei): Plumb a gatewayNodeID in here and also find a way to
+		// express that whatever this txn does should not count towards lease
+		// placement stats.
+		txn := client.NewTxn(db, 0 /* gatewayNodeID */)
 		opt := client.TxnExecOptions{AutoRetry: true, AutoCommit: true}
 		err := txn.Exec(ctx, opt, func(ctx context.Context, txn *client.Txn, opt *client.TxnExecOptions) error {
 			var err error
@@ -746,7 +749,10 @@ func backupResumeHook(typ jobs.Type) func(context.Context, *jobs.Job) error {
 		var tables []*sqlbase.TableDescriptor
 
 		{
-			txn := client.NewTxn(job.DB())
+			// TODO(andrei): Plumb a gatewayNodeID in here and also find a way to
+			// express that whatever this txn does should not count towards lease
+			// placement stats.
+			txn := client.NewTxn(job.DB(), 0 /* gatewayNodeID */)
 			opt := client.TxnExecOptions{AutoRetry: true, AutoCommit: true}
 			if err := txn.Exec(ctx, opt, func(ctx context.Context, txn *client.Txn, opt *client.TxnExecOptions) error {
 				txn.SetFixedTimestamp(details.EndTime)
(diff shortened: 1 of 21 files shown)


mjibson:
Review status: 0 of 21 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/ccl/sqlccl/backup.go, line 346 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18767#-KuvozcxzPLR1tJ29nxT:-Kuvozcyr5ZIz2ds16cg:bqqlhqv) ([raw file](https://github.com/cockroachdb/cockroach/blob/70c164076fa55c752485b3803156ccc3341c3c2b/pkg/ccl/sqlccl/backup.go#L346)):*
> ```Go
> 		// express that whatever this txn does should not count towards lease
> 		// placement stats.
> 		txn := client.NewTxn(db, 0 /* getwayNodeID */)
> ```

getway -> gateway

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18767)*

a-robinson:
:lgtm:

---

Reviewed 21 of 21 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18767#-:-KuyPJLKvwsCom7s8_0T:bnfp4nl)*

andreimatei:
Review status: 19 of 21 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/ccl/sqlccl/backup.go, line 346 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18767#-KuvozcxzPLR1tJ29nxT:-KuzN9o-nbfpmGKNzjsL:b-uitehm) ([raw file](https://github.com/cockroachdb/cockroach/blob/70c164076fa55c752485b3803156ccc3341c3c2b/pkg/ccl/sqlccl/backup.go#L346)):*

getway -> gateway
</blockquote></details>

Done.

I do think that we should introduce the notion of a quick-getaway node, though :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18767)*

PR #18769, opened 2017-09-26
cherrypick-1.1: server,ui: fix the display of the number of indexes per table

knz:
Cherry pick of #18603.

cc @cockroachdb/release


vilterp:
Why are there dups in the server response?

knz:
The server response contains one line per column in the index. This is a direct RPC transcript of the SQL command "SHOW INDEXES". Look at the output of that statement for details.

couchand (approved):

PR #18770, opened 2017-09-26
sql: improve implementation of WaitForOneVersion()

vivekmenezes:
This implementation reads the table descriptor as well as scans
the lease table in the same transaction thereby guaranteeing that
the entire operation happens at a particular timestamp.

### DIFF
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -239,34 +239,55 @@ func (s LeaseStore) release(ctx context.Context, stopper *stop.Stopper, table *t
 func (s LeaseStore) WaitForOneVersion(
 	ctx context.Context, tableID sqlbase.ID, retryOpts retry.Options,
 ) (sqlbase.DescriptorVersion, error) {
-	desc := &sqlbase.Descriptor{}
 	descKey := sqlbase.MakeDescMetadataKey(tableID)
-	var tableDesc *sqlbase.TableDescriptor
+	errNotOneVersion := errors.New("not one version")
 	for r := retry.Start(retryOpts); r.Next(); {
-		// Get the current version of the table descriptor non-transactionally.
-		//
-		// TODO(pmattis): Do an inconsistent read here?
-		if err := s.db.GetProto(context.TODO(), descKey, desc); err != nil {
+		var version sqlbase.DescriptorVersion
+		err := s.db.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
+			desc := &sqlbase.Descriptor{}
+			if err := txn.GetProto(ctx, descKey, desc); err != nil {
+				return err
+			}
+			tableDesc := desc.GetTable()
+			if tableDesc == nil {
+				return errors.Errorf("ID %d is not a table", tableID)
+			}
+			version = tableDesc.Version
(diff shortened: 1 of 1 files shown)


andreimatei:
Why is this an improvement? The question about which leases are active and which aren't shouldn't have anything to do with the timestamp of some transaction (I think). I'd like to understand more what the deal with this patch is. 

---

---

*[pkg/sql/lease.go, line 235 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18770#-KuzsEat7lZXAQoFsfIN:-KuzsEat7lZXAQoFsfIO:b-5yoevc) ([raw file](https://github.com/cockroachdb/cockroach/blob/9bfa6564654a349515694fd54c976d21535865a6/pkg/sql/lease.go#L235)):*
> After returning there can only be versions of the descriptor >= to the
> // returned version

This statement is too vague given all the recent advancements. Please consider clarifying it.

---

*[pkg/sql/lease.go, line 255 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18770#-Kuzt9uPnCIM2KotXPY1:-Kuzt9uPnCIM2KotXPY2:b-1kui8e) ([raw file](https://github.com/cockroachdb/cockroach/blob/9bfa6564654a349515694fd54c976d21535865a6/pkg/sql/lease.go#L255)):*
> ```Go
> 				return errors.Errorf("ID %d is not a table", tableID)
> 			}
> 			version = tableDesc.Version
> ```

nit: rename to `latestVersion`

---

*[pkg/sql/lease.go, line 266 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18770#-KuzuwGjVVlbv3sGEzfE:-KuzuwGjVVlbv3sGEzfF:b-erzyvw) ([raw file](https://github.com/cockroachdb/cockroach/blob/9bfa6564654a349515694fd54c976d21535865a6/pkg/sql/lease.go#L266)):*
> ```Go
> 			// lease to take more time to expire.
> 			expiration := parser.DTimestamp{
> 				Time: time.Unix(0, txn.OrigTimestamp().WallTime).Add(-time.Second),
> ```

I think the use of `s.clock.Now()` was better... I'm probably missing something, as I was saying in the PR comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18770#-:-KuzuP896zpx3wi08MX2:bs9le6s)*

vivekmenezes:
I don't believe there is a problem with the code as it stands. The only problem is whenever I look at it I need to reason about whether it works and even after looking at it leaves me uncomfortable. In particular, everywhere in the lease code we are using the notion of database time, in this code as it stands, we are reading a table descriptor at god knows what timestamp, we're comparing the expiration time read through a transaction against a call to s.clock.Now() rather than   using the timestamp of the transaction used to execute the SQL. What this change does is bring this code in line with all the other lease code (use transaction timestamps everywhere). Thoughts?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18770#-:-Kv-ztOWFbU3OsU0A_V2:b-z5pu33)*

PR #18778, opened 2017-09-26
storage: Fix tracking of keys written in replica.writeStats

### LINKED ISSUE #18607: storage: replica write QPS computation seems incorrect
I was working on #15997 when I wondered why I was getting the following suspiciously low stats for a replica that was serially plowing through batches of 100 ResolveIntent each:

```
“stats”: {
        “queriesPerSecond”: 2.140486885553871,
        “writesPerSecond”: 0.15809450129414426
      },
```

Spelunking a bit, it seems that we use the following to compute the write QPS:

```go
r.writeStats.recordCount(math.Max(float64(rResult.Delta.KeyCount), 1), 0)
```

But `rResult.Delta` is the MVCC stats diff. That is, if you write a version of an existing key, nothing changes. Similarly, when you resolve an intent, nothing changes.

This computation just doesn't seem right except when the operation exclusively creates new keys. I think for the right value we need to look at the `BatchRequest`. It doesn't seem trivial to compute the intuitively "correct" quantity, and perhaps we should clarify further what that means.

a-robinson:
Using the MVCCStats delta was incorrect, since it only counted inserts,
not updates or deletes.

Manually tested that updates to the same key now get reflected properly
in the writes per second row of the range debug page.

Fixes #18607

### DIFF
--- a/pkg/storage/engine/batch.go
+++ b/pkg/storage/engine/batch.go
@@ -498,3 +498,12 @@ func (r *RocksDBBatchReader) varstring() ([]byte, error) {
 	}
 	return s, nil
 }
+
+// RocksDBBatchCount provides an efficient way to get the count of mutations
+// in a RocksDB Batch representation.
+func RocksDBBatchCount(repr []byte) (int, error) {
+	if len(repr) < headerSize {
+		return 0, errors.Errorf("batch repr too small: %d < %d", len(repr), headerSize)
+	}
+	return int(binary.LittleEndian.Uint32(repr[countPos:headerSize])), nil
+}
--- a/pkg/storage/engine/batch_test.go
+++ b/pkg/storage/engine/batch_test.go
@@ -244,8 +244,14 @@ func TestBatchRepr(t *testing.T) {
 		if err != nil {
 			t.Fatalf("%+v", err)
 		}
-		if count, expected := r.Count(), 3; count != expected {
-			t.Fatalf("bad count: expected %d, but found %d", expected, count)
+		const expectedCount = 3
+		if count := r.Count(); count != expectedCount {
+			t.Fatalf("bad count: RocksDBBatchReader.Count expected %d, but found %d", expectedCount, count)
+		}
+		if count, err := RocksDBBatchCount(repr); err != nil {
+			t.Fatal(err)
+		} else if count != expectedCount {
+			t.Fatalf("bad count: RocksDBBatchCount expected %d, but found %d", expectedCount, count)
 		}
 
 		var ops []string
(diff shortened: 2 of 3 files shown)


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18778#-:-KuyW-ORgIkp8NQZrHBB:bnfp4nl)*

tbg (approved):
LGTM, though @petermattis should comment on the cost of this new code (I assume it's fine).

petermattis:
@tschottdorf The cost is low. We could make it lower by avoiding allocations and extracting a call to just retrieve the count from `WriteBatch.Data`. We'd have to remove the usage of `buf.Reader`, but that is easy.

a-robinson:
Yeah, putting more allocations in the hot path here isn't ideal. I'm missing how to avoid allocating the `buf.Reader`, though. Unless you mean (1) copying the `buf.Reader` implementation and providing a way to instantiate one of our custom readers on the stack or (2) writing our own binary parsing code to pull out the count rather than using `encoding/binary`.

I was going to cherry pick this into 1.1 because right now the counts are quite incorrect, but I don't know about adding more allocations to this code path so late.

tschottdorf:
No strong feelings, but the `WriteBatch` version doesn't have the allocation in the first place.

a-robinson:
Yeah, but I think keys modified may be the better metric here, at least if the multi-node performance characteristics are similar to the single-node characteristics I just tested out. Running kv with `--batch=1` and `--block-size=8192` "stabilizes" (for some definition of "stabilizes") around 2500 inserts/second, for a throughput of 20.5MB. Running kv with `--batch=128` and `--block-size=64` stabilized around 22000 inserts/second, for a total throughput of 1.4MB. Since each of those requests was technically the same amount of data, it looks pretty clear that the number of keys touched has a greater affect on throughput than the amount of data. Thus it seems as though keys mutated is a better stand-in for load on the range than bytes written.

petermattis:
> Yeah, putting more allocations in the hot path here isn't ideal. I'm missing how to avoid allocating the buf.Reader, though. Unless you mean (1) copying the buf.Reader implementation and providing a way to instantiate one of our custom readers on the stack or (2) writing our own binary parsing code to pull out the count rather than using encoding/binary.

I'll find the codes shortly to show you what I meant.

petermattis:
You want to extract this code from `RocksDBBatchBuilder.ApplyRepr`

```
func (b *RocksDBBatchBuilder) ApplyRepr(repr []byte) error {
	if len(repr) < headerSize {
		return errors.Errorf("batch repr too small: %d < %d", len(repr), headerSize)
	}
	b.maybeInit()
	pos := len(b.repr)
	data := repr[headerSize:]
	b.grow(len(data))
	copy(b.repr[pos:], data)
	b.count += int(binary.LittleEndian.Uint32(repr[countPos:headerSize]))
	return nil
}
```

The key bit is: `binary.LittleEndian.Uint32(repr[countPos:headerSize])`.

a-robinson:
Awesome, thanks @petermattis! Updated.

petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18778#-:-Kuz__uZTw--8EcyzJ-e:bnfp4nl)*

PR #18783, opened 2017-09-26
*: s/time.Unix/timeutil.Unix/ to ensure UTC

### LINKED ISSUE #18774: cli: regression in test Example_sql_format
I simply ran `make test PKG=./pkg/cli` on a recent master on my Linux system - no other changes.
This seems to have regressed with #18676.

```
--- FAIL: Example_sql_format (0.15s)
got:
sql -e create database t; create table t.times (bare timestamp, withtz timestamptz)
CREATE TABLE
sql -e insert into t.times values ('2016-01-25 10:10:10', '2016-01-25 10:10:10-05:00')
INSERT 1
sql -e select * from t.times
bare    withtz
2016-01-25 05:10:10+00:00       2016-01-25 15:10:10+00:00
# 1 row
want:
sql -e create database t; create table t.times (bare timestamp, withtz timestamptz)
CREATE TABLE
sql -e insert into t.times values ('2016-01-25 10:10:10', '2016-01-25 10:10:10-05:00')
INSERT 1
sql -e select * from t.times
bare    withtz
2016-01-25 10:10:10+00:00       2016-01-25 15:10:10+00:00
# 1 row
```

dt:
timeutil.Unix wraps stdlib's time.Unix but ensures the result is in UTC, not Local.
time.Unix() uses the time.Local zone, but that is host-dependant.

Since we expect geographically diverse clusters of interchangable nodes, host-local
time zones are not desireable.

Fixes #18774.


knz:
Huh, I don't understand this. Please extend the commit message with a discussion of what's going on here.

knz:
LGTM.

cc @cockroachdb/build @cockroachdb/core

dt:
After thinking about this a bit more, I became worried there were probably other paths by which we'd let a `time.Local`'ed `time.Time` make its way into a datum, so I just replaced it with a much more compressive change to replace `time.Unix` with a wrapper to ensure it always gets `UTC`ed.

pkg/sql/parser/datum.go
```diff
@@ -38,10 +38,15 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
 	"github.com/cockroachdb/cockroach/pkg/util/duration"
 	"github.com/cockroachdb/cockroach/pkg/util/ipaddr"
+	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 	"github.com/cockroachdb/cockroach/pkg/util/uint128"
 	"github.com/cockroachdb/cockroach/pkg/util/uuid"
 )
 
+func init() {
+	time.Local = time.UTC
```
tamird: um

pkg/util/timeutil/time.go
```diff
@@ -37,7 +37,7 @@ var UnixEpoch = time.Unix(0, 0)
 // time, usec microseconds since UnixEpoch. In Go's current time.Time
```
tamird: this comment is no longer correct.

pkg/util/timeutil/now_windows.go
```diff
@@ -41,5 +42,5 @@ func init() {
 func Now() time.Time {
```
tamird: this comment still talks about local time.

pkg/util/timeutil/now_unix.go
```diff
@@ -22,5 +22,5 @@ import (
 
 // Now returns the current local time.
```
tamird: this comment still talks about local time.

knz:
Reviewed 37 of 38 files at r1, 10 of 10 files at r2.

---

*[pkg/sql/parser/datum.go, line 47 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783#-KuzMetuMCbnsvcTeFvC-r2-47:-KuzPWf1WJl-y61vlnI7:b-lmy66l) ([raw file](https://github.com/cockroachdb/cockroach/blob/9f3fa797ea2d9cd01774b0df2d8c1a1b6118ce0d/pkg/sql/parser/datum.go#L47)):*

um
</blockquote></details>

Yeah now you've introduced timeutil this is not warranted any more.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783)*

pkg/sql/parser/datum.go
dt: heh, oh, yeah, that was what this commit was supposed to avoid.

pkg/util/timeutil/now_unix.go
dt: Done.

pkg/util/timeutil/now_windows.go
dt: Done.

pkg/util/timeutil/time.go
dt: Done.

dt:
Okay, various import cycles and comments fixed up now. RFAL!

knz:
LGTM with questions

---

Reviewed 10 of 18 files at r3, 10 of 10 files at r4.

---

*[pkg/rpc/context_test.go, line 528 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783#-KuzyXabuI6Orbj9sVFC:-KuzyXacyj-9CI4iGvMP:b-dl8zch) ([raw file](https://github.com/cockroachdb/cockroach/blob/3095785e9a78dd1d37a22f6f37e56d1fd696f8aa/pkg/rpc/context_test.go#L528)):*
> ```Go
> 	ac.time = time.Add(ac.getAdvancementInterval())
> 	ac.Unlock()
> 	return time.UnixNano()
> ```

Why the revert?

---

*[pkg/util/timeutil/now_windows.go, line 44 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783#-Kuzyl5FIw4EquiM3ZRG:-Kuzyl5FIw4EquiM3ZRH:bg7tlnx) ([raw file](https://github.com/cockroachdb/cockroach/blob/3095785e9a78dd1d37a22f6f37e56d1fd696f8aa/pkg/util/timeutil/now_windows.go#L44)):*
> ```Go
> 	var ft windows.Filetime
> 	windows.GetSystemTimePreciseAsFileTime(&ft)
> 	return time.Unix(0, ft.Nanoseconds()).UTC()
> ```

Why not use timeutil here too?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783#-:-KuzynSN5dYLw3nvUnQA:b-bnc1mu)*

dt:

---

*[pkg/rpc/context_test.go, line 528 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783#-KuzyXabuI6Orbj9sVFC:-KuzzFM8TtqmAu0V9HFN:b7s4l6m) ([raw file](https://github.com/cockroachdb/cockroach/blob/3095785e9a78dd1d37a22f6f37e56d1fd696f8aa/pkg/rpc/context_test.go#L528)):*

Why the revert?
</blockquote></details>

`time` there is a local var named time, not the package `time`.

---

*[pkg/util/timeutil/now_windows.go, line 44 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783#-Kuzyl5FIw4EquiM3ZRG:-KuzzMdmdM8nJ6Kqiab5:b-c5t0iu) ([raw file](https://github.com/cockroachdb/cockroach/blob/3095785e9a78dd1d37a22f6f37e56d1fd696f8aa/pkg/util/timeutil/now_windows.go#L44)):*

Why not use timeutil here too?
</blockquote></details>

we're *in* `timeutil` here -- we could just use `Unix()` but eh: this file is one of the few exempted from the timeutil linter and thus assumed to be responsible doing the right thing on its own.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18783)*

knz:
Aw, thanks.

dt:
@knz hmm, this feels like a bit big for a cherry-pick at this point... thoughts?

knz:
It's spread out but the change is systematic. I do not see a problem. If in doubt you can wait for 1.1.1 but it should make it to 1.1 eventually (together with the other ts fix in #18764)

Issue #18784, opened 2017-09-26
storage: move evaluation of BatchRequests to a separate package


tschottdorf:
Broken out from https://github.com/cockroachdb/cockroach/issues/18779.

tschottdorf:
There's a bit of movement yet to do, but I don't think it makes sense to cherry-pick that into 2.0, so I'm moving the milestone.

tschottdorf:
Not sure it's worth tracking this. There's a bit left to do (around EndTransaction) but this issue won't fix that. Someone running into it and getting annoyed will.

nvanbenschoten:
@tschottdorf more to do beyond https://github.com/cockroachdb/cockroach/pull/26917?

tschottdorf:
I think there's some stuff around the commit triggers that isn't really home in replica_command.go, see this comment from the PR you linked:

```
The one-phase commit logic still lives in storage and, unfortunately,
ends up tightly coupled with the EndTransaction machinery in batcheval,
but this is a step in the right direction. By contrast, the extraction
of writeInitial[Replica]State into the stateloader package goes quite
smoothly.
```

PR #18785, opened 2017-09-26
storage: factor out replicaStateLoader and abortCache

tbg:
This is one of the self-contained threads to unravel for #18784.

We probably don't want to merge this just yet, but I think it's worth
basing other refactoring explorations on top of.

### DIFF
--- a/pkg/storage/abortcache/abort_cache.go
+++ b/pkg/storage/abortcache/abort_cache.go
@@ -12,7 +12,7 @@
 // implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
-package storage
+package abortcache
 
 import (
 	"golang.org/x/net/context"
@@ -28,7 +28,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/uuid"
 )
 
-// The AbortCache sets markers for aborted transactions to provide
+// An Instance sets markers for aborted transactions to provide
 // protection against an aborted but active transaction not reading
 // values it wrote (due to its intents having been removed).
 //
@@ -36,7 +36,7 @@ import (
 // is cleared from a range, and is consulted before read commands are processed
 // on a range.
 //
-// The AbortCache stores responses in the underlying engine, using keys derived
+// The Instance stores responses in the underlying engine, using keys derived
 // from Range ID and txn ID.
 // Note that the epoch number is not used to query the cache: once aborted, even
 // higher epochs are prohibited from reading data. That's because, for better or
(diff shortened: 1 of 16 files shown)


petermattis:
:lgtm: though I agree we don't want to merge this yet.

---

Review status: 0 of 16 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18785#-:-Kuz-8jTutxWq9UbLm5n:b-a5dxcm)*

bdarnell:
:lgtm:

---

Review status: 0 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/storage/abortcache/abort_cache.go, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18785#-KuzQwfJXHK5ZVqKucCw:-KuzQwfKOexCDP2OLvLa:b-mpfjh2) ([raw file](https://github.com/cockroachdb/cockroach/blob/48c4a46fd40f5cbea2021780a7eee58ab7166b4e/pkg/storage/abortcache/abort_cache.go#L53)):*
> ```Go
> // this the "AbortSpan" instead, but large parts of the code still call this
> // "Instance". We should settle for one and rename everything post-yellow.
> type Instance struct {
> ```

`type Instance` seems wrong. I think the stuttery `abortcache.AbortCache` is better.

Nit: I don't like that the filename has an underscore and the directory doesn't. 

---

*[pkg/storage/stateloader/instance.go, line 55 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18785#-KuzRSHCy4cyu5uYNzD6:-KuzRSHCy4cyu5uYNzD7:bqat0kv) ([raw file](https://github.com/cockroachdb/cockroach/blob/48c4a46fd40f5cbea2021780a7eee58ab7166b4e/pkg/storage/stateloader/instance.go#L55)):*
> ```Go
> // struct with a mutex, and temporary loaders may be created when
> // locking is less desirable than an allocation.
> type Instance struct {
> ```

Again, I'd rather call this `stateloader.StateLoader`.

---

*[pkg/storage/stateloader/instance.go, line 60 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18785#-KuzRMWt8sNlgm4XIlCF:-KuzRMWt8sNlgm4XIlCG:b-4j54ou) ([raw file](https://github.com/cockroachdb/cockroach/blob/48c4a46fd40f5cbea2021780a7eee58ab7166b4e/pkg/storage/stateloader/instance.go#L60)):*
> ```Go
> }
> 
> // Make creates in Instance.
> ```

/in/an/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18785#-:-KuzSRLJ4DUad6SLvlq_:bnfp4nl)*

tschottdorf:
TFTRs! I'll get back to this once we've released 1.1 and are ready to land refactors.

tschottdorf:
Obsolete; the commits have been reworked and landed (or close to) as #19441 and #19401.

PR #18793, opened 2017-09-26
storage: fix data race in RequestLeaseRequest.Lease

petermattis:
Found by failure of TestLeaseExtensionNotBlockedByRead. We can't mutate
RequestLeaseRequest.Lease.Expiration which is shared with the caller.

### DIFF
--- a/pkg/storage/replica_command.go
+++ b/pkg/storage/replica_command.go
@@ -2029,11 +2029,12 @@ func evalRequestLease(
 
 	// MIGRATION(tschottdorf): needed to apply Raft commands which got proposed
 	// before the StartStasis field was introduced.
-	if args.Lease.DeprecatedStartStasis == nil {
-		args.Lease.DeprecatedStartStasis = args.Lease.Expiration
+	newLease := args.Lease
+	if newLease.DeprecatedStartStasis == nil {
+		newLease.DeprecatedStartStasis = newLease.Expiration
 	}
-	isExtension := prevLease.Replica.StoreID == args.Lease.Replica.StoreID
-	effectiveStart := args.Lease.Start
+	isExtension := prevLease.Replica.StoreID == newLease.Replica.StoreID
+	effectiveStart := newLease.Start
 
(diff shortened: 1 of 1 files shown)


pkg/storage/replica_command.go
```diff
@@ -2062,6 +2062,10 @@ func evalRequestLease(
 			return newFailedLeaseTrigger(false /* isTransfer */), rErr
 		}
 		if args.Lease.Type() == roachpb.LeaseExpiration {
+			// NB: Avoid mutating pointers in the argument which might be shared with
+			// the caller.
+			t := *args.Lease.Expiration
```
tschottdorf: This still smells, we shouldn't be changing `args` at all. Just make `args := *(previous stuff)` so that we really have a local copy here (in addition to your change here). But we also need to switch out `Lease` if it's a pointer field.

petermattis:

---

*[pkg/storage/replica_command.go, line 2067 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18793#-KuzF1vC-hsfRqLdcW0H-r1-2067:-KuzZTVjX6r3gU4e8eyx:b-f95zp2) ([raw file](https://github.com/cockroachdb/cockroach/blob/2436e58ad177a6fa8b9a39ca6a7dfe55653d1538/pkg/storage/replica_command.go#L2067)):*

This still smells, we shouldn't be changing `args` at all. Just make `args := *(previous stuff)` so that we really have a local copy here (in addition to your change here). But we also need to switch out `Lease` if it's a pointer field.
</blockquote></details>

Ok, I've changed this to stop mutating `args.Lease`. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18793)*

bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18793#-:-Kv3UVwzP3YcIyk7LI9T:bnfp4nl)*

Issue #18779, opened 2017-09-26
core: refactoring projects


cuongdo:
This is a master list for core refactoring work. The theme is breaking up the massive `storage` package. Separate GitHub issues should be created as work or discussion begins on an item.

Ideas:

- [ ] Break apart the larger types, like Replica.
  - [ ] Remove unrelated stuff from Replica.
- [ ] Simplify Pre-Raft / Raft flow.
- [ ] Add explicit instructions, such as bumping the transaction timestamp, that KV layer can send to `TxnCoordSender`. This would add grepability and understandablity.
- [ ] Pull queues and caches into their own package.
- [ ] Pull node liveness out of storage.
- [ ] Centralize split code.
- [ ] Add file-level comments that describe the general purpose and scope of that file.
- [ ] Move `RaftTransport` out of `storage`
- [ ] Move `raftScheduler` out of `storage`
- [x] #18784: Move evaluation of `BatchRequest`s to a separate package
- [ ] Enforce lock ordering, preferably programmatically

As you start to work on items, please put your name or GitHub id next to the task to claim it.

a6802739:
@cuongdo, could I have a try for this part?

petermattis:
@a6802739 The above checklist is a grab bag of refactoring ideas, not all of which may pan out. Give us some time to make a bit of progress on the margins and get a better feel for what the refactoring will look like. I'm sure there will be some tasks that emerge for you to tackle.

a6802739:
@petermattis, Thank you very much.

a-robinson:
Also, as suggested in https://github.com/cockroachdb/cockroach/issues/18657#issuecomment-334758036: remove the locking of `Replica.raftMu` in `Replica.propose`

tschottdorf:
https://github.com/cockroachdb/cockroach/pull/19133 has some more abstract refactoring ideas on how to make the "non-raft" portion of Replica clearer.

tschottdorf:
Closing as there isn't anything very specific left here.

PR #18796, opened 2017-09-26
[DNM] storage: factor out evaluation of BatchRequest

tbg:
The first two commits are #18785.

----

See #18779.

This doesn't carry out all the work but does all the basic required
refactoring (mod cleanups).

I don't love that we're passing `ReplicaEvalContext` through an interface.
That should be avoidable and doing so should be instructive (I suspect
we'll trim some fat in the process):

Many methods roughly take the following form:

```go
// GetMVCCStats returns the Replica's MVCCStats. func (rec
ReplicaEvalContext) GetMVCCStats() enginepb.MVCCStats {
   r := rec.repl
   r.mu.RLock()
defer r.mu.RUnlock()
return *r.mu.state.Stats, nil
}
```

and to define this method inside of the `batcheval` package, we'd have
instead

```go
type ReplicaEvalContext {
   // ...
   state struct {
       *sync.Mutex
       *storagebase.ReplicaState
   }
}
```

### DIFF
--- a/pkg/ccl/storageccl/add_sstable.go
+++ b/pkg/ccl/storageccl/add_sstable.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/keys"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/storage"
+	"github.com/cockroachdb/cockroach/pkg/storage/batcheval"
 	"github.com/cockroachdb/cockroach/pkg/storage/engine"
 	"github.com/cockroachdb/cockroach/pkg/storage/engine/enginepb"
 	"github.com/cockroachdb/cockroach/pkg/storage/storagebase"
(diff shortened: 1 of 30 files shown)


petermattis:
Very nice! I'm not in love with the name `batcheval`, but I don't have a better suggestion right now and the bike shedding there can wait until later.

---

Review status: 0 of 30 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18796#-:-Kv2Vrz4MOp4E1nHpDnv:b-wshphr)*

tschottdorf:
Let the bike shed begin, @petermattis. I'd prefer not having to carry out a rename once I brush this up.

I like `batcheval` and have no better name. @andreimatei suggested `eval` which I find is a bit ambiguous but could also work.

petermattis:
`eval` feels too general. `kveval`? Shorter than `batcheval` and not ambiguous like `eval`.

a-robinson:
I like `batcheval`, FWIW

petermattis:
I'm also ok with `batcheval`. I don't have a strong preference between that and `kveval`.

tschottdorf:
Superseded by #19888.

PR #18771, opened 2017-09-26
sql: remove the list of descriptor IDs from the output of SHOW JOBS.

knz:
Since descriptor IDs are not (yet) part of the public UI of
CockroachDB, they should not be printed out by documented statements
like SHOW.

Suggested in https://github.com/cockroachdb/docs/pull/1872#issuecomment-332006659.
Resolves the discussion on that issue.

PR #18797, opened 2017-09-26
cherrypick-1.1: sql: remove the list of descriptor IDs from the output of SHOW JOBS.

knz:
Cherry-picks #18771.

Since descriptor IDs are not (yet) part of the public UI of
CockroachDB, they should not be printed out by documented statements
like SHOW.

Needed to complete the 1.1 docs.

cc @cockroachdb/release


benesch:
:shipit:

PR #18799, opened 2017-09-26
[DNM] acceptanceccl: test restore with no io limit

madelynnblue:
This is a test PR to see how the Azure L machines perform without this limit.


mjibson:
Ran this test 4 times and it never failed.

benesch:
Wait, we should merge this then, no?

mjibson:
Yes I guess?

PR #18800, opened 2017-09-26
cherrypick-1.1: storage: Fix tracking of keys written in replica.writeStats

### LINKED ISSUE #18607: storage: replica write QPS computation seems incorrect

a-robinson:
Using the MVCCStats delta was incorrect, since it only counted inserts,
not updates or deletes.

Manually tested that updates to the same key now get reflected properly
in the writes per second row of the range debug page.

Fixes #18607 

@cockroachdb/release - seemingly worth cherrypicking because the count is completely broken and misleading in the "Keys Written per Second per Store" admin UI graph and our debug pages without this.

### DIFF
--- a/pkg/storage/engine/batch.go
+++ b/pkg/storage/engine/batch.go
@@ -498,3 +498,12 @@ func (r *RocksDBBatchReader) varstring() ([]byte, error) {
 	}
 	return s, nil
 }
+
+// RocksDBBatchCount provides an efficient way to get the count of mutations
+// in a RocksDB Batch representation.
+func RocksDBBatchCount(repr []byte) (int, error) {
+	if len(repr) < headerSize {
+		return 0, errors.Errorf("batch repr too small: %d < %d", len(repr), headerSize)
+	}
+	return int(binary.LittleEndian.Uint32(repr[countPos:headerSize])), nil
+}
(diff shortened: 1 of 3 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18800#-:-Kv3_xwrxSLKu7HzSZGF:bnfp4nl)*

PR #18805, opened 2017-09-26
cherrypick-1.1: sql,keys: bugfix to rowfetcher for COUNT(*)

jordanlewis:
d5c9932 introduced a critical bug to rowfetcher that prevented it from
correctly understanding tables with multiple column families when no
columns from the table were marked as needed, such as in the case of
`COUNT(*)`.

This commit fixes the bug by reusing the logic from EnsureSafeSplitKey
to advance the rowfetcher's key to the family id section as it must,
without needing to decode any of the prefix of the key.

Also add a test to ensure COUNT(*) and multiple column families interact
properly.


jordanlewis:
cc @cockroachdb/release

dt (approved):

PR #18807, opened 2017-09-26
sql: Include Range ID in the output of SHOW TESTING_RANGES

a-robinson:
The new output looks like:

On multiple occasions when debugging I've wanted to be able to look up
debug pages for the ranges that are backing a particular table, but
figuring out the relevant range IDs is an awkward, indirect process.
There may be reasons for not wanting to expose this, so let me know
if you're aware of any. (cc @cockroachdb/core)

```
root@:26257/> show testing_ranges from table foo.bar;
+-----------+---------+----------+----------+--------------+
| Start Key | End Key | Range ID | Replicas | Lease Holder |
+-----------+---------+----------+----------+--------------+
| NULL      | /1      |       18 | {1}      |            1 |
| /1        | /2      |       19 | {1}      |            1 |
| /2        | /3      |       20 | {1}      |            1 |
| /3        | /4      |       21 | {1}      |            1 |
| /4        | NULL    |       22 | {1}      |            1 |
+-----------+---------+----------+----------+--------------+
```

I'll fix up the broken logic tests if you're open to this idea. It may make the logic tests require a little more maintenance unless we filter the output of some of the test cases to not return the Range ID.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/distsql_agg
+++ b/pkg/sql/logictest/testdata/logic_test/distsql_agg
@@ -22,7 +22,7 @@ INSERT INTO data SELECT a, b, c, d FROM
 
 # Verify data placement.
 query TTTI colnames
-SHOW TESTING_RANGES FROM TABLE data
+SELECT "Start Key", "End Key", "Replicas", "Lease Holder" FROM [SHOW TESTING_RANGES FROM TABLE data]
 ----
 Start Key  End Key  Replicas  Lease Holder
 NULL       /1       {1}       1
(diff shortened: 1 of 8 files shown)


knz:
I like the idea.

bdarnell (approved):

a-robinson:
TFTR! I've fixed up the logic tests, and also rearranged the output columns to come after the start key and end key (since the output rows are sorted by the start key, not by the range ID).

RaduBerinde (commented):
LGTM. You can consider hiding the new column so it only shows up if you select it specifically (to avoid the added selects in logictests).

PR #18809, opened 2017-09-26
ui: remove hidden and broken query plan screen

couchand:
This hidden query plan screen is broken.  Removing it cuts our bundle
size by 9%.  We can come back and fix this later, but for now it
shouldn't be bloating our bundle.


mrtracy:
LGTM, and i agree that this is worth cherry picking for 1.1; a no-risk file size reduction.

mrtracy (approved):

Issue #15959, opened 2017-05-16
sql: Cockroach_Nightlies_Steady_6Medium timing out


vivekmenezes:
The acceptance test: Cockroach_Nightlies_Steady_6Medium

https://teamcity.cockroachdb.com/viewType.html?buildTypeId=Cockroach_Nightlies_Steady_6Medium

is supposed to take many hours (we haven't run it very often to have a reasonable expectation), but it is expected to finish within the configured deadline of 24 hours. The last two attempts I've made to run it have hit the deadline.

Investigating this further is waiting on the schema change progress reporting change #15936 that will allow us to see if the test needs more time or it's simply getting stuck.

vivekmenezes:
related to #12424

jordanlewis:
This test doesn't ever complete anymore, with or without #17078. I'm looking into it. Last time I ran it, with #17078, it failed at 90% percent due to #17180.

jordanlewis:
This time, nodes are spitting out "heartbeat failed" messages hundreds of times per second, e.g.:

```
E170727 01:35:53.517489 2220486 storage/replica_range_lease.go:195  [n1,s1,r2111/1:/System/tsd/cr.node.txn.…] heartbeat failed on epoch increment
E170727 01:35:53.519059 2220514 storage/replica_range_lease.go:195  [n1,s1,r1991/1:/Table/51/1/316{50653…-95042…}] heartbeat failed on epoch increment
E170727 01:35:53.521140 2220517 storage/replica_range_lease.go:195  [n1,s1,r220/1:/Table/51/1/760{44165…-89456…}] heartbeat failed on epoch increment
E170727 01:35:53.523139 2220491 storage/replica_range_lease.go:195  [n1,s1,r1228/1:/Table/51/1/3{599755…-604286…}] heartbeat failed on epoch increment
E170727 01:35:53.525162 2220494 storage/replica_range_lease.go:195  [n1,s1,r818/1:/Table/51/1/4{796024…-800562…}] heartbeat failed on epoch increment
E170727 01:35:53.527490 2220496 storage/replica_range_lease.go:195  [n1,s1,r1590/1:/Table/51/1/51{67959…-72546…}] heartbeat failed on epoch increment
E170727 01:35:53.529389 2220501 storage/replica_range_lease.go:195  [n1,s1,r1276/1:/Table/51/1/4{668472…-714681…}] heartbeat failed on epoch increment
[...]

Issue #18817, opened 2017-09-27
acceptance: allocator tests failing during cluster init


cuongdo:
During last night's runs of the allocator tests, the initial cluster failed to start. This is a new problem that started with last night's runs. The messages in [one of the cockroach.log files](https://teamcity.cockroachdb.com/repository/download/Cockroach_AllocatorTest1to310g/362270:id/acceptance/node.0/cockroach/cockroach.log) point to startup/init errors:

```
I170927 06:17:06.272982 200 gossip/server.go:232  [n1] received initial cluster-verification connection from {tcp uprep-1to3s-kickassnorthcutt-cockroach-1:26257}
I170927 06:17:06.276470 40 gossip/client.go:134  [n1] closing client to node 1 (uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257): stopping outgoing client to node 1 (uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257); loopback connection
I170927 06:17:06.296202 196 storage/node_liveness.go:355  [n1,hb] heartbeat failed on epoch increment; retrying
I170927 06:17:07.104318 430 gossip/client.go:129  [n1] started gossip client to uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257
I170927 06:17:07.105706 430 gossip/client.go:134  [n1] closing client to node 1 (uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257): stopping outgoing client to node 1 (uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257); loopback connection
I170927 06:17:08.106772 717 gossip/client.go:129  [n1] started gossip client to uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257
I170927 06:17:08.108127 717 gossip/client.go:134  [n1] closing client to node 1 (uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257): stopping outgoing client to node 1 (uprep-1to3s-kickassnorthcutt-cockroach-1.eastus.cloudapp.azure.com:26257); loopback connection
[...]

PR #18818, opened 2017-09-27
acceptance: refresh allocator test medium data sets

cuongdo:
The medium data sets have been regenerated using TPC-H scale factor 100
data. This results in a similar data size but an 8x increase in range
count, which is great for more thoroughly testing the allocator.

Helps unblock #15959

### DIFF
--- a/build/teamcity-nightly-acceptance.sh
+++ b/build/teamcity-nightly-acceptance.sh
@@ -47,11 +47,11 @@ case $TESTNAME in
     ;;
   TestSteady_6Medium)
     TESTTIMEOUT=24h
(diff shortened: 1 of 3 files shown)


benesch:
Code LGTM! I assume you ran the new tests locally/via TC?

a-robinson (approved):
Do we also need to change the standard deviations specified in the teamcity script, as in https://github.com/cockroachdb/cockroach/pull/18212?

cuongdo:
@benesch I'm waiting for #18817 to be fixed before running. I'll definitely run the affected tests prior to merging.

@a-robinson Good catch. I've updated the standard deviations.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18818#-:-Kv2osKhpEr_Jmi6FTuw:bkqoan4)*

a-robinson (approved):

PR #18828, opened 2017-09-27
storage,storage/engine: reduce encoded size of MVCCNetworkStats

petermattis:
Changed most of the fields of MVCCNetworkStats to variable width
encodings. This is not compatible with the MVCCStats encoding, so
changed ReplicatedEvalResult to contain both old_delta (MVCCStats) and
delta (MVCCRangeStats) fields. Added VersionMVCCNetworkStats so that we
can do a version dance and continue to send
ReplicatedEvalResult.old_delta until all binaries are running a version
that can interpret ReplicatedEvalResult.delta.

This reduces the encoded size of stats in proposals from 63 bytes to 21
bytes in the common case.


petermattis:
@tschottdorf Heh, I imagine you're going to frown at this given our recent discussion.

pkg/storage/storagebase/proposer_kv.proto
```diff
@@ -92,7 +92,8 @@ message ReplicatedEvalResult {
   bool is_consistency_related = 9;
   // The stats delta corresponding to the data in this WriteBatch. On
   // a split, contains only the contributions to the left-hand side.
-  storage.engine.enginepb.MVCCNetworkStats delta = 10 [(gogoproto.nullable) = false];
+  storage.engine.enginepb.MVCCStats old_delta = 10;
```
tschottdorf: `deprecated_delta`? and a pointer at the version which deprecates it.

tbg (approved):
No frowning here! I feel like I understand this change (famous last words).

pkg/storage/storagebase/proposer_kv.proto
petermattis: Yeah, that's a better name. Done.

bdarnell:
:lgtm:

---

Reviewed 6 of 10 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv5BGIjKGgO1MQHN_OQ:bd9y322) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*
> ```Go
> var (
> 	// VersionMVCCNetworkStats is https://github.com/cockroachdb/cockroach/pull/18828.
> 	VersionMVCCNetworkStats = roachpb.Version{Major: 1, Minor: 1, Unstable: 2}
> ```

We should really be waiting until 1.1 is done before adding new versions like this. It would be messy if one of the last-minute 1.1 changes needed to insert a version "behind" these new ones. 

---

*[pkg/storage/raft_log_queue_test.go, line 265 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv59yMEtcQ2BsY2YwFT:-Kv59yMEtcQ2BsY2YwFU:b-f8jyq) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/storage/raft_log_queue_test.go#L265)):*
> ```Go
> 	}{
> 		// Lots of small KVs.
> 		{RaftLogQueueStaleThreshold * 4, 5},
> ```

Should we be doing something more future-proof here? For example, would `{RaftLogQueueStaleSize / 5, 5}` compute the right number of entries?

---

*[pkg/storage/engine/enginepb/mvcc_network_stats.proto, line 20 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5ADYJPCy18x230qIF:-Kv5ADYKlxA0AeSi10kL:b-irini7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/storage/engine/enginepb/mvcc_network_stats.proto#L20)):*
> ```protobuf
> import "gogoproto/gogo.proto";
> 
> // MVCCNetworkStats is convertible to MVCCStats, but uses variable width
> ```

The name `MVCCNetworkStats` should probably be changed. This encoding *is* persisted to disk, just not in the official range stats key. How about `MVCCStatsDelta`?

---

*[pkg/storage/engine/enginepb/mvcc_network_stats.proto, line 30 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5AdRbrFtShKNTdTPR:-Kv5AdRbrFtShKNTdTPS:b-yilmhx) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/storage/engine/enginepb/mvcc_network_stats.proto#L30)):*
> ```protobuf
>   sfixed64 intent_age = 2;
>   sfixed64 gc_bytes_age = 3 [(gogoproto.customname) = "GCBytesAge"];
>   int64 live_bytes = 4;
> ```

These values can be negative, and the `int64` type uses a very inefficient encoding for negative numbers (bigger than `sfixed64`). These should be `sint64` instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-:-Kv5BkwBhcHtUOj7UgTm:bnfp4nl)*

petermattis:
Review status: 7 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7hFSnJOPYoBAHjaQi:bimu2hf) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

We should really be waiting until 1.1 is done before adding new versions like this. It would be messy if one of the last-minute 1.1 changes needed to insert a version "behind" these new ones. 
</blockquote></details>

I think the 1.1 migrations used versions like `1.0.x`. So if we need another 1.1 migration we would add `1.0.4`. The 1.2 migrations will use values like `1.1.x`. Am I misunderstanding this? Cc @tschottdorf

---

*[pkg/storage/raft_log_queue_test.go, line 265 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv59yMEtcQ2BsY2YwFT:-Kv7oyEHiO3MR21qga7U:bb14jhd) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/storage/raft_log_queue_test.go#L265)):*

Should we be doing something more future-proof here? For example, would `{RaftLogQueueStaleSize / 5, 5}` compute the right number of entries?
</blockquote></details>

Did something similar, though with fewer entries.

---

*[pkg/storage/engine/enginepb/mvcc_network_stats.proto, line 20 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5ADYJPCy18x230qIF:-Kv7lOAZfdIOImg_pgHG:bp21ern) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/storage/engine/enginepb/mvcc_network_stats.proto#L20)):*

The name `MVCCNetworkStats` should probably be changed. This encoding *is* persisted to disk, just not in the official range stats key. How about `MVCCStatsDelta`?
</blockquote></details>

Yeah, the name needs adjustment. I'm actually planning on extending this structure in the future to contain the `{Raft,Lease}AppliedIndex`, so perhaps `MVCCRangeState` might make sense.

---

*[pkg/storage/engine/enginepb/mvcc_network_stats.proto, line 30 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5AdRbrFtShKNTdTPR:-Kv7lbTBy4LNpJuqp7fb:b-qu6n94) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/storage/engine/enginepb/mvcc_network_stats.proto#L30)):*

These values can be negative, and the `int64` type uses a very inefficient encoding for negative numbers (bigger than `sfixed64`). These should be `sint64` instead.
</blockquote></details>

Good point. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

bdarnell:
Review status: 7 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7pgdWXAgeBM1rP63L:b2w2168) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

I think the 1.1 migrations used versions like `1.0.x`. So if we need another 1.1 migration we would add `1.0.4`. The 1.2 migrations will use values like `1.1.x`. Am I misunderstanding this? Cc @tschottdorf
</blockquote></details>

Yes, you're right about the numbering, but the problem is that if we introduced 1.0-4, any clusters already running this master branch would have a greater version setting without actually running the migration. These clusters would probably need to be wiped. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

petermattis:
Review status: 7 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7sO-e1rEGfqs5JGs7:b-iz3am4) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

Yes, you're right about the numbering, but the problem is that if we introduced 1.0-4, any clusters already running this master branch would have a greater version setting without actually running the migration. These clusters would probably need to be wiped. 
</blockquote></details>

Right, though that would only affect clusters that have upgraded the version. Just using a master binary on a cluster that was previously running 1.1 won't bump the version unless I've misunderstood something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

bdarnell:
Review status: 7 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7tK9k41ykNuFkcHSZ:b5undrw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

Right, though that would only affect clusters that have upgraded the version. Just using a master binary on a cluster that was previously running 1.1 won't bump the version unless I've misunderstood something.
</blockquote></details>

Right, but starting up a new cluster from the master branch would have the higher version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

petermattis:
Review status: 7 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7tgPae3POfuH7MYRB:b-ecuxc7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

Right, but starting up a new cluster from the master branch would have the higher version.
</blockquote></details>

Yep. Reverting this stuff now doesn't seem particularly useful, but perhaps I should add some better commentary here about the effect of bumping `BinaryServerVersion`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

bdarnell:
Review status: 7 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7uS5Uyc2Icm_tx7Fs:brkttmh) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

Yep. Reverting this stuff now doesn't seem particularly useful, but perhaps I should add some better commentary here about the effect of bumping `BinaryServerVersion`.
</blockquote></details>

Yeah, given all the perils of running straight from master we shouldn't go back and fix this, I'm just pointing this out in light of our general policy on avoiding refactorings on master until the release is out the door.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

petermattis:
Review status: 6 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7vtqFNPjfULfYvAja:b-s4ckqo) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

Yeah, given all the perils of running straight from master we shouldn't go back and fix this, I'm just pointing this out in light of our general policy on avoiding refactorings on master until the release is out the door.
</blockquote></details>

This isn't the same as a refactoring as it doesn't burden cherry-picks. We probably shouldn't be downgrading clusters from master to the previous release without wiping as the instabilities on master can persist after the downgrade and cause us great consternation. I think there might be utility in bumping the minor version on master immediately when the release branch is created. This will prevent any accidental downgrades from master to the previous release on a newly created cluster. 

I've added a comment above explaining the danger in bumping `BinaryServerVersion`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

bdarnell:
Review status: 6 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7xeon7_wgtIKrR-cT:bpp4nr8) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

This isn't the same as a refactoring as it doesn't burden cherry-picks. We probably shouldn't be downgrading clusters from master to the previous release without wiping as the instabilities on master can persist after the downgrade and cause us great consternation. I think there might be utility in bumping the minor version on master immediately when the release branch is created. This will prevent any accidental downgrades from master to the previous release on a newly created cluster. 

I've added a comment above explaining the danger in bumping `BinaryServerVersion`.
</blockquote></details>

I'm not talking about downgrading from master to an older release. Adding new versions below `BinaryServerVersion` can break clusters that just move forward from one master release to the next, since they are unable to coordinate their activation of whatever change is controlled by the new version. 

Our versioning scheme is predicated on versions being (effectively) monotonic and gapless. The only safe way to add new versions is to add them on master and never cherry-pick them onto a branch without ensuring that that branch has *all* versions up to that point. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

petermattis:
Review status: 6 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv7zBPrZIQnThvFvDcs:b8n2vhr) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

I'm not talking about downgrading from master to an older release. Adding new versions below `BinaryServerVersion` can break clusters that just move forward from one master release to the next, since they are unable to coordinate their activation of whatever change is controlled by the new version. 

Our versioning scheme is predicated on versions being (effectively) monotonic and gapless. The only safe way to add new versions is to add them on master and never cherry-pick them onto a branch without ensuring that that branch has *all* versions up to that point. 
</blockquote></details>

Ah, got it. So when would be a safe time to add a new version? I can't imagine we'd want to cherry-pick a change into 1.1 at this point that added a new version. Nor can I imagine cherry-picking a change into 1.1.x that added a new version, but perhaps I'm lacking imagination there. Updated the commentary again. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

bdarnell:
Review status: 6 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/settings/cluster/cluster_version.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv5BGIjKGgO1MQHN_OP:-Kv81RT4B-Clpk_EEXrF:b-z6j69n) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7e98ccccb7d56fb78d81c097ba45c4e1f67d035/pkg/settings/cluster/cluster_version.go#L37)):*

Ah, got it. So when would be a safe time to add a new version? I can't imagine we'd want to cherry-pick a change into 1.1 at this point that added a new version. Nor can I imagine cherry-picking a change into 1.1.x that added a new version, but perhaps I'm lacking imagination there. Updated the commentary again. PTAL.
</blockquote></details>

I'd say anytime after release would be fine. I'd also hope we wouldn't need to cherry-pick anything that introduced a new version after the freeze, but the late introduction of the versioning mechanism combined with the amount of stuff that's gone in so late in this cycle makes it hard to be sure. (until I looked back at the logs I thought we had cherry-picked something with a new version into `release-1.1` at some point)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

bdarnell:
Review status: 6 of 10 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/settings/cluster/cluster_version.go, line 38 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv82WISfMOVDOgStWJv:-Kv82WISfMOVDOgStWJw:b5x1dck) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f0220ec66ceacc0733986e871442bb9cab11c4a/pkg/settings/cluster/cluster_version.go#L38)):*
> ```Go
> // perhaps MinimumSupportedVersion, if necessary). Note that the version
> // upgrade process requires the versions as seen by a cluster to be
> // monotonic. Once we've added 1.1.2 (VersionMVCCNetworkStats), we can't go
> ```

We style these as 1.1-2 to distinguish them from patch releases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

petermattis:
Review status: 6 of 10 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/settings/cluster/cluster_version.go, line 38 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828#-Kv82WISfMOVDOgStWJv:-Kv83L5bjEghsv5FVyA-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f0220ec66ceacc0733986e871442bb9cab11c4a/pkg/settings/cluster/cluster_version.go#L38)):*

We style these as 1.1-2 to distinguish them from patch releases.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18828)*

PR #18829, opened 2017-09-27
teamcity-reset-nightlies: clean up storage accounts too

benesch:
VM disks (VHDs) are backed by Azure storage accounts, which means the
VHDs leak whenever their VMs leak, which is frequently. Since the
storage accounts were previously associated with their own resource
group, deleting the cockroach-nightly resource group would not delete
the storage accounts. Prior to this commit, one of the storage accounts
had over 16TB worth of data.

Plug the leak by moving the storage accounts into the cockroach-nightly
resource group, so they get deleted when cockroach-nightly is deleted,
and teach teamcity-reset-nightlies to recreate them.

### DIFF
--- a/pkg/acceptance/terraform/azure/variables.tf
+++ b/pkg/acceptance/terraform/azure/variables.tf
@@ -60,11 +60,14 @@ variable "azure_vhd_storage_account" {
   # Using a static storage account name also has the benefit of making it
   # easier to expunge leaked VHDs, which continue to occur as of 2017-08-09.
   # Only lowercase letters and numbers are allowed by Azure.
-  default = "cockroachnightlyvhd"
+  #
+  # N.B. Keep this in sync with build/teamcity-reset-nightlies.sh.
+  default = "cockroachnightlyeastvhd"
 }
 
 variable "vhd_storage_container" {
-  # Must belong to ${var.azure_vhd_storage_account}.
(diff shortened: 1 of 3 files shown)


pkg/acceptance/terraform/azure/variables.tf
```diff
@@ -60,11 +60,14 @@ variable "azure_vhd_storage_account" {
   # Using a static storage account name also has the benefit of making it
   # easier to expunge leaked VHDs, which continue to occur as of 2017-08-09.
   # Only lowercase letters and numbers are allowed by Azure.
-  default = "cockroachnightlyvhd"
+  #
+  # N.B. Keep this in sync with build/teamcity-reset-nightlies.sh.
+  default = "cockroachnightlyeastvhd"
```
mjibson: This says `east` and backup_test.go below is `west`. Is that correct?

madelynnblue (approved):

pkg/acceptance/terraform/azure/variables.tf
benesch: Yup, exactly. eastus doesn't support Ls-series VMs 🤦‍♂️, which we need for the backup/restore tests.

benesch:
Merging so that the nightlies aren't broken tonight.

PR #18833, opened 2017-09-27
storage: try to annotate context with replica in HandleRaftResponse

nvb:
During debug sessions, we've noticed that certain Raft responses were
not logging with a replica tag even when they could have been. This
change annotates contexts in `HandleRaftResponse` with the corresponding
replica whenever possible, improving the grep-ability of these log messages.

### DIFF
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -3201,19 +3201,25 @@ func (s *Store) processRaftRequest(
 	return nil
 }
 
-// HandleRaftResponse implements the RaftMessageHandler interface.
+// HandleRaftResponse implements the RaftMessageHandler interface. Per the
+// interface specification, an error is returned if and only if the underlying
+// Raft connection should be closed.
 // It requires that s.mu is not held.
 func (s *Store) HandleRaftResponse(ctx context.Context, resp *RaftMessageResponse) error {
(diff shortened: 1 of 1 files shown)


pkg/storage/store.go
```diff
@@ -3228,26 +3234,24 @@ func (s *Store) HandleRaftResponse(ctx context.Context, resp *RaftMessageRespons
...
+				log.Errorf(ctx, "unable to add to replica GC queue: %s", err)
 			} else if added {
-				log.Infof(replCtx, "added to replica GC queue (peer suggestion)")
+				log.Infof(ctx, "added to replica GC queue (peer suggestion)")
 			}
 		case *roachpb.StoreNotFoundError:
 			log.Warningf(ctx, "raft error: node %d claims to not contain store %d for replica %s: %s",
 				resp.FromReplica.NodeID, resp.FromReplica.StoreID, resp.FromReplica, val)
-			return val.GetDetail()
+			return val.GetDetail() // close Raft connection.
```
tschottdorf: ignorable nit: no punctuation in inline comments 🙈

tbg (approved):

pkg/storage/store.go
nvanbenschoten: 🙈

Issue #18825, opened 2017-09-27
pm cluster r401 has diverged


tschottdorf:
```
Build Tag:    v1.1-alpha.20170817-1138-gfe3bf0e
Build Time:   2017/09/26 21:02:38
Distribution: CCL
Platform:     linux amd64
Go Version:   go1.9
C Compiler:   gcc 6.3.0
Build SHA-1:  fe3bf0e06aacbe75780fb9e33317457cfdb9c7bf
Build Type:   release-gnu
```

(but updated nightly, which I've disabled for now)

Key takeaways: the replicas are caught up (not in this picture, but I've reloaded the page many times and they basically are) and so they should be in the same state. But they're not, in particular:

- n2 has a newer lease
- n2 has a higher LeaseAppliedIndex
- the stats are different among all three of them
- n3 is behind by one TruncatedTerm

Interestingly, n2's newer lease thinks that n1 has the lease while the other two nodes think it's n1. This makes for the interesting situation in which everyone's lease requests fail due to having been based on an invalid proposer.

> cannot replace lease repl=(n1,s1):15 start=1506492612.062313478,0 epo=184 pro=1506492612.062316401,0 with repl=(n2,s2):13 start=1506521413.653484999,0 epo=124 pro=1506521413.653514920,0: 

This is obviously quite troubling and I'll try to dig up more information on what happened to this range.

![out](https://user-images.githubusercontent.com/5076964/30918653-2cf164d8-a36d-11e7-82ca-8c52b86f90c4.png)

tschottdorf:
The relevant period for the (combined) cluster logs is sparse:

```
170927 04:37:24.412805 89 storage/replica_proposal.go:453  [n4,s4,r401/12:/System/tsd/cr.node.exec.…] new range lease repl=(n4,s4):12 start=1506431353.294933057,0 epo=163 pro=1506487044.047877054,0 following repl=(n4,s4):12 start=1506431353.294933057,0 epo=162 pro=1506487028.879821657,0
170927 06:00:10.339505 110 storage/replica_proposal.go:453  [n2,s2,r401/13:/System/tsd/cr.node.exec.…] new range lease repl=(n2,s2):13 start=1506492010.162671820,0 epo=121 pro=1506492010.162674497,0 following repl=(n4,s4):12 start=1506431353.294933057,0 epo=163 pro=1506487044.047877054,0
[...]

PR #18835, opened 2017-09-27
roachpb: fix lease equivalence after nullability refactor

tbg:
In #18689, we changed the nullability of some fields from false to true to reduce the encoded size
of some protobufs, one of which was the `Lease` message.

This introduced a bug in which the "same" lease (but differing in nullability of one of the fields)
was considered non-equivalent.

This in turn lead to ranges that couldn't [get a new lease] and [replica divergence] through the
following mechanism:

- node 1 runs the nullable version, nodes 2 and 3 don't
- the lease has a zero (not nil!) Expiration
- node 2 sends a valid lease request
- all nodes update their in-memory state, but node 1 sees a zero field, 2 and 3 a nil one
- whenever node 1 sends a proposal under the "correct" lease, it sends a zero field
- whenever node 2 and 3 apply these commands, they discard them
- whenever node 1 applies these commands, well, it does!

I also saw the case of a range that had this message (in fact, this is where I got the scent):

```
on n3: command proposed from replica (n3,s3):69 with
repl=(n3,s3):69 start=1506492010.468037043,0 epo=138 pro=1506492010.468041892,0
incompatible to
repl=(n3,s3):69 start=1506492010.468037043,0 epo=138 pro=1506492010.468041892,0
```

(yes, this is `n3` deciding that its own lease isn't equivalent to itself). I think this basically
boils down to the fact that the lease request code doesn't verbatim pass the lease through but
unintentionally scrubs the nil fields:

```go
// ProposerLease will be the same as PrevLease, so with zero field
leaseReq = &roachpb.RequestLeaseRequest{
    Span:      reqSpan,
    Lease:     reqLease,
    PrevLease: status.Lease, // has the zero field
    // has no zero field
}
```

In effect this means that new versions that base their epoch leases off of a previous one will have
trouble (which is essentially all leases). This explains why the PM cluster ground to a halt.

PS: I'm pretty worried that we have more of those bugs lurking.

[get a new lease]: https://github.com/cockroachdb/cockroach/issues/18825#issuecomment-332563514
[replica divergence]: https://github.com/cockroachdb/cockroach/issues/18825#issue-261000347

### DIFF
--- a/pkg/roachpb/data.go
+++ b/pkg/roachpb/data.go
@@ -1118,10 +1118,19 @@ func (l Lease) Equivalent(ol Lease) bool {
 	// and then set to nil.
 	switch l.Type() {
 	case LeaseEpoch:
+		// Ignore expirations. This seems benign but since we changed the
+		// nullability of this field in the 1.2 cycle, it's crucial and
+		// tested in TestLeaseEquivalence.
+		l.Expiration, ol.Expiration = nil, nil
+
 		if ol.Epoch != nil && *l.Epoch == *ol.Epoch {
 			l.Epoch, ol.Epoch = nil, nil
 		}
 	case LeaseExpiration:
+		// See the comment above, though this field's nullability wasn't
+		// changed. We nil it out for completeness only.
+		l.Epoch, ol.Epoch = nil, nil
+
 		// For expiration-based leases, extensions are considered equivalent.
 		if !ol.GetExpiration().Less(l.GetExpiration()) {
 			l.Expiration, ol.Expiration = nil, nil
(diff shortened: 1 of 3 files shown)


petermattis:
:lgtm: nice job tracking this down. Apologies for creating this work in the first place. I now have much more paranoia about changing a field to nullable.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18835#-:-Kv3jZW_UoYQ16pIbhf5:bo5d8ga)*

Issue #18600, opened 2017-09-19
Bug: disk_row_container invalid row length 0, expected 2 when running TPC-H query 8 


richardwu:
When running [TPC-H's](https://github.com/cockroachdb/loadgen/tree/master/tpch) query 8 with a scale-factor of 1 (with default options everywhere else)
```
./tpch --queries=8
```
 on a 3-node cluster locally with environment variables `COCKROACH_PROPOSER_EVALUATED_KV=true COCKROACH_ENTERPRISE_ENABLED=true` (and with a CCL license):
```
./cockroach start --insecure --host=localhost --port=26257 --http-port=8080 --store=cockroach-data/node0 &
sleep 3
./cockroach start --insecure --host=localhost --port=26258 --http-port=8081 --store=cockroach-data/node1 --join=localhost:26257 &
./cockroach start --insecure --host=localhost --port=26259 --http-port=8082 --store=cockroach-data/node2 --join=localhost:26257 &

CockroachDB node starting at 2017-09-19 10:40:32.921068629 -0400 EDT m=+6.900215001 (took 6.9s)
build:      CCL 46fa327e6 @ 2017/09/19 14:40:06 (go1.9)
admin:      http://localhost:8082
sql:        postgresql://root@localhost:26259?application_name=cockroach&sslmode=disable
logs:       /Users/richardwu/go/src/github.com/cockroachdb/cockroach/cockroach-data/node2/logs
store[0]:   path=/Users/richardwu/go/src/github.com/cockroachdb/cockroach/cockroach-data/node2
status:     restarted pre-existing node
clusterID:  b72fae7c-032f-45dc-932d-21e7046cd796
nodeID:     2
CockroachDB node starting at 2017-09-19 10:40:32.999973842 -0400 EDT m=+6.979079430 (took 7.0s)
build:      CCL 46fa327e6 @ 2017/09/19 14:40:06 (go1.9)
admin:      http://localhost:8081
sql:        postgresql://root@localhost:26258?application_name=cockroach&sslmode=disable
logs:       /Users/richardwu/go/src/github.com/cockroachdb/cockroach/cockroach-data/node1/logs
store[0]:   path=/Users/richardwu/go/src/github.com/cockroachdb/cockroach/cockroach-data/node1
status:     restarted pre-existing node
clusterID:  b72fae7c-032f-45dc-932d-21e7046cd796
nodeID:     3
[...]

PR #18838, opened 2017-09-27
distsql: pre-reserve memory needed to mark rows in HashJoiner build phase

asubiotto:
A situation was uncovered by #18600, where the HashJoiner would run out
of memory in the probe phase. This was because we had made an assumption
that we wouldn't hit a memory limit if the buffer phase filled up at
most 2/3 of the limit with both streams , since the marking
infrastructure would take up only a fraction of 1/3 (the chosen stream).

This assumption failed to take into account other limits shared with
other processors. This change pre-reserves the memory needed for the
probe phase in the build phase so that we can keep a single point in the
code where we fall back to disk while not relying on any limit
assumptions.

### DIFF
--- a/pkg/sql/distsqlrun/hash_row_container.go
+++ b/pkg/sql/distsqlrun/hash_row_container.go
@@ -119,9 +119,7 @@ const sizeOfBool = int64(unsafe.Sizeof(false))
 // memRowContainer corresponding to matching rows.
 // NOTE: Once a row is marked, adding more rows to the hashMemRowContainer
 // results in undefined behavior. It is not necessary to do otherwise for the
-// current usage of hashMemRowContainer and allows us to assume that a memory
-// error can only occur at the start of the marking phase, thus not having to
-// deal with half-emitted buckets and marks when falling back to disk.
+// current usage of hashMemRowContainer.
 type hashMemRowContainer struct {
 	*memRowContainer
 	columnEncoder
@@ -133,6 +131,11 @@ type hashMemRowContainer struct {
 	// marked specifies for each row in memRowContainer whether that row has
 	// been marked. Used for iterating over unmarked rows.
 	marked []bool
+
+	// markMemoryReserved specifies whether the hashMemRowContainer's memory
+	// account already accounts for the memory needed to mark the rows in the
+	// hashMemRowContainer.
+	markMemoryReserved bool
 
 	// buckets contains the indices into memRowContainer for a given group
 	// key (which is the encoding of storedEqCols).
@@ -219,6 +222,20 @@ func (h *hashMemRowContainer) addRowToBucket(
 	return nil
 }
 
+// reserveMarkMemoryMaybe is a utility function to grow the
+// hashMemRowContainer's memory account by the memory needed to mark all rows.
+// It is a noop if h.markMemoryReserved is true.
+func (h *hashMemRowContainer) reserveMarkMemoryMaybe(ctx context.Context) error {
+	if h.markMemoryReserved {
+		return nil
+	}
+	if err := h.bucketsAcc.Grow(ctx, sizeOfBoolSlice+(sizeOfBool*int64(h.Len()))); err != nil {
+		return err
+	}
+	h.markMemoryReserved = true
+	return nil
+}
+
 // hashMemRowBucketIterator iterates over the rows in a bucket.
 type hashMemRowBucketIterator struct {
 	*hashMemRowContainer
@@ -267,8 +284,8 @@ func (i *hashMemRowBucketIterator) Mark(ctx context.Context, mark bool) error {
 		log.Fatal(ctx, "hash mem row container not set up for marking")
 	}
 	if i.marked == nil {
-		if err := i.bucketsAcc.Grow(ctx, sizeOfBoolSlice+(sizeOfBool*int64(i.Len()))); err != nil {
-			return err
+		if !i.markMemoryReserved {
+			panic("mark memory should have been reserved already")
 		}
 		i.marked = make([]bool, i.Len())
 	}
--- a/pkg/sql/distsqlrun/hashjoiner.go
+++ b/pkg/sql/distsqlrun/hashjoiner.go
@@ -69,7 +69,8 @@ func (p hashJoinPhase) String() string {
 //     choose the right stream and read it and buffer it until the end.
 //
 //  2. Build phase: in this phase we build the buckets from the rows stored
-//     in the first phase.
+//     in the first phase. If temp storage is enabled and we ran out of memory
+//     in the buffer phase, this phase falls back to disk.
 //
 //  3. Probe phase: in this phase we process all the rows from the other stream
 //     and look for matching rows from the stored stream using the map.
@@ -176,16 +177,10 @@ func (h *hashJoiner) Run(ctx context.Context, wg *sync.WaitGroup) {
 		limitedMon.Start(ctx, rowContainerMon, mon.BoundAccount{})
 		defer limitedMon.Stop(ctx)
 
-		// Override initialBufferSize to be a third of this processor's memory
+		// Override initialBufferSize to be half of this processor's memory
 		// limit. We consume up to h.initialBufferSize bytes from each input
-		// stream. If the chosen stream is fully consumed and does not go over
-		// this limit, h.initialBufferSize bytes (the last third) are budgeted
-		// to construct a hash map from these rows. We do not expect the hash
-		// map structure to consume more than the memory used to store the rows
-		// themselves.
-		// This assumption allows us to only worry about falling back to disk in
-		// the buffer phase.
-		h.initialBufferSize = limit / 3
+		// stream.
+		h.initialBufferSize = limit / 2
 
 		rowContainerMon = &limitedMon
 	}
@@ -195,51 +190,60 @@ func (h *hashJoiner) Run(ctx context.Context, wg *sync.WaitGroup) {
 	defer h.rows[leftSide].Close(ctx)
 	defer h.rows[rightSide].Close(ctx)
 
-	var storedRows hashRowContainer
-	defer func() {
-		if storedRows != nil {
-			storedRows.Close(ctx)
-		}
-	}()
+	row, earlyExit, err := h.bufferPhase(ctx)
 
-	if earlyExit, err := h.bufferPhase(ctx, useTempStorage, &storedRows); earlyExit || err != nil {
+	bufferPhaseOom := false
+	if pgErr, ok := pgerror.GetPGCause(err); ok && pgErr.Code == pgerror.CodeOutOfMemoryError {
+		bufferPhaseOom = true
+	}
+
+	// Exit if earlyExit is true or we received an error that is not a memory
+	// error.
+	if earlyExit || !(err == nil || bufferPhaseOom) {
 		if err != nil {
 			// We got an error. We still want to drain. Any error encountered while
 			// draining will be swallowed, and the original error will be forwarded to
 			// the consumer.
-			log.Infof(ctx, "initial buffering phase error %s", err)
+			log.Infof(ctx, "buffer phase error %s", err)
 		}
 		DrainAndClose(ctx, h.out.output, err /* cause */, h.leftSource, h.rightSource)
 		return
+	}
+
+	// Build hashRowContainer. If we didn't get an oom error in the buffer
+	// phase, the build phase will attempt to build an in-memory representation,
+	// falling back to disk according to useTempStorage.
+	storedRows, earlyExit, err := h.buildPhase(
+		ctx, useTempStorage, !bufferPhaseOom, /* attemptMemoryBuild */
+	)
+	if earlyExit || err != nil {
+		if err != nil {
+			// We got an error. We still want to drain. Any error encountered while
+			// draining will be swallowed, and the original error will be forwarded to
+			// the consumer.
+			log.Infof(ctx, "build phase error %s", err)
+		}
+		DrainAndClose(ctx, h.out.output, err /* cause */, h.leftSource, h.rightSource)
+		return
+	}
+	defer storedRows.Close(ctx)
+
+	// If the buffer phase returned a row, this row has not been added to any
+	// row container, since it is the row that caused a memory error. We add
+	// this row to our built container (which is a hashDiskRowContainer in this
+	// case).
+	if row != nil {
+		if err := storedRows.AddRow(ctx, row); err != nil {
+			log.Infof(ctx, "unable to add row to disk %s", err)
+			DrainAndClose(ctx, h.out.output, err /* cause */, h.leftSource, h.rightSource)
+			return
+		}
 	}
 
 	// From this point, we are done with the source for h.storedSide.
 	srcToClose := h.leftSource
 	if h.storedSide == leftSide {
 		srcToClose = h.rightSource
-	}
-
-	// If storedRows is not nil, there was a memory limit reached in the buffer
-	// phase so we have already fallen back to a disk-based hashRowContainer.
-	// Otherwise, we have to build an in-memory hashRowContainer from
-	// h.rows[h.storedSide].
-	if storedRows == nil {
-		storedMemRows := makeHashMemRowContainer(&h.rows[h.storedSide])
-		if err := storedMemRows.Init(
-			ctx,
-			shouldEmitUnmatchedRow(h.storedSide, h.joinType),
-			h.rows[h.storedSide].types,
-			h.eqCols[h.storedSide],
-		); err != nil {
-			// We got an error. We still want to drain. Any error encountered
-			// while draining will be swallowed, and the original error will be
-			// forwarded to the consumer.
-			err = errors.Wrap(err, "error creating hash row container")
-			log.Info(ctx, err)
-			DrainAndClose(ctx, h.out.output, err /* cause */, h.leftSource, h.rightSource)
-			return
-		}
-		storedRows = &storedMemRows
 	}
 
 	log.VEventf(ctx, 1, "build phase complete")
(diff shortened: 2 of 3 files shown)


arjunravinarayan:
I'm not super happy with this PR, because it seems like a little more refactoring than absolutely necessary, but you should take my opinion with a grain of salt, because I'm working with a healthy amount of :dog: and am not sure I can articulate clearly whether you can make this PR more minimal. In either case, great job on fixing this bug. I'd like @andreimatei to weigh in before we merge this.

:LGTM:

---

Reviewed 3 of 3 files at r1.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 198 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8CVfWPfEP387xwNU2:-Kv8CVfWPfEP387xwNU3:b1eq9t1) ([raw file](https://github.com/cockroachdb/cockroach/blob/e203bd9ac79ebfef32978a155f408a5e78cf1380/pkg/sql/distsqlrun/hashjoiner.go#L198)):*
> ```Go
> 	defer h.rows[rightSide].Close(ctx)
> 
> 	var storedRows hashRowContainer
> ```

Why rope in so much code-move refactoring into this commit (moving some amount of logic into `buildPhase`?) It's not entirely clear to me why it's necessary.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 305 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8D5768eCuK1esoJOB:-Kv8D5768eCuK1esoJOC:b-5luff) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L305)):*
> ```Go
> //    in-memory hashRowContainer representation.
> func (h *hashJoiner) buildPhase(
> 	ctx context.Context, useTempStorage bool, buildMemory bool,
> ```

s/`buildMemory`/`buildInMemory`/ makes this clearer, I was confused momentarily.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-:-Kv8QSMuE_uNMrSWOXAb:b-97y5z4)*

asubiotto:
Thanks for the review. I definitely thought about the amount of refactoring while writing this change but I think that it's small enough to be OK. The issue stems from the fact that I blurred the line between the buffer and build phases in a previous change: we build from disk if we encounter a memory error in the buffer phase and then in the level above, build from memory if we did not encounter a memory error.

The solution to this bug requires us to have a memHashRowContainer built so that we can pre-reserve the memory and fall back to disk if anything goes wrong. Thus the memHashRowContainer must be built before the diskHashRowContainer. This means either moving the in memory build phase to precede the disk build phase in the buffer phase or extract the disk build phase into a function that is called after building in-memory and where it currently is. I decided to clean up my previous mistake and instead move the disk build phase out of the buffer phase and put the build phase (of both memory and disk) behind a function. I think the code is better off this way and the change is small/localized enough that it will not introduce cherrypicking woes.

I have no issue with going with the other solution though so please let me know what you think.

andreimatei:
:lgtm:

I think the refactoring was quite necessary (and it's good we're cherry-picking it). Ideally I think it would have been better to do it in a different commit than the memory reservation change, but I shouldn't be the one to cast that stone :)

---

Reviewed 1 of 3 files at r1.

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 122 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8ohUa2uHFrMvNHlFW:-Kv8ohUa2uHFrMvNHlFX:b-j2pyyv) ([raw file](https://github.com/cockroachdb/cockroach/blob/e203bd9ac79ebfef32978a155f408a5e78cf1380/pkg/sql/distsqlrun/hash_row_container.go#L122)):*
> ```Go
> // NOTE: Once a row is marked, adding more rows to the hashMemRowContainer
> // results in undefined behavior. It is not necessary to do otherwise for the
> // current usage of hashMemRowContainer and allows us to assume that a memory
> ```

Is the undefined behavior described by this note still necessary given that you're seemingly taking away the reason for it?

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 228 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8p71chKVsKyL-2i9d:-Kv8p71chKVsKyL-2i9e:b-v8zieb) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hash_row_container.go#L228)):*
> ```Go
> // memory account by the memory needed to mark all rows. It is a noop if
> // h.markMemoryReserved is true.
> func (h *hashMemRowContainer) reserveMarkMemory(ctx context.Context) error {
> ```

if the `!h.markMemoryReserved` condition is important, I'd rename this to `reserveMarkMemoryMaybe()`

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 229 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8p0DyUutWQZ94ozlw:-Kv8p0DyUutWQZ94ozlx:b-exxew4) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hash_row_container.go#L229)):*
> ```Go
> // h.markMemoryReserved is true.
> func (h *hashMemRowContainer) reserveMarkMemory(ctx context.Context) error {
> 	if !h.markMemoryReserved {
> ```

nit: invert the condition and return early so you can unident the bulk of this function.

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 286 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv94ouomHSZyZ05w0Oz:-Kv94ouomHSZyZ05w0P-:b-fm7dpk) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hash_row_container.go#L286)):*
> ```Go
> 	}
> 	if i.marked == nil {
> 		if err := i.reserveMarkMemory(ctx); err != nil {
> ```

do we expect for this call to ever not be a no-op? If no, I'd find a way to assert that memory has already been reserved.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 210 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv91c3Iva51-n4Kooop:-Kv91c3Iva51-n4Koooq:bui60ay) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L210)):*
> ```Go
> 	// attempt to build an in-memory representation, falling back to disk
> 	// according to useTempStorage.
> 	storedRows, earlyExit, err := h.buildPhase(ctx, useTempStorage, err == nil /* buildMemory */)
> ```

I find everything around the use of this `err` very hard to read. The fact that `err != nil`at this point implies that it's a OOM error is quite obscure, and also the condition above for calling `DrainAndClose()` is a bit inscrutable. 
I suggest introducing a `oom bool` above.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 223 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv92amxRxktSl77pUEn:-Kv92amxRxktSl77pUEo:b-orfupl) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L223)):*
> ```Go
> 	defer storedRows.Close(ctx)
> 
> 	// Add the row that caused a memory error in the buffer phase if there is
> ```

Shouldn't this be moved above the  call to `buildPhase()`? Seems very randomly placed here.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 298 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv919yXIvZ0psdzF5h2:-Kv919yXIvZ0psdzF5h3:b-5h5062) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L298)):*
> ```Go
> 
> // buildPhase constructs our internal hash map of rows seen. This is done
> // entirely from one stream (chosen during initial buffering).
> ```

nit: hint to h.storedSide

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 303 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv9-p-BEj1v5r79KOml:-Kv9-p-BEj1v5r79KOmm:b-2o6zl3) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L303)):*
> ```Go
> //    storage if necessary.
> //  - buildMemory specifies whether the build phase should attempt to build an
> //    in-memory hashRowContainer representation.
> ```

as opposed to what? Consider giving a hint as to why this would ever be set to false - I guess it's when we already happen to know that there's no memory available. Then again, why not let the code always try to use memory and fallback?
Also, if we keep the arg, we should document and assert that useTempStorage is set when this is set.

Final nit: if we keep the arg, I'd suggest we name them `attemptMemoryBuild` and `fallbackToDiskBuild`.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 324 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv950U6dDRqwrUHoN-P:-Kv950U6dDRqwrUHoN-Q:b4c9lyr) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L324)):*
> ```Go
> 		// in-memory rows in this phase to not have to worry about hitting a
> 		// memory limit in the probe phase.
> 		err := storedMemRows.reserveMarkMemory(ctx)
> ```

:thumbsup: 

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 403 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8qDUmwXMpsvTuxSqR:-Kv8qDUmwXMpsvTuxSqS:b972kgl) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L403)):*
> bufferPhase attempts to read a portion

It doesn't "attempt to read", it reads :)

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 405 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8qHka72M3J-ghmx_8:-Kv8qHka72M3J-ghmx_9:b667xne) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L405)):*
> ```Go
> // bufferPhase attempts to read a portion of both streams into memory (up to
> // h.initialBufferSize) in the hope that one of them is small and should be used
> // as h.storedSide. The phase attempts to consume all the rows from the chosen
> ```

Is the last sentence finished?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-:-Kv95QxHPXy0JPlax4yl:b-aqnvpz)*

asubiotto:
Review status: 1 of 3 files reviewed at latest revision, 13 unresolved discussions.

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 122 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8ohUa2uHFrMvNHlFW:-KvSZS3EZKi8LSfth2-3:bimcpvz) ([raw file](https://github.com/cockroachdb/cockroach/blob/e203bd9ac79ebfef32978a155f408a5e78cf1380/pkg/sql/distsqlrun/hash_row_container.go#L122)):*

Is the undefined behavior described by this note still necessary given that you're seemingly taking away the reason for it?
</blockquote></details>

The only reason there is undefined behavior is because it is not necessary to support adding more rows once we start marking. It is not necessary behavior, simply not worth time.

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 228 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8p71chKVsKyL-2i9d:-KvS_sMpVM5VYYcJ3ZTf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hash_row_container.go#L228)):*

if the `!h.markMemoryReserved` condition is important, I'd rename this to `reserveMarkMemoryMaybe()`
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 229 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8p0DyUutWQZ94ozlw:-KvSalA3nm-yKwL_2Owm:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hash_row_container.go#L229)):*

nit: invert the condition and return early so you can unident the bulk of this function.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/hash_row_container.go, line 286 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv94ouomHSZyZ05w0Oz:-KvSeFhA1dnSJ_UFfn9P:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hash_row_container.go#L286)):*

do we expect for this call to ever not be a no-op? If no, I'd find a way to assert that memory has already been reserved.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 198 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8CVfWPfEP387xwNU2:-KvSb9uuev1manDkgzv_:bgifv49) ([raw file](https://github.com/cockroachdb/cockroach/blob/e203bd9ac79ebfef32978a155f408a5e78cf1380/pkg/sql/distsqlrun/hashjoiner.go#L198)):*

Why rope in so much code-move refactoring into this commit (moving some amount of logic into `buildPhase`?) It's not entirely clear to me why it's necessary.
</blockquote></details>

Explained in main thread.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 210 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv91c3Iva51-n4Kooop:-KvSlZDhcyLZCZNLDHJh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L210)):*

I find everything around the use of this `err` very hard to read. The fact that `err != nil`at this point implies that it's a OOM error is quite obscure, and also the condition above for calling `DrainAndClose()` is a bit inscrutable. 
I suggest introducing a `oom bool` above.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 223 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv92amxRxktSl77pUEn:-KvSlm6A_tf0znEVfrFZ:but4ewm) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L223)):*

Shouldn't this be moved above the  call to `buildPhase()`? Seems very randomly placed here.
</blockquote></details>

This needs to be after `buildPhase` since we are adding the row to the built container. Rewrote the comment.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 298 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv919yXIvZ0psdzF5h2:-KvStG-qNs4C6NfkK1Ks:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L298)):*

nit: hint to h.storedSide
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 303 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv9-p-BEj1v5r79KOml:-KvSyltldhiZy0Qcue9o:bswbfmu) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L303)):*

as opposed to what? Consider giving a hint as to why this would ever be set to false - I guess it's when we already happen to know that there's no memory available. Then again, why not let the code always try to use memory and fallback?
Also, if we keep the arg, we should document and assert that useTempStorage is set when this is set.

Final nit: if we keep the arg, I'd suggest we name them `attemptMemoryBuild` and `fallbackToDiskBuild`.
</blockquote></details>

Updated comment.

Since we already have knowledge of whether we should fall back to disk or not (disregarding marking), why not use that knowledge? I prefer to keep this as it is.

re assertions I'm not sure what you mean. That we assert that `useTempStorage` is true when `buildMemory` is true? It's a valid case to try to build in memory, try fall back to disk, and see that external storage is disabled.

Renamed `buildMemory` but I think `fallbackToDiskBuild` is more confusing than `useTempStorage`. This is what the boolean is called at the caller level as well (where we figure out whether we can use temp storage)

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 305 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8D5768eCuK1esoJOB:-KvT0q2I92qJj6dhsosU:b13gejq) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L305)):*

s/`buildMemory`/`buildInMemory`/ makes this clearer, I was confused momentarily.
</blockquote></details>

Renamed to `attemptMemoryBuild`

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 403 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8qDUmwXMpsvTuxSqR:-KvT0xaqxE6WpH9X0XKB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L403)):*

> bufferPhase attempts to read a portion

It doesn't "attempt to read", it reads :)

</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/hashjoiner.go, line 405 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838#-Kv8qHka72M3J-ghmx_8:-KvT1Z8UN8I7tg8GC3my:bjhrbs9) ([raw file](https://github.com/cockroachdb/cockroach/blob/51a8153ff04df55282c50812673473e2300c5d64/pkg/sql/distsqlrun/hashjoiner.go#L405)):*

Is the last sentence finished?
</blockquote></details>

Yes, updated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18838)*

Issue #15894, opened 2017-05-12
migrations: Improve the experience of restarting a server without quorum


a-robinson:
It has now happened multiple times that a user has created a 3 node cluster, stopped them all, and then tried to restart them one at a time using `--background`. Unfortunately, the first server can't successfully come up without at least one other node's kv subsystem being up and running so that there's a quorum of nodes to enable checking for migrations. It's confusing that the way you brought up your nodes the first time (one at a time) doesn't work for subsequent restarts. This happened most recently in #15889.

We should make this better in 1.1. That could mean one of two things:

1. Shout a message to stdout explaining why the server isn't starting if it gets stuck waiting on kv requests in the migration package.
1. Put an optimization into the migrations package to locally persist the migrations that it has run (rather than only storing it in kv) such that it can determine whether or not it has run all migrations without needing to contact the rest of the cluster.

dianasaur323:
@a-robinson I also marked this down as a common issue. Is there a solution we could consider by documenting how to correctly restart a cluster?

dianasaur323:
nvm, I see it here: https://github.com/cockroachdb/docs/issues/1403

bdarnell:
It's not just migrations, it's that `--background` means "block until ready to serve", which requires contact with the leader of range 1. As I said in https://github.com/cockroachdb/cockroach/issues/15889#issuecomment-301337898, `--background` was designed for single-node use and probably doesn't make sense for multi-node clusters. We should be encouraging either foreground processes in separate shells or the shell `&` operator instead of `--background`.

PR #18840, opened 2017-09-27
cherrypick-1.1: server: add warning message for freezing at startup with --background

m-schneider:
Before if a user restarted a server with the --background flag the node
would freeze at startup waiting for another node for quorum.

Now the node will show a warning message if the server doesn't come up
within 30 seconds.

Closes: #15894


bdarnell (approved):

jseldess:
No doc updates needed here.

PR #18843, opened 2017-09-27
sql: modify BYTES->STRING casts to match postgres

### LINKED ISSUE #17969: teamcity: failed tests on master: test/TestDumpRandom
The following tests appear to have failed:

[#336579](https://teamcity.cockroachdb.com/viewLog.html?buildId=336579):

```
--- FAIL: test/TestDumpRandom (1.490s)
test_log_scope.go:78: test logs captured to: /go/src/github.com/cockroachdb/cockroach/artifacts/logTestDumpRandom278895581
test_log_scope.go:61: use -show-logs to present logs inline
dump_test.go:360: random seed: 1560662087743386392
dump_test.go:370: random iteration 0: 203 rows
dump_test.go:370: random iteration 1: 451 rows
dump_test.go:471: pq: could not parse '\X �
	>���:I������X�%k���'	<��1 
------- Stdout: -------
test logs left over in: /go/src/github.com/cockroachdb/cockroach/artifacts/logTestDumpRandom278895581

```
Please assign, take a look and update the issue accordingly.

justinj:
Fixes #17969.

Also introduce the hex form of ENCODE/DECODE to replace this
functionality.

Detailed explanation of why this was necessary:
https://github.com/cockroachdb/cockroach/issues/17969#issuecomment-331317893

This is a backwards-incompatible (but important) change and I think it
should be cherry-picked to 1.1.


knz:

---

*[pkg/sql/parser/datum.go, line 1018 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4DvfT3bdw9CHMigD3:-Kv4DvfU4RHrs1ldEhYA:bw9wgwm) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/datum.go#L1018)):*
> ```Go
> func (d *DBytes) Format(buf *bytes.Buffer, f FmtFlags) {
> 	buf.WriteByte('b')
> ```

Use WriteString instead of 2 calls to WriteByte.

---

*[pkg/sql/parser/datum.go, line 1022 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4Dypm6hnLNLl8F-yn:-Kv4Dypm6hnLNLl8F-yo:bhj8yvn) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/datum.go#L1022)):*
> ```Go
> 	b := string(*d)
> 	buf.WriteString("\\x")
> 	buf.WriteString(hexEncodeString(b))
> ```

`hexEncodeString(buf, b)`

---

*[pkg/sql/parser/encode.go, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4EBTni71tsrBwHZQw:-Kv4EBTni71tsrBwHZQx:b1297qa) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/encode.go#L45)):*
> ```Go
> 	}
> 	// hexMap is a mapping from each byte to the `\x%%` hex form as a []byte.
> 	hexMap [256][]byte
> ```

Huh wait I hadn't seen that earlier. Don't do that!

instead of writing

`buf.WriteBytes(hexMap[v])`

do

```
buf.WriteString("\x")
writeHexDigit(buf, v/16)
writeHexDigit(buf, v%16)
```

with 

```
func writeHexDigit(buf *bytes.Buffer, v int) {
   if v < 10 {
       buf.WriteByte('0' + v)
   } else {
      buf.WriteByte('a' + (v - 16))
  }
}
```

ditto for `rawHexMap`.

---

*[pkg/sql/parser/encode.go, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4DrOyk6jsnl8ZVyeb:-Kv4DrOzXtMOWUYtdFPJ:bjm5cek) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/encode.go#L65)):*
> ```Go
> }
> 
> func hexEncodeString(in string) string {
> ```

Make this take a `bytes.Buffer` as argument and use it.

---

*[pkg/sql/parser/eval.go, line 2424 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4Er8-PWRtiJzszkxJ:-Kv4Er8-PWRtiJzszkxK:b-vu906w) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/eval.go#L2424)):*
> ```Go
> 			s = t.Contents
> 		case *DBytes:
> 			s = "\\x" + hexEncodeString(string(*t))
> ```

bytes.Buffer

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

knz:
LGTM on principle, nit on implementation

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-:-Kv4FQaG54yC0yVeweyg:bkrnw2e)*

justinj:
Review status: 7 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/parser/datum.go, line 1018 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4DvfT3bdw9CHMigD3:-Kv4YkSAzkTlNIdFnuUL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/datum.go#L1018)):*

Use WriteString instead of 2 calls to WriteByte.
</blockquote></details>

Done.

---

*[pkg/sql/parser/datum.go, line 1022 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4Dypm6hnLNLl8F-yn:-Kv4YtB-mUcbSdBqjs4D:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/datum.go#L1022)):*

`hexEncodeString(buf, b)`
</blockquote></details>

Done.

---

*[pkg/sql/parser/encode.go, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4EBTni71tsrBwHZQw:-Kv4Z0c2HBFW7UUzUvm7:b-c4nq7w) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/encode.go#L45)):*

Huh wait I hadn't seen that earlier. Don't do that!

instead of writing

`buf.WriteBytes(hexMap[v])`

do

```
buf.WriteString("\x")
writeHexDigit(buf, v/16)
writeHexDigit(buf, v%16)
```

with 

```
func writeHexDigit(buf *bytes.Buffer, v int) {
   if v < 10 {
       buf.WriteByte('0' + v)
   } else {
      buf.WriteByte('a' + (v - 16))
  }
}
```

ditto for `rawHexMap`.

</blockquote></details>

What makes you say this? Admittedly the perf of this section probably isn't critical, but a [benchmark](https://gist.github.com/justinj/f08ee41342b6d238d40825806f78b1b3) implies to me that the `hexMap` way is a little faster:
```
BenchmarkSlice-8          500000              2928 ns/op
BenchmarkFunc-8           500000              3874 ns/op
```

---

*[pkg/sql/parser/encode.go, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4DrOyk6jsnl8ZVyeb:-Kv4ZLChhsejeOoR8Xwh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/encode.go#L65)):*

Make this take a `bytes.Buffer` as argument and use it.
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval.go, line 2424 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4Er8-PWRtiJzszkxJ:-Kv4ZTNVvybWMcw9T56o:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/eval.go#L2424)):*

bytes.Buffer
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

knz:
Reviewed 4 of 4 files at r2.

---

*[pkg/sql/parser/encode.go, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4EBTni71tsrBwHZQw:-Kv7sttzFym4gzw1m2ri:b-oans63) ([raw file](https://github.com/cockroachdb/cockroach/blob/bde674337e157fdafebf16a22e43f5e516a9d597/pkg/sql/parser/encode.go#L45)):*

What makes you say this? Admittedly the perf of this section probably isn't critical, but a [benchmark](https://gist.github.com/justinj/f08ee41342b6d238d40825806f78b1b3) implies to me that the `hexMap` way is a little faster:
```
BenchmarkSlice-8          500000              2928 ns/op
BenchmarkFunc-8           500000              3874 ns/op
```
</blockquote></details>

Ahem. 

I was about to write that such a benchmark is not realistic, because in most cases you don't print many hex representations in a row, and then the cache miss overhead of pulling an array off memory is going to dominate, which a benchmark doesn't show. 

But then I noticed we use this to print a bunch of bytes in a row, which amortizes the memory cost, so I'll just shut up. However see my other comment below.

---

*[pkg/sql/parser/encode.go, line 252 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv7tH6TkTnYV5d4KSh_:-Kv7tH6TkTnYV5d4KSha:bu2vk5d) ([raw file](https://github.com/cockroachdb/cockroach/blob/a1ee7c68f5e1f66bb9522420e2dad43721e1f9cd/pkg/sql/parser/encode.go#L252)):*
> ```Go
> 
> 	for i := range hexMap {
> 		hexMap[i] = []byte(fmt.Sprintf("\\x%02x", i))
> ```

This is alloc-heavy. Change to pre-allocate a bytes buffer of 768 bytes, fill it and assign the slices into hexMap.
Ditto below (512 bytes).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

justinj:
Ran into some issues since I didn't fully understand how strings beginning with `b` are scanned - should be fixed now.

---

Review status: 4 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/parser/encode.go, line 252 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv7tH6TkTnYV5d4KSh_:-KvDu46XbA5wAxyxzRGo:b-p3rx6c) ([raw file](https://github.com/cockroachdb/cockroach/blob/a1ee7c68f5e1f66bb9522420e2dad43721e1f9cd/pkg/sql/parser/encode.go#L252)):*

This is alloc-heavy. Change to pre-allocate a bytes buffer of 768 bytes, fill it and assign the slices into hexMap.
Ditto below (512 bytes).
</blockquote></details>

Very good idea!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-:-KvCyWOsf6nZcFEH2GTU:b-tm1pe3)*

justinj:
@jordanlewis would you mind taking a look at this as well? I think this should go into 1.1 but the timeline makes me a bit nervous, wouldn't mind some extra eyes.

pkg/sql/parser/eval_test.go
```diff
@@ -606,7 +606,7 @@ func TestEval(t *testing.T) {
 		{`CAST('123' AS int) + 1`, `124`},
 		{`CAST(NULL AS int)`, `NULL`},
 		{`'hello'::char(2)`, `'he'`},
-		{`'hello'::bytes`, `b'hello'`},
+		{`'hello'::bytes`, `b'\x68\x65\x6c\x6c\x6f'`},
```
jordanlewis: Like we discussed I think it would be better to have this stringify to '\x68656c6c6f' for pg compatibility.

justinj:
Review status: 1 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/parser/builtins.go, line 684 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4CdhMqK-pl3GNUU53-r3-684:-KvYbMsXjr9oh2KiDSvx:b-qf019f) ([raw file](https://github.com/cockroachdb/cockroach/blob/b32324e0305d6fc0154abee66825bbf716668bb4/pkg/sql/parser/builtins.go#L684)):*

probably should link to the relevant pg documentation for these two functions.
</blockquote></details>

Do we usually do that for pg functions? Regardless, done (and added `Info` field)

---

*[pkg/sql/parser/eval_test.go, line 609 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4CdhMqK-pl3GNUU58-r3-609:-KvYqSGzHT75MtuD8gSs:bawdy60) ([raw file](https://github.com/cockroachdb/cockroach/blob/b32324e0305d6fc0154abee66825bbf716668bb4/pkg/sql/parser/eval_test.go#L609)):*

Like we discussed I think it would be better to have this stringify to '\x68656c6c6f' for pg compatibility.
</blockquote></details>

Unfortunately, like we discussed after, this breaks distSQL (for specifics not 100% clear to me), but more fundamentally it seems like there is an (unspoken?) invariant that the type of a datum should be implied by its formatting (at least under `FmtParsable`) which this formatting would not satisfy. Not sure what the appropriate thing to do is here.

Note that we can't change this to `b'\x68656c6c6f'` because only the first two digits will be considered hex, and the rest as literals, unless we change how the `b` sigil is read.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

knz:
Review status: 1 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/sql/parser/eval_test.go, line 609 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4CdhMqK-pl3GNUU58-r3-609:-KvYumxhTVmJIVJEQiQ7:bazubps) ([raw file](https://github.com/cockroachdb/cockroach/blob/b32324e0305d6fc0154abee66825bbf716668bb4/pkg/sql/parser/eval_test.go#L609)):*

Unfortunately, like we discussed after, this breaks distSQL (for specifics not 100% clear to me), but more fundamentally it seems like there is an (unspoken?) invariant that the type of a datum should be implied by its formatting (at least under `FmtParsable`) which this formatting would not satisfy. Not sure what the appropriate thing to do is here.

Note that we can't change this to `b'\x68656c6c6f'` because only the first two digits will be considered hex, and the rest as literals, unless we change how the `b` sigil is read.
</blockquote></details>

"invariant that the type of a datum should be implied by its formatting" - this is actually not true, e.g. for timestamps, decimal NaN etc. Check out `f.disambiguateDatumTypes`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

justinj:
Review status: 1 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/sql/parser/eval_test.go, line 609 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4CdhMqK-pl3GNUU58-r3-609:-KvYwReC0GaqDpK5myzp:bjg8um3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b32324e0305d6fc0154abee66825bbf716668bb4/pkg/sql/parser/eval_test.go#L609)):*

"invariant that the type of a datum should be implied by its formatting" - this is actually not true, e.g. for timestamps, decimal NaN etc. Check out `f.disambiguateDatumTypes`.
</blockquote></details>

OH! This comment cleared up so much!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

justinj:
Review status: 1 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/parser/datum.go, line 209 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4CdhMqK-pl3GNUU54-r3-209:-KvZ3AfrYOpX8L4fljcP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b32324e0305d6fc0154abee66825bbf716668bb4/pkg/sql/parser/datum.go#L209)):*

allowBackslashXFormat should be documented in the comment, even though it sounds obvious. It's going to be confusing for people who need to call `ParseDByte` otherwise.
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval_test.go, line 609 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kv4CdhMqK-pl3GNUU58-r3-609:-KvZ3COsXT6i8jsWyehK:bkbv7o7) ([raw file](https://github.com/cockroachdb/cockroach/blob/b32324e0305d6fc0154abee66825bbf716668bb4/pkg/sql/parser/eval_test.go#L609)):*

OH! This comment cleared up so much!
</blockquote></details>

Ok, I think the issues here are resolved.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843)*

knz:
:lgtm: but beware of the docs

---

Reviewed 4 of 10 files at r3, 7 of 7 files at r4.

---

*[pkg/sql/parser/builtins.go, line 682 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-Kw4mMJOX9Q8UK_aOzS7:-Kw4mMJOX9Q8UK_aOzS8:bbuz9qv) ([raw file](https://github.com/cockroachdb/cockroach/blob/3588bffc9aa6bd1fed7746911f09168b4d864cc8/pkg/sql/parser/builtins.go#L682)):*
> ```Go
> 				return NewDString(data), nil
> 			},
> 			Info: "Encodes `data` in the text format specified by `format` (only \"hex\" is supported).",
> ```

You need to regenerate the builtin function docs.
`make generate PKG=./docs/...`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18843#-:-Kw4mcm1PVY2VKcp4FZ4:b-7rubcv)*

PR #18844, opened 2017-09-27
sql: table lease refresh done asynchronously 

lgo:
This commit completes changes the lease refreshing flow from a periodic
renewal to renewing the lease asynchronously on table access if the
lease will expire soon (the first part was reverting the periodic
    renewal in ec9f73b).

Before, table lease acquisition was done through a routine which kept
renewing a lease indefinitely. This meant the routine to renew the lease
would not end after a table was truncated or otherwise discard.

This will continue to prevent consecutive accesses to the table lease
from blocking due to lease expiration as the previous flow did.

When we asynchronously attempt to acquire a lease we create a goroutine
which only lives for the length of the attempt. If it fails, the same
process will repeat until the existing lease expires, when the error
will become user-facing.

cc: @vivekmenezes @andreimatei

### DIFF
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -41,9 +41,6 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 )
 
-// TODO(pmattis): Periodically renew leases for tables that were used recently and
-// for which the lease will expire soon.
-
 var (
 	// LeaseDuration is the mean duration a lease will be acquired for. The
 	// actual duration is jittered using LeaseJitterMultiplier.
@@ -55,6 +52,9 @@ var (
 	// [(1-LeaseJitterMultiplier) * LeaseDuration, (1+LeaseJitterMultiplier) * LeaseDuration]
 	// Exported for testing purposes only.
 	LeaseJitterMultiplier = 0.25
+	// LeaseRenewalTimeout is the time until expiry when we want to acquire a new
+	// lease.
+	LeaseRenewalTimeout = 1 * time.Minute
 )
 
 // tableVersionState holds the state for a table version. This includes
@@ -96,6 +96,15 @@ func (s *tableVersionState) String() string {
 // at the given timestamp
 func (s *tableVersionState) hasExpired(timestamp hlc.Timestamp) bool {
 	return !timestamp.Less(s.expiration)
+}
+
+// leaseHasEnoughLifeLeft checks whether the table descriptor lease has
+// enough life: that it expires further then LeaseRenewalTimeout
+// duration from now.
+func (s *tableVersionState) leaseHasEnoughLifeLeft() bool {
+	// Get the current time and move it forward by the lease renewal timeout.
+	t := hlc.Timestamp{WallTime: hlc.UnixNano() + int64(LeaseRenewalTimeout)}
+	return t.Less(s.expiration)
 }
 
 func (s *tableVersionState) incRefcount() {
@@ -569,6 +578,9 @@ type tableState struct {
 		// If set, leases are released from the store as soon as their
 		// refcount drops to 0, as opposed to waiting until they expire.
 		dropped bool
+		// Indicates whether any routines have blocked during lease acquisition.
+		// Used for testing purposes only.
+		hasBlockedDuringAcquisition bool
 	}
 }
 
@@ -581,26 +593,35 @@ func (t *tableState) acquire(
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
+	// When a lease is about to expire but has not yet (determined by
+	// leaseHasEnoughLifeLeft), then ensure an asynchronous routine to renew the
+	// lease is started.
+	if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) && !s.leaseHasEnoughLifeLeft() && t.mu.acquiring == nil {
+		_, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	// Acquire a lease if no lease exists or if the latest lease is
 	// about to expire. Repeatedly check to ensure either a lease is
 	// acquired or we receive an error. This is required for when another
 	// routine acquires a lease and releases it immediately.
 	for s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp); s = t.mu.active.findNewest() {
-		// Wait for any existing lease acquisition.
-		t.acquireWait()
-		// If the rcently acquired lease is valid for use, short-circuit.
-		if s = t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) {
-			break
-		}
-		errChan, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
-		if err != nil {
-			return nil, err
-		}
-		t.acquireWait()
-		if err = <-errChan; err != nil {
-			return nil, err
+		if t.mu.acquiring == nil {
+			errChan, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+			if err != nil {
+				return nil, err
+			}
+			t.acquireWait()
+			if err = <-errChan; err != nil {
+				return nil, err
+			}
+		} else if t.mu.acquiring != nil {
+			t.acquireWait()
 		}
 	}
+
 	return t.findForTimestamp(ctx, timestamp, m)
 }
 
@@ -784,6 +805,9 @@ func (t *tableState) removeInactiveVersions(m *LeaseManager) {
 func (t *tableState) acquireWait() {
 	// Spin until no lease acquisition is in progress.
 	for acquiring := t.mu.acquiring; acquiring != nil; acquiring = t.mu.acquiring {
+		// Indicates that a routine has blocked at all on another routine acquiring.
+		// Used for testing.
+		t.mu.hasBlockedDuringAcquisition = true
 		// We're called with mu locked, but need to unlock it while we wait
 		// for the in-progress lease acquisition to finish.
 		t.mu.Unlock()
@@ -816,7 +840,10 @@ func (t *tableState) acquireNodeLease(
 
 	// Notify when lease is being acquired.
 	t.mu.acquiring = make(chan struct{})
-	errChan := make(chan error)
+	// Notify the caller when we get an asynchronous acquisition error.
+	// Make it buffered so the task doesn't block if the caller ignores
+	// the channel.
+	errChan := make(chan error, 1)
 
 	err := t.stopper.RunAsyncTask(
 		ctx, "sql.tableState: async table lease acquisition", func(ctx context.Context) {
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -46,9 +46,15 @@ import (
 
 var (
 	// LeaseDuration is the mean duration a lease will be acquired for. The
-	// actual duration is jittered in the range
-	// [0.75,1.25]*LeaseDuration. Exported for testing purposes only.
+	// actual duration is jittered using LeaseJitterMultiplier.
+	// Exported for testing purposes only.
 	LeaseDuration = 5 * time.Minute
+	// LeaseJitterMultiplier is the factor that we use to randomly jitter the lease
+	// duration when acquiring a new lease and the lease renewal timeout. The
+	// range of the actual lease duration will be
+	// [(1-LeaseJitterMultiplier) * LeaseDuration, (1+LeaseJitterMultiplier) * LeaseDuration]
+	// Exported for testing purposes only.
+	LeaseJitterMultiplier = 0.25
 )
 
 // tableVersionState holds the state for a table version. This includes
@@ -123,9 +129,9 @@ type LeaseStore struct {
 }
 
 // jitteredLeaseDuration returns a randomly jittered duration from the interval
-// [0.75 * leaseDuration, 1.25 * leaseDuration].
+// [(1-LeaseJitterMultiplier) * leaseDuration, (1+LeaseJitterMultiplier) * leaseDuration].
 func jitteredLeaseDuration() time.Duration {
-	return time.Duration(float64(LeaseDuration) * (0.75 + 0.5*rand.Float64()))
+	return time.Duration(float64(LeaseDuration) * (1 - LeaseJitterMultiplier + 2*LeaseJitterMultiplier*rand.Float64()))
 }
 
 // acquire a lease on the most recent version of a table descriptor.
@@ -574,17 +580,27 @@ func (t *tableState) acquire(
 ) (*tableVersionState, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
-	// Wait for any existing lease acquisition.
-	t.acquireWait()
 
 	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
-	if s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp) {
-		if err := t.acquireNodeLease(ctx, m, hlc.Timestamp{}); err != nil {
+	// about to expire. Repeatedly check to ensure either a lease is
+	// acquired or we receive an error. This is required for when another
+	// routine acquires a lease and releases it immediately.
+	for s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp); s = t.mu.active.findNewest() {
+		// Wait for any existing lease acquisition.
+		t.acquireWait()
+		// If the rcently acquired lease is valid for use, short-circuit.
+		if s = t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) {
+			break
+		}
+		errChan, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+		if err != nil {
+			return nil, err
+		}
+		t.acquireWait()
+		if err = <-errChan; err != nil {
 			return nil, err
 		}
 	}
-
 	return t.findForTimestamp(ctx, timestamp, m)
 }
 
@@ -710,7 +726,14 @@ func (t *tableState) acquireFreshestFromStoreLocked(ctx context.Context, m *Leas
 		minExpirationTime = newestTable.expiration.Add(int64(time.Millisecond), 0)
 	}
 
-	return t.acquireNodeLease(ctx, m, minExpirationTime)
+	// Begin acquiring a new lease and wait for it to finish
+	errChan, err := t.acquireNodeLease(ctx, m, minExpirationTime)
+	if err != nil {
+		return err
+	}
+	t.acquireWait()
+	err = <-errChan
+	return err
 }
 
 // upsertLocked inserts a lease for a particular table version.
@@ -777,23 +800,56 @@ func (t *tableState) acquireWait() {
 // inserted at the end of the lease set (i.e. it will be returned by
 // findNewest() from now on).
 //
+// acquireNodeLease begins an asynchronous process to acquire a lease.
+// Use errChan to receive any errors from the asynchronous process. If
+// error is not nil, expect the error channel to be nil.
+//
+// Use acquireWait to block until the lease is acquired.
+//
 // t.mu needs to be locked.
 func (t *tableState) acquireNodeLease(
 	ctx context.Context, m *LeaseManager, minExpirationTime hlc.Timestamp,
-) error {
+) (chan error, error) {
 	if m.isDraining() {
-		return errors.New("cannot acquire lease when draining")
+		return nil, errors.New("cannot acquire lease when draining")
 	}
 
-	// Notify when lease has been acquired.
+	// Notify when lease is being acquired.
 	t.mu.acquiring = make(chan struct{})
+	errChan := make(chan error)
+
+	err := t.stopper.RunAsyncTask(
+		ctx, "sql.tableState: async table lease acquisition", func(ctx context.Context) {
+			err := t.acquireNodeLeaseBlocking(ctx, m, minExpirationTime)
+			if err != nil {
+				log.Error(ctx, errors.Wrapf(err, "Async TableDescriptor lease acquisition error"))
+				errChan <- err
+			}
+			close(errChan)
+		})
+
+	if err != nil {
+		close(t.mu.acquiring)
+		close(errChan)
+		t.mu.acquiring = nil
+	}
+
+	return errChan, err
+}
+
+// acquireNodeLeaseBlocking does the blocking operation for acquiring a
+// table descriptor lease.
+//
+// t.mu.acquiring must be initialized before calling.
+func (t *tableState) acquireNodeLeaseBlocking(
+	ctx context.Context, m *LeaseManager, minExpirationTime hlc.Timestamp,
+) error {
+	defer t.mu.Unlock()
 	defer func() {
 		close(t.mu.acquiring)
 		t.mu.acquiring = nil
 	}()
-	// We're called with mu locked, but need to unlock it during lease
-	// acquisition.
-	t.mu.Unlock()
+
 	table, err := m.LeaseStore.acquire(ctx, t.id, minExpirationTime)
 	t.mu.Lock()
 	if err != nil {
@@ -801,6 +857,7 @@ func (t *tableState) acquireNodeLease(
 	}
 	t.upsertLocked(ctx, table, m)
 	t.tableNameCache.insert(table)
+
 	return nil
 }
 
--- a/pkg/sql/lease_internal_test.go
+++ b/pkg/sql/lease_internal_test.go
@@ -19,16 +19,20 @@ package sql
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/config"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
+	"github.com/pkg/errors"
 )
 
 func TestTableSet(t *testing.T) {
@@ -557,4 +561,177 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
 	}
 
 	wg.Wait()
+}
+
+// This test makes sure the lease gets renewed automatically in the background
+// if the lease is about to expire.
+func TestLeaseRefreshedAutomatically(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+	// We set LeaseRenewalTimeout to be very close to the duration, so we
+	// renew on access almost immediately. LeaseJitterMultiplier is set to
+	// ensure any newer leases will have a higher expiration timestamp.
+	// We must do this before leaktest.AfterTest to make sure no routines
+	// race when accessing LeaseJitterMultiplier.
+	savedLeaseRenewalTimeout := LeaseRenewalTimeout
+	savedLeaseJitterMultiplier := LeaseJitterMultiplier
+	defer func() {
+		LeaseRenewalTimeout = savedLeaseRenewalTimeout
+		LeaseJitterMultiplier = savedLeaseJitterMultiplier
+	}()
+	LeaseRenewalTimeout = LeaseDuration - 1*time.Nanosecond
+	LeaseJitterMultiplier = 0
+
+	var testAcquiredCount int32
+	var testAsyncAcquiredCount int32
+
+	testingKnobs := base.TestingKnobs{
+		SQLLeaseManager: &LeaseManagerTestingKnobs{
+			LeaseStoreTestingKnobs: LeaseStoreTestingKnobs{
+				// We want to track when leases get acquired and when they are renewed.
+				LeaseAcquiredEvent: func(table sqlbase.TableDescriptor, err error) {
+					atomic.AddInt32(&testAcquiredCount, 1)
+				},
+			},
+		},
+	}
+	s, sqlDB, kvDB := serverutils.StartServer(t, base.TestServerArgs{Knobs: testingKnobs})
+	defer s.Stopper().Stop(context.TODO())
+	leaseManager := s.LeaseManager().(*LeaseManager)
+
+	if _, err := sqlDB.Exec(`
+CREATE DATABASE t;
+CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
+`); err != nil {
+		t.Fatal(err)
+	}
+	tableDesc := sqlbase.GetTableDescriptor(kvDB, "t", "test")
+	descID := tableDesc.ID
+
+	// Acquire the first lease.
+	ts, e1, err := leaseManager.Acquire(context.TODO(), s.Clock().Now(), descID)
+	if err != nil {
+		t.Fatal(err)
+	} else if err := leaseManager.Release(ts); err != nil {
+		t.Fatal(err)
+	}
+
+	// Reset hasBlockedDuringAcquisition as the first acqusition will always block.
+	leaseManager.mu.tables[descID].mu.hasBlockedDuringAcquisition = false
+
+	if count := atomic.LoadInt32(&testAcquiredCount); count != 1 {
+		t.Fatalf("expected 1 lease to be acquired, but acquired %d times",
+			count)
+	} else if count := atomic.LoadInt32(&testAsyncAcquiredCount); count != 0 {
+		t.Fatalf("expected no leases to be acquired asynchronously, but got %d asynchronously",
+			count)
+	}
+
+	testutils.SucceedsSoon(t, func() error {
+		// Acquire another lease. At first this will be the same lease, but
+		// eventually another elase will be acquired asynchronously and we
+		// will get that one.
+		_, e2, err := leaseManager.Acquire(context.TODO(), s.Clock().Now(), descID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		defer func() {
+			if err := leaseManager.Release(ts); err != nil {
+				t.Fatal(err)
+			}
+		}()
+
+		// We check for the new expiry time because if our past acquire triggered
+		// the background refresh, the next lease we get will be the result of the
+		// background refresh.
+		if e2.WallTime <= e1.WallTime {
+			return errors.Errorf("expected new lease expiration (%s) to be after old lease expiration (%s)",
+				e2, e1)
+		} else if count := atomic.LoadInt32(&testAcquiredCount); count < 2 {
+			return errors.Errorf("expected at least 2 leases to be acquired, but acquired %d times",
+				count)
+		} else if hasBlockedDuringAcquisition := leaseManager.mu.tables[descID].mu.hasBlockedDuringAcquisition; hasBlockedDuringAcquisition {
+			t.Fatalf("expected repeated lease acquisition to not block, but hasBlockedDuringAcquisition is true")
+		}
+		return nil
+	})
+}
+
+// This test makes sure that async lease refreshing doesn't block any routines
+func TestLeaseRefreshDoesntBlock(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+
+	// We set LeaseRenewalTimeout to be very close to the duration, so we
+	// renew on access almost immediately. We also set the
+	// LeaseJitterMultiplier to ensure renewal timeouts happen immediately
+	// (and not jittered).
+	savedLeaseRenewalTimeout := LeaseRenewalTimeout
+	savedLeaseJitterMultiplier := LeaseJitterMultiplier
+	defer func() {
+		LeaseRenewalTimeout = savedLeaseRenewalTimeout
+		LeaseJitterMultiplier = savedLeaseJitterMultiplier
+	}()
+	LeaseRenewalTimeout = LeaseDuration - 1*time.Millisecond
+	LeaseJitterMultiplier = 0
+
+	var testAcquiredCount int32
+
+	testingKnobs := base.TestingKnobs{
+		SQLLeaseManager: &LeaseManagerTestingKnobs{
+			LeaseStoreTestingKnobs: LeaseStoreTestingKnobs{
+				// We want to track when leases get acquired and when they are renewed.
+				LeaseAcquiredEvent: func(table sqlbase.TableDescriptor, err error) {
+					atomic.AddInt32(&testAcquiredCount, 1)
+				},
+			},
+		},
+	}
+	s, sqlDB, kvDB := serverutils.StartServer(t, base.TestServerArgs{Knobs: testingKnobs})
+	defer s.Stopper().Stop(context.TODO())
+	leaseManager := s.LeaseManager().(*LeaseManager)
+
+	if _, err := sqlDB.Exec(`
+CREATE DATABASE t;
+CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
+`); err != nil {
+		t.Fatal(err)
+	}
+	tableDesc := sqlbase.GetTableDescriptor(kvDB, "t", "test")
+	descID := tableDesc.ID
+
+	// Acquire the first lease.
+	ts, _, err := leaseManager.Acquire(context.TODO(), s.Clock().Now(), descID)
+	if err != nil {
+		t.Fatal(err)
+	} else if err := leaseManager.Release(ts); err != nil {
+		t.Fatal(err)
+	}
+
+	// Reset hasBlockedDuringAcquisition as the first acqusition will always block.
+	leaseManager.mu.tables[descID].mu.hasBlockedDuringAcquisition = false
+
+	var wg sync.WaitGroup
+	numRoutines := 40
+	wg.Add(numRoutines)
+	for i := 0; i < numRoutines; i++ {
+		time.Sleep(1 * time.Millisecond)
+		go func() {
+			defer wg.Done()
+			ts, _, err := leaseManager.Acquire(context.TODO(), s.Clock().Now(), descID)
+			if err != nil {
+				t.Error(err)
+				return
+			} else if err := leaseManager.Release(ts); err != nil {
+				t.Error(err)
+				return
+			}
+		}()
+	}
+	wg.Wait()
+
+	if hasBlockedDuringAcquisition := leaseManager.mu.tables[descID].mu.hasBlockedDuringAcquisition; hasBlockedDuringAcquisition {
+		t.Fatalf("expected repeated lease acquisition to not block, but hasBlockedDuringAcquisition is true")
+	} else if acquiredCount := atomic.LoadInt32(&testAcquiredCount); acquiredCount < 2 {
+		t.Fatalf("expected to acquire at least 2 leases. acquired %d times", acquiredCount)
+	}
 }
--- a/pkg/sql/lease_test.go
+++ b/pkg/sql/lease_test.go
@@ -223,8 +223,10 @@ func TestLeaseManager(testingT *testing.T) {
 	// table and expiration.
 	l1, e1 := t.mustAcquire(1, descID)
 	l2, e2 := t.mustAcquire(1, descID)
-	if l1.ID != l2.ID || e1 != e2 {
+	if l1.ID != l2.ID {
 		t.Fatalf("expected same lease, but found %v != %v", l1, l2)
+	} else if e1 != e2 {
+		t.Fatalf("expected same lease timestamps, but found %v != %v", e1, e2)
 	}
 	t.expectLeases(descID, "/1/1")
 	// Node 2 never acquired a lease on descID, so we should expect an error.


lego:
@andreimatei you mentioned rolling back the other patch from the release. Should I revert the merge into 1.1 and when/if cherrypicking this one, then undo the revert and this?

(although a handful of this code also just reverts the past commit, should I instead revert the commit and put this changes as a separate commit?)

andreimatei:
What I would do is revert the previous commits (on both master and release-1.1) and then go forward with this patch from a clean slate.

---

Review status: 0 of 3 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 101 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4EeGWAXF2MUPbXQYD:-Kv4EeGWAXF2MUPbXQYE:b-ff744z) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L101)):*
> ```Go
> }
> 
> // willExpired checks if the table is within the renewal deadline for the
> ```

s/willExpired/willExpire

I'd rename this to something like `leaseHasEnoughLifeLeft()` (I think we once had a function with a similar name.

Also, the comment says "renewal deadline for the transaction"; I don't know what that means, and also it's unclear which transaction you're talking about. This method does not take in a transaction. See below.

Finally, I wouldn't make this method take in a timestamp. I think that's unnecessary noise. In relation with a transaction, a lease (more generally, a descriptor) is valid or invalid, period. Whether it expires in a nanosecond or a millennium doesn't matter (well, it matters a tiny bit for Snapshot transactions, as the life left in leases can dictate how much the txn can be pushed - but I don't think that's particularly important and also since the refresh is async, you're not helping the current txn in any way).
So, the question that this method has to answer, in my opinion, is unrelated to the specifics of any particular transactions. The question is "given that there's evidence for clients' continued interest in this table, should I proactively renew the lease?". So, we should look at the current clock value (there's got to be a clock around here somewhere).

---

*[pkg/sql/lease.go, line 104 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4HEVdPAOjZWrQGtdc:-Kv4HEVdPAOjZWrQGtdd:b-zij0w8) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L104)):*
> ```Go
> // transaction.
> func (s *tableVersionState) willExpire(timestamp hlc.Timestamp) bool {
> 	return !timestamp.Add(int64(LeaseRenewalTimeout), 0).Less(s.expiration)
> ```

this method only makes sense for leases, so make that clear in name/comments and assert in code that the `tableVersionState` refers to a lease (and not to another type of descriptor). See `tableVersionState.leased`.

---

*[pkg/sql/lease.go, line 579 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4Hhqqkv3v0MrOL41N:-Kv4Hhqqkv3v0MrOL41O:brpy5bv) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L579)):*
> ```Go
> 		dropped bool
> 		// A flag to denote that we have already started a renewal.
> 		renewalInProgress bool
> ```

how's this different from `acquiring != nil` ?

---

*[pkg/sql/lease.go, line 593 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4SpYzpbcKJlsO8glR:-Kv4SpYzpbcKJlsO8glS:b-anvvie) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L593)):*
> ```Go
> 
> 	// Block and wait for the new lease to be acquired on another routine.
> 	t.acquireWait()
> ```

can you please take the opportunity to add a TODO with my name here to not wait indiscriminately and link to #18851

---

*[pkg/sql/lease.go, line 613 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4JuSwjOYP3e98hHTH:-Kv4JuSwjOYP3e98hHTI:bhta06a) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L613)):*
> ```Go
> // acquireAlways will acquire a new lease. This is used for acquiring a new
> // lease asynchronously before our previous one has expired.
> func (t *tableState) acquireAlways(m *LeaseManager) {
> ```

I don't understand the name of this method. "always" as opposed to "if you'd like"? :)
I'd don't know what a better name would be because I don't think the method deserves to exist in the first place. If it does need to exist, it seems very similar to `acquireFreshestFromStoreLocked()`.

I suggest you inline it at the call site. If you take the following suggestion and still need a name, something like `acquireIfNotInProgress` would perhaps be better.

---

*[pkg/sql/lease.go, line 623 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4JFicX9qmIzfmSgzc:-Kv4JFidiKwrI4NHI58a:bpflv22) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L623)):*
> ```Go
> 	// Block and wait for the new lease to be acquired on another routine to
> 	// prevent a race condition of entering acquireNodeLease
> 	t.acquireWait()
> ```

if a lease acquisition is in progress, then I think this method should consider that to be good enough and short-circuit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-:-Kv4EI-Q5sOVD59LYH5V:bkp0d0g)*

vivekmenezes:
The easiest way to do that is to revert it in this PR and make your change on top of it. That way this PR is documenting the revert and the new change.

I like the general theme in this PR and we should move forward with it. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-:-Kv5X5VArw6erAnBvsa3:b4yell)*

vivekmenezes:
One thing that needs to get fixed is routines waiting on a thread that is acquiring a lease. We thought that the periodic acquisition of a lease would prevent other routines from waiting on a lease acquisition. Turns out that is not true! All routines wait on acquireWait().

So we need to make it such that a routine waits on acquireWait() only if it attempts to actually acquire a lease. Furthermore when it comes out of the loop it needs to recheck that a new lease has been refreshed to see whether it still needs to acquire a new lease.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-:-Kv79aOiqWKIf4r2I_P9:b2u6i6w)*

lego:
Yep, I realized I missed that bit when I first crafted this patch. `acquire` now reflects that.

---

Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[pkg/sql/lease.go, line 101 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4EeGWAXF2MUPbXQYD:-Kv7eBpH0EefRoNgHSd5:bcggfdo) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L101)):*

s/willExpired/willExpire

I'd rename this to something like `leaseHasEnoughLifeLeft()` (I think we once had a function with a similar name.

Also, the comment says "renewal deadline for the transaction"; I don't know what that means, and also it's unclear which transaction you're talking about. This method does not take in a transaction. See below.

Finally, I wouldn't make this method take in a timestamp. I think that's unnecessary noise. In relation with a transaction, a lease (more generally, a descriptor) is valid or invalid, period. Whether it expires in a nanosecond or a millennium doesn't matter (well, it matters a tiny bit for Snapshot transactions, as the life left in leases can dictate how much the txn can be pushed - but I don't think that's particularly important and also since the refresh is async, you're not helping the current txn in any way).
So, the question that this method has to answer, in my opinion, is unrelated to the specifics of any particular transactions. The question is "given that there's evidence for clients' continued interest in this table, should I proactively renew the lease?". So, we should look at the current clock value (there's got to be a clock around here somewhere).
</blockquote></details>

Yep, makes sense regarding the timestamp. Done, I've renamed the function, I think the comment is more clear now and removed the timestamp part.

---

*[pkg/sql/lease.go, line 104 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4HEVdPAOjZWrQGtdc:-Kv5auen98j0Th0MUGCi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L104)):*

this method only makes sense for leases, so make that clear in name/comments and assert in code that the `tableVersionState` refers to a lease (and not to another type of descriptor). See `tableVersionState.leased`.
</blockquote></details>

Done.

---

*[pkg/sql/lease.go, line 579 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4Hhqqkv3v0MrOL41N:-Kv4TpGuHW1Q8TkxtKrY:b9dq60m) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L579)):*

how's this different from `acquiring != nil` ?
</blockquote></details>

When creating a new goroutine `acquireAlways` to start the acquisition this flag is used to prevent another goroutine from beginning the asynchronous acquisition as well. This is where the routine has begun the asynchronous acquisition but the channel `acquiring` isn't initialized yet in `acquireNodeLease`.

The naming is bad, so I've renamed it as well as more details to the comment.

---

*[pkg/sql/lease.go, line 593 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4SpYzpbcKJlsO8glR:-Kv4Uj3MG_jbLGzc80ee:b-rr97ow) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L593)):*

can you please take the opportunity to add a TODO with my name here to not wait indiscriminately and link to #18851
</blockquote></details>

Can do. Additionally, I just realized there are a few mistakes here. I've shifted things around with comments that should detail it, but it appears the only use of `acquireWait` does not concern itself with the table version (unless I'm misunderstanding the usage in `acquireFreshestFromStoreLocked`).

---

*[pkg/sql/lease.go, line 613 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4JuSwjOYP3e98hHTH:-Kv5jcScn5zoFEJMd0yH:b4gjrj6) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L613)):*

I don't understand the name of this method. "always" as opposed to "if you'd like"? :)
I'd don't know what a better name would be because I don't think the method deserves to exist in the first place. If it does need to exist, it seems very similar to `acquireFreshestFromStoreLocked()`.

I suggest you inline it at the call site. If you take the following suggestion and still need a name, something like `acquireIfNotInProgress` would perhaps be better.
</blockquote></details>

I've renamed it. I don't think it should be inlined just because it's doing a non-trivial amount of work.

---

*[pkg/sql/lease.go, line 623 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4JFicX9qmIzfmSgzc:-Kv4UVH6Uy94K2ghLhVl:b1a65f6) ([raw file](https://github.com/cockroachdb/cockroach/blob/610e6f23f21435ac7d209e4837df2a13b1b7e80b/pkg/sql/lease.go#L623)):*

if a lease acquisition is in progress, then I think this method should consider that to be good enough and short-circuit.
</blockquote></details>

Yep, you're right on that, thanks!

---

*[pkg/sql/lease.go, line 602 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv7erATZ4OGBmSiql6i:-Kv7erAUFmG6K1VOqfUq:b-1tqcwk) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff3bdf313b76856fefde901330c52626b006a047/pkg/sql/lease.go#L602)):*
> ```Go
> 	// leaseHasEnoughLifeLeft, then ensure an asynchronous routine to renew the
> 	// lease is created.
> 	if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) && !t.mu.asyncRenewalInProgress && s.leaseHasEnoughLifeLeft() {
> ```

This condition is a bit complicated but I wasn't sure how to simplify this anymore, but it just does as the comment says.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-:-Kv7eWjAmjWLdLvsoPpX:b-v8o9oj)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease_internal_test.go, line 603 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv8YumVyieusYtCIr76:-Kv8YumVyieusYtCIr77:bbo0ayy) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease_internal_test.go#L603)):*
> ```Go
> 	savedLeaseRenewalTimeout := LeaseRenewalTimeout
> 	defer func() {
> 		LeaseRenewalTimeout = savedLeaseRenewalTimeout
> ```

Finally got the tests to a working state, but there's a randomness problem with the jitter of the lease expiration which is what we check that we have newer leases.

I had to take out changing `LeaseJitterMultiplier` which helps avoid that one problem because recovering the jitter multiplier actually caused race conditions on its own during tests, so this test failed ~1/10 times. `LeaseJitterMultiplier` is used within the `db.Txn` part of lease acquisition, when we've released locks so there wasn't an easy way to get around that.

If the test happens to acquire a lease with a 6.25 minute lease (maximum jitter), the chances the test acquires one with an even later expiry is low.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844)*

lego:
Sorry about the email churn on pushes, I accidentally pushed an older version when I moved to the work laptop.

PLTAL. I've added a test that checks both that no routines are blocking and leases are coming from the async routine and not synchronously.

---

Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-:-Kv8Zd-mY7kXenuHxMmU:bfn4pqa)*

pkg/sql/lease.go
```diff
@@ -41,14 +41,20 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 )
 
-// TODO(pmattis): Periodically renew leases for tables that were used recently and
-// for which the lease will expire soon.
-
```
vivekmenezes: you can rename the commit

sql: table lease refresh done asynchronously

pkg/sql/lease.go
```diff
@@ -41,14 +41,20 @@ import (
...
 	// LeaseDuration is the mean duration a lease will be acquired for. The
-	// actual duration is jittered in the range
-	// [0.75,1.25]*LeaseDuration. Exported for testing purposes only.
+	// actual duration is jittered using LeaseJitterMultiplier.
+	// Exported for testing purposes only.
 	LeaseDuration = 5 * time.Minute
+	// LeaseJitterMultiplier is the factor that we use to randomly jitter the lease
+	// duration when acquiring a new lease and the lease renewal timeout. The
+	// range of the actual lease duration will be
+	// [(1-LeaseJitterMultiplier) * LeaseDuration, (1-LeaseJitterMultiplier) * LeaseDuration]
```
vivekmenezes: // [(1-LeaseJitterMultiplier) * LeaseDuration, (1+LeaseJitterMultiplier) * LeaseDuration]

pkg/sql/lease.go
```diff
@@ -92,6 +98,16 @@ func (s *tableVersionState) hasExpired(timestamp hlc.Timestamp) bool {
 	return !timestamp.Less(s.expiration)
 }
 
+// leaseHasEnoughLifeLeft checks whether the table descriptor lease has
+// enough life: that it expires further then LeaseRenewalTimeout
+// duration from now.
+func (s *tableVersionState) leaseHasEnoughLifeLeft() bool {
+	// Get the current time and move it forward by the lease renewal timeout.
+	t := hlc.Timestamp{WallTime: hlc.UnixNano()}
```
vivekmenezes: t := s.Clock().Now().Add(LeaseRenewalTimeout)

pkg/sql/lease.go
```diff
@@ -563,6 +579,15 @@ type tableState struct {
 		// If set, leases are released from the store as soon as their
 		// refcount drops to 0, as opposed to waiting until they expire.
 		dropped bool
+		// asyncRenewalInProgress indicates an asynchronous routine to renew the
+		// table descriptor lease has begun. The flag is required because when
+		// acquisition is done asynchronously, we can have a race condition where
+		// another routine attempts to also renew the lease before the async routine
+		// acquires t.mu.
+		asyncRenewalInProgress bool
```
vivekmenezes: i think you can eliminate the need for this and use acquiring != nil as long as you restructure acquireNodeLease() to take in an asynchronous boolean which asks it to acquire the lease asynchronously and then structure it such that the channel is closed by the function that does the lease acquisition

pkg/sql/lease.go
```diff
@@ -574,20 +599,65 @@ func (t *tableState) acquire(
...
 	defer t.mu.Unlock()
-	// Wait for any existing lease acquisition.
-	t.acquireWait()
 
-	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
+	// When a lease is about to expire but has not yet (determined by
+	// leaseHasEnoughLifeLeft), then ensure an asynchronous routine to renew the
+	// lease is started.
+	if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) && !t.mu.asyncRenewalInProgress && !s.leaseHasEnoughLifeLeft() {
```
vivekmenezes: if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) &&  !s.leaseHasEnoughLifeLeft() && acquiring != nil {
}

pkg/sql/lease.go
```diff
@@ -574,20 +599,65 @@ func (t *tableState) acquire(
...
-		if err := t.acquireNodeLease(ctx, m, hlc.Timestamp{}); err != nil {
-			return nil, err
+		// Block and wait for the new lease to be acquired on another routine.
+		t.acquireWait()
+
+		// Check again, and acquire synchronously if the lease was already released.
+		if s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp) {
+			if err := t.acquireNodeLease(ctx, m, hlc.Timestamp{}); err != nil {
+				return nil, err
+			}
```
vivekmenezes: great!

pkg/sql/lease.go
```diff
@@ -574,20 +599,65 @@ func (t *tableState) acquire(
...
-	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
+	// When a lease is about to expire but has not yet (determined by
+	// leaseHasEnoughLifeLeft), then ensure an asynchronous routine to renew the
+	// lease is started.
+	if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) && !t.mu.asyncRenewalInProgress && !s.leaseHasEnoughLifeLeft() {
+		// Set a flag denoting async acquisition has begun to prevent this from
+		// happening multiple times.
+		t.mu.asyncRenewalInProgress = true
+		go t.acquireIfRenewalNotInProgress(m)
```
vivekmenezes: use t.stopper.RunAsyncTask() to prevent goroutines leaks in tests,  but as I said above you want to call this from within acquireNodeLease()

pkg/sql/lease.go
```diff
@@ -761,6 +831,10 @@ func (t *tableState) removeInactiveVersions(m *LeaseManager) {
 func (t *tableState) acquireWait() {
 	// Spin until no lease acquisition is in progress.
 	for acquiring := t.mu.acquiring; acquiring != nil; acquiring = t.mu.acquiring {
+		// Indicates that a routine has blocked at all on another routine acquiring.
+		// Used for testing.
+		t.mu.hasBlockedDuringAcquisition = true
+
```
vivekmenezes: You probably need a test that runs a ton of acquisitions and ensures that they don't get blocked.

vivekmenezes (commented):
Nice! Sorry that you had to redo this change. Sometimes we just have to iterate. Thanks

lego:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions.

---

*[pkg/sql/lease.go, line 52 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r8-52:-Kv8wkYWX9EB6datQwGZ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease.go#L52)):*

// [(1-LeaseJitterMultiplier) * LeaseDuration, (1+LeaseJitterMultiplier) * LeaseDuration]
</blockquote></details>

Done.

---

*[pkg/sql/lease.go, line 106 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r8-106:-Kv8w_vGkr9pV7-EHrx3:bpb6h5r) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease.go#L106)):*

t := s.Clock().Now().Add(LeaseRenewalTimeout)
</blockquote></details>

There's no clock function on tableVersionState (or tableVersion for that matter). 

Could this impact tests? Right now I'm getting weird behavior in the new test with lots of routines acquiring where it doesn't seem to cause it to renew frequently (not sure if it's just the test, or the code itself, I haven't seen any issue here). 

---

*[pkg/sql/lease.go, line 587 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r8-587:-Kv9JA2gjX4rhYsSBj7u:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease.go#L587)):*

i think you can eliminate the need for this and use acquiring != nil as long as you restructure acquireNodeLease() to take in an asynchronous boolean which asks it to acquire the lease asynchronously and then structure it such that the channel is closed by the function that does the lease acquisition
</blockquote></details>

Done.

---

*[pkg/sql/lease.go, line 606 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r8-606:-Kv9JCbSJVl385zjByx-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease.go#L606)):*

if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) &&  !s.leaseHasEnoughLifeLeft() && acquiring != nil {
}

</blockquote></details>

Done.

---

*[pkg/sql/lease.go, line 610 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r8-610:-Kv9JDtAstNLgwOOhmqB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease.go#L610)):*

use t.stopper.RunAsyncTask() to prevent goroutines leaks in tests,  but as I said above you want to call this from within acquireNodeLease()
</blockquote></details>

Done.

---

*[pkg/sql/lease.go, line 837 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r8-837:-Kv9YuKMkQmNgYgpTjZ0:b-fxpe04) ([raw file](https://github.com/cockroachdb/cockroach/blob/44adf8e94d816a12b182d90f7c89ed55b5747c0c/pkg/sql/lease.go#L837)):*

You probably need a test that runs a ton of acquisitions and ensures that they don't get blocked.
</blockquote></details>

Done (but the test isn't passing right now, as mentioned in the other comment re: the timestamp generation).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844)*

vivekmenezes:
@lego a thought I had this morning about this PR. I'd like us to explore the opportunity to make 
acquireNodeLease *always* asynchronous. So in the places where it is expected to be synchronous we need to call acquireWait() to know when it's done.

If we were to make the above change, I'd suggest splitting up this PR into two changes. The first change will involve making acquireNodeLease() asynchronous and fitting that implementation into the current code to see how it affects all synchronous code sites. You can then add this change over that  to call it from the new site and not need to wait after it's called (along with the new unittests)

lego:
Done. I think the tests are good now, ensuring that the calls don't block and that it refreshes the leases in the background.

---

Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-:-Kv9Z0u07A51qNVz2ReR:bde5wde)*

pkg/sql/lease.go
```diff
@@ -785,22 +820,56 @@ func (t *tableState) acquireNodeLease(
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
 	defer func() {
 		close(t.mu.acquiring)
 		t.mu.acquiring = nil
 	}()
-	// We're called with mu locked, but need to unlock it during lease
-	// acquisition.
+
+	if m.isDraining() {
```
vivekmenezes: you've already checked this. no need to check it again

pkg/sql/lease.go
```diff
@@ -785,22 +820,56 @@ func (t *tableState) acquireNodeLease(
...
+		err := errors.New("cannot acquire lease when draining")
+		t.mu.latestAcquisitionErr = err
+		return err
+	}
+
 	t.mu.Unlock()
 	table, err := m.LeaseStore.acquire(ctx, t.id, minExpirationTime)
```
vivekmenezes: I'd make this the first line of this function. Get rid of everything above.

vivekmenezes (commented):
Great stuff!

I've only reviewed the first commit. Once we've converged on that we can start looking at the second commit. Thanks for doing this!

pkg/sql/lease.go
lego: I'm a bit skeptical of this because if the system starts draining by the time this routine begins.

Sidenote: Is draining what happens when cockroach is cleaning quitting? Can we possibly run into an error state if we don't stop here while draining is true?

pkg/sql/lease.go
```diff
@@ -563,6 +569,12 @@ type tableState struct {
 		// If set, leases are released from the store as soon as their
 		// refcount drops to 0, as opposed to waiting until they expire.
 		dropped bool
+		// latestAcquisitionErr is used to communicating errors received
+		// when acquiring a lease on the asynchronous task. It is only
+		// non-nil if there have been no errors or if there was an error and
+		// a routine blocking on acquisition has not picked up the error
+		// yet. It is also cleared when we begin another acquisition.
+		latestAcquisitionErr error
```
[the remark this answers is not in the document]
lego: Done. Great idea! This also alleviates something I was FUD about the second commit.

pkg/sql/lease.go
```diff
@@ -785,22 +820,56 @@ func (t *tableState) acquireNodeLease(
 		return errors.New("cannot acquire lease when draining")
 	}
 
-	// Notify when lease has been acquired.
+	// Notify when lease is being acquired.
 	t.mu.acquiring = make(chan struct{})
+	t.mu.latestAcquisitionErr = nil
+
+	err := t.stopper.RunAsyncTask(
+		context.Background(), "sql.tableState: async table lease acquisition", func(ctx context.Context) {
```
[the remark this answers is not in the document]
lego: Done.

Although in the past experiences of having a long-running worker take in the context I'm a bit skeptical. But since it's a more short-lived task I'll give it a try :P.

In the case of a request spawning, a background refresh, passing through the current request context I think is not correct because the task can outlive the request. (But I'm also not quite sure what a context is or is for).

pkg/sql/lease.go
lego: Done.

pkg/sql/lease.go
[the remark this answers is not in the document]
benesch: (Have you read https://github.com/cockroachdb/cockroach/blob/master/docs/tech-notes/contexts.md?)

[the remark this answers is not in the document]
lego: > You can go a long way without understanding exactly what the point is (until you can’t), but that’s no way to live.
The story of my life right now. Thanks @benesch!

pkg/sql/lease.go
lego: Done.

pkg/sql/lease.go
```diff
@@ -574,17 +586,29 @@ func (t *tableState) acquire(
 ) (*tableVersionState, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
+
 	// Wait for any existing lease acquisition.
 	t.acquireWait()
```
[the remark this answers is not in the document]
lego: Done 👍

pkg/sql/lease.go
```diff
@@ -574,17 +586,29 @@ func (t *tableState) acquire(
...
 	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
-	if s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp) {
-		if err := t.acquireNodeLease(ctx, m, hlc.Timestamp{}); err != nil {
-			return nil, err
+	// about to expire. Repeatedly check to ensure either a lease is
+	// acquired or we receive an error. This is required for when another
+	// routine acquires a lease and releases it immediately.
+	for {
+		if s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp) {
```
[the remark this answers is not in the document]
lego: Done 👍

pkg/sql/lease.go
```diff
@@ -574,17 +580,27 @@ func (t *tableState) acquire(
...
 	// Acquire a lease if no lease exists or if the latest lease is
-	// about to expire.
-	if s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp) {
-		if err := t.acquireNodeLease(ctx, m, hlc.Timestamp{}); err != nil {
+	// about to expire. Repeatedly check to ensure either a lease is
+	// acquired or we receive an error. This is required for when another
+	// routine acquires a lease and releases it immediately.
+	for s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp); s = t.mu.active.findNewest() {
+		// Wait for any existing lease acquisition.
+		t.acquireWait()
```
vivekmenezes: i think you can have acquireWait() return a boolean indicating if it waited. If it waited here you can just ```continue``` . That way you don't need to do the check below

pkg/sql/lease.go
```diff
@@ -574,17 +580,27 @@ func (t *tableState) acquire(
...
+		// If the rcently acquired lease is valid for use, short-circuit.
+		if s = t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) {
+			break
+		}
+		errChan, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+		if err != nil {
+			return nil, err
+		}
+		t.acquireWait()
```
vivekmenezes: do you really need to call acquireWait() again?

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
+// Use errChan to receive any errors from the asynchronous process. If
+// error is not nil, expect the error channel to be nil.
+//
+// Use acquireWait to block until the lease is acquired.
+//
 // t.mu needs to be locked.
 func (t *tableState) acquireNodeLease(
 	ctx context.Context, m *LeaseManager, minExpirationTime hlc.Timestamp,
-) error {
+) (chan error, error) {
```
vivekmenezes: I think you can make this method only return a ```chan error```

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
 	if m.isDraining() {
-		return errors.New("cannot acquire lease when draining")
+		return nil, errors.New("cannot acquire lease when draining")
 	}
 
-	// Notify when lease has been acquired.
+	// Notify when lease is being acquired.
 	t.mu.acquiring = make(chan struct{})
+	errChan := make(chan error)
```
vivekmenezes: create this channel at the top and use it throughout. Use a buffer of 1 so that doesn't block on send or receive

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
+
+	err := t.stopper.RunAsyncTask(
+		ctx, "sql.tableState: async table lease acquisition", func(ctx context.Context) {
+			err := t.acquireNodeLeaseBlocking(ctx, m, minExpirationTime)
+			if err != nil {
+				log.Error(ctx, errors.Wrapf(err, "Async TableDescriptor lease acquisition error"))
+				errChan <- err
+			}
+			close(errChan)
```
vivekmenezes: you don't need to close the channel

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
+	// Notify when lease is being acquired.
 	t.mu.acquiring = make(chan struct{})
+	errChan := make(chan error)
+
+	err := t.stopper.RunAsyncTask(
+		ctx, "sql.tableState: async table lease acquisition", func(ctx context.Context) {
+			err := t.acquireNodeLeaseBlocking(ctx, m, minExpirationTime)
+			if err != nil {
+				log.Error(ctx, errors.Wrapf(err, "Async TableDescriptor lease acquisition error"))
+				errChan <- err
```
vivekmenezes: return err even when nil

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
+//
+// t.mu.acquiring must be initialized before calling.
+func (t *tableState) acquireNodeLeaseBlocking(
+	ctx context.Context, m *LeaseManager, minExpirationTime hlc.Timestamp,
+) error {
+	defer t.mu.Unlock()
 	defer func() {
 		close(t.mu.acquiring)
 		t.mu.acquiring = nil
 	}()
```
vivekmenezes: I think moving the close and setting to nil of acquiring is cleaner done in acquireNodeLease(). You already do that when you hit an error

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
+			if err != nil {
+				log.Error(ctx, errors.Wrapf(err, "Async TableDescriptor lease acquisition error"))
+				errChan <- err
+			}
+			close(errChan)
+		})
+
+	if err != nil {
+		close(t.mu.acquiring)
+		close(errChan)
```
vivekmenezes: just return the error on the channel. no need to close it.

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
+		t.mu.acquiring = nil
+	}
+
+	return errChan, err
```
vivekmenezes: return errChan

pkg/sql/lease.go
```diff
@@ -777,30 +800,64 @@ func (t *tableState) acquireWait() {
...
-	// We're called with mu locked, but need to unlock it during lease
-	// acquisition.
-	t.mu.Unlock()
+
 	table, err := m.LeaseStore.acquire(ctx, t.id, minExpirationTime)
 	t.mu.Lock()
 	if err != nil {
 		return err
 	}
 	t.upsertLocked(ctx, table, m)
```
vivekmenezes: lock () and defer Unlock() just above this

pkg/sql/lease.go
```diff
@@ -98,6 +98,15 @@ func (s *tableVersionState) hasExpired(timestamp hlc.Timestamp) bool {
 	return !timestamp.Less(s.expiration)
 }
 
+// leaseHasEnoughLifeLeft checks whether the table descriptor lease has
+// enough life: that it expires further then LeaseRenewalTimeout
+// duration from now.
+func (s *tableVersionState) leaseHasEnoughLifeLeft() bool {
+	// Get the current time and move it forward by the lease renewal timeout.
+	t := hlc.Timestamp{WallTime: hlc.UnixNano() + int64(LeaseRenewalTimeout)}
+	return t.Less(s.expiration)
```
vivekmenezes: I'd return !t.Less(s.expiration) here and call this method shouldPreemptivelyRenewLease(). The current name is easy to confuse with hasExpired()

vivekmenezes: I also think you can use the clock in tableVersionState instead of calling hlc.UnixNano()

pkg/sql/lease.go
```diff
@@ -581,26 +593,35 @@ func (t *tableState) acquire(
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
+	// When a lease is about to expire but has not yet (determined by
+	// leaseHasEnoughLifeLeft), then ensure an asynchronous routine to renew the
+	// lease is started.
+	if s := t.mu.active.findNewest(); s != nil && !s.hasExpired(timestamp) && !s.leaseHasEnoughLifeLeft() && t.mu.acquiring == nil {
```
vivekmenezes: I don't believe you need hasExpired() here. If it's true you'll acquire the lease here and wait in acquireWait() below

pkg/sql/lease.go
```diff
@@ -581,26 +593,35 @@ func (t *tableState) acquire(
...
+			errChan, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+			if err != nil {
+				return nil, err
+			}
+			t.acquireWait()
+			if err = <-errChan; err != nil {
+				return nil, err
+			}
+		} else if t.mu.acquiring != nil {
+			t.acquireWait()
```
vivekmenezes: why this change?

pkg/sql/lease.go
```diff
@@ -816,7 +840,10 @@ func (t *tableState) acquireNodeLease(
 
 	// Notify when lease is being acquired.
 	t.mu.acquiring = make(chan struct{})
-	errChan := make(chan error)
+	// Notify the caller when we get an asynchronous acquisition error.
+	// Make it buffered so the task doesn't block if the caller ignores
+	// the channel.
+	errChan := make(chan error, 1)
```
vivekmenezes: move this into the previous commit

vivekmenezes (commented):
Thank you for you patience on this PR!

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 47 unresolved discussions, some commit checks failed.

---

*[pkg/sql/lease.go, line 828 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r10-828:-KvV76Nj7OzudnzH2jeg:b-ueh9e) ([raw file](https://github.com/cockroachdb/cockroach/blob/61c166d286bb2cabb12597dfd263b99b402fc1f7/pkg/sql/lease.go#L828)):*
> In the case of a request spawning, a background refresh, passing through the current request context I think is not correct because the task can outlive the request.

Bad folklore. `RunAsyncTask()` uses `ForkCtxSpan()`, which is explicitly documented to allow the child ctx to outlive the parent context.  

---

*[pkg/sql/lease.go, line 49 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvUyQ2FFV_GsNdccpYJ:-KvUyQ2FFV_GsNdccpYK:b-jfpj06) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L49)):*
> ```Go
> var (
> 	// LeaseDuration is the mean duration a lease will be acquired for. The
> 	// actual duration is jittered using LeaseJitterMultiplier.
> ```

since we almost removed this effect, I'd suggest adding a comment to explain the point of jittering: to prevent a cascade of concurrent re-acquisitions for a bunch of leases acquired at the same time.

---

*[pkg/sql/lease.go, line 57 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvUxiZVb0dBjNb-V3j_:-KvUxiZVb0dBjNb-V3ja:bcn2t65) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L57)):*
> ```Go
> 	// [(1-LeaseJitterMultiplier) * LeaseDuration, (1+LeaseJitterMultiplier) * LeaseDuration]
> 	// Exported for testing purposes only.
> 	LeaseJitterMultiplier = 0.25
> ```

The name seems wrong. How about `LeaseJitterFraction` ?

---

*[pkg/sql/lease.go, line 134 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvUyk9joHLx_sud6Wlb:-KvUyk9joHLx_sud6Wlc:b60xnch) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L134)):*
> ```Go
> // [(1-LeaseJitterMultiplier) * leaseDuration, (1+LeaseJitterMultiplier) * leaseDuration].
> func jitteredLeaseDuration() time.Duration {
> 	return time.Duration(float64(LeaseDuration) * (1 - LeaseJitterMultiplier + 2*LeaseJitterMultiplier*rand.Float64()))
> ```

line is over 100 chars

---

*[pkg/sql/lease.go, line 586 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvUz4dCdFNAbCT0TmVe:-KvUz4dCdFNAbCT0TmVf:bals6qf) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L586)):*
> ```Go
> 	// Acquire a lease if no lease exists or if the latest lease is
> 	// about to expire. Repeatedly check to ensure either a lease is
> 	// acquired or we receive an error. This is required for when another
> ```

s/for when/when

---

*[pkg/sql/lease.go, line 588 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvUzLF_TehK39noMIrM:-KvUzLF_TehK39noMIrN:b60xnch) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L588)):*
> ```Go
> 	// acquired or we receive an error. This is required for when another
> 	// routine acquires a lease and releases it immediately.
> 	for s := t.mu.active.findNewest(); s == nil || s.hasExpired(timestamp); s = t.mu.active.findNewest() {
> ```

line is over 100 chars

---

*[pkg/sql/lease.go, line 591 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV-24U_HUpVZ2mN4qA:-KvV-24U_HUpVZ2mN4qB:b-ii84u6) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L591)):*
> ```Go
> 		// Wait for any existing lease acquisition.
> 		t.acquireWait()
> 		// If the rcently acquired lease is valid for use, short-circuit.
> ```

rcently

---

*[pkg/sql/lease.go, line 807 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV50WJuL7UqkDcfwCS:-KvV50WJuL7UqkDcfwCT:bqhhyra) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L807)):*
> ```Go
> // error is not nil, expect the error channel to be nil.
> //
> // Use acquireWait to block until the lease is acquired.
> ```

This guidance seems wrong. Callers to this function should use the returned channel (and reacquire the lock if they had released it). There's not much reason for them to use `acquireWait()`.
`acquireWait` should be used by other suckers that didn't get to call `acquireNodeLease` because they saw `t.mu.acquiring` set.

Like I said elsewhere, I suspect the code would be simplified significantly if we used (our fork or) `x/sync/singleflight`.

---

*[pkg/sql/lease.go, line 809 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV8h69gnzGvRrE8zXO:-KvV8h69gnzGvRrE8zXP:bm5n9dq) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L809)):*
> ```Go
> // Use acquireWait to block until the lease is acquired.
> //
> // t.mu needs to be locked.
> ```

this should also say that callers *have to* release the lock while waiting on the channel.

---

*[pkg/sql/lease.go, line 810 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV29DK9FIzZj84oZH8:-KvV29DK9FIzZj84oZH9:b-vrljrg) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L810)):*
> ```Go
> //
> // t.mu needs to be locked.
> func (t *tableState) acquireNodeLease(
> ```

nit: take the opportunity to rename this to `acquireNodeLeaseLocked()` since it requires `t.mu` to be held.

---

*[pkg/sql/lease.go, line 812 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r11-812:-KvV0vxKGaRCGU7y20Vy:bfpls9c) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L812)):*

I think you can make this method only return a ```chan error```
</blockquote></details>

+1. Fill the channel immediately when you currently return an error directly.

Better yet, check out and consider using `x/sync/singleflight`, which Nathan recently forked and improved into crdb, to replace `acquireWait()` and `t.mu.acquiring`.

---

*[pkg/sql/lease.go, line 844 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV3w8Ij1EbdwUUdmXP:-KvV3w8Ij1EbdwUUdmXQ:b-435a3p) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L844)):*
> ```Go
> //
> // t.mu.acquiring must be initialized before calling.
> func (t *tableState) acquireNodeLeaseBlocking(
> ```

the naming convention is the other way around:  what's now `acquireNodeLease` should be `acquireNodeLease(Locked)Async` and this one should be `acquireNodeLease(Locked)f`. Then the comment doesn't need to mention anything about "blocking" - everything is "blocking" unless otherwise indicated.

---

*[pkg/sql/lease.go, line 847 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV90j3oTDFNIEkowON:-KvV90j3oTDFNIEkowOO:b-z6y8p) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L847)):*
> ```Go
> 	ctx context.Context, m *LeaseManager, minExpirationTime hlc.Timestamp,
> ) error {
> 	defer t.mu.Unlock()
> ```

don't defer this before you've taken the lock!

---

*[pkg/sql/lease.go, line 860 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV3huzpUPyl6hL7CVQ:-KvV3huzpUPyl6hL7CVR:b-kxdwii) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L860)):*
> ```Go
> 	t.upsertLocked(ctx, table, m)
> 	t.tableNameCache.insert(table)
> ```

nit: superfluous diff

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844)*

vivekmenezes:
If you replace acquiring can you do it in another commit. Perhaps the first
one. You can also do it in another pr before this one

lego:
*[pkg/sql/lease.go, line 107 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-Kv4DPdVHew3Vzx3paHg-r12-107:-KvX_Un9cHQdtRu3iP9X:b2oulkc) ([raw file](https://github.com/cockroachdb/cockroach/blob/02e1911ce934f0c68e791c791d9168bc11293688/pkg/sql/lease.go#L107)):*
> tableVersionState
I don't think there's a clock accessible through `tableVersionState`. It's neither on that struct of the embedded `TableDescriptor`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844)*

lego:
*[pkg/sql/lease.go, line 807 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV50WJuL7UqkDcfwCS:-KvXhvEzT1fF0QnaDDL7:bmtunyy) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L807)):*

This guidance seems wrong. Callers to this function should use the returned channel (and reacquire the lock if they had released it). There's not much reason for them to use `acquireWait()`.
`acquireWait` should be used by other suckers that didn't get to call `acquireNodeLease` because they saw `t.mu.acquiring` set.

Like I said elsewhere, I suspect the code would be simplified significantly if we used (our fork or) `x/sync/singleflight`.
</blockquote></details>

For the case when we only have synchronous codepaths, I do agree singleflight will simplify things.

When we introduce the async renewal codepath, i think this may complicate using singlepath quite a bit but I have a few ideas where most of this can be reduced. (wow, our fork is pretty wicked with Do _and_ DoChan!)

Going to put the first commit into a separate PR to simplify things, and I'll only continue this PR once that one settles.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844)*

vivekmenezes:
Review status: 0 of 3 files reviewed at latest revision, 47 unresolved discussions, some commit checks failed.

---

*[pkg/sql/lease.go, line 807 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844#-KvV50WJuL7UqkDcfwCS:-KvY7ye8uDq4exRERd5J:b-13gxhn) ([raw file](https://github.com/cockroachdb/cockroach/blob/167c13c2a415b6880739c818fae44fc9aa7cc053/pkg/sql/lease.go#L807)):*

For the case when we only have synchronous codepaths, I do agree singleflight will simplify things.

When we introduce the async renewal codepath, i think this may complicate using singlepath quite a bit but I have a few ideas where most of this can be reduced. (wow, our fork is pretty wicked with Do _and_ DoChan!)

Going to put the first commit into a separate PR to simplify things, and I'll only continue this PR once that one settles.
</blockquote></details>

sounds good. Just refer to the issue in all the PRs

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18844)*

lego:
Closing in favour of #19005 (same solution, different approach)

PR #18845, opened 2017-09-27
[WIP] rfc: debug snapshots

couchand:
This RFC proposes expanding the operation of the debug zip command to
include the time series metrics and more of the debug pages, as well as
developing a tool to browse the data based on the admin UI.

### DIFF
--- a/docs/RFCS/20170927_debug_snapshots.md
+++ b/docs/RFCS/20170927_debug_snapshots.md
@@ -1,0 +1,186 @@
+- Feature Name: Debug Snapshots
+- Status: draft
+- Start Date: 2017-09-27
+- Authors: Andrew Dona-Couch
+- RFC PR: (PR # after acceptance of initial draft)
+- Cockroach Issues: #16584, #14584
+
+# Summary
+
+Debugging issues in the field is a significant and growing part of the ongoing
+development of CockroachDB.  This is supported by the `debug zip` command, which
+provides a quick export of much debugging information from a cluster, allowing
+us to investigate and diagnose issues without needing direct access.  This RFC
+proposes an expansion of the data available in this export, as well as the
+development of a tool based on the admin UI to help our team quickly navigate
+the data exported.
+
+# Motivation
+
+We regularly ask users to run `cockroach debug zip` and send us the output, and
+separately to take screenshots of the various time series graphs in the admin UI.
+While this process has done okay till now, there are a few drawbacks that we'd
+like to avoid:
+
+- Not every debug page is exported.
+- The screenshotting is entirely manual.
+- It requires either knowing in advance which time series we care about or lots
+  of scrolling and clicking to get all the screenshots.
+- Users don't always capture the right screenshots, adding costly round trips.
+- The result is a static view that isn't browsable.
+- We don't get the underlying data, just a low-fidelity view of it.
+
+By extending the existing `debug zip` command to export the time series data,
+we can roll up all of the user's manual effort into a single command invocation,
+and also make sure we get the data we need to solve their problem.
+
+# Guide-level explanation
+
+There are two main components in this debug system: the export command of the
+Cockroach CLI, and the new browsing UI for the output of this command.
+
+An export of useful debugging information can be made by running
+`cockroach debug zip [filename]` (with the appropriate connection and
+authentication flags for the cluster).  After a little while this will produce
+the requested file which can be sent along to an engineer on support duty for
+further debugging.  This should be the primary way a user gives us insight into
+their cluster; we might even think about requesting it from the start in the
+GitHub issue template (though it would need to be submitted separately).
+
+The new browsing UI will provide an interface largely the same as the built-in
+admin UI.  Rather than showing the metrics and configuration of a running
+cluster, shows the data from the debug export.  This should be as simple as
+specifying the source data file and then browsing the otherwise-familiar admin
+UI pages.
+
+# Reference-level explanation
+
+Enhancements to the `debug zip` command come in two basic flavors:
+
+- exporting more of the debug endpoints (e.g. problem ranges) and
+- exporting time series data.
+
+Adding additional debug endpoints to the zip file is a no-brainer.  It is a
+straightforward continuation of the existing tooling, and so could probably be
+done as a starter project or by an external contributor without much guidance.
+
+Some candidate debug pages include:
+
+- problem ranges,
+- allocator,
+- certs,
+- raft debug,
+- profiling pages:
+  - heap dumps
+  - gorouting dumps
+- etc.
+
+Exporting the time series data will require more thoughtful design, so that is
+where the focus of the rest of this RFC will be.  The main questions are about
+the scope of data exported and the format we write it to disk.  It seems
+reasonable to write the time series data in the raw protobuf format that we use
+to transfer it over the wire to the admin UI: this is an already-established
+serialization format, and may make the UI changes easier.
+
+The amount of data we want to export is an open question.  In the long-term, we
+may want some amount of configurability, since some problems are not noticed as
+soon as they occur.  However, most issues that we debug are noticed relatively
+quickly, so for a first pass we could pick a time limit and just use that.
+
+The other side of this project is building a tool to browse the exported data.
+This should be as close to the experience of using the admin UI as possible - it
+should feel to the support engineer as though they are looking at the admin UI
+of the cluster in question as of the time the issue was identified.  For good
+software engineering reasons, we should try to use the admin UI codebase as well.
+
+Since this tool is primarily intended to be used by Cockroach Labs engineers, it
+doesn't need to be built into the `cockroach` binary.  Of course it could be,
+which would have a certain elegance, but if it would be easier to build as a
+separate tool we should feel justified in going down that route.
+
+A few specific changes need to be made to the admin UI code to support this:
+
+- disabling ongoing data refresh
+- changing the time period selection to work well with the recorded time period
+- other changes as needed
+
+The backend of this tool can be built in one of two ways.  Either we just spin
+up a one-node cluster and import the zipped data, or we write a bespoke server
+for this purpose.  There are reasons to think the either of these choices will
+be faster than the other, but the evidence seems to lean towards writing a
+bespoke tool for this purpose.  It wouldn't require us to add a bunch of
+conditional code paths for the debugging endpoints, which could be fraught.  The
+new tool would just implement the various gRPC endpoints, but backed by the
+debug zip rather than the running cluster, as well as serving the admin UI code.
+
+## Detailed design
+
+TODO: detailed design
+
+Outline both "how it works" and "what needs to be changed and in which order to get there."
+
+Describe the overview of the design, and then explain each part of the
+implementation in enough detail that reviewers will be able to
+identify any missing pieces. Make sure to call out interactions with
+other active RFCs.
+
+## Drawbacks
+
+This proposal will take a non-zero amount of developer effort.  However, one can
+argue that it will, in the long run, save much more developer effort.
+
+There will be some changes necessary to the main `cockroach` binary (more if we
+build the browsing tool into the binary).  But, these won't be extensive.
+
+Some users will be hesitant to send us a zip file of detailed information about
+their running production cluster, and would be more comfortable with the current,
+screenshot-driven system.  In that case, there's nothing stopping us from
+continuing to use screenshots, though if it ends up being a significant fraction
+of our users the development time will have been wasted.
+
+## Rationale and Alternatives
+
+### Automating screenshots
+
+Using something like [this POC](https://github.com/tschottdorf/roacheteer), we
+could automate the screenshot process, relieving the burden on the users.
+However, we still only get a low-fidelity view of the data, giving us less
+useful information for the debugging process.
+
+### A new "everything" view + print-to-PDF
+
+Making an "everything" view and using print-to-pdf instead of screenshots could
+also be an alternative. Still low-fidelity (although we could customize the
+display for print purposes, e.g. by including legends on the graphs instead of
+only showing them on hover), but would be easier for users than a
+headless-browser-based screenshot tool.
+
+Low fidelity can be a good thing if users (especially those in regulated
+industries) feel more comfortable sending us a PDF full of graphs than logs that
+may contain sensitive information. This will matter more as these customers move
+from the proof of concept phase into production.
+
+### Just add more of the debug pages
+
+We'll get some benefit by just expanding the number of exported debug pages.
+The problem ranges page, in particular, would be very helpful.  This would get
+us part of the way there, but more and more the debug pages are getting fairly
+dynamic, and the time series data is also really useful.
+
+### Business as usual
+
+We continue expecting users to take screenshots of time series we need.  This
+requires more back-and-forth with our engineers on support, making the debug
+process take longer as well as making everyone's experience more frustrating.
+
+## Unresolved questions
+
+- How much data is too much?  How much is enough?
+- Build in to the binary or a separate tool?
+- What's the best way to receive the zip files?  Until now we've just had users
+  send them to us over e-mail.  A future improvement would be to create some
+  kind of "black hole" drop box for users to send us the debug files.
+- What's the debug story for wedged clusters?  We're necessarily using this tool
+  when something's gone wrong, so it needs to work when some things are broken.
+  We might allow the debug export to fall back to an inconsistent read if we
+  can't successfully get what we're looking for consistently.


tschottdorf:
Thanks for writing this up, I think this is a good direction. I'd be nice to have some back-of-envelope computations that relate expected time series dump size with the amount of history we want (for example: 1 week at 10s resolution = 512mb) as that'll give us a realistic idea of the size of the dump vs its contents. You could just let a node sit for a day and then take the size of the RocksDB directory as an indicator. For better ideas, talk to @mrtracy :smile: 

---

---

*[docs/RFCS/20170927_debug_snapshots.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4L64CdQwMvm2ax_vN:-Kv4L64CdQwMvm2ax_vO:b-ezrmkr) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L22)):*
> ```Markdown
> We regularly ask users to run `cockroach debug zip` and send us the output, and
> separately to take screenshots of the various time series graphs in the admin UI.
> While this process has worked acceptably till now, there are a few drawbacks
> ```

I think it's up for debate whether this has worked "acceptably". :smile: 

---

*[docs/RFCS/20170927_debug_snapshots.md, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4LCjNUCjPi9nVafUp:-Kv4LCjNUCjPi9nVafUq:b-papmx) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L30)):*
> ```Markdown
>   of scrolling and clicking to get all the screenshots.
> - The result is a static view that isn't browsable.
> - We don't get the underlying data, just a low-fidelity view of it.
> ```

Users often don't send us quite what we asked for. An additional round-trip can take hours and a lot of time is also spent verbalizing the request and giving (hopefully) good instructions.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 47 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4LOTRqjc_fnKr6WKr:-Kv4LOTRqjc_fnKr6WKs:b-w9fzvv) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L47)):*
> ```Markdown
> further debugging.  This should be the primary way a user gives us insight into
> their cluster; we might even think about requesting it from the start in the
> GitHub issue template.
> ```

I like this, but! most of that stuff is sensitive (though not extremely so) and so users can't just upload it (besides, it might be too large). We should consider running a "black hole" that users can throw the file into. This might be out of scope for this PR, but it's perhaps relatively straightforward (an AWS key which has write-only access to some bucket?)

---

*[docs/RFCS/20170927_debug_snapshots.md, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4MNe9KmTfV75UqDZK:-Kv4MNe9KmTfV75UqDZL:b-x38ovf) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L107)):*
> ```Markdown
> for this purpose.  There are reasons to think the either of these choices will
> be faster than the other, but I, as the author of this RFC, expect that it will
> be easier to write a bespoke tool.  (TODO: justify)
> ```

I initially thought the import was easier, but I've since come around to your point. Not so much because of the time series -- rather because I think it'll be really annoying to point the `ui` at the problemranges and navigate everything when it's a live cluster the ui is talking to.

The bespoke tool would just implement the various grpc server interfaces and serve up static-ish responses. I actually think the timeseries might be the one where the real server wins out most, but I expect that to be outweighed by the other stuff, and you should be able to set up a slim server that omits many of the moving parts but can serve timeseries scans just fine.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4N7gn_gsg7UneaXON:-Kv4N7gn_gsg7UneaXOO:brq3tm7) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L139)):*
> ```Markdown
> ### Just add more of the debug pages
> 
> We'll get some benefit by just expanding the number of exported debug pages.
> ```

Debug pages now tend to be fairly dynamic (think problem ranges), so they're not so easy to just export any more. Or, at least curl is far from enough. You need to render and execute in a headless browser.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-:-Kv4NIGNzE7dYmeusY-e:b-1v63er)*

tschottdorf:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845)*

bdarnell:
This seems like a good time to introduce two different modes of `debug zip`: one that includes as much as we can, and one that only includes less sensitive information. The difference IMHO is mainly about whether any keys from the database could be included (so timeseries stuff is OK to include in the non-sensitive zip, but logs and some debug pages only go in the full version)

For the less-sensitive debug zip, it would be great if we could serve it from the UI so you could get it without fumbling with CLI certs. 

---

---

*[docs/RFCS/20170927_debug_snapshots.md, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv7qPihS1HaySCVPxYY:-Kv7qPiiiwwVX7zxillf:b-4j8kpy) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L66)):*
> ```Markdown
> done as a starter project or by an external contributor without much guidance.
> 
> Some candidate debug pages include:
> ```

Also the range debug pages for all meta and liveness ranges.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv7r8uaE7T607sN9CCJ:-Kv7r8uaE7T607sN9CCK:b-ayrxhh) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L130)):*
> ```Markdown
> ## Rationale and Alternatives
> 
> ### Automating screenshots
> ```

Making an "everything" view and using print-to-pdf instead of screenshots could also be an alternative. Still low-fidelity (although we could customize the display for print purposes, e.g. by including legends on the graphs instead of only showing them on hover), but would be easier for users than a headless-browser-based screenshot tool. 

Low fidelity can be a good thing if users (especially those in regulated industries) feel more comfortable sending us a PDF full of graphs than logs that may contain sensitive information. This will matter more as these customers move from the proof of concept phase into production. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-:-Kv7ooLqkLCY8CiToAwr:b-ntbvrz)*

couchand:
TFTR, you both provided helpful feedback as I go back to finish this up.  Good idea about getting hard numbers about the data, that will definitely help us make specific decisions.  I also agree that there's room for a few configuration flags on the debug zip command - the area I had been thinking about was the size/detail tradeoff, but the sensitivity issue is a good one, too.

---

---

*[docs/RFCS/20170927_debug_snapshots.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4L64CdQwMvm2ax_vN:-Kv8VkCL67vXWEjj708Q:b2nhg5f) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L22)):*

I think it's up for debate whether this has worked "acceptably". :smile: 
</blockquote></details>

Ha!

---

*[docs/RFCS/20170927_debug_snapshots.md, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4LCjNUCjPi9nVafUp:-Kv8VewoR_MlKqGmPPho:b-ylmks0) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L30)):*

Users often don't send us quite what we asked for. An additional round-trip can take hours and a lot of time is also spent verbalizing the request and giving (hopefully) good instructions.
</blockquote></details>

That's a great point, I'll add that.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 47 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4LOTRqjc_fnKr6WKr:-Kv8Vm_Q_pU-YWs2460X:b9yz9or) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L47)):*

I like this, but! most of that stuff is sensitive (though not extremely so) and so users can't just upload it (besides, it might be too large). We should consider running a "black hole" that users can throw the file into. This might be out of scope for this PR, but it's perhaps relatively straightforward (an AWS key which has write-only access to some bucket?)
</blockquote></details>

I had a conversation with @dianasaur323 about the size of the debug zip and this came up there, too.  We both agree that would be very helpful, just from the perspective of making sure we can actually receive it no matter the size (the zip from a running cluster is already pretty big - 40M from one of the prod clusters, for example).  But it's probably another project; I'll add a mention of it here for good measure.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv7qPihS1HaySCVPxYY:-Kv8Wq32evPzWDZfJXwm:bavkljn) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L66)):*

Also the range debug pages for all meta and liveness ranges.
</blockquote></details>

Correct me if I'm wrong, but I believe the zip currently exports the range debug data for every range: https://github.com/cockroachdb/cockroach/blob/93b84fff46964d98cc1fd577657feea7e3f004c2/pkg/cli/zip.go#L217

Please do let me know if that's excluding the meta and liveness ranges somehow that I don't notice.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4MNe9KmTfV75UqDZK:-Kv8XJKFGJ2CxtfaM7vi:bp5u1lv) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L107)):*

I initially thought the import was easier, but I've since come around to your point. Not so much because of the time series -- rather because I think it'll be really annoying to point the `ui` at the problemranges and navigate everything when it's a live cluster the ui is talking to.

The bespoke tool would just implement the various grpc server interfaces and serve up static-ish responses. I actually think the timeseries might be the one where the real server wins out most, but I expect that to be outweighed by the other stuff, and you should be able to set up a slim server that omits many of the moving parts but can serve timeseries scans just fine.
</blockquote></details>

Just what I was thinking, thanks for giving me a few specific details to get started writing this up.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv7r8uaE7T607sN9CCJ:-Kv8X_3C9hal-UcHP5pN:b-p5c8hn) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L130)):*

Making an "everything" view and using print-to-pdf instead of screenshots could also be an alternative. Still low-fidelity (although we could customize the display for print purposes, e.g. by including legends on the graphs instead of only showing them on hover), but would be easier for users than a headless-browser-based screenshot tool. 

Low fidelity can be a good thing if users (especially those in regulated industries) feel more comfortable sending us a PDF full of graphs than logs that may contain sensitive information. This will matter more as these customers move from the proof of concept phase into production. 
</blockquote></details>

That's a good point, we can work around some of the limitations of the screenshot-style solution by adding an everything view (which might be useful on it's own, anyway).  I'll write up this alternative.

I just ran a print-to-pdf experiment, and it's almost completely unhelpful at the moment.  So if we were going to go that route, we'd also have to put some attention into getting the printing layout to look right.

I wonder about the sensitivity aspect: I can see your point when it relates to the debug pages above, but I'm not sure I get it regarding detailed time-series data.  Is the concern there mostly about their gut feeling?

---

*[docs/RFCS/20170927_debug_snapshots.md, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4N7gn_gsg7UneaXON:-Kv8YlfcTt6fuTileLXg:bb7j7c8) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L139)):*

Debug pages now tend to be fairly dynamic (think problem ranges), so they're not so easy to just export any more. Or, at least curl is far from enough. You need to render and execute in a headless browser.
</blockquote></details>

For the purposes of this alternative, I just mean we'll add more types of JSON export.  For pages like the problem ranges, that would just be a bunch of different JSON files, much like the ranges exports are now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-:-Kv8Z3hQG9ZInw1NM4ip:b-2dxh3a)*

bdarnell:

---

*[docs/RFCS/20170927_debug_snapshots.md, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv7qPihS1HaySCVPxYY:-Kv8fl4rGVSGpgavTUo2:b-7qvpuv) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L66)):*

Correct me if I'm wrong, but I believe the zip currently exports the range debug data for every range: https://github.com/cockroachdb/cockroach/blob/93b84fff46964d98cc1fd577657feea7e3f004c2/pkg/cli/zip.go#L217

Please do let me know if that's excluding the meta and liveness ranges somehow that I don't notice.
</blockquote></details>

Oh, I didn't know that. (I obviously haven't used debug zips much)

---

*[docs/RFCS/20170927_debug_snapshots.md, line 130 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv7r8uaE7T607sN9CCJ:-Kv8fvYZUz3mvwWvk52c:b-j88tgd) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L130)):*

That's a good point, we can work around some of the limitations of the screenshot-style solution by adding an everything view (which might be useful on it's own, anyway).  I'll write up this alternative.

I just ran a print-to-pdf experiment, and it's almost completely unhelpful at the moment.  So if we were going to go that route, we'd also have to put some attention into getting the printing layout to look right.

I wonder about the sensitivity aspect: I can see your point when it relates to the debug pages above, but I'm not sure I get it regarding detailed time-series data.  Is the concern there mostly about their gut feeling?
</blockquote></details>

The time-series data is not sensitive, but by using a "human-readable" format like PDF instead of exporting the raw data, it's more apparent that we're not going to intentionally or unintentionally sneak anything else out. (there's plenty of room for shenanigans in a PDF if we were being evil, but paranoid users also have counter-measures like re-encoding the PDF themselves)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845)*

nvanbenschoten:

---

*[docs/RFCS/20170927_debug_snapshots.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-KwM8E5H1TRfadbylhNm:-KwM8E5ITaZU7vDdGGJp:bd3wov3) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L60)):*
> ```Markdown
> 
> - exporting more of the debug endpoints (e.g. problem ranges) and
> - exporting time series data.
> ```

One thing I'm worried about here is that this information is often unavailable on wedged clusters, for various reasons. I love the idea to expose this in the happy case, but as with anything when dealing with debug information, we should make sure we're able to get _some_ information out even if the rest isn't possible. In practice, this means expecting failures when retrieving any sources of information and reporting the failures instead of erroring out of the entire `debug zip` process. If we do this correctly, these isolated failures of different subsystems can actually be good indications of where investigation should be focused.  

---

*[docs/RFCS/20170927_debug_snapshots.md, line 72 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-KwMA8T08onIP_v42Fp-:-KwMA8T08onIP_v42Fp0:brrl9lw) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L72)):*
> ```Markdown
> - certs,
> - raft debug,
> - etc.
> ```

Profiling pages like heap and goroutine dumps would also be helpful, although we're probably not persisting either at the moment.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 90 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-KwMAppNsN3hqPIjnjp7:-KwMAppNsN3hqPIjnjp8:b32v57) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L90)):*
> ```Markdown
> should feel to the support engineer as though they are looking at the admin UI
> of the cluster in question as of the time the issue was identified.  For good
> software engineering reasons, we should try to use the admin UI codebase as well.
> ```

+1

---

*[docs/RFCS/20170927_debug_snapshots.md, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4MNe9KmTfV75UqDZK:-KwMBAL5p_JOuKWfHEyI:br42md4) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L107)):*

Just what I was thinking, thanks for giving me a few specific details to get started writing this up.
</blockquote></details>

Yeah, this approach makes the most sense to me as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845)*

bdarnell:

---

*[docs/RFCS/20170927_debug_snapshots.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-KwM8E5H1TRfadbylhNm:-Kwf1hBhrgdVSor8Sa2u:brhtxbq) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L60)):*

One thing I'm worried about here is that this information is often unavailable on wedged clusters, for various reasons. I love the idea to expose this in the happy case, but as with anything when dealing with debug information, we should make sure we're able to get _some_ information out even if the rest isn't possible. In practice, this means expecting failures when retrieving any sources of information and reporting the failures instead of erroring out of the entire `debug zip` process. If we do this correctly, these isolated failures of different subsystems can actually be good indications of where investigation should be focused.  
</blockquote></details>

We have the option of using inconsistent reads to get timeseries (and other) data. The UI used to do this by default but we switched to consistent reads because inconsistent reads are allowed to be arbitrarily far behind. It would make sense (for both the UI and debug zip) to attempt a consistent read first then fall back to an inconsistent read to get *something* when things are broken. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845)*

dianasaur323:
Everything overall looks good to me, although I am concerned with scope. I think you mentioned this in one of our admin UI meetings. It's looking like it may be increasingly difficult to implement this in its entirety in this release. Probably a discussion for another day though.

On a separate note, I may have missed this while reading through, but have you thought about how the user will export this to us? What command will they run, and how much data are they allowed to specify? I think our time selector goes pretty far back. 

Finally, this capability sounds quite similar to the Oracle AWR tool, which allows users to export performance stats and browse them offline. Just food for thought.

---

Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/20170927_debug_snapshots.md, line 47 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-Kv4LOTRqjc_fnKr6WKr:-KxLh98lW9FSZ7cOPBxp:b-fk5eyi) ([raw file](https://github.com/cockroachdb/cockroach/blob/661b23cf6503b8ec2572043c8ca02b497ffd1c11/docs/RFCS/20170927_debug_snapshots.md#L47)):*

I had a conversation with @dianasaur323 about the size of the debug zip and this came up there, too.  We both agree that would be very helpful, just from the perspective of making sure we can actually receive it no matter the size (the zip from a running cluster is already pretty big - 40M from one of the prod clusters, for example).  But it's probably another project; I'll add a mention of it here for good measure.
</blockquote></details>

Yes, I think the privacy stuff is something we should be careful of, since we don't do any scrubbing. Because of that, we might not want to make this the default way to request for more data, or at least put a disclaimer that they need to be careful about sending us private data.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18845#-:-KxLj5A1cRnYYmWi8oZ5:bk6e1p2)*

cuongdo:
@couchand Could you iterate on this RFC? It's important to continue to investigate scope out this work.

BramGruneir:
I just want to bump this back up.  We should look into this for 2.2.  It obviously needs a lot more work and every decision should be revisited now, but this shouldn't be left on the backlog.

cc: @piyush-singh, @awoods187

knz:
Discussed this offline with @piyush-singh : this work is still in-scope but not prioritized yet. We have linked this RFC PR in our product database, and intend to refer to it when we re-prioritize the work.

Until then, I am going to close the PR, so as to reduce our open PR metrics since we are not expecting progress anytime soon.

PR #18834, opened 2017-09-27
storage: don't set {,TxnSpan}GCThreshold to zero-values

petermattis:
Fixes #18831

Issue #18811, opened 2017-09-27
stability: longer and more extensive testing of mixed version clusters


petermattis:
`TestVersionUpgrade` was initially failing in #18689 due to improper handling of mixed-version clusters. Our testing of such clusters is quite weak at the moment. We should probably have a nightly test that runs a mixed-version cluster and perhaps manually test such clusters periodically.

tschottdorf:
What could have caught https://github.com/cockroachdb/cockroach/issues/18825#issue-261000347?

I think a test that runs a cluster under high pressure (to incentivize many lease changes) and with aggressive timings while cycling the binary versions back and forth for an hour, verifying at the end that certain invariants are met (for example, index equal => statemachine equal) would've done, as this is essentially what triggered #18825.

Probably a local acceptance test would do. This doesn't seem very dependent on the cloud, but some random latency wouldn't hurt.

tschottdorf:
Closing for https://github.com/cockroachdb/cockroach/issues/15851.

PR #18848, opened 2017-09-27
protoutil: fuzz fields with changed nullability

tbg:
Changing the nullability of proto fields is tricky. PRs #18825 introduced (at least) two (#18835 and
\#18834) bugs that we found relatively quickly, but not in CI.

In the spirit of advancing #18811, here's an attempt at making our CI provoke these types of errors
more aggressively.

Basically, establish a struct tag `cockroachdb:"randnullable"` that instructs `protoutil.Marshal` to
randomly turn `nil` fields into pointers to the zero value. This is allocation heavy (due to the
need to avoid data races), so it is disabled with the `bench` build tag.

We can then annotate fields whose nullability we change and get more coverage of the implications
of that change.

I was able to reproduce both #18835 (with `TestStoreRangeLease` going into an infinite loop) and
\#18834 (with a lot of assertion crashes).

@petermattis I'm probably missing a few fields. Care to point out which ones should also be fuzzed?


petermattis:
Seems like a reasonable approach. I'd like to see this fuzzing done for all tests and excluded from normal builds and benchmarks. Not sure how to accomplish the last part.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-:-Kv4u08O2dwNQWffl0CZ:bw6bzc3)*

tschottdorf:
Rebasing on top of #18939 made this a lot more awkward because now the following pattern is fraught:

```go
	// All of the Cockroach protos implement MarshalTo and Size. So we marshal
	// directly into the Value.RawBytes field instead of allocating a separate
	// []byte and copying.
	size := msg.Size()
	v.RawBytes = make([]byte, headerSize+size)
	if _, err := protoutil.MarshalTo(msg, v.RawBytes[headerSize:]); err != nil {
		return err
	}
```

since `msg.Size()` isn't actually the size of buffer we need (if we add zero values, we'll add a little more).

@petermattis, any ideas? I could make the interface more complicated to accommodate all of this, but wish there was a more straightforward option.

petermattis:
Approach generally looks good.

Adding the `bench` tag is slightly unfortunate for my workflow as I frequently run `go` directly. Perhaps we should only enable fuzzing for race builds. Is there any effect on test time for some of the slower packages such as `storage`?

---

Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/roachpb/data.go, line 377 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTF2Sa6rigy9rD89nm:-KvTF2Sa6rigy9rD89nn:bdmdzxo) ([raw file](https://github.com/cockroachdb/cockroach/blob/671b06adecf7e8b833e8fde153eabbfc53dbfcd0/pkg/roachpb/data.go#L377)):*
> ```Go
> 	}
> 	// TODO(tschottdorf): this is awkward since it might deep-clone and protoutil.Marshal
> 	// below will do it again. It's a bit hard to move this though.
> ```

Not sure if I'm understanding this comment. `protoutil.Marshal` doesn't deep-clone. 

---

*[pkg/roachpb/data.go, line 378 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTFAPTA8zCYarxKR1-:-KvTFAPTA8zCYarxKR10:bkzp90l) ([raw file](https://github.com/cockroachdb/cockroach/blob/671b06adecf7e8b833e8fde153eabbfc53dbfcd0/pkg/roachpb/data.go#L378)):*
> ```Go
> 	// TODO(tschottdorf): this is awkward since it might deep-clone and protoutil.Marshal
> 	// below will do it again. It's a bit hard to move this though.
> 	msg = protoutil.MaybeFuzz(msg)
> ```

Seems fine to do this here, though it is unfortunate it couldn't be done in `MarshalTo`. 

---

*[pkg/roachpb/data.go, line 1085 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTFQRH8Uf5JeiVEKOG:-KvTFQRH8Uf5JeiVEKOH:b-h8h2ov) ([raw file](https://github.com/cockroachdb/cockroach/blob/671b06adecf7e8b833e8fde153eabbfc53dbfcd0/pkg/roachpb/data.go#L1085)):*
> ```Go
> 
> // BootstrapLease returns the lease to persist for the range of a freshly bootstrapped store. The
> // returned lease is morally "empty" but has a few fields set to null instead of nil because some
> ```

The use of `null` here is a bit confusing. I think it would be better to say: `has a few fields set to non-nil zero values because some ...`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-:-KvTFjmhdUTVI0OJ9WH0:b3futou)*

tschottdorf:
`141.416s` (no fuzzing) vs `160.493s` (fuzzing) in `race`. Updated.

---

Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/roachpb/data.go, line 377 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTF2Sa6rigy9rD89nm:-KvTUTNnN7MYb_20zxPp:bsxhe3m) ([raw file](https://github.com/cockroachdb/cockroach/blob/671b06adecf7e8b833e8fde153eabbfc53dbfcd0/pkg/roachpb/data.go#L377)):*

Not sure if I'm understanding this comment. `protoutil.Marshal` doesn't deep-clone. 
</blockquote></details>

`protoutil.Marshal` (in this code) calls `MaybeFuzz` again and eats another deep clone.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-:-KvT_Pcj2Fc5m8w0i1Cz:b-944kxi)*

tschottdorf:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/roachpb/data.go, line 378 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTFAPTA8zCYarxKR1-:-KvTh8MRxHEYwDT0arVz:b-pbdqz3) ([raw file](https://github.com/cockroachdb/cockroach/blob/671b06adecf7e8b833e8fde153eabbfc53dbfcd0/pkg/roachpb/data.go#L378)):*

Seems fine to do this here, though it is unfortunate it couldn't be done in `MarshalTo`. 
</blockquote></details>

Agreed.

---

*[pkg/roachpb/data.go, line 1085 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTFQRH8Uf5JeiVEKOG:-KvThHaeK9B9M8vKFgLf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/671b06adecf7e8b833e8fde153eabbfc53dbfcd0/pkg/roachpb/data.go#L1085)):*

The use of `null` here is a bit confusing. I think it would be better to say: `has a few fields set to non-nil zero values because some ...`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848)*

petermattis:
:lgtm:

I'll take another look tonight to see if there are additional fields that should be fuzzed.

---

Review status: 0 of 19 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/util/protoutil/marshal.go, line 73 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTi7dSkd-8zyOhisnq:-KvTi7dSkd-8zyOhisnr:butmkcx) ([raw file](https://github.com/cockroachdb/cockroach/blob/e86fc630d6dc5e459b4830c706645e85d2119062/pkg/util/protoutil/marshal.go#L73)):*
> ```Go
> }
> 
> // MarshalTo encodes pb into the wire format. It is used throughout the code
> ```

`s/MarshalTo/MarshalToWithoutFuzzing/g`

---

*[pkg/util/protoutil/randnullability.go, line 102 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTif7P5J07kDFaY80d:-KvTif7P5J07kDFaY80e:b69hpxp) ([raw file](https://github.com/cockroachdb/cockroach/blob/e86fc630d6dc5e459b4830c706645e85d2119062/pkg/util/protoutil/randnullability.go#L102)):*
> ```Go
> 			if bytes.Contains(buf, []byte(`.Benchmark`)) {
> 				log.Fatalf(ctx, "must run with `-tags bench`")
> 			}
> ```

This check is no longer necessary, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-:-KvTiwra7sK8ibp2C0Wv:bbti2l9)*

tschottdorf:
Review status: 0 of 19 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/util/protoutil/marshal.go, line 73 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTi7dSkd-8zyOhisnq:-KvTkpPW5q0SXkX3g3-8:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e86fc630d6dc5e459b4830c706645e85d2119062/pkg/util/protoutil/marshal.go#L73)):*

`s/MarshalTo/MarshalToWithoutFuzzing/g`
</blockquote></details>

Done.

---

*[pkg/util/protoutil/randnullability.go, line 102 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848#-KvTif7P5J07kDFaY80d:-KvTl-nHz5RQ8KP6itNY:b79glk1) ([raw file](https://github.com/cockroachdb/cockroach/blob/e86fc630d6dc5e459b4830c706645e85d2119062/pkg/util/protoutil/randnullability.go#L102)):*

This check is no longer necessary, right?
</blockquote></details>

Yep, removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18848)*

PR #18858, opened 2017-09-28
sql: inform non-retryable transactions early

### LINKED ISSUE #16678: storage: inform non-retryable transactions early that they can't commit once they've been pushed
We have this policy of allowing transactions to continue even when we know that they won't be able to commit, because we want them to lay down their intents in the hope that, after a restart, the transaction will write to the same keys. Since intents have been laid down early, those intents will keep other transactions away, so that our original txn is more likely to commit on its 2nd attempt.
This happens at least in two cases: when a txn has its timestamp bumped by having one of its writes run into the timestamp cache and on a WriteTooOldError.

This mechanism is cool, but not helpful for transactions that have no intention to retry - e.g. SQL explicit txns that didn't use the `savepoint cockroach_restart` statement. These guys would benefit from being told early that they can't commit. And also it'd make debugging them easier, as the conflict causing them to fail would be more easily apparent.

a6802739:
Fixes #16678, @andreimatei, PLAT. Thank you very much.

### DIFF
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -1607,8 +1607,19 @@ func (e *Executor) execStmtInOpenTxn(
 		if err != nil {
 			return
 		}
-
-		if txnState.State() == AutoRetry && txnState.isSerializableRestart() {
+		// Generally, if the transaction is serializable and it has been pushed,
+		// we don't deal with it until commit time (when we return a retriable error).
+		// This is because we want the transaction to continue and lay down all the intents it wants,
+		// so that a future attempt is more likely to succeed.
+		// However, there are two situations when we do want to detect the push and restart early:
+		//
+		// 1. If we can still automatically retry the txn, we go ahead and retry now -
+		// if we'd execute further statements, we probably wouldn't be allowed to retry automatically any more.
+		//
+		// 2. If the client is not following the client-directed retries protocol,
+		// then technically there won't be any retries,
+		// so the client would not benefit from letting the transaction run more statements.
+		if txnState.isSerializableRestart() && (txnState.State() == AutoRetry || !txnState.retryIntent) {
 			// If we just ran a statement synchronously, then the parallel queue
 			// would have been synchronized first, so this would be a no-op.
 			// However, if we just ran a statement asynchronously, then the
@@ -1629,15 +1640,8 @@ func (e *Executor) execStmtInOpenTxn(
 					return
 				}
 			}
-
-			// If we can still automatically retry the txn, and we detect that the
-			// transaction won't be allowed to commit because its timestamp has been
-			// pushed, we go ahead and retry now - if we'd execute further statements,
-			// we probably wouldn't be allowed to retry automatically any more.
 			txnState.mu.txn.Proto().Restart(
 				0 /* userPriority */, 0 /* upgradePriority */, e.cfg.Clock.Now())
-			// Force an auto-retry by returning a retryable error to the higher
-			// levels.
 			err = roachpb.NewHandledRetryableTxnError(
 				"serializable transaction timestamp pushed (detected by SQL Executor)",
 				txnState.mu.txn.ID(),
--- a/pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
+++ b/pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
@@ -13,6 +13,10 @@ GRANT ALL on t TO testuser
 statement ok
 BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE
 
+# Without client-directed retries, the serializability violation would be detected early.
+statement ok
+SAVEPOINT cockroach_restart
+
 # The SELECT forces the timestamp to be chosen.
 query I
 SELECT * FROM t
@@ -33,9 +37,13 @@ user root
 statement ok
 INSERT INTO t(a) VALUES (1)
 
-# Commit will send an EndTransaction which will result in a retry error.
+# RELEASE will send an EndTransaction which will result in a retriable error.
 statement error pgcode 40001 retry txn.*
-COMMIT
+RELEASE SAVEPOINT cockroach_restart
+
+#ROLLBACK TO SAVEPOINT
+statement OK
+ROLLBACK TO SAVEPOINT cockroach_restart
 
 # Check that the connection is usable after the failed COMMIT.
 query I
--- a/pkg/sql/txn_restart_test.go
+++ b/pkg/sql/txn_restart_test.go
@@ -1676,13 +1676,11 @@ CREATE TABLE t.test (k INT PRIMARY KEY);
 		t.Fatal(err)
 	}
 
-	// Two tests, one where the txn is pushed while the transaction can still be
-	// retried, one where it's pushed too late for auto-retries.
-	// moveOutOfAutoRetry dictates whether we move the transaction from state
-	// AutoRetry to Open before starting the batch of statements that will
-	// encounter the push. When moveOutOfAutoRetry is set, we expect to get a
-	// retriable error at commit time. When it's not, we expect to see no error
-	// because the transaction is retried automatically.
+	// Two tests, one where the txn is pushed while the transaction can still be retried,
+	// one where it's pushed too late for auto-retries. moveOutOfAutoRetry dictates whether
+	// we move the transaction from state AutoRetry to Open before starting the batch of statements
+	// that will encounter the push. When moveOutOfAutoRetry is set, we expect to get a retriable error.
+	// When it's not, we expect to see no error because the transaction is retried automatically.
 	for i, moveOutOfAutoRetry := range []bool{false, true} {
 		t.Run(fmt.Sprintf("%t", moveOutOfAutoRetry),
 			func(t *testing.T) {
@@ -1709,20 +1707,22 @@ CREATE TABLE t.test (k INT PRIMARY KEY);
 				//   - a marker statement recognized by the filter.
 				//   - a write that will conflict with the read done by the filter and
 				//     will cause the txn's timestamp to be pushed.
-				if _, err := tx.Exec(
+				_, err = tx.Exec(
 					fmt.Sprintf(
 						`SHOW DATABASES; SELECT '%s'; INSERT INTO t.test VALUES (%d)`,
-						marker, i)); err != nil {
-					t.Fatal(err)
-				}
-
-				err = tx.Commit()
+						marker, i))
 				if moveOutOfAutoRetry {
 					if !isRetryableErr(err) {
 						t.Fatalf("expected retryable error, got: %v", err)
 					}
+					if err := tx.Rollback(); err != nil {
+						t.Fatal(err)
+					}
 				} else {
 					if err != nil {
+						t.Fatal(err)
+					}
+					if err := tx.Commit(); err != nil {
 						t.Fatal(err)
 					}
 				}


knz:
Thank you for your contribution!

The code looks good to me (Andrei will check though),  however please extend the commit message to explain your motivation and reasoning for this change.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-:-KvIipEuIlntCZ0ua0AT:b-haajey)*

andreimatei:
+1 - the commit title and description need more words.

---

---

*[pkg/sql/executor.go, line 1603 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-KvSnU4LMeen0i4EJbIj:-KvSnU4LMeen0i4EJbIk:b66hhpn) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b5cc15868022cf0e6392f376e269b01faa99a4e/pkg/sql/executor.go#L1603)):*
> ```Go
> 				roachpb.Transaction{},
> 			)
> 			err = txnState.updateStateAndCleanupOnErr(err, e)
> ```

I think the code would benefit from uniting this new branch with the previous one; I think they need to do exactly the same thing.

---

*[pkg/sql/executor.go, line 1604 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-KvSlTZr0HTRe1OluIOn:-KvSlTZr0HTRe1OluIOo:bwt3u23) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b5cc15868022cf0e6392f376e269b01faa99a4e/pkg/sql/executor.go#L1604)):*
> ```Go
> 			)
> 			err = txnState.updateStateAndCleanupOnErr(err, e)
> 		} else if txnState.State() != AutoRetry && !txnState.retryIntent && txnState.isSerializableRestart() {
> ```

the `txnState.State() != AutoRetry` condition is superfluous, right? If `isSerializableRestart()` is true, than `txnState.State() != AutoRetry` is known (otherwise the previous branch would have been taken). I'd remove because I'd find the flow easier to read without it.

Also nit: make the `txnState.isSerializableRestart()` condition first, as that's the important one that hints to what the code is about to do. Feel free to do the same in the above branch (I probably should have written it this way, although it's less important there).

---

*[pkg/sql/executor.go, line 1604 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-KvSn105FjRqL4tWl8d0:-KvSn105FjRqL4tWl8d1:b-dqzvo4) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b5cc15868022cf0e6392f376e269b01faa99a4e/pkg/sql/executor.go#L1604)):*
> ```Go
> 			)
> 			err = txnState.updateStateAndCleanupOnErr(err, e)
> 		} else if txnState.State() != AutoRetry && !txnState.retryIntent && txnState.isSerializableRestart() {
> ```

This branch needs a comment as to what it's doing and why. The branch above has one (starts with "If we can still automatically retry the txn"). That comment got pushed by code changes below the point where it was supposed to be - at the beginning of the branch :)

---

*[pkg/sql/executor.go, line 1605 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-KvSmlxIzMGS8x53YOw9:-KvSmlxIzMGS8x53YOwA:b-ak6u11) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b5cc15868022cf0e6392f376e269b01faa99a4e/pkg/sql/executor.go#L1605)):*
> ```Go
> 			err = txnState.updateStateAndCleanupOnErr(err, e)
> 		} else if txnState.State() != AutoRetry && !txnState.retryIntent && txnState.isSerializableRestart() {
> 			err = roachpb.NewHandledRetryableTxnError(
> ```

I think we need to drain the parallel queue like in the branch above (the comment as to why that's needed seems to apply equally here too, no?).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-:-KvSkLccUbL5TXXqb6eL:ba3xoi5)*

andreimatei:
Do you know what's going on with the test failures?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-:-KvSo1zJ5_yDyH8VEX5Z:bapo2w2)*

a6802739:
@andreimatei @knz , Thank you for your review, I think the test failure is `TestPushedTxnDetection`. I have added the description in the commit message.

I couldn't login in `Reviewable` for now, don't know why.

Please help me have a review again.

andreimatei:
I suggest the following commit title and message:

```
sql: inform non-retriable transactions early of serializability errors

Fixes #16678 

We have a policy of allowing transactions to continue even when we know that they won't be able to commit, because we want them to lay down their intents in the hope that, after a restart, the transaction will write to the same keys. Since intents have been laid down early, those intents will keep other transactions away, so that our original txn is more likely to commit on its 2nd attempt. This happens at least in two cases: when a txn has its timestamp bumped by having one of its writes run into the timestamp cache and on a WriteTooOldError.
This mechanism is not helpful for transactions that have no intention to retry - e.g. SQL explicit txns that didn't use the `SAVEPOINT cockroach_restart` statement. These guys would benefit from being told early that they can't commit, and also it'd make debugging them easier, as the conflict causing them to fail would be more easily apparent.

This patch detects serializability violations early for transactions that are not going to be retried, extending the detection we already had for auto-retry transactions. When a violation is detected and `txnState.retryIntent` is not set, the client gets the retriable error as soon as a statements detects the condition.
```

---

Review status: 0 of 3 files reviewed at latest revision, 12 unresolved discussions, all commit checks successful.

---

*[pkg/sql/executor.go, line 1607 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1OdV6YtMS4PrPSQGD:-Kw1OdV6YtMS4PrPSQGE:bvxcveh) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/executor.go#L1607)):*
> ```Go
> 			return
> 		}
> 		if txnState.isSerializableRestart() {
> ```

something's wrong here... We *only* want to return the retriable error `if txnState.isSerializableRestart() && (txnState.State() == AutoRetry || !txnState.retryIntent)`. As written, you're returning the error regardless of `txnState.State()` and `txnState.retryIntent`.

---

*[pkg/sql/executor.go, line 1631 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1QUi_o6JIVE8_IYf6:-Kw1QUi_o6JIVE8_IYf7:b-a4n4vu) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/executor.go#L1631)):*
> ```Go
> 			}
> 			if txnState.State() == AutoRetry {
> 				// If we can still automatically retry the txn, and we detect that the
> ```

please move this comment to just above the `if txnState.isSerializableRestart() ` block, and augment it with the new case we're handling. Say that we want to propagate the error if either:
a) we're in AutoRetry, for the reason described here or,
b) if !retryIntent, because the client would not benefit from letting the transaction run more statements.

---

*[pkg/sql/executor.go, line 1635 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1Q1WtEx5U0eK4xnYt:-Kw1O8b3Jb5f3PMXjgZ8:b-3f3nqu) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/executor.go#L1635)):*
> ```Go
> 				// pushed, we go ahead and retry now - if we'd execute further statements,
> 				// we probably wouldn't be allowed to retry automatically any more.
> 				txnState.mu.txn.Proto().Restart(
> ```

Although not strictly needed, it also wouldn't hurt to always do this `Restart()` call (if `txnState.State() != AutoRetry`, then it won't matter since the `updateStateAndCleanupOnErr()` call below will abort the txn.
I think doing always would simplify the code, so I'd remove the `if`.

---

*[pkg/sql/txn_restart_test.go, line 1682 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1TtNlsDtqmdyJOCZM:-Kw1TtNlsDtqmdyJOCZN:bkgb1p3) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/txn_restart_test.go#L1682)):*
> ```Go
> 	// moveOutOfAutoRetry dictates whether we move the transaction from state
> 	// AutoRetry to Open before starting the batch of statements that will
> 	// encounter the push. When moveOutOfAutoRetry is set, we expect to get a
> ```

this comment is now stale

---

*[pkg/sql/txn_restart_test.go, line 1719 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1ScVKScVXcqPcdTNq:-Kw1ScVKScVXcqPcdTNr:b-y3yoe4) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/txn_restart_test.go#L1719)):*
> ```Go
> 						t.Fatalf("expected retryable error, got: %v", err)
> 					}
> 					if err := tx.Rollback(); err != nil {
> ```

what's going on here? Why did you make the `tx.Commit()` conditional?

I think the way in which you want to modify this test 

---

*[pkg/sql/txn_restart_test.go, line 1734 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1UDNBH2O00VVsA1qy:-Kw1UDNBH2O00VVsA1qz:b-tqywq4) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/txn_restart_test.go#L1734)):*
> ```Go
> }
> 
> func TestPushedTxnRetriableError(t *testing.T) {
> ```

please add a comment about what this test is trying to test and how it's testing it.

---

*[pkg/sql/txn_restart_test.go, line 1743 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1U797Ebfbx06ajBhM:-Kw1U797Ebfbx06ajBhN:binw671) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/txn_restart_test.go#L1743)):*
> ```Go
> 	defer cleanup()
> 
> 	db, err := gosql.Open("postgres", pgURL.String())
> ```

why did you need to create this `db`, as opposed to just using `sqlDB`? You can call `Begin` twice on a `sql.DB` to get two different transactions/connections.

---

*[pkg/sql/logictest/testdata/logic_test/serializable_eager_restart, line 31 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kw1RNHgTALVYNEs5FSd:-Kw1RNHgTALVYNEs5FSe:b-5b0ea9) ([raw file](https://github.com/cockroachdb/cockroach/blob/73e8141f3037d3a976843f135c1ab995edbedcc7/pkg/sql/logictest/testdata/logic_test/serializable_eager_restart#L31)):*
> ```
> user root
> 
> # The insert increases the candidate timestamp, and it will encounter a restart error.
> ```

I don't think you want to change this test the way you're doing it. This test is trying to test precisely that the transaction restart *late* - see the beginning of the file.
What you want to do is make the condition you're introducing in this patch not fire for this test. So, add a `SAVEPOINT cockroach_restart` to the beginning of the transaction (and also change the COMMIT below to a `RELEASE SAVEPOINT cockroach_restart`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-:-Kw1MOv9YSmztOgacilv:b-lyaifj)*

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
```diff
@@ -29,13 +32,17 @@ SELECT * FROM t
...
-# until an EndTransaction.
+# until RELEASE SAVEPOINT.
 statement ok
 INSERT INTO t(a) VALUES (1)
 
-# Commit will send an EndTransaction which will result in a retry error.
+# RELEASE SAVEPOINT.
 statement error pgcode 40001 retry txn.*
-COMMIT
+RELEASE SAVEPOINT cockroach_restart
```
a6802739: I found in `TestLogic/default/serializable_eager_restart` and `TestLogic/distsql/serializable_eager_restart`, it reports error like `expected "retry txn.*", but found <nil>`.

Actually, it seems it will only in parallel condition, it will return an retriable error.

pkg/sql/txn_restart_test.go
```diff
@@ -1708,27 +1708,76 @@ CREATE TABLE t.test (k INT PRIMARY KEY);
...
-					t.Fatal(err)
-				}
-
-				err = tx.Commit()
+						marker, i))
 				if moveOutOfAutoRetry {
 					if !isRetryableErr(err) {
 						t.Fatalf("expected retryable error, got: %v", err)
 					}
+					if err := tx.Rollback(); err != nil {
```
a6802739: My understanding is that this txn couldn't commit now, so I just Rollback it.

Sorry, I couldn't quite understand what do you mean here.

pkg/sql/txn_restart_test.go
```diff
@@ -1708,27 +1708,76 @@ CREATE TABLE t.test (k INT PRIMARY KEY);
...
 				}
 			})
 	}
 }
 
+// TestPushedTxnRetriableError is just used to verify we should return an retry error
+// early to the client for explicit txn whose timestamp has been pushed by other txns.
+// We start two txns in this test, txn1 did a write first, then txn2 did a read, when txn1 try to
+// update the record that txn2 has read, it's timestamp will be pushed forward.
+// So txn1 will get a retriable error immediately.
```
a6802739: I have add a comment here about this test.

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
a6802739: I found in `TestLogic/default/serializable_eager_restart` and `TestLogic/distsql/serializable_eager_restart`, it reports error like `expected "retry txn.*", but found <nil>`.

Actually, it seems only in parallel condition, it will return an retriable error.

a6802739 (commented):
@andreimatei , Thanks for you review. I couldn't login in `Reviewable` right now.

pkg/sql/executor.go
```diff
@@ -1604,45 +1604,50 @@ func (e *Executor) execStmtInOpenTxn(
...
-				if _, ok := parErr.(*roachpb.HandledRetryableTxnError); !ok {
-					err = parErr
-					return
+		// If we can still automatically retry the txn, and we detect that the
+		// transaction won't be allowed to commit because its timestamp has been
+		// pushed, we go ahead and retry now - if we'd execute further statements,
+		// we probably wouldn't be allowed to retry automatically any more.
+		// We want to propagate the error if either:
+		// 1.we're in AutoRetry, for the reason described above.
+		// 2.retryIntent is not set, because the client would not benefit from letting the transaction run more statements.
```
andreimatei: line is over 100 chars

The comment is also badly structured; the two paragraphs are not well integrated. I suggest:
```
Generally, if the transaction is serializable and it has been pushed, we don't deal with it until commit time (when we return a retriable error). This is because we want the transaction to continue and lay down all the intents it wants, so that a future attempt is more likely to succeed. However, there are two situations when we do want to detect the push and restart early:
1. If we can still automatically retry the txn, we go ahead and retry now - if we'd execute further statements, we probably wouldn't be allowed to retry automatically any more.
2. If the client is not following the client-directed retries protocol, then technically there won't be any retries, so the client would not benefit from letting the transaction run more statements.
```

pkg/sql/executor.go
```diff
@@ -1604,45 +1604,50 @@ func (e *Executor) execStmtInOpenTxn(
...
-				"serializable transaction timestamp pushed (detected by SQL Executor)",
-				txnState.mu.txn.ID(),
-				// No updated transaction required; we've already manually updated our
-				// client.Txn.
-				roachpb.Transaction{},
-			)
-			err = txnState.updateStateAndCleanupOnErr(err, e)
+			// if we already knew the serializable transaction has been pushed by
+			// another transaction, we should return an retriable error to the higher level.
+			if txnState.State() == AutoRetry || !txnState.retryIntent {
```
andreimatei: looks like you've tried to address my comment about only returning the error sometimes, but the code is still wrong. You also don't want to run the `txnState.mu.txn.Proto().Restart()` above unless we're going to return this error.
The whole big block that starts with `if txnState.isSerializableRestart()` should start with `if txnState.isSerializableRestart() && (AutoRetry || !retryIntent)`.

pkg/sql/txn_restart_test.go
```diff
@@ -1680,7 +1680,7 @@ CREATE TABLE t.test (k INT PRIMARY KEY);
 	// moveOutOfAutoRetry dictates whether we move the transaction from state
 	// AutoRetry to Open before starting the batch of statements that will
 	// encounter the push. When moveOutOfAutoRetry is set, we expect to get a
-	// retriable error at commit time. When it's not, we expect to see no error
+	// retriable error. When it's not, we expect to see no error
```
andreimatei: nit: please re-wrap comments at 80 columns

pkg/sql/txn_restart_test.go
andreimatei: This looks good. Never mind what I was saying.

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
```diff
@@ -29,13 +32,17 @@ SELECT * FROM t
 user root
 
 # The insert increases the candidate timestamp, but a restart won't occur
-# until an EndTransaction.
+# until RELEASE SAVEPOINT.
```
andreimatei: The `EndTransaction` comment was good - it talks about when the error is detected by crdb at the kv layer. It is detected when an `EndTransaction` request is sent. The `EndTransaction` will indeed be sent by a `RELEASE SAVEPOINT`, but that doesn't need to be explained necessarily.
I'd revert this line.

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
andreimatei: That is unexpected, right? Can you debug? You can run just a single logic test with something like (change the test names):
```
make testshort PKG=./pkg/sql/logictest  TESTTIMEOUT=2m TESTFLAGS="-v -show-logs -show-sql --vmodule=executor=2" TESTS=TestLogic/distsql-disk/postgresjoin
```

If you fix the server code, I can look too.

andreimatei (changes requested):

a6802739:
@andreimatei  Thank you very much. I will try to debug it. Thank you for your kind review.

andreimatei:
I'd start by fixing the server code. It might be that the extra `Proto.Restart()` in there is causing that test failure.

pkg/sql/txn_restart_test.go
a6802739: @andreimatei, sorry, I don't understand what do you mean here?

pkg/sql/executor.go
a6802739: Sorry, I just couldn't understand why we should run the `txnState.mu.txn.Proto().Restart()` when `txnState.isSerializableRestart() && !retryIntent`?

a6802739:
@andreimatei, Yeah, you are right. I have fixed it now.

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
```diff
@@ -33,9 +36,13 @@ user root
 statement ok
 INSERT INTO t(a) VALUES (1)
 
-# Commit will send an EndTransaction which will result in a retry error.
+# RELEASE SAVEPOINT.
```
andreimatei: change to `# RELEASE will send an EndTransaction which will result in a retriable error.`

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 19 unresolved discussions, all commit checks successful.

---

*[pkg/sql/executor.go, line 1641 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-Kv6HVAxi2Rj-FJ5ZmlW-r3-1641:-KwAyYH8semNHWbLg0uH:byfix6f) ([raw file](https://github.com/cockroachdb/cockroach/blob/43ab7e338037ad99840f57dceef166941481720a/pkg/sql/executor.go#L1641)):*
> Sorry, I just couldn't understand why we should run the txnState.mu.txn.Proto().Restart() when txnState.isSerializableRestart() && !retryIntent?

It doesn't matter whether we run `txnState.mu.txn.Proto().Restart()` when `!retryIntent`. It does, however, matter that we *do not* run `txnState.mu.txn.Proto().Restart()` when we're not going to return the `HandledRetryableTxnError`. The code you had written when I made this comment would run `txnState.mu.txn.Proto().Restart()` when it shouldn't have.
All clear?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858)*

pkg/sql/txn_restart_test.go
andreimatei: Isn't this test testing the same thing as the one above, in the `moveOutOfAutRetry=true` case? I don't think it adds anything; I would remove it.

andreimatei (changes requested):

pkg/sql/executor.go
```diff
@@ -1626,15 +1637,10 @@ func (e *Executor) execStmtInOpenTxn(
...
-
-			// If we can still automatically retry the txn, and we detect that the
-			// transaction won't be allowed to commit because its timestamp has been
-			// pushed, we go ahead and retry now - if we'd execute further statements,
-			// we probably wouldn't be allowed to retry automatically any more.
 			txnState.mu.txn.Proto().Restart(
 				0 /* userPriority */, 0 /* upgradePriority */, e.cfg.Clock.Now())
-			// Force an auto-retry by returning a retryable error to the higher
-			// levels.
+			// if we already knew the serializable transaction has been pushed by
```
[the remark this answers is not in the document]
a6802739: Done.

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
```diff
@@ -13,6 +13,9 @@ GRANT ALL on t TO testuser
 statement ok
 BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE
 
+statement ok
```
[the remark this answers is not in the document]
a6802739: Done.

pkg/sql/logictest/testdata/logic_test/serializable_eager_restart
a6802739: Done.

a6802739:
@andreimatei, about `txnState.mu.txn.Proto().Restart()`, if we are not going to return `HandledRetryableTxnError`, we shouldn't run it, right?

Thank you for your kind explanation.

andreimatei:
> about txnState.mu.txn.Proto().Restart(), if we are not going to return HandledRetryableTxnError, we shouldn't run it, right?

correct

---

Review status: 0 of 3 files reviewed at latest revision, 19 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18858#-:-KwGKeM8XDMjDpPlJbjF:b-txf6zo)*

andreimatei:
@a6802739 I think this is ready to merge, right?

a6802739:
yeah, l think so.

2017年10月17日星期二，Andrei Matei <notifications@github.com> 写道：

PR #18804, opened 2017-09-26
storage: have serializable restarts take precedence over deadline expiration

andreimatei:
Before this patch, if a Serializable txn was pushed and the push caused
the txn to exceed its deadline, the client would get a "deadline
exceeded" error back. What the client should have gotten is the
retriable error, as that would allow it to retry. This patch fixes this
by inverting the order of error generation. As a result, Serializable
txns shouldn't see "deadline exceeded" errors any more.

Touches #18684

Issue #18684, opened 2017-09-21
sql: TransactionStatusError: transaction deadline exceeded


jordanlewis:
When running TPCC with more than a single thread, occasionally a transaction will fail with the following error:

` pq: TransactionStatusError: transaction deadline exceeded`

What is this about? What is the referenced deadline? It seems like a bug that this message is propagated to the user, but what internal condition is causing this scenario and why?

Any ideas? cc @andreimatei @bdarnell

bdarnell:
I believe that message means a table descriptor lease has expired. cc @vivekmenezes @lego

lego:
@jordanlewis how frequent is the error? If it is frequent enough it will hopefully be easy to test whether the lease expiring and re-acquisition is the issue since I have an in-flight PR right now (relevant PR #18606, issue #17227).

That is peculiar that it happens specifically with more than one thread. There are a few places in the lease acquisition where I can see race conditions causing starvation but only if the lease is immediately released, or _maybe_ under heavy load. This could go either two ways: the above PR fixes it because lease acquisition is taking too long, or it's a race-condition causing a thread to be starved. I would think the former would happen with also one thread though, so it's less likely to be that.

vivekmenezes:
This bug is happening because we are using a lease very close to its expiration deadline (we changed that recently to do that in https://github.com/cockroachdb/cockroach/pull/16842 by taking out the notion of MinLeaseDuration). I suspect we are seeing this here because a transaction with SNAPSHOT ISOLATION is using a table descriptor with an unexpired lease but is having its COMMIT pushed beyond the lease expiration (which is what the txn expiration is set to).

lego's PR will fix this because it will reacquire a lease way before it expires and thus transactions will use leases with expirations further out into the future (like a minute).

jordanlewis:
All transactions in TPCC are set to Serializable isolation.

vivekmenezes:
if you can log the txn.OrigTimestamp() and expiration in getTableVersion() after the call to AcquireByName() and panic instead of returning the error with the deadline timestamp and the txn timestamp we might get a better clue. Thanks!

vivekmenezes:
of course there is also the possibility that we should be using OrigTimestamp() when acquiring leases. I think I need to ask around as to what API to use to get the COMMIT timestamp for a transaction as it's running

vivekmenezes:
[...]

PR #18859, opened 2017-09-28
Revert "sql: Refresh table leases asynchronously on access"

### LINKED ISSUE #17227: sql: refresh table leases asynchronously

lgo:
This reverts commit ec9f73b.
The commit was attempting to fix a lease deadline bug, #18684,
and close #17227 by not blocking on table descriptor lease acquisition.
The fix prevents the first issue, but does not actually resolve the
second issue. There was also a flaky test added which will be fixed with the
revert, #18827.

Even after this commits revert, #18804 fixes the first bug for SERIALIZABLE
transactions.

A replacement patch is in-progress at #18844.

### DIFF
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -41,23 +41,14 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 )
 
+// TODO(pmattis): Periodically renew leases for tables that were used recently and
+// for which the lease will expire soon.
+
 var (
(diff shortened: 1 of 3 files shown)


tschottdorf:
Please give rationale :-)

vivekmenezes (approved):
Add a rationale in the commit message and merge. Thanks

vivekmenezes (commented):
Also reference the issues it was fixing

lego:
Updated the commit message! :)

vivekmenezes:
LGTM

Issue #18665, opened 2017-09-21
ui: Consider upgrade from D3 v3 to v4 (or v5!)


mrtracy:
The Admin UI currently uses D3 version 3.  Version 4 of D3 is a non-trivial upgrade, but claims a number of benefits:

+ Modularity (import only the features needed)
+ More consistent parameters to callback functions
+ More in-depth typescript support

Our usage of v3 is not currently crippling, but it might be better to upgrade sooner rather than later.

vilterp:
Welp, now there's v5: https://github.com/d3/d3/releases/tag/v5.0.0

PR #18865, opened 2017-09-28
[WIP] ui: replace NVD3 with vanilla D3

### LINKED ISSUE #18620: ui: Render graphs with D3 directly, replacing use of NVD3
We have used the NVD3 library to draw our metrics graphs since the creation of the Admin UI.  However, we are rapidly outgrowing the restraints of this system:

 + We have numerous graph-related bugs that are proving difficult to analyze. As the graphing logic is largely contained in this third party library (which is not exceptionally well documented), we do not have a sufficient intuition of how to analyze or fix these bugs.
+ We have a desire to add more advanced features to graphs (#13378, #13752) which require features that do not intuitively fit into the NVD3 framework.

NVD3 is a library meant to provide a reasonable level of graph utility for a wide array of use cases; we now need a more advanced level of graph utility for a narrower use case.  Therefore, we should begin rendering our graphs using D3 directly; this will afford us greater control and understanding of the rendering process, allowing us to debug issues more effectively and add advanced features more easily.

---

EDIT 2019: Related issues.

Features this will directly or indirectly enable:
- #5980 - "step chart"
- #10014 - "log y scale"
- #12899 - "second axis"
- #13378 - "events on graphs"
- #13389 - "histogram on hover"
- #13752 - "zoom on time scale"
- #16855 - "show configured cache size"
- #20021 - "ghost time series"
- #27934 - "boxplot for distribution"

Bugs this will likely obselete (let's not repeat them, though):
- #12973
- #16505 
- #17552
- #18452
- #19271
- #19490
- #21630
- #22572
- #23480
- #27094
- #27954
- #30286
- #30287
- #33717

couchand:
Obviously not even close to ready to merge.  Putting it up here to get people's eyes on it/work on this effort out in the open.

will close #18620, and incidentally #18665, too.

### DIFF
--- a/pkg/ui/src/js/charts.js
+++ b/pkg/ui/src/js/charts.js
@@ -65,17 +65,11 @@ function renderChart(config) {
   x.domain([d3.min(xExtents, function (d) { return d[0]; }), d3.max(xExtents, function (d) { return d[1]; })]);
   y.domain([d3.min(yExtents, function (d) { return d[0]; }), d3.max(yExtents, function (d) { return d[1]; })]);
 
-  var legend = chart.append("g")
-    .attr("class", "nv-legendWrap")
-    .attr("transform", "translate(0,-30)")
-    .append("g")
-    .attr("class", "nvd3 nv-legend")
-    .attr("transform", "translate(0,5)")
-    .append("g")
-    .attr("transform", "translate(" + (width - config.metrics.length * 70) + ",5)");
+  var legend = chart.selectAll("g.nv-legend")
+    .attr("transform", "translate(" + (width - config.metrics.length * 70) + ",10)");
 
   var legendsBase = legend.selectAll("g.nv-series")
-    .data(config.metrics);
+    .data(config.metrics.length === 1 ? [] : config.metrics);
 
   var legendsNew = legendsBase.enter()
     .append("g")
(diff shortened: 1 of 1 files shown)


cuongdo:
This has gone 2 weeks without feedback. @mrtracy and @vilterp: could you please provide feedback?

mrtracy:
Although i'm not sure what stage of development you are in here, I have a few high level comments:

1. This still looks like a Javascript/jQuery plugin, not a Typescript/React component.  I have been assuming you've been writing this style for rapid prototyping, but just in case, this eventually should be converted to Typescript, and exports should be managed in ES6 style (i.e. use `exports` instead of `window.charts`).

2. From a high-level design, this component is rendering an array of charts, not a single chart.  I understand where you are coming from; this is a solution for rendering the line across multiple components.  However, I really think this is escaping into d3 rendering far too early.  If anything, I think we should be moving some components *out* of the d3 rendering pipeline (the legend immediately comes to mind).  We should constrain the d3 rendering to only the components where d3 has a substantial advantage over JSX for that purpose (anything that needs to render along an axis).

In order to meet that constraint, I think the component must focus on rendering the visual core of a single chart, leaving elements like the title and surrounding div to be rendered in JSX.  If we need to coordinate something between charts, we should do that at the React level; although, I actually think the current method of rendering the line across multiple charts is fine.

couchand:
Yep, wrapping in a React component and adding typings as appropriate remain todo items (you can track the progress at #18620, though I notice now that doesn't mention types anywhere).

I agree that it probably makes sense to make the D3 component just render a single chart, as that's much easier to follow.  I didn't write it this way to support the linked guideline (I intend that to be managed by the Redux state), but rather because it was much quicker to write this by relying on only a single dependency.

I don't think it makes sense to have the legend rendered separately, for a few reasons.  Most significantly that would require more work to coordinate colors and selection states, and generally to keep the series in sync between the legend and the lines on the chart.

I also don't think the current method of rendering the linked guideline is sufficient.  Particularly, when we go to implement event annotations, we explicitly need to know the time being hovered, not just the pixel position.

couchand:
Ok, though pieces of this code will live on, this PR is too old and not adding any value at this point.

PR #18867, opened 2017-09-28
cherrypick-1.1: util/log: parse logs assuming UTC

dt:
72d005482 switched `formatLogEntry` to use `timeutil.Unix` and thus changed the `time.Time` passed to `formatHeader` to have a UTC location, meaning the log timestamps switched to being printed in UTC rather than the host's local TZ.

This is actually likely desireable in a distributed system, as it makes correlating events across nodes in a geographically diverse cluster easier, but it broke our parsing of the logs, which assumed local.

We probably shouldn't assume any particular offset anyway when reading a log file that might have been written with a different offset. We could add explicit offsets, to avoid assuming anything, but as mentioned above, just switching to UTC everywhere has other benefits too.# Please enter the commit message for your changes. Lines starting

### DIFF
--- a/build/style_test.go
+++ b/build/style_test.go
@@ -222,7 +222,7 @@ func TestStyle(t *testing.T) {
 
 		if err := stream.ForEach(stream.Sequence(
 			filter,
-			stream.GrepNot(`^util/(log|syncutil|timeutil|tracing)/\w+\.go\b`),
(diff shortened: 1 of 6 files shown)


dt:
er, might need another one in here, waiting for now.

dt:
okay, added #18868 too.

petermattis (approved):
LGTM

PR #18868, opened 2017-09-28
util/log: use UTC in log tests too, remove linter exceptions

dt:
We had previously whitelisted a few logging-related files from being required to use timeutil.
Requiring them to do so as well ensures they too use the helpers that return UTC rather than Local.


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18868#-:-Kv8eHSCYfkKTI5ch6Uh:bnfp4nl)*

PR #18891, opened 2017-09-28
storage: add TestProposalOverhead

### LINKED ISSUE #18738: storage: test that we don't regress on the space overhead of proposals
From [a comment](https://github.com/cockroachdb/cockroach/pull/18689#issuecomment-331906918) on #18689:

> Do you think we adequately protect against accidental re-blowup here in future changes? I assume those will be rare enough and scrutinized anyway, but just curious if there would be enough test breakage for the developer to realize that it's a really bad idea to make these structs any bigger than they need to be.

Perhaps something as simple as looking at the encoded size of the zero value of `storagebase.RaftCommand` would be sufficient.

petermattis:
Rework StoreConfig.TestingProposalFilter to take a ProposalFilterArgs
which contains the actual proposal.

Fixes #18738

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -2918,17 +2918,18 @@ func (r *Replica) propose(
 	// TODO(irfansharif): This int cast indicates that if someone configures a
 	// very large max proposal size, there is weird overflow behavior and it
 	// will not work the way it should.
-	if proposal.command.Size() > int(MaxCommandSize.Get(&r.store.cfg.Settings.SV)) {
+	proposalSize := proposal.command.Size()
+	if proposalSize > int(MaxCommandSize.Get(&r.store.cfg.Settings.SV)) {
 		// Once a command is written to the raft log, it must be loaded
 		// into memory and replayed on all replicas. If a command is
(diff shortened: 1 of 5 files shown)


tschottdorf:
:lgtm:

Mildly worried that it might be flaky since there are timestamps in the proposal.

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18891#-:-Kv91ZqvCCmO6UcO1gs5:bvetvyu)*

petermattis:
`hlc.Timestamp.WallTime` encodes to 9 bytes which is the maximum size for an `int64` field. The timestamps will not be getting any larger. It is somewhat unfortunate that we're not using `sfixed64` for `hlc.Timestamp.WallTime` as it would be both smaller and faster to encode/decode. We use timestamps all over the place, but changing this now would be quite difficult.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18891#-:-Kv9JylNJF2cXL9sRRN0:bncf769)*

PR #18893, opened 2017-09-28
sql: fix handling of errors on Commit and Rollback 

### LINKED ISSUE #18853: sql, storage: errors on ROLLBACK lead to panics
A client gave us a panic:

```
E170927 14:50:01.485045 1104 util/log/crash_reporting.go:114  [n1] a panic has occurred!
panic: unexpected txnState when cleaning up: NoTxn [recovered]
        panic: panic while executing "ROLLBACK": unexpected txnState when cleaning up: NoTxn [recovered]
        panic: panic while executing "ROLLBACK": unexpected txnState when cleaning up: NoTxn [recovered]
        panic: panic while executing "ROLLBACK": unexpected txnState when cleaning up: NoTxn
goroutine 1104 [running]:
github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).Recover(0xc4205279e0, 0x7fe47bf3fbc0, 0xc4203f4360)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:183 +0xb2
panic(0x1b3b340, 0xc423209380)
        /usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/sql/pgwire.(*v3Conn).serve.func1(0xc422b90300, 0x7fe47bf3fbc0, 0xc422ba4090)
        /go/src/github.com/cockroachdb/cockroach/pkg/sql/pgwire/v3.go:423 +0xa1
panic(0x1b3b340, 0xc423209380)
        /usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/sql.(*Session).maybeRecover(0xc422baf000, 0x1cfe948, 0x9, 0xc422eae392, 0x8)
        /go/src/github.com/cockroachdb/cockroach/pkg/sql/session.go:1346 +0x28d
panic(0x1a7fc40, 0xc42326d500)
        /usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/sql.(*Executor).execStmtInOpenTxn.func1(0xc422a22598, 0xc422baf088, 0xc4206b8000, 0x0, 0xc42295d680, 0xc422baf000)
        /go/src/github.com/cockroachdb/cockroach/pkg/sql/executor.go:1407 +0x1f8
github.com/cockroachdb/cockroach/pkg/sql.(*Executor).execStmtInOpenTxn(0xc4206b8000, 0xc422baf000, 0x2abc2c0, 0x2fc1a60, 0x0, 0x0, 0x0, 0x14e856558a9a3a1e, 0x1d600000001, 0xc42309f260, ...)
        /go/src/github.com/cockroachdb/cockroach/pkg/sql/executor.go:1496 +0xdd8
github.com/cockroachdb/cockroach/pkg/sql.(*Executor).execSingleStatement(0xc4206b8000, 0xc422baf000, 0x2abc2c0, 0x2fc1a60, 0x0, 0x0, 0x0, 0x14e856558a9a3a1e, 0x1d600000001, 0xc42309f260, ...)
```

And a log message.

``` 
sql/executor.go:1651  [client=172.31.7.52:49144,user=root,n1] txn rollback failed: HandledRetryableTxnError: TransactionAbortedError: txn aborted "sql txn" id=3dd8473e key=/Table/54/1/"job_node12_145"/16/0 rw=true pri=0.09419308 iso=SERIALIZABLE stat=PENDING epo=1 ts=1506546759.929383122,0 orig=1506546759.929383122,0 max=1506546757.485196014,0 wto=false rop=false seq=2
```

I think we have a problem in the Executor because an error on Rollback is handled particularly and transitions the state to `NoTxn` in an unexpected place (`RollbackSQLTxn()`). It may be an artifact of the fact that `RollbackSQLTxn()` is called from multiple states.

Separately, it seems funky to me that we can get a retriable error from an `EndTransaction` request.

andreimatei:
Fixes #18853

Ever since we've introduced results streaming, there's been great
ambiguity about the handling of errors that happen on commit and
rollback statements. The execution of these statements didn't use to
return any errors ever, but that has changed because result streaming
introduced the possibility of communication error happening during
statement execution, and these errors can never be swallowed (and so we
also stopped swallowing other errors at the time).
Because COMMIT and ROLLBACK do their own state transitions, the fact
that they started returning an error confused a higher layer
(execStmtInOpenTxn()) which also does state transitions when it sees
errors. That crashed.

This patch resolves the problem by establishing a protocol for state
transitions such that particular statements can coordinate with
execStmtInOpenTxn() around the needed transitions. COMMIT and ROLLBACK
use this mechanism to inform the higher layers that they have sometimes
taken care of all transitions.
Morally, this change is in the spirit of envisioned future changes to
the session state machine: state transitions should/will be centralized
into one level of the code, not sprinkled everywhere.

### DIFF
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -1402,7 +1402,13 @@ func (e *Executor) execStmtInAbortedTxn(
 	switch s := stmt.AST.(type) {
 	case *parser.CommitTransaction, *parser.RollbackTransaction:
 		if txnState.State() == RestartWait {
-			return rollbackSQLTransaction(txnState, res)
+			transition := rollbackSQLTransaction(txnState, res)
+			if transition.transitionDependentOnErrType {
+				log.Fatal(session.Ctx(), "transitionDependentOnErrType should not be "+
+					"returned by rollbackSQLTransaction")
+			}
+			txnState.resetStateAndTxn(transition.targetState)
+			return transition.err
 		}
 		// Reset the state to allow new transactions to start.
 		// The KV txn has already been rolled back when we entered the Aborted state.
@@ -1547,6 +1553,9 @@ func (e *Executor) execStmtInOpenTxn(
 
 	sessionEventf(session, "%s", stmt)
 
+	var explicitStateTransition bool
+	var transition stateTransition
+
 	// Do not double count automatically retried transactions.
 	if automaticRetryCount == 0 {
 		e.updateStmtCounts(stmt)
@@ -1554,7 +1563,28 @@ func (e *Executor) execStmtInOpenTxn(
 
 	// After the statement is executed, we might have to do state transitions.
 	defer func() {
-		if err != nil {
+		// There's two cases to handle, depending on the two categories of
+		// statements we have:
+		// - Some statements (COMMIT, ROLLBACK) explicitly decide what state
+		// transition they want. For them, explicitStateTransition is set and
+		// transition is filled in, and so we obey their instructions.
+		// - Most statements don't set explicitStateTransition. For them we do state
+		// transitions based on the error they returned (if any).
+
+		if explicitStateTransition {
+			if err != nil {
+				panic(fmt.Sprintf("unexpected return err when explicitStateTransition is set: %+v", err))
+			}
+			err = transition.err
+			if transition.transitionDependentOnErrType {
+				if err == nil {
+					panic(fmt.Sprintf("missing err when transitionDependentOnErrType is set"))
+				}
+				err = txnState.updateStateAndCleanupOnErr(err, e)
+			} else {
+				txnState.resetStateAndTxn(transition.targetState)
+			}
+		} else if err != nil {
 			if !txnState.TxnIsOpen() {
 				panic(fmt.Sprintf("unexpected txnState when cleaning up: %v", txnState.State()))
 			}
@@ -1563,9 +1593,17 @@ func (e *Executor) execStmtInOpenTxn(
 			if firstInTxn && isBegin(stmt) {
 				// A failed BEGIN statement that was starting a txn doesn't leave the
 				// txn in the Aborted state; we transition back to NoTxn.
+				//
+				// TODO(andrei): BEGIN should use the explicitStateTransition mechanism.
 				txnState.resetStateAndTxn(NoTxn)
 			}
-		} else if txnState.State() == AutoRetry && txnState.isSerializableRestart() {
+		}
+
+		if err != nil {
+			return
+		}
+
+		if txnState.State() == AutoRetry && txnState.isSerializableRestart() {
 			// If we just ran a statement synchronously, then the parallel queue
 			// would have been synchronized first, so this would be a no-op.
 			// However, if we just ran a statement asynchronously, then the
@@ -1630,7 +1668,9 @@ func (e *Executor) execStmtInOpenTxn(
 	case *parser.CommitTransaction:
 		// CommitTransaction is executed fully here; there's no planNode for it
 		// and a planner is not involved at all.
-		return commitSQLTransaction(txnState, commit, res)
+		transition = commitSQLTransaction(txnState, commit, res)
+		explicitStateTransition = true
+		return nil
 
 	case *parser.ReleaseSavepoint:
 		if err := parser.ValidateRestartCheckpoint(s.Savepoint); err != nil {
@@ -1638,7 +1678,9 @@ func (e *Executor) execStmtInOpenTxn(
 		}
 		// ReleaseSavepoint is executed fully here; there's no planNode for it
 		// and a planner is not involved at all.
-		return commitSQLTransaction(txnState, release, res)
+		transition = commitSQLTransaction(txnState, release, res)
+		explicitStateTransition = true
+		return nil
 
 	case *parser.RollbackTransaction:
 		// Turn off test verification of metadata changes made by the
@@ -1646,7 +1688,9 @@ func (e *Executor) execStmtInOpenTxn(
 		session.testingVerifyMetadataFn = nil
 		// RollbackTransaction is executed fully here; there's no planNode for it
 		// and a planner is not involved at all.
-		return rollbackSQLTransaction(txnState, res)
+		transition = rollbackSQLTransaction(txnState, res)
+		explicitStateTransition = true
+		return nil
 
 	case *parser.Savepoint:
 		if err := parser.ValidateRestartCheckpoint(s.Name); err != nil {
@@ -1789,12 +1833,35 @@ func stmtAllowedInImplicitTxn(stmt Statement) bool {
 	return false
 }
 
+// stateTransition allows statement execution to request state transitions as a
+// side-effect of their execution. Currently only specific statements use this
+// mechanism; the majority rely on the handling of the errors they return to
+// affect state transitions.
+type stateTransition struct {
+	// transitionDependentOnErrType, if set, means that the error should be
+	// interpreted to decide on the state transition (e.g. retriable errors may
+	// cause different transitions than non-retriable errors).
+	// If set, err must also be set, and targetSet is ignored.
+	transitionDependentOnErrType bool
+	// err is used by statements to inform that their execution encountered an
+	// error. The error will be eventually passed to pgwire, which serializes it
+	// to the client.
+	// The error can be a communication error (e.g. sending results to the client
+	// failed); these errors will be recognized by pgwire and the connection will
+	// be terminated.
+	err error
+
+	// targetState is the state to which we'll transition.
+	targetState TxnStateEnum
+}
+
 // rollbackSQLTransaction executes a ROLLBACK statement. The transaction is
 // rolled-back and the statement result is written to res.
 //
-// No matter what happens with the statement execution, the session state is
-// moved to NoTxn.
-func rollbackSQLTransaction(txnState *txnState, res StatementResult) error {
+// Returns the desired state transition. Since ROLLBACK is not concerned with
+// retriable errors, stateTransition.transitionDependentOnErrType is guaranteed
+// to be false.
+func rollbackSQLTransaction(txnState *txnState, res StatementResult) stateTransition {
 	if !txnState.TxnIsOpen() && txnState.State() != RestartWait {
 		panic(fmt.Sprintf("rollbackSQLTransaction called on txn in wrong state: %s (txn: %s)",
 			txnState.State(), txnState.mu.txn.Proto()))
@@ -1802,17 +1869,18 @@ func rollbackSQLTransaction(txnState *txnState, res StatementResult) error {
 	err := txnState.mu.txn.Rollback(txnState.Ctx)
 	if err != nil {
 		log.Warningf(txnState.Ctx, "txn rollback failed: %s", err)
-		txnState.resetStateAndTxn(NoTxn)
-		res.BeginResult((*parser.RollbackTransaction)(nil))
-		if closeErr := res.CloseResult(); closeErr != nil {
-			return closeErr
-		}
-		return err
 	}
 	// We're done with this txn.
-	txnState.resetStateAndTxn(NoTxn)
 	res.BeginResult((*parser.RollbackTransaction)(nil))
-	return res.CloseResult()
+	if closeErr := res.CloseResult(); closeErr != nil {
+		return stateTransition{
+			targetState: Aborted,
+			err:         closeErr,
+		}
+	}
+	return stateTransition{
+		targetState: NoTxn,
+	}
 }
 
 type commitType int
@@ -1822,9 +1890,12 @@ const (
 	release
 )
 
-// commitSqlTransaction executes a COMMIT or RELEASE SAVEPOINT statement. The
-// transaction is committed and results are writtern to res.
-func commitSQLTransaction(txnState *txnState, commitType commitType, res StatementResult) error {
+// commitSQLTransaction executes a COMMIT or RELEASE SAVEPOINT statement. The
+// transaction is committed and the statement result is written to res.
+func commitSQLTransaction(
+	txnState *txnState, commitType commitType, res StatementResult,
+) stateTransition {
+
 	if !txnState.TxnIsOpen() {
 		panic(fmt.Sprintf("commitSqlTransaction called on non-open txn: %+v", txnState.mu.txn))
 	}
@@ -1835,21 +1906,40 @@ func commitSQLTransaction(txnState *txnState, commitType commitType, res Stateme
 		// Errors on COMMIT need special handling: if the errors is not handled by
 		// auto-retry, COMMIT needs to finalize the transaction (it can't leave it
 		// in Aborted or RestartWait). Higher layers will handle this with the help
-		// of `txnState.commitSeen`, set above.
-		return err
+		// of `txnState.commitSeen`, set above. As far as the layer interpreting our
+		// stateTransition is concerned, transitionDependentOnErrType is the right
+		// thing to do without concern for the special nature of COMMIT.
+		return stateTransition{
+			transitionDependentOnErrType: true,
+			err: err,
+		}
 	}
 
+	var transition stateTransition
 	switch commitType {
 	case release:
 		// We'll now be waiting for a COMMIT.
-		txnState.resetStateAndTxn(CommitWait)
+		transition.targetState = CommitWait
 	case commit:
 		// We're done with this txn.
-		txnState.resetStateAndTxn(NoTxn)
+		transition.targetState = NoTxn
 	}
 
 	res.BeginResult((*parser.CommitTransaction)(nil))
-	return res.CloseResult()
+	if err := res.CloseResult(); err != nil {
+		transition = stateTransition{
+			err: err,
+			// The state after a communication error doesn't really matter, as the
+			// session will not receive more statements. We're using Aborted for
+			// consistency with most error code paths which don't have special
+			// handling for communication errors at the Executor level.
+			//
+			// TODO(andrei): Introduce a dedicated state to represent broken sessions
+			// that should not accept any statement anymore.
+			targetState: Aborted,
+		}
+	}
+	return transition
 }
 
 // exectDistSQL converts a classic plan to a distributed SQL physical plan and
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -1689,9 +1689,6 @@ func (e *Executor) execStmtInOpenTxn(
 			txnState.mu.txn.Proto().Restart(
 				0 /* userPriority */, 0 /* upgradePriority */, hlc.Timestamp{})
 		}
-		if err != nil {
-			return err
-		}
 		return nil
 
 	case *parser.Prepare:
@@ -1793,7 +1790,10 @@ func stmtAllowedInImplicitTxn(stmt Statement) bool {
 }
 
 // rollbackSQLTransaction executes a ROLLBACK statement. The transaction is
-// rolled-back results are written to res. All errors are swallowed.
+// rolled-back and the statement result is written to res.
+//
+// No matter what happens with the statement execution, the session state is
+// moved to NoTxn.
 func rollbackSQLTransaction(txnState *txnState, res StatementResult) error {
 	if !txnState.TxnIsOpen() && txnState.State() != RestartWait {
 		panic(fmt.Sprintf("rollbackSQLTransaction called on txn in wrong state: %s (txn: %s)",
--- a/pkg/sql/session_test.go
+++ b/pkg/sql/session_test.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -26,8 +27,11 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/base"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/security"
 	"github.com/cockroachdb/cockroach/pkg/sql"
+	"github.com/cockroachdb/cockroach/pkg/storage"
+	"github.com/cockroachdb/cockroach/pkg/storage/storagebase"
 	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/sqlutils"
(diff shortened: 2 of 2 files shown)


andreimatei:
I plan on cherry-picking this into 1.1.

bdarnell:
A lot of the change here has to do with the behavior of `COMMIT` (which is the only user of the error-dependent transition), but the bug report that motivated this change (and the test) were only about errors returned by `ROLLBACK`. For 1.1, isn't there a more targeted fix we could make inside `rollbackSQLTransaction`? It looks to me like the important part of this change is that we set the transaction state to aborted when `CloseResult` returns an error. Am I missing something? 

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/sql/executor.go, line 1597 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9R3irQUtBqDzVHARh:-Kv9R3irQUtBqDzVHARi:bhplop) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1597)):*
> ```Go
> 				// txn in the Aborted state; we transition back to NoTxn.
> 				//
> 				// TODO(andrei): BEGIN should use the explicitStateTransition mechanism.
> ```

Will `BEGIN` use `transitionDependentOnErrType`? If it doesn't, I think it would be better to isolate that whole mechanism in the two call sites to `commitSQLTransaction`. 

---

*[pkg/sql/executor.go, line 1861 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9PPvQag3s4oLH8FHl:-Kv9PPvRgPU_dprnApXb:b-l4ipq0) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1861)):*
> ```Go
> // retriable errors, stateTransition.transitionDependentOnErrType is guaranteed
> // to be false.
> func rollbackSQLTransaction(txnState *txnState, res StatementResult) stateTransition {
> ```

If `rollbackSQLTransaction` is guaranteed not to return a `stateTransition` with `transitionDependentOnErrorType`, and one of the two call sites doesn't even want a `stateTransition` in the first place, it seems better to make `rollbackSQLTransaction` return a `targetState, error` pair (and then construct a `stateTransition` in `execStmtInTransaction()`)

---

*[pkg/sql/executor.go, line 1876 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9PmesZvF3DhSnNWvf:-Kv9PmesZvF3DhSnNWvg:b-owswey) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1876)):*
> ```Go
> 	res.BeginResult((*parser.RollbackTransaction)(nil))
> 	if closeErr := res.CloseResult(); closeErr != nil {
> 		transition = stateTransition{
> ```

This can be `return stateTransition{...}`, and then the `transition` variable defined above can be moved down to the final return.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-:-Kv9RcDYbV294oherGVI:b84basq)*

andreimatei:
We had the same problem with `ROLLBACK` as with `COMMIT`, that's why `commitSQLTransaction` also changed.
In the commit case, only communication errors (i.e. errors from `res.CloseResult()`) would have caused the panic, whereas in the rollback case any error will do.

The point of this PR is that both `rollbackSQLTxn` and `commitSQLTxn` don't call `resetStateAndTxn()` any more; instead they delegate to the higher layer which doesn't freak out any more because someone changed the state from underneath it when an error is returned.

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/sql/executor.go, line 1597 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9R3irQUtBqDzVHARh:-Kv9Y8VBGgZ35YW43IIt:buc7gsv) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1597)):*

Will `BEGIN` use `transitionDependentOnErrType`? If it doesn't, I think it would be better to isolate that whole mechanism in the two call sites to `commitSQLTransaction`. 
</blockquote></details>

The plan is for a larger refactoring where more things use a mechanism in this spirit. Everything about the mechanism itself and the place where state transitions will happen is likely to change.

---

*[pkg/sql/executor.go, line 1861 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9PPvQag3s4oLH8FHl:-Kv9W82h0Nsx7LmB5v1b:b-ympoze) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1861)):*
> and one of the two call sites doesn't even want a stateTransition in the first place

Which call site is that? `execStmtInAbortedTxn` also does a state transition.

---

*[pkg/sql/executor.go, line 1876 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9PmesZvF3DhSnNWvf:-Kv9WYgSuOsCDj1yRUp6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1876)):*

This can be `return stateTransition{...}`, and then the `transition` variable defined above can be moved down to the final return.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-:-Kv9Wm91oZpmn2sISETk:b-tm4slv)*

knz:
Reviewed 2 of 2 files at r4.

---

*[pkg/sql/executor.go, line 1568 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9cqDFhoRFG_YAWdo3:-Kv9cqDGYuMQKsFSFOQy:b2cbd0a) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1568)):*
> ```Go
> 		// There's two cases to handle, depending on the two categories of
> 		// statements we have:
> 		// - Some statements explicitly decide what state transition they want. For
> ```

Give an example in the comment.

---

*[pkg/sql/executor.go, line 1576 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9cxq_ZOH-KqMGFNLx:-Kv9cxq_ZOH-KqMGFNLy:b-jvq1bu) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1576)):*
> ```Go
> 		if explicitStateTransition {
> 			if err != nil {
> 				panic(fmt.Sprintf("unexpected return err when explicitStateTransition is set: %s", err))
> ```

print error with `%+v`

---

*[pkg/sql/executor.go, line 1600 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9dh6hK5OVURQm4cAP:-Kv9dh6hK5OVURQm4cAQ:b-26ko7y) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1600)):*
> ```Go
> 				txnState.resetStateAndTxn(NoTxn)
> 			}
> 			return
> ```

I think the `return` must be outside of the closing `}` below, with an additional `if err != nil { return }`. Otherwise, the control flow can go through with `err != nil`, because it could be set in the first `if` branch above.

---

*[pkg/sql/executor.go, line 1620 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9dM0LC7OSURyrBGkv:-Kv9dM0M5BpIE0r_IWih:bs1w0dk) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1620)):*
> ```Go
> 				// If synchronizing results in a non-retryable error, it takes priority.
> 				if _, ok := parErr.(*roachpb.HandledRetryableTxnError); !ok {
> 					err = parErr
> ```

I would recommend if err is not nil at this point, merging the two errors together
```
if err != nil {
    err = errors.Wrapf(parErr, "while cleaning up after: %+v", err)
} else {
    err = parErr
}
```

ditto below

---

*[pkg/sql/session_test.go, line 246 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eMLWxjGzSCW9mtsy:-Kv9eMLWxjGzSCW9mtsz:b-cv5jb3) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L246)):*
> ```Go
> 					// the test there's two EndTransactions being sent - one is the direct
> 					// result of the test's call to tx.Rollback(), the second is sent by
> 					// the TxnCoordSender - indirectly triggered by the fact that, on the
> ```

Even if what you write here is correct, it doesn't feel right.
@tschottdorf is this something that should be investigated further?
If so Andrei please file an issue and add a TODO here.

---

*[pkg/sql/session_test.go, line 295 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eZ41GWNUM_O7zwJH:-Kv9eZ41GWNUM_O7zwJI:b4gtqtq) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L295)):*
> ```Go
> 
> 	if atomic.LoadInt64(&injectedErr) == 0 {
> 		t.Fatal("Test didn't inject the error. It must have failed to find " +
> ```

we do lowercase in messages. `t.Fatal("test didn't inject the error - it must have ...`
Also no final period.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893)*

andreimatei:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[pkg/sql/executor.go, line 1568 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9cqDFhoRFG_YAWdo3:-Kv9fofHYz50K9KyPTVF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1568)):*

Give an example in the comment.
</blockquote></details>

Done.

---

*[pkg/sql/executor.go, line 1576 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9cxq_ZOH-KqMGFNLx:-Kv9fX6_YPsXS4JZUh0Y:b-5fbvgf) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1576)):*

print error with `%+v`
</blockquote></details>

why? "%s" is what we always use for errs...

---

*[pkg/sql/executor.go, line 1600 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9dh6hK5OVURQm4cAP:-Kv9hE6MmAF6obneG6ee:b-o8dk5e) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1600)):*

I think the `return` must be outside of the closing `}` below, with an additional `if err != nil { return }`. Otherwise, the control flow can go through with `err != nil`, because it could be set in the first `if` branch above.
</blockquote></details>

done.

It was fine, because you can't be in state `AutoRetry` if that branch was taken. You also can't be in state `AutoRetry` if this current branch was taken, which says that just deleting the `return` should also be fine. I don't know what the cleanest thing to do is;  what we really should do is what we were talking about a while ago - stop propagating query execution errors (as opposed to communication errors or other errors that have to interrupt a session).

---

*[pkg/sql/executor.go, line 1620 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9dM0LC7OSURyrBGkv:-Kv9ihQQMo5tPtFLlUcI:bdemnma) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1620)):*

I would recommend if err is not nil at this point, merging the two errors together
```
if err != nil {
    err = errors.Wrapf(parErr, "while cleaning up after: %+v", err)
} else {
    err = parErr
}
```

ditto below
</blockquote></details>

I made it clear that you can't get here with `err != nil`, so this is moot, right?

---

*[pkg/sql/session_test.go, line 246 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eMLWxjGzSCW9mtsy:-Kv9jdBJqqzuyoy1U_qJ:bjny09z) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L246)):*

Even if what you write here is correct, it doesn't feel right.
@tschottdorf is this something that should be investigated further?
If so Andrei please file an issue and add a TODO here.
</blockquote></details>

It's right enough, I think. The `TxnCoordSender` would have cleaned up the state of the transaction if the `EndTransaction` would have succeeded - because it would have gotten a response with an updated `Transaction` proto with a final status. Since we injected the error, as far as it is concerned, the transaction is still alive and well, and the client can attempt to roll it back again.
The SQL layer decides to not retry anything, and cancels the transaction context. Now `TxnCoordSender` is convinced to give up on it and cleans up.

It's not very satisfying, but I don't think there's much to see here.

I separately have an item to investigate about eliminating some errors that `EndTransaction` can return - the thing should not fail, for the most part. What should happen when it does fail... probably not very important.

---

*[pkg/sql/session_test.go, line 295 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eZ41GWNUM_O7zwJH:-Kv9iw54WvmsuRTrjrnN:b-um19kt) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L295)):*

we do lowercase in messages. `t.Fatal("test didn't inject the error - it must have ...`
Also no final period.
</blockquote></details>

done
I protest, though. Are you sure this rule was ever established for fatal messages? It has been established for errors (for better or worse), because they might get wrapped. But this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893)*

bdarnell:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/sql/executor.go, line 1861 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9PPvQag3s4oLH8FHl:-KvDOCcfWIy_MEmpQ321:b-bprc10) ([raw file](https://github.com/cockroachdb/cockroach/blob/29f32fc57e9ff043fbdfda9bafab8c8a547a50df/pkg/sql/executor.go#L1861)):*

> and one of the two call sites doesn't even want a stateTransition in the first place

Which call site is that? `execStmtInAbortedTxn` also does a state transition.
</blockquote></details>

`execStmtInAbortedTxn` does a state transition, but it doesn't need the `transitionDependentOnErrorType` logic. That's why I'm suggesting it might be more natural to return `targetState, err` instead of this struct with other functionality that is not needed in this case. But I don't feel strongly if this is more compatible with future evolution.

---

*[pkg/sql/executor.go, line 1576 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9cxq_ZOH-KqMGFNLx:-KvDO-7Og7OOjxD7iiCP:b-881mdu) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1576)):*

why? "%s" is what we always use for errs...
</blockquote></details>

`%+v` includes the stack trace where the error was generated, which would be useful in this case. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-:-KvDPminhpRee7f-wD6o:bnfp4nl)*

knz:
LGTM too modulo %+v and comment suggestion

---

Reviewed 2 of 2 files at r5.

---

*[pkg/sql/executor.go, line 1600 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9dh6hK5OVURQm4cAP:-KvIiD-OJ7rSp_TRRHxk:b-q0q535) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1600)):*

done.

It was fine, because you can't be in state `AutoRetry` if that branch was taken. You also can't be in state `AutoRetry` if this current branch was taken, which says that just deleting the `return` should also be fine. I don't know what the cleanest thing to do is;  what we really should do is what we were talking about a while ago - stop propagating query execution errors (as opposed to communication errors or other errors that have to interrupt a session).
</blockquote></details>

I agree there's more thinking to be done. You could add it as  a TODO comment here.
However I strongly feel that pulling the `return` out was the right thing to do now, because it makes the code more readable and it enables local reasoning.

---

*[pkg/sql/executor.go, line 1620 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9dM0LC7OSURyrBGkv:-KvIiQCZZYIBQqohhLYp:blupe9a) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1620)):*

I made it clear that you can't get here with `err != nil`, so this is moot, right?
</blockquote></details>

Correct. Thanks.

---

*[pkg/sql/session_test.go, line 246 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eMLWxjGzSCW9mtsy:-KvIiZxhJ341TNLMG-IU:b-xpntp0) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L246)):*

It's right enough, I think. The `TxnCoordSender` would have cleaned up the state of the transaction if the `EndTransaction` would have succeeded - because it would have gotten a response with an updated `Transaction` proto with a final status. Since we injected the error, as far as it is concerned, the transaction is still alive and well, and the client can attempt to roll it back again.
The SQL layer decides to not retry anything, and cancels the transaction context. Now `TxnCoordSender` is convinced to give up on it and cleans up.

It's not very satisfying, but I don't think there's much to see here.

I separately have an item to investigate about eliminating some errors that `EndTransaction` can return - the thing should not fail, for the most part. What should happen when it does fail... probably not very important.
</blockquote></details>

Okay. Care about copy-pasting your explanation into the comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-:-KvIieAxNbuN-ZdqK0dM:b-o4zrjf)*

andreimatei:

---

*[pkg/sql/executor.go, line 1576 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9cxq_ZOH-KqMGFNLx:-KvSbVG4TYf5u3mK_J75:b-qki4o4) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/executor.go#L1576)):*

`%+v` includes the stack trace where the error was generated, which would be useful in this case. 
</blockquote></details>

Done.

OK, so this stack printing works because the errors created by `pkg/errors` implement the `Formatter` interface, and it doesn't hurt other errors because there's special handling in the printf functions for errors that makes it so that `Error()` is always called to produce a string.

---

*[pkg/sql/session_test.go, line 246 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eMLWxjGzSCW9mtsy:-KvScOLJ6SeQ12uALko_:b-sj1uds) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L246)):*

Okay. Care about copy-pasting your explanation into the comment?
</blockquote></details>

not particularly :)
I think the existing comment is sufficient for this test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893)*

knz:
Reviewed 1 of 2 files at r7.
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/sql/session_test.go, line 246 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eMLWxjGzSCW9mtsy:-KvSf0e9FlmPRJ6zSKl8:b-7kx5uv) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L246)):*

not particularly :)
I think the existing comment is sufficient for this test.
</blockquote></details>

It's fine if you don't want to extend the comment in *this file*.

However leaving this piece of knowledge to rot in a reviewable comment is **not** an option. Please find *some* good place to write this information as a comment in the source base. It might go well in `executor.go` next to the handling of commit/abort.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893)*

andreimatei:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/sql/session_test.go, line 246 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893#-Kv9eMLWxjGzSCW9mtsy:-KvSjebyww334KmYyZu3:bkdyshn) ([raw file](https://github.com/cockroachdb/cockroach/blob/6c9caad418bf1290373a179b1c3d415152536265/pkg/sql/session_test.go#L246)):*

It's fine if you don't want to extend the comment in *this file*.

However leaving this piece of knowledge to rot in a reviewable comment is **not** an option. Please find *some* good place to write this information as a comment in the source base. It might go well in `executor.go` next to the handling of commit/abort.
</blockquote></details>

What precisely is the "piece of knowledge" that you find so valuable? The way I see it, there's no particular surprise here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18893)*

PR #18899, opened 2017-09-28
sqlccl: support into_db in IMPORT

### LINKED ISSUE #18876: sqlccl: verify db exists when into_db is used with IMPORT CSV
https://github.com/cockroachdb/cockroach/blob/14c5215987e857d881c46749dbcf42f1a724ae1d/pkg/ccl/sqlccl/csv.go#L922

madelynnblue:
Fixes #18876

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -27,6 +27,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/build"
 	"github.com/cockroachdb/cockroach/pkg/ccl/storageccl"
 	"github.com/cockroachdb/cockroach/pkg/config"
+	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/keys"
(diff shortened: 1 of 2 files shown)


dt (approved):

mjibson:
@bdarnell thoughts about cherry picking this into 1.1? I'm fine to leave it out, because it's easy enough to work around by setting your session database. (And, it's not that the syntax was allowed but ignored: it will still error in 1.1 if this option is present, so users won't be confused.)

bdarnell:
This is about parity between IMPORT and RESTORE, right? Let's leave it out of 1.1 since the feature is still marked experimental anyway.

mjibson:
Yes. Fine with me.

dt:
@bdarnell @mjibson huh, I might be coming around to pro-cherrypick on this one: the error message if the session DB is empty says to supply the `into_db` option... which then fails, so it ends up looking kinda like we have a bug. We could just patch the error message, but given that it comes from a reused helper, cherrypicking this might be just as conservative.

PR #18900, opened 2017-09-28
server, ui: add quiescent raft state to the debug page

### LINKED ISSUE #18450: ui: problemranges page should show whether replica is quiesced
It shows "dormant" but not quiesced.

@cuongdo for triage.

BramGruneir:
Instead of being considered `dormant` when there is no raft state, if the range is quiescent, it will now show the replica as having that state and colour the whole replica accordingly.

Closes #18450.

screen shot 2017-09-28 at 13 32 04


couchand:
:lgtm:

My only question is if we want the entire range table to be lavender.  That seems pretty... intense.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-:-Kv9316S9mHr4WAosUEb:buo6w2m)*

BramGruneir:
I had to change it from lavender to violet, the lavender was too light.

And yes, since all the values have been frozen as the range is quiesced.  

Dormant replicas are all grey. 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-:-Kv93JVWKVCn_KXkYSFF:bjds3h8)*

tschottdorf:
Thanks! My one comment would be that the color could be confusing. Maybe it isn't needed since the description now says `quiescent`?

tschottdorf:
(or just have them grey as well)

BramGruneir:
I wanted to differentiate it from the grey dormant replicas.  I could try for a softer colour if that would help.

But if you see some quiesced and some non, then we should know we have a problem right away.  Dormants should now only happen when a replica is waiting to be GCed or when it is brand new and isn't part of the raft group yet.

BramGruneir:
On second thought, I like your idea of just showing the quiescent state and leaving the rest as normal.  I'm going to go ahead and make that change.

BramGruneir:
Here's an updated screen shot (of a range with some clear issues!):

screen shot 2017-09-28 at 16 51 37

bdarnell:
Quiescence occurs in combination with some other state - a quiescent node is still either a leader or a follower. (I don't think it's possible to be both quiesced and dormant)

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-:-Kv9HDANgtutbaqwZAYF:b-gpkcjq)*

BramGruneir:
When a range is quiescent (or dormant), `replica.RaftStatus()` returns a nil.

So there is no
- raft state
- raft leader
- vote
- term
- applied index
- commit index

I'm going to convert those values to a "-" when a range is quiescent or dormant.

bdarnell:
Are you sure `replica.RaftStatus()` returns nil for a quiesced range? Quiescence doesn't set `internalRaftGroup` to nil.

BramGruneir:
I'll double check in a bit.

BramGruneir:
Hmm... so you are correct, it's not always dormant.  Especially right after I restart my local cluster. 
 I think what I was seeing was that sometimes the ranges could be both dormant and quiescent.

I'm going to factor this into the change and not make quiescent a new raft state, but just another field on the range debug page.

BramGruneir:
Ok, I've changed this up as I described earlier.

For states that are dormant, they don't show any details about their raft state, since they have none:
screen shot 2017-10-02 at 17 13 57

A quiesced range now looks like this:
screen shot 2017-10-02 at 17 14 08

And a non-quiesced range now looks like this:
screen shot 2017-10-02 at 17 18 52

BramGruneir:
So I think this is good to go, PTAL.

couchand:
Reviewed 2 of 9 files at r1.
Review status: 2 of 9 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/ui/src/views/reports/containers/range/rangeTable.tsx, line 201 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-KvXY1z196mI0Hvk-khu:-KvXY1z196mI0Hvk-khv:bdds1sn) ([raw file](https://github.com/cockroachdb/cockroach/blob/52de287d2359c99e47ffb241ffd35a4f3b96601c/pkg/ui/src/views/reports/containers/range/rangeTable.tsx#L201)):*
> ```XML
>   // contentNoValue returns an empty value if the condition is false, and if
>   // true, executes and returns the content function.
>   contentNoValue(
> ```

how about `contentIf` or `showIf` or something like that?

---

*[pkg/ui/src/views/reports/containers/range/rangeTable.tsx, line 399 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-KvXYlohbhhhhqNO6ZNm:-KvXYlohbhhhhqNO6ZNn:b-y4f7jm) ([raw file](https://github.com/cockroachdb/cockroach/blob/52de287d2359c99e47ffb241ffd35a4f3b96601c/pkg/ui/src/views/reports/containers/range/rangeTable.tsx#L399)):*
> ```XML
>         );
>       }
>       const dormant = (raftState.value[0] === "dormant");
> ```

nit: no parens

---

*[pkg/ui/src/views/reports/containers/range/rangeTable.tsx, line 413 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-KvXYRUwyuDK5vMff_m-:-KvXYRUwyuDK5vMff_m0:b1h5o9f) ([raw file](https://github.com/cockroachdb/cockroach/blob/52de287d2359c99e47ffb241ffd35a4f3b96601c/pkg/ui/src/views/reports/containers/range/rangeTable.tsx#L413)):*
> ```XML
>         problems: this.contentProblems(info.problems, awaitingGC),
>         raftState: raftState,
>         quiescent: info.quiescent ? rangeTableQuiescent : rangeTableEmptyContent,
> ```

This might be overkill, but how about `this.contentIf(quiescent, () => rangeTableQuiescent)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900)*

BramGruneir:
Review status: 1 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/ui/src/views/reports/containers/range/rangeTable.tsx, line 201 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-KvXY1z196mI0Hvk-khu:-KvXbaWjqzNl6v_7l70K:bve87f1) ([raw file](https://github.com/cockroachdb/cockroach/blob/52de287d2359c99e47ffb241ffd35a4f3b96601c/pkg/ui/src/views/reports/containers/range/rangeTable.tsx#L201)):*

how about `contentIf` or `showIf` or something like that?
</blockquote></details>

That's a much better name.  Done.

---

*[pkg/ui/src/views/reports/containers/range/rangeTable.tsx, line 399 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-KvXYlohbhhhhqNO6ZNm:-KvXbcOWf0I9BFDbcGsQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/52de287d2359c99e47ffb241ffd35a4f3b96601c/pkg/ui/src/views/reports/containers/range/rangeTable.tsx#L399)):*

nit: no parens
</blockquote></details>

Done.

---

*[pkg/ui/src/views/reports/containers/range/rangeTable.tsx, line 413 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900#-KvXYRUwyuDK5vMff_m-:-KvXbcutUry0H4FvP70j:b1pm2dk) ([raw file](https://github.com/cockroachdb/cockroach/blob/52de287d2359c99e47ffb241ffd35a4f3b96601c/pkg/ui/src/views/reports/containers/range/rangeTable.tsx#L413)):*

This might be overkill, but how about `this.contentIf(quiescent, () => rangeTableQuiescent)`
</blockquote></details>

The reason I didn't go with that is because there's no execution cost associated with the constants.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18900)*

PR #18910, opened 2017-09-29
storage,util/hlc: convert hlc.Timestamp to proto3

petermattis:
Convert hlc.Timestamp to proto3 so that a logical component of 0 will
not be encoded on the wire. This is complicated by needing to leave the
"below raft" encoding for MVCCMetadata unchanged. Added LegacyTimestamp
to accomplish this which is convertible to Timestamp but has the legacy
encoding. Changed the MVCCMetadata timestamp fields to use
LegacyTimestamp. Note that the timestamp in MVCCMetadata.Txn (type
TxnMeta) does not need to change because we never encoding an
MVCCMetadata with a non-nil Txn field below raft. Extended the
TrackRaftProtos mechanism to verify that TxnMeta is never encoded blow
raft.


bdarnell:
:lgtm:

---

Reviewed 21 of 21 files at r1.

---

*[pkg/storage/track_raft_protos.go, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18910#-KvDn5v2TqxjXgiYZ2oG:-KvDn5v2TqxjXgiYZ2oH:b-ynb6wq) ([raw file](https://github.com/cockroachdb/cockroach/blob/1eead4839153e660abb35bef386373cdd0af5ae1/pkg/storage/track_raft_protos.go#L62)):*
> ```Go
> 		t := reflect.TypeOf(pb)
> 
> 		// Special handling for MVCCMetadata: we expect MVCCMetadata to be
> ```

I don't understand why this is needed. Shouldn't this be traversed automatically?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18910#-:-KvDnVMT95iJ1aKa82iZ:bnfp4nl)*

petermattis:

---

*[pkg/storage/track_raft_protos.go, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18910#-KvDn5v2TqxjXgiYZ2oG:-KvDo3OLbD8YJwlXLFoK:binkg17) ([raw file](https://github.com/cockroachdb/cockroach/blob/1eead4839153e660abb35bef386373cdd0af5ae1/pkg/storage/track_raft_protos.go#L62)):*

I don't understand why this is needed. Shouldn't this be traversed automatically?
</blockquote></details>

That's what I thought initially, but that's not the way `protoutil.Interceptor` works. The interceptor is only invoked for the top-level `proto.Message` passed to `prototutil.Marshal`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18910)*

Issue #6248, opened 2016-04-22
sql: improve client.Batch creation on INSERT/DELETE/UPDATE


petermattis:
We currently perform a lot of allocations when constructing a `client.Batch` during INSERT/DELETE/UPDATE. For example, `Batch.CPut` performs allocations while marshaling values (via `client.marshalValue`), but those could be avoided if the caller marshaled directly into a `roachpb.Value`. We could also use chunk allocation to minimize the allocations necessary for `roachpb.Value.RawBytes`. `Batch.CPut` also performs ~3 allocations to create the `*roachpb.ConditionalPutRequest`, fill in the expected value (silliness is causing this to always be heap allocated) and to create the `client.KeyValue` for the result.

None of these should be too difficult to fix, though we need to coordinate any surgery with work such as #6211.

petermattis:
We should check to see if this is still worthwhile.

jordanlewis:
This is still a thing, though I think the number of allocations caused by this are still few compared to other sources.

PR #18912, opened 2017-09-29
 sql: exponentially grow Datum alloc chunks

nvb:
`DatumAlloc` used to allocate fixed chunks of 16 `Datums` at a time. This
was an improvement over allocating a single `Datum` at a time, but meant
that `Datum` decoding still required an allocation count that grew
linearly with respect to the number of rows in a scan.

This commit changes `DatumAlloc` so that it grows the chunk allocation
size exponentially as more `Datums` are needed. This means that `Datum`
decoding now only requires an allocation count that grows logarithmically
with respect to the number of rows in a scan. It uses a maximum allocation
size so we avoid allocating slices that are too big.

```
name                                     old time/op    new time/op    delta
Scan/Cockroach/count=100000/limit=0-16      156ms ± 4%     156ms ± 3%     ~     (p=1.000 n=23+23)

name                                     old alloc/op   new alloc/op   delta
Scan/Cockroach/count=100000/limit=0-16     48.1MB ± 1%    47.5MB ± 0%     ~     (p=0.433 n=25+24)

name                                     old allocs/op  new allocs/op  delta
Scan/Cockroach/count=100000/limit=0-16       258k ±17%      208k ± 0%  -19.33%  (p=0.000 n=25+24)
```

### DIFF
--- a/pkg/sql/bench_test.go
+++ b/pkg/sql/bench_test.go
@@ -509,16 +509,25 @@ func runBenchmarkScan(b *testing.B, db *gosql.DB, count int, limit int) {
 		b.Fatal(err)
 	}
 
-	var buf bytes.Buffer
-	buf.WriteString(`INSERT INTO bench.scan VALUES `)
-	for i := 0; i < count; i++ {
-		if i > 0 {
-			buf.WriteString(", ")
+	inserted := 0
+	for inserted < count {
+		const maxInsertSize = 100000
+		insertBatch := count - inserted
+		if insertBatch > maxInsertSize {
+			insertBatch = maxInsertSize
 		}
-		fmt.Fprintf(&buf, "(%d)", i)
-	}
-	if _, err := db.Exec(buf.String()); err != nil {
-		b.Fatal(err)
+		var buf bytes.Buffer
+		buf.WriteString(`INSERT INTO bench.scan VALUES `)
+		for i := 0; i < insertBatch; i++ {
+			if i > 0 {
+				buf.WriteString(", ")
+			}
+			fmt.Fprintf(&buf, "(%d)", i+inserted)
+		}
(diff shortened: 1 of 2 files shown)


petermattis:
Are any other benchmarks affected? (No need to run them all, but perhaps sample a few). The reduction in allocations is nice, but it would be nicer if there is a performance improvement. Also, https://github.com/cockroachdb/cockroach/issues/6248 might be more fruitful.

---

---

*[pkg/sql/sqlbase/table.go, line 956 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18912#-KvDqRNlq5nkvX3QTmnB:-KvDqRNlq5nkvX3QTmnC:b958840) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e7cb72e1ad9ab0a637a1878ac0d951372368e7d/pkg/sql/sqlbase/table.go#L956)):*
> ```Go
> 	if l == 0 {
> 		l = a.allocLen(cap(*buf))
> 		*buf = make([]parser.DInt, l, l)
> ```

The 3rd parameter isn't necessary here and below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18912#-:-KvDqYR0nXLOW02VKZAE:b-1m2scy)*

knz:
@nvanbenschoten mind sharing the rationale to close this? (is there any lesson to learn?)

nvanbenschoten:
Mainly because I thought this was https://github.com/cockroachdb/cockroach/pull/19050 without looking too closely. I'm going to re-open and try this out again at some point though because it seems like a pretty clear (if minor) win.

petermattis:
@nvanbenschoten Are you planning to do anything about this?

nvanbenschoten:
No, closing.

Issue #13193, opened 2017-01-27
sql,kv: investigate returning EncDatum format from KV on scans


RaduBerinde:
Since there is measurable overhead from serializing/deserializing keys/values we would likely get a benefit from a more efficient data encoding between the KV and SQL layers.

The idea (suggested by @petermattis) is to break apart the encoded keys/values (using `encoding.PeekLength` and such) and return rows of `EncDatum`s (or something that directly maps to `EncDatum`s) directly from KV. KV would understand how to parse values (column families etc); there would be a special mode or variant of ScanRequest which has the key and value types passed down from SQL.

This would also allow us to avoid sending out data for columns the scan doesn't need.

petermattis:
I believe the major inefficiency in the existing `ScanResponse` is our use of protos:

```
message Value {
  optional bytes raw_bytes = 1;
  optional util.hlc.Timestamp timestamp = 2;
}

message KeyValue {
  optional bytes key;
  optional Value value;
}

message ScanResponse {
  repeated KeyValue rows;
}
```

When we unmarshal a `ScanResponse`, each `KeyValue` requires 2 allocations: one for the key and one for the value. There is also a small bit of overhead in the protobuf encoding. My thinking is that a flag in the `ScanRequest` could indicate an alternate return format that is composed of SQL rows, encoded compactly:

```
message ScanResponse {
  ...
  optional group sql_rows {
    repeated int32 row_starts;
    required bytes data;
  }
}
```

`sql_rows.row_starts` would be the index into `sql_rows.data` of the start of each row. Each row would be a series of `EncDatums`. Lots more details to figure out here such as how to specify in the `ScanRequest` how to combine multiple key/values into a single row and which columns to return, but none of that seems insurmountable.

RaduBerinde:
I think this would improve performance significantly for most queries, especially when we have "wide" tables. I will work on a proposal for the API.

tschottdorf:
You could also make the change Peter suggested and then see how it improves the actual performance.

RaduBerinde:
What change is that exactly? There are a lot of details missing from the rough proposal above. I want to come up with a complete proposal before I go ahead with all the implementation (which is anything but trivial).

tschottdorf:
Sorry, didn't mean to say don't make a proposal. Meant to say don't make a full-fledged proposal before you've managed to get some sanity checks that this is useful. That's likely obvious to you anyway, so sorry about the noise.

[...]

PR #18913, opened 2017-09-29
[WIP] *: optionally return Scan response data in RocksDB batch repr format

petermattis:
This reduces the size of a Scan response containing 10k "kv" keys from
410 KiB to 322 KiB. If we omit sending the timestamp this can further
reduce the size to 234 KiB. This can be seen as a primitive version of
 #13193. The response size savings are compelling and we'd likely see
even more space savings if we switched to an encoded datum format.

Unfortunately, this change doesn't seem to have an effect on performance
even when forcing the RPC to go over the network. The test I performed
was to load 1m rows using `kv` and then run `select count(*) test.kv`,
flipping `ReturnBatchRepr` on and off.

There are bits here that deserve to be merged regardless. Specifically,
the changes to RocksDBBatchReader and the usage of roachpb.KeyValue
instead of client.KeyValue in sql. I'll split these pieces out into
separate PRs.


petermattis:
Cc @RaduBerinde, @bdarnell, @tschottdorf, @spencerkimball

bdarnell:
> Unfortunately, this change doesn't seem to have an effect on performance
> even when forcing the RPC to go over the network.

What configurations did you test? I'd expect a bigger difference for a multi-DC cluster like `indigo` than a single-DC one. 

---

Review status: 0 of 21 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18913#-:-KvEM2TKAmH8xtZ9tYyI:b-3ehfz4)*

petermattis:
> What configurations did you test? I'd expect a bigger difference for a multi-DC cluster like indigo than a single-DC one.

I only tested locally with the "localServer" optimization disabled. I agree this will make a bigger difference on wide area networks. Seems worth exploring further.

---

Review status: 0 of 21 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18913#-:-KvENU0ZFwtccSZCtooB:b-xuddfc)*

nvanbenschoten:
I just ran this against TPCC and I'm not really seeing any movement in the metrics. It's giving at most a **1%** improvement in `ns/op`, but that doesn't look statistically significant.

petermattis:
That's strange. I wonder why your PR showed improvement and mine didn't. Both should reduce allocations, while this one will also reduce the size of responses.

petermattis:
With #21395 merged, the low-level `MVCCScan` op now generates RocksDB batch repr format. I plan to take another look at this PR to see if it moves the needle. I'm also planning to look at how this affects performance on a geo-distributed cluster.

petermattis:
#26553 has replaced this PR.

PR #18914, opened 2017-09-29
storage/engine: reduce allocations in RocksDBBatchReader

petermattis:
Rather than use a bytes.Reader for reading the batch repr, read from a
[]byte. This allows us to return safe slices from
RocksDBBatchReader.{Key,Value} and removes the need to temporarily copy
the key and value.


bdarnell:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18914#-:-KvENDAwj83bnSwo7ppo:bnfp4nl)*

PR #18915, opened 2017-09-29
acceptance: identify time span with index creation problems

vivekmenezes:
related to #18533

### DIFF
--- a/pkg/acceptance/allocator_test.go
+++ b/pkg/acceptance/allocator_test.go
@@ -71,6 +71,9 @@ type allocatorTest struct {
 	// Run some schema changes during the rebalancing.
 	RunSchemaChanges bool
 
+	// start load time.
+	startLoad time.Time
+
 	f *terrafarm.Farmer
 }
 
@@ -155,6 +158,7 @@ func (at *allocatorTest) Run(ctx context.Context, t *testing.T) {
 	at.f.Assert(ctx, t)
 
 	log.Infof(ctx, "starting load on cluster")
+	at.startLoad = timeutil.Now()
 	if err := at.f.StartLoad(ctx, "block_writer"); err != nil {
 		t.Fatal(err)
 	}
@@ -177,7 +181,9 @@ func (at *allocatorTest) Run(ctx context.Context, t *testing.T) {
 			"SELECT COUNT(newcol) FROM tpch.customer AS OF SYSTEM TIME %s",
 			"SELECT COUNT(c_name) FROM tpch.customer@foo AS OF SYSTEM TIME %s",
 		}
-		if err := at.runSchemaChanges(ctx, t, schemaChanges, validationQueries); err != nil {
+		if err := at.runSchemaChanges(
+			ctx, t, schemaChanges, validationQueries, nil,
+		); err != nil {
 			t.Fatal(err)
 		}
 		// These schema changes are run later because the above schema
@@ -187,16 +193,23 @@ func (at *allocatorTest) Run(ctx context.Context, t *testing.T) {
 		// being updated.
 		log.Info(ctx, "running schema changes over datablocks.blocks")
 		schemaChanges = []string{
-			"ALTER TABLE datablocks.blocks ADD COLUMN newcol DECIMAL DEFAULT (DECIMAL '1.4')",
+			"ALTER TABLE datablocks.blocks ADD COLUMN created_at TIMESTAMP DEFAULT now()",
 			"CREATE INDEX foo ON datablocks.blocks (block_id)",
 		}
 		// All these return the same result.
 		validationQueries = []string{
 			"SELECT COUNT(*) FROM datablocks.blocks AS OF SYSTEM TIME %s",
-			"SELECT COUNT(newcol) FROM datablocks.blocks AS OF SYSTEM TIME %s",
+			"SELECT COUNT(created_at) FROM datablocks.blocks AS OF SYSTEM TIME %s",
 			"SELECT COUNT(block_id) FROM datablocks.blocks@foo AS OF SYSTEM TIME %s",
 		}
-		if err := at.runSchemaChanges(ctx, t, schemaChanges, validationQueries); err != nil {
+		// Queries to hone in on index validation problems.
+		indexValidationQueries := []string{
+			"SELECT COUNT(created_at) FROM datablocks.blocks@primary AS OF SYSTEM TIME %s WHERE created_at > $1 AND created_at < $2",
+			"SELECT COUNT(block_id) FROM datablocks.blocks@foo AS OF SYSTEM TIME %s WHERE created_at > $1 AND created_at < $2",
+		}
+		if err := at.runSchemaChanges(
+			ctx, t, schemaChanges, validationQueries, indexValidationQueries,
+		); err != nil {
 			t.Fatal(err)
 		}
 	}
@@ -218,7 +231,11 @@ func (at *allocatorTest) RunAndCleanup(ctx context.Context, t *testing.T) {
 }
 
 func (at *allocatorTest) runSchemaChanges(
-	ctx context.Context, t *testing.T, schemaChanges []string, validationQueries []string,
+	ctx context.Context,
+	t *testing.T,
+	schemaChanges []string,
+	validationQueries []string,
+	indexValidationQueries []string,
 ) error {
 	db, err := gosql.Open("postgres", at.f.PGUrl(ctx, 0))
 	if err != nil {
@@ -228,30 +245,20 @@ func (at *allocatorTest) runSchemaChanges(
 		_ = db.Close()
 	}()
 
-	errChan := make(chan error)
-	for i := range schemaChanges {
-		cmd := schemaChanges[i]
-		go func(i int) {
-			start := timeutil.Now()
-			log.Infof(ctx, "starting schema change: %s", cmd)
-			if _, err := db.Exec(cmd); err != nil {
-				errChan <- errors.Errorf("hit schema change error: %s, for %s, in %s", err, cmd, timeutil.Since(start))
-				return
-			}
-			log.Infof(ctx, "completed schema change: %s, in %s", cmd, timeutil.Since(start))
-			errChan <- nil
-			// TODO(vivek): Monitor progress of schema changes and log progress.
-		}(i)
-	}
-
-	for range schemaChanges {
-		if err := <-errChan; err != nil {
-			t.Fatal(err)
+	for _, cmd := range schemaChanges {
+		start := timeutil.Now()
+		log.Infof(ctx, "starting schema change: %s", cmd)
+		if _, err := db.Exec(cmd); err != nil {
+			t.Fatalf("hit schema change error: %s, for %s, in %s", err, cmd, timeutil.Since(start))
 		}
+		log.Infof(ctx, "completed schema change: %s, in %s", cmd, timeutil.Since(start))
+		// TODO(vivek): Monitor progress of schema changes and log progress.
 	}
 
 	log.Info(ctx, "validate applied schema changes")
-	if err := at.ValidateSchemaChanges(ctx, t, validationQueries); err != nil {
+	if err := at.ValidateSchemaChanges(
+		ctx, t, validationQueries, indexValidationQueries,
+	); err != nil {
 		t.Fatal(err)
 	}
 	return nil
@@ -259,7 +266,7 @@ func (at *allocatorTest) runSchemaChanges(
 
 // The validationQueries all return the same result.
 func (at *allocatorTest) ValidateSchemaChanges(
-	ctx context.Context, t *testing.T, validationQueries []string,
+	ctx context.Context, t *testing.T, validationQueries []string, indexValidationQueries []string,
 ) error {
 	// Sleep for a bit before validating the schema changes to
 	// accommodate for time differences between nodes. Some of the
@@ -278,16 +285,21 @@ func (at *allocatorTest) ValidateSchemaChanges(
 		_ = db.Close()
 	}()
 
-	var now string
-	if err := db.QueryRow("SELECT cluster_logical_timestamp()").Scan(&now); err != nil {
+	var nowString string
+	if err := db.QueryRow("SELECT cluster_logical_timestamp()").Scan(&nowString); err != nil {
 		t.Fatal(err)
 	}
+	var nowInNanos int64
+	if _, err := fmt.Sscanf(nowString, "%d", &nowInNanos); err != nil {
+		t.Fatal(err)
+	}
+	now := timeutil.Unix(0, nowInNanos)
 
 	// Validate the different schema changes
 	var eCount int64
 	for i := range validationQueries {
 		var count int64
-		q := fmt.Sprintf(validationQueries[i], now)
+		q := fmt.Sprintf(validationQueries[i], nowString)
 		if err := db.QueryRow(q).Scan(&count); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/acceptance/allocator_test.go
```diff
@@ -278,16 +290,14 @@ func (at *allocatorTest) ValidateSchemaChanges(
 		_ = db.Close()
 	}()
 
-	var now string
-	if err := db.QueryRow("SELECT cluster_logical_timestamp()").Scan(&now); err != nil {
-		t.Fatal(err)
-	}
+	now := timeutil.Now()
+	nowString := fmt.Sprintf("%d.0000000000", now.UnixNano())
```
vivekmenezes: I had to stop using the cluster logical time because a cluster logical time is a decimal which is read as a string, which then needs to be converted to a decimal and to a time.Time for this change. Since we already have a 5 second sleep above i thought this was fine.

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[pkg/acceptance/allocator_test.go, line 245 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvENx0fGlcEc0_VR_bu:-KvENx0fGlcEc0_VR_bv:b-tgol2a) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L245)):*
> ```Go
> 			// Guarantee a schema change execution order.
> 			// We need this because the schema validation code depends on the
> 			// the schema changes executing in a specific order.
> ```

Why are we starting a goroutine for each schema change if we want them to execute serially? Why not just one goroutine in a loop?

---

*[pkg/acceptance/allocator_test.go, line 294 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEJOUCM_7DYbhy8MR7-r1-294:-KvEPIeshAhzD0Hih8Hr:b2nr7jn) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L294)):*

I had to stop using the cluster logical time because a cluster logical time is a decimal which is read as a string, which then needs to be converted to a decimal and to a time.Time for this change. Since we already have a 5 second sleep above i thought this was fine.
</blockquote></details>

We should keep using `cluster_logical_timestamp` for the `as of system time` queries. (and we should always aim to get rid of arbitrary sleeps wherever we can) I think we can use a separate non-logical timestamp for the extra validation (if it matters that these are slightly different values, then the logical ticks will also be important and the test should be reworked to use HLC timestamps everywhere)

---

*[pkg/acceptance/allocator_test.go, line 313 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQL47th-BB4DBk_Mv:-KvEQL47th-BB4DBk_Mw:bc2yzqc) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L313)):*
> ```Go
> 			// Investigate index creation problems. Always run this so we know
> 			// it works.
> 			if extraValidate && i == 2 {
> ```

Where does this `2` come from?

---

*[pkg/acceptance/allocator_test.go, line 314 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQOdSj_7UlTigBEJW:-KvEQOdSj_7UlTigBEJX:b-d3zldj) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L314)):*
> ```Go
> 			// it works.
> 			if extraValidate && i == 2 {
> 				sp := span{start: now.Add(-30 * time.Minute), end: now}
> ```

Why this time span? Why not the whole database? (this is only run once at the end of the test, right?)

---

*[pkg/acceptance/allocator_test.go, line 327 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEPb_UhkzEJ2wAIwIy:-KvEPb_VyatNfXjlYQ-V:b-bmp53h) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L327)):*
> ```Go
> }
> 
> type span struct {
> ```

s/span/timeSpan/g

---

*[pkg/acceptance/allocator_test.go, line 335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQxsF__oh4EhZ7sqg:-KvEQxsF__oh4EhZ7sqh:b5tf70q) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L335)):*
> ```Go
> func (at *allocatorTest) checkIndexOverTimeSpan(ctx context.Context, db *gosql.DB, s span) (bool, error) {
> 	var eCount int64
> 	if err := db.QueryRow("SELECT COUNT(created_at) FROM datablocks.blocks@primary WHERE created_at > $1 AND created_at < $2", s.start, s.end).Scan(&eCount); err != nil {
> ```

Why are these queries hard-coded while the rest of the test uses queries that are passed in?  Why do the other validation queries use `AS OF SYSTEM TIME` while these don't? 

---

*[pkg/acceptance/allocator_test.go, line 371 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvERz-Ff7kcZhWKhbyK:-KvERz-Ff7kcZhWKhbyL:b-ueiqv9) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L371)):*
> ```Go
> 		}
> 		if leftState && rightState {
> 			log.Infof(ctx, "problem seen over [%s, %s]", s.start, s.end)
> ```

Printing log(N) messages as we recurse seems noisy. 

---

*[pkg/acceptance/allocator_test.go, line 372 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvESBBJpUOJnBUOgEtg:-KvESBBJpUOJnBUOgEth:bpskdfz) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L372)):*
> ```Go
> 		if leftState && rightState {
> 			log.Infof(ctx, "problem seen over [%s, %s]", s.start, s.end)
> 			// pick the middle of the original span.
> ```

This is odd. If there were two inconsistencies at the start and end of the span, we may just drop the search here and never print anything more detailed. We need to recurse into both the left and right sides here instead of this "middle" span. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915)*

vivekmenezes:
Review status: 0 of 1 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/acceptance/allocator_test.go, line 245 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvENx0fGlcEc0_VR_bu:-KvPz1oma4-pZliuU1OO:b7nt1os) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L245)):*

Why are we starting a goroutine for each schema change if we want them to execute serially? Why not just one goroutine in a loop?
</blockquote></details>

The goroutine was added to allow for schema changes getting queued up and ensuring that queued up schema changes can wait for many minutes before execution.

---

*[pkg/acceptance/allocator_test.go, line 294 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEJOUCM_7DYbhy8MR7-r1-294:-KvQERgynjGhmUlWkgrB:b-axcqh2) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L294)):*

We should keep using `cluster_logical_timestamp` for the `as of system time` queries. (and we should always aim to get rid of arbitrary sleeps wherever we can) I think we can use a separate non-logical timestamp for the extra validation (if it matters that these are slightly different values, then the logical ticks will also be important and the test should be reworked to use HLC timestamps everywhere)
</blockquote></details>

Not sure what you're recommending here. 

---

*[pkg/acceptance/allocator_test.go, line 313 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQL47th-BB4DBk_Mv:-KvQ0TDuoMKmS0Y-2cQx:b-amcdhs) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L313)):*

Where does this `2` come from?
</blockquote></details>

removed the need for this magic number by moving this validation code above. I needed the check to run it once.

---

*[pkg/acceptance/allocator_test.go, line 314 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQOdSj_7UlTigBEJW:-KvQ0cBVp-mjWS0QS2jB:b-aeaujo) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L314)):*

Why this time span? Why not the whole database? (this is only run once at the end of the test, right?)
</blockquote></details>

It's only run at the end of the group of schema changes. I've changed this to the start time of the load generators on the  cluster.

---

*[pkg/acceptance/allocator_test.go, line 327 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEPb_UhkzEJ2wAIwIy:-KvQ1WjIB3Cws99ZaCPx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L327)):*

s/span/timeSpan/g
</blockquote></details>

Done.

---

*[pkg/acceptance/allocator_test.go, line 335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQxsF__oh4EhZ7sqg:-KvQ4OvE1WtWhQiiSDvK:b-b1873y) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L335)):*

Why are these queries hard-coded while the rest of the test uses queries that are passed in?  Why do the other validation queries use `AS OF SYSTEM TIME` while these don't? 
</blockquote></details>

I've moved them to the top where they get passed in. Since they validate an index using the timestamp stored in the row they do not need to use ```AS OF SYSTEM TIME```

---

*[pkg/acceptance/allocator_test.go, line 371 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvERz-Ff7kcZhWKhbyK:-KvQ4hfdaW74-8ge4FSM:bgvxanw) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L371)):*

Printing log(N) messages as we recurse seems noisy. 
</blockquote></details>

I've cleaned this up to be more sensible. The code now logs small time spans with problems and time spans without problems.

---

*[pkg/acceptance/allocator_test.go, line 372 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvESBBJpUOJnBUOgEtg:-KvPyIY796sKi3GpJXZ0:bhx2fks) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L372)):*

This is odd. If there were two inconsistencies at the start and end of the span, we may just drop the search here and never print anything more detailed. We need to recurse into both the left and right sides here instead of this "middle" span. 
</blockquote></details>

Ive changed this code to be more sensible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915)*

vivekmenezes:
@jordanlewis I'd really like to merge this so we can start running it as soon as possible. thanks

jordanlewis:
I will take a look this morning.

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/acceptance/allocator_test.go, line 245 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvENx0fGlcEc0_VR_bu:-KvSZ0EK3bjz9bJngkvc:bdjkg6c) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L245)):*

The goroutine was added to allow for schema changes getting queued up and ensuring that queued up schema changes can wait for many minutes before execution.
</blockquote></details>

One goroutine would also queue up the schema changes without introducing arbitrary delays. If we want to test a schema change that blocks for a while waiting for its turn it should probably be a separate test instead of getting tied in to this one.

---

*[pkg/acceptance/allocator_test.go, line 294 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEJOUCM_7DYbhy8MR7-r1-294:-KvSZgfie6zhMy-6yr1D:b-5nfrvo) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L294)):*

Not sure what you're recommending here. 
</blockquote></details>

1. It's not clear to me that `AS OF SYSTEM TIME` is required in `validationQueries`. If they are run after the system has reached a steady state, it shouldn't be necessary. If we do need `AS OF SYSTEM TIME` I think we also need it for `indexValidationQueries`.
2. If we need `AS OF SYSTEM TIME`, I think we also need to use `cluster_logical_timestamp()`, instead of client `time.Now()`. 

---

*[pkg/acceptance/allocator_test.go, line 335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQxsF__oh4EhZ7sqg:-KvS_rsvnZqpQN8Vn0sD:b-nsmb9y) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L335)):*

I've moved them to the top where they get passed in. Since they validate an index using the timestamp stored in the row they do not need to use ```AS OF SYSTEM TIME```
</blockquote></details>

That doesn't follow. `now()` timestamps can be inserted out of order from the perspective of an outside observer (we saw this in practice with the jepsen tests). If the table is receiving writes during this validation, we need to use `AS OF SYSTEM TIME` in all the validation queries; if not we don't need it anywhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915)*

vivekmenezes:

---

*[pkg/acceptance/allocator_test.go, line 245 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvENx0fGlcEc0_VR_bu:-KvTVTvhZ3D1CskQ0zq0:bpethik) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L245)):*

One goroutine would also queue up the schema changes without introducing arbitrary delays. If we want to test a schema change that blocks for a while waiting for its turn it should probably be a separate test instead of getting tied in to this one.
</blockquote></details>

yeah there isn't a good enough reason to queue up the schema changes here. I've removed the use of a goroutine here.

---

*[pkg/acceptance/allocator_test.go, line 294 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEJOUCM_7DYbhy8MR7-r1-294:-KvTXSV7u-9zxZobT2Et:b-2y0xj6) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L294)):*

1. It's not clear to me that `AS OF SYSTEM TIME` is required in `validationQueries`. If they are run after the system has reached a steady state, it shouldn't be necessary. If we do need `AS OF SYSTEM TIME` I think we also need it for `indexValidationQueries`.
2. If we need `AS OF SYSTEM TIME`, I think we also need to use `cluster_logical_timestamp()`, instead of client `time.Now()`. 
</blockquote></details>

We strictly do not need to use ```AS OF SYSTEM TIME``` . We use it to pick a snapshot of the database at which we can check whether the size of the table matches up to the size of the new column and index. Data is still being written to the database, so if were to not use it we'd have to run all the validation queries in a single transaction, and furthermore need to add in another loop to retry whenever they fail.

reverted back to use the cluster_logical_timestamp()

---

*[pkg/acceptance/allocator_test.go, line 335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-KvEQxsF__oh4EhZ7sqg:-KvTZ5RGuGgT3Md87SPe:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/148a973ebb9b4d750b06a2b4b42e6da458e00fce/pkg/acceptance/allocator_test.go#L335)):*

That doesn't follow. `now()` timestamps can be inserted out of order from the perspective of an outside observer (we saw this in practice with the jepsen tests). If the table is receiving writes during this validation, we need to use `AS OF SYSTEM TIME` in all the validation queries; if not we don't need it anywhere.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915)*

jordanlewis:
:lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18915#-:-KvThhkRXOnl2UY_5LCQ:bg7igkv)*

PR #18916, opened 2017-09-29
ui: fix undefined.replica_id errors

### LINKED ISSUE #18035: ui: TypeError while loading range debug page on broken range
On a cluster affected by #17989, trying to load up a problematic range on a functioning node only displays the message below. It may be related to the fact that node 1 doesn't is supposed to have a copy of replica 357 but hasn't yet processed the split that created it. This is annoying, but not incredibly urgent.

screen shot 2017-08-30 at 2 09 54 pm

BramGruneir:
And as a concequence of this, also display when a replica is awaiting being gced.

Fixes #18035.

### DIFF
--- a/pkg/ui/src/views/reports/containers/range/index.tsx
+++ b/pkg/ui/src/views/reports/containers/range/index.tsx
@@ -129,7 +129,10 @@ class Range extends React.Component<RangeProps, {}> {
         (info => RangeInfo.IsLeader(info)),
         (info => FixLong(info.raft_state.applied).toNumber()),
         (info => FixLong(info.raft_state.hard_state.term).toNumber()),
-        (info => RangeInfo.GetLocalReplica(info).replica_id),
+        (info => {
+          const localReplica = RangeInfo.GetLocalReplica(info);
+          return _.isNil(localReplica) ? 0 : localReplica.replica_id;
+        }),
       ],
       ["desc", "desc", "desc", "asc"],
     );
--- a/pkg/ui/src/views/reports/containers/range/print.ts
+++ b/pkg/ui/src/views/reports/containers/range/print.ts
@@ -7,8 +7,24 @@ import { LongToMoment, TimestampToMoment } from "src/util/convert";
 
 export const dateFormat = "Y-MM-DD HH:mm:ss";
 
-export function PrintReplicaID(rangeID: Long, rep: protos.cockroach.roachpb.ReplicaDescriptor$Properties) {
-  return `n${rep.node_id} s${rep.store_id} r${rangeID.toString()}/${rep.replica_id}`;
+// PrintReplicaID prints our standard replica identifier. If the replica is nil,
+// it uses passed in store, node and replica IDs instead.
+export function PrintReplicaID(
+  rangeID: Long,
+  rep: protos.cockroach.roachpb.ReplicaDescriptor$Properties,
+  nodeID?: number,
+  storeID?: number,
+  replicaID?: Long,
+) {
+  if (!_.isNil(rep)) {
+    return `n${rep.node_id} s${rep.store_id} r${rangeID.toString()}/${rep.replica_id}`;
+  }
+  // Fall back to the passed in node, store and replica IDs. If those are nil,
+  // use a question mark instead.
+  const nodeIDString = _.isNil(nodeID) ? "?" : nodeID.toString();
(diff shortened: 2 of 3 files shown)


BramGruneir:
Here you can see the new "AwaitingGC" replica issue.
screen shot 2017-09-29 at 15 24 21

I'd like to cherry pick this for 1.1, and check for over-replicated range and awaiting gc replicas on the server level for 1.2.

petermattis:
I don't see an urgent need for this in 1.1. Perhaps in 1.1.1. Maybe. We really need to raise the bar for cherry-picks.

BramGruneir:
The reason I think this should be for 1.1 is because if there is a replica awaiting GC, it will not display the page at all.

See the PM cluster's range 70 `/#/reports/range/70`, which current reports 

screen shot 2017-09-29 at 17 12 40

And looks like the earlier posted image with this change.

petermattis:
> The reason I think this should be for 1.1 is because if there is a replica awaiting GC, it will not display the page at all.

Yet this isn't a crash, data corruption, or query correctness problem. On the other hand, this change does only touch UI code and should only affect the debug pages which is a stronger argument for inclusion.

couchand:
:lgtm: Just adds some defensive coding to the debug pages, so seems pretty safe.  Since it will help us debugging issues in the field, I'd lean towards including, though I also agree with the general point about raising the bar for cherry-picks.

---

Reviewed 4 of 4 files at r1.

---

*[pkg/ui/src/views/reports/containers/range/index.tsx, line 134 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18916#-KvSTymyeVc-K8BoUABg:-KvSTymzGpfWU4XebYlW:b-hb9yta) ([raw file](https://github.com/cockroachdb/cockroach/blob/73c0896c5f3ce6753a2ff2fdc55f6353df27c587/pkg/ui/src/views/reports/containers/range/index.tsx#L134)):*
> ```XML
>         (info => {
>           const localReplica = RangeInfo.GetLocalReplica(info);
>           return _.isNil(localReplica) ? 0 : localReplica.replica_id;
> ```

aside: the second line is `R.propOr('replica_id', 0, localReplica)`

---

*[pkg/ui/src/views/reports/containers/range/print.ts, line 19 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18916#-KvSUHuTAYsUqCdAh4mZ:-KvSUHuTAYsUqCdAh4m_:b-gv029w) ([raw file](https://github.com/cockroachdb/cockroach/blob/73c0896c5f3ce6753a2ff2fdc55f6353df27c587/pkg/ui/src/views/reports/containers/range/print.ts#L19)):*
> ```XML
> ) {
>   if (_.isNil(rep)) {
>     const actualNodeID = _.isNil(nodeID) ? "?" : nodeID.toString();
> ```

nit: the names seem misleading here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18916#-:-KvSV-sMIIx2IA2V3s3d:b-ockstj)*

BramGruneir:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/ui/src/views/reports/containers/range/print.ts, line 19 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18916#-KvSUHuTAYsUqCdAh4mZ:-KvShTNImjpL6AnGsp-6:bsvntc0) ([raw file](https://github.com/cockroachdb/cockroach/blob/73c0896c5f3ce6753a2ff2fdc55f6353df27c587/pkg/ui/src/views/reports/containers/range/print.ts#L19)):*

nit: the names seem misleading here
</blockquote></details>

Changed their names, and moved the fallback processing to after the default return.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18916)*

BramGruneir:
TFTR.  I'll cherry pick this and we can bike shed on if this should be cherry picked to 1.1.0 or 1.1.1 in that PR.

---

Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18916#-:-KvShaYBzQiYHCyAcGF2:beso6r)*

PR #18917, opened 2017-09-29
cmd/docgen: check in generated function docs

dt:
switch to cobra commands and add a default output to the cockroach repo docs folder.
checking in the sources will allow CI to ensure that they are re-generated along with the
changes to the functions from which they are generated.

going forward, the docs repo will simply copy these in (or symlink them?) before building.


mjibson:
What is the plan for the CI changes? Can we include those here, so that these docs are generated during `make generate`?

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[.gitignore, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvEWOVkmoyR_y7Vlk2m:-KvEWOVkmoyR_y7Vlk2n:bxoloc4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c45ead528962289598cab5833a94b4e09726d82b/.gitignore#L27)):*
> ```
> # Generated files containing include paths.
> zcgo_flags*.go
> docgen
> ```

I don't understand the need for this.

---

*[pkg/cmd/docgen/funcs.go, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvEVUkVFF32h56pMCBq:-KvEVOJEzeJSnOmDZ_j-:b-zbr0ur) ([raw file](https://github.com/cockroachdb/cockroach/blob/c45ead528962289598cab5833a94b4e09726d82b/pkg/cmd/docgen/funcs.go#L37)):*
> ```Go
> 	Short: "generate markdown documentation of functions and operators",
> 	RunE: func(cmd *cobra.Command, args []string) error {
> 		outDir := "docs/generated/sql"
> ```

filepath.Join?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvEWSt2O9SpsNmTbWLE:b-9w8hzn)*

benesch:
:lgtm: modulo @mjibson's comments

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvEYKgiF_ZZBXZTWnK5:b-armo8f)*

dt:
as far as including an invocation in `make generate`:

maybe i'm mistaken, but AFAIK, the convention is that `go generate path/to/foo` (re)generates files _in foo_. If these docs live under `/docs` though, that would not be included be the default generation spec of `./pkg/...`.

Some possible approaches:
1) Change the default pkg spec for `make generate` to include `docs/...`.
2) `go generate pkg/sql/parser` is just weird and also generates files outside its path, in `docs/`.
3) Just have these files live under `pkg/sql/parser` instead of `docs/`
3.1) generate in pkg/sql/parser but have symlink between there and docs/... and pkg/sql/parser
4) not include these in bare `make generate`, and require explicit `make generate PKG=./docs/...` instead.

Of these, 1 or 4 feels the cleanest to me?

---

---

*[.gitignore, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvEWOVkmoyR_y7Vlk2m:-KvEql5GZwFdhjiTJYJY:bn1ek0z) ([raw file](https://github.com/cockroachdb/cockroach/blob/c45ead528962289598cab5833a94b4e09726d82b/.gitignore#L27)):*

I don't understand the need for this.
</blockquote></details>

I guess I was assuming a workflow of `go build ./pkg/cmd/docgen && ./docgen ...` in which case it could be annoying to have git see the `docgen` binary. we could suggest `go run` instead, but I always find its file vs package arguments clunkier to use.

---

*[pkg/cmd/docgen/funcs.go, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvEVUkVFF32h56pMCBq:-KvErEWn1MQBgGu4F8yc:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c45ead528962289598cab5833a94b4e09726d82b/pkg/cmd/docgen/funcs.go#L37)):*

filepath.Join?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvEtUNOe044zuhgE7Ir:b-k66noq)*

jseldess (commented):
Thanks, @dt. I don't understand exactly how how our workflows changes with this PR. Could you please document that?

knz:
My preference goes for option 1.

:+1: on documenting the process change in the commit message at the least, and also in either CONTRIBUTING.md or build/README.md, or both.

---

Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---

*[.gitignore, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvEWOVkmoyR_y7Vlk2m:-KvSfv704yPRQQ5RBj-n:bkfdm70) ([raw file](https://github.com/cockroachdb/cockroach/blob/c45ead528962289598cab5833a94b4e09726d82b/.gitignore#L27)):*

I guess I was assuming a workflow of `go build ./pkg/cmd/docgen && ./docgen ...` in which case it could be annoying to have git see the `docgen` binary. we could suggest `go run` instead, but I always find its file vs package arguments clunkier to use.
</blockquote></details>

I think I would be surprised (if not annoyed) that `go build ./pkg/cmd/docgen` produces a binary in the current directory instead of `bin/` or even `pkg/cmd/docgen`. What about you add `docgen` to the list of program dependencies in `common.mk` lines 162-179?

We already have the proper gitignore for that and the binary would live in a non-surprising location.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvSgiyIwoJVDHDNsvX7:b-xz9nqe)*

dt:
Eh, so I ended up going for 2) -- it isn't quite idiomatic `go:generate` but sql/parser already used Makefile-driven generation so it seemed like made sense to just drop in there. 

`go build` is going to drop a binary in cwd unless you hand it `-o`, but with the makefile managing it, i'm no longer concerned about using `go run` so I've killed the ignore.

I also added some comments in a `doc.go` explaining the files and how they're handled.

Thanks, RFAL.

knz:
Reviewed 3 of 3 files at r3.

---

*[pkg/sql/parser/doc.go, line 18 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSuXyPCxEOMqcw-BQO:-KvSuXyQSggMUFOitZ8v:b-hsl2ch) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/doc.go#L18)):*
> ```Go
> 
> /*
> Documentation of the SQL functions and operators is generated by the `docgen`
> ```

I had understood you wanted to use some new go:generate directives in this file.

If you are just writing prose, then please extend README.md instead.

---

*[pkg/sql/parser/Makefile, line 6 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSuhVjpe-W_gBGrwkn:-KvSuhVjpe-W_gBGrwko:b3yzuq4) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/Makefile#L6)):*
> ```Makefile
> SHELL = /usr/bin/env bash
> 
> TARGETS = sql.go keywords.go reserved_keywords.go helpmap_test.go help_messages.go doc.go
> ```

s/doc.go/extdocs

---

*[pkg/sql/parser/Makefile, line 79 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSujY1a_z3_W5A0guG:-KvSujY2ZT75KIKcGJc0:boefvqw) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/Makefile#L79)):*
> ```Makefile
> 	gofmt -s -w $@
> 
> doc.go:
> ```

```make
.PHONY: extdocs
extdocs:
     go run ../../cmd/docgen/*.go functions ../../../docs/generated/sql
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917)*

benesch:

---

*[pkg/sql/parser/Makefile, line 79 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSujY1a_z3_W5A0guG:-KvSx378tM4CKJjzVn9D:b-uk80s0) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/Makefile#L79)):*

```make
.PHONY: extdocs
extdocs:
     go run ../../cmd/docgen/*.go functions ../../../docs/generated/sql
```
</blockquote></details>

Given the amount of `..`s here, I think I'd prefer if this lived in the top level Makefile. I think we can make the claim that `extdocs` will eventually generate docs for more than just `pkg/sql/parser`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917)*

dt:
Review status: 5 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/sql/parser/Makefile, line 6 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSuhVjpe-W_gBGrwkn:-KvSy2EFjeyYVHutHdyi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/Makefile#L6)):*

s/doc.go/extdocs
</blockquote></details>

Done.

---

*[pkg/sql/parser/Makefile, line 79 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSujY1a_z3_W5A0guG:-KvSy50B-QZKKzmo9pCc:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/Makefile#L79)):*

```make
.PHONY: extdocs
extdocs:
     go run ../../cmd/docgen/*.go functions ../../../docs/generated/sql
```
</blockquote></details>

Done.

---

*[.gitignore, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvEWOVkmoyR_y7Vlk2m:-KvSy6RwMko4EKpzSneL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c45ead528962289598cab5833a94b4e09726d82b/.gitignore#L27)):*

I think I would be surprised (if not annoyed) that `go build ./pkg/cmd/docgen` produces a binary in the current directory instead of `bin/` or even `pkg/cmd/docgen`. What about you add `docgen` to the list of program dependencies in `common.mk` lines 162-179?

We already have the proper gitignore for that and the binary would live in a non-surprising location.
</blockquote></details>

Done.

---

*[pkg/sql/parser/doc.go, line 18 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSuXyPCxEOMqcw-BQO:-KvSy5_lW-msYbapv73g:btrjjk4) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/doc.go#L18)):*

I had understood you wanted to use some new go:generate directives in this file.

If you are just writing prose, then please extend README.md instead.
</blockquote></details>

I did originally, but then ended up just using the Makefile since it is already managing generation for the package.

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917)*

dt:
Review status: 5 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/Makefile, line 79 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-KvSujY1a_z3_W5A0guG:-KvT4okGahDS8DVlwv6O:b-jzufpl) ([raw file](https://github.com/cockroachdb/cockroach/blob/700532928dd394a51f9b1abd3ddb4185aedf49fa/pkg/sql/parser/Makefile#L79)):*

Done.
</blockquote></details>

Hm, went with this since it is already handling generating user-facing docs (the help text) from the parser, and function docs so it seemed like a good fit (particularly for the diagrams). I kind of like that existing workflows that include regenerating the parser will just start regenerated the parser-derived docs as well without needing to learn a new step. If it is just the `../../` you don't like, we could also move cmd/docgen under pkg/sql/parser if it is only intended to be  run by the makefile rather than by hand.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917)*

mjibson:
:lgtm:

I have no opinion on where the go:generate or Make stuff lives. Whatever makes sense.

---

Review status: 5 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvTHTO99wUpOOKtXuYX:b8oqy5y)*

benesch:
@dt sgtm. Trying to avoid another recursive Make mess a la protobuf.mk but
this is fine by me for now!

dt:
Hmm. The diagram generation takes several minutes, so I might have changed my mind on including it in `pkg/sql/parser`'s default `make`.

mjibson:
Oh yeah. And if we are going to run this for each PR, then we need to install the railroad generator and java on the agents, because we shouldn't be hitting the generator website that much. Maybe we can punt on auto-generating.

dt:
I just meant that the time required is an argument for a standalone, top-level command like @benesch suggested, that you'd explicitly run when you're done iterating on a `parser` change, since iterating on, and particularly rebasing, a parser change frequently involves re-running yacc quite a bit, and I'd go crazy if that included a multi-minute docgen step.

dt:
Okay, in light of the realization that diagram generation is far too slow to include in regular `parser` code generation, i've removed docgen from the `pkg/sql/parser` gen, since that is usually run repeatedly while iterating on a change. Indeed, it's probably too slow even for inclusion in a bare `make generate`, so I've just updated the CI invocation to add `docs/...` to its PKG.

mjibson:
:lgtm:

---

Review status: 5 of 9 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvTfbakGeFrlQO4Nmn4:bnfp4nl)*

benesch:
:lgtm:

---

Review status: 5 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18917#-:-KvTrOGp2d8edV1mmE58:bnfp4nl)*

Issue #17597, opened 2017-08-11
sql: get rid of the Session.TableCollection


andreimatei:
Currently the Session has a TableCollection member: https://github.com/cockroachdb/cockroach/blob/57423bafbfda80679cab22204b8c4ccb436297c6/pkg/sql/session.go#L248

It contains descriptors and leases used by the current transaction. I believe the idea was, at some point, that transactions should reuse descriptors.
I think it'd be great if we managed to get rid of this member. First of all, if this structure is per-transaction and not per-session, then it's placed in the wrong place.
More importantly, as of recently, we're good about giving correct descriptors to transactions based on the transaction's timestamp. I'd be great if this stayed the single mechanism by which correct descriptors are used.

There might be some considerations around this TableCollection around table name -> table id mappings. There I forgot what the story is and I don't know what transactional considerations apply as tables get renamed. But in any case, this doesn't have to do with descriptors themselves.

vivekmenezes:
@andreimatei TableCollection is holding a lot of stuff and I'm planning to add more :-) so perhaps we should just close this issue

andreimatei:
Oh no :)
Any thoughts on the points, though? 
> First of all, if this structure is per-transaction and not per-session, then it's placed in the wrong place.
Also everything about the timestamp being sufficient.

Basically, perhaps except for uncommitted tables, why does the thing need to exist?

vivekmenezes:
well I'm working on a change that's going to add uncommittedCreatedTables to this data structure. I thing the way this works is you use it as a dumping ground until it becomes unwieldy and then someone comes around and breaks it up and renews order. But I still think that doesn't warrant an issue.

andreimatei:
The problem is that it's not clear (at least not to me) what stuff is dumped in this dumping ground. The last instance where I ran into this  is in relation with "internal statements" - when one executes something using the `InternalExecutor` in the context of a session, does the internal executor need to share the `TableCollection`? What if it further executes the statement in the context of the current txn? I've got no idea. Stemming for the fact that it's not clear to me why it exists in the first place. Any documentation you can add would be most valuable.

lucy-zhang:
Closing this because it doesn't really reflect our current plans for the TableCollection anymore - I think the current plan is to move it into a separate leasing/caching-related package and have a more principled API for it, as a way to store leased objects and uncommitted objects. "Getting rid of it" is probably too strong. cc @ajwerner

PR #18918, opened 2017-09-29
sql: fix stale txn deadline

andreimatei:
Touches #18684

Before this patch, we were relying to TableCollection to reset a txn's
deadline when it detected that the transaction had restarted and a new
set of descriptors needed to be acquired. However, there was a bug:
Executor.execSchemaChanges() would reset the TableCollection in a way
that prevented it from resetting the txn deadline in the future (the
idea is that tc.timestamp was being set to nil, which disarmed the txn
deadline reset code).
This meant that the txn continued with the old deadline after restarting
and so, if it ended up being pushed above the original deadline, it
couldn't commit.

This patch moves the deadline reset out of TableCollection, into the
Executor. The TableCollection's role in resetting the deadline was
questionable to begin withi (as opposed to the TC's role in _updating_
the timestamp, that one is good). Now the reset is tied to a SQL
transaction restart.

### DIFF
--- a/pkg/sql/session.go
+++ b/pkg/sql/session.go
@@ -1243,6 +1243,7 @@ func (ts *txnState) updateStateAndCleanupOnErr(err error, e *Executor) error {
 		// Note that TransactionAborted is also a retriable error, handled here;
 		// in this case cleanup for the txn has been done for us under the hood.
 		ts.SetState(RestartWait)
+		ts.mu.txn.ResetDeadline()
 	}
 	return err
 }
@@ -1314,8 +1315,6 @@ func (scc *schemaChangerCollection) queueSchemaChanger(schemaChanger SchemaChang
 func (scc *schemaChangerCollection) execSchemaChanges(
 	ctx context.Context, e *Executor, session *Session,
 ) error {
-	// Release the leases once a transaction is complete.
-	session.tables.releaseTables(ctx)
 	if e.cfg.SchemaChangerTestingKnobs.SyncFilter != nil {
 		e.cfg.SchemaChangerTestingKnobs.SyncFilter(TestingSchemaChangerCollection{scc})
 	}
--- a/pkg/sql/table.go
+++ b/pkg/sql/table.go
@@ -268,7 +268,6 @@ type TableCollection struct {
 // of a transaction retry.
 func (tc *TableCollection) resetForTxnRetry(ctx context.Context, txn *client.Txn) {
 	if tc.timestamp != (hlc.Timestamp{}) && tc.timestamp != txn.OrigTimestamp() {
-		txn.ResetDeadline()
 		tc.releaseTables(ctx)
 	}
 }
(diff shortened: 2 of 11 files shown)


andreimatei:
I'll cherry-pick this in 1.1.

bdarnell:
:lgtm:

---

Reviewed 1 of 1 files at r1, 10 of 10 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18918#-:-KvJigf-E9Xw1ONHh4kw:bnfp4nl)*

pkg/sql/session.go
```diff
@@ -1243,6 +1243,7 @@ func (ts *txnState) updateStateAndCleanupOnErr(err error, e *Executor) error {
 		// Note that TransactionAborted is also a retriable error, handled here;
 		// in this case cleanup for the txn has been done for us under the hood.
```
vivekmenezes: Can we do the reset in this under the hood cleanup?

pkg/sql/table.go
```diff
@@ -268,7 +268,6 @@ type TableCollection struct {
 // of a transaction retry.
 func (tc *TableCollection) resetForTxnRetry(ctx context.Context, txn *client.Txn) {
 	if tc.timestamp != (hlc.Timestamp{}) && tc.timestamp != txn.OrigTimestamp() {
-		txn.ResetDeadline()
 		tc.releaseTables(ctx)
```
vivekmenezes: Can you add a TODO to remove tc.timestamp . I think we should have the txn retry always call releaseTables

vivekmenezes (approved):

andreimatei:
Raphael, what kind of example do you want to see in the commit message, more concretely? I think the explanation of the problem is clear: "This meant that the txn continued with the old deadline after restarting
and so, if it ended up being pushed above the original deadline, it couldn't commit."

---

Review status: 1 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/session.go, line 1244 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18918#-KvEvQqEha0U_hsvlpBi-r2-1244:-KvShLuljMLUc-6uXfId:b-2togj) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ce9e77f2ea57faf0b3760d6e8a7dad97ab9b6fd/pkg/sql/session.go#L1244)):*

Can we do the reset in this under the hood cleanup?
</blockquote></details>

not really. The "under the hood cleanup" happens in `TxnCoordSender`, which doesn't have access to the `client.Txn`.

---

*[pkg/sql/table.go, line 271 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18918#-KvEvQqEha0U_hsvlpBj-r2-271:-KvSiQokYrvTnEfeeO1C:bie88hy) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ce9e77f2ea57faf0b3760d6e8a7dad97ab9b6fd/pkg/sql/table.go#L271)):*

Can you add a TODO to remove tc.timestamp . I think we should have the txn retry always call releaseTables
</blockquote></details>

I had filed #17597 a while ago about figuring out what TableCollection generally is and whether it has a point (I don't think it does). I'd rather do something about the whole structure rather than dealing with the timestamp field in particular.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18918#-:-KvSjB0fndphxPaOqPdo:btq1j9y)*

vivekmenezes:
LGTM

PR #18920, opened 2017-09-30
sql/sqlbase: convert txnKVFetcher to use roachpb.KeyValue

petermattis:
For historical reasons, txnKVFetcher was converting roachpb.KeyValue
into client.KeyValue. Switch to using roachpb.KeyValue to avoid the
conversion.

Switch to using the count of keys in the response header instead of
asking for the length of returned key/value pairs slice which results in
slightly simpler code.


danhhz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18920#-:-KvSnQdKAX-E7dClXLgn:bnfp4nl)*

PR #18925, opened 2017-09-30
build: adjust build tag

petermattis:
Current master reports the build tag as
`v1.1-alpha.20170817-1197-g081f0bab8`. This is quite confusing given
that the most recent 1.1 release is actually v1.1-beta.20170928. Switch
back to using `git describe --exact-match` and generate our own tag
format: `unstable-<date>-<sha>`. With this change, the build tag is
reported as: `unstable-20170929-081f0bab8`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -350,7 +350,9 @@ $(ARCHIVE).tmp: .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
 # For details, see the "Possible timestamp problems with diff-files?" thread on
 # the Git mailing list (http://marc.info/?l=git&m=131687596307197).
 .buildinfo/tag: | .buildinfo
-	@{ git describe --tags --dirty 2> /dev/null || git rev-parse --short HEAD; } | tr -d \\n > $@
+	@{ git describe --tags --exact-match 2>/dev/null || \
+	  echo v1.x-unstable.$$(git show -s --date=format:'%Y%m%d' --format=%cd HEAD)-$$(git rev-parse --short HEAD); } | \
+	  tr -d \\n > $@
 
 .buildinfo/basebranch: | .buildinfo
 	@git describe --tags --abbrev=0 | tr -d \\n > $@


petermattis:
This changes the behavior from #18565. I don't see how to use `git describe` without `--exact-match` and have it give something reasonable. Perhaps somebody else has a better idea.

petermattis:
Note that this causes "unstable" build tags to sort before releases and alphas. For example:

```
unstable.20170930-a19c0a5
v1.0.6
v1.1-beta.20170921
v1.1-beta.20170928
```

We could make "unstable" more prominent and have it sort afterward by doing:

```
v1.0.6
v1.1-beta.20170921
v1.1-beta.20170928
vUNSTABLE.20170930-a19c0a5
```

knz:
I don't think this approach is constructive - it adds code where a simple solution is possible.

I think instead we should place a tag on the `master` branch called `v1.x-current`, reset to the latest split point each time we make a release fork. This way the build tag on master becomes `v1.x-current.NNNN` which is reasonable, until we tag `v1.2-alpha` when it will become even further reasonable.

petermattis:
> I don't think this approach is constructive - it adds code where a simple solution is possible.

Felt constructive to me. How else would you have preferred this discussion take place?

I'm agnostic towards the actual solution, though I'm curious if the `v1.1-alpha.20170817-1197-g081f0bab8` is readable by anyone else. Without looking at the docs, what does `-1197-` indicate? And what does the `g` at the start of the sha indicate? I think there is value in including the date of the commit in these current/unstable builds. With the tag approach you've proposed we'd have `v1.x-current-1197-g081f0bab8`. Yes, I can easily look up the commit date from the sha, but we have tools such as `roachperf` which use the build tags for naming the output directory. 

Let the bikeshedding begin.

knz:
On 30-09-17 13:04, Peter Mattis wrote:
> How else would you have preferred this discussion take place?

Hah. I hadn't considered that. Thanks for this thoughtful reminder.

> I'm agnostic towards the actual solution, though I'm curious if the
> |v1.1-alpha.20170817-1197-g081f0bab8| is readable by anyone else.

I think as more people become seasoned with git over time, this concern
will dissipate.

> Without looking at the docs, what does |-1197-| indicate?

Once a dev sees the value increase with each commit they make, they'll
get the memo quickly. Users will simply notice the value increases over
time and correlates with "how much was changed". This correlation may
not be conscious/documented but I think it will emerge quickly, and I
find it useful.

> And what does the |g| at the start of the sha indicate?

Straight from the man page for "git describe":

"The "g" prefix stands for "git" and is used to allow describing the
version of a software depending on the SCM the software is managed with.
This is useful in an environment where people may use different SCMs."

> I think there is value in
> including the date of the commit in these current/unstable builds.

I'm open to this idea, but FWIW I never had that need myself.

> [...] we have tools such as |roachperf| which use the
> build tags for naming the output directory.

1) the build tag starting with "v1.N" does not give you a "natural order
= lexical order" either once N goes from 9 to 10.

2) once you start deriving your sort key using an additional sed, the
commit count can be used to sort the directory names in "ascending
commit order" just as well as a timestamp.


-- 
Raphael 'kena' Poss

petermattis:
> > I'm agnostic towards the actual solution, though I'm curious if the
> > |v1.1-alpha.20170817-1197-g081f0bab8| is readable by anyone else.
>
> I think as more people become seasoned with git over time, this concern will dissipate.

Perhaps. I've seen this format many times and always have to refer to the documentation. Repeated viewings have not brought understanding. 

Let's wait for some other people to chime in.

bdarnell:
We should maintain the convention that all tags starting with `v` are correctly sortable by semver rules, so if we want to place a tag on master at the branch split point it would probably need to be `v1.2-alpha.20170824`. Or I suppose it could be something that doesn't start with `v` at all. 

FWIW, I like the `git describe` tag (except for the `g`). However, I was recently confused by the fact that magit was showing my status in a similar format, and thought something had gone wrong since it was using alpha instead of the most recent tag. 

Date of last commit can be misleading: cherry-picks keep their original date (although on master where the head commit should always be a merge, this wouldn't be a problem). However, it seems like a potentially useful thing to include. I'd keep the `git describe` output in the "tag" field, though, and add a separate "date of last commit" field (just like we have build date as a separate build-info field).

petermattis:
> Date of last commit can be misleading: cherry-picks keep their original date (although on master where the head commit should always be a merge, this wouldn't be a problem). However, it seems like a potentially useful thing to include. I'd keep the git describe output in the "tag" field, though, and add a separate "date of last commit" field (just like we have build date as a separate build-info field).

As long as the cherry-pick was merged using a merge commit, the tag will contain the date of the merge commit, not the cherry-pick. I just verified this.

bdarnell:
Yes, but if I'm making a custom build from a non-release branch I'm probably not using merge commits.

petermattis:
> Yes, but if I'm making a custom build from a non-release branch I'm probably not using merge commits.

Good point. I still prefer the `v1.2-unstable` approach in this PR as it will provide a more accurate date that `git describe --tags`, but I'm willing to concede defeat.

Should I create a tag for `v1.2-alpha`? And at which sha? And should I strip the `g` from the output of `git describe --tags`?

knz:
On 01-10-17 20:02, Peter Mattis wrote:
> Should I create a tag for |v1.2-alpha|? 

I like the idea.

> And at which sha?

The branch split point perhaps?

> And should I
> strip the |g| from the output of |git describe --tags|?

Please don't!

The full string generated by `git describe` including the `g` is a valid
input to `git log`, `git checkout`, `git cherry-pick` etc. If you strip
the `g`, `git` doesn't recognize the string as a commit id.

-- 
Raphael 'kena' Poss

bdarnell:
Yeah, don't remove the `g`. I wouldn't include it if i were designing `git describe` in a vacuum, but now that it's here we should just use it instead of inventing an almost-but-not-quite-compatible format.

Creating a `v1.2-alpha` tag at the split point sounds fine to me.

petermattis:
What is the split point? `git merge-base master upstream/release-1.1`?

bdarnell:
Yes (8633ca46d53c987b229562c0652694c53f3bae79)

petermattis:
Ok, done.

PR #18928, opened 2017-09-30
sql: don't capture contextHolder closure in EvalContext

jordanlewis:
Previously, evalContext stored a closure that returns a context.Context.
This caused unnecessary allocations when creating a new EvalCtx via
session.evalCtx. Instead of keeping a closure, keep a CtxProvider
interface that Session now implements.


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18928#-:-KvJXu8i_K49458tXOjQ:bnfp4nl)*

Issue #18919, opened 2017-09-30
perf: quantify network space overhead


petermattis:
Similar to #18659 I decided to look at the overhead in `Batch{Request,Response}` for simple "kv" reads (i.e. a single `Scan{Request,Response}` in the batch). 

```
  BatchRequest:               191
  BatchRequest.Header:        156
  BatchRequest.Header.Txn:    122
  BatchRequest.Requests[0]:   30
  BatchResponse:              189
  BatchResponse.Header:       138
  BatchResponse.Header.Txn:   122
  BatchResponse.Responses[0]: 46
```

The `Scan{Request,Response}` consume exactly 20% of the total bytes sent/received. The request and response txn consume 64% of the bytes. The response txn is nearly identical to the request txn with only `Transaction.MaxTimestamp` having changed. 

There are some small gains to be had by converting `Batch{Request,Response}`, `Transaction` and `TxnMeta` to proto3 to avoid encoding zero value primitive fields. There is a large gain to be had by only returning the delta of the transaction in the response. 

Cc @andreimatei who suggested last week that we develop a more principled way to update the transaction upon receiving a response.

tschottdorf:
+1 to only returning a transaction delta (at least for common cases).

petermattis:
@tschottdorf Do you have time to tackle only sending a transaction delta in the response? I think you're the most knowledgeable person about this code. If not, perhaps you can give guidance to someone else. Feels like a fairly small project, though I might be missing subtleties.

tschottdorf:
Ack, assigning.

andreimatei:
I think we probably agree that what should be returned by the server is not a brute "delta", but a specific set of instructions about what the server wants the client to do (to the transaction).

petermattis:
@andreimatei That might be cleaner, though it also might add an unfortunate migration story.

tschottdorf:
@andreimatei I was picturing something like the below which gets terminated at `TxnCoordSender` (i.e. `client.Txn` would still see the full proto). This is provisional and I have to figure out what to stick in the error messages.

I'm not too worried about the migration with regards to correctness. We can add the new field and populate it only when the cluster version indicates this is safe.

Let me know if you pictured something different.

```proto
message TxnMetaDelta {
  // Never needed.
  // optional bytes id = 1;

  // Never changes.
  // optional IsolationType isolation = 2;

  // Never returned in a reply.
[...]

PR #5891, opened 2016-04-06
all: migrate to proto3 where possible

tamird:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach/5891)

PR #18930, opened 2017-10-01
roachpb: convert api.proto and data.proto to proto3

petermattis:
This shaves 6% off the encoded size of `Batch{Request,Response}` for a
simple scan operation.

See #18919


petermattis:
This touches enough files that we should probably hold off on merging for a few weeks. Also, should I convert more files to `proto3`? `api.proto` contains the most important conversions as it touches RPC messages that contain primitive fields.

bdarnell:
I'm nervous about this - we've broken things by making changes like this before (although at least this would be at the beginning of a cycle). See #5891 for more debate.

`api.proto` and `data.proto` are the big risks - if we get them converted to proto3 we might as well go all the way and get everything moved over. But it might be safer to leave them as-is and convert only selected messages to proto3 so we can give the change more scrutiny (Transaction and TxnMeta look like the biggest wins).

petermattis:
#5891 was pre proposer-evaluated kv so there was a lot more stuff below raft. That said, recent changes to protos have proven fragile. I'd go so far as to say that this PR will almost certainly a surprising/subtle bug. Not sure how to mitigate that.

bdarnell:
Yeah, it's gotten much better with propEvalKV (we didn't even have the below-raft-protos test when that was first proposed). They're even talking about bringing unknown-field preservation back to proto3, which counters one of my proto2-forever arguments. 

A 6% space savings isn't huge and most of that probably comes from a couple of messages, so we could minimize the risk by just not changing most of the protos. Or if we decide to go 100% proto3, we should probably go ahead and rip off the bandaid.

petermattis:
Agreed that the 6% savings isn't huge, but I think it paves the way for only sending back a `Transaction` with the fields we want to update. So yeah, just converting `TxnMeta` and `Transaction` would be the minimal approach, though `roachpb.Header` might be useful too.

petermattis:
I took another look at this PR this morning, and my preference is to convert more protos to proto3 rather than less. Proposer-evaluated KV means we have relatively few protos that need to remain proto2. Yeah, it is possible that we'll have some headaches in the transition, but I think they'll be minor. Converting a field from not-nullable to nullable is more problematic.

I've added another commit which converts `TxnMeta` to proto3.

petermattis:
@bdarnell PTAL

bdarnell:
:lgtm:

---

Reviewed 9 of 20 files at r1, 37 of 37 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18930#-:-KxKKy8uvLRw18cfqgIP:bnfp4nl)*

PR #18939, opened 2017-10-02
*: add protoutil.MarshalTo

petermattis:
Get rid of the few random calls to protoutil.Interceptor by using
protoutil.MarshalTo. Previously this had not been done because of an
observed slowdown in BenchmarkValueSetProto. This slow down was caused
by the additional interface conversions. Avoid such conversions by
adding protoutil.Message which extends proto.Message with MarshalTo and
Size methods. Note that justification for this change is the removal of
the random calls to protoutil.Interceptor, not the small performance
improvement.

name             old time/op  new time/op  delta
ValueSetProto-8  45.3ns ± 6%  38.0ns ± 3%  -15.94%  (p=0.000 n=10+9)

### DIFF
--- a/pkg/config/migration_test.go
+++ b/pkg/config/migration_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
+	"github.com/cockroachdb/cockroach/pkg/util/protoutil"
 	"github.com/gogo/protobuf/proto"
 )
 
@@ -26,7 +27,7 @@ func TestMigrateZoneConfig(t *testing.T) {
 	defer leaktest.AfterTest(t)()
 
 	testCases := []struct {
-		input, want proto.Message
+		input, want protoutil.Message
 	}{
 		{
 			&ZoneConfig{
--- a/pkg/gossip/gossip.go
+++ b/pkg/gossip/gossip.go
@@ -848,7 +848,7 @@ func (g *Gossip) addInfoLocked(key string, val []byte, ttl time.Duration) error 
 
 // AddInfoProto adds or updates an info object. Returns an error if info
 // couldn't be added.
-func (g *Gossip) AddInfoProto(key string, msg proto.Message, ttl time.Duration) error {
+func (g *Gossip) AddInfoProto(key string, msg protoutil.Message, ttl time.Duration) error {
 	bytes, err := protoutil.Marshal(msg)
 	if err != nil {
 		return err
--- a/pkg/internal/client/util.go
+++ b/pkg/internal/client/util.go
@@ -19,11 +19,10 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/gogo/protobuf/proto"
-
 	"github.com/cockroachdb/apd"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/util/duration"
+	"github.com/cockroachdb/cockroach/pkg/util/protoutil"
 )
 
 // TODO(pmattis): The methods in this file needs tests.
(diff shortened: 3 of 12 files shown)


nvanbenschoten:
Do we need a lint to avoid any uses of `proto.Message` going forward?

---

---

*[pkg/storage/engine/mvcc_test.go, line 3326 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-KvSlbL9WR_5F31w2v5p:-KvSlbL9WR_5F31w2v5q:bh5cgye) ([raw file](https://github.com/cockroachdb/cockroach/blob/af656e6bd4dbe6d1224e805399b143ebe9024a33/pkg/storage/engine/mvcc_test.go#L3326)):*
> ```Go
> 
> // encodedSize returns the encoded size of the protobuf message.
> func encodedSize(msg protoutil.Message, t *testing.T) int64 {
> ```

No need for `t *testing.T` anymore. Actually, this entire function is pretty useless now.

---

*[pkg/util/protoutil/marshal.go, line 46 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-KvSm-OnkEVhbcrA55cS:-KvSm-OnkEVhbcrA55cT:bxk2q0d) ([raw file](https://github.com/cockroachdb/cockroach/blob/af656e6bd4dbe6d1224e805399b143ebe9024a33/pkg/util/protoutil/marshal.go#L46)):*
> ```Go
> // base to intercept calls to pb.MarshalTo.
> func MarshalTo(pb Message, dest []byte) (int, error) {
> 	Interceptor(pb)
> ```

Is `Interceptor` always set? Seems like it should be avoided when possible if it's just going to be a no-op anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-:-KvSmC2PqCC5wz6_O4cx:bq2rts4)*

petermattis:

---

*[pkg/storage/engine/mvcc_test.go, line 3326 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-KvSlbL9WR_5F31w2v5p:-KvSrJGD2w4B7v2IwiIv:biq7tcn) ([raw file](https://github.com/cockroachdb/cockroach/blob/af656e6bd4dbe6d1224e805399b143ebe9024a33/pkg/storage/engine/mvcc_test.go#L3326)):*

No need for `t *testing.T` anymore. Actually, this entire function is pretty useless now.
</blockquote></details>

Good point. Removed it completely.

---

*[pkg/util/protoutil/marshal.go, line 46 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-KvSm-OnkEVhbcrA55cS:-KvSrT53bCugPvLXfVd6:biwd8qy) ([raw file](https://github.com/cockroachdb/cockroach/blob/af656e6bd4dbe6d1224e805399b143ebe9024a33/pkg/util/protoutil/marshal.go#L46)):*

Is `Interceptor` always set? Seems like it should be avoided when possible if it's just going to be a no-op anyway.
</blockquote></details>

Yes, it is always set. Let me see if it makes a difference to avoid calling it...

Nope, I can't measure a difference, so I'm going to leave as-is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939)*

bdarnell:
:lgtm: 

A lint sounds like a good idea, although we do touch a few protobufs (notably from the `raftpb` package) that only implement `proto.Message`, not `protoutil.Message`. I think we always refer to them by their concrete type instead of the `Message` interface, though.

---

Reviewed 11 of 12 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-:-KvSwEzSeO5oXkxZjlvS:b98h4d)*

petermattis:
The various `raftpb` protos implement `protoutil.Message` or this wouldn't have compiled. For example, `raftpb.HardState.{MarshalTo,Size}` exist.

We still have quite a few uses of `proto.Message`. The issue is with the read path where we'd have to add a `protoutil.Unmarshal` or go through an interface conversion. Let me follow up with another PR which adds `protoutil.Unmarshal` and adds the lint. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18939#-:-KvSx3QGc1w59JCWLYf2:bd94eow)*

PR #18941, opened 2017-10-02
cherrypick-1.1 ui: fix undefined.replica_id errors

### LINKED ISSUE #18035: ui: TypeError while loading range debug page on broken range

BramGruneir:
Cherry-picking #18916
cc @cockroachdb/release

And as a consequence of this, also display when a replica is awaiting being gced.

Fixes #18035.

### DIFF
--- a/pkg/ui/src/views/reports/containers/range/index.tsx
+++ b/pkg/ui/src/views/reports/containers/range/index.tsx
@@ -129,7 +129,10 @@ class Range extends React.Component<RangeProps, {}> {
         (info => RangeInfo.IsLeader(info)),
         (info => FixLong(info.raft_state.applied).toNumber()),
         (info => FixLong(info.raft_state.hard_state.term).toNumber()),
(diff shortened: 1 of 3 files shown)


BramGruneir:
There is some debate as to if this is valid for 1.1.  I think it is, as currently the range debug page will fail to load on any range that has some replicas awaiting GC.

couchand (approved):

bdarnell:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18941#-:-KvT0uYUHPCcWkkYd0CT:bnfp4nl)*

PR #18942, opened 2017-10-02
distsql: introduce float and reduce precision for float/decimal comparison in TestDistAggregationTable

### LINKED ISSUE #18691: sql/distsqlplan: TestDistAggregationTable/VARIANCE/int1/50 failed under stress

### LINKED ISSUE #18693: sql/distsqlplan: TestDistAggregationTable/VARIANCE/dec1/100 failed under stress
SHA: https://github.com/cockroachdb/cockroach/commits/1d5c72d03ced1cb44c3f00dc8a656a5fb347f18b

Parameters:
```
TAGS=
GOFLAGS=-race
```

Stress build found a failed test: https://teamcity.cockroachdb.com/viewLog.html?buildId=358214&tab=buildLog

```

```

richardwu:
Should fix #18691, #18693, #18921, #18923 

This PR introduces two modifications to TestDistAggregationTable:
1. Introduce floating point tests
2. Reduce precision required for float and decimal comparisons

The former is self-explanatory.

RE the latter: previously we observed that stress testing results for
multi-stage (local) vs single-stage (non-local) distributed aggregations
exhibited off-by-last-digit errors.

An attempt to remedy this issue by increasing the precision of
intermediate calculations was introduced in #18701.

Recent nightly stress tests reveal that in some cases, the non-local
results may [truncate at precision 19](https://teamcity.cockroachdb.com/viewLog.html?buildId=365868&tab=buildLog&_focus=1200) or still have a [20th precision off-by-one discrepancy](https://teamcity.cockroachdb.com/viewLog.html?buildId=364882&tab=buildLog&_focus=1371) with the local results.

This is inherent in the running algorithms we use for local vs non-local
aggregation calculations of `STDDEV` and `VARIANCE`, and I don't see any
other way to avoid this generally besides reducing the precision we
compare at in our tests.

Float results are even more skewed hence we reduce the precision for
comparison to 10 (after the decimal point; arbitrarily decided such that
test cases pass almost always w/ confirmation bias). Note floating point
operations are expected to involve significant precision loss.

### DIFF
--- a/pkg/sql/distsqlplan/aggregator_funcs_test.go
+++ b/pkg/sql/distsqlplan/aggregator_funcs_test.go
@@ -32,6 +32,18 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/uuid"
 )
 
+var (
+	// compareCtx is a decimal context with reduced precision for comparing decimal
+	// results.  For some decimal aggregate operations there will often be
+	// unavoidable off-by-last-digit errors, which is semantically okay.
+	compareCtx = parser.DecimalCtx.WithPrecision(parser.DecimalCtx.Precision - 1)
+	// floatPrecFmt is the format string with a precision of 10 (after
+	// decimal point) specified for float comparisons. Float aggregation
+	// operations involve unavoidable off-by-last-few-digits errors, which
+	// is expected.
+	floatPrecFmt = "%.10f"
+)
+
 // runTestFlow runs a flow with the given processors and returns the results.
 // Any errors stop the current test.
 func runTestFlow(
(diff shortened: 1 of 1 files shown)


mjibson:

---

*[pkg/sql/distsqlplan/aggregator_funcs_test.go, line 267 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTK1_FO_TsxL3LDWeW:-KvTK1_FO_TsxL3LDWeX:b3z6hz2) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/distsqlplan/aggregator_funcs_test.go#L267)):*
> ```Go
> 			rowDist := rowsDist[0][i]
> 			rowNonDist := rowsNonDist[0][i]
> 			if rowDist.Type.String() != rowNonDist.Type.String() {
> ```

Can we use .FamilyEqual instead of comparing strings?

---

*[pkg/sql/distsqlplan/aggregator_funcs_test.go, line 268 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTKDR7Vail2HbDHFQQ:-KvTKDR7Vail2HbDHFQR:b-hyazzg) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/distsqlplan/aggregator_funcs_test.go#L268)):*
> ```Go
> 			rowNonDist := rowsNonDist[0][i]
> 			if rowDist.Type.String() != rowNonDist.Type.String() {
> 				t.Errorf("different type for column %d (dist: %s non-dist: %s)", i, rowDist.Type.String(), rowNonDist.Type.String())
> ```

t.Fatalf here probably

---

*[pkg/sql/distsqlplan/aggregator_funcs_test.go, line 272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTJbbwaQVO4xtKBgC9:-KvTJbbwaQVO4xtKBgCA:b-3zsyf5) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/distsqlplan/aggregator_funcs_test.go#L272)):*
> ```Go
> 
> 			var equiv bool
> 			switch rowDist.Datum.(type) {
> ```

use `switch t := rowDist.Datum.(type) {` here, then you can use `t` below instead of re-casting rowDist.Datum

---

*[pkg/sql/parser/decimal.go, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTIs5kAuGpHrM2wqwH:-KvTIs5kAuGpHrM2wqwI:bn0ortl) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/parser/decimal.go#L44)):*
> ```Go
> 	// operations there will often be unavoidable off-by-last-digit errors,
> 	// which is semantically okay.
> 	CompareCtx = DecimalCtx.WithPrecision(DecimalCtx.Precision - 1)
> ```

Since this is only used in a test, define it in the test instead of globally.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942)*

richardwu:
I also introduced float (non-NULL and NULL) columns in the test (after revisiting my comment that float operations are unaffected, they are indeed severely imprecise and I've also reduced the precision required for its tests). I've set a conservative precision for floats from what I've observed works for local tests, but they may fail when we stress test it with a ton of inputs.

I'll revisit that when the time comes but for now I believe a precision of 10 after the decimal point (for floats) suffices.

---

Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlplan/aggregator_funcs_test.go, line 267 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTK1_FO_TsxL3LDWeW:-KvTpj903s849X8jrxzv:bykdvko) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/distsqlplan/aggregator_funcs_test.go#L267)):*

Can we use .FamilyEqual instead of comparing strings?
</blockquote></details>

Done. I also realised that this was previously doing comparisons between the protoclass `sql.ColumnType`s which involves other fields. Changing this to comparing between their datums' `Type`s shouldn't be a problem.

---

*[pkg/sql/distsqlplan/aggregator_funcs_test.go, line 268 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTKDR7Vail2HbDHFQQ:-KvTq-ZnaSy0mAfjZgcI:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/distsqlplan/aggregator_funcs_test.go#L268)):*

t.Fatalf here probably
</blockquote></details>

Done.

---

*[pkg/sql/distsqlplan/aggregator_funcs_test.go, line 272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTJbbwaQVO4xtKBgC9:-KvTq0cBVuViiiAtiBDQ:bgdii6g) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/distsqlplan/aggregator_funcs_test.go#L272)):*

use `switch t := rowDist.Datum.(type) {` here, then you can use `t` below instead of re-casting rowDist.Datum
</blockquote></details>

Done (I chose `typedDist` over `t` since `t` conflicts with `t *testing.T`).

---

*[pkg/sql/parser/decimal.go, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-KvTIs5kAuGpHrM2wqwH:-KvTq9rzI6_sRIFZVgeG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d96a4805df5282e1a376ee46ad591c463b8b2a0d/pkg/sql/parser/decimal.go#L44)):*

Since this is only used in a test, define it in the test instead of globally.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18942#-:-KvTqAxk2oU7f2QX4pGP:b-tubeik)*

madelynnblue (approved):

PR #18943, opened 2017-10-02
cherrypick 1.1: sql: fix stale txn deadline

andreimatei:
cc @cockroachdb/release 

Touches #18684

Before this patch, we were relying on TableCollection to reset a txn's
deadline when it detected that the transaction had restarted and a new
set of descriptors needed to be acquired. However, there was a bug:
Executor.execSchemaChanges() would reset the TableCollection in a way
that prevented it from resetting the txn deadline in the future (the
idea is that tc.timestamp was being set to nil, which disarmed the txn
deadline reset code).
This meant that the txn continued with the old deadline after restarting
and so, if it ended up being pushed above the original deadline, it
couldn't commit.

This patch moves the deadline reset out of TableCollection, into the
Executor. The TableCollection's role in resetting the deadline was
questionable to begin with (as opposed to the TC's role in _updating_
the timestamp; that one is good). Now the reset is tied to a SQL
transaction restart.

### DIFF
--- a/pkg/ccl/sqlccl/backup.go
+++ b/pkg/ccl/sqlccl/backup.go
@@ -345,7 +345,7 @@ func makeBackupDescriptor(
 		opt := client.TxnExecOptions{AutoRetry: true, AutoCommit: true}
 		err := txn.Exec(ctx, opt, func(ctx context.Context, txn *client.Txn, opt *client.TxnExecOptions) error {
 			var err error
-			txn.SetFixedTimestamp(endTime)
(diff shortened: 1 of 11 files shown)


bdarnell:
:lgtm:

---

Review status: 0 of 11 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18943#-:-KvT3UJHhANKCNHysJPA:bnfp4nl)*

vivekmenezes:
LGTM

PR #18950, opened 2017-10-02
storage: fix inability to update liveness in mixed version clusters

### LINKED ISSUE #18940: stability: cyan unavailability after upgrade
`cyan` was upgraded from 8ec7c2b to 081f0ba on 2017-09-30-02:00:00 (rolling upgrade by continuous deployment).

During the upgrade process, cyan went from healthy to complete range unavailability:
cyan_summary

The following days had multiple failures (most recent were OOM, have not looked at older ones yet), multiple other upgrade cycles by continuous deployment, but health remains the same:
cyan_symmary_4d

Grabbing logs now before they get GCed. Will investigate further.
Pausing Continuous deployment.

petermattis:
The encoding of `storage.Liveness` changed in #18910 which caused node
liveness updates to fail in mixed version clusters due to failed
conditional puts. Because the encoding changed, continued attempts to
update the liveness would never succeed. Changed `Liveness.expiration`
to use the `LegacyTimestamp` type.

Audited the code and the only protos that are used as the expected value
in conditional put operations are `storage.Liveness` and
`roachpb.RangeDescriptor`. Extended TestBelowRaftProtos to verify the
encoding of these protos does not change unexpectedly.

Consistency checks were failing in mixed version clusters because of an
explicit call to marshal an `hlc.Timestamp` in `Replica.sha512`.

Fixes #18940


mberhault:
LGTM. I'll roll it out to `cyan` once merged.

tschottdorf:
:lgtm:

Looks like #18848 would've had trouble finding this, since for each field member we decide once about whether to insert zeros, and then stick with it (otherwise you potentially fire the state assertions)

---

Reviewed 44 of 44 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18950#-:-KvTTkK-sqUj2YEUzuAy:buh9e65)*

a-robinson:
But our acceptance tests had no trouble finding this (or the previous proto nullability issue) - https://github.com/cockroachdb/cockroach/issues/18817

petermattis:
@tschottdorf Also, the problem here wasn't related to change a non-nullable field to nullable, so #18848 would definitely not have found anything.

bdarnell:
This works for now, but it feels like the wrong solution: Instead of decoding a proto and re-encoding it to use as the target of a conditional put, we should be using the original raw bytes as the expected value for cput. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18950#-:-KvY06rgDMz35T230I9B:b-5imiuo)*

petermattis:
> This works for now, but it feels like the wrong solution: Instead of decoding a proto and re-encoding it to use as the target of a conditional put, we should be using the original raw bytes as the expected value for cput.

That's a good point, though I can't even follow in the code where the old liveness is getting decoded. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18950#-:-KvYO4V1T2VgjBmNeAfJ:b-gr5udc)*

tschottdorf:
Pinging https://github.com/cockroachdb/cockroach/issues/18219 since the code changes may overlap.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18950#-:-KvYOuboAi5CoiH_oDoR:b-4oa5dm)*

PR #18951, opened 2017-10-02
build/deploy: install root CAs in docker image

madelynnblue:
Fixes #18905

PR #18959, opened 2017-10-02
cherrypick-1.1: build/deploy: install root CAs in docker image

### LINKED ISSUE #18905: build/deploy: missing ca certificates
Attempting to run a BACKUP from the deploy docker image:

```
x509: failed to load system roots and no roots provided
```

I think deploy/Dockerfile just needs `RUN apt-get update && apt-get -y upgrade && apt-get install -y ca-certificates`

But I can't test it because of #18904

madelynnblue:
Fixes #18905 
Cherrypick of #18951

### DIFF
--- a/build/deploy/Dockerfile
+++ b/build/deploy/Dockerfile
@@ -2,6 +2,13 @@ FROM debian:8.9
 
 MAINTAINER Tobias Schottdorf <tobias.schottdorf@gmail.com>
 
+# Install root CAs so we can make SSL connections to phone home and
+# do backups to GCE/AWS/Azure.
+RUN apt-get update && \
+	apt-get -y upgrade && \
+	apt-get install -y ca-certificates  && \
+	rm -rf /var/lib/apt/lists/*
+
 RUN mkdir -p /cockroach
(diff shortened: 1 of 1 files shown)


a-robinson (approved):

bdarnell:
Since docker caches/reuses intermediate layers, how do we ensure that each release has the latest CA certs at the time it is built?

benesch:
StackOverflow suggests something like this: https://github.com/moby/moby/issues/1996#issuecomment-185872769

a-robinson:
That'd be nice, or just using `docker build --no-cache`. Although is this a problem in practice given that our releases are built on ephemeral teamcity workers?

bdarnell:
If the teamcity workers are sufficiently isolated, we may be fine here, but we should probably add `--no-cache` to be safe on the `docker build` commands for releases.

benesch:
Just for posterity: TeamCity workers are intentionally *not* isolated so that e.g. a build will not need to redownload the builder image if it's running on an agent that's previously run a build with the same builder image version. Less of a concern with releases, since we release once a week and agents rarely survive longer than a day, but still not a guarantee.

PR #18966, opened 2017-10-03
docs: use clock offset consistently

yaojingguo:
`clock skew` and `clock offset` mean the same thing in the design doc. And as mentioned in [What is the precise definition of clock skew in the design doc?](https://forum.cockroachlabs.com/t/what-is-the-precise-definition-of-clock-skew-in-the-design-doc/980),  `clock offset` is more precise. So it is better to use clock offset consistently.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -357,7 +357,7 @@ There are several scenarios in which transactions interact:
   In that case, we would need to take this value into account, but
   we just don't know. Hence the transaction restarts, using instead
   a future timestamp (but remembering a maximum timestamp used to
-  limit the uncertainty window to the maximum clock skew). In fact,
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -469,7 +469,7 @@ Please see [pkg/roachpb/data.proto](https://github.com/cockroachdb/cockroach/blo
 
 **Choosing a Timestamp**
 
-A key challenge of reading data in a distributed system with clock skew
+A key challenge of reading data in a distributed system with clock offset
```
a-robinson: I think this line was ok with the word "skew", since it's the fact that the machines have clock skew that causes an offset between their clocks.

a-robinson (commented):
Thanks @yaojingguo!

PR #18975, opened 2017-10-03
cherrypick-1.1: distsql: pre-reserve memory needed to mark rows in HashJoiner build phase

asubiotto:
A situation was uncovered by #18600, where the HashJoiner would run out
of memory in the probe phase. This was because we had made an assumption
that we wouldn't hit a memory limit if the buffer phase filled up at
most 2/3 of the limit with both streams , since the marking
infrastructure would take up only a fraction of 1/3 (the chosen stream).

This assumption failed to take into account other limits shared with
other processors. This change pre-reserves the memory needed for the
probe phase in the build phase so that we can keep a single point in the
code where we fall back to disk while not relying on any limit
assumptions.


andreimatei:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18975#-:-KvXn6i-lqSgCeD45JPr:bnfp4nl)*

PR #18978, opened 2017-10-03
distsql: truncate for decimal comparisons and reduce precision for float comparisons in TestDistAggregationTable

### LINKED ISSUE #18968: sql/distsqlplan: TestDistAggregationTable/VARIANCE/dec1/50 failed under stress
SHA: https://github.com/cockroachdb/cockroach/commits/b36405997c61ed77cf758863c55a73c3db833461

Parameters:
```
TAGS=
GOFLAGS=
```

Stress build found a failed test: https://teamcity.cockroachdb.com/viewLog.html?buildId=367016&tab=buildLog

```

```

richardwu:
🤞 fixes #18968, #18971 

PR #18942 reduced precision for decimal comparisons between
local and non-local aggregations in TestDistAggregationTable as well as
introduced floating point aggregations.

Precision was reduced from 20 to 19 with a 19 precision apd.Context.
This however did not amend the case where we have boundary digits (i.e.
4 and 5) in the 20th decimal place which may cause the 19th precision
digits to also differ. Instead, the context should always truncate the
20th digit which is admissible for decimals.

Nightlies also revealed that a precision of 10 after the decimal for
floats was too precise. This PR reduces it down to 3 which should be
:100: and aligns with how much we round in logic tests.

### DIFF
--- a/pkg/sql/distsqlplan/aggregator_funcs_test.go
+++ b/pkg/sql/distsqlplan/aggregator_funcs_test.go
@@ -20,6 +20,7 @@ import (
 
 	"golang.org/x/net/context"
 
+	"github.com/cockroachdb/apd"
 	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/sql/distsqlrun"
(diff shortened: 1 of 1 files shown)


madelynnblue (approved):

pkg/sql/distsqlplan/aggregator_funcs_test.go
```diff
@@ -33,15 +34,25 @@ import (
...
 	// unavoidable off-by-last-digit errors, which is semantically okay.
-	compareCtx = parser.DecimalCtx.WithPrecision(parser.DecimalCtx.Precision - 1)
-	// floatPrecFmt is the format string with a precision of 10 (after
+	compareCtx = &apd.Context{
+		Precision: parser.DecimalCtx.Precision - 1,
+		// We truncate down instead of rounding since the last digit
+		// may cause the new last digit to also be off by one.
+		// For example, if we use the default RoundHalfUp and the last
+		// digits of the decimals we want to compare are 4 and 5,
+		// then the former will be rounded down and the latter will
```
RaduBerinde: Won't this still fail for 0.19 vs 0.20? (both could be valid if the real result is close to 0.195). Seems like the right thing to do is not to truncate but be less strict when comparing.

mjibson: Oh yes, you're correct. I thought it would be ok because of how we handle 

https://github.com/cockroachdb/cockroach/blob/aad470f0cedacd5bff4b7a09a479a2b9c4ab8466/pkg/sql/logictest/testdata/logic_test/aggregate#L809

But that only works because we know the expected output and so can be more confident it won't have the rollover you describe. For reference, here's how apd tests do their 1ulp comparison:

https://github.com/cockroachdb/apd/blob/b1ce49cb2a474f4416531e7395373eaafaa4fbe2/gda_test.go#L672

richardwu: Good point, I'll push out another PR again to remedy this.

PR #18979, opened 2017-10-03
build: also build terminfo libs as deps

knz:
Needed for #18531.

I will update the PR with the docker.io image tag when I have checked locally that #18531 passes all tests with this. You can start looking at the general gist of the change in the mean time.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial-20170619
+FROM ubuntu:xenial-20170915
 
 MAINTAINER Tamir Duberstein <tamird@gmail.com>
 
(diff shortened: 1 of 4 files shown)


build/Dockerfile
```diff
@@ -87,6 +87,43 @@ RUN mkdir build src && cd build \
...
+
+# Run the two builds.
+# We need to manually disable secure_getenv and cdefs on the musl build
+# because they are not available in the musl sysroot.
+RUN mkdir ncurses \
+ && curl -fsSL http://ftp.gnu.org/gnu/ncurses/ncurses-6.0.tar.gz | tar --strip-components=1 -C ncurses -xz \
+ && cd ncurses \
+ && patch -p0 <../ncurses.patch \
+ && export cf_cv_type_of_bool='unsigned char' \
+ && export cf_cv_working_poll=yes \
```
benesch: This looks suspicious. It deserves a comment at least, I think?

benesch (approved):
Approach LGTM, but would be good to get @bdarnell's buy-in too.

build/Dockerfile
knz: Done.

knz:
I checked that the binaries generated using this builder image properly work on CentOS (in particular the symbol versioning story is solid, thanks to the `--without-sym-versioning` flag to the embedded terminfo's `configure`). Now pushing to docker.io.

knz:
Ok the image tag is populated. PTAL

knz:
cc @dt (I think you were also working on extending the builder image. Perhaps you want to use this one as base.)

dt:
ugh, I already built and pushed `20171003-192752`

benesch:
LGTM modulo rebuilding on top of @dt's change!

bdarnell:
I'm still not fond of dynamically linking anything but glibc, but I guess if it works on centos 6 and current ubuntu, LGTM

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[build/Dockerfile, line 93 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18979#-Kv_QOy345w9bR7gAbu-:-Kv_QOy4BUxhF8499E84:byu4nt0) ([raw file](https://github.com/cockroachdb/cockroach/blob/9efa7e01678269302d91ad14c8c5d82f7b98750a/build/Dockerfile#L93)):*
> ```Dockerfile
> # (on BSD or BSD-derived like macOS it's already built-in; on windows we don't need it.)
> #
> # The patch is needed to work around a bug in Debian mawk, see
> ```

According to https://askubuntu.com/questions/561621/choosing-awk-version-on-ubuntu-14-04 if the `gawk` package is installed it will be used in place of `mawk` as `/usr/bin/awk` (weird, but OK). Installing `gawk` above should be an alternative to this patch, although I don't know if it's worth switching. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18979#-:-Kv_Wo2qX2lPjwMknPvN:b-6ib0lx)*

knz:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[build/Dockerfile, line 93 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/18979#-Kv_QOy345w9bR7gAbu-:-KvaCBJZwzJcB6ItMvcf:b-v85tk6) ([raw file](https://github.com/cockroachdb/cockroach/blob/9efa7e01678269302d91ad14c8c5d82f7b98750a/build/Dockerfile#L93)):*

According to https://askubuntu.com/questions/561621/choosing-awk-version-on-ubuntu-14-04 if the `gawk` package is installed it will be used in place of `mawk` as `/usr/bin/awk` (weird, but OK). Installing `gawk` above should be an alternative to this patch, although I don't know if it's worth switching. 
</blockquote></details>

`gawk` is already installed and `/usr/bin/awk` points to it. However the ncurses Makefile forces `mawk` so that doesn't help :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18979)*

PR #18980, opened 2017-10-03
Cherrypick 1.1:  server, ui: add quiescent to the debug page

### LINKED ISSUE #18450: ui: problemranges page should show whether replica is quiesced

BramGruneir:
This change allows the range debug page to display if a range is quiescent.
There are also a few small changes to the range debug page:
- reordered the top options slightly, so the lease type is above the lease status
- removed the capitalization of Epoch and Expiration as it doesn't fit with the rest of the page
- when a replica is dormant, it's raft state now shows dashes instead of zeros for values it doesn't have

Closes #18450.

This is a Cherry Pick of #18900, see that issue for more details.
cc @cockroachdb/release


bdarnell (approved):

BramGruneir:
So should I merge this, or wait until 1.1.1?

bdarnell:
Wait for 1.1.1.

PR #18981, opened 2017-10-03
cmd/docgen: add grammar diagram generation to gen docs 

dt:
This pulls the diagram generator in from the docs repo, and includes it in the
generation of `docs/generated`, as well as its output, meaning that changes to
the grammar will also require updating the generated diagrams.

Our CI workers download and decrypt Railroad.jar to avoid excessive requests to
the API.


mjibson:
The yacc parser is already present at https://github.com/cockroachdb/cockroach/tree/master/pkg/internal/rsg/yacc. Is it possible to reuse that code so there's only one copy in the repo?

dt:
@mjibson aha! perfect. done.

benesch:
:lgtm: provided @bdarnell and @mjibson are comfortable with the encryption scheme. Also, you might want to consider gitattributing `-merge` on the generated docs.

Thanks, @dt!

---

Reviewed 10 of 10 files at r1, 80 of 80 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18981#-:-KvYCdpcLATlZTxxO7LW:b-jms0ix)*

dt:
Short of building and production-izing our own diagram API server, i'm not sure there we can do much better as long as we're re-generating in PRs (since we're running, essentially, arbitrary code).

mjibson:
Sounds like we need to figure out how to rebuild the agent image then. I'd just dump the railroad.jar file there and drop the encryption hoo-ha.

jordanlewis:
The postgres-test image has Java. That's how the Java acceptance test works.

dt:
well, we run builds (and generation) in the builder, not directly on the agent image (or in the acceptance test image), and fortunately rebuilding the builder is pretty well documented, is a bit slow. doing so now.

knz:
Just be sure not to embed the railroad.jar inside the builder image. We don't have permission to redistribute.

dt:
@knz yep, we're not doing that, but thanks for double checking.

knz:
Could you perhaps extract the dockerfile change into a separate PR? This one you could merge straight ahead, so I can move forward with #18979.

knz:
you can regen this PR and make it merge, or pull out the now-first commit into a separate PR and merge that?

bdarnell:
So now teamcity will enforce that diagrams have been generated, but how are developers supposed to generate those diagrams? Does everyone who changes sql.y need to manually get our internal copy of railroad.jar? What about external contributors? Is there an easy way to get the generated diagrams out of the failed teamcity build? 

If the encryption key is supplied to the CI builders, isn't it trivial for a rogue PR to extract Railroad.jar? Hardening our CI environment to the point where we can safely use non-redistributable components in our build process seems challenging. 

---

Reviewed 2 of 92 files at r3, 5 of 10 files at r5.
Review status: 7 of 92 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18981#-:-KvbtzM7oaiARl0sCJ7Q:b7mt577)*

knz:
Good call on the fact our env vars are exposed to rogue PRs. This is already a problem I think.

There's an option to use an online web service instead of railroad.jar. It's much slower but it gets things done.

dt:
@bdarnell yep, updating sql.y will involve re-generating the diagrams, which was the goal, since that ensures that the updates actually happen, and that any breaking changes get caught and fixed, rather than waiting until someone tries to update the docs days or weeks later. If you don't have railroad.jar, we generate by hitting the public web API (same as today in docs repo). 

@knz: yeah, we've discussed ability to exfiltrate the env var, or an auth'ed s3 bucket, or a non-public build script that runs before the public ones, and various other schemes at _great length_ in teamcity slack.  At the end of the day, we're running arbitrary code in PR build, so as mentioned above, short of building a new diagram API server and productionizing it, we're not going to find an perfect solution.

dt:
@knz we ran the proposal by the author and he's OK with it.

dt:
@benesch I could go either way on `-merge`. Personally I used to always say that generated content should always be `-merge`, but consensus elsewhere here (like protos) has been that we're not making that a rule, and in most cases have said that if git manages to auto-merge, we're OK. We've also lived without it so far in docs repo so it seems to be okay?

benesch (approved):
SGTM

PR #18986, opened 2017-10-03
*: consistently use versioned URLs to docs

### LINKED ISSUE #18984: Fix version number in links to docs site
Links to the docs site generated from the CLI can contain links to "relative versions," e.g., stable. This could lead to problems because an old binary will always link to the newest version's content and if the process changes, the instructions we link to will not provide the correct responses to the user.

### Steps to Repro
~~~
./cockroach start --insecure
~~~

Generates the following response:
~~~
*
* WARNING: RUNNING IN INSECURE MODE!
* 
* - Your cluster is open for any client that can access <all your IP addresses>.
* - Any user, even root, can log in without providing a password.
* - Any user, connecting as root, can read or write any data in your cluster.
* - There is no network encryption nor authentication, and thus no confidentiality.
* 
* Check out how to secure your cluster: https://www.cockroachlabs.com/docs/stable/secure-a-cluster.html
*
~~~

### Expected Behavior

Link to docs should be:

~~~
https://www.cockroachlabs.com/docs/$(binary major version)/secure-a-cluster.html
~~~

e.g.

~~~
https://www.cockroachlabs.com/docs/v1.0/secure-a-cluster.html
~~~

### Actual Behavior

Link uses `/stable`, which is only as valid as long as this version has `stable` pointed to it.

benesch:
Previously, some links to the docs website hardcoded the "stable"
version alias. Since the "stable" content is updated with every release,
the link points to increasingly irrelevant content for old binaries.
Instead, derive a docs "permalink" from the binary's version. (This
logic previously existed in sql/parser; this commit simply applies that
logic to docs URLs in other packages.)

Fix #18984.

@knz can you review? (The reviewers field is broken.)

### DIFF
--- a/pkg/base/docs.go
+++ b/pkg/base/docs.go
@@ -1,0 +1,29 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package base
+
+import (
+	"fmt"
+
+	"github.com/cockroachdb/cockroach/pkg/build"
+)
+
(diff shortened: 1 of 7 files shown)


knz:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18986#-:-KvYDKGNTKWLR8mdJAte:bnfp4nl)*

mrtracy:
LGTM

benesch:
TFTRs!

PR #18987, opened 2017-10-03
sqlccl: remove WriteBatch import option

dt:


benesch (approved):
Obviously you'll want @danhhz's review, but LGTM for what it's worth.

mjibson:
Can we also remove it from roachpb?

benesch:
I was assuming we'd need to wait for another release cycle to pass before removing the command implementation, but I guess it's reasonable that a mixed 1.1 and 1.2 cluster won't be able to complete a restore until all nodes upgraded to 1.2? (IIRC it's possible for a 1.1 restore to issue WriteBatch commands if the cluster setting has been toggled appropriately.)

dt:
@mjibson I did _not_ remove the command itself or the tests of it in this commit, just the usage of it in import (e.g. RESTORE). Removing the command itself can happen, if we choose to do so, separately.

danhhz:
:lgtm: yeah, the WriteBatch implementation needs to stay at least through 1.2

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18987#-:-KvYgUDU-9TcR6LbV3Xq:bi3d1jd)*

PR #18991, opened 2017-10-03
docs: name the tech notes after date of first commit

knz:
Mirrors the RFC naming convention.


bdarnell:
RFCs are dated because they are in large part historical artifacts; they capture discussion at a particular time and we do not generally go back and make substantial changes to them as the product evolves. Tech notes, on the other hand, tend to be more "evergreen" and should be updated to match reality. Given that, I'm not sure that a date-based naming convention makes sense here.

knz:
Yes I agree that tech notes will more naturally evolve over time.

Except that this evolution may or may not happen "on time". At any point of time, some tech notes might still be outdated. I guess I would be happy if we had a comment to "sort the tech notes by date of last commit". Is there such a thing? I could put this command in the README instead of renaming the files.

bdarnell:
`ls -lt` will be close enough in most cases if you've been maintaining a local checkout (there's probably a git command that may be more precise). 

I'm not sure how useful this would be, though. While the directory is small it doesn't really matter. When it gets bigger, I doubt that "sort by time of last update" is going to be the best way to make sense of its contents.

knz:
Thanks for the feedback! Closing this.

PR #18996, opened 2017-10-03
build: Build Docker images with the --no-cache flag

a-robinson:
To avoid ending up with old image components like CA certs.

Follow-up to a comment on #18959


benesch (approved):
Ooh, even better. LGTM!

bdarnell:
Should `build/deploy/rhel/build.sh` also be updated (and does the dockerfile need updating to include the RHEL equivalent of the ca-certificates package)?

a-robinson:
`--no-cache` wouldn't hurt. The base image (registry.access.redhat.com/rhel-atomic) comes with the most recent `ca-certificates` package and using `--no-cache` will ensure we uses the latest such registry.access.redhat.com/rhel-atomic image, so there's no need to do anything with it (plus the atomic image actually comes without `yum`).

PR #18997, opened 2017-10-03
engine: Make temp directory cleanup resilient to rename errors

### LINKED ISSUE #18994: storage: cleaning up old temporary directory doesn't work on OverlayFS
The logic in `engine.cleanupTempStorageDirs` tries to avoid blocking during the startup path by renaming the old temporary storage directory from its normal name to a `TO-DELETE-` prefixed name so that a new temporary directory can be created in its place. The `TO-DELETE-` directory then gets deleted in the background to avoid blocking anything.

Unfortunately, if the old temporary storage directory was created in a lower OverlayFS layer, the rename operation will always fail (https://github.com/moby/moby/issues/25409). This breaks use cases such as those that initialize a database in a certain way as part of a Docker build process so that the resulting image can be used in reproducible tests.

It wouldn't be a huge deal if overlay2 wasn't the default storage driver for Docker on Mac (and possibly other environments like Windows, but I didn't check).

cc @arjunravinarayan - I'm going to just fall back to removing the file if `os.Rename` fails. Is there anything wrong with that approach?

a-robinson:
Fixes #18994

### DIFF
--- a/pkg/storage/engine/temp_engine.go
+++ b/pkg/storage/engine/temp_engine.go
@@ -99,7 +99,11 @@ func cleanupTempStorageDirs(ctx context.Context, path string, wg *sync.WaitGroup
 		toDeleteFull := filepath.Join(path, fileToDelete.Name())
 		if toDeleteFull != deletionDir {
 			if err := os.Rename(toDeleteFull, filepath.Join(deletionDir, fileToDelete.Name())); err != nil {
-				return err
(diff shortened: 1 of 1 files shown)


arjunravinarayan:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/18997#-:-KvYntbOtizbsQqCrqDe:bnfp4nl)*

PR #19000, opened 2017-10-03
cherrypick-1.1: engine: Make temp directory cleanup resilient to rename errors

### LINKED ISSUE #18994: storage: cleaning up old temporary directory doesn't work on OverlayFS

a-robinson:
Fixes #18994 

@cockroachdb/release - very low risk since it only runs in a code path when we'd previously just `log.Fatal`, and fixes a bug that's blocking a user's efforts to run tests against us, but I'll leave the decision to @bdarnell

### DIFF
--- a/pkg/storage/engine/temp_engine.go
+++ b/pkg/storage/engine/temp_engine.go
@@ -99,7 +99,11 @@ func cleanupTempStorageDirs(ctx context.Context, path string, wg *sync.WaitGroup
 		toDeleteFull := filepath.Join(path, fileToDelete.Name())
 		if toDeleteFull != deletionDir {
 			if err := os.Rename(toDeleteFull, filepath.Join(deletionDir, fileToDelete.Name())); err != nil {
-				return err
+				// Fall back to just removing the file if rename fails (e.g. due to #18994).
+				if rmErr := os.Remove(toDeleteFull); rmErr != nil {
+					return errors.Errorf("failed to remove file %q (%s) after failing to rename it (%s)",
+						toDeleteFull, rmErr, err)
+				}
 			}
 		}
 	}


bdarnell (approved):
I think we should hold this until 1.1.1.

a-robinson (commented):
👍

a-robinson:
On second thought, I think this does make sense to include in 1.1.0 unless that door is really shut quite hard. It's a system crash that's directly affecting a customer that can't be worked around without asking everyone involved to change their docker storage driver. Given the minimal risk involved, I'd lean toward not making them wait an extra week or two for a fix.

bdarnell:
I guess since we already bumped the rc1 commit we can throw this in too, although let me say that what this customer is doing seems a bit nuts to me and I'm not sure we want to support it (it results in a lot of clusters out there with the same ID).

a-robinson:
Is it actually nuts, though? Wanting to be able to run tests against a database that always starts from the exact same state doesn't seem that crazy to me. Consistency of environment across different machines and instantiations is one of docker's big selling points.

bdarnell:
My concern is that we don't generally support doing things by copying rocksdb files around. If you attempt to back up your database by making filesystem-level backups of the rocksdb state, you'll break things in subtle ways. A cleaner (but admittedly slower) way to do this would be to make a backup of a database in the desired state and then restore it on each use. 

(and if someone tries to extrapolate from this and start up a multi-node cluster by baking things into a docker image, it won't work)

richardwu:
Just a heads up, the temp directory cleanup will be cleaned up once [this PR](https://github.com/cockroachdb/cockroach/pull/18544) is merged and comments addressed (which is targeted for 1.1.x).

PR #19001, opened 2017-10-03
distsql: add `distsql-disk` configuration to logic tests (incl. bigtest)

asubiotto:
Previously we only used the `distsql-disk` configuration in `postgresjoin`.

### DIFF
--- a/build/teamcity-sqllogictest.sh
+++ b/build/teamcity-sqllogictest.sh
@@ -9,7 +9,7 @@ mkdir -p artifacts
 # the all-on/all-off strategy BULIDER_HIDE_GOPATH_SRC gives us.
 export BUILDER_HIDE_GOPATH_SRC=0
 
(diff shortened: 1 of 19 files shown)


RaduBerinde:
For many of the files, it doesn't make much sense to run in this configuration (e.g. event_log, dependencies, inet). It would be useful to trim them down to avoid making logic tests take 30% longer.

asubiotto:
Done, PTAL. I left a couple of files (`collatedstring_*`, `datetime`) where I think it's interesting to test the ORDER BY statements.

RaduBerinde:
What about the `distsql_*` tests? Should we have a `5node-disk` config as well?

asubiotto:
Added a `5node-distsql-disk` config

RaduBerinde:
LGTM 💯

PR #19005, opened 2017-10-03
sql: Add background renewal for table descriptor leases 

### LINKED ISSUE #17227: sql: refresh table leases asynchronously

lgo:
This commit changes TableDescriptor lease acquisition to acquire a new
lease before the current lease expires. This prevents routines from
fully blocking when there are no valid leases.

If renewal continously fails, the error will become user-facing if no
lease is acquired yet by the time a lease is expired.

This patch also adds a benchmark for `LeaseManager.AcquireByName`
when cached. LeaseManager.AcquireByName is a critical path for
statements and generally when accessing the table descriptor. The
case when the lease is cached serves the majority of operations.

Closes #17227.

cc: @vivekmenezes @andreimatei

### DIFF
--- a/pkg/base/config.go
+++ b/pkg/base/config.go
@@ -85,6 +85,10 @@ const (
 	// that we use to randomly jitter the lease duration when acquiring a
 	// new lease and the lease renewal timeout.
 	DefaultTableDescriptorLeaseJitterFraction = 0.25
+
+	// DefaultTableDescriptorLeaseRenewalTimeout is the default time
+	// before a lease expires when acquisition to renew the lease begins.
+	DefaultTableDescriptorLeaseRenewalTimeout = time.Minute
 )
 
 var defaultRaftElectionTimeoutTicks = envutil.EnvOrDefaultInt(
@@ -489,6 +493,10 @@ type LeaseManagerConfig struct {
 	// randomly jitter the lease duration when acquiring a new lease and
 	// the lease renewal timeout.
 	TableDescriptorLeaseJitterFraction float64
+
+	// DefaultTableDescriptorLeaseRenewalTimeout is the default time
+	// before a lease expires when acquisition to renew the lease begins.
+	TableDescriptorLeaseRenewalTimeout time.Duration
 }
 
 // NewLeaseManagerConfig initializes a LeaseManagerConfig with default values.
@@ -496,5 +504,6 @@ func NewLeaseManagerConfig() *LeaseManagerConfig {
 	return &LeaseManagerConfig{
 		TableDescriptorLeaseDuration:       DefaultTableDescriptorLeaseDuration,
 		TableDescriptorLeaseJitterFraction: DefaultTableDescriptorLeaseJitterFraction,
+		TableDescriptorLeaseRenewalTimeout: DefaultTableDescriptorLeaseRenewalTimeout,
 	}
 }
--- a/pkg/sql/lease.go
+++ b/pkg/sql/lease.go
@@ -40,10 +40,8 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil/singleflight"
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
+	"github.com/cockroachdb/cockroach/pkg/util/tracing"
 )
-
-// TODO(pmattis): Periodically renew leases for tables that were used recently and
-// for which the lease will expire soon.
 
 // tableVersionState holds the state for a table version. This includes
 // the lease information for a table version.
@@ -122,6 +120,9 @@ type LeaseStore struct {
 	// range of the actual lease duration will be
 	// [(1-leaseJitterFraction) * leaseDuration, (1+leaseJitterFraction) * leaseDuration]
 	leaseJitterFraction float64
+	// leaseRenewalTimeout is the time before a lease expires when
+	// acquisition to renew the lease begins.
+	leaseRenewalTimeout time.Duration
 
 	testingKnobs LeaseStoreTestingKnobs
 	memMetrics   *MemoryMetrics
@@ -564,6 +565,11 @@ type tableState struct {
 	tableNameCache *tableNameCache
 	stopper        *stop.Stopper
 
+	// renewalInProgress is an atomic indicator for when a renewal for a
+	// lease has begun. This is atomic to prevent multiple routines from
+	// entering renewal initialization.
+	renewalInProgress int32
+
 	mu struct {
 		syncutil.Mutex
 
@@ -614,6 +620,7 @@ func (t *tableState) acquire(
 			return nil, result.Err
 		}
 	}
+
 	return t.findForTimestamp(ctx, timestamp, m)
 }
 
@@ -970,6 +977,33 @@ func (t *tableState) purgeOldVersions(
 	return err
 }
 
+// startLeaseRenewal starts a singleflight.Group to acquire a lease.
+// This function blocks until lease acquisition completes.
+// t.renewalInProgress must be set to 1 before calling.
+func (t *tableState) startLeaseRenewal(
+	ctx context.Context, m *LeaseManager, tableVersion *tableVersionState,
+) {
+	resultChan, _ := t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
+		log.VEventf(ctx, 1,
+			"background lease renewal beginning for tableID=%d tableName=%q",
+			t.id, tableVersion.TableDescriptor.Name)
+		token, err := t.acquireNodeLease(ctx, m, hlc.Timestamp{})
+		if err != nil {
+			log.Errorf(ctx,
+				"background lease renewal for tableID=%d tableName=%q failed: %s",
+				t.id, tableVersion.TableDescriptor.Name, err)
+		} else {
+			log.VEventf(ctx, 1,
+				"background lease renewal finished for tableID=%d tableName=%q",
+				t.id, tableVersion.TableDescriptor.Name)
+		}
+		return token, err
+
+	})
+	<-resultChan
+	atomic.StoreInt32(&t.renewalInProgress, 0)
+}
+
 // LeaseAcquireBlockType is the type of blocking result event when
 // calling LeaseAcquireResultBlockEvent.
 type LeaseAcquireBlockType int
@@ -1144,6 +1178,7 @@ type LeaseManager struct {
 	// Not protected by mu.
 	tableNames   tableNameCache
 	testingKnobs LeaseManagerTestingKnobs
+	ambientCtx   log.AmbientContext
 	stopper      *stop.Stopper
 }
 
@@ -1151,6 +1186,7 @@ type LeaseManager struct {
 //
 // stopper is used to run async tasks. Can be nil in tests.
 func NewLeaseManager(
+	ambientCtx log.AmbientContext,
 	nodeID *base.NodeIDContainer,
 	db client.DB,
 	clock *hlc.Clock,
@@ -1166,6 +1202,7 @@ func NewLeaseManager(
 			nodeID:              nodeID,
 			leaseDuration:       cfg.TableDescriptorLeaseDuration,
 			leaseJitterFraction: cfg.TableDescriptorLeaseJitterFraction,
+			leaseRenewalTimeout: cfg.TableDescriptorLeaseRenewalTimeout,
 			testingKnobs:        testingKnobs.LeaseStoreTestingKnobs,
 			memMetrics:          memMetrics,
 		},
@@ -1173,7 +1210,8 @@ func NewLeaseManager(
 		tableNames: tableNameCache{
 			tables: make(map[tableNameCacheKey]*tableVersionState),
 		},
-		stopper: stopper,
+		ambientCtx: ambientCtx,
+		stopper:    stopper,
 	}
 
 	lm.mu.Lock()
@@ -1192,7 +1230,9 @@ func nameMatchesTable(table *sqlbase.TableDescriptor, dbID sqlbase.ID, tableName
 // the timestamp. It returns the table descriptor and a expiration time.
 // A transaction using this descriptor must ensure that its
 // commit-timestamp < expiration-time. Care must be taken to not modify
-// the returned descriptor.
+// the returned descriptor. Renewal of a lease may begin in the
+// background. Renewal is done in order to prevent blocking on future
+// acquisitions.
 func (m *LeaseManager) AcquireByName(
 	ctx context.Context, timestamp hlc.Timestamp, dbID sqlbase.ID, tableName string,
 ) (*sqlbase.TableDescriptor, hlc.Timestamp, error) {
@@ -1200,6 +1240,25 @@ func (m *LeaseManager) AcquireByName(
 	tableVersion := m.tableNames.get(dbID, tableName, timestamp)
 	if tableVersion != nil {
 		if !timestamp.Less(tableVersion.ModificationTime) {
+
+			// Atomically check and begin a renewal if one has not already
+			// been set.
+
+			durationUntilExpiry := time.Duration(tableVersion.expiration.WallTime - timestamp.WallTime)
+			if durationUntilExpiry < m.LeaseStore.leaseRenewalTimeout {
+				if t := m.findTableState(tableVersion.ID, false /* create */); t != nil &&
+					atomic.CompareAndSwapInt32(&t.renewalInProgress, 0, 1) {
+					// Start the renewal. When it finishes, it will reset t.renewalInProgress.
+					if err := t.stopper.RunAsyncTask(context.Background(),
+						"lease renewal", func(ctx context.Context) {
+							ctx, _ = tracing.EnsureContext(ctx, m.ambientCtx.Tracer, "lease renewal")
+							t.startLeaseRenewal(ctx, m, tableVersion)
+						}); err != nil {
+						return &tableVersion.TableDescriptor, tableVersion.expiration, err
+					}
+				}
+			}
+
 			return &tableVersion.TableDescriptor, tableVersion.expiration, nil
 		}
 		if err := m.Release(&tableVersion.TableDescriptor); err != nil {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -276,8 +276,16 @@ func NewServer(cfg Config, stopper *stop.Stopper) (*Server, error) {
 	if leaseManagerTestingKnobs := cfg.TestingKnobs.SQLLeaseManager; leaseManagerTestingKnobs != nil {
 		lmKnobs = *leaseManagerTestingKnobs.(*sql.LeaseManagerTestingKnobs)
 	}
-	s.leaseMgr = sql.NewLeaseManager(&s.nodeIDContainer, *s.db, s.clock, lmKnobs,
-		s.stopper, &s.internalMemMetrics, s.cfg.LeaseManagerConfig)
+	s.leaseMgr = sql.NewLeaseManager(
+		s.cfg.AmbientCtx,
+		&s.nodeIDContainer,
+		*s.db,
+		s.clock,
+		lmKnobs,
+		s.stopper,
+		&s.internalMemMetrics,
+		s.cfg.LeaseManagerConfig,
+	)
 	s.leaseMgr.RefreshLeases(s.stopper, s.db, s.gossip)
 
 	// We do not set memory monitors or a noteworthy limit because the children of
--- a/pkg/sql/lease_test.go
+++ b/pkg/sql/lease_test.go
@@ -44,7 +44,7 @@ import (
 )
 
 type leaseTest struct {
-	*testing.T
+	testing.TB
 	server                   serverutils.TestServerInterface
 	db                       *gosql.DB
 	kvDB                     *client.DB
@@ -53,13 +53,13 @@ type leaseTest struct {
 	cfg                      *base.LeaseManagerConfig
 }
 
-func newLeaseTest(t *testing.T, params base.TestServerArgs) *leaseTest {
+func newLeaseTest(tb testing.TB, params base.TestServerArgs) *leaseTest {
 	if params.LeaseManagerConfig == nil {
 		params.LeaseManagerConfig = base.NewLeaseManagerConfig()
 	}
-	s, db, kvDB := serverutils.StartServer(t, params)
+	s, db, kvDB := serverutils.StartServer(tb, params)
 	leaseTest := &leaseTest{
-		T:      t,
+		TB:     tb,
 		server: s,
 		db:     db,
 		kvDB:   kvDB,
@@ -893,4 +893,46 @@ INSERT INTO t.timestamp VALUES ('a', 'b');
 	if err := <-errChan; err != nil {
 		t.Fatal(err)
 	}
+}
+
+// BenchmarkLeaseAcquireByNameCached benchmarks the AcquireByName
+// acquisition code path if a valid lease exists and is contained in
+// tableNameCache. In particular this benchmark is done with
+// parallelism, which is important to also benchmark locking.
+func BenchmarkLeaseAcquireByNameCached(b *testing.B) {
+	defer leaktest.AfterTest(b)()
+	params, _ := createTestServerParams()
+
+	t := newLeaseTest(b, params)
+	defer t.cleanup()
+
+	if _, err := t.db.Exec(`
+CREATE DATABASE t;
+CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
+`); err != nil {
+		t.Fatal(err)
+	}
+
+	tableDesc := sqlbase.GetTableDescriptor(t.kvDB, "t", "test")
+	dbID := tableDesc.ParentID
+	tableName := tableDesc.Name
+	leaseManager := t.node(1)
+
+	// Acquire the lease so it is put into the tableNameCache.
+	_, _, err := leaseManager.AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, tableName)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	b.ResetTimer()
+
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			_, _, err := leaseManager.AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, tableName)
+			if err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
+
 }
--- a/pkg/sql/lease_test.go
+++ b/pkg/sql/lease_test.go
@@ -39,7 +39,9 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
+	"github.com/cockroachdb/cockroach/pkg/util/tracing"
 	"github.com/pkg/errors"
 )
 
@@ -185,12 +187,16 @@ func (t *leaseTest) mustPublish(ctx context.Context, nodeID uint32, descID sqlba
 	}
 }
 
+// node gets a LeaseManager corresponding to a mock node. A new lease
+// manager is initialized for each node. This allows for more complex
+// inter-node lease testing.
 func (t *leaseTest) node(nodeID uint32) *sql.LeaseManager {
 	mgr := t.nodes[nodeID]
 	if mgr == nil {
 		nc := &base.NodeIDContainer{}
 		nc.Set(context.TODO(), roachpb.NodeID(nodeID))
 		mgr = sql.NewLeaseManager(
+			log.AmbientContext{Tracer: tracing.NewTracer()},
 			nc, *t.kvDB,
 			t.server.Clock(),
 			t.leaseManagerTestingKnobs,
@@ -928,11 +934,109 @@ CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
 
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
-			_, _, err := leaseManager.AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, tableName)
+			_, _, err := leaseManager.AcquireByName(
+				context.TODO(),
+				t.server.Clock().Now(),
+				dbID,
+				tableName,
+			)
 			if err != nil {
 				b.Fatal(err)
 			}
 		}
 	})
 
+}
+
+// This test makes sure leases get renewed automatically in the
+// background if the lease is about to expire, without blocking. We
+// first acquire a lease, then continue to re-acquire it until another
+// lease is renewed.
+func TestLeaseRenewedAutomatically(testingT *testing.T) {
+	defer leaktest.AfterTest(testingT)()
+
+	var testAcquiredCount int32
+	var testAcquisitionBlockCount int32
+
+	params, _ := createTestServerParams()
+	params.Knobs = base.TestingKnobs{
+		SQLLeaseManager: &sql.LeaseManagerTestingKnobs{
+			LeaseStoreTestingKnobs: sql.LeaseStoreTestingKnobs{
+				// We want to track when leases get acquired and when they are renewed.
+				// We also want to know when acquiring blocks to test lease renewal.
+				LeaseAcquiredEvent: func(_ sqlbase.TableDescriptor, _ error) {
+
+					atomic.AddInt32(&testAcquiredCount, 1)
+				},
+				LeaseAcquireResultBlockEvent: func(_ sql.LeaseAcquireBlockType) {
+					atomic.AddInt32(&testAcquisitionBlockCount, 1)
+				},
+			},
+		},
+	}
+	params.LeaseManagerConfig = base.NewLeaseManagerConfig()
+	// The lease jitter is set to ensure newer leases have higher
+	// expiration timestamps.
+	params.LeaseManagerConfig.TableDescriptorLeaseJitterFraction = 0.0
+	// The renewal timeout is set to be the duration, so background
+	// renewal should begin immediately after accessing a lease.
+	params.LeaseManagerConfig.TableDescriptorLeaseRenewalTimeout =
+		params.LeaseManagerConfig.TableDescriptorLeaseDuration
+
+	t := newLeaseTest(testingT, params)
+	defer t.cleanup()
+
+	if _, err := t.db.Exec(`
+CREATE DATABASE t;
+CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR);
+`); err != nil {
+		t.Fatal(err)
+	}
+
+	tableDesc := sqlbase.GetTableDescriptor(t.kvDB, "t", "test")
+	dbID := tableDesc.ParentID
+	tableName := tableDesc.Name
+
+	// Acquire the first lease.
+	ts, e1, err := t.node(1).AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, tableName)
+	if err != nil {
+		t.Fatal(err)
+	} else if err := t.release(1, ts); err != nil {
+		t.Fatal(err)
+	} else if count := atomic.LoadInt32(&testAcquiredCount); count != 1 {
+		t.Fatalf("expected 1 lease to be acquired, but acquired %d times",
+			count)
+	}
+
+	// Reset testAcquisitionBlockCount as the first acqusition will always block.
+	atomic.StoreInt32(&testAcquisitionBlockCount, 0)
+
+	testutils.SucceedsSoon(t, func() error {
+		// Acquire another lease. At first this will be the same lease, but
+		// eventually we will asynchronously renew a lease and our acquire will get
+		// a newer lease.
+		ts, e2, err := t.node(1).AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, tableName)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer func() {
+			if err := t.release(1, ts); err != nil {
+				t.Fatal(err)
+			}
+		}()
+
+		// We check for the new expiry time because if our past acquire triggered
+		// the background renewal, the next lease we get will be the result of the
+		// background renewal.
+		if e2.WallTime <= e1.WallTime {
+			return errors.Errorf("expected new lease expiration (%s) to be after old lease expiration (%s)",
+				e2, e1)
+		} else if count := atomic.LoadInt32(&testAcquiredCount); count < 2 {
+			return errors.Errorf("expected at least 2 leases to be acquired, but acquired %d times",
+				count)
+		} else if blockCount := atomic.LoadInt32(&testAcquisitionBlockCount); blockCount > 0 {
+			t.Fatalf("expected repeated lease acquisition to not block, but blockCount is: %d", blockCount)
+		}
+		return nil
+	})
 }
--- a/pkg/sql/schema_changer_test.go
+++ b/pkg/sql/schema_changer_test.go
@@ -40,11 +40,13 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
 	"github.com/cockroachdb/cockroach/pkg/util/protoutil"
 	"github.com/cockroachdb/cockroach/pkg/util/retry"
 	"github.com/cockroachdb/cockroach/pkg/util/stop"
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
+	"github.com/cockroachdb/cockroach/pkg/util/tracing"
 )
 
 // asyncSchemaChangerDisabled can be used to disable asynchronous processing
@@ -175,6 +177,7 @@ func TestSchemaChangeProcess(t *testing.T) {
 	stopper := stop.NewStopper()
 	cfg := base.NewLeaseManagerConfig()
 	leaseMgr := sql.NewLeaseManager(
+		log.AmbientContext{Tracer: tracing.NewTracer()},
 		&base.NodeIDContainer{},
 		*kvDB,
 		hlc.NewClock(hlc.UnixNano, time.Nanosecond),


lego:
Ignore the first commit which is part of #18989.

lego:
While acquiring a new lease and re-acquiring (not getting a new lease) seem to be performing about the same as before (might still be worth profiling these), it appears that initializing an async acquisition is far slower than I would hope.

```
ACQUIRING NEW LEASE
Time: 3.967999ms

RE-ACQUIRING (no new lease)
Time: 940.374µs

BEGIN ASYNC ACQUISITION
Time: 2.135845ms
```

pkg/sql/lease.go
```diff
@@ -1115,16 +1141,35 @@ func nameMatchesTable(table *sqlbase.TableDescriptor, dbID sqlbase.ID, tableName
...
+				t.mu.group.DoChan("acquire", func() (interface{}, error) {
+					// defer tracing.FinishSpan(span)
+					log.Warningf(context.Background(), "TableDescriptor async renewal")
+					err := t.acquireNodeLease(context.Background(), m, hlc.Timestamp{})
+					if err != nil {
+						log.Error(context.Background(), errors.Wrapf(err, "TableDescriptor async lease acquisition failed"))
+					}
+					return nil, err
+				})
+			}
```
vivekmenezes: I'm not happy about the code above because it is likely to be non performant. Instead I think what we can do is add a timer to tableVersionState and modify m.tablename.get() above to check if the timer is set. If it is not set, it can create a timer using ``` func AfterFunc(d Duration, f func()) *Timer``` from the time package. The function can renew the lease when it fires and reset the timer to nil.

Additionally when the tableVersionState is released the timer should be stopped. If AfterFunc() is called during the stop then it should wait until Afterfunc sets the timer to nil rather than setting it to nil on its own.

Thoughts?

lego: I think the timer check and creation don't belong in `m.tableNames.get` because the method is currently is just focused on retrieving the lease from the table cache. Otherwise, a few additional parameters need to get passed in _just_ to do the timer initialization to renew (LeaseManager, table ID).

Instead, the code added there could be replaced with
```
if ...timer == nil {
   timer = timer.AfterFunc(s.durationUntilRenewal(), func() { m.Acquire(ctx, hcl.Timestamp{}, tableVersion.ID) })
}
```

The uncertainty I have is where the timer should live. You suggest that we reset the timer when we release a `tableVersionState`, so in that case, it lives there. If we get a new lease, we will have a new tableVersionState, so won't we end up initializing a second timer? 

(unrelated to your comment) I realized this code could instead be restructured to just call `LeaseManager.Acquire` and instead put the `DoChan`  in `tableState.acquire`. The only addition to `AcquireByName` is to check for preemptive renewal and then call `LeaseManager.Acquire`. That's much cleaner because 1) it's not exposing the `DoChan` internals here, and 2) `LeaseManager.Acquire` already does the exact same table name lookup. I was also confused when `tableState.acquire` wasn't always called previously, so this would make more sense.

That's why I suggested the AfterFunc to be much simpler above.

lego:
Rebased the patch on the singleflight branch as that's now merged.

Previously this patch was a bit more complicated, which resulted in +1ms just when the lease renewal was being kicked off. Instead, this patch is now simplified and on first lease access, it creates a timer to fire later at the desired refresh time. It also uses a lockless manner to enter the timer initialization and broadcast that it has been initialized.

 Sidenote: A lock is still involved in `findTableState`, although the lock is held for a very short period. Because of this, I created a benchmark to assess if this impacts anything. It seems that there is not an impactful performance change, as I think this code path is run about once per statement/plan. (I believe the methodology is correct, it uses a parallel benchmark to attempt to cause lock contention.)

```
name                        old time/op  new time/op  delta
LeaseAcquireByNameCached-8   406ns ± 0%   507ns ± 0%   ~     (p=1.000 n=1+1)
```

---

---

*[pkg/sql/lease.go, line 1220 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwhhxqOuC7B0URqe23j:-KwhhxqPlgqLxsnOk9oj:b3d8y1c) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1220)):*
> ```Go
> 			if t := m.findTableState(tableVersion.ID, true); atomic.CompareAndSwapInt32(&t.mu.renewalTimerSet, 0, 1) {
> 				durationUntilRenewal := time.Duration(tableVersion.expiration.WallTime-hlc.UnixNano()) - m.LeaseStore.leaseRenewalTimeout
> 				// innerCtx, span := tracing.ForkCtxSpan(ctx, "[async] lease refresh")
> ```

@andreimatei any idea why `ForkCtxSpan` would not work? When this line is uncommented and `innerCtx` is used within the `DoChan` spawned it seems to always result in `context canceled` errors.

My best guess is that the current span is a `noopSpan` so a no new context is made and we just use the short-lived parent context. If this is true, then `ForkCtxSpan` is incorrect in stating it `Returns the new context`. Otherwise, I'm at a loss.

Any suggestions for a workaround? `context.Background()`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-Kw_hwGL0AXMBXqfvnHf:b-g3o87x)*

bdarnell:
I haven't looked at the code, but should we take this opportunity to make table leases much shorter? We wanted a long duration because renewals are disruptive, but if that's no longer true it would be nice to have much shorter leases so that you don't have to wait 5 minutes for a schema change after killing a node (which is something that our users seem to run into with some regularity). 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-KwklX4Uaizh7ijX4b-2:b-97hep)*

lego:
Are there any other cons for shortening the table lease life?

By allowing table versions to increment faster (say 2 minutes), will this cause issues for long running queries or other processes? My assumption is no, as it seems schema changes only continue when those queries are finished -- from the table descriptor lease RFC
> [schema changing will] wait for the table descriptor change to propagate to all nodes in the cluster and all uses of the previous version to finish.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-Kwkm7Hy1wkit9__GVg1:bgroxdw)*

lego:
Another con is renewals can still disruptive but to a lesser extent. If the new lifespan is 2 minutes, and only one query happens every 2 minutes (or very close to 2 minutes), queries will still get blocked because of acquisition. (And not disruptive if you have any amount of queries prior to the expiry).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-Kwko6nt9UAr-vELN0RG:b-ela5uo)*

andreimatei:
If anybody has complained about not being able to do schema changes after a node was kill-9'ed as Ben claims, than a duration of a minute of so sounds good to me. However, if there's indications of leases not being released when a node was more gently killed, than that's likely a bug.

Long running queries are not supposed to be affected by schema changes except if they're running in SNAPSHOT isolation and they're pushed. However, that's theory. I'm not sure to what extent the code implements the desires of the latest version of that RFC.

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/lease.go, line 1220 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwhhxqOuC7B0URqe23j:-KwpNP-pC-pVrBICJbxe:bbk5scl) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1220)):*

@andreimatei any idea why `ForkCtxSpan` would not work? When this line is uncommented and `innerCtx` is used within the `DoChan` spawned it seems to always result in `context canceled` errors.

My best guess is that the current span is a `noopSpan` so a no new context is made and we just use the short-lived parent context. If this is true, then `ForkCtxSpan` is incorrect in stating it `Returns the new context`. Otherwise, I'm at a loss.

Any suggestions for a workaround? `context.Background()`?
</blockquote></details>

`ForkCtxSpan` refers to creating a new span, different from the caller's span. It has nothing to do with the parent-child relationship between the two contexts - meaning that the returned ctx is derived from the caller's ctx and hence it inherits the cancellation.
If you don't want that (and you don't), then you have to pass in a context that's not derived from the caller's (like `context.Background()` or an annotation thereof).

The tricky thing in this case is that you have to integrate with the server's stopper (and also separately draining?) so that the acquisition doesn't happen if... something.
Not sure how to do that, exactly. For the stopper part, you probably have to start either a task or a worker that calls `timer.Stop()` when the stopper is quiescing. Also check out `timer.WithCancel` for stopping an acquisition if the acquisition actually started.

---

*[pkg/sql/lease.go, line 1224 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwpMKdwPAg4rrTOZewu:-KwpMKdwPAg4rrTOZewv:b-bce5r7) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1224)):*
> ```Go
> 				t.mu.preemptiveRenewalTimer = time.AfterFunc(durationUntilRenewal, func() {
> 					t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
> 						// defer tracing.FinishSpan(span)
> ```

who is supposed to finish `span` when the `DoChan` call joins an existing group?

---

*[pkg/sql/lease.go, line 1229 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwpOTxCEZWo0lCEyl3b:-KwpOTxCEZWo0lCEyl3c:bkr85vl) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1229)):*
> ```Go
> 						token, err := t.acquireNodeLease(innerCtx, m, hlc.Timestamp{})
> 						if err != nil {
> 							log.Error(innerCtx, errors.Wrapf(err, "lease acquisition failed"))
> ```

nit: if you're logging an error, there's no reason to create a new error which will just be thrown away. Just use `log.Errorf()` to provide the extra message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-KwpKxq5LDHlIumTVWRa:b-nnr5h9)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/lease.go, line 1220 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwhhxqOuC7B0URqe23j:-KwpfyOMZ7_gPYbw3ps3:bpngx5z) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1220)):*

`ForkCtxSpan` refers to creating a new span, different from the caller's span. It has nothing to do with the parent-child relationship between the two contexts - meaning that the returned ctx is derived from the caller's ctx and hence it inherits the cancellation.
If you don't want that (and you don't), then you have to pass in a context that's not derived from the caller's (like `context.Background()` or an annotation thereof).

The tricky thing in this case is that you have to integrate with the server's stopper (and also separately draining?) so that the acquisition doesn't happen if... something.
Not sure how to do that, exactly. For the stopper part, you probably have to start either a task or a worker that calls `timer.Stop()` when the stopper is quiescing. Also check out `timer.WithCancel` for stopping an acquisition if the acquisition actually started.
</blockquote></details>

acquireNodeLease` already handles draining, so I think it's fine on that account. (Or is more necessary?)

As for stopping the acquisition before the timer starts, we could potentially do one of two things. I think the first option is a better approach.

One option is retrofit `RefreshLeases` to do it:
```go
case <-s.ShouldQuiesce():
  // Where m is the MeaseManager
  m.Lock()
  for _, t := m.mu.tables {
    if atomic.LoadInt32(&t.mu.renewalTimerSet) == 1 {
      t.mu.preemptiveRenewalTimer.Stop()
    }
  }
  m.Unlock()
```

This could also just be a separate goroutine we initialize alongside `RefreshRoutines` (or inside `NewLeaseManager`... not sure about the preferred style I'd imagine inside NewLeaseManager due to it being a "requirement" to stop correctly). Both strategies only need one goroutine to any amount of tables.

Alternatively, for each renewal we begin we can also start a cleanup routine which will stop the timer:
```go
if  ... {
   // Use a channel to finish the cleanup worker if the timer fires before ShouldQuiesce is true.
   c := make(chan, struct{})
   time.AfterFunc(...,
      // Signal the cleanup worker that it's no longer needed.
      c <- struct{}{} 
      DoChan(..., func() {
         // ...
      })
   })
  // Start the clean up worker
  t.stopper.RunWorker(context.Background(), func(ctx context.Context) {
    select {
      <-t.stopper.ShouldQuiesce():
        t.mu.preemptiveRenewalTimer.Stop()
        return
      <-c:
        return
    }
  })
}
```

---

*[pkg/sql/lease.go, line 1224 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwpMKdwPAg4rrTOZewu:-KwpsD-xdFfcBDAiv4c9:b13zd3l) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1224)):*

who is supposed to finish `span` when the `DoChan` call joins an existing group?
</blockquote></details>

Good catch. The context creation (and span cleanup) should happen entirely inside the DoChan closure, or in the AfterFunc closure.

I've also realized there is a problem: who sets `renewalTimerSet` back to `0` if this joins an existing group. To get around this we can instead set that value back after the `DoChan` results channel returns.

One last thing. Would it be preferable to replace the atomic values with enums? They will need to be wrapped with `int32(...)` though :/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/lease.go, line 1220 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwhhxqOuC7B0URqe23j:-Kwva1thXKiHyZp1XQp2:beoqt1k) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1220)):*

acquireNodeLease` already handles draining, so I think it's fine on that account. (Or is more necessary?)

As for stopping the acquisition before the timer starts, we could potentially do one of two things. I think the first option is a better approach.

One option is retrofit `RefreshLeases` to do it:
```go
case <-s.ShouldQuiesce():
  // Where m is the MeaseManager
  m.Lock()
  for _, t := m.mu.tables {
    if atomic.LoadInt32(&t.mu.renewalTimerSet) == 1 {
      t.mu.preemptiveRenewalTimer.Stop()
    }
  }
  m.Unlock()
```

This could also just be a separate goroutine we initialize alongside `RefreshRoutines` (or inside `NewLeaseManager`... not sure about the preferred style I'd imagine inside NewLeaseManager due to it being a "requirement" to stop correctly). Both strategies only need one goroutine to any amount of tables.

Alternatively, for each renewal we begin we can also start a cleanup routine which will stop the timer:
```go
if  ... {
   // Use a channel to finish the cleanup worker if the timer fires before ShouldQuiesce is true.
   c := make(chan, struct{})
   time.AfterFunc(...,
      // Signal the cleanup worker that it's no longer needed.
      c <- struct{}{} 
      DoChan(..., func() {
         // ...
      })
   })
  // Start the clean up worker
  t.stopper.RunWorker(context.Background(), func(ctx context.Context) {
    select {
      <-t.stopper.ShouldQuiesce():
        t.mu.preemptiveRenewalTimer.Stop()
        return
      <-c:
        return
    }
  })
}
```
</blockquote></details>

For now, I implemented the first approach I've suggested.

---

*[pkg/sql/lease.go, line 1224 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwpMKdwPAg4rrTOZewu:-Kwv_qPU5fpLzfJYvp7p:bsqddsf) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1224)):*

Good catch. The context creation (and span cleanup) should happen entirely inside the DoChan closure, or in the AfterFunc closure.

I've also realized there is a problem: who sets `renewalTimerSet` back to `0` if this joins an existing group. To get around this we can instead set that value back after the `DoChan` results channel returns.

One last thing. Would it be preferable to replace the atomic values with enums? They will need to be wrapped with `int32(...)` though :/
</blockquote></details>

Fixed both the span issue and what I mentioned.

---

*[pkg/sql/lease.go, line 1229 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwpOTxCEZWo0lCEyl3b:-Kwv_q5UblF4pfaSBDn3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b9ec6a88fa77435f69cb80e5f17e9f50b1ec4973/pkg/sql/lease.go#L1229)):*

nit: if you're logging an error, there's no reason to create a new error which will just be thrown away. Just use `log.Errorf()` to provide the extra message.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

andreimatei:
as discussed, maybe we can get rid of the timer

---

---

*[pkg/base/config.go, line 91 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9Kv8VmJu1yDjQpiDk:-Kx9Kv8VmJu1yDjQpiDl:b-p8bn5r) ([raw file](https://github.com/cockroachdb/cockroach/blob/7905053556311e2e79178740bd717dba2c03af0f/pkg/base/config.go#L91)):*
> ```Go
> 	// DefaultTableDescriptorLeaseRenewalTimeout is the default time
> 	// before a lease expires when acquisition to renew the lease begins.
> 	DefaultTableDescriptorLeaseRenewalTimeout = 1 * time.Minute
> ```

nit: `1 *` is not necessary

---

*[pkg/sql/lease.go, line 1219 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9LUGd-TeDBhosa0PR:-Kx9LUGd-TeDBhosa0PS:bfnaqo1) ([raw file](https://github.com/cockroachdb/cockroach/blob/7905053556311e2e79178740bd717dba2c03af0f/pkg/sql/lease.go#L1219)):*
> ```Go
> 			// been set. The new timer is set to fire leaseRenewalTimeout
> 			// before the lease expires to execute a lease renewal function.
> 			if t := m.findTableState(tableVersion.ID, true); atomic.CompareAndSwapInt32(&t.mu.renewalTimerSet, 0, 1) {
> ```

please comment the `true` -> it's in the style guide now!

---

*[pkg/sql/lease_test.go, line 931 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9KAEhBZvZxL5vN-Dd:-Kx9KAEhBZvZxL5vN-De:bcak4fg) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf2a02a43ecfbb17f5749f372ae03b48ea194f8a/pkg/sql/lease_test.go#L931)):*
> ```Go
> 	b.RunParallel(func(pb *testing.PB) {
> 		for pb.Next() {
> 			_, _, err := leaseManager.AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, tableName)
> ```

line over 100 chars. Please set a limit in your editor. Probably above too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-Kx9MiVRI5h7WRV_ZF0J:bdmd6ne)*

lego:
Done! :)

---

---

*[pkg/base/config.go, line 91 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9Kv8VmJu1yDjQpiDk:-Kx9NNAYbQPTp0u1Stjz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7905053556311e2e79178740bd717dba2c03af0f/pkg/base/config.go#L91)):*

nit: `1 *` is not necessary
</blockquote></details>

Done.

---

*[pkg/sql/lease.go, line 1219 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9LUGd-TeDBhosa0PR:-Kx9NucZquGx2UKWmvml:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7905053556311e2e79178740bd717dba2c03af0f/pkg/sql/lease.go#L1219)):*

please comment the `true` -> it's in the style guide now!
</blockquote></details>

Done.

---

*[pkg/sql/lease_test.go, line 931 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9KAEhBZvZxL5vN-Dd:-Kx9Q20FjMyYegF4fxEj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf2a02a43ecfbb17f5749f372ae03b48ea194f8a/pkg/sql/lease_test.go#L931)):*

line over 100 chars. Please set a limit in your editor. Probably above too.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-Kx9RBaAN5war9dE8a0_:b-elbe8b)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/lease.go, line 1219 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9LUGd-TeDBhosa0PR:-Kx9UCFz_yBnl1lg1ds0:bgja53d) ([raw file](https://github.com/cockroachdb/cockroach/blob/7905053556311e2e79178740bd717dba2c03af0f/pkg/sql/lease.go#L1219)):*

Done.
</blockquote></details>

(Also adjusted the bool to `false`. If a `tableVersion` is found, we should not need to create a `tableState`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 590 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9ZjfBp1ZxfSeCm2SE:-Kx9ZjfBp1ZxfSeCm2SF:b-xboaxn) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L590)):*
> ```Go
> 		// This is atomically updated to prevent multiple routines from
> 		// entering renewal initialization.
> 		renewalInProgress int32
> ```

if this is an atomic, it should be commented as such. If it can be accessed without holding `mu`, why is it under `mu`?

---

*[pkg/sql/lease.go, line 1202 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9YOAWrfkM6DokwQw-:-Kx9YOAWrfkM6DokwQw0:b-b4eemo) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1202)):*
> ```Go
> // A transaction using this descriptor must ensure that its
> // commit-timestamp < expiration-time. Care must be taken to not modify
> // the returned descriptor. A timer may be set to renew a lease in the
> ```

there's a couple of comments about "timer" left in the commit. They should all go, right?

---

*[pkg/sql/lease.go, line 1216 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9YWUM2A1fiEeBPMyi:-Kx9YWUM2A1fiEeBPMyj:bkqom94) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1216)):*
> ```Go
> 			durationUntilExpiration := time.Duration(tableVersion.expiration.WallTime - hlc.UnixNano())
> 			durationUntilRenewal := durationUntilExpiration - m.LeaseStore.leaseRenewalTimeout
> 			if t := m.findTableState(tableVersion.ID, false /* create */); durationUntilRenewal < 0 && atomic.CompareAndSwapInt32(&t.mu.renewalInProgress, 0, 1) {
> ```

I don't even need to measure that line to know...

---

*[pkg/sql/lease.go, line 1216 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9YnBMpwF15Oo1EX59:-Kx9YnBMpwF15Oo1EX5A:b-8onedu) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1216)):*
> ```Go
> 			durationUntilExpiration := time.Duration(tableVersion.expiration.WallTime - hlc.UnixNano())
> 			durationUntilRenewal := durationUntilExpiration - m.LeaseStore.leaseRenewalTimeout
> 			if t := m.findTableState(tableVersion.ID, false /* create */); durationUntilRenewal < 0 && atomic.CompareAndSwapInt32(&t.mu.renewalInProgress, 0, 1) {
> ```

`t ` can be nil and crash the condition, no? Or if it can't be nil for some reason, let's assert that.

---

*[pkg/sql/lease.go, line 1221 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9_SpZDEcTMC5KiEZZ:-Kx9_SpZDEcTMC5KiEZ_:b-as8jsa) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1221)):*
> ```Go
> 					innerCtx = t.stopper.WithCancel(innerCtx)
> 					resultChan, _ := t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
> 						defer tracing.FinishSpan(span)
> ```

who closes the span if this call is not the "group leader"? Or are we relying on this always being the group leader because of the `renewalInProgress` check, let's assert that we indeed were the leader using the `DoChan` bool result. But what if someone is acquiring a lease through another code path?

---

*[pkg/sql/lease.go, line 1222 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9_80KU37NFbLi08Vq:-Kx9_80KU37NFbLi08Vr:bpt2g8k) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1222)):*
> ```Go
> 					resultChan, _ := t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
> 						defer tracing.FinishSpan(span)
> 						log.Infof(innerCtx, "lease acquisition beginning")
> ```

`Infof` is way to high. Also below. Also please put more info in these messages - the table name or smth

---

*[pkg/sql/lease_test.go, line 931 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9KAEhBZvZxL5vN-Dd:-Kx9XcGSruHctazXVMNd:b-9ce6wj) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf2a02a43ecfbb17f5749f372ae03b48ea194f8a/pkg/sql/lease_test.go#L931)):*

Done.
</blockquote></details>

Thanks. As a nit for the future - it's great if you can fix things in the right commit. What I do is address comments based on the revision on which they were left (assuming the reviewer reviewed commit by commit). Reviewable shows you the review on which the comment was left in the top-right corner of the comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 590 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9ZjfBp1ZxfSeCm2SE:-Kx9hS4H5RrbHoKSM4_U:b-wx9bxr) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L590)):*

if this is an atomic, it should be commented as such. If it can be accessed without holding `mu`, why is it under `mu`?
</blockquote></details>

Done. (Yep, no reason it needed to be under there as it didn't use `mu`.)

---

*[pkg/sql/lease.go, line 1202 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9YOAWrfkM6DokwQw-:-Kx9i03ZWfkLV9yWantk:b-snh328) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1202)):*

there's a couple of comments about "timer" left in the commit. They should all go, right?
</blockquote></details>

Done here and elsewhere.

---

*[pkg/sql/lease.go, line 1216 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9YWUM2A1fiEeBPMyi:-Kx9g4p3PE-3h4JDa_Ey:b-bwgxqw) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1216)):*

I don't even need to measure that line to know...
</blockquote></details>

Yea go-vet was recombining this in what seems to be sane line breaks :/. I guess breaking on the && is the most reasonable allowed line break.

---

*[pkg/sql/lease.go, line 1216 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9YnBMpwF15Oo1EX59:-Kx9iUTZ7g6uWOC_5ifE:be6zh18) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1216)):*

`t ` can be nil and crash the condition, no? Or if it can't be nil for some reason, let's assert that.
</blockquote></details>

I don't believe `t` can be nil here right now, but in if we begin deleting `tableState` (e.g. if a table is deleted) then this check will be required, so I've added it.

---

*[pkg/sql/lease.go, line 1221 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9_SpZDEcTMC5KiEZZ:-Kx9gS1Rz9GozBXpi4hI:b2twuld) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1221)):*

who closes the span if this call is not the "group leader"? Or are we relying on this always being the group leader because of the `renewalInProgress` check, let's assert that we indeed were the leader using the `DoChan` bool result. But what if someone is acquiring a lease through another code path?
</blockquote></details>

Oof yea I shuffled the `innerCtx` again after removing the timer. The span can be initialized inside the `DoChan` func fine so it's only created and cleaned up if this call runs.

---

*[pkg/sql/lease.go, line 1222 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9_80KU37NFbLi08Vq:-Kx9l54CFx7a_tu3CwKU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease.go#L1222)):*

`Infof` is way to high. Also below. Also please put more info in these messages - the table name or smth
</blockquote></details>

Done.

---

*[pkg/sql/lease_test.go, line 931 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9KAEhBZvZxL5vN-Dd:-Kx9gt2V-EHdUwv1JQRj:b-viap9k) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf2a02a43ecfbb17f5749f372ae03b48ea194f8a/pkg/sql/lease_test.go#L931)):*

Thanks. As a nit for the future - it's great if you can fix things in the right commit. What I do is address comments based on the revision on which they were left (assuming the reviewer reviewed commit by commit). Reviewable shows you the review on which the comment was left in the top-right corner of the comment.
</blockquote></details>

TIL!

(The revisions don't seem to explicitly state a corresponding commit, besides the fact that there are now two "active" revisions. Is that right?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

andreimatei:
LGTM but one of the tests seems to have failed on TC

---

Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/lease.go, line 1222 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9yrCLSUNn02vfbe-0:-Kx9yrCLSUNn02vfbe-1:bjc6d0p) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease.go#L1222)):*
> ```Go
> 				t.stopper.RunWorker(context.Background(), func(ctx context.Context) {
> 					resultChan, _ := t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
> 						innerCtx, span := tracing.ForkCtxSpan(ctx, "[async] lease refresh")
> ```

nit: overwrite `ctx` to not have two contexts in scope.

---

*[pkg/sql/lease.go, line 1225 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9ySf55r7CQbqfs5Kq:-Kx9ySf55r7CQbqfs5Kr:bs6hx1e) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease.go#L1225)):*
> ```Go
> 						innerCtx = t.stopper.WithCancel(innerCtx)
> 						defer tracing.FinishSpan(span)
> 						if log.V(2) {
> ```

`VEventf()` instead of this construct.

If you want `if V(2)` for whatever reason, there's a newer `ExpensiveLogEnabled` or smth that you should use (it's linked from `V()`).

---

*[pkg/sql/lease_test.go, line 931 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9KAEhBZvZxL5vN-Dd:-KxA0WQYl-86M0DMwjd2:b8b9k5h) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf2a02a43ecfbb17f5749f372ae03b48ea194f8a/pkg/sql/lease_test.go#L931)):*

TIL!

(The revisions don't seem to explicitly state a corresponding commit, besides the fact that there are now two "active" revisions. Is that right?)
</blockquote></details>

each revision is a commit. If you look at the beginning of the review comments, you see the commit message for each revision.
Another pro tip: always rebase before pushing so that the current generation of commits form a suffix of the revisions (i.e. always force all commits to be rewritten by doing a rebase).

---

*[pkg/sql/lease_test.go, line 126 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxA0ICFA5A0Q8Q6cTEV:-KxA0ICGHJLzcDGidTzk:b-iksoun) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease_test.go#L126)):*
> ```Go
> 	nodeID uint32, dbID sqlbase.ID, name string,
> ) (*sqlbase.TableDescriptor, hlc.Timestamp, error) {
> 	return t.node(nodeID).AcquireByName(context.TODO(), t.server.Clock().Now(), dbID, name)
> ```

is this wrapper worth it? I had to look a bunch for it, and I didn't know what the literal `1` in all the calls means.

---

*[pkg/sql/lease_test.go, line 953 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxA-OQsi64B8mIvdfAj:-KxA-OQsi64B8mIvdfAk:b-gmi7f4) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease_test.go#L953)):*
> ```Go
> // This test makes sure the lease gets renewed automatically in the background
> // if the lease is about to expire, without blocking.
> func TestLeaseRefreshedAutomatically(testingT *testing.T) {
> ```

"refresh" is called "renewal" (async renewal?) elsewhere, isn't it?

---

*[pkg/sql/lease_test.go, line 1043 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9zScyJW6ZPR4uHlUy:-Kx9zScyJW6ZPR4uHlUz:b2j2a4w) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease_test.go#L1043)):*
> ```Go
> 
> // This test makes sure that async lease refreshing doesn't block any routines
> func TestLeaseRefreshDoesntBlock(testingT *testing.T) {
> ```

This is an unusual test. If I imaging correctly (and I have to imagine because the comment you've added doesn't really say anything - what's "any routines"?), it's testing that an acquisition is not blocked by a renewal. Is there a risk of that happening / is the risk worth this complicated test? What kind of programming mistake would make the async renewal sync?
Your choice.

---

*[pkg/sql/lease_test.go, line 1043 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9zzXgrGiHzzirdbXY:-Kx9zzXgrGiHzzirdbXZ:b2j8tli) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease_test.go#L1043)):*
> ```Go
> 
> // This test makes sure that async lease refreshing doesn't block any routines
> func TestLeaseRefreshDoesntBlock(testingT *testing.T) {
> ```

"refresh" is called "renewal" in other code, isn't it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-KxA0uT87ryDPQ6_v8vB:b-bn29r6)*

lego:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions.

---

*[pkg/sql/lease.go, line 1222 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9yrCLSUNn02vfbe-0:-KxAW-pETD-QWlih3nab:b-xvofn0) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease.go#L1222)):*

nit: overwrite `ctx` to not have two contexts in scope.
</blockquote></details>

So having `ctx` blows a linter rule for shadowing the scope, so that was the breaking point for moving this code to a separate function.

---

*[pkg/sql/lease.go, line 1225 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9ySf55r7CQbqfs5Kq:-KxATHTm3jLdsBDYY6jn:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease.go#L1225)):*

`VEventf()` instead of this construct.

If you want `if V(2)` for whatever reason, there's a newer `ExpensiveLogEnabled` or smth that you should use (it's linked from `V()`).
</blockquote></details>

Done.

---

*[pkg/sql/lease_test.go, line 126 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxA0ICFA5A0Q8Q6cTEV:-KxARvBrefa8hYQ5rK5S:bbc0a00) ([raw file](https://github.com/cockroachdb/cockroach/blob/165da7007cfcc3f61b42331dae4cbd3289e16c0c/pkg/sql/lease_test.go#L126)):*

is this wrapper worth it? I had to look a bunch for it, and I didn't know what the literal `1` in all the calls means.
</blockquote></details>

Yea, that's a fair point. It's clearer with the direct call to `node`, and I also added a comment to `node`.

---

*[pkg/sql/lease_test.go, line 953 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxA-OQsi64B8mIvdfAj:-KxAW15czIg1x-_axOmt:bmb98js) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease_test.go#L953)):*

"refresh" is called "renewal" (async renewal?) elsewhere, isn't it?
</blockquote></details>

Fixed here and elsewhere.

---

*[pkg/sql/lease_test.go, line 1043 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9zzXgrGiHzzirdbXY:-KxAVmzUsAq7UjRosAgh:b-793g3i) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease_test.go#L1043)):*

"refresh" is called "renewal" in other code, isn't it?
</blockquote></details>

Done. (a instances of refresh in the tests were updated)

---

*[pkg/sql/lease_test.go, line 1043 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9zScyJW6ZPR4uHlUy:-KxAW3qQPgeYjw9bwLEr:bldy0da) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease_test.go#L1043)):*

This is an unusual test. If I imaging correctly (and I have to imagine because the comment you've added doesn't really say anything - what's "any routines"?), it's testing that an acquisition is not blocked by a renewal. Is there a risk of that happening / is the risk worth this complicated test? What kind of programming mistake would make the async renewal sync?
Your choice.
</blockquote></details>

Uhh yea it is peculiar. There is no risk of it happening and is an attempted test to make sure it doesn't block. Something that the previous test also asserts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

pkg/base/config.go
```diff
@@ -79,12 +79,16 @@ const (
 	// the jitter fraction. Jittering is done to prevent multiple leases
 	// from being renewed simultaneously if they were all acquired
 	// simultaneously.
-	DefaultTableDescriptorLeaseDuration = 5 * time.Minute
+	DefaultTableDescriptorLeaseDuration = 2 * time.Minute
```
vivekmenezes: No strong reason to reduce this other than the case where a node dies in the middle of a schema change. That's a pretty rare case, and the tradeoff here is that we renew leases more frequently.

pkg/sql/lease.go
```diff
@@ -1192,14 +1233,27 @@ func nameMatchesTable(table *sqlbase.TableDescriptor, dbID sqlbase.ID, tableName
...
+
+			// Atomically check and begin a renewal if one has not already
+			// been set.
+			durationUntilExpiration := time.Duration(tableVersion.expiration.WallTime - hlc.UnixNano())
+			durationUntilRenewal := durationUntilExpiration - m.LeaseStore.leaseRenewalTimeout
+			if t := m.findTableState(tableVersion.ID, false /* create */); t != nil &&
+				durationUntilRenewal < 0 && atomic.CompareAndSwapInt32(&t.renewalInProgress, 0, 1) {
+				// Start the renewal. When it finishes, it will reset t.renewalInProgress.
+				t.stopper.RunWorker(context.Background(), t.renewalFn(m, tableVersion))
+			}
```
vivekmenezes: RunWorker() is used to start goroutines that run forever. How about modifying this code as:

```
if time.Duration(tableVersion.expiration.WallTime - timestamp.WallTime) < m.LeaseStore.leaseRenewalTimeout {
   if t := m.findTableState(tableVersion.ID, false /* create */); t != nil && atomic.CompareAndSwapInt32(&t.renewalInProgress, 0, 1) {
 				// Start the renewal. When it finishes, it will reset t.renewalInProgress.
 				if err := t.stopper.RunTaskAsync(context.Background(), t.renewalFn(m, tableVersion)); err != {
return &tableVersion.TableDescriptor, tableVersion.expiration, err
}
 	}
}
```

pkg/sql/lease.go
```diff
@@ -970,6 +978,38 @@ func (t *tableState) purgeOldVersions(
...
+// renewalFn returns a closure used for renewing a lease.
+// t.renewalInProgress must be set to 1 before calling.
+func (t *tableState) renewalFn(
+	m *LeaseManager, tableVersion *tableVersionState,
+) func(context.Context) {
+	return func(ctx context.Context) {
+		resultChan, _ := t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
+			var span opentracing.Span
+			ctx, span = tracing.ForkCtxSpan(ctx, "[async] lease renewal")
+			ctx = t.stopper.WithCancel(ctx)
```
vivekmenezes: you shouldn't need all this opentracing stuff

vivekmenezes (approved):
Nicely done!

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 1222 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9yrCLSUNn02vfbe-0:-KxBPZQP-uflRYz1yrTk:b-menzhg) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease.go#L1222)):*

So having `ctx` blows a linter rule for shadowing the scope, so that was the breaking point for moving this code to a separate function.
</blockquote></details>

I see. As a matter of fact I wasn't thinking of shadowing ctx; I was thinking of overwriting it.
```
var span Span
ctx, span = tracing.Fork()
```
But I guess that wouldn't be so nice, since ctx is captured from the outer scope and it might be weird to change it (even though `ctx` is not currently used elsewhere). The fact that you're forced to capture is the real problem here; I think the `DoChan()` interface should take a ctx and pass it to the closure (like the stopper interface). @nvanbenschoten, right? I think we should do that now before the interface gets more used.

Short of that, moving the closure to a different method sounds good to me, but the function that returns a function is weird and I don't see why it's necessary. 
```
t.stopper.RunWorker(context.Background(), func(ctx) {
  t.renewalFn(ctx, m, tableVersion))
})
```

---

*[pkg/sql/lease.go, line 989 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxBZ7i6_-syfi4XVAuT:-KxBZ7i6_-syfi4XVAuU:bjzq9dj) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L989)):*
> ```Go
> 		resultChan, _ := t.mu.group.DoChan(acquireGroupKey, func() (interface{}, error) {
> 			var span opentracing.Span
> 			ctx, span = tracing.ForkCtxSpan(ctx, "[async] lease renewal")
> ```

I think ForkCtxSpan is not gonna do what you want - it's not gonna create a span since there's no spans in contexts passed by `RunWorker` and `RunAsyncTask`. If you want to have a span here (which I think is a good thing, I'm not sure why Vivek doesn't like it), you should manually create a root span (see `EnsureCtx` and other functions around).

---

*[pkg/sql/lease.go, line 990 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxBaPrJfw_GITQvXYUk:-KxBaPrKDOdEZYFBKcH-:bjj6kio) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L990)):*
> ```Go
> 			var span opentracing.Span
> 			ctx, span = tracing.ForkCtxSpan(ctx, "[async] lease renewal")
> 			ctx = t.stopper.WithCancel(ctx)
> ```

since you've told me that `acquireNodeLease` will short-circuit if the server is draining, I'm not sure the `stopper.WithCancel()` is necessary / gives us anything any more.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

vivekmenezes:
I'm okay with adding in the new span

lego:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

*[pkg/base/config.go, line 82 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KwhYImNMD2km4uFbQ26-r10-82:-KxDsw8nZv2Fq8O80J1O:b-eq1e5y) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/base/config.go#L82)):*

No strong reason to reduce this other than the case where a node dies in the middle of a schema change. That's a pretty rare case, and the tradeoff here is that we renew leases more frequently.
</blockquote></details>

Okay, I've changed it back to 5 minutes for now. It's easy to change in the future!

---

*[pkg/sql/lease.go, line 1222 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-Kx9yrCLSUNn02vfbe-0:-KxDoanWuk0-ld3--1u_:b-gfwkpr) ([raw file](https://github.com/cockroachdb/cockroach/blob/4652a1d30afb01bf3c212b1c79f9026d40c3f22a/pkg/sql/lease.go#L1222)):*

I see. As a matter of fact I wasn't thinking of shadowing ctx; I was thinking of overwriting it.
```
var span Span
ctx, span = tracing.Fork()
```
But I guess that wouldn't be so nice, since ctx is captured from the outer scope and it might be weird to change it (even though `ctx` is not currently used elsewhere). The fact that you're forced to capture is the real problem here; I think the `DoChan()` interface should take a ctx and pass it to the closure (like the stopper interface). @nvanbenschoten, right? I think we should do that now before the interface gets more used.

Short of that, moving the closure to a different method sounds good to me, but the function that returns a function is weird and I don't see why it's necessary. 
```
t.stopper.RunWorker(context.Background(), func(ctx) {
  t.renewalFn(ctx, m, tableVersion))
})
```
</blockquote></details>

Ah yea, I had done a very roundabout away to pass in the context variables. Thanks!

---

*[pkg/sql/lease.go, line 989 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxBZ7i6_-syfi4XVAuT:-KxDrDJDzi5ItPRIMg2I:b5tghhi) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L989)):*

I think ForkCtxSpan is not gonna do what you want - it's not gonna create a span since there's no spans in contexts passed by `RunWorker` and `RunAsyncTask`. If you want to have a span here (which I think is a good thing, I'm not sure why Vivek doesn't like it), you should manually create a root span (see `EnsureCtx` and other functions around).
</blockquote></details>

:+1:. What I added uses `tracing.NewTracer()`. Should this instead use the server's `AmbientCtx` to get the tracer? (I'm not too certain what the tracer will do, in particular, if we aren't crossing boundaries for this span.)

---

*[pkg/sql/lease.go, line 990 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxBaPrJfw_GITQvXYUk:-KxDs9I4fYzCFqyQW0Im:b-86hfw2) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L990)):*

since you've told me that `acquireNodeLease` will short-circuit if the server is draining, I'm not sure the `stopper.WithCancel()` is necessary / gives us anything any more.
</blockquote></details>

If both the stopper only stops when the server is stopping and draining happens when the server is gracefully closing, then I agree there's no purpose.

(removed it).

---

*[pkg/sql/lease.go, line 1255 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KvZ01COSbtB0vdT5bvy-r10-1255:-KxDljLZhcOxMnceTA6I:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L1255)):*

RunWorker() is used to start goroutines that run forever. How about modifying this code as:

```
if time.Duration(tableVersion.expiration.WallTime - timestamp.WallTime) < m.LeaseStore.leaseRenewalTimeout {
   if t := m.findTableState(tableVersion.ID, false /* create */); t != nil && atomic.CompareAndSwapInt32(&t.renewalInProgress, 0, 1) {
 				// Start the renewal. When it finishes, it will reset t.renewalInProgress.
 				if err := t.stopper.RunTaskAsync(context.Background(), t.renewalFn(m, tableVersion)); err != {
return &tableVersion.TableDescriptor, tableVersion.expiration, err
}
 	}
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

pkg/sql/lease.go
```diff
@@ -1192,14 +1227,32 @@ func nameMatchesTable(table *sqlbase.TableDescriptor, dbID sqlbase.ID, tableName
...
 	if tableVersion != nil {
 		if !timestamp.Less(tableVersion.ModificationTime) {
+
+			// Atomically check and begin a renewal if one has not already
+			// been set.
+
+			if time.Duration(tableVersion.expiration.WallTime-timestamp.WallTime) < m.LeaseStore.leaseRenewalTimeout {
+				if t := m.findTableState(tableVersion.ID, false /* create */); t != nil && atomic.CompareAndSwapInt32(&t.renewalInProgress, 0, 1) {
+					// Start the renewal. When it finishes, it will reset t.renewalInProgress.
+					if err := t.stopper.RunAsyncTask(context.Background(), "lease renewal", func(ctx context.Context) {
```
vivekmenezes: I believe you can pass in the current context here

lego:
Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 1247 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KvZ01COSbtB0vdT5bvy-r11-1247:-KxEGWlKtPm9NHGm_exs:b-1444ec) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1247)):*

I believe you can pass in the current context here
</blockquote></details>

Unfortunately not. The task routine may not begin until the session is closed and by then the existing context is canceled. (Which occurs every time in practice.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

vivekmenezes:
Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 1247 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KvZ01COSbtB0vdT5bvy-r11-1247:-KxEZml0seq9RreH_9rB:b-9ndcxm) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1247)):*

Unfortunately not. The task routine may not begin until the session is closed and by then the existing context is canceled. (Which occurs every time in practice.)
</blockquote></details>

Looking at the implementation of RunAsyncTask() it looks like the context is replaced with a new context before running the goroutine

ctx, span := tracing.ForkCtxSpan(ctx, taskName)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 989 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxBZ7i6_-syfi4XVAuT:-KxEhHkuYT8r8IpIXblZ:bf4laz7) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L989)):*

:+1:. What I added uses `tracing.NewTracer()`. Should this instead use the server's `AmbientCtx` to get the tracer? (I'm not too certain what the tracer will do, in particular, if we aren't crossing boundaries for this span.)
</blockquote></details>

We Don't Like It When People Call NewTracer ™. If you create a new tracer, you gotta call Configure() on it; and in any case the tracer is supposed to be a singleton.
Indeed, you should get a tracer from an AmbientContext. Either use the server's AmbientContext or give a new one to the `LeaseManager` or to whomever.

---

*[pkg/sql/lease.go, line 1244 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxEeuU8fEvdoMbpqwh7:-KxEeuU8fEvdoMbpqwh8:bq4ddzp) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1244)):*
> ```Go
> 			// been set.
> 
> 			if time.Duration(tableVersion.expiration.WallTime-timestamp.WallTime) < m.LeaseStore.leaseRenewalTimeout {
> ```

looooong. style guide bro.

---

*[pkg/sql/lease.go, line 1245 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxEfAIxXKhV3xW-mtfp:-KxEfAIxXKhV3xW-mtfq:b-nvdfkv) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1245)):*
> ```Go
> 
> 			if time.Duration(tableVersion.expiration.WallTime-timestamp.WallTime) < m.LeaseStore.leaseRenewalTimeout {
> 				if t := m.findTableState(tableVersion.ID, false /* create */); t != nil && atomic.CompareAndSwapInt32(&t.renewalInProgress, 0, 1) {
> ```

looong

---

*[pkg/sql/lease.go, line 1247 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KvZ01COSbtB0vdT5bvy-r11-1247:-KxEgSd3dQ5ubF-tOFjd:b-vxyhu4) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1247)):*

Looking at the implementation of RunAsyncTask() it looks like the context is replaced with a new context before running the goroutine

ctx, span := tracing.ForkCtxSpan(ctx, taskName)
</blockquote></details>

True, but the ctx passed to the closure is still a child of the passed-in context (so it inherits the cancellation).
(I don't know if that behavior makes sense, but it is the current code).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

lego:
Going to merge later today if there are no more comments. Thanks a ton folks! :tada:

---

Review status: 0 of 5 files reviewed at latest revision, 9 unresolved discussions.

---

*[pkg/sql/lease.go, line 989 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxBZ7i6_-syfi4XVAuT:-KxEoWZim3TAoTtxX8Y_:b7hot6n) ([raw file](https://github.com/cockroachdb/cockroach/blob/c6a05112f15d25cea04105b9a0874d04f00eff72/pkg/sql/lease.go#L989)):*

We Don't Like It When People Call NewTracer ™. If you create a new tracer, you gotta call Configure() on it; and in any case the tracer is supposed to be a singleton.
Indeed, you should get a tracer from an AmbientContext. Either use the server's AmbientContext or give a new one to the `LeaseManager` or to whomever.
</blockquote></details>

:+1: sgtm. Done. (pulled in the context into lease manager).

---

*[pkg/sql/lease.go, line 1244 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxEeuU8fEvdoMbpqwh7:-KxEogbGwPq4Guq8J520:b-j2kt9d) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1244)):*

looooong. style guide bro.
</blockquote></details>

Erf... dealt with :'(

---

*[pkg/sql/lease.go, line 1245 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxEfAIxXKhV3xW-mtfp:-KxEojMcuy4bh54t9DTF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1245)):*

looong
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-KxEols1i-jgAXH-BKTV:bcn5vp)*

andreimatei:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-:-KxEqC7zDkOMMs_AXX03:bnfp4nl)*

lego:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lease.go, line 1244 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxEeuU8fEvdoMbpqwh7:-KxEpuqAaRfOKzHQuGcQ:ba7n0r5) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1244)):*

Erf... dealt with :'(
</blockquote></details>

TBH distributing a basic set of git hooks in the style guide would maybe be a good idea. e.g. I've since added a `make lint` pre-push and now a pre-commit line length check.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

jordanlewis:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/sql/lease.go, line 1244 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005#-KxEeuU8fEvdoMbpqwh7:-KxF8FyYRKds-Ams9Pf5:bxvazpn) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0bfa704c6796ca41eaf5160f9b372c25a48057d/pkg/sql/lease.go#L1244)):*

TBH distributing a basic set of git hooks in the style guide would maybe be a good idea. e.g. I've since added a `make lint` pre-push and now a pre-commit line length check.
</blockquote></details>

@lego it'd probably be useful if you shared those hooks - git hooks are hard to write!

Also, we should consider adding long line fixes to `crlfmt` which exists just for this purpose.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19005)*

PR #19007, opened 2017-10-03
vendor: bump dep and re-sync with lock

knz:
Due to a bug in the previously vendored version of `dep`, `dep` was
failing to properly check whether the contents of `vendor` were in
sync with `Gopkg.lock`. This in turn allowed `vendor` to diverge
without anyone noticing.

This patch uses the latest version of `dep` to vendor itself and
re-sync the vendor tree.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -22,10 +22,11 @@
   revision = "a2fdd780c9a50455cecd249b00bdc3eb73a78e31"
 
 [[projects]]
-  branch = "2.x"
+  branch = "parse-constraints-with-dash-in-pre"
   name = "github.com/Masterminds/semver"
   packages = ["."]
-  revision = "139cc0982c53f1367af5636b12b7643cd03757fc"
+  revision = "a93e51b5a57ef416dac8bb02d11407b6f55d8929"
+  source = "https://github.com/carolynvs/semver.git"
 
 [[projects]]
   name = "github.com/Masterminds/vcs"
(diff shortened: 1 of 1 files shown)


dt (approved):
Thanks for digging into this!

knz:
The merit is yours! I wouldn't have known where to start if it were not for your suggestion to reset `vendor`.

Gopkg.lock
```diff
@@ -22,10 +22,11 @@
   revision = "a2fdd780c9a50455cecd249b00bdc3eb73a78e31"
 
 [[projects]]
-  branch = "2.x"
+  branch = "parse-constraints-with-dash-in-pre"
   name = "github.com/Masterminds/semver"
   packages = ["."]
-  revision = "139cc0982c53f1367af5636b12b7643cd03757fc"
+  revision = "a93e51b5a57ef416dac8bb02d11407b6f55d8929"
+  source = "https://github.com/carolynvs/semver.git"
```
RaduBerinde: Where did this fork come from?

knz: `dep ensure` decided it should be there. I suppose it should have been there since the beginning.

dt: https://github.com/golang/dep/blob/master/Gopkg.toml#L4

dt: we use one-big-vendor for our tools's deps and our own deps (arguably, we shouldn't).

`dep` depends on a fork of semvar under one of the individual dep maintainers' accounts, so now we do too.

RaduBerinde: 👍 Thanks!

dt: I can investigate moving us to a use separate `build/vendor` for tools and their deps, which should help free up more of our deps (and minimize the probability of a conflict at least for awhile).

benesch: I'm leery of yet another submodule—especially one that needs to be managed by dep. Have we hit actual incompatibilities yet?

PR #18508, opened 2017-09-15
sql: store information about keys in orderingInfo, elide distinctNode

RaduBerinde:
#### sql: store information about keys in orderingInfo

We are now able to store information about column sets that form "keys", i.e.
all rows are distinct when projected on that set.

The existing structure had a very limited form of this, only keeping track if
the columns in the ordering form a key. Allowing arbitrary sets will enable more
optimizations.

#### sql: elide distinctNode if we have a key

An immediate application of the new key information is to elide distinct nodes
when possible.

PR #19008, opened 2017-10-03
sql: don't enable "kv tracing" for SHOW TRACE FOR <stmt>

andreimatei:
A "kv trace" is the a set of log messages around SQL<->KV interaction.
The concept is not very well defined at the moment, but it comprises two
categories of messages:
1. Details of KV requests being sent for execution, in particular
pretty-printed scan requests.
2. Per-row messages. In particular, every row that's being returned by
KV for processing by SQL.

These messages are the only ones that are returned as results of
SHOW _KV_ TRACE FOR <stmt>/SESSION.
The situation gets interesting for SHOW TRACE FOR ... (no KV specified).
In this case, before this patch, all the "kv trace" messages were
included in the results. This patch changes that and eliminates the
per-row messages. This is an improvement, per-row messages (as opposed
to per-RPC or higher level) is too verbose for most people that want
traces.

Tactically, a couple of "kv messages" that weren't per row are now
always logged (and hence traced) regardless of whether kv tracing is
enabled or not, and otherwise kv tracing is now no longer enabled for
SHOW TRACE FOR.

This was already the case before this patch, but now this comes to the
foreground: "kv tracing" means one thing externally (as in SHOW KV TRACE
FOR ...), versus internally, in the code, where "kvTracedEnabled" and
similarly-named members refer to "per-row" messages. A round of renames
is in order, but that's left for another day.

Benchstat - no difference
https://gist.github.com/andreimatei/c8ed8f850b59a9eec67b2fb190160108

This patch has no influence on the new index-join benchmark:
```
make bench PKG=./pkg/sql BENCHES=BenchmarkIndexJoin TESTFLAGS='-count 10'
make bench PKG=./pkg/sql BENCHES=BenchmarkIndexJoin TESTFLAGS='-test.benchmem -count 10'

[andrei:~/work/src/github.com/cockroachdb/cockroach] sql-kv-tracing+ ± benchstat /tmp/{before,after}
name         old time/op  new time/op  delta
IndexJoin-8  8.54ms ± 4%  8.54ms ± 1%   ~     (p=0.605 n=9+9)

[andrei:~/work/src/github.com/cockroachdb/cockroach] sql-kv-tracing+ ± benchstat /tmp/{before-mem,after-mem}
IndexJoin-8    8.64ms ± 2%    8.59ms ± 6%    ~     (p=0.243 n=9+10)

IndexJoin-8    3.62MB ± 0%    3.62MB ± 0%  -0.03%  (p=0.000 n=9+9)

IndexJoin-8     11.6k ± 0%     11.6k ± 0%    ~     (p=0.314 n=10+10)
```

### DIFF
--- a/pkg/sql/bench_test.go
+++ b/pkg/sql/bench_test.go
@@ -1033,3 +1033,42 @@ func BenchmarkPlanning(b *testing.B) {
 		})
 	}
 }
+
+// BenchmarkIndexJoin measure an index-join with 1000 rows.
+func BenchmarkIndexJoin(b *testing.B) {
+	// The table will have an extra column not contained in the index to force a
+	// join with the PK.
+	create := `
+		 CREATE TABLE tidx (
+				 k INT NOT NULL,
+				 v INT NULL,
+				 extra STRING NULL,
+				 CONSTRAINT "primary" PRIMARY KEY (k ASC),
+				 INDEX idx (v ASC),
+				 FAMILY "primary" (k, v, extra)
+		 )
+		`
+	// We'll insert 1000 rows with random values below 1000 in the index. We'll
+	// then query the index with a query that retrieves all the data (but the
+	// optimizer doesn't know that).
+	insert := "insert into tidx(k,v) select generate_series(1,1000), (random()*1000)::int"
+	s, db, _ := serverutils.StartServer(b, base.TestServerArgs{UseDatabase: "bench"})
+	defer s.Stopper().Stop(context.TODO())
+
+	if _, err := db.Exec(`CREATE DATABASE bench`); err != nil {
+		b.Fatal(err)
+	}
+	if _, err := db.Exec(create); err != nil {
+		b.Fatal(err)
+	}
+	if _, err := db.Exec(insert); err != nil {
+		b.Fatal(err)
+	}
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		if _, err := db.Exec("select * from tidx where v < 1000"); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
--- a/pkg/sql/logictest/testdata/logic_test/show_trace
+++ b/pkg/sql/logictest/testdata/logic_test/show_trace
@@ -31,7 +31,6 @@ SELECT span, message, operation FROM [SHOW TRACE FOR SELECT 1]
 (0,0)  === SPAN START: sql txn implicit ===  sql txn implicit
 (0,1)  === SPAN START: starting plan ===     starting plan
 (0,2)  === SPAN START: consuming rows ===    consuming rows
-(0,2)  output row: [1]                       consuming rows
 (0,2)  plan completed execution              consuming rows
 (0,2)  resources released, stopping trace    consuming rows
 
@@ -44,7 +43,6 @@ SELECT span, message, operation FROM [SHOW TRACE FOR EXECUTE x]
 (0,0)  === SPAN START: sql txn implicit ===  sql txn implicit
 (0,1)  === SPAN START: starting plan ===     starting plan
 (0,2)  === SPAN START: consuming rows ===    consuming rows
-(0,2)  output row: [1]                       consuming rows
 (0,2)  plan completed execution              consuming rows
 (0,2)  resources released, stopping trace    consuming rows
 
--- a/pkg/sql/session.go
+++ b/pkg/sql/session.go
@@ -1419,7 +1419,12 @@ type SessionTracing struct {
 // span that have already been created will not be traced).
 //
 // StopTracing() needs to be called to finish this trace.
-func (st *SessionTracing) StartTracing(recType tracing.RecordingType, traceKV bool) error {
+//
+// Args:
+// kvTracingEnabled: If set, the traces will also include "KV trace" messages -
+//   verbose messages around the interaction of SQL with KV. Some of the messages
+//   are per-row.
+func (st *SessionTracing) StartTracing(recType tracing.RecordingType, kvTracingEnabled bool) error {
 	if st.enabled {
 		return errors.Errorf("already tracing")
 	}
@@ -1433,7 +1438,7 @@ func (st *SessionTracing) StartTracing(recType tracing.RecordingType, traceKV bo
 
 	tracing.StartRecording(sp, recType)
 	st.enabled = true
-	st.kvTracingEnabled = traceKV
+	st.kvTracingEnabled = kvTracingEnabled
 	st.recordingType = recType
 	return nil
 }
--- a/pkg/sql/show.go
+++ b/pkg/sql/show.go
@@ -334,7 +334,7 @@ WHERE message LIKE 'fetched: %'
 		plan.Close(ctx)
 		return nil, err
 	}
-	tracePlan, err := p.makeTraceNode(stmtPlan)
+	tracePlan, err := p.makeTraceNode(stmtPlan, n.OnlyKVTrace /* kvTracingEnabled */)
 	if err != nil {
 		plan.Close(ctx)
 		stmtPlan.Close(ctx)
--- a/pkg/sql/sqlbase/fk.go
+++ b/pkg/sql/sqlbase/fk.go
@@ -235,21 +235,19 @@ func makeFKInsertHelper(
 	return h, nil
 }
 
-func (h fkInsertHelper) checkAll(ctx context.Context, row parser.Datums, traceKV bool) error {
+func (h fkInsertHelper) checkAll(ctx context.Context, row parser.Datums) error {
 	if len(h.fks) == 0 {
 		return nil
 	}
 	for idx := range h.fks {
-		if err := h.checkIdx(ctx, idx, row, traceKV); err != nil {
+		if err := h.checkIdx(ctx, idx, row); err != nil {
 			return err
 		}
 	}
 	return h.checker.runCheck(ctx, nil, row)
 }
 
-func (h fkInsertHelper) checkIdx(
-	ctx context.Context, idx IndexID, row parser.Datums, traceKV bool,
-) error {
+func (h fkInsertHelper) checkIdx(ctx context.Context, idx IndexID, row parser.Datums) error {
 	fks := h.fks
 	for i, fk := range fks[idx] {
 		nulls := true
@@ -322,21 +320,19 @@ func makeFKDeleteHelper(
 	return h, nil
 }
 
-func (h fkDeleteHelper) checkAll(ctx context.Context, row parser.Datums, traceKV bool) error {
+func (h fkDeleteHelper) checkAll(ctx context.Context, row parser.Datums) error {
 	if len(h.fks) == 0 {
 		return nil
 	}
 	for idx := range h.fks {
-		if err := h.checkIdx(ctx, idx, row, traceKV); err != nil {
+		if err := h.checkIdx(ctx, idx, row); err != nil {
 			return err
 		}
 	}
-	return h.checker.runCheck(ctx, row, nil)
+	return h.checker.runCheck(ctx, row, nil /* newRow */)
 }
 
-func (h fkDeleteHelper) checkIdx(
-	ctx context.Context, idx IndexID, row parser.Datums, traceKV bool,
-) error {
+func (h fkDeleteHelper) checkIdx(ctx context.Context, idx IndexID, row parser.Datums) error {
 	for i := range h.fks[idx] {
 		if err := h.checker.addCheck(row, &h.fks[idx][i]); err != nil {
 			return err
@@ -381,12 +377,12 @@ func makeFKUpdateHelper(
 }
 
 func (fks fkUpdateHelper) checkIdx(
-	ctx context.Context, idx IndexID, oldValues, newValues parser.Datums, traceKV bool,
+	ctx context.Context, idx IndexID, oldValues, newValues parser.Datums,
 ) error {
-	if err := fks.inbound.checkIdx(ctx, idx, oldValues, traceKV); err != nil {
+	if err := fks.inbound.checkIdx(ctx, idx, oldValues); err != nil {
 		return err
 	}
-	return fks.outbound.checkIdx(ctx, idx, newValues, traceKV)
+	return fks.outbound.checkIdx(ctx, idx, newValues)
 }
 
 // CollectSpans implements the FkSpanCollector interface.
--- a/pkg/sql/sqlbase/kvfetcher.go
+++ b/pkg/sql/sqlbase/kvfetcher.go
@@ -103,10 +103,6 @@ type txnKVFetcher struct {
 	// rangeInfos are deduped, so they're not ordered in any particular way and
 	// they don't map to kvFetcher.spans in any particular way.
 	rangeInfos []roachpb.RangeInfo
-
-	// traceKV determines whether or not to emit a log message every time a Scan
-	// is sent.
-	traceKV bool
 }
 
 func (f *txnKVFetcher) getRangesInfo() []roachpb.RangeInfo {
@@ -172,7 +168,6 @@ func makeKVFetcher(
 	useBatchLimit bool,
 	firstBatchLimit int64,
 	returnRangeInfo bool,
-	traceKV bool,
 ) (txnKVFetcher, error) {
 	if firstBatchLimit < 0 || (!useBatchLimit && firstBatchLimit != 0) {
 		return txnKVFetcher{}, errors.Errorf("invalid batch limit %d (useBatchLimit: %t)",
@@ -206,7 +201,6 @@ func makeKVFetcher(
 		useBatchLimit:   useBatchLimit,
 		firstBatchLimit: firstBatchLimit,
 		returnRangeInfo: returnRangeInfo,
-		traceKV:         traceKV,
 	}, nil
 }
 
@@ -230,16 +224,15 @@ func (f *txnKVFetcher) fetch(ctx context.Context) error {
 		}
 	}
 
-	if f.traceKV {
-		scanStr := "Scan "
+	if log.ExpensiveLogEnabled(ctx, 2) {
+		buf := bytes.NewBufferString("Scan ")
 		for i, span := range f.spans {
 			if i != 0 {
-				scanStr += ", "
+				buf.WriteString(", ")
 			}
-			scanStr += span.String()
+			buf.WriteString(span.String())
 		}
-
-		log.VEventf(ctx, 2, scanStr)
+		log.VEvent(ctx, 2, buf.String())
 	}
 
 	// Reset spans in preparation for adding resume-spans below.
--- a/pkg/sql/sqlbase/rowfetcher.go
+++ b/pkg/sql/sqlbase/rowfetcher.go
@@ -229,7 +229,7 @@ func (rf *RowFetcher) StartScan(
 		firstBatchLimit++
 	}
 
-	f, err := makeKVFetcher(txn, spans, rf.reverse, limitBatches, firstBatchLimit, rf.returnRangeInfo, traceKV)
+	f, err := makeKVFetcher(txn, spans, rf.reverse, limitBatches, firstBatchLimit, rf.returnRangeInfo)
 	if err != nil {
 		return err
 	}
--- a/pkg/sql/sqlbase/rowwriter.go
+++ b/pkg/sql/sqlbase/rowwriter.go
@@ -250,7 +250,7 @@ func (ri *RowInserter) InsertRow(
 		}
 	}
 
-	if err := ri.Fks.checkAll(ctx, values, traceKV); err != nil {
+	if err := ri.Fks.checkAll(ctx, values); err != nil {
 		return err
 	}
 
@@ -619,12 +619,14 @@ func (ru *RowUpdater) UpdateRow(
 	}
 
 	if rowPrimaryKeyChanged {
-		if err := ru.Fks.checkIdx(ctx, ru.Helper.TableDesc.PrimaryIndex.ID, oldValues, ru.newValues, traceKV); err != nil {
+		if err := ru.Fks.checkIdx(
+			ctx, ru.Helper.TableDesc.PrimaryIndex.ID, oldValues, ru.newValues,
+		); err != nil {
 			return nil, err
 		}
 		for i := range newSecondaryIndexEntries {
 			if !bytes.Equal(newSecondaryIndexEntries[i].Key, secondaryIndexEntries[i].Key) {
-				if err := ru.Fks.checkIdx(ctx, ru.Helper.Indexes[i].ID, oldValues, ru.newValues, traceKV); err != nil {
+				if err := ru.Fks.checkIdx(ctx, ru.Helper.Indexes[i].ID, oldValues, ru.newValues); err != nil {
 					return nil, err
 				}
 			}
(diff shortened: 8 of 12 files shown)


andreimatei:
I probably want to cherry-pick this in some 1.1.x release.

jordanlewis:
Ah I see - this explains why `SHOW KV TRACE` actually currently returns *more* stuff than `SHOW TRACE`, which seems a little unintuitive to me. I like this idea.

I'm surprised that no logic tests change here. I think we should have at least a modicum of test coverage - our traces are actually a very important feature that we can't allow regression in.

---

Reviewed 2 of 2 files at r1.
Review status: 2 of 8 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/sql/session.go, line 1427 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZE1kRd2tBHp81l8Hi:-KvZE1kSYHwxVYOSELQ-:bfz4vpl) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/session.go#L1427)):*
> ```Go
> //   verbose messages around the interaction of SQL with KV. Some of the messages
> //   are per-row.
> func (st *SessionTracing) StartTracing(recType tracing.RecordingType, KVTracingEnabled bool) error {
> ```

nit: capitalized argument names seem weird, or at least not "Roman" with respect to our codebase's Rome...

---

*[pkg/sql/trace.go, line 44 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZFVY_8ozvP4QZxsFF:-KvZFVY_8ozvP4QZxsFG:b2yzult) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/trace.go#L44)):*
> ```Go
> 	// If set, the trace will also include "KV trace" messages - verbose messages
> 	// around the interaction of SQL with KV. Some of the messages are per-row.
> 	KVTracingEnabled bool
> ```

nit: i don't think this should be capitalized since it's not exported. Or is it?

---

*[pkg/sql/trace.go, line 59 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZFZMyue5HLNvNiAkj:-KvZFZMyue5HLNvNiAkk:b-t3slf1) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/trace.go#L59)):*
> ```Go
> //   verbose messages around the interaction of SQL with KV. Some of the
> //   messages are per-row.
> func (p *planner) makeTraceNode(plan planNode, KVTracingEnabled bool) (planNode, error) {
> ```

nit: caps again.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZF0ROAeMSnNp1agq7:-KvZF0ROAeMSnNp1agq8:bxwy0u4) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/sqlbase/kvfetcher.go#L227)):*
> ```Go
> 	}
> 
> 	scanStr := "Scan "
> ```

Hmmm.... not sure if this makes sense. This emits at least 1 Scan event per read operation. Isn't that pretty expensive?

In general it's not obvious to me that we should ever be doing this much unconditional tracing. But I guess that's another fish to fry - we already do a ton of unconditional tracing (including internals like time spent in the command queue). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-:-KvZFrOk7RXvucKvP6Uo:bpk69s4)*

andreimatei:
> Ah I see - this explains why SHOW KV TRACE actually currently returns more stuff than SHOW TRACE, which seems a little unintuitive to me. I like this idea.

Well, no; before this patch, `SHOW TRACE FOR <stmt>` would include all the messages from `SHOW KV TRACE FOR <stmt>`, so the former should always be a superset. This patch makes the two uncomparable: `SHOW KV TRACE FOR <stmt>` will return a restricted set of messages, but include some per-row ones. The per-row ones are excluded from `SHOW TRACE FOR <session>`. Similarly with the recording done for `SET tracing cluster` vs `SET tracing kv`.
If this is not what you currently see, show me.

"KV tracing" is generally a concept that I have mixed feelings on (I was not involved in creating it). I don't have a clear opinion on whether it deserves to exists and, if it does, exactly what semantics it should have. The thing that I'm interested in and that this patch does, however, is for non-kv tracing to not include per-row messages.


> I'm surprised that no logic tests change here. I think we should have at least a modicum of test coverage - our traces are actually a very important feature that we can't allow regression in.

What test did you have in mind exactly? There is test coverage for tracing, kv and regular. However, the tests are never about checking exactly what a trace looks like; that'd be unwieldy. This PR eliminates a few messages from non-kv traces. How did you expect that to have been tested?
I don't see a particular need for extra testing, but if you have a reasonable idea, I'm opened.

---

Review status: 2 of 8 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/sql/session.go, line 1427 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZE1kRd2tBHp81l8Hi:-Kv_HF9s3eGhNNy17Z0t:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/session.go#L1427)):*

nit: capitalized argument names seem weird, or at least not "Roman" with respect to our codebase's Rome...
</blockquote></details>

Done.

---

*[pkg/sql/trace.go, line 44 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZFVY_8ozvP4QZxsFF:-Kv_GWz_IZ7MoFj7YDKJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/trace.go#L44)):*

nit: i don't think this should be capitalized since it's not exported. Or is it?
</blockquote></details>

Done.

---

*[pkg/sql/trace.go, line 59 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZFZMyue5HLNvNiAkj:-Kv_H95i8H1aInlfeMyH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/trace.go#L59)):*

nit: caps again.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZF0ROAeMSnNp1agq7:-Kv_Er6ueeVZHWqhpMEx:b79zxk1) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/sqlbase/kvfetcher.go#L227)):*

Hmmm.... not sure if this makes sense. This emits at least 1 Scan event per read operation. Isn't that pretty expensive?

In general it's not obvious to me that we should ever be doing this much unconditional tracing. But I guess that's another fish to fry - we already do a ton of unconditional tracing (including internals like time spent in the command queue). 
</blockquote></details>

The key here is "per read operation" - per batch sent for execution. These are big operations. Generally they're RPCs. Each one of them generates tons of messages already. One more here is not a big deal.

Unconditional logging is not particularly expensive when tracing and the respective logging is turned off - we've optimized this a bunch (when tracing/logging is enabled, that's another story). I believe we ran some tests about it. There must be some overhead for the boxing that the untyped `Eventf()` interface must do... We should test again at some point.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-:-Kv_BcLL8N9Iqiy6PBuT:bmu3wiq)*

knz:
Reviewed 2 of 2 files at r1, 6 of 6 files at r3, 6 of 6 files at r4.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZF0ROAeMSnNp1agq7:-Kva1sHcq30tTLK-az4S:bw106x7) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/sqlbase/kvfetcher.go#L227)):*

The key here is "per read operation" - per batch sent for execution. These are big operations. Generally they're RPCs. Each one of them generates tons of messages already. One more here is not a big deal.

Unconditional logging is not particularly expensive when tracing and the respective logging is turned off - we've optimized this a bunch (when tracing/logging is enabled, that's another story). I believe we ran some tests about it. There must be some overhead for the boxing that the untyped `Eventf()` interface must do... We should test again at some point.

</blockquote></details>

Yes we have ran tests for this and yes it is rather expensive. Sufficiently expensive in fact that Peter had added a separate boolean for this block of code (and related code) before, so it wouldn't run on the regular code path. Check the history of the code. @petermattis what do you think? Right now this code is alloc-heavy (especially the string appends). At the least use a bytes.Buffer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

andreimatei:
Review status: 2 of 8 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZF0ROAeMSnNp1agq7:-KvbreBFajSmVfKnMc6o:b7o8e7l) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/sqlbase/kvfetcher.go#L227)):*

Yes we have ran tests for this and yes it is rather expensive. Sufficiently expensive in fact that Peter had added a separate boolean for this block of code (and related code) before, so it wouldn't run on the regular code path. Check the history of the code. @petermattis what do you think? Right now this code is alloc-heavy (especially the string appends). At the least use a bytes.Buffer.
</blockquote></details>

ah, I though we were arguing about the logging itself, not the code around it. `bytes.Buffer` is a good idea. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

knz:
Reviewed 5 of 6 files at r6.
Review status: 7 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZF0ROAeMSnNp1agq7:-Kvli7000G-3sp2ws02p:b-biair1) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/sqlbase/kvfetcher.go#L227)):*

ah, I though we were arguing about the logging itself, not the code around it. `bytes.Buffer` is a good idea. Done.
</blockquote></details>

`buf.writeString(span.String())`  is still one allocation too many. Make a `Format` method on `span` and give it `buf` as argument.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

knz:
Also, same request as I made to Radu in #18508: please run a micro-benchmark (E.g. exactly the one Radu used) before and after this PR, and report the results. 1) I want to see the exact effects on allocations. @jordanlewis will be interested too. 2) it will give you (like everyone in the team is getting their chance too soon) experience with running these micro-benchmarks and attaching the results to the commit message as a matter of course.

RaduBerinde:
LGTM

jordanlewis:
LGTM

andreimatei:
Check out the new `ExpensiveLogEnabled()`.

---

Review status: 2 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-KvZF0ROAeMSnNp1agq7:-Kvmz_X4NIstJjgQxfa2:b-dvhxh2) ([raw file](https://github.com/cockroachdb/cockroach/blob/eca98ea2a80df677baa30f90cd5076d8bae76299/pkg/sql/sqlbase/kvfetcher.go#L227)):*

`buf.writeString(span.String())`  is still one allocation too many. Make a `Format` method on `span` and give it `buf` as argument.
</blockquote></details>

I'll give you satisfaction - the pretty printing of spans does cost a few alloc (assuming that's a meaningful "cost" :P).
I've gated this behind a new `ExpensiveLogEnabled()` - check it out.
I've added a link to bench results in the PR description.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-:-Kvn-0e7Si_a1EODb4vE:b-7bic23)*

knz:
Review status: 2 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kvn0VpjARpwX6gC0ImK:bbqt25g) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*
> ```Go
> 	}
> 
> 	if log.ExpensiveLogEnabled(ctx, 2) {
> ```

No! We want these spans do be printed in SHOW KV TRACE in all cases.

What about you restore the `traceKV` conditional for this one!?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

andreimatei:
Review status: 2 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kvn0uHwVvMt7CxcW0Wu:b-7mqqoc) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*

No! We want these spans do be printed in SHOW KV TRACE in all cases.

What about you restore the `traceKV` conditional for this one!?
</blockquote></details>

They are printed with `SHOW [KV] TRACE`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

knz:
Review status: 2 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kvn1CkrjWt_h7OrSBfG:b-bvasau) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*

They are printed with `SHOW [KV] TRACE`.
</blockquote></details>

The conditional to establish this is complicated. Do you understand what SpanInContext does? Do you understand what VDepth does?  It's not much cheaper than hitting the heap allocator!
Testing a simple boolean is the  way to go.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

andreimatei:
Review status: 2 of 10 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kvn1whgOBIMBovV30lo:bvk7vif) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*

The conditional to establish this is complicated. Do you understand what SpanInContext does? Do you understand what VDepth does?  It's not much cheaper than hitting the heap allocator!
Testing a simple boolean is the  way to go.
</blockquote></details>

Testing *which* boolean? The "kvTrace" one is the wrong one to trace.
I've given you the benchmarks and they show nothing. I'm telling you that we log a million things per KV request; the condition to evaluate whether this message should be constructed is not more expensive than any of the other a million `log.VEvent()` calls. What else can I do for you?
P.S. What do we do to not have this conversation for every `VEvent()` & friends call?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

RaduBerinde:
:lgtm:

---

Review status: 2 of 10 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/util/log/log.go, line 192 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvo636bDM9sNvHWpT6m:-Kvo636bDM9sNvHWpT6n:b-7qa4mn) ([raw file](https://github.com/cockroachdb/cockroach/blob/ecc293e0ff7774b40f72046bb7e67f0f33ddc898/pkg/util/log/log.go#L192)):*
> ```Go
> // attached to ctx. In particular, if some OpenTracing collection is enabled
> // (e.g. LightStep), that, by itself, does NOT cause the expensive messages to
> // be enabled. SHOW TRACE FOR <stmt> and friends, on the other hand, does cause
> ```

[nit] do

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-:-Kvo6waJDxrBMrLJbWI0:bnfp4nl)*

knz:
On 06-10-17 21:01, Andrei Matei wrote:
> I'm telling you that we log a million things per KV request; the
> condition to evaluate whether this message should be constructed is not
> more expensive than any of the other a million |log.VEvent()| calls.

The fact we have too much overhead on the hot path is undeniable. This
should be reduced, not increased.

> What else can I do for you?

A before/after benchmark comparison on an index join of 1000 rows, for
example, counting both the time overhead and the allocations.

-- 
Raphael 'kena' Poss

andreimatei:
Review status: 2 of 11 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kw1HkwrAibfagSmXbmM:b7brcjz) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*
[Pasting here from your email response. Please be aware that email responds are posted at top comments, so they break conversations.]

The fact we have too much overhead on the hot path is undeniable. This should be reduced, not increased.

> What else can I do for you?

A before/after benchmark comparison on an index join of 1000 rows, for example, counting both the time overhead and the allocations.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

andreimatei:
Added a benchmark and another commit changing a log message in traceNode.

---

Review status: 2 of 11 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kw1IPFRbVDYrCbxhNm4:b-v35ae4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*

[Pasting here from your email response. Please be aware that email responds are posted at top comments, so they break conversations.]

The fact we have too much overhead on the hot path is undeniable. This should be reduced, not increased.

> What else can I do for you?

A before/after benchmark comparison on an index join of 1000 rows, for example, counting both the time overhead and the allocations.
</blockquote></details>

I added a commit with the benchmark. No difference.
```
[andrei:~/work/src/github.com/cockroachdb/cockroach] sql-kv-tracing+ ± benchstat /tmp/{before,after}
name         old time/op  new time/op  delta
IndexJoin-8  8.54ms ± 4%  8.54ms ± 1%   ~     (p=0.605 n=9+9)


[andrei:~/work/src/github.com/cockroachdb/cockroach] sql-kv-tracing+ ± benchstat /tmp/{before-mem,after-mem}
IndexJoin-8    8.64ms ± 2%    8.59ms ± 6%    ~     (p=0.243 n=9+10)

IndexJoin-8    3.62MB ± 0%    3.62MB ± 0%  -0.03%  (p=0.000 n=9+9)

IndexJoin-8     11.6k ± 0%     11.6k ± 0%    ~     (p=0.314 n=10+10)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-:-Kw1IQgv0u7WrG0F4AVL:bdczo3b)*

knz:
:lgtm: modulo extending the commit message with the bench results

---

Reviewed 4 of 8 files at r10, 1 of 1 files at r11, 6 of 8 files at r12, 2 of 2 files at r13.
Review status: 11 of 12 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kw1moo7VgH3NdrkxMoQ:bnemfv1) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*

I added a commit with the benchmark. No difference.
```
[andrei:~/work/src/github.com/cockroachdb/cockroach] sql-kv-tracing+ ± benchstat /tmp/{before,after}
name         old time/op  new time/op  delta
IndexJoin-8  8.54ms ± 4%  8.54ms ± 1%   ~     (p=0.605 n=9+9)


[andrei:~/work/src/github.com/cockroachdb/cockroach] sql-kv-tracing+ ± benchstat /tmp/{before-mem,after-mem}
IndexJoin-8    8.64ms ± 2%    8.59ms ± 6%    ~     (p=0.243 n=9+10)

IndexJoin-8    3.62MB ± 0%    3.62MB ± 0%  -0.03%  (p=0.000 n=9+9)

IndexJoin-8     11.6k ± 0%     11.6k ± 0%    ~     (p=0.314 n=10+10)
```
</blockquote></details>

Brilliant, thanks. Please copy this into the commit message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-:-Kw1n3LUCBgyHgG8rn45:b-lnevib)*

andreimatei:
Review status: 2 of 12 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/sqlbase/kvfetcher.go, line 227 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008#-Kvn0VpjARpwX6gC0ImJ:-Kw1sLsP2rxadhN-j6Rz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2d7d3ee3281e2c0101ebd2aea9d1ea629a74ebe/pkg/sql/sqlbase/kvfetcher.go#L227)):*

Brilliant, thanks. Please copy this into the commit message.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19008)*

Issue #18822, opened 2017-09-27
backup/restore: improved usability work in scope for 1.2


dianasaur323:
Tracking issue for usability work to be accomplished in 1.2 for backup / restore. In order of priority. 

- [ ] Allow restore into a non-existing database https://github.com/cockroachdb/cockroach/issues/19055
- [ ] Save credentials as cluster settings
- [x] Better compatibility with Google Cloud https://github.com/cockroachdb/cockroach/pull/19018#pullrequestreview-68083735
- [ ] Export grants in restore https://github.com/cockroachdb/cockroach/issues/18636 (stretch)

cc @danhhz @mjibson @dt @benesch @cuongdo for comments!

cuongdo:
Could you clarify how GC compatibility should be improved? Also, in what specific ways should restoring interleaved tables be improved?

dt:
I think the GCE fix is supporting explicit credentials, not just instance auth.

dianasaur323:
@cuongdo yep, my understanding was that you currently have to be running within google cloud platform in order to backup / restore into Google Cloud, and I think we should have a better story of compatibility there.

For interleaved tables, my understanding is that currently, we limit what portion of the interleaved table you can restore. I'm moving outside of my comfortable zone of what I know here, so @dt will have to confirm, but my understanding is that the outer "section(?)" of the interleaved table can be restored, but the inner "section(?)" can't without also restoring the full interleaved table.

dt:
yeah, currently it's all-or-nothing when restoring interleaved tables.

dianasaur323:
So I removed the multiple table and interleaved table support, since our 1.2 issue list is aggressive at this point. I figured we might be able to keep these last 3 small usability nits, since they seem like they shouldn't take too long to implement. Correct me if I'm wrong?

mjibson:
LGTM

dianasaur323:
ps, noticed that I was missing export of grants for restore. I think we can do a better job, and since we are investing more time on restore than on import this time around, we can try to get this, and mark it as a stretch if we don't.

dt:
we export grants, we just don't restore them, and instead treat the tables more like newly created tables (inherit from db in which they're created). this is because there were some open questions around how to handle matching users (or not) that we wanted to resolve first.

bdarnell:
[...]

PR #19018, opened 2017-10-04
storageccl: add setting for GCS key

madelynnblue:
Previously the `gs` storage scheme for backup/restore could only use
instance (implicit) auth. The other schemes allowed to pass auth
secrets in the URLS. However Google's auth stuff is much longer
than just an id and secret key: it is a JSON file with various
fields. Marshaling that to base64 to put into a URL would have been a
bit overblown. Also, being forced to hardcode those auth keys forced
operators to execute queries with them even when they would prefer
to have the keys more hidden.

After the recent change to disallow non-root from even viewing cluster
settings, we can now put these auth keys into the cluster settings
and configure the storage providers to use them. We are starting with
Google because it was the only one lacking before.

Add a cluster setting `cloudstorage.gs.default.key` that holds the
contents of a Google Cloud Storage JSON key file. The namespacing is:
`cloudstorage.<scheme>.<name>.<key/id/secret/etc>`. The `name` field
here will start as `default` for all schemes. We may allow users to
add their own custom-named keys, but that will require some changes
to the settings infra, and so may be done later. But the namespacing
is done as is to allow this future enhancement.

Add an AUTH parameter to GCS export storages. There are three valid
values. If the AUTH parameter is not present, the storage will check
if the key setting exists and use it if it does, otherwise fall back
to auth from the instance/environment (hereafter called implicit). A
value of `default` only checks the key, and a value of `implicit`
only uses implicit auth.

See #18822


dt:
@bdarnell brought up some concerns about saving credentials in settings on the feature issue #18822 that we should probably resolve before moving forward here.

dt:
#19031

pkg/ccl/storageccl/export_storage.go
```diff
@@ -45,6 +50,19 @@ const (
...
+	// key in a URI.
+	AuthParam         = "AUTH"
+	authParamImplicit = "implicit"
+	authParamDefault  = "default"
+
+	cloudstoragePrefix           = "cloudstorage"
+	cloudstorageGS               = cloudstoragePrefix + ".gs"
+	cloudstorageDefault          = ".default"
+	cloudstorageGSDefault        = cloudstorageGS + cloudstorageDefault
+	cloudstorageGSJsonkey        = ".jsonkey"
```
dt: 🚲🏡: arguably we don't need `jsonkey`, but it keep symmetry with the other providers where we'll have suffixes for `.key` and `.secret` or whatever.

pkg/ccl/storageccl/export_storage.go
```diff
@@ -427,11 +454,41 @@ func (g *gcsStorage) Conf() roachpb.ExportStorage {
...
+		}
+		if key != "" {
+			source, err := google.JWTConfigFromJSON([]byte(key), scope)
+			if err != nil {
+				return nil, errors.Wrap(err, "creating GCS oauth token source")
+			}
+			opts = append(opts, option.WithTokenSource(source.TokenSource(ctx)))
+		}
+	case authParamImplicit:
+		// do nothing; use implicit params
```
dt: nit: i'd link to SDK doc on behavior.

pkg/ccl/cliccl/load.go
```diff
@@ -169,7 +169,7 @@ func runLoadShow(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	desc, err := sqlccl.ReadBackupDescriptorFromURI(ctx, basepath)
+	desc, err := sqlccl.ReadBackupDescriptorFromURI(ctx, basepath, nil /* cluster.Settings */)
```
dt: here+elsewhere: if you felt so inclined, I'd love to see a alias like `settings.NoSetttingsValues = nil` so we could just use that in place of the commented nil args`nil /* ...*/`

dt (approved):
We also have `TestCloudBackupRestoreGoogleCloudStorage` in sqlccl's tests if you wanted to exercise from SQL too.

pkg/ccl/cliccl/load.go
mjibson: Done.

pkg/ccl/storageccl/export_storage.go
mjibson: Changed to `.key`

pkg/ccl/storageccl/export_storage.go
mjibson: Done.

mjibson:
RFAL. Since the cluster settings stuff is in, I think this can go in now.

mjibson:
ping @dt

dt:
:lgtm:

---

Reviewed 2 of 11 files at r1, 7 of 11 files at r3, 1 of 4 files at r4.
Review status: 5 of 16 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19018#-:-KwBpWToeOkE7T7Zt0hR:bnfp4nl)*

PR #19023, opened 2017-10-04
sql: allow fetching zone configs transactionally

benesch:
sql.GetZoneConfig has useful logic for recursively looking up zone
configs (i.e., first check for a table config, then a database config,
then fall back to the default config), but can only use the cached
SystemConfig. Add a function sql.GetZoneConfigInTxn that uses the same
internal logic but performs consistent lookups using a client.Txn.

For upcoming geo-partitioning work. @mberhault can you take a look or suggest someone else who should?

### DIFF
--- a/pkg/sql/config.go
+++ b/pkg/sql/config.go
@@ -15,8 +15,12 @@
 package sql
 
 import (
+	"golang.org/x/net/context"
+
 	"github.com/cockroachdb/cockroach/pkg/config"
+	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/keys"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
 )
 
@@ -26,10 +30,13 @@ func init() {
 	config.ZoneConfigHook = GetZoneConfig
 }
 
-// GetZoneConfig returns the zone config for the object with 'id'.
-func GetZoneConfig(cfg config.SystemConfig, id uint32) (config.ZoneConfig, bool, error) {
+func getZoneConfig(
+	id uint32, get func(roachpb.Key) (*roachpb.Value, error),
+) (config.ZoneConfig, bool, error) {
 	// Look in the zones table.
-	if zoneVal := cfg.GetValue(sqlbase.MakeZoneKey(sqlbase.ID(id))); zoneVal != nil {
+	if zoneVal, err := get(sqlbase.MakeZoneKey(sqlbase.ID(id))); err != nil {
+		return config.ZoneConfig{}, false, err
+	} else if zoneVal != nil {
 		// We're done.
 		zone, err := config.MigrateZoneConfig(zoneVal)
 		return zone, true, err
@@ -37,7 +44,9 @@ func GetZoneConfig(cfg config.SystemConfig, id uint32) (config.ZoneConfig, bool,
 
 	// No zone config for this ID. We need to figure out if it's a database
 	// or table. Lookup its descriptor.
-	if descVal := cfg.GetValue(sqlbase.MakeDescMetadataKey(sqlbase.ID(id))); descVal != nil {
+	if descVal, err := get(sqlbase.MakeDescMetadataKey(sqlbase.ID(id))); err != nil {
+		return config.ZoneConfig{}, false, err
+	} else if descVal != nil {
 		// Determine whether this is a database or table.
 		var desc sqlbase.Descriptor
 		if err := descVal.GetProto(&desc); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/sql/config.go
```diff
@@ -26,39 +30,69 @@ func init() {
...
+// cached system config.
+func GetZoneConfig(cfg config.SystemConfig, id uint32) (config.ZoneConfig, bool, error) {
+	return getZoneConfig(id, func(key roachpb.Key) (*roachpb.Value, error) {
+		return cfg.GetValue(key), nil
+	})
+}
+
+// GetZoneConfigInTxn is like GetZoneConfig, but uses the provided transaction
+// to perform lookups instead of the cached system config.
+func GetZoneConfigInTxn(
```
mberhault: When would this be used? We call `GetZoneConfig` **a lot** within pretty tight loops (think the various queues).
Is the plan to stop gossiping the zone configs, or parts of it?

benesch: Certainly not planning to stop gossiping the zone config! Rather I'm moving parts of the `./cockroach zone` CLI into SQL, and I want to be able to read/update a zone config within a transaction. Basically, I'm implementing

```
ALTER TABLE db.tbl EXPERIMENTAL SET ZONE 'some yaml'
```

The CLI currently accomplishes this by running:

```
BEGIN;
SELECT config FROM system.zones WHERE id = 51;
--If missing...
SELECT config FROM system.zones WHERE id = 50;
--If still missing...
SELECT config FROM system.zones WHERE id = 0;
UPDATE system.zones SET config = 'some derived config' WHERE id = 51;
COMMIT;
```

(That PR should be out for review later today.)

mberhault: Ah, so it's purely for settings purposes. Consumption of the zone configs would still be through the gossiped (and cached) version? If so, this sees reasonable.

benesch: Yep, exactly!

danhhz:
:lgtm: though I would really like to see a benchmark of `GetZoneConfig` (the one used in the hot path) before and after

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19023#-:-Kvh-MkKjiCMmgyMp92o:b-98obzg)*

benesch:
Done. PTAL.

```
$ ./bin/benchstat old.txt new.txt 
name             old time/op  new time/op  delta
GetZoneConfig-8   840ns ± 3%   836ns ± 1%   ~     (p=0.589 n=10+9)
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19023#-:-Kvh_3jDdFjtYTkszEVl:b-e3ggdi)*

danhhz:
benchmark lgtm modulo one comment. can you rerun your benchstat with -benchmem turned on plz? if allocations don't change, feel free to merge without another round of feedback from me

---

---

*[pkg/sql/config_test.go, line 267 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19023#-Kvi6v7tXvOZN5SdUwK7:-Kvi6v7tXvOZN5SdUwK8:bxh9q1j) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4112b51f8d39619ed68d1f9165bd731b4a48a21/pkg/sql/config_test.go#L267)):*
> ```Go
> 	for i := 0; i < b.N; i++ {
> 		cfg.GetZoneConfigForKey(keys.MakeTablePrefix(keys.MaxReservedDescID + 1))
> 	}
> ```

`b.StopTimer()` at the end, so you don't time your defer

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19023#-:-Kvi6zcOmfEOtBZsWsNl:btf1rjh)*

benesch:
Looks like Go is smart enough to hoist the closure:

```
$ ./bin/benchstat old.txt new.txt 
GetZoneConfig-8     779ns ± 1%     800ns ± 1%  +2.62%  (p=0.000 n=9+10)

GetZoneConfig-8      118B ± 0%      118B ± 0%    ~     (all equal)

GetZoneConfig-8      5.00 ± 0%      5.00 ± 0%    ~     (all equal)
```

benesch:
TFTR!

---

---

*[pkg/sql/config_test.go, line 267 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19023#-Kvi6v7tXvOZN5SdUwK7:-Kvr7aqBNGvf5u5g5dRC:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b4112b51f8d39619ed68d1f9165bd731b4a48a21/pkg/sql/config_test.go#L267)):*

`b.StopTimer()` at the end, so you don't time your defer
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19023#-:-Kvr7clim6SuprXGay8b:b-814g8k)*

PR #19025, opened 2017-10-04
ui: add explanatory comment to install link

couchand:
I went back to check that #18986 had hit every docs link in `pkg/ui` and saw that this one wasn't using `util/docs`.  There's a good reason for that, but since it might not be immediately obvious to everyone in the future, I added a comment.


vilterp:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19025#-:-KvblaEf-dFdoz5U5K1P:bnfp4nl)*

PR #19026, opened 2017-10-04
ui: replace checked-in typing of dagre-layout w/ DefinitelyTyped version

### LINKED ISSUE #19003: ui: use TypeScript declarations from DefinitelyTyped for dagre-layout, once they exist
For the [command-queue PR](https://github.com/cockroachdb/cockroach/pull/18344), I checked in typescript definitions for dagre-layout, since they weren't in DefinitelyTyped. I've now opened a PR to add them to DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/20257

Once this merges, we should delete the checked-in version from our repo.

vilterp:
My PR to add types for `dagre-layout` to DefinitelyTyped was added; now they no longer need to be checked in.

Fixes #19003


couchand (approved):

benesch:
Heads up: typings should go in dev dependencies, not dependencies. This doesn't matter yet, but will after #18349 lands. (Hopefully the error message is obvious enough; the build will fail if you add a package to the wrong place.)

PR #19034, opened 2017-10-04
distsql: check for within 1ulp instead of truncating for decimal comparisons in TestDistAggregationTable

richardwu:
As noted by @RaduBerinde in
https://github.com/cockroachdb/cockroach/pull/18978#discussion_r142716315,
truncating the local and non-local decimal results to 19 precision still
misses an edge case (e.g. when the last two decimal places are 19 and
20).

@mjibson pointed out a test case in the apd package that checks that two decimals
are within 1ulp for some high-precision decimal calculations. This PR
uses the same idiom to check for decimal results that are within 1ulp
which was deemed as an acceptable threshold.

If we happen to discover that 1ulp is not lax enough, we can increase it to a
higher threshold.

### DIFF
--- a/pkg/sql/distsqlplan/aggregator_funcs_test.go
+++ b/pkg/sql/distsqlplan/aggregator_funcs_test.go
@@ -16,11 +16,11 @@ package distsqlplan
 
 import (
 	"fmt"
+	"math/big"
 	"testing"
 
 	"golang.org/x/net/context"
 
-	"github.com/cockroachdb/apd"
 	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/sql/distsqlrun"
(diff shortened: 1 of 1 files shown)


madelynnblue (approved):
LGTM. Did you stress this to make sure it works?

richardwu:
Yup, stressed test for a few minutes with decimal variance calculations.

RaduBerinde:
LGTM

> If we happen to discover that 1ulp is not lax enough, we can increase it to a
higher threshold.

Technically, that would be a bug no? We should use as much internal precision as necessary to guarantee 1ulp

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19034#-:-KvdJ3k30UZrhK5mPupl:b-5ccjqq)*

richardwu:
I haven't fully gone through the derivation of the [parallel algorithm](http://prod.sandia.gov/techlib/access-control.cgi/2008/086212.pdf) from the [local algorithm](https://www.johndcook.com/blog/standard_deviation/) for `STDDEV/VARIANCE`, but it appears they're mathematically sound thus we should expect an allowable max 1ulp difference. I'll update the commit message.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19034#-:-KvdKqaSitVoFPExqvsB:b-5m050a)*

PR #19040, opened 2017-10-04
storage: add assertions that detect HardState proto change

### LINKED ISSUE #17146: storage: improve assertions that prevent HardState proto change
A picture says about as much as a hundred words. From https://reviewable.io/reviews/cockroachdb/cockroach/17051#-KpKIEcA5LZyVrn41G6M

image

pydima:
Closes #17146.

### DIFF
--- a/pkg/storage/below_raft_protos_test.go
+++ b/pkg/storage/below_raft_protos_test.go
@@ -73,13 +73,22 @@ var belowRaftGoldenProtos = map[reflect.Type]fixture{
 	},
 	reflect.TypeOf(&raftpb.HardState{}): {
 		populatedConstructor: func(r *rand.Rand) protoutil.Message {
+			type expectedHardState struct {
+				Term             uint64
+				Vote             uint64
+				Commit           uint64
+				XXX_unrecognized []byte
+			}
+			// Conversion fails if new fields are added to `HardState`, in which case this method
+			// and the expected sums should be updated.
+			var _ = expectedHardState(raftpb.HardState{})
+
 			n := r.Uint64()
-			// NB: this would rot if HardState ever picked up more (relevant)
-			// fields, but that's unlikely.
 			return &raftpb.HardState{
-				Term:   n % 3,
-				Vote:   n % 7,
-				Commit: n % 11,
+				Term:             n % 3,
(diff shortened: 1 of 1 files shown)


tschottdorf:
Thanks @pydima!

I think you can simplify this by just adding a test here:

```go
func TestHardStateKnownFields(t *testing.T) {
  populatedHardState := belowRaftGoldenProtos[reflect.TypeOf(&raftpb.HardState{})].populatedConstructor(zeroSeededRand).(*raftpb.HardState)
  // Use reflection to walk all the values and explode if you find a zero value (except for `XXX_unrecognized`)
  // If one is found, Fatal with a hint to update populatedConstructor for *HardState
}
```

petermattis:
If the goal is to detect when there are additional fields added to `raftpb.HardState`, another approach would be to make a parallel structure that is convertible:

```
type expectedHardState struct {
	Term             uint64
	Vote             uint64
	Commit           uint64
	XXX_unrecognized []byte
}

var _ = expectedHardState(raftpb.HardState{})  // This will not compile if fields are added to raftpb.HardState
```

tschottdorf:
That's much easier. Good idea!

pydima:
Thanks, that's really much simpler. I've updated PR.

pkg/storage/below_raft_protos_test.go
```diff
@@ -74,8 +85,6 @@ var belowRaftGoldenProtos = map[reflect.Type]fixture{
 	reflect.TypeOf(&raftpb.HardState{}): {
 		populatedConstructor: func(r *rand.Rand) protoutil.Message {
 			n := r.Uint64()
-			// NB: this would rot if HardState ever picked up more (relevant)
-			// fields, but that's unlikely.
 			return &raftpb.HardState{
```
tschottdorf: Try just adding `XXX_unrecognized: nil,` here and see if that allows you to remove the linter warning.

pydima: yep, you are right, thanks! fixed that.

tschottdorf:
Thanks! I made some small edits to your branch (hope that's ok, just wanted to save you the hassle) and will merge on green.

PR #19041, opened 2017-10-04
sql: add crdb_internal.ranges

madelynnblue:
This is slightly different from SHOW TESTING_RANGES in that it:
is runnable only by root, shows all ranges, doesn't truncate table
IDs from the output, doesn't show lease holders. The bulkio team
has needed (and written) this many times to investigate various
split-related things.

Also sort that silly list.

### DIFF
--- a/pkg/sql/crdb_internal.go
+++ b/pkg/sql/crdb_internal.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/cockroachdb/cockroach/pkg/build"
 	"github.com/cockroachdb/cockroach/pkg/keys"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/security"
 	"github.com/cockroachdb/cockroach/pkg/server/serverpb"
 	"github.com/cockroachdb/cockroach/pkg/settings"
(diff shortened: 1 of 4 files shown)


knz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19041#-:-KvdGoLrIiu_PbVjPJdr:bnfp4nl)*

danhhz:
That's for adding this!

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/crdb_internal.go, line 1337 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19041#-Kvgw3Jo6yMDDzleC4-D:-Kvgw3Jo6yMDDzleC4-E:b-n3scov) ([raw file](https://github.com/cockroachdb/cockroach/blob/b09a4b59bfd76766900ee57c1191d1fc60a112a2/pkg/sql/crdb_internal.go#L1337)):*
> ```Go
> 				parser.NewDString(keys.PrettyPrint(desc.StartKey.AsRawKey())),
> 				parser.NewDString(keys.PrettyPrint(desc.EndKey.AsRawKey())),
> 				parser.NewDInt(parser.DInt(len(desc.Replicas))),
> ```

I would greatly prefer this be which replicas and not how many. Not sure the command really replaces my manual kv scanning of the range descriptors otherwise

optional: I'd also be really nice to have lease holder information. I see the `TODO(radu): this will be slow if we have a lot of ranges; find a way to make this part optional` comment on the `SHOW TESTING_RANGES` impl, but it's useful enough that I think adding it now and speeding it up later is probably better than omitting it. One rpc per range isn't really that bad in a cold path like this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19041#-:-Kvgx58m2lNVWbBzigRe:be2sk2r)*

RaduBerinde:
This table was proposed in an RFC but was never implemented. Here are the details in case you want to conform to that schema: https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20170314_sql_split_syntax.md#4-crdb_internalranges-and-ranges_cached-system-table

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19041#-:-Kvh5eahT_b39PREP1Yk:bhpm8bj)*

RaduBerinde:
*[pkg/sql/crdb_internal.go, line 1337 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19041#-Kvgw3Jo6yMDDzleC4-D:-Kvh6L9eOiWVsAuD7irQ:b-alwuv3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b09a4b59bfd76766900ee57c1191d1fc60a112a2/pkg/sql/crdb_internal.go#L1337)):*

I would greatly prefer this be which replicas and not how many. Not sure the command really replaces my manual kv scanning of the range descriptors otherwise

optional: I'd also be really nice to have lease holder information. I see the `TODO(radu): this will be slow if we have a lot of ranges; find a way to make this part optional` comment on the `SHOW TESTING_RANGES` impl, but it's useful enough that I think adding it now and speeding it up later is probably better than omitting it. One rpc per range isn't really that bad in a cold path like this
</blockquote></details>

I agree that having leaseholder and replicas would be helpful (ideally we would remove `SHOW TESTING_RANGES` altogether once this table is available).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19041)*

PR #19043, opened 2017-10-04
docgen: pick up upsert change from docs repo

dt:
The diagram code import picked up a version before https://github.com/cockroachdb/docs/commit/e8dae27e3e32da75989d77fd9dcec976217e814f.
This applies the same diff there and re-generates the diagrams.


jseldess:
@dt, https://github.com/cockroachdb/docs/pull/1946 (lots of 1.1 diagram fixes) might've gotten missed, too. And there are a handful of open 1.1 PRs with subsequent diagram changes, namely:

- https://github.com/cockroachdb/docs/pull/1954
- https://github.com/cockroachdb/docs/pull/1934
- https://github.com/cockroachdb/docs/pull/1914
- https://github.com/cockroachdb/docs/pull/1872

And there are likely to be more before the release. 

So it's best to wait on making this change official until after the release. That ok?

dt:
@jseldess yep, as those changes merge in docs over the next week or two, I'll keep the code here up-to-date too, so that ideally there are no diffs between `docs/generated` and the files in the docs repo. That'll also give us a QA period to ensure that it is working as expected.

madelynnblue (approved):

PR #19047, opened 2017-10-05
base: remove trailing slash from DocsURLBase

### LINKED ISSUE #19039: cli/sql: duplicate forward slash in generated URLs for contextual help
```
> select * from [TAB]
[...]
See also:
  SHOW TABLES
  https://www.cockroachlabs.com/docs/v1.1//table-expressions.html
```

Note how there is a double slash: `v1.1//table`

benesch:
pkg/sql/parser was not expecting DocsURLBase to have a trailing slash,
resulting in double-slashed URLs like

    https://www.cockroachlabs.com/docs/v1.2//FOO.html

Fixes #19039.

### DIFF
--- a/pkg/base/docs.go
+++ b/pkg/base/docs.go
@@ -15,15 +15,13 @@
 package base
 
 import (
-	"fmt"
-
 	"github.com/cockroachdb/cockroach/pkg/build"
 )
 
 // DocsURLBase is the root URL for the version of the docs associated with this
 // binary.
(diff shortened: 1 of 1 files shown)


knz (approved):

knz:
You'll need to regenerate the docs (I think because some of the builtin function docs contain URL refs)

benesch:
Hmm, none of the docs actually changed. Looks like I just hit the nondeterminism of the diagrams before @dt fixed it.

PR #19048, opened 2017-10-05
docgen: handle missing jar more gracefully

dt:
We originally always fell back to the public webservice if the jar was missing, but that was made stricter to avoid inadvertantly regressing where we intended to use the jar, like CI builds.

Unfortunately that made OSS contributors unable to run the 'go:generate' driven invocation out-of-the-box (a manual invocation would work).
This makes the strict handling optional, so we can enable it in CI to prevent regression, but do the fallback elsewhere.

Also splits the generation targets up to allow more granular re-generation as desired.


pkg/cmd/docgen/diagrams.go
```diff
@@ -158,7 +160,12 @@ var diagramCmd = func() *cobra.Command {
 			if railroadJar != "" {
 				_, err := os.Stat(railroadJar)
 				if err != nil {
-					log.Fatalf("%s not found\n", railroadJar)
+					if envutil.EnvOrDefaultBool("COCKROACH_REQUIRE_RAILROAD", false) {
+						log.Fatalf("%s not found\n", railroadJar)
+					} else {
+						log.Printf("%s not found, falling back to slower web service (employees can find it on google drive).", railroadJar)
```
benesch: optional nit: the "it" in the parenthetical has an awkward antecedent

docs/generated/sql/diagrams/doc.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package diagrams
+
+//go:generate go run ../../../../pkg/cmd/docgen/main.go ../../../../pkg/cmd/docgen/funcs.go ../../../../pkg/cmd/docgen/diagrams.go grammar --base ./diagrams --railroad ../../../../build/Railroad.jar --addr ../../../../pkg/sql/parser/sql.y
```
benesch: We should consider `go install`ing docgen and letting the Makefile keep it up to date. (In a future PR.) Go makes this so damn gross.

benesch (approved):

docs/generated/sql/diagrams/doc.go
dt: yeah, i just didn't want to force re-building it whenever we built tools, since it does depend on e.g. sql/parser, since not everyone actually needs to run it -- only those making sql lang changes do. I guess we make everyone build protoc even if they don't make proto changes though, so eh. I wish we had better just-in-time tool binary mgmt.

mjibson: MORE DOTS

madelynnblue (approved):

pkg/cmd/docgen/diagrams.go
dt: Done.

dt:
huh. the output is not deterministic!

I think it is the `replace` maps (we iterate them), when the order of replacements matters (one contains other).

dt:
okay, well ideally we would have non-overlapping replacements, or, short of that, an ordered-slice of replacements, but sorting them for now should get determinism / cut flakiness while we look at a real fix.

dt:
merging this on the prior LGTMs, even though i added a sort since then -- more than happy to revise further in a follow up, but just want to kill the non-determinism before it starts flaking people's builds in the morning.

Issue #17172, opened 2017-07-23
perf: explore pushing parts of MVCC or KV evaluation into C++


petermattis:
We've observed in the past that minimizing Go/C++ crossings in `storage/engine/mvcc.go` is a useful performance optimization. But many of the interesting methods in `mvcc.go` still perform multiple Cgo calls. And stepping up a level, evaluation of a KV operation can often perform multiple MVCC operations.

The challenge in pushing more logic into C++ is in the Go/C++ boundary. Currently, Go code manually fills in C (not C++) structures and passes those C structures to C routines. When data is returned from C++ the Go code manually copies it from the C structs. So far, we've avoided passing any complicated structure and thus the C++ interface is very basic, knowing only about strings and timestamps.

Cgo does not provide a facility for passing Go structures to C. I think we can provide that facility ourselves. Specifically, we can use `reflect` to auto-generate C++ code that can access a Go data structure. Using `reflect` we can find the types and byte offsets of any field for a given Go type. Auto-generating C++ code from these byte offsets is then relatively straightforward.

Note that passing an arbitrary Go structure to C++ would violate the rules in https://golang.org/cmd/cgo/#hdr-Passing_pointers. Those rules are somewhat stricter than necessary and we'd definitely be playing on the edge here. The auto-generated code could only be used for reading the Go structure, not for writing as we'd definitely not want to deal with supporting the GC write barrier. This implies that returning a structure from C++ to Go would be more limited. A final limitation is that this approach would not allow C++ to iterate over a Go map, but so far none of the protos in `roachpb` use Go maps. 

The proposed steps of exploration are:
1. Write a tool to auto-generate C++ for accessing a Go data structure.
2. Convert one of the MVCC routines (e.g. `MVCCGet`) to C++ and benchmark the performance improvement.
3. Evaluate where it makes sense to push Go code into C++ (the decision might be no change from the status quo).

bdarnell:
Go has two ways to interface with foreign code: cgo and assembly. The assembly interface doesn't interact with the scheduler, so it's much faster (but it has limitations such as not being suitable for code that may make system calls). There are tricks like https://github.com/minio/c2goasm for calling C code via the assembly interface; we may be able to do something like this to reduce the cost of crossing the boundary for frequently-called functions.

petermattis:
[...]

PR #19049, opened 2017-10-05
storage: avoid cacheRequest allocations for each request

nvb:
We used to allocate a new `cacheRequest` and usually the `read` `[]Span`
that it contained for each request. These allocations happened often enough
that I was seeing them show up on TPCC heap profiles. We now batch the
`cacheRequest` allocation in with `endCmds`, which saves at least one
allocation per request.

I wasn't able to see any benchmarks movement because of this change, but
the allocations are no longer showing up in the profiles and it seems
like a clear win.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -1892,6 +1892,10 @@ type endCmds struct {
 	repl *Replica
 	cmds batchCmdSet
 	ba   roachpb.BatchRequest
+
+	// The following fields are used to batch ts cache allocations.
+	cr    cacheRequest
+	spans [1]roachpb.Span
 }
 
 // done removes pending commands from the command queue and updates
@@ -1908,8 +1912,8 @@ func (ec *endCmds) done(br *roachpb.BatchResponse, pErr *roachpb.Error, retry pr
 			// evaluating the request.
 			log.Fatal(context.Background(), err)
 		}
-		creq := makeCacheRequest(&ec.ba, br, span)
 
+		creq := ec.makeCacheRequest(br, span)
 		if ec.repl.store.Clock().MaxOffset() == timeutil.ClocklessMaxOffset {
 			// Clockless mode: all reads count as writes.
 			creq.writes, creq.reads = append(creq.writes, creq.reads...), nil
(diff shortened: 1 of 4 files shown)


petermattis:

---

*[pkg/storage/replica.go, line 1897 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19049#-KvgkZ4X6zpD7QP0GA7_:-KvgkZ4X6zpD7QP0GA7a:b-cpla3s) ([raw file](https://github.com/cockroachdb/cockroach/blob/4a56bc788257f5cc78ce9760e63887ac620bacf1/pkg/storage/replica.go#L1897)):*
> ```Go
> 
> 	// The following fields are used to batch ts cache allocations.
> 	cr    cacheRequest
> ```

This has the downside of pinning the other `endCmds` memory (which is already 144 bytes) to the lifetime of `cacheRequest`. At the very least, this will throw off the memory accounting in the timestamp cache. Since this isn't showing any performance improvement, I'm thumbs down on this change as I'd rather explore #17172 and move the timestamp cache into C++. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19049)*

nvanbenschoten:

---

*[pkg/storage/replica.go, line 1897 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19049#-KvgkZ4X6zpD7QP0GA7_:-KvgnNPUIT1_hzWFo0nl:b-y6dv0m) ([raw file](https://github.com/cockroachdb/cockroach/blob/4a56bc788257f5cc78ce9760e63887ac620bacf1/pkg/storage/replica.go#L1897)):*

This has the downside of pinning the other `endCmds` memory (which is already 144 bytes) to the lifetime of `cacheRequest`. At the very least, this will throw off the memory accounting in the timestamp cache. Since this isn't showing any performance improvement, I'm thumbs down on this change as I'd rather explore #17172 and move the timestamp cache into C++. 
</blockquote></details>

Yeah, that's a good point. My first run at this just threw `cacheRequests` in a `sync.Pool`, since we have such easy access to their lifetimes. Seems like that would have similar benefits to this without the downside you mentioned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19049)*

nvanbenschoten:
This work was subsumed by #19418.

PR #19050, opened 2017-10-05
[experiment] bufalloc: create custom allocator for []roachpb.KeyValue

nvb:
The [`[]roachpb.KeyValue` in `mvccScanInternal`](https://github.com/cockroachdb/cockroach/blob/16cb6dc42c23c0d8e73401217dd13c51874901ef/pkg/storage/engine/mvcc.go#L1568) was showing up as the
second largest culprit of allocations when running TPCC, just behind
[`bufalloc.ByteAllocator.Copy` in `MVCCIterate`](https://github.com/cockroachdb/cockroach/blob/16cb6dc42c23c0d8e73401217dd13c51874901ef/pkg/storage/engine/mvcc.go#L1708). This makes sense because
this slice will collect all KV pairs for `ScanRequests` and carry them
back to the client.

I figured that we could improve this by pooling slices, and that led to
the current design of pooling different capacity slices and managing
slice appends.

Over a series of 5 one-minute TPCC runs from before and after the change,
this approach has shown to decrease the `ns/op` metric from an average
of `73638065.2 ns/op` to an average of `63700720.2 ns/op`, a speedup of
**13.5%**.

The code is pretty rough here, but it's an interesting starting point
that demonstrates a clear target for optimization.

### DIFF
--- a/pkg/internal/client/batch.go
+++ b/pkg/internal/client/batch.go
@@ -18,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
+	"github.com/cockroachdb/cockroach/pkg/util/bufalloc"
 )
 
 const (
@@ -209,6 +210,7 @@ func (b *Batch) fillResults() error {
 						dst.Key = src.Key
 						dst.Value = &src.Value
 					}
+					bufalloc.ReleaseKVSlice(t.Rows)
 				}
 			case *roachpb.ReverseScanRequest:
 				if result.Err == nil {
@@ -220,6 +222,7 @@ func (b *Batch) fillResults() error {
 						dst.Key = src.Key
 						dst.Value = &src.Value
 					}
+					bufalloc.ReleaseKVSlice(t.Rows)
 				}
 			case *roachpb.DeleteRequest:
 				row := &result.Rows[k]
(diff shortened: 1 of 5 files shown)


petermattis:
Interesting. I'm surprised it had this large an impact. https://github.com/cockroachdb/cockroach/pull/18913 might have a larger effect. I never tested that with TPC-C.

jordanlewis:
@nvanbenschoten it would be useful to see which queries were sped up most dramatically by this change. One way to do that is to pass `-ops-stats` to `tpcc` and compare before and after to see which operation got the biggest improvement. I assume it's one of the aggregates but I'd like to know for sure.

Another thing you can do is specify a particular transaction and run only that one - with `-mix=delivery=1` for instance to get only delivery transactions.

nvanbenschoten:
@petermattis, they aren't mutually exclusive, are they? Either way, since I have TPCC setup I'll take a look at how much #18913 moves the needle.

@jordanlewis yeah I'd expect it to be whichever queries fetch the largest set of KVs. I'll take a look.

petermattis:
> @petermattis, they aren't mutually exclusive, are they? Either way, since I have TPCC setup I'll take a look at how much #18913 moves the needle.

#18913 does away with the `roachpb.KeyValue` in the response which would make this change unnecessary.

nvanbenschoten:
Oh right, I didn't realize that we wouldn't even call `MVCCScan` for `ScanRequests`.

PR #19052, opened 2017-10-05
storageccl: remove index id during rekey of primary index

### LINKED ISSUE #16317: backup/restore: extra empty range created on restore

madelynnblue:
During backup, spans are created based on their index, so the first
span for a table would be something like /Table/52/1. During restore,
we would restore these spans as they are during backup, resulting in a
span from /Table/51/<whatever> - /Table/52/1. The next time the split
queue is run it would ensure a table is split at the correct place
(/Table/52), leaving a forever-empty range at /Table/52 - /Table/52/1.

To fix this, teach the rekey method to recognize primary index start
spans and have it move those to the start of the table span.

Fixes #16317

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -343,6 +343,24 @@ func backupAndRestore(
 		if rowCount != int64(numAccounts) {
 			t.Fatalf("expected %d rows but found %d", numAccounts, rowCount)
 		}
+
+		// Verify there's no /Table/51 - /Table/51/1 empty span.
+		{
+			var count int
+			sqlDBRestore.QueryRow(`
+			SELECT count(*) FROM crdb_internal.ranges
+			WHERE start_key = (
+				('/Table/' ||
+				(SELECT table_id FROM crdb_internal.tables
+					WHERE database_name = $1 AND name = $2
+				)::STRING) ||
+				'/1'
+			)
+		`, "data", "bank").Scan(&count)
+			if count != 0 {
+				t.Fatal("unexpected span start at primary index")
+			}
+		}
 	}
 }
 func TestBackupRestoreSystemJobs(t *testing.T) {
--- a/pkg/ccl/storageccl/key_rewriter.go
+++ b/pkg/ccl/storageccl/key_rewriter.go
@@ -11,6 +11,7 @@ package storageccl
 import (
 	"bytes"
 
+	"github.com/cockroachdb/cockroach/pkg/keys"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
 	"github.com/cockroachdb/cockroach/pkg/util/encoding"
@@ -191,7 +192,9 @@ func (kr *KeyRewriter) RewriteKey(key []byte) ([]byte, bool, error) {
 }
 
 // RewriteSpan returns a new span with both Key and EndKey rewritten using
-// RewriteKey. An error is returned if either was not matched for rewrite.
+// RewriteKey. Span start keys for the primary index will be rewritten to
+// contain just the table ID. That is, /Table/51/1 -> /Table/51. An error
+// is returned if either was not matched for rewrite.
 func (kr *KeyRewriter) RewriteSpan(span roachpb.Span) (roachpb.Span, error) {
 	newKey, ok, err := kr.RewriteKey(append([]byte(nil), span.Key...))
 	if err != nil {
@@ -199,6 +202,15 @@ func (kr *KeyRewriter) RewriteSpan(span roachpb.Span) (roachpb.Span, error) {
 	}
 	if !ok {
 		return roachpb.Span{}, errors.Errorf("could not rewrite key: %s", span.Key)
+	}
+	// Modify all spans that begin at the primary index to instead begin at the
+	// start of the table. That is, change a span start key from /Table/51/1 to
+	// /Table/51. Otherwise a permanently empty span at /Table/51-/Table/51/1
+	// will be created.
+	if b, id, idx, err := sqlbase.DecodeTableIDIndexID(newKey); err != nil {
+		return roachpb.Span{}, errors.Wrapf(err, "could not rewrite key: %s", span.Key)
+	} else if idx == 1 && len(b) == 0 {
+		newKey = keys.MakeTablePrefix(uint32(id))
 	}
 	newEndKey, ok, err := kr.RewriteKey(append([]byte(nil), span.EndKey...))
 	if err != nil {
--- a/pkg/ccl/storageccl/key_rewriter_test.go
+++ b/pkg/ccl/storageccl/key_rewriter_test.go
@@ -12,6 +12,7 @@ import (
 	"bytes"
 	"testing"
 
+	"github.com/cockroachdb/cockroach/pkg/keys"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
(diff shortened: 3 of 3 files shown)


danhhz:
Please add some more detail to the commit message, including what the issue was, how it was happening, and why this fixes it

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/storageccl/key_rewriter.go, line 207 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvDG5SGN5fWAu1XxY:-KvgvDG5SGN5fWAu1XxZ:b-ttmqpz) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter.go#L207)):*
> ```Go
> 	}
> 	if b, id, idx, err := sqlbase.DecodeTableIDIndexID(newKey); err != nil {
> 		return roachpb.Span{}, errors.Wrapf(err, "could not rewrite key: %s", span.Key)
> ```

this probably wants a distinct error message so we can figure out which branch was taken if some customer hits this

---

*[pkg/ccl/storageccl/key_rewriter.go, line 211 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvPfGQ6ARVZM8ojYr:-KvgvPfGQ6ARVZM8ojYs:bfh049n) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter.go#L211)):*
> ```Go
> 		newKey = keys.MakeTablePrefix(uint32(id))
> 	}
> 	// Modify all spans that begin at the primary index to instead begin at the
> ```

This comments seems oddly placed. Move above the `sqlbase.DecodeTableIDIndexID` block?

---

*[pkg/ccl/storageccl/key_rewriter_test.go, line 147 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvY6OCB-eCICpnR9c:-KvgvY6OCB-eCICpnR9d:b-vvqidy) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter_test.go#L147)):*
> ```Go
> 	})
> 
> 	t.Run("span", func(t *testing.T) {
> ```

seems like we want a RESTORE level test, too? it's probably too easy for this to regress otherwise

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-:-KvgvcweYKAxJlIJBlAW:b-rw49wb)*

mjibson:
Added commit log details.

---

---

*[pkg/ccl/storageccl/key_rewriter.go, line 207 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvDG5SGN5fWAu1XxY:-KvhfJutbX11XWUomk-j:bhr37ti) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter.go#L207)):*

this probably wants a distinct error message so we can figure out which branch was taken if some customer hits this
</blockquote></details>

We already duplicate the error message 4 other times. This one wraps the error from DecodeTableIDIndexID, which will be unique enough to tell it's from this line.

---

*[pkg/ccl/storageccl/key_rewriter.go, line 211 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvPfGQ6ARVZM8ojYr:-KvhfXfmKr-eseV_HlG-:b-a4wcvi) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter.go#L211)):*

This comments seems oddly placed. Move above the `sqlbase.DecodeTableIDIndexID` block?
</blockquote></details>

Oops. Done.

---

*[pkg/ccl/storageccl/key_rewriter_test.go, line 147 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvY6OCB-eCICpnR9c:-KvhnRIiOjHaWxKUKH79:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter_test.go#L147)):*

seems like we want a RESTORE level test, too? it's probably too easy for this to regress otherwise
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-:-KvhrDkXWmA-s-tKNukN:b-bnqxuc)*

danhhz:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup_test.go, line 350 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-Kvi6FtfoOPDh3fkDww7:-Kvi6FtgHDspCcjv0o8E:b71ogfd) ([raw file](https://github.com/cockroachdb/cockroach/blob/1962e344cf6dd7480184de7fd24e888bb3491687/pkg/ccl/sqlccl/backup_test.go#L350)):*
> ```Go
> 		{
> 			var count int
> 			sqlDBRestore.QueryRow(`
> ```

it might be more maintainable to count the number of ranges that are part of the table span and assert that they're equal on the original and restored table

---

*[pkg/ccl/storageccl/key_rewriter.go, line 207 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-KvgvDG5SGN5fWAu1XxY:-Kvi6clFXfWps0FNPTxQ:bb7m07t) ([raw file](https://github.com/cockroachdb/cockroach/blob/6ab57ec7d1ca9d85559c9132a0ad062fd3fa9aed/pkg/ccl/storageccl/key_rewriter.go#L207)):*

We already duplicate the error message 4 other times. This one wraps the error from DecodeTableIDIndexID, which will be unique enough to tell it's from this line.
</blockquote></details>

okay

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-:-Kvi6fofCCAvO2vSbgKL:bnfp4nl)*

mjibson:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup_test.go, line 350 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052#-Kvi6FtfoOPDh3fkDww7:-KviCIUe5ZzJQ5x1B9AB:bqa7dxy) ([raw file](https://github.com/cockroachdb/cockroach/blob/1962e344cf6dd7480184de7fd24e888bb3491687/pkg/ccl/sqlccl/backup_test.go#L350)):*

it might be more maintainable to count the number of ranges that are part of the table span and assert that they're equal on the original and restored table
</blockquote></details>

I don't think that would have caught this bug because the split queue doesn't run fast enough to make the new split which would trigger the test failure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19052)*

PR #19060, opened 2017-10-05
base,server,distsqlrun: specify a Monitor instead of a capacity limit for temp storage

andreimatei:
Before this patch, "temp storage" was limited to a fixed max capacity
through a config options expressed as a limit in bytes. Eventually, the
DistSQL server created a BytesMonitor that will enforce that limit.
This patch replaces, from the perspective of the various specs that come
into play in creating temp storage, the limit with a BytesMonitor that's
under the control of whoever is creating the Server. So, it's no longer
up to the DistSQL server to create the monitor; it is passed in from up
above.
This patch does not change any behavior; the monitors that are passed in
for temp storage have the same limit as before and model exclusive use
of "disk space/memory capacity out of thin air". However, the door is
now open to high-level architecting of resource usage; the resources
used for temp storage can, in principle, be integrated into general,
higher-level resource limits/accounting in crdb (e.g. the disk space
being used can be integrated with other users of the same storage
device competing for capacity).


andreimatei:
The first commit is #18544; ignore here.

RaduBerinde:
LGTM

PR #19062, opened 2017-10-05
rfc: JSONB encoding

justinj:


bdarnell:
:lgtm:

---

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhhRjvijUUX4usHcAy:-KvhhRjvijUUX4usHcAz:b-o0bntk) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L79)):*
> ```Markdown
> experiments will be needed.
> 
> The root JSON object is always a container, but it might be a `scalar`
> ```

Add a comment that scalar containers are defined below. I was confused by this initially. (maybe move the container header definition up here instead of in the `array` section)

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhjJoSm9FYHZsadmJB:-KvhjJoSm9FYHZsadmJC:brqe4p) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L99)):*
> ```Markdown
> differs from JavaScript's `number` type as Postgres JSONB `number`s are
> arbitrary-precision.
> In CockroachDB, `number` will be encoded using CockroachDB's `DECIMAL` encoding.
> ```

Why DECIMAL instead of FLOAT? JSON officially only supports double precision, although many implementations allow int64 for integer values. Does postgres do this? Do we?

Would it be worthwhile to switch between different numeric sub-types depending on the value? 

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 219 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhkTSYQzA5v618y4iC:-KvhkTSYQzA5v618y4iD:b-ifpzzx) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L219)):*
> ```Markdown
> # Unresolved questions
> 
> - What should be done about `$STRIDE`, should it be used as written here, or
> ```

What value does postgres use?

I would keep the length-vs-offset flag bit, but for the initial implementation only use lengths (effectively an infinite stride). The kind of random access to an array where the offset makes a difference sounds rare to me, but postgres may have a good reason for it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-:-Kvhn4ERPjB4a1RoSSJM:bnfp4nl)*

justinj:

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhhRjvijUUX4usHcAy:-KvhwwrTCdN3x87uIgUL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L79)):*

Add a comment that scalar containers are defined below. I was confused by this initially. (maybe move the container header definition up here instead of in the `array` section)
</blockquote></details>

Done.

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhjJoSm9FYHZsadmJB:-KvhsiIERvwHXKxCJ1WP:bf03tmu) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L99)):*

Why DECIMAL instead of FLOAT? JSON officially only supports double precision, although many implementations allow int64 for integer values. Does postgres do this? Do we?

Would it be worthwhile to switch between different numeric sub-types depending on the value? 
</blockquote></details>

Postgres uses `NUMERIC` for the numbers in their JSON objects, rather than floats (which indeed is a difference from official JSON).

I guess this is a question of whether we want to be more JSON-compliant or more Postgres-compliant, personally I feel like it makes more sense to do what Postgres does here, given that we already have a more-or-less `NUMERIC`-compliant encoding in `DECIMAL`.

Using a different encoding based on the value seems like a good thing to test, we have a couple extra type flags that could possibly be used for that purpose. I added a note to experiment with it. I'm not sure how substantial the space or speed difference for a `DECIMAL` encoding is compared to an int64.

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 219 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhkTSYQzA5v618y4iC:-KvhuI-cuZlUHJUWIz_Q:b1xh83z) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L219)):*

What value does postgres use?

I would keep the length-vs-offset flag bit, but for the initial implementation only use lengths (effectively an infinite stride). The kind of random access to an array where the offset makes a difference sounds rare to me, but postgres may have a good reason for it.
</blockquote></details>

[They use 32](https://github.com/postgres/postgres/blob/90627cf98a8e7d0531789391fd798c9bfcc3bc1a/src/include/utils/jsonb.h#L181). It's not clear where they got that number from. @jordanlewis suggested it being so large might imply that they believe there to be some kind of power law distribution to the size of JSON objects (mostly small, with a few really large ones), which seems plausible to me.

That sounds reasonable to me. It will be possible to get more data once this is functional, and if we keep the flag we can change this if we find offsets to be beneficial.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062)*

petermattis:
:lgtm:

---

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 92 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-Kvhrb2Otr2urTAjVBFA:-Kvhrb2Otr2urTAjVBFB:b-91aetq) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L92)):*
> ```Markdown
> ### `string`
> 
> `string`s are encoded as normal UTF-8 strings.
> ```

Is this using `encoding.EncodeBytesValue`? Similarly, for `true`, `false`, `null`, are they using `encoding.Encode{Null,Bool}Value`? 

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhjJoSm9FYHZsadmJB:-KvhrBBQKdl40JqvFMP6:b94h1ba) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L99)):*

Why DECIMAL instead of FLOAT? JSON officially only supports double precision, although many implementations allow int64 for integer values. Does postgres do this? Do we?

Would it be worthwhile to switch between different numeric sub-types depending on the value? 
</blockquote></details>

Also curious about this decision. I believe DECIMAL is a lot slower than FLOAT for encoding/decoding and definitely for manipulation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-:-KvhsRRudH2moatDRwCf:bg7igk9)*

justinj:

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 92 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-Kvhrb2Otr2urTAjVBFA:-KvhzOMOHzcF3XJIXPWX:b-gxbv9y) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L92)):*

Is this using `encoding.EncodeBytesValue`? Similarly, for `true`, `false`, `null`, are they using `encoding.Encode{Null,Bool}Value`? 
</blockquote></details>

It could, it doesn't need to though, since we already have the length and type information from the JEntry header. There's no need to use the `true`, `false`, and `null` encodings since they have no encoding at all in JSON besides their JEntry, whose format is different from our existing value tags.

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhjJoSm9FYHZsadmJB:-Kvi-EaJU9sko6pZKj9j:bseq4vn) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L99)):*

Also curious about this decision. I believe DECIMAL is a lot slower than FLOAT for encoding/decoding and definitely for manipulation.
</blockquote></details>

It does seem like the type flags could give us a way to avoid using `DECIMAL`s when we can get away with an int or a float (because we have to switch on them regardless), so I will definitely experiment with doing that. cc @mjibson who might also have insights on this topic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062)*

mjibson:

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhjJoSm9FYHZsadmJB:-KviILSPEhU_ePGlZtaY:btngnxx) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L99)):*

It does seem like the type flags could give us a way to avoid using `DECIMAL`s when we can get away with an int or a float (because we have to switch on them regardless), so I will definitely experiment with doing that. cc @mjibson who might also have insights on this topic.
</blockquote></details>

I don't have much to add. I guess just balancing what the spec is with what postgres does and what users expect? I think people doing operations on JSON numbers should know they are floats and so are limited like floats.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062)*

knz:
LGTM too, although for the sake of completeness you could enhance this with an example encoding of a few example json values.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-:-KviOhLubWayMURl4kBk:bxsx3t8)*

mjibson:

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvhjJoSm9FYHZsadmJB:-KvjbNBe__s0efYm9cBi:bgsh7g5) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L99)):*

We definitely can't *always* use floats if we want Postgres compatibility. If numbers were floats, the following wouldn't work:

```
jordan=# insert into j values (1, '{"a": 2.34400000}');
jordan=# select * from j;
 a |         b
---+-------------------
 1 | {"a": 2.34400000}
```
</blockquote></details>

Then looks like we need decimals.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062)*

justinj:

---

*[docs/RFCS/20171005_jsonb_encoding.md, line 81 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-KvikYaMPOPgJtlFbCaF:-KvmqtmVcYfZO6tm2GPW:b-zdjb9w) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4b0bdc490855c58c572ec41e9c3802749c3c5e2/docs/RFCS/20171005_jsonb_encoding.md#L81)):*

This doesn't seem strictly necessary to me - couldn't you add a version field later to the ColumnDescriptor/ColumnType proto? 
</blockquote></details>

My first thought was that then you couldn't have multiple versions coexisting in a column, but upon reflection, if the format is dramatically different (and thus will have very different performance characteristics) that doesn't sound like something you would want anyway, so agreed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062)*

a-robinson (approved):
👍

m-schneider:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19062#-:-Kw186JSinzMP3nUJ-1F:bnfp4nl)*

justinj:
@knz I added a couple examples, good idea.

Also, this RFC is going to enter the final comment period.

PR #19063, opened 2017-10-05
ui: Convert Metrics Query Thunk -> Redux Saga

mrtracy:
Converts the redux thunk used to run metrics queries (for our time
series graphs) into a redux saga. Sagas are an alternative framework for
asynchronous actions which use generators instead of promises.

Metrics query was chosen as the first thunk to convert because it has
the most complex behavior (the transparent batching of requests from
different graphs). Other reducers which use metric queries still use
redux thunk.  In the interest of keeping this PR focused on the saga
conversion, I have not modified the reducer for metrics, or any other
component, in any significant way.

Advantages of this move:

+ Thunk has been broken up into three different generator functions
which call each other. Importantly, because these are saga generators,
the higher level functions can be tested in isolation without having to
invoke the lower-level functions; this allows us to put a strong focus
on each part of the process without needing complicated mocks or
orchestration in the tests.

In my opinion, the readability and focus of the new tests is enough to
justify the switch to saga on its own.

+ Sagas are executed as a long-running process loop which listens for
dispatched actions; this is in contrast to thunk, which dispatches
functions as actions. This allows us to have a significant amount of
control over parallel invocations while using locally-scoped variables
(as an alternative to package globals or redux store state).

This has a minor benefit for this particular method, but I anticipate
greater benefits for other sagas as they are written.

New Dependencies Required:
+ Redux Saga
+ Babel Polyfill

Polyfill has been hooked up in karma.conf.js and webpack.config.js. The
new dependencies are not a major contributor to package size.

### DIFF
--- a/pkg/ui/karma.conf.js
+++ b/pkg/ui/karma.conf.js
@@ -26,6 +26,7 @@ module.exports = function(config) {
 
     // list of files / patterns to load in the browser
     files: [
+      "src/polyfills.ts",
       "src/**/*.spec.*",
     ],
 
@@ -48,7 +49,7 @@ module.exports = function(config) {
     // preprocess matching files before serving them to the browser
     // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
     preprocessors: {
-      "src/**/*.spec.*": ["webpack", "sourcemap"],
+      "src/**": ["webpack", "sourcemap"],
     },
 
     // test results reporter to use
--- a/pkg/ui/package.json
+++ b/pkg/ui/package.json
@@ -26,7 +26,9 @@
     "react-select": "^1.0.0-rc.4",
     "react-sticky": "^5.0.5",
     "redux": "^3.6.0",
+    "redux-saga": "^0.15.6",
     "redux-thunk": "^2.2.0",
+    "regenerator-runtime": "^0.11.0",
     "reselect": "^3.0.0",
     "webcola": "3.3.5",
     "whatwg-fetch": "^2.0.3"
--- a/pkg/ui/src/index.tsx
+++ b/pkg/ui/src/index.tsx
@@ -7,6 +7,8 @@ import "react-select/dist/react-select.css";
 import "styl/app.styl";
 import "src/js/sim/style.css";
 
+import "src/polyfills";
+
 import * as protobuf from "protobufjs/minimal";
 import Long from "long";
 
--- a/pkg/ui/src/polyfills.ts
+++ b/pkg/ui/src/polyfills.ts
@@ -1,0 +1,10 @@
+// This file imports polyfills necessary for both the unit tests and the
+// deployed bundle.
+//
+// If you're tempted to instead import babel-polyfill, which bundles the most
+// common polyfills, fair warning: babel-polyfill, jsdom, and Node 6 are
+// mutually incompatible.
+//
+// TODO(benesch): reevaluate babel-polyfill when we upgrade to Node 8.
+
+import "regenerator-runtime/runtime";
--- a/pkg/ui/src/redux/metrics.spec.ts
+++ b/pkg/ui/src/redux/metrics.spec.ts
@@ -1,11 +1,12 @@
 import { assert } from "chai";
 import _ from "lodash";
 import Long from "long";
-import { Action } from "redux";
-import fetchMock from "src/util/fetch-mock";
 
+import { delay } from "redux-saga";
+import { AllEffect, call, ForkEffect, put, take } from "redux-saga/effects";
+import { queryTimeSeries } from "src/util/api";
 import * as protos from "src/js/protos";
-import * as api from "src/util/api";
+
 import * as metrics from "./metrics";
 
 type TSRequest = protos.cockroach.ts.tspb.TimeSeriesQueryRequest;
@@ -147,11 +148,13 @@ describe("metrics reducer", function() {
     });
   });
 
-  describe("queryMetrics asynchronous action", function() {
+  describe("saga functions", function() {
     type timespan = [Long, Long];
+    const shortTimespan: timespan = [Long.fromNumber(400), Long.fromNumber(500)];
+    const longTimespan: timespan = [Long.fromNumber(0), Long.fromNumber(500)];
 
     // Helper function to generate metrics request.
-    const createRequest = function(ts: timespan, ...names: string[]): TSRequest {
+    function createRequest(ts: timespan, ...names: string[]): TSRequest {
       return new protos.cockroach.ts.tspb.TimeSeriesQueryRequest({
         start_nanos: ts[0],
         end_nanos: ts[1],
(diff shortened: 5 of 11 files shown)


couchand:
:lgtm: Wow this looks really good.  I agree that the tests are much cleaner now.  I'd still advocate for moving the request queue into the store to allow that last inner closure to be ripped out, but I know you don't agree on that so I won't push it.  Too bad about polyfill, particularly since I'm not a fan of generators in the first place, but it's not a big deal.  Thanks for taking this on, looking forward to seeing the rest of it.

---

Review status: 0 of 10 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 170 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw66uDNE-R3UeLbIzHL:-Kw66uDNE-R3UeLbIzHM:bginqex) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L170)):*
> ```XML
>     }
> 
>     function createResponse(...queries: protos.cockroach.ts.tspb.TimeSeriesQueryResponse.Result$Properties[]) {
> ```

Looks like the spread isn't really needed here, every caller already has an array.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 184 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw63si2zouybHyUuspm:-Kw63si3iTI9HQae2qbb:b3i0imh) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L184)):*
> ```XML
>       it("initially waits for incoming request objects", function () {
>         const saga = metrics.queryMetricsSaga();
>         assert.deepEqual(saga.next().value, take(metrics.REQUEST));
> ```

This is sort of a weird assertion, on the configuration rather than the effect, but I guess it works...

---

*[pkg/ui/src/redux/metrics.spec.ts, line 197 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw648Ge-J2ijZx8OswQ:-Kw648Ge-J2ijZx8OswR:b604o1c) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L197)):*
> ```XML
>         const sendBatchesFork = saga.next().value as ForkEffect;
>         assert.isDefined(sendBatchesFork.FORK);
>         assert.deepEqual(saga.next().value, take(metrics.REQUEST));
> ```

It's just a bit unfortunate that this test has to be so coupled to the precise arrangement of yields, but I guess that's part of the saga's interface?  As you know, I'm not all that familiar with sagas, so I don't know if this is common practice.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 247 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw65vQ4ctmc3TeF1p8P:-Kw65vQ4ctmc3TeF1p8Q:bs49b3p) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L247)):*
> ```XML
>         assert.isArray(allEffect.ALL);
>         assert.deepEqual(allEffect.ALL, [
>           call(metrics.sendRequestsAsBatch, [requests[0], requests[2], requests[3]]),
> ```

It might be nice to name these various requests so the assertion isn't so magical.

---

*[pkg/ui/src/redux/metrics.ts, line 252 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw64y1OQITXR29FieI-:-Kw64y1OQITXR29FieI0:bhqwzrw) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L252)):*
> ```XML
>   let requests: WithID<TSRequest>[] = [];
> 
>   while (true) {
> ```

Maybe wrap in a `takeAll` instead of manually `while(true)`ing?

---

*[pkg/ui/src/redux/metrics.ts, line 296 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw67N-T0W3NPQtQ6Vf3:-Kw67N-T0W3NPQtQ6Vf4:bxmbat) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L296)):*
> ```XML
>  * sendRequestsAsBatch sends the supplied requests in a single batch.
>  */
> export function* sendRequestsAsBatch(requests: WithID<TSRequest>[]) {
> ```

Can the names of these functions be made more distinct?

---

*[pkg/ui/src/redux/metrics.ts, line 302 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw67XOj93cc7fbakmar:-Kw67XOj93cc7fbakmas:b8ryts) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L302)):*
> ```XML
> 
>   let response: protos.cockroach.ts.tspb.TimeSeriesQueryResponse;
>   try {
> ```

Wow it's hurting my brain trying to figure out how a plain `try`/`catch` works here.  I guess this is just generator magic, but it seems really counter intuitive to me.  I'm gonna need to go look at the transpiled output to figure out what it ends up as.

---

*[pkg/ui/src/redux/state.ts, line 33 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw68--IYj9Ny1BYD5MD:-Kw68--IYj9Ny1BYD5ME:bunsr1y) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/state.ts#L33)):*
> ```XML
>   const sagaMiddleware = createSagaMiddleware();
> 
>   const store = createStore(
> ```

It might be nice to split up the rest configuration a bit, maybe not in this PR if you'd like to keep it focused.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-:-Kw68vsXSPL0W-Ds5y0-:b-fyeszo)*

couchand:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063)*

mrtracy:

---

*[pkg/ui/src/redux/metrics.spec.ts, line 170 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw66uDNE-R3UeLbIzHL:-KwkCv-eKQGa9Oev0pjh:b-207vx6) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L170)):*

Looks like the spread isn't really needed here, every caller already has an array.
</blockquote></details>

Good point, removed the spread.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 184 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw63si2zouybHyUuspm:-KwkCHRsPn6gXix2rqRY:brlspes) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L184)):*

This is sort of a weird assertion, on the configuration rather than the effect, but I guess it works...
</blockquote></details>

`yield take(metrics.REQUEST)` is the effect; the saga we are testing is only responsible for asking for the next action.  Redux Saga will accept this and eventually yield back the next matching action.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 197 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw648Ge-J2ijZx8OswQ:-KwkIvHAJt1cd06AscGn:b-q1zjix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L197)):*

It's just a bit unfortunate that this test has to be so coupled to the precise arrangement of yields, but I guess that's part of the saga's interface?  As you know, I'm not all that familiar with sagas, so I don't know if this is common practice.
</blockquote></details>

This is very precisely validating the control flow of the saga; we're verifying what directives the saga is giving to the redux saga system, given a certain pattern of input.

To decouple it from the precise behavior, you could write some sort of "mock-saga-runner" which would execute the output of the saga under test more automatically.  I think the output of the mock runner would be 1.) A Store and 2.) A list of pending operations which need additional test input.

This would allow us to obviate the need for precise validation of certain sequences (i.e. sequences of puts), and it would probably provide a better interface for analyzing pending requests (e.g. forks and calls). 

In other words, its likely a good idea - but I think it's beyond the scope of this PR, because that mock object is a little complex.  I think we can accept the coupling for now; it's still superior to the previous configuration, since we get much clearer insight into the high-level functioning of these actions that we did with thunks.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 247 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw65vQ4ctmc3TeF1p8P:-KwkMqW_tPfwoQ9Q2iBu:b-5her6t) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L247)):*

It might be nice to name these various requests so the assertion isn't so magical.
</blockquote></details>

I've tried to make it a little more readable; the the test data is now initially sorted into `shortRequests` and `longRequests`, which are combined into a `mixedRequests` using a flatmap function.  No more numeric accesses.

---

*[pkg/ui/src/redux/metrics.ts, line 252 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw64y1OQITXR29FieI-:-KwkF8GPAlquQfaQmYtD:b-gignf5) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L252)):*

Maybe wrap in a `takeAll` instead of manually `while(true)`ing?
</blockquote></details>

`takeAll` creates a forked process for every message, but we are exercising greater control over the forking process.

---

*[pkg/ui/src/redux/metrics.ts, line 296 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw67N-T0W3NPQtQ6Vf3:-KwkPPSDXWstAuKwpf1Q:b-o6ls22) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L296)):*

Can the names of these functions be made more distinct?
</blockquote></details>

Renamed to "batchAndSendRequests" and "sendRequestBatch", that's a little more distinct.  Ultimately, they are performing a related function and it's going to be difficult to differentiate the names too much.

---

*[pkg/ui/src/redux/metrics.ts, line 302 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw67XOj93cc7fbakmar:-Kw6XEmzUzrGprWoQbH5:bcs31xo) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L302)):*

Wow it's hurting my brain trying to figure out how a plain `try`/`catch` works here.  I guess this is just generator magic, but it seems really counter intuitive to me.  I'm gonna need to go look at the transpiled output to figure out what it ends up as.
</blockquote></details>

When you want to resume a generator, you can use either `iterator.next(result)` or `iterator.throw(error)`.  The caller can thus make the generator throw an exception at any yield statement.  Saga will throw an exception into the generator if the call()ed function throws (or, in the case of an async function, if the call()ed function returns a promise which is rejected).

---

*[pkg/ui/src/redux/state.ts, line 33 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw68--IYj9Ny1BYD5MD:-KwkFS2oOJAqLg5AKeE1:b-c3pmlx) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/state.ts#L33)):*

It might be nice to split up the rest configuration a bit, maybe not in this PR if you'd like to keep it focused.
</blockquote></details>

Yes, I'd like to keep it focused on the saga as much as possible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063)*

mrtracy:
Thanks to @benesch for the assist with incompatibility shenanigans.

couchand:
Reviewed 10 of 10 files at r2.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 184 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw63si2zouybHyUuspm:-KwpWVJOprNROIKarVBR:bme06li) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L184)):*

`yield take(metrics.REQUEST)` is the effect; the saga we are testing is only responsible for asking for the next action.  Redux Saga will accept this and eventually yield back the next matching action.
</blockquote></details>

I see, same discussion below.

---

*[pkg/ui/src/redux/metrics.spec.ts, line 197 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw648Ge-J2ijZx8OswQ:-KwpWZYNBr4mP2C23xWZ:bw1vbn8) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L197)):*

This is very precisely validating the control flow of the saga; we're verifying what directives the saga is giving to the redux saga system, given a certain pattern of input.

To decouple it from the precise behavior, you could write some sort of "mock-saga-runner" which would execute the output of the saga under test more automatically.  I think the output of the mock runner would be 1.) A Store and 2.) A list of pending operations which need additional test input.

This would allow us to obviate the need for precise validation of certain sequences (i.e. sequences of puts), and it would probably provide a better interface for analyzing pending requests (e.g. forks and calls). 

In other words, its likely a good idea - but I think it's beyond the scope of this PR, because that mock object is a little complex.  I think we can accept the coupling for now; it's still superior to the previous configuration, since we get much clearer insight into the high-level functioning of these actions that we did with thunks.
</blockquote></details>

That sounds like a good idea, but yeah, I agree out of scope for this change.  Made an issue for it: #19375

---

*[pkg/ui/src/redux/metrics.spec.ts, line 247 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw65vQ4ctmc3TeF1p8P:-KwpXl2-_Ks1w5SZWajN:b-qbxl8r) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.spec.ts#L247)):*

I've tried to make it a little more readable; the the test data is now initially sorted into `shortRequests` and `longRequests`, which are combined into a `mixedRequests` using a flatmap function.  No more numeric accesses.
</blockquote></details>

Looks great, thanks.

---

*[pkg/ui/src/redux/metrics.ts, line 252 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw64y1OQITXR29FieI-:-KwpY1vcwBSKCxLk6AfA:b7snix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L252)):*

`takeAll` creates a forked process for every message, but we are exercising greater control over the forking process.
</blockquote></details>

Oh, I meant `takeEvery`, don't know if that was confusing.  I have to admit not having tried it, but would something like this work?

```
function* queryMetricsSaga() {
  yield takeEvery(REQUEST, handleAction);

  function* handleAction(requestAction: PayloadAction<WithID<TSRequest>>) {
    // Dispatch action to underlying store.
    yield put(requestAction);
    requests.push(requestAction.payload);
 
    // If no other requests are queued, fork a process which will send the
    // request (and any other subsequent requests that are queued).
    if (requests.length === 1) {
      yield fork(sendRequestsAfterDelay);
    }
  }

  // other closure stuff here ...
}
```

Not sure if that's a net win, but it seems maybe more idiomatic?

---

*[pkg/ui/src/redux/metrics.ts, line 302 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063#-Kw67XOj93cc7fbakmar:-KwpZY4HbpaRvzNqT-z6:bpkigfb) ([raw file](https://github.com/cockroachdb/cockroach/blob/a297eda049556820b31386c5c338140c6a16a858/pkg/ui/src/redux/metrics.ts#L302)):*

When you want to resume a generator, you can use either `iterator.next(result)` or `iterator.throw(error)`.  The caller can thus make the generator throw an exception at any yield statement.  Saga will throw an exception into the generator if the call()ed function throws (or, in the case of an async function, if the call()ed function returns a promise which is rejected).
</blockquote></details>

I guess my comment was not about redux-saga, but rather just complaining about the generator syntax.  I don't know how to get a consistent mental model of what's happening with generator.  What does `yield` do?  I had thought it was like a return, but you could resume later.  But if the caller can "throw into", that completely breaks that idea.

I just don't like generators, I guess.  I don't think think the concept adds any particular value, and it requires me to think so hard about the specific details of the implementation.  It's not an abstraction that lets me ignore the details, which is what I want out of language features.

This isn't to say we shouldn't go forward here, just that I'm still not a fan of generators.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19063)*

PR #19068, opened 2017-10-05
sql/parser: some scanner optimizations

jordanlewis:
Go's ToUpper and ToLower functions don't assume anything about their
input strings, so they have to do a bunch of work to properly deal with
Unicode characters.

There's 2 hot-path places in the scanner where we already know that the
input is ASCII-only, so we can skip a lot of that work.

https://github.com/golang/go/issues/17859 is an open issue against Go to
improve these functions if the input's ASCII, for what it's worth.

```
name     old time/op    new time/op    delta
Parse-8    39.7µs ± 1%    34.0µs ± 1%  -14.35%  (p=0.000 n=10+9)

name     old alloc/op   new alloc/op   delta
Parse-8    6.99kB ± 0%    6.75kB ± 0%   -3.43%  (p=0.000 n=10+10)

name     old allocs/op  new allocs/op  delta
Parse-8       193 ± 0%       194 ± 0%   +0.52%  (p=0.000 n=10+10)
```


pkg/sql/parser/scan.go
```diff
@@ -520,8 +520,23 @@ func (s *Scanner) scanIdent(lval *sqlSymType) {
 	start := s.pos - 1
 	for ; isIdentMiddle(s.peek()); s.pos++ {
 	}
-	lval.str = Name(s.in[start:s.pos]).Normalize()
-	if id, ok := keywords[strings.ToUpper(lval.str)]; ok {
+	// We know that the identifier we've seen so far is ASCII, so we don't need
+	// to unicode normalize. Instead, just upper and lowercase
+	bLower := make([]byte, s.pos - start)
+	bUpper := make([]byte, s.pos - start)
```
nvanbenschoten: Perhaps we could lazily allocate these in the case where `s.in[start:s.pos]` is either all lowercase or all uppercase.

pkg/sql/parser/scan.go
```diff
@@ -795,14 +812,28 @@ func isNonKeywordBareIdentifier(s string) bool {
 			return false
 		}
 	}
-	upper := strings.ToUpper(s)
-	if _, ok := reservedKeywords[upper]; ok {
+
+	// If we made it this far, we know there's no unicode characters in our
```
nvanbenschoten: How do we know this?

pkg/sql/parser/scan.go
```diff
@@ -520,8 +520,23 @@ func (s *Scanner) scanIdent(lval *sqlSymType) {
 	start := s.pos - 1
 	for ; isIdentMiddle(s.peek()); s.pos++ {
 	}
-	lval.str = Name(s.in[start:s.pos]).Normalize()
-	if id, ok := keywords[strings.ToUpper(lval.str)]; ok {
+	// We know that the identifier we've seen so far is ASCII, so we don't need
```
nvanbenschoten: Could you explain how we know this?

pkg/sql/parser/scan.go
```diff
@@ -795,14 +812,28 @@ func isNonKeywordBareIdentifier(s string) bool {
...
 	}
-	upper := strings.ToUpper(s)
-	if _, ok := reservedKeywords[upper]; ok {
+
+	// If we made it this far, we know there's no unicode characters in our
+	// string, so we can use a naive unicode-unaware uppercase algorithm that's
+	// faster than Strings.ToUpper.
+	// See https://github.com/golang/go/issues/17859 for ToUpper performance.
+	isLower := true
+	b := make([]byte, len(s))
```
nvanbenschoten: Same thing here, it's too bad we have to allocate even if `s` already is all uppercase.

nvb (commented):
Nice!

knz:

---

*[pkg/sql/parser/scan.go, line 539 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviPyS3BjUqXevIYSf-:-KviPyS4xeGZ3MMSQL3l:bluikxo) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L539)):*
> ```Go
> 	}
> 	lval.str = string(bLower)
> 	if id, ok := keywords[string(bUpper)]; ok {
> ```

woah, what is this madness. Since we have to compute the lowercase version anyway, what about populating the keyword[] array with lowercase keys too? There is no reason to have that indexed by uppercase strings. This way you save half of the work!

Then one step further: what about having `Scanner.in` be a `[]byte` instead of `string`, ditto for `lval.str` and do the lowercasing in-place and just aliasing the slices everywhere. This way no need to allocate one byte array per identifier or literal. The things could only be converted/copied to string when strictly needed e.g. for StrVal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068)*

jordanlewis:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/parser/scan.go, line 523 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviDx_M4rbjYRaZq7uz-r1-523:-KvildiLTV2m60XtxyH0:ba94sbp) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L523)):*

Could you explain how we know this?
</blockquote></details>

We actually don't. I was confused. Fixed.

---

*[pkg/sql/parser/scan.go, line 526 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviDx_M4rbjYRaZq7uz-r1-526:-KvimVgzlMP7kioTEdgm:b1b7bz3) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L526)):*

Perhaps we could lazily allocate these in the case where `s.in[start:s.pos]` is either all lowercase or all uppercase.
</blockquote></details>

Fixed this.

---

*[pkg/sql/parser/scan.go, line 539 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviPyS3BjUqXevIYSf-:-KviVUoW5sfsyg2I8SHn:btwfol0) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L539)):*

woah, what is this madness. Since we have to compute the lowercase version anyway, what about populating the keyword[] array with lowercase keys too? There is no reason to have that indexed by uppercase strings. This way you save half of the work!

Then one step further: what about having `Scanner.in` be a `[]byte` instead of `string`, ditto for `lval.str` and do the lowercasing in-place and just aliasing the slices everywhere. This way no need to allocate one byte array per identifier or literal. The things could only be converted/copied to string when strictly needed e.g. for StrVal.
</blockquote></details>

Great point I didn't realize that this was the only place `keywords` was used. Fixed that :)

I'll get to this other smart idea you mention in the next patch.

---

*[pkg/sql/parser/scan.go, line 816 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviDx_M4rbjYRaZq7uz-r1-816:-KviVeClmXwzaHTLgl17:bzidu6) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L816)):*

How do we know this?
</blockquote></details>

We don't I was wrong. Fixed.

---

*[pkg/sql/parser/scan.go, line 821 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviDx_M4rbjYRaZq7uz-r1-821:-KviVhSdnDHXfdZmKwdf:bt2vr1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L821)):*

Same thing here, it's too bad we have to allocate even if `s` already is all uppercase.
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068)*

jordanlewis:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/parser/scan.go, line 821 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KviDx_M4rbjYRaZq7uz-r1-821:-KvivE6mzN3jhD13TZDj:bj4d6ec) ([raw file](https://github.com/cockroachdb/cockroach/blob/92aa5c05aa42ed272fececd14b9470754e49d6f7/pkg/sql/parser/scan.go#L821)):*

Fixed.
</blockquote></details>

Actually fixed now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068)*

knz:
:lgtm: modulo missing explanatory comment.

---

Reviewed 5 of 5 files at r2.

---

*[pkg/sql/parser/scan.go, line 521 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KvlkvlDqxjF1SzCMKs5:-KvlkvlDqxjF1SzCMKs6:b9pzjl5) ([raw file](https://github.com/cockroachdb/cockroach/blob/dc35cc050af88d55f3d236efa97e04d58e61dd98/pkg/sql/parser/scan.go#L521)):*
> ```Go
> func (s *Scanner) scanIdent(lval *sqlSymType) {
> 	start := s.pos - 1
> 	isASCII := true
> ```

Please add an explanatory comment here to explain the optimization. The reader must understand when they read your comment "why wasn't this code written in a more simple way?"

---

*[pkg/sql/parser/scan.go, line 827 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KvlkmD5HAt8p0d6COtT:-KvlkmD6VZzUmj2dDB9Q:bxxynib) ([raw file](https://github.com/cockroachdb/cockroach/blob/dc35cc050af88d55f3d236efa97e04d58e61dd98/pkg/sql/parser/scan.go#L827)):*
> ```Go
> 		return false
> 	}
> 	isASCII := s[0] < utf8.RuneSelf
> ```

ditto explanatory comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-:-KvlkxGEqHuSVjM7Dybg:b-qpljyb)*

jordanlewis:
TFTRs!

---

Review status: 4 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/parser/scan.go, line 521 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KvlkvlDqxjF1SzCMKs5:-KvmECNRaJPyVPzej3MF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dc35cc050af88d55f3d236efa97e04d58e61dd98/pkg/sql/parser/scan.go#L521)):*

Please add an explanatory comment here to explain the optimization. The reader must understand when they read your comment "why wasn't this code written in a more simple way?"
</blockquote></details>

Done.

---

*[pkg/sql/parser/scan.go, line 827 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-KvlkmD5HAt8p0d6COtT:-KvmED0wvA2s9e8e633_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dc35cc050af88d55f3d236efa97e04d58e61dd98/pkg/sql/parser/scan.go#L827)):*

ditto explanatory comment.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19068#-:-KvmEBSJREzqIDs8k3Vt:bz5mair)*

PR #19077, opened 2017-10-06
build: separate out race logic tests

dt:
The testrace build is going over 20min lately, so pulling logic tests (~12-15min) into their own build should help a bit.
A follow up step might be to split `default` from the rest of the logic tests if testlogicrace is still the slowest build. This includes the env vars needed for that, so if/when we decide to do so, it could just be a TC change.

### DIFF
--- a/build/teamcity-testlogicrace.sh
+++ b/build/teamcity-testlogicrace.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+export BUILDER_HIDE_GOPATH_SRC=1
+
+mkdir -p artifacts
+
+build/builder.sh env \
+	make testrace \
+  PKG=./pkg/sql/logictest
+	TESTFLAGS='-v' \
+	2>&1 \
+	| tee artifacts/testlogicrace.log \
+	| go-test-teamcity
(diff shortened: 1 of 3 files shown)


dt:
experimenting a bit to see how much we actually win here -- will need to actually add a (conditional) tc step before this is ready to move forward though, and depending on what timings look like, it might be worth splitting testlogicrace along default vs not-default while we're at it (since adding/changing build steps is kinda tricky / not atomic).

dt:
Okay, on this run:

`testlogicrace`: 12m23s
https://teamcity.cockroachdb.com/viewLog.html?buildId=370200&tab=buildResultsDiv&buildTypeId=Cockroach_UnitTests_Testlogicrace

`testrace` with the skip logic tests flag: 15m:07s
https://teamcity.cockroachdb.com/viewLog.html?buildId=370204&tab=buildResultsDiv&buildTypeId=Cockroach_UnitTests_Testrace

Compared to ~18-20m+ runs of full `testrace`, so it will shave a few mins off of total CI cycle.
Compilation is now 20% of that, so we could potentially save some more time if we found a way to reduce recompilation (about a third of that time in c-deps which change rarely and seem like we might be able to do some caching).

dt:
Another run: 
13m:02s `testrace` -logic
13m:05s `testlogicrace`

jordanlewis (approved):
LGTM. Let's do it!

PR #19078, opened 2017-10-06
sql: remove some allocation-heavy closures

jordanlewis:
Previously multiSourceInfo.findColumn used a closure to do some commonly
repeated work. Allocating this closure accounted for around 2.5% of the
total allocations recorded during a short run of TPCC. It's totally
pointless to keep it around, so extract it into its own function.

Also, typeCheckSameTypedExprs used a lot of closures to do its work. Create a
state object to pass around the arguments to avoid this problem, similar
to the strategy employed by typeCheckOverloadedExprs.

```
TypeCheck-8    15.1µs ± 1%    14.2µs ± 1%   -5.59%  (p=0.000 n=10+10)

TypeCheck-8    1.26kB ± 0%    0.68kB ± 0%  -45.86%  (p=0.000 n=10+10)

TypeCheck-8      38.0 ± 0%      32.0 ± 0%  -15.79%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/sql/data_source.go
+++ b/pkg/sql/data_source.go
@@ -791,6 +791,21 @@ func (src *dataSourceInfo) checkDatabaseName(tn parser.TableName) (parser.TableN
 	return tn, nil
 }
 
+func findColHelper(
+	src *dataSourceInfo, c *parser.ColumnItem, colName string, iSrc, srcIdx, colIdx, idx int,
+) (int, int, error) {
+	col := src.sourceColumns[idx]
+	if col.Name == colName {
+		if colIdx != invalidColIdx {
+			return invalidSrcIdx, invalidColIdx, pgerror.NewErrorf(pgerror.CodeAmbiguousColumnError,
+				"column reference %q is ambiguous", parser.ErrString(c))
+		}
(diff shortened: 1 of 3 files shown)


knz:
:lgtm:

What's really sad to me is that a proper compiler should be able to extract closures into a static function automatically if they are known to not escape their scope... The algorithm to do so is known since the early 1990's.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19078#-:-Kvl0-C0Yf6Dg70Mi3sr:b6rb6gb)*

jordanlewis:
Added a benchmark for completeness.

```
TypeCheck-8    15.1µs ± 1%    14.2µs ± 1%   -5.59%  (p=0.000 n=10+10)

TypeCheck-8    1.26kB ± 0%    0.68kB ± 0%  -45.86%  (p=0.000 n=10+10)

TypeCheck-8      38.0 ± 0%      32.0 ± 0%  -15.79%  (p=0.000 n=10+10)
```

knz:
Yay :tada: :100:

jordanlewis:
TFTR!

PR #19079, opened 2017-10-06
sql: slim down nameResolutionVisitor

jordanlewis:
nameResolutionVisitor previously allocated a slice of column offsets to
keep track of the number of columns that each data source was
responsible for. This showed up prominently in allocations profiles.

Instead, store the column offset right on the data source info.

@knz I think this is a little sketchy since there is some leaking abstraction, but I'm submitting it for discussion anyway. Perhaps we can think of a better idea.

### DIFF
--- a/pkg/sql/data_source.go
+++ b/pkg/sql/data_source.go
@@ -124,6 +124,12 @@ type dataSourceInfo struct {
 	// column but might be different if the statement renames
 	// them using AS.
 	sourceAliases sourceAliases
+
+	// colOffset is the offset of the first column in this dataSourceInfo in the
+	// multiSourceInfo array it is part of.
+	// The value is populated and used during name resolution, and shouldn't get
+	// touched by anything but the nameResolutionVisitor without care.
+	colOffset int
 }
 
 // planDataSource contains the data source information for data
--- a/pkg/sql/select_name_resolution.go
+++ b/pkg/sql/select_name_resolution.go
@@ -35,7 +35,6 @@ const invalidColIdx = parser.InvalidColIdx
 type nameResolutionVisitor struct {
 	err        error
 	sources    multiSourceInfo
-	colOffsets []int
 	iVarHelper parser.IndexedVarHelper
 	searchPath parser.SearchPath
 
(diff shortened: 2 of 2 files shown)


knz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19079#-:-Kvl9-Wqnu42tQgg1avW:bnfp4nl)*

jordanlewis:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/data_source.go, line 128 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19079#-Kvljd6RhauBDUGcGiOT:-KvmB6GQrHtTRTMrYLDA:bfei5l7) ([raw file](https://github.com/cockroachdb/cockroach/blob/27aa2dbfb222bda3a7fa37f4ff020471f02d1987/pkg/sql/data_source.go#L128)):*

You can fix the "abstraction leak" durably by changing/extending the definition of this field in the comment:

```
// colOffset is the offset of the first column in this dataSourceInfo in the multiSourceInfo array it is part of.
// The value is currently populated during name resolution.
```
</blockquote></details>

Thanks for the suggestion; done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19079)*

knz:
Thanks!

---

Reviewed 1 of 1 files at r2.

---

*[pkg/sql/data_source.go, line 128 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19079#-Kvljd6RhauBDUGcGiOT:-KvmCmE8_xLoP2m92i51:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/27aa2dbfb222bda3a7fa37f4ff020471f02d1987/pkg/sql/data_source.go#L128)):*

Thanks for the suggestion; done.
</blockquote></details>

:+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19079#-:-KvmCoPOI19kqhL-9dMo:b-ucfknh)*

PR #19080, opened 2017-10-06
log: improve error reporting

tbg:
In particular, this should allow errors like "write: no space left on device" to
expose themselves on sentry.

### DIFF
--- a/pkg/util/log/crash_reporting.go
+++ b/pkg/util/log/crash_reporting.go
@@ -16,11 +16,13 @@ package log
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"runtime"
 	"runtime/debug"
 	"strings"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	raven "github.com/getsentry/raven-go"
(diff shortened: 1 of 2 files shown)


bdarnell:
:lgtm:

---

---

*[pkg/util/log/crash_reporting.go, line 202 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-KwRDnl7KuKVQtC3JgDL:-KwRDnl7KuKVQtC3JgDM:b-pmigbn) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4152a99fd839e2fc44f0f4e84dd983751a7cafd/pkg/util/log/crash_reporting.go#L202)):*
> ```Go
> }
> 
> func censor(r interface{}) string {
> ```

s/censor/redact/g

---

*[pkg/util/log/crash_reporting.go, line 203 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-KwRDwyqDajGg1IKZEno:-KwRDwyrTkO4Ekyw-lyt:b-sk02xk) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4152a99fd839e2fc44f0f4e84dd983751a7cafd/pkg/util/log/crash_reporting.go#L203)):*
> ```Go
> 
> func censor(r interface{}) string {
> 	handlePathError := func(v *os.PathError) string {
> ```

These local functions look more confusing than just doing these inline in the type switch (as long as they're not reused, and even then you can often handle it with a fallthrough (handleSafeType looks like the only exception)). 

---

*[pkg/util/log/crash_reporting_test.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-KwRGC_5zJ9mPw0eICpu:-KwRGC_5zJ9mPw0eICpv:b-tveccl) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4152a99fd839e2fc44f0f4e84dd983751a7cafd/pkg/util/log/crash_reporting_test.go#L73)):*
> ```Go
> 		},
> 		{
> 			format: "", rs: []interface{}{os.NewSyscallError("write", syscall.Errno(28))},
> ```

Errno values are not portable; use `errno.ENOSPC`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-:-KwRGWdpZNpc3l6M6Uzd:bnfp4nl)*

tschottdorf:

---

*[pkg/util/log/crash_reporting.go, line 202 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-KwRDnl7KuKVQtC3JgDL:-Kw_NMu-plLHI6cEznp3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4152a99fd839e2fc44f0f4e84dd983751a7cafd/pkg/util/log/crash_reporting.go#L202)):*

s/censor/redact/g
</blockquote></details>

Done.

---

*[pkg/util/log/crash_reporting.go, line 203 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-KwRDwyqDajGg1IKZEno:-Kw_NO1aUDA5q7ZxgSvv:bvjbepa) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4152a99fd839e2fc44f0f4e84dd983751a7cafd/pkg/util/log/crash_reporting.go#L203)):*

These local functions look more confusing than just doing these inline in the type switch (as long as they're not reused, and even then you can often handle it with a fallthrough (handleSafeType looks like the only exception)). 
</blockquote></details>

Done. I opted to avoid fallthroughs though.

---

*[pkg/util/log/crash_reporting_test.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080#-KwRGC_5zJ9mPw0eICpu:-Kw_NnngTC5suhDiMmS4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d4152a99fd839e2fc44f0f4e84dd983751a7cafd/pkg/util/log/crash_reporting_test.go#L73)):*

Errno values are not portable; use `errno.ENOSPC`. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19080)*

PR #19081, opened 2017-10-06
sql/parser: Print TimestampTZ with correct time zone.

rytaft:
Make Format() print the correct time zone for TimestampTZ types.
Currently, TimestampTZ is always printed in UTC time, regardless
of the time zone.

### DIFF
--- a/pkg/sql/parser/datum.go
+++ b/pkg/sql/parser/datum.go
@@ -1698,7 +1698,7 @@ func (d *DTimestampTZ) Format(buf *bytes.Buffer, f FmtFlags) {
 	if !f.bareStrings {
 		buf.WriteByte('\'')
 	}
-	buf.WriteString(d.UTC().Format(TimestampOutputFormat))
+	buf.WriteString(d.Time.Format(TimestampOutputFormat))
 	if !f.bareStrings {
 		buf.WriteByte('\'')
 	}
--- a/pkg/sql/parser/eval_test.go
+++ b/pkg/sql/parser/eval_test.go
@@ -633,7 +633,7 @@ func TestEval(t *testing.T) {
 		{`('2010-09-28'::date)::timestamp`, `'2010-09-28 00:00:00+00:00'`},
 		{`'2010-09-28 12:00:00.1-04'::timestamp`, `'2010-09-28 16:00:00.1+00:00'`},
 		{`'2010-09-28 12:00:00.1-04'::timestamp::text`, `'2010-09-28 16:00:00.1+00:00'`},
-		{`'2010-09-28 12:00:00.1-04'::timestamptz::text`, `'2010-09-28 16:00:00.1+00:00'`},
+		{`'2010-09-28 12:00:00.1-04'::timestamptz::text`, `'2010-09-28 12:00:00.1-04:00'`},
 		{`'12h2m1s23ms'::interval`, `'12h2m1s23ms'`},
 		{`'12h2m1s23ms'::interval::text`, `'12h2m1s23ms'`},
 		{`interval '1'`, `'1s'`},
@@ -726,7 +726,7 @@ func TestEval(t *testing.T) {
 		{`extract_duration(millisecond from '20s30ms40µs'::interval)`, `20030`},
 		{`extract_duration(microsecond from '12345ns'::interval)`, `12`},
 		// Need two interval ops to verify the return type matches the return struct type.
-		{`'2010-09-28 12:00:00.1-04:00'::timestamptz - '0s'::interval - '0s'::interval`, `'2010-09-28 16:00:00.1+00:00'`},
+		{`'2010-09-28 12:00:00.1-04:00'::timestamptz - '0s'::interval - '0s'::interval`, `'2010-09-28 12:00:00.1-04:00'`},
 		{`'12h2m1s23ms'::interval + '1h'::interval`, `'13h2m1s23ms'`},
 		{`'12 hours 2 minutes 1 second'::interval + '1h'::interval`, `'13h2m1s'`},
 		{`'PT12H2M1S'::interval + '1h'::interval`, `'13h2m1s'`},
--- a/pkg/sql/parser/format_test.go
+++ b/pkg/sql/parser/format_test.go
@@ -136,6 +136,10 @@ func TestFormatExpr(t *testing.T) {
 			`('3s')[interval]`},
 		{`date '2003-01-01'`, FmtShowTypes,
 			`('2003-01-01')[date]`},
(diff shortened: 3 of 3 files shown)


dt:
:lgtm:

I'm curious where this came up? my recollection of the PG docs is that TimestampTZ's behavior is just that it renders in the _session_ offset, so where we send and receive them, in pgwire, we always do something like `.In(ctx.GetLocation())`, but I didn't realize we surfaced/relied on the raw datum `Format()` output. Indeed, if so, so we need the session location plumbed though, or to ensure they're localized earlier?

---

---

*[pkg/sql/parser/datum.go, line 1701 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-Kvk1TFse4F6fWenVf2P:-Kvk1TFse4F6fWenVf2Q:bcp918p) ([raw file](https://github.com/cockroachdb/cockroach/blob/a01e091ec9702933e9b97497a8ef9a66c58e55d0/pkg/sql/parser/datum.go#L1701)):*
> ```Go
> 		buf.WriteByte('\'')
> 	}
> 	buf.WriteString(d.In(d.Location()).Format(TimestampOutputFormat))
> ```

I might be mis-remembering, but I think `d.In(d.Location())` just makes a copy with the location to whatever it was already, so, if that's true, I think you can just remove the `.UTC()` and get the same behavior.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-:-Kvk1qtrMN06bpUQ6TwG:b-z1xsz4)*

rytaft:
Thanks! I noticed it because I was writing new test cases for the date_trunc() function I'm adding, and tests were failing that should have succeeded.  So I'm not sure that Format() is getting used in many places other than for testing.  But it seems that there is probably another place where .UTC() is being called instead of .In(d.getLocation()), since I just tried the following:

```
root@:26257/> SELECT TIMESTAMP WITH TIME ZONE '2001-08-27 12:30:23.345678+04' as TSTZ;
+----------------------------------+
|               tstz               |
+----------------------------------+
| 2001-08-27 08:30:23.345678+00:00 |
+----------------------------------+
(1 row)
```

As you can see, it's ignoring the time zone and printing the timestamp in UTC.

Should I still commit this change for Format(), or try to fix the other issue as well for the command line output?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-:-Kvmy633NuZyxw_iNnzl:b8tttu5)*

rytaft:

---

*[pkg/sql/parser/datum.go, line 1701 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-Kvk1TFse4F6fWenVf2P:-Kvn1vt-xX8sgvjLeHn5:bj93q5j) ([raw file](https://github.com/cockroachdb/cockroach/blob/a01e091ec9702933e9b97497a8ef9a66c58e55d0/pkg/sql/parser/datum.go#L1701)):*

I might be mis-remembering, but I think `d.In(d.Location())` just makes a copy with the location to whatever it was already, so, if that's true, I think you can just remove the `.UTC()` and get the same behavior.
</blockquote></details>

It doesn't compile if I just remove the .UTC(), but it works if I change it to d.Time.Format.

I've amended the commit to use that instead.  Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081)*

dt:
I think that example is expected behavior, assuming your session offset is 0.

One thing that is a bit confusing is that the spec calls it a `timestamp with time zone` but it would more accurately be called a `timestamp that is aware that time zones are a thing` instead. `timestamp with time zone` doesn't _have_ or _store_ a time zone at all -- on disk it is just an integer number seconds/nano seconds from jan 1 1970 UTC, exactly the same as `timestamp without time zone` (aka plain `timestamp`). The only difference between the timestamp types is their behavior, namely that the `tz` variant parses and renders _in the session offset_. 

So, in your example, it parses the string, including the `+4` offset, but then it renders it in the session offset, which appears to be 0. If you did `SET TIME ZONE -5` I expect you'd see it render as `... 03:30... -05:00`.

As far as `Format` change though:

In most cases, the `pgwire` package is responsible for formatting all our internal datums for clients according to postgres's spec. Timestamps are usually rendered by the `formatTs()` method in there, which uses the session location to localize appropriately. This is the function that rendered the one in your example above.

I'm not exactly sure where we surface the result of `Datum.Format` -- we use it to print expressions, so it shows up in `dump`, in logs and messages, and in things like `SHOW CREATE TABLE` for `DEFAULT` or `CHECK` expressions. The main rule for Format is that it must round-trip if passed back to `parser`. Since we parse the offset string, this should change be fine, and it is more informative than always shifting to UTC first.

That said, most `time.Time`s will be UTC anyway: any value read from disk, or returned from e.g. `now()`, etc starts with no offset and it isn't until rendering in `pgwire` or in a few of the builtins that the session location is applied to it, but in the event there is a non-zero offset, i'd just as soon see it here.

---

---

*[pkg/sql/parser/datum.go, line 1701 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-Kvk1TFse4F6fWenVf2P:-KvnCcSpP_VkKxNHTEj1:b-9jev4h) ([raw file](https://github.com/cockroachdb/cockroach/blob/a01e091ec9702933e9b97497a8ef9a66c58e55d0/pkg/sql/parser/datum.go#L1701)):*

It doesn't compile if I just remove the .UTC(), but it works if I change it to d.Time.Format.

I've amended the commit to use that instead.  Thanks!

</blockquote></details>

Ah, yep, the joys of embedded structs. Looks good!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-:-KvnCajdxZ_ILHNtjpAn:bx1ingi)*

rytaft:
Ohh ok that makes sense.  Thanks for the info!  Once I get the checks to complete successfully I'll merge this in.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-:-KvnlRlQYuXYDkIuTQtI:b-kbpzqc)*

dt:
I think the CI failure was just a sad executor - at least I glanced at it and saw a bunch of `/usr/bin/ld: final link failed: No space left on device` so I just re-ran it and it is green now.

rytaft:
Awesome thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19081#-:-KvnqE98PhUgtCqqITxq:bxn9vqu)*

PR #19085, opened 2017-10-06
makefile: error if TESTS is set with make bench

RaduBerinde:


knz (approved):

Makefile
```diff
@@ -24,8 +24,21 @@
 
 PKG          := ./pkg/...## Which package to run tests against, e.g. "./pkg/storage".
 TAGS         :=
-TESTS        :=.## Tests to run for use with `make test`.
-BENCHES      :=## Benchmarks to run for use with `make bench`.
+
+ifeq ($(MAKECMDGOALS),bench)
```
benesch: This won't catch cases like `make build bench`. Can you use `ifneq ($(findstring bench,$(MAKECMDGOALS))),)` instead?

Makefile
```diff
@@ -24,8 +24,21 @@
...
+ifneq ($(TESTS),)
+$(error TESTS specified with "make bench"; did you mean BENCHES?)
+endif
+TESTS := -
+BENCHES :=. ## Benchmarks to run for use with `make bench`.
```
benesch: nit: spacing here looks funny.

benesch (approved):

Makefile
RaduBerinde: Done

Makefile
RaduBerinde: Fixed

benesch:
🎉 LGTM! Thanks, Radu!

RaduBerinde:
Actually, `ifneq ($(findstring bench,$(MAKECMDGOALS))),)` seems to always pass regardless of the target

RaduBerinde:
Ah, there was an extra paren.

PR #19088, opened 2017-10-06
sql: add equivalency groups from equality filters + related fixes

RaduBerinde:
#### sql: fix a case where we aren't setting the constant columns correctly

If we have a scan without a preferred order, the current code fails to set
`exactPrefix` correctly. This results in not setting the constant columns for
the exact prefix. For example: `SELECT * FROM kv WHERE k = 1` should have k as a
constant column.

The fix is to always calculate exactPrefix.

#### sql: refine orderingInfo with equalities from scan filters

Look for equality expressions in filters and add constant columns and
equivalency groups accordingly.

#### sql: refine filterNode orderingInfo with the filter expression

Apply the same treatment we applied to the scan node.

#### sql: improve orderingInfo property propagation for joins

Propagate all constant columns (existing code was based on the incorrect
assumption that we can only propagate constant *equality* columns); and
propagate key property when the equality columns form keys on both sides.

```
name                                                             old time/op    new time/op    delta
Planning/SELECT_*_FROM_abc-4                                        211µs ± 3%     211µs ± 2%    ~     (p=0.971 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_a_>_5_ORDER_BY_a-4                 190µs ± 6%     187µs ± 4%    ~     (p=0.190 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_b_=_5-4                            187µs ± 6%     188µs ± 7%    ~      (p=0.661 n=10+9)
Planning/SELECT_*_FROM_abc_WHERE_b_=_5_ORDER_BY_a-4                 195µs ± 9%     196µs ± 7%    ~      (p=0.968 n=10+9)
Planning/SELECT_*_FROM_abc_WHERE_c_=_5-4                            188µs ± 5%     184µs ± 9%    ~     (p=0.280 n=10+10)
Planning/SELECT_*_FROM_abc_JOIN_abc_AS_abc2_ON_abc.a_=_abc2.a-4     298µs ± 1%     297µs ± 5%    ~     (p=0.684 n=10+10)

name                                                             old alloc/op   new alloc/op   delta
Planning/SELECT_*_FROM_abc-4                                       31.7kB ± 0%    31.7kB ± 0%    ~       (p=0.928 n=7+7)
Planning/SELECT_*_FROM_abc_WHERE_a_>_5_ORDER_BY_a-4                22.6kB ± 0%    22.8kB ± 0%  +0.85%    (p=0.000 n=9+8)
Planning/SELECT_*_FROM_abc_WHERE_b_=_5-4                           23.5kB ± 0%    23.5kB ± 0%  +0.28%  (p=0.000 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_b_=_5_ORDER_BY_a-4                24.4kB ± 0%    24.6kB ± 0%  +0.79%   (p=0.000 n=10+9)
Planning/SELECT_*_FROM_abc_WHERE_c_=_5-4                           23.4kB ± 0%    23.4kB ± 0%  +0.33%  (p=0.000 n=10+10)
Planning/SELECT_*_FROM_abc_JOIN_abc_AS_abc2_ON_abc.a_=_abc2.a-4    71.6kB ± 0%    71.6kB ± 0%  +0.02%  (p=0.000 n=10+10)

name                                                             old allocs/op  new allocs/op  delta
Planning/SELECT_*_FROM_abc-4                                          354 ± 0%       354 ± 0%    ~     (p=1.000 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_a_>_5_ORDER_BY_a-4                   370 ± 0%       376 ± 0%  +1.62%  (p=0.000 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_b_=_5-4                              347 ± 0%       346 ± 0%  -0.29%  (p=0.000 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_b_=_5_ORDER_BY_a-4                   397 ± 0%       403 ± 0%  +1.51%  (p=0.000 n=10+10)
Planning/SELECT_*_FROM_abc_WHERE_c_=_5-4                              343 ± 0%       342 ± 0%  -0.29%  (p=0.000 n=10+10)
Planning/SELECT_*_FROM_abc_JOIN_abc_AS_abc2_ON_abc.a_=_abc2.a-4       716 ± 0%       717 ± 0%  +0.14%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/sql/expand_plan.go
+++ b/pkg/sql/expand_plan.go
@@ -34,7 +34,7 @@ func (p *planner) expandPlan(ctx context.Context, plan planNode) (planNode, erro
 	if err != nil {
 		return plan, err
 	}
-	plan = simplifyOrderings(plan, nil)
+	plan = p.simplifyOrderings(plan, nil)
 	return plan, nil
 }
 
@@ -422,33 +422,33 @@ func translateOrdering(desiredDown sqlbase.ColumnOrdering, r *renderNode) sqlbas
 // This determination cannot be done directly as part of the doExpandPlan
 // recursion (using desiredOrdering) because some nodes (distinctNode) make use
 // of whatever ordering the underlying node happens to provide.
-func simplifyOrderings(plan planNode, usefulOrdering sqlbase.ColumnOrdering) planNode {
+func (p *planner) simplifyOrderings(plan planNode, usefulOrdering sqlbase.ColumnOrdering) planNode {
 	if plan == nil {
 		return nil
 	}
 
 	switch n := plan.(type) {
 	case *createTableNode:
-		n.sourcePlan = simplifyOrderings(n.sourcePlan, nil)
+		n.sourcePlan = p.simplifyOrderings(n.sourcePlan, nil)
 
 	case *updateNode:
-		n.run.rows = simplifyOrderings(n.run.rows, nil)
+		n.run.rows = p.simplifyOrderings(n.run.rows, nil)
 
 	case *insertNode:
-		n.run.rows = simplifyOrderings(n.run.rows, nil)
+		n.run.rows = p.simplifyOrderings(n.run.rows, nil)
 
 	case *deleteNode:
-		n.run.rows = simplifyOrderings(n.run.rows, nil)
+		n.run.rows = p.simplifyOrderings(n.run.rows, nil)
 
 	case *explainDistSQLNode:
-		n.plan = simplifyOrderings(n.plan, nil)
+		n.plan = p.simplifyOrderings(n.plan, nil)
 
 	case *traceNode:
-		n.plan = simplifyOrderings(n.plan, nil)
+		n.plan = p.simplifyOrderings(n.plan, nil)
 
 	case *explainPlanNode:
 		if n.expanded {
-			n.plan = simplifyOrderings(n.plan, nil)
+			n.plan = p.simplifyOrderings(n.plan, nil)
 		}
 
 	case *indexJoinNode:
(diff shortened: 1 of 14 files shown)


knz:
I'm sorry I haven't caught the bug fixed by the first commit. 

---

Reviewed 5 of 5 files at r1, 3 of 3 files at r2, 8 of 8 files at r3.
Review status: 11 of 17 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/filter.go, line 37 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19088#-KvmBA3d3h89KtqwIvr9:-KvmBA3d3h89KtqwIvrA:b-6adu3m) ([raw file](https://github.com/cockroachdb/cockroach/blob/0cdde254f280c7e123d79368c249370933d664af/pkg/sql/filter.go#L37)):*
> ```Go
> }
> 
> func (f *filterNode) computeOrdering() {
> ```

if `simplifyOrdering` and the `computeOrdering` method had an `*EvalCtx` parameter (I would actually even pass the planner as reference throughout) you wouldn't need to add the planner as field in filterNode.

(Note: also if we plumbed the planner through all plan optimizations (as it probably should) we could remove the planner reference from all planNodes. I was planning to do that soon, but feel free to get a head start.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19088#-:-KvmAKE_1MT5NA2hKigT:bb8vuqr)*

knz:
Nice improvement throughout.

LGTM but I'd prefer if we didn't add fields to filterNode (in general we should be removing fields -- with orderinginfo as an exception, since it's a clear property).

---

Reviewed 6 of 6 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19088#-:-KvmC6Lu31KCqSeYLOOy:b-fhbzou)*

RaduBerinde:
Done, I didn't feel good about adding that planner field either.

knz:
:lgtm:

---

Reviewed 5 of 11 files at r5, 6 of 6 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19088#-:-KvmHpw_fPrHjImmzub2:bnfp4nl)*

PR #19096, opened 2017-10-06
cli: Improve help message for debug zip command

a-robinson:
Previously if you didn't provide a file argument, it would just print
out:

Error: exactly one argument is required
Failed running "debug"

Which doesn't indicate what argument is required and also kind of makes
it look like you were just running `cockroach debug`, not
`cockroach debug zip`.

Found by @dianasaur323

### DIFF
--- a/pkg/cli/zip.go
+++ b/pkg/cli/zip.go
@@ -18,7 +18,6 @@ package cli
 import (
 	"archive/zip"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


tschottdorf:
Why not the usual?

```go
	if len(args) != 1 {
		return usageAndError(cmd)
	}
```

a-robinson:
Because I'm a bad. Thanks for reminding me of that!

tbg (approved):

Issue #17370, opened 2017-08-01
perf: investigate gRPC packeting optimizations


petermattis:
gRPC creates unnecessary packets during unary response processing. An upstream PR (https://github.com/grpc/grpc-go/pull/1373) to fix this was buggy and rejected and we're waiting on a real fix from the gRPC folks. But above and beyond reducing the number of packets per response to 1, we can look at reducing it below 1 by combining multiple responses into a single packet. Additionally, flushing the write buffer to the connection currently blocks other goroutines. We should investigate a pipelined approach where the write buffer is filled and when it needs to be flushed a new buffer is swapped into place while the buffer is written to the connection.

irfansharif:
could you add more explicit details here? Not as yet familiar with the code paths where this would have a measurable effect or even how I'd go about measuring this? Exec latencies, RTT latencies? 

I also recall some prior experiments you ran investigating this effectively reducing {99,95}-th percentile latencies (?), if there's anything relevant pertaining to this it'd be nice to add.

petermattis:
See https://github.com/grpc/grpc-go/blob/master/transport/http2_server.go#L816 and https://github.com/grpc/grpc-go/blob/master/transport/http2_client.go#L681. The measurable effect of reducing the number of packets (or perhaps the benefit is from reducing the number of system calls) is measurably lower latencies which directly impact performance. Also see https://github.com/grpc/grpc-go/pull/1373 which demonstrated some of the benefits, but had a deadlock scenario. What I'd like to explore here is even more aggressively reducing flushes when there are multiple goroutines sending concurrently. The tricky part here is how to do this without introducing deadlock scenarios caused by the application level flow control. But perhaps experimental code doesn't need to be perfect in order to show whether further investment is worthwhile.

petermattis:
In addition to the Cockroach measured network latencies, I also look at the packet/sec numbers that the prometheus node exporter provides. For a given queries/sec, lower packet/sec is usually better.

irfansharif:
Simply aggregating flushes for a given request into one.
```
___ops/s_____MB/s__p50(ms)__p95(ms)__p99(ms)
  +5.81%   +5.67%  -34.65%   -3.98%   -5.44%
```

Before:
![image](https://user-images.githubusercontent.com/10536690/29427790-f0f6e612-8358-11e7-96db-38f55d5aa003.png)

After:
[...]

PR #19098, opened 2017-10-06
[WIP] experimental cRPC protocol

petermattis:
The cRPC protocol is built on top of http2 using only http2 data
frames. The core read and write loops are similar to gRPC, but trimmed
down to their essence. A ton of stuff is missing before this could
actually be used, such as context cancellation propagation, trace
propagation, chunking of large requests/responses into frames, and
proper error handling. This PR was done as a proof of concept to see
what performance impact replacing gRPC with something else could
achieve.

On a single-node cluster with the local server optimization disabled,
gRPC on a read-mostly workload shows:
```
   10.0s        0         111904        11189.4      1.4      1.2      3.0      5.8     18.9
```
cRPC:
```
   10.0s        0         183923        18389.3      0.9      0.7      1.8      3.7     62.9
```
Unfortunately, the effects on a multi-node cluster are less dramatic.

I think it would be possible to make cRPC wire compatible with
gRPC. This would require sending a headers frame with every RPC, but the
headers are small (7 bytes) and constant for this use case. We'd also
have to send a settings frame as the first frame on the connection, but
that is trivial.

See #17370.

### DIFF
--- a/pkg/kv/transport.go
+++ b/pkg/kv/transport.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	"golang.org/x/net/context"
-	"google.golang.org/grpc"
 
 	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
@@ -42,7 +41,7 @@ type SendOptions struct {
 
 type batchClient struct {
 	remoteAddr string
-	conn       *grpc.ClientConn
+	crpcConn   *rpc.CRPCClientConn
 	client     roachpb.InternalClient
 	args       roachpb.BatchRequest
 	healthy    bool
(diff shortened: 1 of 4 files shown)


petermattis:
Sending this in the interest of sharing my experimentation. This is not meant to be a proposal that we go forward with this path.

petermattis:
Cc @a-robinson

petermattis:
Also, take a look at https://github.com/petermattis/pinger which allows comparison between grpc and crpc (a.k.a. the `y` protocol).

petermattis:
Testing this PR against current master (after rebasing) shows a 25% improvement in throughput on a 6-node cluster running `kv --read-percent=95`. Not sure why I wasn't seeing a benefit on a multi-node cluster before.

a-robinson:
I did a similar experiment of using grpc bidirectional streams for all batch requests/responses ([code here](https://github.com/a-robinson/cockroach/commits/grpc-stream)). All numbers below are from `kv --read-percent=95`

**On my laptop**

Head (e90d6202c0bf3e65a5723eb787d96882796a1271) with local server optimization removed:
```
   10.0s        0         111705        11168.9      1.4      1.2      3.0      6.0     31.5
```

Naive streams (non-pooled, reconnect every time, i.e. the second commit on my branch):
```
   10.0s        0         107300        10727.6      1.5      1.2      3.1      6.3     25.2
```

sync.Pool'ed streams:
```
   10.0s        0         137018        13699.0      1.2      1.0      2.5      6.0     28.3
```

This is a 23% improvement.

**On a 6 node roachprod cluster**

Head, load on just 1 node:
```
  120.6s        0         530804         4402.9      1.8      0.7     12.1     26.2    121.6
```

sync.Pool'ed streams, load on just 1 node:
```
  130.1s        0         594868         4572.3      1.7      0.6     13.6     26.2    318.8
```

Head, load on all 6 nodes:
9.8k to start, 9.25k stable
screen shot 2017-11-14 at 11 41 57 am

sync.Pool'ed streams, load on all 6 nodes:
Started around 13k, leveled off around 11.5k
screen shot 2017-11-14 at 11 20 36 am

This is a 24% improvement, so very consistent with the local macbook results. However, the fact that it's consistent is inconsistent with your crpc results that showed much less improvement on a 6 node cluster. It's also worth noting that using grpc streams means we'd lose out on built-in context propagation/cancellation.

However, what's weird is that if I do a bunch of range splits (by passing `--splits=8` to each of the 6 nodes' kv processes), the two basically even out at just over 15k QPS:
(head)
screen shot 2017-11-14 at 11 45 42 am
(sync.Pool'ed streams)
screen shot 2017-11-14 at 11 28 19 am

I don't yet have any good suggestions for why the improvement disappears if the ranges are well split.

petermattis:
Using a `sync.Pool` for the streams isn't kosher as you never close the streams. Probably fine for experimentation. 

> It's also worth noting that using grpc streams means we'd lose out on built-in context propagation/cancellation.

Because the context used to create the stream wouldn't be the one we're using for the RPC? I hadn't thought about that. That's unfortunate and makes this approach a lot less viable. Cc @bdarnell

a-robinson:
I tried re-running the `--splits=8` numbers just to verify I didn't mess something up, and the throughput difference of using grpc streams really does mostly go away with the splits, falling down to less than a 5% improvement. For comparison, even with pre-splits, cRPC shows a 20% improvement (18.5k vs 15.5k qps). It's very odd that splits would make the rpc layer's performance less important, since I'd think that the higher throughput enabled by splits would only make the rpc layer more important.

petermattis:
Closing this PR because we're unlikely to proceed with it any time soon. I'll keep my branch around.

PR #19104, opened 2017-10-06
cli,config,sql: rearrange zone config logic

benesch:
See individual commit messages for details. In summary:

* Move logic for parsing zone config "CLI specifiers", like `.default` and `db.tbl`, to `pkg/config`.
* Expose the CLI specifier and YAML config for each zone in a new virtual table `crdb_internal.zones`.
* Make our CLI client dumber by using the logic in `config` and the new virtual table in `sql`.

All this paves the way for new experimental SQL syntax to manipulate zone configs. I'll send a followup PR with that work shortly. The motivation for all of this work is partitioning: teaching zone configs about partitioning, and writing tests that modify zone configs, will be much easier with all the logic centralized somewhere.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -28,9 +28,148 @@ import (
 
 	"github.com/cockroachdb/cockroach/pkg/keys"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
+	"github.com/cockroachdb/cockroach/pkg/sql/parser"
 	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
 )
+
+// Several ranges outside of the SQL keyspace are given special names so they
+// can be targeted by zone configs.
+const (
+	DefaultZoneName    = "default"
+	MetaZoneName       = "meta"
+	SystemZoneName     = "system"
+	TimeseriesZoneName = "timeseries"
+)
+
+// NamedZones maps named zones to their pseudo-table ID that can be used to
+// install an entry into the system.zones table.
+var NamedZones = map[string]uint32{
+	DefaultZoneName:    keys.RootNamespaceID,
+	MetaZoneName:       keys.MetaRangesID,
+	SystemZoneName:     keys.SystemRangesID,
+	TimeseriesZoneName: keys.TimeseriesRangesID,
+}
+
+// NamedZonesByID is the inverse of NamedZones: it maps pseudo-table IDs to
+// their zone names.
+var NamedZonesByID = func() map[uint32]string {
+	out := map[uint32]string{}
+	for name, id := range NamedZones {
+		out[id] = name
+	}
+	return out
+}()
+
+// ZoneSpecifier identifies either a special named zone or the zone associated
+// with a particular database or table. Either NamedZone or Database must be
+// non-empty.
+type ZoneSpecifier struct {
+	NamedZone string
+	Database  string
+	Table     string
+}
+
+// ZoneSpecifierFromID creates a zone specifier for the zone with the given ID.
+func ZoneSpecifierFromID(
+	id uint32,
+	resolveID func(id uint32) (parentID uint32, name string, err error),
+) (ZoneSpecifier, error) {
+	if name, ok := NamedZonesByID[id]; ok {
+		return ZoneSpecifier{NamedZone: name}, nil
+	}
+	parentID, name, err := resolveID(id)
+	if err != nil {
+		return ZoneSpecifier{}, err
+	}
+	if parentID == keys.RootNamespaceID {
+		return ZoneSpecifier{Database: name}, nil
+	}
+	_, db, err := resolveID(parentID)
+	if err != nil {
+		return ZoneSpecifier{}, err
+	}
+	return ZoneSpecifier{Database: db, Table: name}, nil
+}
+
+// ParseCLIZoneSpecifier converts a single string s identifying a zone, as would
+// be used to name a zone on the command line, to a ZoneSpecifier. A valid CLI
+// zone specifier is either 1) a database or table reference of the form
+// DATABASE[.TABLE], or 2) a special named zone of the form [.NAME].
+func ParseCLIZoneSpecifier(s string) (ZoneSpecifier, error) {
+	if len(s) > 0 && s[0] == '.' {
+		name := s[1:]
+		if name == "" {
+			return ZoneSpecifier{}, errors.New("missing zone name")
+		}
+		return ZoneSpecifier{NamedZone: name}, nil
+	}
+	// ParseTableName is not vulnerable to SQL injection, so passing s directly
+	// is safe. See #8389 for details.
+	tn, err := parser.ParseTableName(s)
+	if err != nil {
+		return ZoneSpecifier{}, fmt.Errorf("malformed name: %s", s)
+	}
+	db := tn.Database()
+	if db == "" {
+		// No database was specified, so interpret the table name as the database.
+		db = tn.Table()
+		return ZoneSpecifier{Database: db}, nil
+	}
+	table := tn.Table()
+	return ZoneSpecifier{Database: db, Table: table}, nil
+}
+
+// ResolvePath converts a zone specifier to a slice of zone config IDs that
+// might apply to the zone. The last zone config ID to actually exist in
+// system.zones is the zone config that applies.
+//
+// Note that GetZoneConfigForKey performs the same logic without explicitly
+// constructing a path. This function exists to support the CLI.
+func (zs *ZoneSpecifier) ResolvePath(
+	resolveName func(parentID uint32, name string) (id uint32, err error),
+) ([]uint32, error) {
+	path := []uint32{keys.RootNamespaceID}
+	if zs.NamedZone != "" {
+		if zs.NamedZone == DefaultZoneName {
+			return path, nil
+		}
+		if id, ok := NamedZones[zs.NamedZone]; ok {
+			return append(path, id), nil
+		}
+		return nil, fmt.Errorf("%q is not a built-in zone", zs.NamedZone)
+	}
+	if zs.Database != "" {
+		databaseID, err := resolveName(path[len(path)-1], zs.Database)
+		if err != nil {
+			return nil, err
+		}
+		path = append(path, databaseID)
+	}
+	if zs.Table != "" {
+		tableID, err := resolveName(path[len(path)-1], zs.Table)
+		if err != nil {
+			return nil, err
+		}
+		path = append(path, tableID)
+	}
+	return path, nil
+}
+
+// CLISpecifier converts a ZoneSpecifier to a CLI zone specifier as described in
+// ParseCLIZoneSpecifier.
+func (zs *ZoneSpecifier) CLISpecifier() string {
+	if zs.NamedZone != "" {
+		return "." + zs.NamedZone
+	}
+	if zs.Table != "" {
+		return (&parser.TableName{
+			DatabaseName: parser.Name(zs.Database),
+			TableName:    parser.Name(zs.Table),
+		}).String()
+	}
+	return parser.Name(zs.Database).String()
+}
 
 const (
 	// minRangeMaxBytes is the minimum value for range max bytes.
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -15,11 +15,14 @@
 package config_test
 
 import (
+	fmt "fmt"
 	"reflect"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/kr/pretty"
 	"gopkg.in/yaml.v2"
 
 	"github.com/cockroachdb/cockroach/pkg/config"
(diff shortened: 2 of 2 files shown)


knz:
Thanks for tackling this. 

May I however request you opt for a `show` statement instead of an internal vtable, for two reasons:

1) the zone information should be well documented, whereas we do not publicly document crdb_internal
2) Given our current architecture around virtual tables currently not the right way to go wrt streaming behavior and memory usage. You do get the proper resource behavior however using `delegateQuery()` and an intermediate planNode and then document to users who want to filter/join this further to use `select * from [show zones]`

---

Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 1 of 1 files at r3.

---

*[pkg/config/config.go, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqp0SGuFKt65tlhVad:-Kvqp0SGuFKt65tlhVae:b-8bpt7r) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L37)):*
> ```Go
> 
> // Several ranges outside of the SQL keyspace are given special names so they
> // can be targeted by zone configs.
> ```

Extend the commit message to list these explicitly. Add the docs-todo label to the PR.

---

*[pkg/config/config.go, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqp6E47EmpQTIioC3K:-Kvqp6E47EmpQTIioC3L:b-lx61a5) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L45)):*
> ```Go
> )
> 
> // NamedZones maps named zones to their psuedo-table ID that can be used to
> ```

pseudo

---

*[pkg/config/config.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqp9evEw2ha0z65_1q:-Kvqp9evEw2ha0z65_1r:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L54)):*
> ```Go
> }
> 
> // NamedZonesByID is the inverse of NamedZones: it maps psuedo-table IDs to
> ```

ditto

---

*[pkg/config/config.go, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqpXOH90f-8oQyGZwU:-KvqpXOH90f-8oQyGZwV:b-hctacj) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L66)):*
> ```Go
> // ZoneSpecifier identifies either a special named zone or the zone associated
> // with a particular database or table. Either NamedZone or Database must be
> // non-nil; Table should only be non-nil if Database is non-nil.
> ```

It's extremely unusual in Go to see pointers to strings (because strings are immutable and string references lightweight). A comment must be present here to explain what is going on.

Note that strings being immutable means that extracting a slice from a string does not copy data.

---

*[pkg/config/config.go, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqprLFfjBAwNdMBVXM:-KvqprLGOAU30HvCIan8:b7md49l) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L79)):*
> ```Go
> ) (ZoneSpecifier, error) {
> 	if name, ok := NamedZonesByID[id]; ok {
> 		return ZoneSpecifier{NamedZone: &name}, nil
> ```

... therefore I think here you'd be fine with `NamedZone: name`.

The only argument that strongly mandates a pointer to string is when you need a (usually transient) reference to a string reference, because you want to observe how a string is being mutated over time from the perspective of an orthogonal component. I do not see this pattern clearly here.

---

*[pkg/config/config_test.go, line 544 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqqG1GZR7xodipGNpP:-KvqqG1GZR7xodipGNpQ:bvcwh57) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config_test.go#L544)):*
> ```Go
> 	config.NamedZones = map[string]uint32{
> 		config.DefaultZoneName: 0,
> 		"carl":                 42,
> ```

:sparkling_heart: 

---

*[pkg/config/config_test.go, line 607 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqqTYrrRUnvmTpQgZu:-KvqqTYrrRUnvmTpQgZv:b8g0mas) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config_test.go#L607)):*
> ```Go
> 		{`.`, nil, `"" is not a built-in zone`},
> 		{`".table."`, nil, `".table." not found`},
> 		{`".".".table."`, []uint32{0, 53, 54}, ""},
> ```

Unrelated: this reminds me that we could humor our users and remind them about the rules for fully qualified names:

"""
_Usually a fully qualified name is written like this: `exampledb.customer.address` or `test.kv.v` and can contain at most one or two periods. For example:_

```sql
create database ".𛲄·.·";
create table "𛲄.𛲄·"("𛲄.·.𛲄" int);
select "𛲄.𛲄·"."𛲄.·.𛲄" from ".𛲄·.·"."𛲄.𛲄·"."𛲄.·.𛲄";
```
"""

@jseldess what do you think?

---

*[pkg/config/config_test.go, line 624 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqsyU2yPfA3T6Le3Fj:-KvqsyU2yPfA3T6Le3Fk:b-d738fj) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config_test.go#L624)):*
> ```Go
> 			}
> 			if e, a := tc.path, path; !reflect.DeepEqual(e, a) {
> 				t.Errorf("#%d: path did not match expected path:%s\n",
> ```

`path: %s` (missing space)

---

*[pkg/sql/crdb_internal.go, line 1356 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqt5mHQ5sIKYIGDzil:-Kvqt5mHQ5sIKYIGDzim:bn55xn5) ([raw file](https://github.com/cockroachdb/cockroach/blob/e2f2eedb220d55725c7a4eede21ca1e122c7c529/pkg/sql/crdb_internal.go#L1356)):*
> ```Go
> CREATE TABLE crdb_internal.zones (
> 	id            INT NOT NULL,
>   cli_specifier STRING NOT NULL,
> ```

weird indentation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-:-KvqtpmZNC3OcHTJbUiH:b-9sbklu)*

benesch:
Hmm, that's an interesting point. I actually added the vtable to _support_ my planned `EXPERIMENTAL SHOW ZONE CONFIGURATION` statement, and I modeled the design after `SHOW var`, which (inefficiently, as you say), queries into a vtable. Let me rework that syntax to avoid the vtable, see how it feels, and update this PR accordingly.

---

---

*[pkg/config/config.go, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqprLFfjBAwNdMBVXM:-KvrAFwn6AoB-KY7UZbM:b-edjfo4) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L79)):*

... therefore I think here you'd be fine with `NamedZone: name`.

The only argument that strongly mandates a pointer to string is when you need a (usually transient) reference to a string reference, because you want to observe how a string is being mutated over time from the perspective of an orthogonal component. I do not see this pattern clearly here.
</blockquote></details>

Well, the pattern I'm trying to replicate is...

```
type ZoneSpecifier = NamedZone of str | Database of str | Table of (str, str)
```

Do you have a more idiomatic way of representing that particular ADT in Go? :D

I opted for string pointers to distinguish between the nil string and the empty string. For example, if we used strings instead of `*strings`, this code in `ParseCLIZoneSpecifier` below

```
if len(s) > 0 && s[0] == '.' {
	name := s[1:]
	return ZoneSpecifier{NamedZone: name}, nil
}
```

would return a completely empty `ZoneSpecifier`. I guess that's ok? `ResolvePath` could just return a "no zone specified" error if all three fields are empty.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-:-KvrBNl14oPs6iXtHnHX:b-wowhw1)*

knz:
if the vtable was guaranteed to only have few entries regardless of the deployment I wouldn't say anything. The zone table OTOH can grow arbitrarily large.

---

---

*[pkg/config/config.go, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqprLFfjBAwNdMBVXM:-KvrUebw0HnDSHXJbnL5:btsllwo) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L79)):*

Well, the pattern I'm trying to replicate is...

```
type ZoneSpecifier = NamedZone of str | Database of str | Table of (str, str)
```

Do you have a more idiomatic way of representing that particular ADT in Go? :D

I opted for string pointers to distinguish between the nil string and the empty string. For example, if we used strings instead of `*strings`, this code in `ParseCLIZoneSpecifier` below

```
if len(s) > 0 && s[0] == '.' {
	name := s[1:]
	return ZoneSpecifier{NamedZone: name}, nil
}
```

would return a completely empty `ZoneSpecifier`. I guess that's ok? `ResolvePath` could just return a "no zone specified" error if all three fields are empty.
</blockquote></details>

We do not allow the empty string as valid database or table name. It seems to me you can use that to disambiguate. Isn't that the case?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-:-KvrUS42txPDJ6mUu7fB:b-tmr1tm)*

benesch:
Yes, I agree the size of system.zones could be problematic. If you just
want the row for one zone, filtering the vtable will be needlessly
inefficient. But if your intention is to list all zones, each row requires
resolving the table and database name, and I worry that performing this
lookup row by row via a planNode will be much slower than scanning and
retrieving all namespace entries at the start like the vtable does.

Another option: delegate to a query that joins system.zones to
system.namespace twice to resolve the table and database names. Then we
could automatically take advantage of any future join optimizations, but I
have a feeling that query would be quite slow at the moment.

knz:
On 07-10-17 18:09, Nikhil Benesch wrote:
> but I
> have a feeling that query would be quite slow at the moment.

It would not be slower than a vtable: loading a namespace vtable in
memory would require the same scan of the underlying system table.

I'd say go for the join.

-- 
Raphael 'kena' Poss

benesch:
Ok, thanks for the review so far @knz! I've rebased out the second two commits, leaving only the first commit, which adds `config.ZoneSpecifier`, in this PR. I'll send along another PR with the revised SQL approach.

Happy to hold off merging this until that next PR is LGTM'd, but keeping this one around to avoid losing the review comments.

---

Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[pkg/config/config.go, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqp0SGuFKt65tlhVad:-Kvtyb3VasXc1Bqkh_Tg:bgatv4y) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L37)):*

Extend the commit message to list these explicitly. Add the docs-todo label to the PR.
</blockquote></details>

These already exist (and are documented); I'm just duplicating the constants from the CLI.

---

*[pkg/config/config.go, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqp6E47EmpQTIioC3K:-KvtyiUxtWcAqXz-x30R:bo3vl6y) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L45)):*

pseudo
</blockquote></details>

Thanks, done.

---

*[pkg/config/config.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqp9evEw2ha0z65_1q:-Kvtyj4W9KVp1Jaqt6hd:b5t6jp4) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L54)):*

ditto
</blockquote></details>

Also done.

---

*[pkg/config/config.go, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqpXOH90f-8oQyGZwU:-KvtykPyu2uwPGyifCvE:b-n81tul) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L66)):*

It's extremely unusual in Go to see pointers to strings (because strings are immutable and string references lightweight). A comment must be present here to explain what is going on.

Note that strings being immutable means that extracting a slice from a string does not copy data.
</blockquote></details>

Ok, switched to plain ol' `string`s.

---

*[pkg/config/config.go, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqprLFfjBAwNdMBVXM:-KvtynPLVBaFfhInmHMo:b-wpuska) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config.go#L79)):*

We do not allow the empty string as valid database or table name. It seems to me you can use that to disambiguate. Isn't that the case?
</blockquote></details>

Yeah, I can just avoid creating an empty `ZoneSpecifier` in `ParseCLIZoneSpecifier` instead. This seems better.

---

*[pkg/config/config_test.go, line 624 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqsyU2yPfA3T6Le3Fj:-KvtzGBKA5646RTIkf4n:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config_test.go#L624)):*

`path: %s` (missing space)
</blockquote></details>

Done.

---

*[pkg/sql/crdb_internal.go, line 1356 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kvqt5mHQ5sIKYIGDzil:-KvtzHP2-pyVI45Cnqj8:b-525nx1) ([raw file](https://github.com/cockroachdb/cockroach/blob/e2f2eedb220d55725c7a4eede21ca1e122c7c529/pkg/sql/crdb_internal.go#L1356)):*

weird indentation
</blockquote></details>

(Rebased out of this PR.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-:-KvtzK3n-QJNxM0he0_O:b9foaqh)*

jseldess:
*[pkg/config/config_test.go, line 607 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-KvqqTYrrRUnvmTpQgZu:-Kvv8AxX2jIg5Sj81tAt:b-eiz3ox) ([raw file](https://github.com/cockroachdb/cockroach/blob/284abb23aa04e0c3d459bd6ca97e6c9773e5c4b2/pkg/config/config_test.go#L607)):*
I definitely like clarifying this rule somewhere. Just noticed it's not explicitly mentioned in the docs anywhere so opened https://github.com/cockroachdb/docs/issues/1973.

I find "can contain at most one or two periods", in combination with the example you provide, ambiguous, however. I think we need to be more explicit that we're talking about the periods used to mark hierarchy (excluding the periods in entity names).  Perhaps something like this?

> A qualified name uses a single period to indicate a `database.table` relationship or a `table.column` relationship. A fully qualified name can thus contain, at most, one or two periods, excluding periods used in the entity names themselves. For example:

As you can see, I also removed the "Usually" from the start. Was it there for a reason?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104)*

a-robinson:
This mostly :lgtm: but if you're planning large changes to zone configs, mind summarizing your plans somewhere or chatting about them sometime?

---

Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[pkg/config/config.go, line 36 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0Yi5m6MJw5x51PUVC:-Kw0Yi5m6MJw5x51PUVD:bdr90x2) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L36)):*
> ```Go
> )
> 
> // Several ranges outside of the SQL keyspace are given special names so they
> ```

Is there any reason not to just make the CLI code use these rather than duplicating them? That same question applies to most of the logic here.

---

*[pkg/config/config.go, line 102 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0ZpIaEpXmRkmTQZcE:-Kw0ZpIbCYFURFKzfZEf:bw6m7pz) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L102)):*
> ```Go
> 	if len(s) > 0 && s[0] == '.' {
> 		name := s[1:]
> 		if name == "" {
> ```

I thought you said this new logic properly handled databases named "."? It looks like this wouldn't handle that case. I don't think that case is particularly important, but am confused by the difference between your commit message and this code.

---

*[pkg/config/config.go, line 103 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0_3dNR_LRrJnKp_gV:-Kw0_3dNR_LRrJnKp_gW:b-bkaeie) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L103)):*
> ```Go
> 		name := s[1:]
> 		if name == "" {
> 			return ZoneSpecifier{}, errors.New("missing zone name")
> ```

This may just be personal preference, but I always like to include the actual input in errors that claim the input was invalid. For example, `errors.Errorf("invalid zone name %q", s)`

---

*[pkg/config/config.go, line 111 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0_R-nx3rLyuOARMx4:-Kw0_R-nx3rLyuOARMx5:b-j95xjg) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L111)):*
> ```Go
> 	tn, err := parser.ParseTableName(s)
> 	if err != nil {
> 		return ZoneSpecifier{}, fmt.Errorf("malformed name: %s", s)
> ```

nit, but %q is nice for cases when the empty string is passed in

---

*[pkg/config/config.go, line 125 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0aYc9FSmBrI_o0IO0:-Kw0aYc9FSmBrI_o0IO1:b-yq0f9d) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L125)):*
> ```Go
> // ResolvePath converts a zone specifier to a slice of zone config IDs that
> // might apply to the zone. The last zone config ID to actually exist in
> // system.zones is the zone config that applies.
> ```

Referencing `system.zones` and its "last zone config ID" here seems a bit out of place. I don't know if we really want to commit to the ordering of entries in `system.zones` meaning anything. Was this just a copy-paste?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-:-Kw0YJCJpOqt0JzsHyKN:b-trhm77)*

danhhz:
> if you're planning large changes to zone configs, mind summarizing your plans somewhere or chatting about them sometime?

@a-robinson The only thing we're planning outside of what's mentioned in the RFC is a _TESTING ONLY_ sql command to let us set them in the partitioning tests (plus we'll rip out the duplicated logic that 3 or 4 other tests use to get around this being cli only right now).

knz:
Reviewed 4 of 6 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104)*

benesch:
@a-robinson I responded to your feedback here, but the changes are in the superseding PR, #19140.

---

---

*[pkg/config/config.go, line 36 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0Yi5m6MJw5x51PUVC:-Kw1xSsYKXXIEMP7SZ0a:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L36)):*

Is there any reason not to just make the CLI code use these rather than duplicating them? That same question applies to most of the logic here.
</blockquote></details>

Done.

---

*[pkg/config/config.go, line 102 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0ZpIaEpXmRkmTQZcE:-Kw1vj3-ZGOXF76OcFrF:b70rizx) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L102)):*

I thought you said this new logic properly handled databases named "."? It looks like this wouldn't handle that case. I don't think that case is particularly important, but am confused by the difference between your commit message and this code.
</blockquote></details>

Ah, so it only allows databases named `.` when properly escaped with double quotes, as they would be in SQL:

```
./cockroach zone set '"."'
```

Extremely goofy, but at least it's possible.

---

*[pkg/config/config.go, line 103 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0_3dNR_LRrJnKp_gV:-Kw1ws9yESpcFSitCDYM:b-retioh) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L103)):*

This may just be personal preference, but I always like to include the actual input in errors that claim the input was invalid. For example, `errors.Errorf("invalid zone name %q", s)`
</blockquote></details>

In this case, the only input that could generate the error message is the empty string, so I thought this message was actually clearer. Thoughts?

---

*[pkg/config/config.go, line 111 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0_R-nx3rLyuOARMx4:-Kw1x9i1NiiBNNyZwNTH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L111)):*

nit, but %q is nice for cases when the empty string is passed in
</blockquote></details>

Done.

---

*[pkg/config/config.go, line 125 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-Kw0aYc9FSmBrI_o0IO0:-Kw1xBF__xTpUgBPDNbX:bi5z8uv) ([raw file](https://github.com/cockroachdb/cockroach/blob/7494d9cd01592269c06bac7491705942e06434b5/pkg/config/config.go#L125)):*

Referencing `system.zones` and its "last zone config ID" here seems a bit out of place. I don't know if we really want to commit to the ordering of entries in `system.zones` meaning anything. Was this just a copy-paste?
</blockquote></details>

Yeah, this was just poorly worded. "The ID in the path with the highest index that *also* has an entry in system.zones is the zone config that applies" is what I was after.
Anyway, this function is gone in the latest revision.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19104#-:-Kw1xbYhzxlJlUlRdQV1:bszzrl)*

Issue #18860, opened 2017-09-28
sql: NaN = NaN (in Postgres)


petermattis:
In Postgres, `NaN = NaN`. This is a violation of IEEE754:

```
pmattis=# select 'NaN'::float = 'NaN'::float;
 ?column?
----------
 t
(1 row)
```

Cockroach provides the expected result
```
root@:26257/> select 'NaN'::float = 'Nan'::float;
+-----------------------------+
| 'NaN'::FLOAT = 'Nan'::FLOAT |
+-----------------------------+
| false                       |
+-----------------------------+
(1 row)
```

I'm not sure if there is anything to be done here as changing our behavior at this point would be backwards incompatible. And even if we could change it, I think the Postgres behavior is wrong. In Cockroach, `NaNs` can be detected using:

```
root@:26257/> SELECT 'NaN'::FLOAT IS NAN;
+---------------------+
| isnan('NaN'::FLOAT) |
+---------------------+
| true                |
+---------------------+
(1 row)
root@:26257/> SELECT isnan('NaN'::FLOAT);
+---------------------+
| isnan('NaN'::FLOAT) |
+---------------------+
| true                |
+---------------------+
(1 row)
```

The rationale for the Postgres behavior is:

```
Note: In most implementations of the "not-a-number" concept, NaN is not considered equal to any other numeric value (including NaN). In order to allow numeric values to be sorted and used in tree-based indexes, PostgreSQL treats NaN values as equal, and greater than all non-NaN values.
```

In Cockroach, we allow indexing `NaNs` and using the index to exclude `NaNs` from scans just as we index `NULLs`, but we maintain the IEEE754 behavior of `NaN != NaN`.

nvanbenschoten:
There's some inconsistency here. Even though we define `NaN != NaN`, our index encoding is not aware of this like [it is for NULL values](https://github.com/cockroachdb/cockroach/blob/d883c3a1c605aec228fbc9b5f7450869b8946de4/pkg/sql/sqlbase/table.go#L1447). This leads to unexpected errors like the following:
```
root@:26257/test> INSERT INTO b VALUES ('NaN');
pq: duplicate key value (a)=(NaN) violates unique constraint "b_a_key"
```

We should either fix this to make it work like it does for `NULL` or we should follow Postgres' lead of making `NaN = NaN`. I'm torn about what the right choice is, but at this point, things are broken either way so I'm not sure the backward compatibility argument is very strong.

petermattis:
[...]

PR #19106, opened 2017-10-07
sql: improve index constraints for boolean columns

nvb:
This PR includes 4 small changes that together greatly improve the
index constraints, and by extension, the index spans that we generate
when dealing with `BOOL` columns:

1. We now create index constraints for bool `IndexedVars` in `WHERE`
clauses.
2. We now normalize `NOT NOT b` to `b`, which allows us to improve
index selection when that pattern is found in `WHERE` clauses.
3. We now normalize `IS` and `IS NOT` expressions to `EQ` and `NE`
when possible. This allows index selection rules that work only for `EQ`
and `NE` to construct index constraints. This also simplifies things for #18860.
4. We now simplify `v != max` to `v < max` and `v != min` to `v > min`, which
allows us to create index constraints on these expressions.

None of the following queries were able to efficiently use indexes before.
Now they all can:
```
SELECT * FROM t WHERE b -- commit 1
SELECT * FROM t WHERE NOT NOT b -- commit 2
SELECT * FROM t WHERE b IS TRUE -- commit 3
SELECT * FROM t WHERE b IS NOT TRUE -- commit 3/4
SELECT * FROM t WHERE b IS FALSE -- commit 3
SELECT * FROM t WHERE b IS NOT FALSE -- commit 3/4
SELECT * FROM t WHERE b != true -- commit 4
SELECT * FROM t WHERE b != false -- commit 4
```


RaduBerinde:
:lgtm: Nice improvements!

---

---

*[pkg/sql/analyze.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxKnn0cYL507LIjMI:-KvqxKnn0cYL507LIjMJ:b-a0x0un) ([raw file](https://github.com/cockroachdb/cockroach/blob/f1422923b1ccfdd26b7003a9b51c64af3869c2e4/pkg/sql/analyze.go#L133)):*
> ```Go
> //   (a < 1 AND a > 2)  -> false
> //   (a > 1 OR a < 2)   -> true
> //   (a > 1 OR func(b)) -> true
> ```

[nit] perhaps add the new example here (a) -> (a = true)

---

*[pkg/sql/analyze.go, line 1483 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxRYw6xAO14qoW0fg:-KvqxRYw6xAO14qoW0fh:b-23xcch) ([raw file](https://github.com/cockroachdb/cockroach/blob/f1422923b1ccfdd26b7003a9b51c64af3869c2e4/pkg/sql/analyze.go#L1483)):*
> ```Go
> }
> 
> func simplifyBoolVar(evalCtx *parser.EvalContext, n *parser.IndexedVar) (parser.TypedExpr, bool) {
> ```

needs a comment. It's a little unintuitive what it does given its name, arguably "a" is simpler than "a = true"

---

*[pkg/sql/parser/normalize.go, line 450 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxsVzjENWjHZFsQj_:-KvqxsVzjENWjHZFsQja:bjnfzs9) ([raw file](https://github.com/cockroachdb/cockroach/blob/67582bf69529f78bb3e46270b536ce863576dba5/pkg/sql/parser/normalize.go#L450)):*
> ```Go
> 			return NewTypedAndExpr(
> 				NewTypedComparisonExpr(EQ, expr.TypedLeft(), expr.TypedRight()),
> 				NewTypedComparisonExpr(IsNot, expr.TypedLeft(), DNull),
> ```

Is this needed? Would a NULL ever satisfy an equality?

---

*[pkg/sql/parser/normalize.go, line 457 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxxlVBDVFI6ALUZOV:-KvqxxlVBDVFI6ALUZOW:bqrumyx) ([raw file](https://github.com/cockroachdb/cockroach/blob/67582bf69529f78bb3e46270b536ce863576dba5/pkg/sql/parser/normalize.go#L457)):*
> ```Go
> 			return NewTypedOrExpr(
> 				NewTypedComparisonExpr(NE, expr.TypedLeft(), expr.TypedRight()),
> 				NewTypedComparisonExpr(Is, expr.TypedLeft(), DNull),
> ```

Is this needed? isn't NE always true if one of the operands is NULL?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-:-KvqyEMfx5cNiWNIZNAp:bdyovjk)*

knz:
:lgtm: no more comments than radu

---

Reviewed 4 of 4 files at r1, 4 of 4 files at r2, 4 of 4 files at r3, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-:-KvqzgvfaE50n5vr2K5O:b4g3qm)*

nvanbenschoten:
Review status: 1 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/analyze.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxKnn0cYL507LIjMI:-KvrrU8Ey22kx_74R5BZ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f1422923b1ccfdd26b7003a9b51c64af3869c2e4/pkg/sql/analyze.go#L133)):*

[nit] perhaps add the new example here (a) -> (a = true)
</blockquote></details>

Done.

---

*[pkg/sql/analyze.go, line 1483 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxRYw6xAO14qoW0fg:-KvrrUe_G5a9sEu3YjH6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f1422923b1ccfdd26b7003a9b51c64af3869c2e4/pkg/sql/analyze.go#L1483)):*

needs a comment. It's a little unintuitive what it does given its name, arguably "a" is simpler than "a = true"
</blockquote></details>

Done.

---

*[pkg/sql/parser/normalize.go, line 450 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxsVzjENWjHZFsQj_:-KvrqHqCCMufBzfOa0LM:b-wcy447) ([raw file](https://github.com/cockroachdb/cockroach/blob/67582bf69529f78bb3e46270b536ce863576dba5/pkg/sql/parser/normalize.go#L450)):*

Is this needed? Would a NULL ever satisfy an equality?
</blockquote></details>

Added a comment. The gist is that `EQ` exprs don't handle `NULL` values the same way that `IS` exprs do.

---

*[pkg/sql/parser/normalize.go, line 457 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-KvqxxlVBDVFI6ALUZOV:-KvrqSSxe97aoMFChyhl:b-ne6p03) ([raw file](https://github.com/cockroachdb/cockroach/blob/67582bf69529f78bb3e46270b536ce863576dba5/pkg/sql/parser/normalize.go#L457)):*

Is this needed? isn't NE always true if one of the operands is NULL?
</blockquote></details>

Same as above. Also, got rid of `isVar(expr.Left) && `, which wasn't meant to be there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106)*

RaduBerinde:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19106#-:-KvtYViNunQca7rPmguF:bnfp4nl)*

PR #19108, opened 2017-10-07
sql: change help token to ?? (from ?)

justinj:
Unforuntately, `?` collides with a JSON operator, meaning keeping it as
the help token could cause some confusion and parsing issues. Change it
to ?? per discussion.

PR #19109, opened 2017-10-07
sql: add json operators to parser

justinj:
None of these actually do anything yet, but they should parse and format properly.

@knz please only look at the middle commit - first one is #19108 and the last one is
regenerated docs.

I remember Raphael mentioning he'd rather some operators be transformed into function
calls, rather than being actual operators, what are your feelings on that for these ones?
I've done them as normal operators here, wouldn't mind us coming up with some general
guidelines for when that's acceptable vs. not.

cc @dt, is my docs diff supposed to be so large?


dt:
@justinj probably not... I'm guessing you don't have the local jar so it used the web service, and that might no longer match our jar's output. Ugh. I need to find a better solution to this.

justinj:
After getting the local jar all was good!

knz:
:lgtm:

---

Reviewed 5 of 5 files at r2.
Review status: 5 of 8 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19109#-:-Kw-z4OivDDpgjPIeP2O:bnfp4nl)*

knz:
I'm ok with this patch.
My objection stands in general however.

The general design principle when designing an AST that *also* supports generic function calls is "do not add dedicated AST node types for operators unless there is at least one occurrence of special casing that operator later in the execution stage".

For us that would be "only give dedicated node types to JSON operators for which we will have special indexing support", and desugar the rest to function calls.

You can also *start* by desugaring *everything*, and only introduce dedicated node types when the dedicated support is introduced, later.

If you wish to follow this principle I'd be happy to review the corresponding amend to this PR.

petermattis:
@knz Not sure I understand your objection as Justin didn't add dedicated AST node types, only new comparison and binary operators. None of the existing binary operators have special indexing support and not all of the comparison ones do. So there doesn't seem anything too out of the ordinary from this PR in my opinion.

---

Review status: 5 of 8 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19109#-:-Kw05kHhINpw34MrXidB:b-6sjl8v)*

knz:
Yes that's why I'm saying "ok" on this as the first thing in my review.

However the code could desugar to builtin function calls without adding new enumeration values in binOpType.

petermattis:
> However the code could desugar to builtin function calls without adding new enumeration values in binOpType.

I'm not understanding why that would be better. You'd be removing enumeration values in exchange for new builtin functions. The phrase "6 of one, half a dozen of another" comes to mind. And de-sugaring to function calls for these operators would be distinctly different from the existing comparison and binary operators.

knz:
Maybe I wasn't clear, my suggestion implies de-sugaring everything that doesn't have dedicated support in the logical planning, not just the new operators.

petermattis:
Ah, yes, I didn't understand that. I'm still not understanding the benefit, though. Wouldn't de-sugaring everything that doesn't have dedicated support in logical planning be a very large undertaking as you'd be ruining round-tripping of queries? For example, `SELECT 1 + 2` would round-trip to `SELECT plus(1, 2)`. That seems like a lot of busy work and also a step backward, so clearly I must be missing something.

knz:
Your example is highlighting why this would not be a large concern -- we actually have special rules for arithmetic operators in logical planning, in "expression normalization" and "simplification" (your code), so these would not be desugared.

OTOH things like regular expression operators, bit fiddling etc, which are less common, should be desugared.

petermattis:
Ah, got it. Though note that the regular expression operators are normalized if one of the operands is `NULL`.

knz:
I think that all builtin function calls that do not declare explicitly to have a special behavior on NULL arguments get simplified to NULL during type checking.

justinj:
Ok, I'd rather not desugar the ones that will have indexing support at some point ("some point" being within 1.2), just to have change them back after, so I'll desugar `#-`, which is the only added one without future indexing support.

The binary operators which don't appear to get any special treatment are `&`, `|`, `#`, `^`, `||`, `%`, `<<`, and `>>` (all the comparison operators seem to get at normalized or analyzed in some way). Re: names, something like `BITAND(a, b)`, `BITOR(a, b)`, `BITXOR(a, b)`, `POW(a, b)`, `CONCAT_OP(a, b)` (`CONCAT` exists and has different behaviour than `||`), `MOD(a, b)`, `LSHIFT(a, b)`, `RSHIFT(a, b)`? That could maybe be it's own PR but since this PR isn't adding any new functionality anyway I don't see a reason to not just lump that change in here if we're ok with those names.

knz:
All OK except for `concat_op` which probably should be `concat_all`.

petermattis:
I don't see much point in de-sugaring the existing operators (or any of the operators introduced in this PR). Maintaining the ability to round-trip queries is also an important principle for the AST (one which we've not adhered to 100%).

knz:
Since the rest of this comment will derail the discussion (in an important way, mind, but still out of scope for the original change), please consider merging the PR without desugaring and push the desugaring discussion to a separate PR.

Now, regarding this: "Maintaining the ability to round-trip queries is also an important principle for the AST" - I do not know this to be true; given that 1) we certainly have not documented so anywhere 2) I can list several reasons why this is not desirable, in practice and in principle. 

Of course the reasons I have in mind may not be well informed, so before I put my foot in my mouth you could perhaps indulge and explain your position first. After all you were the person who introduced this idea in the first place, so if there's a good reason and it's not documented yet, perhaps you should really have the first shot at explaining.

justinj:
Seems like a good idea, opened an issue for this as #19128.

knz:
thx - let's continue that discussion on the new issue

knz:
:lgtm: with question on empty builtin

---

Reviewed 3 of 4 files at r1, 1 of 6 files at r3, 6 of 6 files at r4.

---

*[pkg/sql/parser/builtins.go, line 1518 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19109#-Kw1oZR1dNr9LdynZFll:-Kw1oZR1dNr9LdynZFlm:bdlfh8h) ([raw file](https://github.com/cockroachdb/cockroach/blob/8d688b368eb66080bd04b048f2e7c0ae965d4351/pkg/sql/parser/builtins.go#L1518)):*
> ```Go
> 	"json_remove_path": {
> 		Builtin{
> 		// TODO(justin): added here so #- can be desugared into it, still needs to be
> ```

Just be careful that if you merge this you may cause the thing to panic if you evaluate a call.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19109#-:-Kw1p1xfVPA0KDSKRo1m:b-bzav36)*

justinj:
Oh, good catch, I had intended to just have an empty list of builtins.

justinj:
TFTRs!

PR #19113, opened 2017-10-08
cmd/docgen: write bnf, skip svg when unchanged

dt:
The SVGs are slow to generate and the diffs are nearly uninteligible.

Writining the BNF to a file next to the generated SVG both results in more human readable diffs
and allows skipping the SVG generation if the BNF has not changed.


docs/generated/sql/diagrams/upsert.bnf
```diff
@@ -0,0 +1 @@
+insert_stmt ::=
```
benesch: Well this looks suspect.

pkg/cmd/docgen/diagrams.go
```diff
@@ -659,6 +659,22 @@ var diagramCmd = func() *cobra.Command {
 						fmt.Printf("%s: (POST REPLACE)\n\n%s\n", s.name, g)
 						return
 					}
+
+					name := strings.Replace(s.name, "_stmt", "", 1)
```
benesch: Not convinced it's worth opening up the potential for a name collision with this, but I don't feel strongly if you find the `_stmt` suffix that bothersome.

pkg/cmd/docgen/diagrams.go
```diff
@@ -659,6 +659,22 @@ var diagramCmd = func() *cobra.Command {
...
+
+					bnfFile := filepath.Join(baseDir, fmt.Sprintf("%s.bnf", name))
+					if bnf, err := ioutil.ReadFile(bnfFile); err != nil && !os.IsNotExist(err) {
+						log.Fatal(s.name, err)
+					} else if err == nil {
+						if bytes.Equal(bnf, g) {
+							fmt.Fprintf(os.Stderr, "no change in bnf for %s...\n", s.stmt)
```
benesch: 👏

benesch (commented):
This is *much* better. Nice! LGTM.

Once we have a proper pipeline set up to move this stuff into the docs repo, I'd vote for removing the generated HTML from source control.

benesch:
Oh, I forgot to mention: the BNFs for `show` and `upsert` seem... very broken?

pkg/cmd/docgen/diagrams.go
dt: this is the `name` from below, i just moved it up.

mjibson:
LGTM, excepting the empty BNFs already noted.

Issue #14768, opened 2017-04-10
storage: prioritization mechanism for foreground/background traffic


petermattis:
#14718 limited the bandwidth for preemptive snapshots (i.e. rebalancing) to 2 MB/sec. This is a blunt instrument. @bdarnell [says](https://github.com/cockroachdb/cockroach/pull/14718#issuecomment-292844403):

> What we really need is some sort of prioritization scheme that would allow snapshots to use the bandwidth only if it's not needed for other stuff. But I don't have any concrete suggestions so maybe we should go ahead and do this anyway.

petermattis:
Perhaps something like [weighted fair queueing](https://en.wikipedia.org/wiki/Weighted_fair_queueing). Here's a [go implementation](https://github.com/tadglines/wfq/blob/master/wfq.go).

tschottdorf:
Would an ideal solution have the limiting sit on the incoming bandwidth and not the outgoing? I'm thinking of the case in which node 1 and 2 both send to node 3, but node 1 streams a snapshot while node 2 has foreground traffic. It would really have to be node 3 which backs off the other two; node 1 and 2 each would try to go at full speed without a chance of determining their relative priorities. Naively, I'd think that WFQ for reading from the connection should work since there's flow control and the sender has to wait once they've filled up their window.

I've searched around for precedent which I'm sure must exist somewhere, but haven't really been successful.

petermattis:
I think we'd want foreground/background traffic prioritization on both the sender and receiver. The receiver might not have any foreground traffic, but the sender might. And vice versa.

cuongdo:
@m-schneider Can you take this on during 1.2? The issue is a little vague now, and this will require an RFC. So, it'd be helpful if you worked with @tschottdorf on making this actionable.

petermattis:
I'd estimate that the coding portion of this is a small fraction of the work. The bigger task is to test the prioritization mechanism under a variety of rebalancing and recovery scenarios on real clusters.

tschottdorf:
+ performance testing, as the mechanism will sit squarely in the hot path.

m-schneider:
@cuongdo Sure sounds like a great project for 1.2!

m-schneider:
[...]

PR #19115, opened 2017-10-08
storage: updating local Store before gossip should not crash

nvb:
Updating a target store's write stats immediately after rebalancing was
recently addressed in #18425. With that change, if `updateLocalStoreAfterRebalance`
is called before the `StorePool` had seen the `StoreDescriptor` in gossip,
it will trigger an NPE. This change fixes this by making the update a
no-op if the descriptor has yet to be seen in gossip.

It's my understanding that this update was best-effort, so ignoring it before gossip
is fine. That may be incorrect though and we may want to do something more here.

### DIFF
--- a/pkg/storage/store_pool.go
+++ b/pkg/storage/store_pool.go
@@ -327,14 +327,20 @@ func (sp *StorePool) deadReplicasGossipUpdate(_ string, content roachpb.Value) {
 	detail.deadReplicas = deadReplicas
 }
 
-// updateLocalStoreAfterRebalance is used to update local copy of target store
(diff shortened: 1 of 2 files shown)


a-robinson (approved):
LGTM. It is indeed a best-effort thing. Where did you see this crash, though?

nvanbenschoten:
I saw the crash when restarting a node that had about 500k ranges on it. I suspect that
the network traffic attempting to catch the node's `Replicas` up was getting in the way of
gossip updates. https://github.com/cockroachdb/cockroach/issues/14768 might help with issues like this in the future.

PR #19133, opened 2017-10-09
[DNM] exploration of Replica's Request "frontend"

tbg:
This is an exploration of the "life of a request" from `(*Store).Send` onward
with the goal of identifying the different components of the request
lifecycle ("stages").

What's hopefully clear from the exposition below is that the process of
obtaining a BatchResponse (or an error) from a BatchRequest is
straightforward. Yet, it is not a trivial exercise to "grok" this path from
our current codebase. Additionally, we're not doing a good job separating the
general tasks associated to this, which are (for writes):

1. "Replica frontend": query and propose to the:
2. "Replica backend": the actual state machine.

No attempt has been made here to untangle 2). That's much more complicated
since a lot of the logic is actually in Store, and there are various
callbacks between Store and Replica. But the frontend 1) seems fairly
straightforward.

The below is pseudocode. In particular, these various "stages" wouldn't be
interchangeable in practice, and none of the types make sense. The idea is
that I wanted to avoid having to nest six levels deep but to basically have
this call structure:

```
func Stage1(data Data) {
    nextData := Preprocess(data)
    prevReturn := Stage2(nextData)
    return Postprocess(prevReturn)
}
```

but again, consider that a presentational device only.

### DIFF
--- a/pkg/storage/refactor_replica.go
+++ b/pkg/storage/refactor_replica.go
@@ -1,0 +1,106 @@
+package storage
+
+import (
+	"golang.org/x/net/context"
+
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
+)
+
+// Store .
+type Store struct{}
+
+// Replica .
+type Replica struct{}
+
+// This is an exploration of the "life of a request" from `(*Store).Send` onward
+// with the goal of identifying the different components of the request
+// lifecycle ("stages").
+//
+// What's hopefully clear from the exposition below is that the process of
+// obtaining a BatchResponse (or an error) from a BatchRequest is
+// straightforward. Yet, it is not a trivial exercise to "grok" this path from
+// our current codebase. Additionally, we're not doing a good job separating the
+// general tasks associated to this, which are (for writes):
+//
+// 1. "Replica frontend": query and propose to the:
+// 2. "Replica backend": the actual state machine.
+//
+// No attempt has been made here to untangle 2). That's much more complicated
+// since a lot of the logic is actually in Store, and there are various
+// callbacks between Store and Replica. But the frontend 1) seems fairly
+// straightforward.
+//
+// The below is pseudocode. In particular, these various "stages" wouldn't be
+// interchangeable in practice, and none of the types make sense. The idea is
+// that I wanted to avoid having to nest six levels deep but to basically have
+// this call structure:
+//
+// ```
+// func Stage1(data Data) {
+//   nextData := Preprocess(data)
+//   prevReturn := Stage2(nextData)
+//   return Postprocess(prevReturn)
+// }
+// ```
+//
+// but again, consider that a presentational device only.
+
+// Send .
+func (s *Store) Send(
+	ctx context.Context, ba roachpb.BatchRequest,
+) (br *roachpb.BatchResponse, pErr *roachpb.Error) {
+	return Nest(
+		AnnotateCtxStage,        // augment ctx
+		SetActiveTimestampStage, // set ba.Timestamp
+		HLCStage,                // hlc.Update
+		ObservedTimestampStage,  // adjust ba.MaxTimestamp, on return populate br.Txn.MaxOffsets
+		// Retrying as necessary, handle WriteIntentError etc from the wrapped ReplicaSendStage.
+		// Internally concerns the intent resolver, the pushTxnQueue, and others. Deserves
+		// an "internal refactor" as well.
+		ConflictResolutionStage,
+		ReplicaSendStage, // *Replica.Send
+	)(ctx, ba)
+}
+
+// Send .
+func (r *Replica) Send(ctx context.Context, ba roachpb.BatchRequest) {
+	Nest(
+		AnnotateCtxStage,
+		DispatchStage, // delegate to one of the HandleX below
+	)
+}
+
+// HandleReadOnlyRequest .
+func (r *Replica) HandleReadOnlyRequest(Input) Output {
+	return Nest(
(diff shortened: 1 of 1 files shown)


petermattis:

---

*[pkg/storage/refactor_replica.go, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19133#-Kw2k-bj-MU5MH2HFrmO:-Kw2k-bj-MU5MH2HFrmP:b-i7o7qs) ([raw file](https://github.com/cockroachdb/cockroach/blob/a7ddada93a2727221d733ca0e9f8d6e29e793adc/pkg/storage/refactor_replica.go#L61)):*
> ```Go
> 		// an "internal refactor" as well.
> 		ConflictResolutionStage,
> 		ReplicaSendStage, // *Replica.Send
> ```

I've seen structuring like this in the past and it always feels artificial. Since the various stages are building up state, it might be beneficial to have an explicit `RequestState` structure and a series of method calls for the stages:

```
state := &RequestState{ctx, ba}
if err := state.AnnotateCtx(); err != nil {
  ...
}
if err := state.SetActiveTimestamp(); err != nil {
  ...
}
if err := state.HLC(); err != nil {
  ...
}
if err := state.ObservedTimestamp(); err != nil {
  ...
}
```

The above has the advantage of providing the re-assuring `if err := ...; err != nil` pattern. Our code wouldn't be idiomatic without it. 

I'm not sure if this approach is really much better, though capturing state in a `RequestState` structure could make the high-level steps clearer. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19133)*

nvanbenschoten:

---

*[pkg/storage/refactor_replica.go, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19133#-Kw2k-bj-MU5MH2HFrmO:-Kw6HyRN6LMyCZj0QNgx:b-suhqnz) ([raw file](https://github.com/cockroachdb/cockroach/blob/a7ddada93a2727221d733ca0e9f8d6e29e793adc/pkg/storage/refactor_replica.go#L61)):*

I've seen structuring like this in the past and it always feels artificial. Since the various stages are building up state, it might be beneficial to have an explicit `RequestState` structure and a series of method calls for the stages:

```
state := &RequestState{ctx, ba}
if err := state.AnnotateCtx(); err != nil {
  ...
}
if err := state.SetActiveTimestamp(); err != nil {
  ...
}
if err := state.HLC(); err != nil {
  ...
}
if err := state.ObservedTimestamp(); err != nil {
  ...
}
```

The above has the advantage of providing the re-assuring `if err := ...; err != nil` pattern. Our code wouldn't be idiomatic without it. 

I'm not sure if this approach is really much better, though capturing state in a `RequestState` structure could make the high-level steps clearer. 
</blockquote></details>

I like the idea of creating a structure like this a lot because it's declarative and clear (I'd use the word monadic, but I don't want that on my conscience). However, I feel like it could get us into the same kinds of issues that the `Sender` interface has. Mainly, it would force all interactions between stages to communicate through a necessarily constrained interface. It would also make and retries between stages difficult.

Also, for the sake of argument, could you define `Nest`? I think we'd actually want to split it up into `NestStore` and `NestReplica` like:

```
type storeFn func(context.Context, *Store, roachpb.BatchRequest) (roachpb.BatchRequest, *roachpb.BatchResponse, *roachpb.Error)
type replFn func(context.Context, *Replica, roachpb.BatchRequest) (roachpb.BatchRequest, *roachpb.BatchResponse, *roachpb.Error)

func NestStore(storeFn...) storeFn
func NestRepl(replFn...) replFn
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19133)*

bdarnell:

---

*[pkg/storage/refactor_replica.go, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19133#-Kw2k-bj-MU5MH2HFrmO:-KwSHazk7QmP4jCpJVco:b-v7djqh) ([raw file](https://github.com/cockroachdb/cockroach/blob/a7ddada93a2727221d733ca0e9f8d6e29e793adc/pkg/storage/refactor_replica.go#L61)):*

I like the idea of creating a structure like this a lot because it's declarative and clear (I'd use the word monadic, but I don't want that on my conscience). However, I feel like it could get us into the same kinds of issues that the `Sender` interface has. Mainly, it would force all interactions between stages to communicate through a necessarily constrained interface. It would also make and retries between stages difficult.

Also, for the sake of argument, could you define `Nest`? I think we'd actually want to split it up into `NestStore` and `NestReplica` like:

```
type storeFn func(context.Context, *Store, roachpb.BatchRequest) (roachpb.BatchRequest, *roachpb.BatchResponse, *roachpb.Error)
type replFn func(context.Context, *Replica, roachpb.BatchRequest) (roachpb.BatchRequest, *roachpb.BatchResponse, *roachpb.Error)

func NestStore(storeFn...) storeFn
func NestRepl(replFn...) replFn
```
</blockquote></details>

I'm not a fan of mutating a RequestState struct - I think part of the problem with the current code is that we shoehorn everything into the Send interface and use BatchRequest/BatchResponse for everything even though there should logically be different signatures at different levels (e.g. some levels require that BatchResponse.Error not be set). I'd rather be explicit about what state is used at which stages (which I guess means factoring out functions corresponding to these stages from the existing too-large methods)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19133)*

PR #19136, opened 2017-10-09
distsql: pre-allocate some slices and index vs appending to slice

richardwu:
In some cases, we append to a pre-allocated slice which is ~10% slower
than indexing into the slice (even with a counter)
```
name      time/op
Append-8  45.6ns ±11%
Idx-8     41.7ns ± 7%
```

Also, we can sometimes pre-allocate a slice by computing the size ahead
of time (even if you have to iterate through the slice). A
microbenchmark shows it's 20% worse to append instead of pre-computing
the size and allocating
```
name      time/op
Append-8  1.35µs ± 1%
Alloc-8   1.13µs ± 7%
```

### DIFF
--- a/pkg/sql/distsql_physical_planner.go
+++ b/pkg/sql/distsql_physical_planner.go
@@ -450,7 +450,7 @@ func makePlanToStreamColMap(numCols int) []int {
 	return m
 }
 
-// indetityMap returns the slice {0, 1, 2, ..., numCols-1}.
+// identityMap returns the slice {0, 1, 2, ..., numCols-1}.
 // buf can be optionally provided as a buffer.
 func identityMap(buf []int, numCols int) []int {
 	buf = buf[:0]
@@ -458,6 +458,15 @@ func identityMap(buf []int, numCols int) []int {
 		buf = append(buf, i)
 	}
 	return buf
+}
+
+// identityMapInPlace returns the modified slice such that it contains
+// {0, 1, ..., len(slice)-1}.
+func identityMapInPlace(slice []int) []int {
+	for i := range slice {
+		slice[i] = i
+	}
+	return slice
 }
 
 // spanPartition is the intersection between a set of spans for a certain
@@ -679,13 +688,15 @@ func getOutputColumnsFromScanNode(n *scanNode) []uint32 {
 			num++
 		}
 	}
-	outputColumns := make([]uint32, 0, num)
+	outputColumns := make([]uint32, num)
+	idx := 0
 	for i := range n.resultColumns {
 		// TODO(radu): if we have a scan with a filter, valNeededForCol will include
 		// the columns needed for the filter, even if they aren't needed for the
 		// next stage.
 		if n.valNeededForCol[i] {
-			outputColumns = append(outputColumns, uint32(i))
+			outputColumns[idx] = uint32(i)
+			idx++
 		}
 	}
 	return outputColumns
@@ -747,12 +758,13 @@ func (dsp *distSQLPlanner) createTableReaders(
 	var p physicalPlan
 	stageID := p.NewStageID()
 
-	for _, sp := range spanPartitions {
+	p.ResultRouters = make([]distsqlplan.ProcessorIdx, len(spanPartitions))
+	for i, sp := range spanPartitions {
 		tr := &distsqlrun.TableReaderSpec{}
 		*tr = spec
 		tr.Spans = make([]distsqlrun.TableReaderSpan, len(sp.spans))
-		for i := range sp.spans {
-			tr.Spans[i].Span = sp.spans[i]
+		for j := range sp.spans {
+			tr.Spans[j].Span = sp.spans[j]
 		}
 
 		proc := distsqlplan.Processor{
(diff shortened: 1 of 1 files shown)


RaduBerinde:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19136#-:-Kw1bQ-bASbvjWZP9ruy:bnfp4nl)*

pkg/sql/distsql_physical_planner.go
```diff
@@ -679,13 +688,15 @@ func getOutputColumnsFromScanNode(n *scanNode) []uint32 {
 			num++
 		}
 	}
-	outputColumns := make([]uint32, 0, num)
+	outputColumns := make([]uint32, num)
```
nvanbenschoten: This one seems crazy to me. Why are we iterating through `n.resultColumns` twice to get the `num` count when we can just append to an empty slice? The following would be idiomatic if we're not sure how selective `valNeededForCol` is:
```
var outputColumns []uint32
for i := range n.resultColumns {
  if n.valNeededForCol[i] {
    outputColumns = append(outputColumns, uint32(i))
  }
}
return outputColumns
```

If it's almost always true, `s/var outputColumns []uint32/outputColumns := make([]uint32, 0, len(n.resultColumns))`.

danhhz:
Was this showing up in profiling? If so, the commit message should include a benchstat of some cockroach component (not the technique) that this is expected to speed up. Otherwise, it smells a bit of premature optimization (and I would argue there's some readability losses here)

pkg/sql/distsql_physical_planner.go
richardwu: This was one of those Go things where it's slightly more efficient to do two `O(n)` scans than to `append` arbitrarily (depending on the % of elements that evaluated to true in the predicate, the higher % the more efficient since it would reduce the # of re-allocs).

This of course sacrifices readability and succinctness and perhaps is not worth the very small efficiency boost.

knz:
On 10-10-17 17:23, Richard Wu wrote:
> This of course sacrifices readability and succinctness and perhaps is
> not worth the very small efficiency boost.
> 

You can determine this by measuring and reporting the measurement.

-- 
Raphael 'kena' Poss

pkg/sql/distsql_physical_planner.go
nvanbenschoten: Do we have benchmarks to back up that it's more efficient? And if we want to avoid reallocs, just do `outputColumns := make([]uint32, 0, len(n.resultColumns)`. Either way, I'd have to agree with others that we should err on the side of readability unless changes are an all-around win (like some of these) or they are demonstrated to provide noticeable perf improvements.

richardwu: I ran a [micro-benchmark](https://gist.github.com/richardwu/a282758538a39975c3023e0e478db77c) with slices of up to 100 elements with a certain % of elements that needs to be added to the slice.

For 10% across 20 samples
```
name          time/op
Append-8      3.37µs ±14%
ExactAlloc-8  3.57µs ± 9%
MaxAlloc-8    3.31µs ±12%
```

For 50% 
```
name          time/op
Append-8      3.59µs ±27%
ExactAlloc-8  3.60µs ±10%
MaxAlloc-8    3.15µs ± 5%
```

And 90%
```
name          time/op
Append-8      4.26µs ±16%
ExactAlloc-8  3.63µs ±13%
MaxAlloc-8    3.26µs ± 9%
```

Of course, max alloc (what @nvanbenschoten suggested) is always the fastest but will almost always over-estimate. Exact alloc performs worse on lower % but better with higher %.

Perhaps for the sake of readability and speed (but sacrificing memory) it's better to always do `make([]uint32, 0, len(n.resultColumns))`. I don't have the exact stats on the predicate in the function (and it seems like it correlates to the % of columns from a table selected in a query) but I reckon it varies quite a lot.

richardwu:
RE: @danhhz so the planning phase in general is not significant to show up in my profiles. I agree that this may be a bit premature, but I've noticed that there're some instances where we repeat the less efficient iteration idiom elsewhere and I reckon it does accumulate across the board.

danhhz:
> RE: @danhhz so the planning phase in general is not significant to show up in my profiles. I agree that this may be a bit premature, but I've noticed that there're some instances where we repeat the less efficient iteration idiom elsewhere and I reckon it does accumulate across the board.

That logic makes me uncomfortable. We should not be optimizing, especially ones that compromise readability/maintainability without more proof than this. In general, I'm 👎 on commits like this given that this was the motivation

danhhz:
> planning phase in general is not significant to show up in my profiles.

Also, what profiles are you talking about here? If it's an overall profile, sure I'm not surprised that this doesn't show up. But if you write a benchmark that only runs planning, an optimization change should move the needle on either time or allocations to be merged

richardwu:
This was an overall profile, but I'll get back to you on a planning-only benchmark. I'll be sure to include numbers on future optimization changes 👍

danhhz:
Thanks! 👌

PR #19140, opened 2017-10-09
cli,config,sql: add experimental zone config SQL syntax

benesch:
Prior to this commit, most of the logic for installing a zone config lived in pkg/cli. Expose an experimental SQL API so that tests can easily manipulate zone configs without reinventing the wheel, then teach the CLI to use the SQL statements so that knowledge of how to set/lookup zone configs is encoded in only once place.

To add a new zone config:

    ALTER RANGE {default|meta|timeseries|system} EXPERIMENTAL CONFIGURE ZONE 'yaml'
    ALTER DATABASE db EXPERIMENTAL CONFIGURE ZONE 'yaml'
    ALTER TABLE [db.]tbl EXPERIMENTAL CONFIGURE ZONE 'yaml'

To delete a zone config:

    ALTER ... EXPERIMENTAL CONFIGURE ZONE NULL

To view a zone config:

    EXPERIMENTAL SHOW ZONE CONFIGURATIONS FOR RANGE {default|meta|timeseries|system}
    EXPERIMENTAL SHOW ZONE CONFIGURATIONS FOR DATABASE db
    EXPERIMENTAL SHOW ZONE CONFIGURATIONS FOR TABLE db.tbl

To list all installed zone configs:

    EXPERIMENTAL SHOW [ALL] ZONE CONFIGURATIONS
    SELECT * FROM crdb_internal.zones

This syntax is *not* intended to be permanent, but rather intended to unblock testing of partitioning (see RFC in #18683). Permanent syntax might look more like:

    ALTER ... CONFIGURE ZONE WITH num_replicas=1, gc.ttlseconds=90

Supersedes and therefore closes #19104.

----

@knz I left the `crdb_internal.zones` vtable in place to serve as the backing for `EXPERIMENTAL SHOW ALL ZONE CONFIGURATIONS`. Showing just one zone config via `EXPERIMENTAL SHOW ZONE CONFIG FOR ...` does something much more efficient. (This are a bit complicated by the fact that `SHOW ZONE CONFIG FOR...` shows the zone config that currently applies to a zone, which is not necessarily the most specific zone that could apply, while `SHOW ALL ZONE CONFIGURATIONS` decodes what's literally in `system.zones`.)


knz:
Very nice patch.

---

Reviewed 25 of 25 files at r1.

---

*[pkg/sql/plan.go, line 439 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25LR4Wq8Y3EMHQMyp:-Kw25LR4Wq8Y3EMHQMyq:b-t3bi1g) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/plan.go#L439)):*
> ```Go
> 	case *parser.ShowUsers:
> 		return p.ShowUsers(ctx, n)
> 	case *parser.ShowZoneConfig:
> ```

Add both things to prepare() underneath to also support placeholders.

---

*[pkg/sql/zone.go, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25ZXoweuIZwaGJxKw:-Kw25ZXpbBY9SrEnzM0d:b-1v9sf9) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/zone.go#L95)):*
> ```Go
> 	var yamlConfig *string
> 	{
> 		datum, err := n.yamlConfig.Eval(&params.p.evalCtx)
> ```

`if datum, err := ...; err != nil { ... } else { switch val := datum.(type) ... }`

no need for the extra pair of braces

---

*[pkg/sql/zone.go, line 161 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25nJcL2UhFsWAaUPS:-Kw25nJcL2UhFsWAaUPT:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/zone.go#L161)):*
> ```Go
> func (*setZoneConfigNode) Close(context.Context)          {}
> func (n *setZoneConfigNode) Values() parser.Datums        { return nil }
> func (n *setZoneConfigNode) FastPathResults() (int, bool) { return n.numAffected, true }
> ```

:+1:

---

*[pkg/sql/zone.go, line 221 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25zXPeMw9cKY1dLyT:-Kw25zXPeMw9cKY1dLyU:bqzo53k) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/zone.go#L221)):*
> ```Go
> 	{
> 		// Determine the CLI specifier for the zone config that actually applies
> 		// without performing another KV lookup.
> ```

Separate function underneath perhaps?

---

*[pkg/sql/parser/sql.y, line 1056 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw24kBUv3Zh7rMqpmpT:-Kw24kBVNxap-3u2qkXa:b-56kpru) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/parser/sql.y#L1056)):*
> ```Yacc
> 
> alter_zone_range_stmt:
>   ALTER RANGE DEFAULT EXPERIMENTAL CONFIGURE ZONE a_expr_const
> ```

Make the third thing a `ctext_expr`, so you can merge the two non-terminals into one, then use the same logic as SET so as to also support placeholders and recognize identifiers as strings.

---

*[pkg/sql/parser/sql.y, line 2445 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw259B_qtxC8RJN7SXG:-Kw259B_qtxC8RJN7SXH:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/parser/sql.y#L2445)):*
> ```Yacc
> 
> show_zone_stmt:
>   EXPERIMENTAL SHOW ZONE CONFIGURATION FOR RANGE name
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-Kw26F0AsbV2tl4L-viw:b-7q9a64)*

benesch:
Ok, RFAL!

---

Review status: 23 of 25 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/plan.go, line 439 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25LR4Wq8Y3EMHQMyp:-KwB4ei8OY32tPc5obKC:b-lzknvo) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/plan.go#L439)):*

Add both things to prepare() underneath to also support placeholders.
</blockquote></details>

(See below.)

---

*[pkg/sql/zone.go, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25ZXoweuIZwaGJxKw:-KwAoRfsEG96a-QtDZlg:b-qtwu96) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/zone.go#L95)):*

`if datum, err := ...; err != nil { ... } else { switch val := datum.(type) ... }`

no need for the extra pair of braces
</blockquote></details>

I agree with you, but go lint does not:

```
sql/zone.go:96:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
```

I removed the extra scope anyway, since leaking `datum` isn't the end of the world.

---

*[pkg/sql/zone.go, line 221 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25zXPeMw9cKY1dLyT:-KwAw6IyYMT_P1ZBN1ql:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/zone.go#L221)):*

Separate function underneath perhaps?
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 1056 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw24kBUv3Zh7rMqpmpT:-KwB3K5gGtcwBr_J-S6Q:b3404g1) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/parser/sql.y#L1056)):*

Make the third thing a `ctext_expr`, so you can merge the two non-terminals into one, then use the same logic as SET so as to also support placeholders and recognize identifiers as strings.
</blockquote></details>

How would you feel about holding off on this for now? :( I tried going down this path, but it forces a fairly substantial refactor to thread a `TypedExpr` through planning, and probably warrants a separate `ZoneSpecifier` struct for the `config` package that doesn't have the `TypedExpr` baggage. I'm not sure it's worth it given that `ALTER TABLE` and `ALTER DATABASE` won't support placeholders and the syntax is still experimental.

I could also extract a new production for the range name:

```
range_name: DEFAULT | name
```

---

*[pkg/sql/parser/sql.y, line 2445 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw259B_qtxC8RJN7SXG:-KwB3lcVUxvM7BDPEs0v:b5erico) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/parser/sql.y#L2445)):*

ditto
</blockquote></details>

(see above.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-KwB4gBWnJC9iofA0FtF:bkqtryl)*

knz:
I think I'm OK with this. However for the future I would recommend that you keep the original PR open and simply redirect the branch to the new patch. This way the reivew conversations are properly preserved.

---

Reviewed 2 of 2 files at r2.

---

*[pkg/sql/zone.go, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw25ZXoweuIZwaGJxKw:-KwBEeix7BP7HN6DOrdz:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/zone.go#L95)):*

I agree with you, but go lint does not:

```
sql/zone.go:96:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
```

I removed the extra scope anyway, since leaking `datum` isn't the end of the world.
</blockquote></details>

:+1:

---

*[pkg/sql/parser/sql.y, line 1056 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw24kBUv3Zh7rMqpmpT:-KwBElDQ6oYR9bWzxbte:byz8fho) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/parser/sql.y#L1056)):*

How would you feel about holding off on this for now? :( I tried going down this path, but it forces a fairly substantial refactor to thread a `TypedExpr` through planning, and probably warrants a separate `ZoneSpecifier` struct for the `config` package that doesn't have the `TypedExpr` baggage. I'm not sure it's worth it given that `ALTER TABLE` and `ALTER DATABASE` won't support placeholders and the syntax is still experimental.

I could also extract a new production for the range name:

```
range_name: DEFAULT | name
```
</blockquote></details>

All right, carry on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-KwBEsqXugXClQ_AaNJj:b6z8lx8)*

danhhz:
I didn't quite get through everything, but here's what I have so far

---

Review status: 24 of 25 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/cli/zone.go, line 32 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBSq2SXRlhXpLEG54u:-KwBSq2SXRlhXpLEG54v:b5tkyte) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L32)):*
> ```Go
> func queryZoneSpecifiers(conn *sqlConn) ([]string, error) {
> 	rows, err := makeQuery(
> 		`SELECT cli_specifier FROM [EXPERIMENTAL SHOW ZONE CONFIGURATIONS] ORDER BY 1`)(conn)
> ```

nit: `ORDER BY cli_specifier` is easier to read

---

*[pkg/cli/zone.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBTQFaQ232Yn0VMn_P:-KwBTQFaQ232Yn0VMn_Q:bgslsy0) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L93)):*
> ```Go
> 	}
> 
> 	cliSpecifier, ok := vals[0].(string)
> ```

why not QueryRow.Scan? ditto elsewhere

---

*[pkg/config/config.go, line 31 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBVEoXr0NpfGeev_44:-KwBVEoXr0NpfGeev_45:b-x32fcg) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config.go#L31)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/keys"
> 	"github.com/cockroachdb/cockroach/pkg/roachpb"
> 	"github.com/cockroachdb/cockroach/pkg/sql/parser"
> ```

dunno if there is anything to be done about it, but adding this dependency to pkg/config is unfortunate

---

*[pkg/config/config_test.go, line 18 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBW98hPKduiuOkBZFE:-KwBW98hPKduiuOkBZFF:b3kre) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L18)):*
> ```Go
> 
> import (
> 	fmt "fmt"
> ```

?

---

*[pkg/config/config_test.go, line 537 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBXbxjQYqA7iko4Mz_:-KwBXbxjQYqA7iko4Mza:b-krgusa) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L537)):*
> ```Go
> }
> 
> func TestZoneSpecifiers(t *testing.T) {
> ```

great tests! :ok_hand: 

---

*[pkg/config/config_test.go, line 540 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBWV0HaN19u9pBlSCz:-KwBWV0HaN19u9pBlSD-:bdlbf1i) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L540)):*
> ```Go
> 
> 	// Simulate exactly two named zones: one named default and one named carl.
> ```

why do we need to swap it for the test? can't the normal entries be used?

---

*[pkg/config/config_test.go, line 554 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBWpvzxLSqpTHeyzCT:-KwBWpvzxLSqpTHeyzCU:b8aod79) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L554)):*
> ```Go
> 	}
> 
> 	// Simulate the following schema:
> ```

I think this comment is out of date compared to below

---

*[pkg/config/config_test.go, line 558 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBX1tlQ-ruA7dB5lN3:-KwBX1tlQ-ruA7dB5lN4:b-nqz76t) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L558)):*
> ```Go
> 	//   CREATE TABLE db.table...
> 	//   CREATE DATABASE "."
> 	//   CREATE TABLE ".table."
> ```

it's a little unclear which database each of these tables is being created in. you could fully specify the CREATE TABLEs?

---

*[pkg/config/config_test.go, line 626 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBXOxxazfR1bwY8y4t:-KwBXOxxazfR1bwY8y4u:bjog0fd) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L626)):*
> ```Go
> 			}
> 			if e, a := tc.id, int(id); !reflect.DeepEqual(e, a) {
> 				t.Errorf("#%d: path did not match expected path: %s\n",
> ```

may as well make these subtests while you're in here

---

*[pkg/sql/zone.go, line 164 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBZZ97H-svQ4H9KLqh:-KwBZZ97H-svQ4H9KLqi:b472dzh) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/zone.go#L164)):*
> ```Go
> 	proto, _, err := GetZoneConfigInTxn(params.ctx, params.p.txn, uint32(targetID))
> 	if err == errNoZoneConfigApplies {
> 		// TODO(benesch): Handling this case shouldn't be the caller's
> ```

can you expand this TODO? I don't understand what is actually to be done here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-KwBaYNjQOmNzKLYrjCw:b-b3p41t)*

benesch:
Review status: 17 of 27 files reviewed at latest revision, 11 unresolved discussions.

---

*[pkg/cli/zone.go, line 32 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBSq2SXRlhXpLEG54u:-KwBey9Q6nNx6-u_xipS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L32)):*

nit: `ORDER BY cli_specifier` is easier to read
</blockquote></details>

Done.

---

*[pkg/cli/zone.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBTQFaQ232Yn0VMn_P:-KwBbUD4i1G8fAv_fGBn:b-6pfjxd) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L93)):*

why not QueryRow.Scan? ditto elsewhere
</blockquote></details>

`conn.QueryRow` returns `[]driver.Value`, not `db.Row`. I won't pretend to understand why.

---

*[pkg/config/config.go, line 31 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBVEoXr0NpfGeev_44:-KwBepicu1TDTgXRBN6t:bb6sh2i) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config.go#L31)):*

dunno if there is anything to be done about it, but adding this dependency to pkg/config is unfortunate
</blockquote></details>

Ack.

---

*[pkg/config/config_test.go, line 18 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBW98hPKduiuOkBZFE:-KwBbePDL3nZnlTsh0Eb:b-9zjd9r) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L18)):*

?
</blockquote></details>

Um, huh. Thanks, goimports.

Fixed, thanks!

---

*[pkg/config/config_test.go, line 540 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBWV0HaN19u9pBlSCz:-KwBccPllWE1nW9NsbeC:bo3pvnf) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L540)):*

why do we need to swap it for the test? can't the normal entries be used?
</blockquote></details>

I *guess*, but I dislike having the tests depend on the actual values in the map, which could plausibly change over time.

---

*[pkg/config/config_test.go, line 554 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBWpvzxLSqpTHeyzCT:-KwBcYzn5mn1PsDHUiV8:bt2vr1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L554)):*

I think this comment is out of date compared to below
</blockquote></details>

Fixed.

---

*[pkg/config/config_test.go, line 558 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBX1tlQ-ruA7dB5lN3:-KwBcZqkM1h0tFKZlrnf:b-oc4vn3) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L558)):*

it's a little unclear which database each of these tables is being created in. you could fully specify the CREATE TABLEs?
</blockquote></details>

Oops, good point. Done.

---

*[pkg/config/config_test.go, line 626 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBXOxxazfR1bwY8y4t:-KwBeSShHcv5z9nrNHw0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config_test.go#L626)):*

may as well make these subtests while you're in here
</blockquote></details>

Done.

---

*[pkg/sql/zone.go, line 164 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBZZ97H-svQ4H9KLqh:-KwBcuI6Qkn0EgYkBsDq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/zone.go#L164)):*

can you expand this TODO? I don't understand what is actually to be done here
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 1056 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-Kw24kBUv3Zh7rMqpmpT:-KwBcv6Q-vo26ODTEHh0:bh6gtfo) ([raw file](https://github.com/cockroachdb/cockroach/blob/b41695d08431bc8cde1cd2d36d398b6076ce1c06/pkg/sql/parser/sql.y#L1056)):*

All right, carry on.
</blockquote></details>

Aha! There's already a production for that: `unrestricted_name`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140)*

a-robinson:
Nice work! My eyes are starting to glaze over at how much sheer code there is, so thanks for the detailed tests.

---

Reviewed 22 of 25 files at r1, 1 of 2 files at r2, 3 of 10 files at r4.
Review status: 20 of 27 files reviewed at latest revision, 19 unresolved discussions, some commit checks pending.

---

*[pkg/cli/zone.go, line 244 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBS-OmYeSSvTMSYmOH:-KwBS-OnxWWeH54akJEs:b-uij4ud) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L244)):*
> ```Go
> 
> 	err = conn.ExecTxn(func(conn *sqlConn) error {
> 		if err := conn.Exec(fmt.Sprintf(`ALTER %s EXPERIMENTAL CONFIGURE ZONE %s`, zs,
> ```

Does there not need to be a space after the last `%s`? Weird.

---

*[pkg/config/config.go, line 31 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBVEoXr0NpfGeev_44:-KwBbB02yNBEd2P6e5C6:b32v57) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config.go#L31)):*

dunno if there is anything to be done about it, but adding this dependency to pkg/config is unfortunate
</blockquote></details>

+1

---

*[pkg/config/config.go, line 36 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBbD_sX4lZuwsMFfvI:-KwBbD_sX4lZuwsMFfvJ:b5nz42u) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config.go#L36)):*
> ```Go
> )
> 
> // Several ranges outside of the SQL keyspace are given special names so they
> ```

I'm mostly ignoring this file under the assumption it hasn't meaningfully changed since #19104. Let me know if that isn't the case.

---

*[pkg/sql/config.go, line 45 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBcw5jkXpzHbfYTVM6:-KwBcw5jkXpzHbfYTVM7:b-fpoj4d) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/config.go#L45)):*
> ```Go
> 	// Look in the zones table.
> 	if zoneVal, err := get(sqlbase.MakeZoneKey(sqlbase.ID(id))); err != nil {
> 		return config.ZoneConfig{}, 0, err
> ```

I'm not a huge fan of returning 0 as a way of indicating that we aren't referring to a zone config, since 0 is a valid zone config ID (for ".default")

---

*[pkg/sql/config.go, line 52 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBcQiokQ_KV27EgbKm:-KwBcQiokQ_KV27EgbKn:b-3ycphw) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/config.go#L52)):*
> ```Go
> 	}
> 
> 	// No zone config for this ID. We need to figure out if it's a database
> ```

Not your doing, but this is really only checking if it's a table. The comment in the `else if` is also completely redundant with this one and should be removed.

---

*[pkg/sql/crdb_internal.go, line 1353 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBdmIivEDFodV3jcMP:-KwBdmIivEDFodV3jcMQ:bq95wx5) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1353)):*
> ```Go
> }
> 
> // crdbInternalZonesTable exposes the zone config overrides in the system.zones
> ```

What are "the zone config overrides"? This comment may need rewording.

---

*[pkg/sql/crdb_internal.go, line 1383 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBeGddUmiGWOhEPYLW:-KwBeGddUmiGWOhEPYLX:bb1kxo0) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1383)):*
> ```Go
> 			id := uint32(parser.MustBeDInt(r[0]))
> 			var cliSpecifier parser.Datum = parser.DNull
> 			if zs, err := config.ZoneSpecifierFromID(id, resolveID); err == nil {
> ```

I assume this is to make the internal table resilient to differences between the actual stored `system.zones` contents and the cached version on this node. Are all users of `crdb_internal.zones` resilient to the `cli_specifier` being missing?

---

*[pkg/sql/zone.go, line 164 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBZZ97H-svQ4H9KLqh:-KwBhE4nn1t44XXNansu:bt10s8b) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/zone.go#L164)):*

Done.
</blockquote></details>

This also might be incorrect as written, right? If a transaction modifies the default zone config, then sets a different zone config, would it grab the old or the new version of the default zone config?

---

*[pkg/sql/zone.go, line 36 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBfccSgw4oBmh84rTS:-KwBfccSgw4oBmh84rTT:b-ibj3y8) ([raw file](https://github.com/cockroachdb/cockroach/blob/de349ce973d64fd40fcf77cd7845a90c8315a8ee/pkg/sql/zone.go#L36)):*
> ```Go
> 	if zs.NamedZone != "" {
> 		return pgerror.NewErrorf(
> 			pgerror.CodeInvalidCatalogNameError, "range %q does not exist", zs.NamedZone)
> ```

s/range/zone?

---

*[pkg/sql/zone.go, line 91 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBgUnv3V6i4jst0dKU:-KwBgUnv3V6i4jst0dKV:b-t3lhm4) ([raw file](https://github.com/cockroachdb/cockroach/blob/de349ce973d64fd40fcf77cd7845a90c8315a8ee/pkg/sql/zone.go#L91)):*
> ```Go
> 		// default zone config.
> 		zs.NamedZone = config.DefaultZoneName
> 	} else if resolvedID != uint32(actualID) {
> ```

What is this cast for?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-KwBhttJs4BI4RlWivDI:b-ni4w06)*

benesch:
Yeah, sorry this ended up being such a behemoth. Couldn't figure out a good way of splitting it up.

---

Review status: 16 of 27 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[pkg/cli/zone.go, line 244 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBS-OmYeSSvTMSYmOH:-KwBlSZJZiXYWEO8DbJ-:b4ak851) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L244)):*

Does there not need to be a space after the last `%s`? Weird.
</blockquote></details>

Not sure I follow. This is what I'm intending to produce:

```
ALTER {RANGE|DATABASE|TABLE} EXPERIMENTAL CONFIGURE ZONE 'foo'
```

---

*[pkg/config/config.go, line 36 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBbD_sX4lZuwsMFfvI:-KwBjQK9PhXjveQShiQw:b-cekuil) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/config/config.go#L36)):*

I'm mostly ignoring this file under the assumption it hasn't meaningfully changed since #19104. Let me know if that isn't the case.
</blockquote></details>

The added functionality is nearly equivalent, but the implementation changed. :/ Specifically, I no longer introduce a separate `config.ZoneSpecifier` type, since converting between a `parser.ZoneSpecifier` and a `config.ZoneSpecifier` resulted in a lot of wasted effort for no real benefit.

---

*[pkg/sql/config.go, line 45 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBcw5jkXpzHbfYTVM6:-KwBkDX-YTqyV7YeLWK2:b2tbaxt) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/config.go#L45)):*

I'm not a huge fan of returning 0 as a way of indicating that we aren't referring to a zone config, since 0 is a valid zone config ID (for ".default")
</blockquote></details>

Me neither. I guess I could use an `int64` and return -1? But then you'd need to cast back to a uint32 everywhere. I'm not convinced this is much worse than returning `config.ZoneConfig{}`. That's _also_ not valid.

---

*[pkg/sql/config.go, line 52 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBcQiokQ_KV27EgbKm:-KwBmCEMZA69pJHdiTbR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/config.go#L52)):*

Not your doing, but this is really only checking if it's a table. The comment in the `else if` is also completely redundant with this one and should be removed.
</blockquote></details>

Done.

---

*[pkg/sql/crdb_internal.go, line 1353 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBdmIivEDFodV3jcMP:-KwBmSQlYewlKunRsLnn:bt2vr1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1353)):*

What are "the zone config overrides"? This comment may need rewording.
</blockquote></details>

Fixed.

---

*[pkg/sql/crdb_internal.go, line 1383 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBeGddUmiGWOhEPYLW:-KwBkXntl6VjJThmoPrd:b-h8zeqb) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1383)):*

I assume this is to make the internal table resilient to differences between the actual stored `system.zones` contents and the cached version on this node. Are all users of `crdb_internal.zones` resilient to the `cli_specifier` being missing?
</blockquote></details>

This reads out of `system.zones` directly, so it should never hit the cache. The problem is that someone can run `DELETE FROM system.zones WHERE id = 0` and remove the `.default` entry; the zone config applies is then the one baked into the binary (`config.DefaultZoneConfig()`), but that one has no CLI specifier. Arcane, I know.

---

*[pkg/sql/zone.go, line 164 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBZZ97H-svQ4H9KLqh:-KwBxX-Uax4KDOV53isZ:b2t02cf) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/zone.go#L164)):*

This also might be incorrect as written, right? If a transaction modifies the default zone config, then sets a different zone config, would it grab the old or the new version of the default zone config?
</blockquote></details>

This should be taken care of by `GetZoneConfigInTxn`, which specifically doesn't use the system config. I added a test to verify, and it seems happy.

---

*[pkg/sql/zone.go, line 36 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBfccSgw4oBmh84rTS:-KwBmXIkvysvkYgic2cr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/de349ce973d64fd40fcf77cd7845a90c8315a8ee/pkg/sql/zone.go#L36)):*

s/range/zone?
</blockquote></details>

Done.

---

*[pkg/sql/zone.go, line 91 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBgUnv3V6i4jst0dKU:-KwBl3wfezwxu1s-yeVD:b-limjqt) ([raw file](https://github.com/cockroachdb/cockroach/blob/de349ce973d64fd40fcf77cd7845a90c8315a8ee/pkg/sql/zone.go#L91)):*

What is this cast for?
</blockquote></details>

Oops, copied and pasted from somewhere where `actualID` was a `sqlbase.ID`. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-KwBxYeZ64-tT2RMYASa:bo1xvtx)*

a-robinson:
:lgtm:

---

Reviewed 1 of 10 files at r4, 1 of 1 files at r5, 4 of 4 files at r6.
Review status: 22 of 27 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[pkg/cli/zone.go, line 244 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBS-OmYeSSvTMSYmOH:-KwGZbDXu9bcDWUEAEG9:b-5yyxjf) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/cli/zone.go#L244)):*

Not sure I follow. This is what I'm intending to produce:

```
ALTER {RANGE|DATABASE|TABLE} EXPERIMENTAL CONFIGURE ZONE 'foo'
```
</blockquote></details>

Apparently I didn't follow -- I thought concatenation was happening with the stuff below. Disregard.

---

*[pkg/sql/config.go, line 45 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBcw5jkXpzHbfYTVM6:-KwG_1_AJfEeusNfebVx:bm6rgr) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/config.go#L45)):*

Me neither. I guess I could use an `int64` and return -1? But then you'd need to cast back to a uint32 everywhere. I'm not convinced this is much worse than returning `config.ZoneConfig{}`. That's _also_ not valid.
</blockquote></details>

Yeah, you're right that the return value makes this hard. Not worth changing, and I suppose not even worth leaving a comment about given the convention that non-error return values are only valid if the returned error is nil.

---

*[pkg/sql/crdb_internal.go, line 1383 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBeGddUmiGWOhEPYLW:-KwGaO57MdyA8F2gGs-0:bhi8rh1) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1383)):*

This reads out of `system.zones` directly, so it should never hit the cache. The problem is that someone can run `DELETE FROM system.zones WHERE id = 0` and remove the `.default` entry; the zone config applies is then the one baked into the binary (`config.DefaultZoneConfig()`), but that one has no CLI specifier. Arcane, I know.
</blockquote></details>

I think I'm more lost than I was before, which is saying something, since I'm now not sure why I thought the cache was coming into play here.

1. Where is the filling in of the baked-in default zone config happening?
2. Since it looks like there could be times when we don't set the CLI specifier (in case of an error), are all users of `crdb_internal.zones` resilient to the `cli_specifier` being missing?

---

*[pkg/sql/zone.go, line 164 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBZZ97H-svQ4H9KLqh:-KwGbbYNwPFFjlE2GqlK:b44tpx) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/zone.go#L164)):*

This should be taken care of by `GetZoneConfigInTxn`, which specifically doesn't use the system config. I added a test to verify, and it seems happy.
</blockquote></details>

👍

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-KwGblIi_ty-OiofWNcx:bnfp4nl)*

benesch:
Review status: 22 of 27 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[pkg/sql/crdb_internal.go, line 1383 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBeGddUmiGWOhEPYLW:-KwGtIxoUVaniG99K-WR:b-rly1qk) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1383)):*

I think I'm more lost than I was before, which is saying something, since I'm now not sure why I thought the cache was coming into play here.

1. Where is the filling in of the baked-in default zone config happening?
2. Since it looks like there could be times when we don't set the CLI specifier (in case of an error), are all users of `crdb_internal.zones` resilient to the `cli_specifier` being missing?
</blockquote></details>

Oof, you're right, I'm very confused. `err == nil` only in the case where `system.namespace` and `system.zones` are inconsistent.

Here's what I propose: I'll propagate the error instead of returning `NULL` for now. I'll file an issue about making the whole apparatus more resilient to `system.zones` having IDs that are not present in `system.namespace`. (`DROP TABLE` was causing an inconsistency like this for a while.) You're totally right that, at present, the CLI might crash if we return a `NULL`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140)*

knz:
:lgtm:

---

Reviewed 8 of 10 files at r4, 1 of 1 files at r5, 4 of 4 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-Kw_SrFSl8JIRG5-GZgO:bnfp4nl)*

benesch:
Thanks for the mammoth reviews, everyone! Merging on green.

---

Review status: 26 of 27 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[pkg/sql/crdb_internal.go, line 1383 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-KwBeGddUmiGWOhEPYLW:-Kw_rEy7vSMfmVe5aqm0:b-896fja) ([raw file](https://github.com/cockroachdb/cockroach/blob/fb219836522bac4cecb83a9b9f7141f957cb9b48/pkg/sql/crdb_internal.go#L1383)):*

Oof, you're right, I'm very confused. `err == nil` only in the case where `system.namespace` and `system.zones` are inconsistent.

Here's what I propose: I'll propagate the error instead of returning `NULL` for now. I'll file an issue about making the whole apparatus more resilient to `system.zones` having IDs that are not present in `system.namespace`. (`DROP TABLE` was causing an inconsistency like this for a while.) You're totally right that, at present, the CLI might crash if we return a `NULL`.


</blockquote></details>

Done!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19140#-:-Kw_tzBXMRVXF-azno2B:br7tpt)*

PR #19143, opened 2017-10-09
cherry-pick 1.1: sql: don't enable "kv tracing" for SHOW TRACE FOR <stmt>

andreimatei:
Cherry-pick of selections from #19008. The whole thing did not apply cleanly/was unnecessary because some of the logging messages it dealt with are not present on this branch.

CC @cockroachdb/release 

A "kv trace" is the a set of log messages around SQL<->KV interaction.
The concept is not very well defined at the moment, but it comprises two
categories of messages:
1. Details of KV requests being sent for execution, in particular
pretty-printed scan requests.
2. Per-row messages. In particular, every row that's being returned by
KV for processing by SQL.

These messages are the only ones that are returned as results of
SHOW _KV_ TRACE FOR <stmt>/SESSION.
The situation gets interesting for SHOW TRACE FOR ... (no KV specified).
In this case, before this patch, all the "kv trace" messages were
included in the results. This patch changes that and eliminates the
per-row messages. This is an improvement, per-row messages (as opposed
to per-RPC or higher level) is too verbose for most people that want
traces.

This was already the case before this patch, but now this comes to the
foreground: "kv tracing" means one thing externally (as in SHOW KV TRACE
FOR ...), versus internally, in the code, where "kvTracedEnabled" and
similarly-named members refer to "per-row" messages. A round of renames
is in order, but that's left for another day.

This patch also introduces log.ExpensiveLogEnabled(), a replacement for
log.V() which also takes tracing into consideration.

### DIFF
--- a/pkg/sql/bench_test.go
+++ b/pkg/sql/bench_test.go
@@ -989,3 +989,42 @@ func BenchmarkWideTable(b *testing.B) {
 		})
 	})
 }
+
+// BenchmarkIndexJoin measure an index-join with 1000 rows.
+func BenchmarkIndexJoin(b *testing.B) {
+	// The table will have an extra column not contained in the index to force a
+	// join with the PK.
+	create := `
+		 CREATE TABLE tidx (
+				 k INT NOT NULL,
+				 v INT NULL,
+				 extra STRING NULL,
+				 CONSTRAINT "primary" PRIMARY KEY (k ASC),
+				 INDEX idx (v ASC),
+				 FAMILY "primary" (k, v, extra)
+		 )
+		`
+	// We'll insert 1000 rows with random values below 1000 in the index. We'll
+	// then query the index with a query that retrieves all the data (but the
+	// optimizer doesn't know that).
+	insert := "insert into tidx(k,v) select generate_series(1,1000), (random()*1000)::int"
+	s, db, _ := serverutils.StartServer(b, base.TestServerArgs{UseDatabase: "bench"})
+	defer s.Stopper().Stop(context.TODO())
+
+	if _, err := db.Exec(`CREATE DATABASE bench`); err != nil {
+		b.Fatal(err)
+	}
+	if _, err := db.Exec(create); err != nil {
+		b.Fatal(err)
+	}
+	if _, err := db.Exec(insert); err != nil {
+		b.Fatal(err)
+	}
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		if _, err := db.Exec("select * from tidx where v < 1000"); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
--- a/pkg/sql/logictest/testdata/logic_test/show_trace
+++ b/pkg/sql/logictest/testdata/logic_test/show_trace
@@ -31,7 +31,6 @@ SELECT span, message, operation FROM [SHOW TRACE FOR SELECT 1]
 (0,0)  === SPAN START: sql txn implicit ===  sql txn implicit
 (0,1)  === SPAN START: starting plan ===     starting plan
 (0,2)  === SPAN START: consuming rows ===    consuming rows
-(0,2)  output row: [1]                       consuming rows
 (0,2)  plan completed execution              consuming rows
 (0,2)  resources released, stopping trace    consuming rows
 
@@ -44,7 +43,6 @@ SELECT span, message, operation FROM [SHOW TRACE FOR EXECUTE x]
 (0,0)  === SPAN START: sql txn implicit ===  sql txn implicit
 (0,1)  === SPAN START: starting plan ===     starting plan
 (0,2)  === SPAN START: consuming rows ===    consuming rows
-(0,2)  output row: [1]                       consuming rows
 (0,2)  plan completed execution              consuming rows
 (0,2)  resources released, stopping trace    consuming rows
 
--- a/pkg/sql/session.go
+++ b/pkg/sql/session.go
@@ -1419,7 +1419,12 @@ type SessionTracing struct {
 // span that have already been created will not be traced).
 //
 // StopTracing() needs to be called to finish this trace.
-func (st *SessionTracing) StartTracing(recType tracing.RecordingType, traceKV bool) error {
+//
+// Args:
+// kvTracingEnabled: If set, the traces will also include "KV trace" messages -
+//   verbose messages around the interaction of SQL with KV. Some of the messages
+//   are per-row.
+func (st *SessionTracing) StartTracing(recType tracing.RecordingType, kvTracingEnabled bool) error {
 	if st.enabled {
 		return errors.Errorf("already tracing")
 	}
@@ -1433,7 +1438,7 @@ func (st *SessionTracing) StartTracing(recType tracing.RecordingType, traceKV bo
 
 	tracing.StartRecording(sp, recType)
 	st.enabled = true
-	st.kvTracingEnabled = traceKV
+	st.kvTracingEnabled = kvTracingEnabled
 	st.recordingType = recType
 	return nil
 }
--- a/pkg/sql/show.go
+++ b/pkg/sql/show.go
@@ -329,7 +329,7 @@ WHERE message LIKE 'fetched: %'
 		plan.Close(ctx)
 		return nil, err
 	}
-	tracePlan, err := p.makeTraceNode(stmtPlan)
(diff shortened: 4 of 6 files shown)


knz:
:lgtm:

---

Reviewed 1 of 1 files at r1, 4 of 4 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-:-Kw4k7wmroxiAxhbpfc1:bnfp4nl)*

bdarnell:

---

*[pkg/sql/bench_test.go, line 994 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-KwSPXvS7hWEC2YzL0N9:-KwSPXvS7hWEC2YzL0NA:boyi2wp) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f02d97303409ebc56798a86855e5bf56d02d63f/pkg/sql/bench_test.go#L994)):*
> ```Go
> 
> // BenchmarkIndexJoin measure an index-join with 1000 rows.
> func BenchmarkIndexJoin(b *testing.B) {
> ```

No need to remove it now but this benchmark should not have been included in the cherry-pick. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143)*

andreimatei:

---

*[pkg/sql/bench_test.go, line 994 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-KwSPXvS7hWEC2YzL0N9:-KwVvIi6OWNJe87Y-p8m:b-zgnkmd) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f02d97303409ebc56798a86855e5bf56d02d63f/pkg/sql/bench_test.go#L994)):*

No need to remove it now but this benchmark should not have been included in the cherry-pick. 
</blockquote></details>

Why not? Shouldn't we encourage test changes to be cherry-picked? Like, none of the drawbacks of cherry-picking apply and all of the upshots apply.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143)*

bdarnell:

---

*[pkg/sql/bench_test.go, line 994 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-KwSPXvS7hWEC2YzL0N9:-KwVvrFq2O58Sm8jeVbi:bzi3uas) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f02d97303409ebc56798a86855e5bf56d02d63f/pkg/sql/bench_test.go#L994)):*

Why not? Shouldn't we encourage test changes to be cherry-picked? Like, none of the drawbacks of cherry-picking apply and all of the upshots apply.
</blockquote></details>

When a code change is cherry-picked, of course the corresponding tests should come with it. But otherwise, test changes need to be justified like any other cherry-pick (in their own PR, not tacked on to something unrelated). My perspective is the opposite of yours: test cherry-picks have a significant drawback in that they have some non-zero chance of introducing flakiness in the release build, and they have relatively little upside because they're getting run on master too. We should only cherry-pick test-only changes if they either reduce flakiness or test a condition which we reasonably believe may differ between the release branch and master at the time the test was introduced. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143)*

andreimatei:

---

*[pkg/sql/bench_test.go, line 994 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-KwSPXvS7hWEC2YzL0N9:-KwVzWG70cs0ia3_Nnb1:b-pnptv2) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f02d97303409ebc56798a86855e5bf56d02d63f/pkg/sql/bench_test.go#L994)):*

When a code change is cherry-picked, of course the corresponding tests should come with it. But otherwise, test changes need to be justified like any other cherry-pick (in their own PR, not tacked on to something unrelated). My perspective is the opposite of yours: test cherry-picks have a significant drawback in that they have some non-zero chance of introducing flakiness in the release build, and they have relatively little upside because they're getting run on master too. We should only cherry-pick test-only changes if they either reduce flakiness or test a condition which we reasonably believe may differ between the release branch and master at the time the test was introduced. 
</blockquote></details>

Well, but, it is my impression (and definitely my personal practice) that we do cherry-picks mostly at the level of a PR. So, it's extra work to... cherry-pick within the PR - both initially when cherry-picking and subsequently when looking at cherry-picks and trying to understand what exactly has been cherry-picked and what hasn't.
Also, as a matter of principle, if a test is flaky, it might be that the test is bad and the code is bad. The former may be more frequent, but the latter is more serious. So, assuming that the cherry-picked tests are supposed to work on the branch that we cherry-picked them on, but they prove to be flaky, we'd like to know, wouldn't we? Now, I'm not saying we should explicitly cherry-pick new tests (ain't nobody got time for that), but when it's easier to do cherry-pick them (like, when they're in a PR that we otherwise cherry-pick), it seems to me like a good thing to do.

In this particular case, you can debate whether the benchmark should have been part of the PR or not. It was, because it came as a result of a conversation we were having on the PR and Raphael for some reason put his foot down and said "don't do the following commit if you don't show me that there's no regression on this new benchmark". And so, technically, if someone were to observe that on the 1.1 branch there's a regression to the benchmark after the next cherry-picked commit, we'd again want to know. (In reality, I haven't actually compared the benchmark on 1.1 before/after the subsequent commit, but in principle one could).


Does this moderate your stance? (I'll take an unqualified "no" :) )

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143)*

bdarnell:

---

*[pkg/sql/bench_test.go, line 994 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-KwSPXvS7hWEC2YzL0N9:-KwW1jJk6QeTanGCN2fS:b-n3yxe) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f02d97303409ebc56798a86855e5bf56d02d63f/pkg/sql/bench_test.go#L994)):*

Well, but, it is my impression (and definitely my personal practice) that we do cherry-picks mostly at the level of a PR. So, it's extra work to... cherry-pick within the PR - both initially when cherry-picking and subsequently when looking at cherry-picks and trying to understand what exactly has been cherry-picked and what hasn't.
Also, as a matter of principle, if a test is flaky, it might be that the test is bad and the code is bad. The former may be more frequent, but the latter is more serious. So, assuming that the cherry-picked tests are supposed to work on the branch that we cherry-picked them on, but they prove to be flaky, we'd like to know, wouldn't we? Now, I'm not saying we should explicitly cherry-pick new tests (ain't nobody got time for that), but when it's easier to do cherry-pick them (like, when they're in a PR that we otherwise cherry-pick), it seems to me like a good thing to do.

In this particular case, you can debate whether the benchmark should have been part of the PR or not. It was, because it came as a result of a conversation we were having on the PR and Raphael for some reason put his foot down and said "don't do the following commit if you don't show me that there's no regression on this new benchmark". And so, technically, if someone were to observe that on the 1.1 branch there's a regression to the benchmark after the next cherry-picked commit, we'd again want to know. (In reality, I haven't actually compared the benchmark on 1.1 before/after the subsequent commit, but in principle one could).


Does this moderate your stance? (I'll take an unqualified "no" :) )
</blockquote></details>

Yeah, if this PR was all of #19008 then I'd be fine with it. But since you're already only taking part of the PR (and didn't include the results of this benchmark in the commit message as you did in the original PR), I would have dropped this commit too. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143)*

andreimatei:

---

*[pkg/sql/bench_test.go, line 994 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143#-KwSPXvS7hWEC2YzL0N9:-KwW3RKxtovKi7wV6B2i:b-3dlml8) ([raw file](https://github.com/cockroachdb/cockroach/blob/5f02d97303409ebc56798a86855e5bf56d02d63f/pkg/sql/bench_test.go#L994)):*

Yeah, if this PR was all of #19008 then I'd be fine with it. But since you're already only taking part of the PR (and didn't include the results of this benchmark in the commit message as you did in the original PR), I would have dropped this commit too. 
</blockquote></details>

Fair enough. I considered being honest here that I have in fact cherry-picked at a low level in this case, but my message was already too long :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19143)*

PR #19144, opened 2017-10-10
sql: redefine NaN = NaN -> true

### LINKED ISSUE #18860: sql: NaN = NaN (in Postgres)

nvb:
Fixes #18860.

This change switches `NaN` equality semantics from defining `NaN` as not-equal
to itself to defining it as equal to itself. This is a very sad violation of
IEEE754 in the name of Postgres compatibility.

Also, we previously defined all comparisons other than `NaN = NaN` as returning
false. This change switches this so that `NaN` acts like a normal value that
is less than all other values. This is needed so that comparison with `NaN`
is consistent with the ordering of `NaN` in indexes.


madelynnblue (commented):
Could you add details to the commit message about what the new intended behavior is? I can mostly tease it out from the code, but I think some prose would be nice commit log documentation.

nvanbenschoten:
Done. I added a second commit that makes `NaN`s compare like they do in indexes.

madelynnblue (approved):

knz:
Please clarify, and test, the relative order of NaN and NULL.
Also see my comment below about testing the consistency of indexing and Compare()

---

Review status: 6 of 10 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/float, line 48 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-KwF0_Lr4uOZ8ny7nKsC:-KwF0_Lr4uOZ8ny7nKsD:berrqy) ([raw file](https://github.com/cockroachdb/cockroach/blob/53bf0ae9d944850e92444f1b34ce2df49f2f9971/pkg/sql/logictest/testdata/logic_test/float#L48)):*
> ```
> 
> query RBBB
> SELECT f, f IS NaN, f = 'NaN', isnan(f) FROM p ORDER BY 1
> ```

1) add both indexed and non-indexed (with forced sort node) versions of this test, to check that the sorting produces the same order.
2) Add a decimal version of this test and the new one created by point 1.
3) Also have NULL as a value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-:-KwF0xdyCU_8Ro-pVYoI:b-7s7jjt)*

knz:
Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144)*

nvanbenschoten:
Isn't that what the changes to `TestDatumOrdering` are testing?

---

Review status: 1 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/logictest/testdata/logic_test/float, line 48 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-KwF0_Lr4uOZ8ny7nKsC:-KxJ4y6LYMc8rDbA3H3E:bh9pea4) ([raw file](https://github.com/cockroachdb/cockroach/blob/53bf0ae9d944850e92444f1b34ce2df49f2f9971/pkg/sql/logictest/testdata/logic_test/float#L48)):*

1) add both indexed and non-indexed (with forced sort node) versions of this test, to check that the sorting produces the same order.
2) Add a decimal version of this test and the new one created by point 1.
3) Also have NULL as a value.
</blockquote></details>

Good points! Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-:-KxJ4oQ644nC31xUHOCS:b-g7a8sy)*

knz:
Yes but I meant with "clarify" to extend the commit message to outline what's being chosen wrt. nulls.

Otherwise LGTM with just minor remarks left.

---

Reviewed 6 of 9 files at r3, 4 of 4 files at r4.

---

*[pkg/sql/logictest/testdata/logic_test/decimal, line 137 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-KxNcdN8KcUbU8iuvNma:-KxNcdN94EIEP5NG27Kn:bahhk6f) ([raw file](https://github.com/cockroachdb/cockroach/blob/16af7669120fe74d8df4a01b1538e305479e9aab/pkg/sql/logictest/testdata/logic_test/decimal#L137)):*
> ```
> 
> query R
> SELECT * FROM s ORDER BY d
> ```

I think these two ordering tests are now redundant with the one below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-:-KxNbmwZiqOL2mdf_xst:b-6qj1r3)*

nvanbenschoten:
Clarified the commit message. TFTRs!

---

Review status: 2 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/logictest/testdata/logic_test/decimal, line 137 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-KxNcdN8KcUbU8iuvNma:-KxhprIw1rsuivxa9Msd:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/16af7669120fe74d8df4a01b1538e305479e9aab/pkg/sql/logictest/testdata/logic_test/decimal#L137)):*

I think these two ordering tests are now redundant with the one below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19144#-:-KxhprfEpeWN1kEgzQHP:bxmohld)*

PR #19146, opened 2017-10-10
acceptance: add consistency checks for photos.photos

vivekmenezes:
So we can ensure that tables not undergoing schema changes are
consistent.


asubiotto (approved):
LGTM but consider putting each `validationQueries` in its own scope

vivekmenezes:
This test is weirdly failing with 
```
allocator_test.go:207: pq: database "photos" does not exist
```

I see database photos in the schema

PR #19148, opened 2017-10-10
sql: improve crdb_internal.ranges

madelynnblue:
This implements the comments in #19041 and completes much of the RFC at
https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20170314_sql_split_syntax.md#4-crdb_internalranges-and-ranges_cached-system-table.

The virtual table infra is not sophisticated enough yet to selectively
disable certain rows or columns, so for now it will display everything.

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -349,7 +349,7 @@ func backupAndRestore(
 			var count int
 			sqlDBRestore.QueryRow(`
 			SELECT count(*) FROM crdb_internal.ranges
-			WHERE start_key = (
+			WHERE start_pretty = (
 				('/Table/' ||
 				(SELECT table_id FROM crdb_internal.tables
 					WHERE database_name = $1 AND name = $2
--- a/pkg/sql/crdb_internal.go
+++ b/pkg/sql/crdb_internal.go
@@ -27,6 +27,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/build"
+	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/keys"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/security"
@@ -1311,15 +1312,43 @@ CREATE TABLE crdb_internal.forward_dependencies (
 var crdbInternalRangesTable = virtualSchemaTable{
 	schema: `
 CREATE TABLE crdb_internal.ranges (
-  range_id  INT,
-  start_key STRING,
-  end_key   STRING,
-  replicas  INT
+  range_id     INT NOT NULL,
+  start_key    BYTES NOT NULL,
+  start_pretty STRING NOT NULL,
+  end_key      BYTES NOT NULL,
+  end_pretty   STRING NOT NULL,
+  database     STRING NOT NULL,
+  "table"      STRING NOT NULL,
+  "index"      STRING NOT NULL,
+  replicas     INT[] NOT NULL,
+  lease_holder INT NOT NULL
 )
 `,
 	populate: func(ctx context.Context, p *planner, prefix string, addRow func(...parser.Datum) error) error {
 		if err := p.RequireSuperUser("read crdb_internal.ranges"); err != nil {
 			return err
+		}
+		descs, err := getAllDescriptors(ctx, p.txn)
+		if err != nil {
+			return err
+		}
+		dbNames := make(map[uint64]string)
+		tableNames := make(map[uint64]string)
+		indexNames := make(map[uint64]map[sqlbase.IndexID]string)
+		parents := make(map[uint64]uint64)
+		for _, desc := range descs {
+			id := uint64(desc.GetID())
+			switch desc := desc.(type) {
+			case *sqlbase.TableDescriptor:
+				parents[id] = uint64(desc.ParentID)
+				tableNames[id] = desc.GetName()
+				indexNames[id] = make(map[sqlbase.IndexID]string)
+				for _, idx := range desc.Indexes {
+					indexNames[id][idx.ID] = idx.Name
+				}
+			case *sqlbase.DatabaseDescriptor:
+				dbNames[id] = desc.GetName()
+			}
 		}
 		ranges, err := scanMetaKVs(ctx, p.txn, roachpb.Span{
 			Key:    keys.MinKey,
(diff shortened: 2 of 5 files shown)


knz:
Very nice. We'll see to filtering on vtables soon enough.

---

Reviewed 3 of 3 files at r1.

---

*[pkg/sql/crdb_internal.go, line 1315 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw4kSd-DDLMnEWHFbxX:-Kw4kSd0NRyLu0MWMaWC:b-6fwlq8) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/crdb_internal.go#L1315)):*
> ```Go
> 	schema: `
> CREATE TABLE crdb_internal.ranges (
>   range_id     INT,
> ```

We use NOT NULL on the columns that indeed cannot be NULL, see the other vtable definitions.

---

*[pkg/sql/crdb_internal.go, line 1322 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw4kXDPduPCg3UFHPr_:-Kw4kXDPduPCg3UFHPra:b-fpo1j7) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/crdb_internal.go#L1322)):*
> ```Go
>   database     STRING,
>   "table"      STRING,
>   "index"        STRING,
> ```

irregular spacing.

---

*[pkg/sql/show_test.go, line 562 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw4l3Txi4qaj5HdRo1O:-Kw4l3TyMnda9JegwgCy:b-jqccrl) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/show_test.go#L562)):*
> ```Go
> 
> 	// These tests would be a better fit for the logic tests, but they do not seem
> 	// to generate splits on their own, so we need to use a normal test instead.
> ```

I do not understand this. There is no reason why a Go test would do any more or less work than a logic test. If you see a difference it may be due to non-deterministic timings (as we've seen with split tests before) which would imply this test is flaky too. Can you investigate further and determine what the difference is, then extend your comment accordingly?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-:-Kw4lMjDNbTcBi5pO5MN:bwt9ht8)*

danhhz (approved):
This is great! :lgtm: but you should also get @knz's approval before merging

RaduBerinde:
LGTM (modulo knz's comments)

---

---

*[pkg/sql/crdb_internal.go, line 1386 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw6rQMo8BFv1wADtCt_:-Kw6rQMo8BFv1wADtCta:b7k2bk7) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/crdb_internal.go#L1386)):*
> ```Go
> 
> 			// Get the lease holder.
> 			// TODO(radu): this will be slow if we have a lot of ranges; find a way to
> ```

In addition to plumbing filters through to this function, we should also plumb needed columns information @knz

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-:-Kw6s4ecYl2p9-G2sG35:b3942xb)*

mjibson:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/crdb_internal.go, line 1315 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw4kSd-DDLMnEWHFbxX:-Kw8nnGC68Ee6MF_RwHo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/crdb_internal.go#L1315)):*

We use NOT NULL on the columns that indeed cannot be NULL, see the other vtable definitions.
</blockquote></details>

Done.

---

*[pkg/sql/crdb_internal.go, line 1322 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw4kXDPduPCg3UFHPr_:-Kw8nnWbs4Xwc1YLBCW1:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/crdb_internal.go#L1322)):*

irregular spacing.
</blockquote></details>

Done.

---

*[pkg/sql/show_test.go, line 562 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw4l3Txi4qaj5HdRo1O:-Kw8rqqlExvE_W7ufuaH:bpj0g4b) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/show_test.go#L562)):*

I do not understand this. There is no reason why a Go test would do any more or less work than a logic test. If you see a difference it may be due to non-deterministic timings (as we've seen with split tests before) which would imply this test is flaky too. Can you investigate further and determine what the difference is, then extend your comment accordingly?
</blockquote></details>

Moved these tests back to the logic tests. I must have been missing something before.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148)*

knz:
LGTM however from the TC output:

- I think the backup/restore tests are telling you something
- there's a testrace failure that needs a separate issue

---

Reviewed 5 of 5 files at r2.

---

*[pkg/sql/crdb_internal.go, line 1386 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw6rQMo8BFv1wADtCt_:-Kw9XCc5Saq8BdihDjAD:b-c661v) ([raw file](https://github.com/cockroachdb/cockroach/blob/f395cb84a22f645eed7b123639394c063547d6e9/pkg/sql/crdb_internal.go#L1386)):*

In addition to plumbing filters through to this function, we should also plumb needed columns information @knz
</blockquote></details>

Agreed, well noted.

---

*[pkg/sql/logictest/testdata/logic_test/ranges, line 241 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-Kw9XJ1DMj5ZfQCemPSh:-Kw9XJ1DMj5ZfQCemPSi:bnz5gn) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf8d544698cad5e15d67b8bacf4060633d0d0ebb/pkg/sql/logictest/testdata/logic_test/ranges#L241)):*
> ```
> # Create and drop things to produce interesting data for crdb_internal.ranges.
> 
> statement ok
> ```

FYI you can have multiple preparatory statements separated by semicolons below a single `statement ok`.
(no need to change this here though - but it does nicely condense the tests.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19148#-:-Kw9Xf9iYeJiniSn6e_E:bbsrqt)*

tschottdorf:
I haven't reviewed this, but just wanted to say thanks because having to figure out which table a range belongs to has always been annoying, until now. I just used this for #19059 and it worked like a charm.

PR #19153, opened 2017-10-10
docs: pre-generate func & op tables in HTML

### LINKED ISSUE #18647: Tables are broken in autogenerated functions documentation
When you regenerate the functions docs, tables are on the page are broken, apparently due to recently added examples:

screen shot 2017-09-20 at 4 44 07 pm

## Steps to reproduce:

1. Go to your clone of the docs repo and open a new branch.
2. Go to your clone of the cockroach repo and make sure you're on the `release-1.1` branch: 
   ```
   git checkout release-1.1
   ```
3. Run: 
   ```
   go run pkg/cmd/docgen/* ../docs/_includes/sql/v1.1
   ```
4. Go back your docs branch and check the status: 
   ```
   git status
   ```
5. You should see the following:
   ```
   modified:   _includes/sql/v1.1/aggregates.md
   modified:   _includes/sql/v1.1/functions.md
   modified:   _includes/sql/v1.1/operators.md
   ```
6. Start a local docs server:
   ```
   make serve
   ```
7. Open http://localhost:4000/docs/dev/functions-and-operators.html in a browser. 

Result: You should see that several tables are broken, seemingly where examples have been added.

knz:
Markdown is a terrible markup format which is unable to process
markup inside table cells.

Instead pre-process each table cell as an individual Markdown
document and embed the resulting HTML in the output.

This patch incidentally also fixes hyperlinks for array arguments and
`collatedstring`.

Fixes #18647.

### DIFF
--- a/pkg/cmd/docgen/funcs.go
+++ b/pkg/cmd/docgen/funcs.go
@@ -24,6 +24,7 @@ import (
 	"sort"
 	"strings"
 
+	markdown "github.com/golang-commonmark/markdown"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
@@ -164,6 +165,7 @@ const notUsableInfo = "Not usable; exposed only for compatibility with PostgreSQ
 func generateFunctions(from map[string][]parser.Builtin, categorize bool) []byte {
 	functions := make(map[string][]string)
 	seen := make(map[string]struct{})
+	md := markdown.New(markdown.XHTMLOutput(true), markdown.Nofollow(true))
 	for name, fns := range from {
 		// NB: funcs can appear more than once i.e. upper/lowercase varients for
 		// faster lookups, so normalize to lowercase and de-dupe using a set.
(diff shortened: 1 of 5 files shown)


benesch:
Thanks, @knz! I worry about introducing yet another Markdown parser, though. I put the docs team through some serious misery when I swapped out the Ruby Markdown parser. Markdown is such an underspecified language that the parsers tend to have severe discrepancies.

On my phone now, but is there any way we can avoid MD entirely in these docs? I can't imagine we're doing much more than wrapping things in backticks.

knz:
Huh, we are. There's a full table in one of the descriptions; we want bullet lists in several of them, many contain multiple paragraphs and nearly all of them contain words between backticks intended to be rendered as code.

Meanwhile embedding HTML in the info strings is also a no-no because we want to dump the info strings also as plain text in the interactive contextual help.

knz:
If you ask me, both the inline description field and the doc page would be authored in rST, because rST always embeds cleanly within rST!!!1!1!!!! kthxbye

knz:
Also:

1) The solution here "works for me"

2) I do not know of any other solution

So if you'd rather see something else, feel free to make a concrete proposal and hold my hand to get there (or even better: propose a superseding PR). I said to @jseldess I could fix this before Oct 15th with this specific solution in mind; I am not able to guarantee any other solution within that timeline.

benesch (approved):
Oh man, that table in table makes me sad. LGTM given the volume of Markdown we use in those table cells.

knz:
Thanks!

pkg/cmd/docgen/funcs.go
```diff
@@ -24,6 +24,7 @@ import (
 	"sort"
 	"strings"
 
+	markdown "github.com/golang-commonmark/markdown"
```
nvanbenschoten: What's up with the alias here?

knz: Hm not sure what happened. Will clean up.

jseldess:
Thanks very much for working out a short-term solution, @knz! I'm very much hoping we can move to rST soon, if not for the whole docs, then at least for table-intensive pages like this.

PR #19150, opened 2017-10-10
sql: avoid a panic with multi-valued SRFs

knz:
This ensures that a multi-valued SRF in render position, e.g.
`SELECT pg_get_keywords()`, is rendered as a tuple, as in Postgres,
instead of crashing the server.

Fixes #19149.

PR #19160, opened 2017-10-10
sql: use a single instance of UnqualifiedStar

### LINKED ISSUE #19159: sql: singleton UnqualifiedStar
Suggested by @nvanbenschoten 

> nit: we should create a single instance of parser.UnqualifiedStar{} and use that everywhere.

knz:
Fixes #19159.

### DIFF
--- a/pkg/sql/parser/var_name.go
+++ b/pkg/sql/parser/var_name.go
@@ -82,11 +82,6 @@ var _ VarName = &ColumnItem{}
 // NormalizeVarName is a no-op for UnqualifiedStar (already normalized)
 func (u UnqualifiedStar) NormalizeVarName() (VarName, error) { return u, nil }
 
-var singletonStarName VarName = UnqualifiedStar{}
-
-// StarExpr is a convenience function that represents an unqualified "*".
-func StarExpr() VarName { return singletonStarName }
-
 // ResolvedType implements the TypedExpr interface.
 func (UnqualifiedStar) ResolvedType() Type {
 	panic("unqualified stars ought to be replaced before this point")
--- a/pkg/sql/parser/name_part.go
+++ b/pkg/sql/parser/name_part.go
@@ -109,6 +109,9 @@ func (l NameList) Format(buf *bytes.Buffer, f FmtFlags) {
 // a '*' as name part of an UnresolvedName.
 type UnqualifiedStar struct{}
 
+// UnqualifiedStarInstance is the singleton instance of UnqualifiedStar.
+var UnqualifiedStarInstance = UnqualifiedStar{}
+
 // Format implements the NodeFormatter interface.
 func (UnqualifiedStar) Format(buf *bytes.Buffer, _ FmtFlags) { buf.WriteByte('*') }
 func (u UnqualifiedStar) String() string                     { return AsString(u) }
--- a/pkg/sql/parser/select.go
+++ b/pkg/sql/parser/select.go
@@ -130,7 +130,7 @@ func (node *SelectExpr) NormalizeTopLevelVarName() error {
 // starSelectExpr is a convenience function that represents an unqualified "*"
 // in a select expression.
 func starSelectExpr() SelectExpr {
-	return SelectExpr{Expr: StarExpr()}
+	return SelectExpr{Expr: UnqualifiedStarInstance}
 }
 
 // Format implements the NodeFormatter interface.
--- a/pkg/sql/parser/sql.go
+++ b/pkg/sql/parser/sql.go
@@ -13744,7 +13744,7 @@ sqldefault:
 		sqlDollar = sqlS[sqlpt-4 : sqlpt+1]
 		//line sql.y:5285
 		{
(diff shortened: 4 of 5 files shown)


nvanbenschoten:
I'm confused. I didn't realize we already were using a singleton. We should have just been calling `StarExpr()` in #19150.

knz:
StarExpr() is a VarName. Not suitable for use as element of UnresolvedName.

pkg/sql/parser/var_name.go
```diff
@@ -82,10 +82,7 @@ var _ VarName = &ColumnItem{}
 // NormalizeVarName is a no-op for UnqualifiedStar (already normalized)
 func (u UnqualifiedStar) NormalizeVarName() (VarName, error) { return u, nil }
 
-var singletonStarName VarName = UnqualifiedStar{}
-
-// StarExpr is a convenience function that represents an unqualified "*".
-func StarExpr() VarName { return singletonStarName }
+var UnqualifiedStarInstance = UnqualifiedStar{}
```
nvanbenschoten: Move this to `name_part.go` near the type definition.

nvb (approved):

pkg/sql/parser/var_name.go
knz: Done.

petermattis:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/parser/name_part.go, line 113 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19160#-Kw684mwWO6FxH5gZz-f:-Kw684mxnM5WORYxKGwl:bbb99zq) ([raw file](https://github.com/cockroachdb/cockroach/blob/74b2140ae2a3e1ba46a5a64c8d1b7571b48cb85a/pkg/sql/parser/name_part.go#L113)):*
> ```Go
> 
> // UnqualifiedStarInstance is the singleton instance of UnqualifiedStar.
> var UnqualifiedStarInstance = UnqualifiedStar{}
> ```

Why is this better than using `UnqualifiedStar{}` everywhere? (I know one reason, but I doubt it is the reason you're using).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19160)*

nvanbenschoten:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/parser/name_part.go, line 113 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19160#-Kw684mwWO6FxH5gZz-f:-Kw68ujB1Idv_QXvfyx4:b-vq5v9e) ([raw file](https://github.com/cockroachdb/cockroach/blob/74b2140ae2a3e1ba46a5a64c8d1b7571b48cb85a/pkg/sql/parser/name_part.go#L113)):*

Why is this better than using `UnqualifiedStar{}` everywhere? (I know one reason, but I doubt it is the reason you're using).
</blockquote></details>

I suggested it in a previous review because there's no reason to allocate multiple instances of this. Thinking about it more though, it's zero sized so it shouldn't allocate when put in an interface anyway. Plus, even if it did, Go may [avoid duplicate allocs](https://golang.org/ref/spec#Size_and_alignment_guarantees).

What reason were you thinking? Well-defined comparisons?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19160)*

knz:
merge or no merge?

petermattis:
I'd say no merge, but don't have a strong preference. 

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/name_part.go, line 113 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19160#-Kw684mwWO6FxH5gZz-f:-KwetVLb_HGLTf9hsdLn:b8zuthk) ([raw file](https://github.com/cockroachdb/cockroach/blob/74b2140ae2a3e1ba46a5a64c8d1b7571b48cb85a/pkg/sql/parser/name_part.go#L113)):*

I suggested it in a previous review because there's no reason to allocate multiple instances of this. Thinking about it more though, it's zero sized so it shouldn't allocate when put in an interface anyway. Plus, even if it did, Go may [avoid duplicate allocs](https://golang.org/ref/spec#Size_and_alignment_guarantees).

What reason were you thinking? Well-defined comparisons?
</blockquote></details>

Assigning `UnqualifiedStar{}` to an interface requires a runtime call that has small but measurable overhead. The difference is so minor that this isn't worth optimizing. I'm pretty sure this change doesn't change comparisons.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19160#-:-KwettBnJ0tjrZxBgkg9:b-48gj5d)*

nvanbenschoten:
Yeah, I'd say no merge. Sorry for the misguided recommendation and thanks for putting this together.

knz:
Thanks for the discussion!

PR #19166, opened 2017-10-10
sql: Add TableDescriptor validation for duplicate indexes

lgo:
Previously we didn't check if an index was a duplicate when running
TableDescriptor.Validate.

This is now checked. NB: Adding this check will be a breaking change if
a duplicate index already exists, as an error will now be thrown.

### DIFF
--- a/pkg/ccl/sqlccl/restore.go
+++ b/pkg/ccl/sqlccl/restore.go
@@ -689,7 +689,7 @@ func restoreTableDescs(
 		}
 
 		for _, table := range tables {
-			if err := table.Validate(ctx, txn); err != nil {
+			if err := table.Validate(ctx, txn, false /* newOrModifiedTable */); err != nil {
 				return err
 			}
 		}
--- a/pkg/sql/create.go
+++ b/pkg/sql/create.go
@@ -438,7 +438,7 @@ func (n *createViewNode) Start(params runParams) error {
 		return err
 	}
 
-	if err = desc.ValidateTable(); err != nil {
+	if err = desc.ValidateTable(true /* newOrModifiedTable */); err != nil {
 		return err
 	}
 
@@ -471,7 +471,7 @@ func (n *createViewNode) Start(params runParams) error {
 	if desc.Adding() {
 		n.p.notifySchemaChange(&desc, sqlbase.InvalidMutationID)
 	}
-	if err := desc.Validate(params.ctx, n.p.txn); err != nil {
+	if err := desc.Validate(params.ctx, n.p.txn, true /* newOrModifiedTable */); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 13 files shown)


lego:
Just to confirm, would we consider indexes different if their directions are different?

Right now the patch disregards direction in checking.

lego:
Yep, just confirmed weird things happen.

There are failures from `lease.go` and `schema_changer.go` when validating the table descriptor, and as a result, a table descriptor lease can't be acquired so you can't run any queries including to drop the duplicate index.

Are there any procedures for running "migrations" between CockroachDB versions?

Alternatively, a sane course of action is to remove the duplicate index, but that may not be possible without first removing it from the descriptor or disabling validation. Seems to be a fair amount of work, but also this would _eventually_ need to be done.

petermattis:
> Just to confirm, would we consider indexes different if their directions are different?

Yes. `INDEX (a ASC, b DESC)` is not the same as `INDEX (a ASC, b ASC)`. The former can be used to optimize a query with `ORDER BY a ASC b DESC` while the latter can't.

I suspect adding this check to `TableDescriptor.Validate` will be problematic. I suggest adding a new method that is only called when a table is created or modified, not when it is retrieved.

mjibson:
I suggest an API change to not take a boolean, but have separate functions instead. See https://ariya.io/2011/08/hall-of-api-shame-boolean-trap for more. Forcing the use of a comment for every single call to that function is an antipattern. Yes, we already do that in many places, but we can stop the bleeding.

danhhz:
Eek! What's the motivation for this? We actually have a reasonable use for exactly duplicated indexes in partitioning, so I'd like to avoid adding this requirement if we can

Concretely, if you have a users table partitioned by country, but want a global index on the id in each region (so inserts are more work, but lookups don't cross oceans), that's easy to do with multiple identical non-partitioned indexes that are each zoned to a different region. All of the other ways we've thought of to do this are pretty gross

lego:
Gotcha.

I was going through schema checks to be as exhaustive as possible, and this was a todo that was easy to hash out. So instead how about we remove the todo for checking indexes are unique, and possibly add a note to `validateTableIndexes` stating duplicate indexes are valid?

benesch:
Also I believe some ORMs create duplicate indices 🤦‍♂️:

https://github.com/cockroachdb/sequelize-cockroachdb/issues/18

danhhz:
> I was going through schema checks to be as exhaustive as possible, and this was a todo that was easy to hash out. So instead how about we remove the todo for checking indexes are unique, and possibly add a note to validateTableIndexes stating duplicate indexes are valid?

SGTM! Thanks!

lego:
closed in favour of #19183

PR #19175, opened 2017-10-11
rfcs: new policy RFC about beta and experimental features

knz:
As discussed.

### DIFF
--- a/docs/RFCS/beta_and_experimental_features.md
+++ b/docs/RFCS/beta_and_experimental_features.md
@@ -1,0 +1,249 @@
+- Feature Name:  Experimental features
+- Status: draft
+- Start Date: 2017-10-11
+- Authors: knz, Diana Hsieh
+- RFC PR: (PR # after acceptance of initial draft)
+- Cockroach Issue: (one or more # from the issue tracker)
+
+# Summary
+
+This RFC proposes to:
+
+1) define the "Experimental" status of new functionality that also has UX
+   aspects (as opposed to new functionality that has no user-visible
+   changes at all, for which this RFC is not applicable).
+2) present recommendations as to how to present beta features to
+   users: command-line flags, SQL language extensions, cluster
+   settings, etc.
+
+In a nutshell, *this RFC intends to make it unambiguous to users when
+functionality is still considered as beta/experimental* -- useful in at least the following contexts:
+
+- when the CockroachDB team reserves itself the right to change UX
+  aspects without notice, and without running afoul of the expectation
+  of backward compatibility.
+- for features that don't "make data easy," it's better for us to
+  clearly mark them as experimental so as not to hurt the perception
+  of our brand.
+
+Out of scope: "internal" experimental features that only support
+another feature but are entirely invisible to users.
+
+# Motivation
+
+In CockroachDB a single feature may pop up in the UX surface
+area in multiple locations, including but not limited to:
+
+- command-line flags;
+- environment variables;
+- cluster settings;
+- session variables;
+- local commands in the `cockroach sql` interactive prompt;
+- the SQL language that can be used both in `cockroach sql` and any client app;
+- SQL contextual help;
+- the online documentation;
+- the web UI (sometimes called "web console" by users);
+- log messages.
+
+For each location where features are visible to users, a user
+discovering the feature may (obviously!) start to use it, and start
+*expecting* the feature to *remain available* in the future. Since
+CockroachDB also advertises backward compatibility across releases,
+this expectation seems reasonable at first.
+
+Yet, when new features are added to CockroachDB, there is a period of
+time where the UX aspects of the feature may not be stabilized yet: we
+may not be certain yet that the UX choices are good, or perhaps early
+users will provide feedback that change our UX choices. Thus, the
+CockroachDB team may want to preserve the ability to change the UX
+aspects without violating the promise of backward compatibility.
+
+This is analogous to the process of "API stabilization" when a library
+or online services adds a new feature -- the feature is initially
+released, then early adopters will try it out, and then based on the
+feedback the API may change slightly before it is stabilized and
+becomes "frozen" for the purpose of defining backward compatibility.
+In this initial phase, the new feature/API is typically marked
+"experimental" in its documentation.
+
+In contrast to libraries or online services, however, the "API" of
+CockroachDB has a large surface area, and there is not a single point
+in the product nor the documentation where "features" are identified
+as a bullet list, where it would be convenient to add the text
+"experimental" next to their description.
+
+More to the point, a SQL database is typically a product where users
+are likely to discover features via experimentation as opposed to
+checking the documentation beforehand. Without a clear mention *within
+the product*, users may fail to learn altogether that a feature is
+experimental, and erroneously start to develop an expectation of
+backward compatibility.
+
+This motivates guidelines to make the experimental nature of new
+functionality more clear throughout the product, not only the
+documentation.
+
+# Guide-level explanation
+
+When implementing new features, first determine whether the feature
+should be considered beta/experimental at all. For this refer to the
+section below.
+
+Then, if applicable, clarify the experimental nature of the feature
+in the UX elements explicitly:
+
+- for the log messages: left out of scope in this RFC. We are not
+  considering marking features as experimental when they touch log
+  messages, as we are not currently providing any guarantees about log
+  formatting / messages.
+- for command-line flags: add the marker `[EXPERIMENTAL]` in the help text (`--help`);
+- for cluster settings: add the marker `[EXPERIMENTAL]` in the setting description (column "Description" in `SHOW ALL CLUSTER SETTINGS`). Also see below.
+- for session variables: add the marker `[EXPERIMENTAL]` in the variable description (column "Description" in `SHOW ALL`, also see below).
+- for local commands in the CLI shell: add the marker `[EXPERIMENTAL]` in the online help (output of `\h`)
+- for environment variables: add the word `EXPERIMENTAL` in the name of the environment variable.
+- for SQL contextual help: add the word `[EXPERIMENTAL]` in the description.
+- for the online documentation: be explicit that the feature is experimental.
+- for the web UI: some discreet UI element that marks the feature as experimental.
+- for SQL language extensions:
+
+  - use the word EXPERIMENTAL in the new syntax.
+  - hide the extension from "stable" docs (make them only appear in
+    the current "dev" version docs).
+
+Gating the feature behind a cluster setting:
+
+If possible, make the feature subject to a cluster setting that
+enables it.
+
+The cluster setting should be named `experimental.name-of-the-feature.enabled`.
+Make it default to `false`.
+
+For example we have `experimental.importcsv.enabled` at the time of
+this writing.
+
+When an experimental feature becomes non-experimental, the setting
+that controlled it will remain (until some later version, e.g. the
+next major version), but it may become an error to attempt to set it
+to false. This allows cluster setup/testing scripts to enable
+experimental features that an application requires without breaking
+when the features lose their experimental status".
+
+Be mindful to distinguish:
+
+- *enable / disable* settings that gate experimental features, these
+  should be *named* with the prefix `experimental.`
+- new tuning settings that configure the behavior of some feature (either
+  experimental or not). These should be *described* (column
+  "description") as `[EXPERIMENTAL]`. The rationale is that when the
+  feature transitions from experimental to non-experimental, the users
+  will want to preserve their configuration for that feature. So the
+  configuration parameter, if set via a cluster setting, must remain
+  valid.
+
+# What is an Experimental feature?
+
+If the feature has **no** user-facing changes:
+
+- If the feature replaces a previous code path entirely (so there is
+  no way to obtain the previous behavior), treat the feature as
+  non-experimental.
+
+- If the feature proposes an alternative code path, and the previous
+  one is still available, treat the feature as experimental and:
+
+  a) verify that the choice of code paths is effectively configurable;
+  b) verify this configuration choice is documented;
+  c) verify that the change is duly documented so users know about the tradeoffs.
+
+If the feature has user-facing changes:
+
+- If the feature is CockroachDB-specific (doesn't aim for
+  compatibility):
+  
+  - if the UX of the feature falls short of our standards / QA at the
+    time of release, treat the feature as experimental.
+
+  - otherwise, treat as non-experimental.
+
+- If the feature aims for compatibility with another product:
+
+  - if the part of its UX that aims to correspond to an equivalent
+    feature does not mirror 1-1 the surface aspects (functional) of a
+    competing product, treat the feature as experimental.
+
+  - if it does not match existing user expectations (e.g., those
+    expectations developed via a competing product) about how it
+    should perform at an extra-functional level (e.g., performance),
+    then treat the feature as experimental. Check that the differences
+    with the competing product(s) are properly documented in the RFC
+    that introduces the new feature, and later in online docs.
+
+  - if it does match user expectations, then:
+
+    - if the change is "small" as defined in the contributor guide,
+      and the feature mirrors 1-to-1 a SQL standard or a competing
+      product, then consider the feature as non-experimental.
+
+    - otherwise, consider the feature as experimental, unless the RFC
+      that introduces the new feature makes a successful argument
+      otherwise.
+
+# Reference-level explanation
+
+## Detailed design
+
+- To prepare for the application of this RFC, the session variable mechanism
+  (`pkg/sql/vars.go`) must be extended to support a "description" field, like
+  cluster settings already have.
+
+- Other features currently experimental at the time of this writing must be
+  reviewed and possibly documented as such in all their UX surface area
+  (e.g. IMPORT CSV, the experimental SQL statements, maybe
+  others. Also see unresolved questions below.)
+
+- The RFC on SQL syntax changes must be updated accordingly.
+
+No other technical changes needed.
+
+## Drawbacks
+
+None known.
+
+## Rationale and Alternatives
+
+The motivation section outlines the reason why "do nothing" is not a
+viable alternative: we do want to commit to backward compatibility and
+*also* want to keep some wiggle room and release new features before
+their UX is finalized; so we must find a way to carve out new features
+away from the scope of backward compatibility.
+
+Some alternatives that were considered:
+
+- design and implement the feature "as finished" and only gate it behind a
+  `experimental` cluster setting. Problems with this approach:
+
+  - some features are not amenable to cluster settings, e.g. flags
+    that control the behavior of `cockroach start` or `cockroach sql` when
+    the server is not active yet;
+
+  - the DBA may know about the experimental nature of the feature,
+    because they set the cluster setting, but then developers with
+    access to the database may start discovering the feature after
+    it's enabled and not understand that the feature is experimental,
+    because nothing else would tell them.
+
+- design and implement the feature "as finished" and only document it
+  in the online docs. Problems with this approach:
+
+  - users may discover features without reading the docs and develop a
+    wrong expectation.
+
+## Unresolved questions
+
+- What is the list of features that are already implemented but still
+  currently considered experimental, and which thus need to be made more
+  explicitly so according to this RFC? Noted during review:
+
+  - IMPORT CSV
+  - disqSQL temporary storage
+  - stats-based rebalancing


danhhz:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtW5_oexDbyKc_kR9:-KwAtW5_oexDbyKc_kRA:b60uheb) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L193)):*
> ```Markdown
> 	the server is not active yet;
> 
>   - the DBA may know about the experimental nature of the feature,
> ```

this is a very compelling argument to me for syntax over cluster settings

---

*[docs/RFCS/beta_and_experimental_features.md, line 215 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtbCLsEKgWtSslax9:-KwAtbCLsEKgWtSslaxA:b-e2eb19) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L215)):*
> ```Markdown
> ## Unresolved questions
> 
> - Is it reasonable to keep the terms "Beta" and "Experimental"
> ```

so far, we've been using experimental to mean both "beta" like IMPORT CSV and "internal only" like https://github.com/cockroachdb/cockroach/pull/19140. the latter is a particularly illustrative example in that we intend to release sql syntax for setting zone configs at some point, but this is _definitely not_ the syntax we'd use for it when we do

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

awoods187:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7EoRP1_PE12Hrx_s:-KwB7EoRP1_PE12Hrx_t:bipprh2) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L22)):*
> ```Markdown
> functionality is still considered as beta/experimental* and when the
> CockroachDB team reserves itself the right to change UX aspects
> without notice without running afoul of the expectation of backward
> ```

repetition of without is awk

---

*[docs/RFCS/beta_and_experimental_features.md, line 71 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7foggbWhYttxHCdd:-KwB7foggbWhYttxHCde:bvkrgqk) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L71)):*
> ```Markdown
> are likely to discover features via experimentation as opposed to
> checking the documentation beforehand. Without a clear mention *within
> the product* a users may fail to learn altogether that a feature is
> ```

nit --remove "a" and add a " , "

---

*[docs/RFCS/beta_and_experimental_features.md, line 75 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7oLkNh-V2RGCCNBd:-KwB7oLkNh-V2RGCCNBe:b-xoiojm) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L75)):*
> ```Markdown
> backward compatibility.
> 
> This motivates guideliens to make the experimental nature of new
> ```

s/ guidelines 

---

*[docs/RFCS/beta_and_experimental_features.md, line 81 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7snFLvJxyWyyBt4y:-KwB7snFLvJxyWyyBt4z:b-yn7d1i) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L81)):*
> ```Markdown
> # Guide-level explanation
> 
> When implementing new features first determine whether the feature
> ```

new features, first

---

*[docs/RFCS/beta_and_experimental_features.md, line 85 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7xtvMTR7T7Ha47ff:-KwB7xtvMTR7T7Ha47fg:b-k2aqz8) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L85)):*
> ```Markdown
> section below.
> 
> Then, if applicable, to display the experimental nature of the feature
> ```

awk phrasing of to display

---

*[docs/RFCS/beta_and_experimental_features.md, line 136 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8J0jWa6j9Nqtaxev:-KwB8J0jWa6j9Nqtaxew:b-7s6wll) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L136)):*
> ```Markdown
> 
> - If the UX of the feature does not mirror 1-1 an equivalent feature
>   in a competing product (e.g. for SQL feature parity), treat the
> ```

e.g.,

---

*[docs/RFCS/beta_and_experimental_features.md, line 159 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8UEahNmySYEskqm-:-KwB8UEahNmySYEskqm0:bb72pdu) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L159)):*
> ```Markdown
> ## Detailed design
> 
> - To prepare for the application of this RFC the session variable mechanism
> ```

RFC,

---

*[docs/RFCS/beta_and_experimental_features.md, line 220 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8tYAqZlsmz-LjlA9:-KwB8tYAqZlsmz-LjlAA:b-txl0rr) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L220)):*
> ```Markdown
> - What is the list of features that are already implemented but still
>   currently considered experimental, and which thus need to be made more
>   explicitly so according to this RFC?
> ```

I think a section is needed in this RFC describing exactly who is responsible for marking a feature as beta/experimental. In my opinion, the product manager should be responsible for making this designation as a part of the new definition of done work

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

nstewart:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 137 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBQTgf6PxJB9yKSkDj:-KwBQTgf6PxJB9yKSkDk:bwiwoyy) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L137)):*
> ```Markdown
> - If the UX of the feature does not mirror 1-1 an equivalent feature
>   in a competing product (e.g. for SQL feature parity), treat the
>   feature as experimental.
> ```

I don't think we should pin this to competing products. I think we should compare against our own quality standards.  So more like "If the UX of the feature falls short of our standards, treat the feature as experimental". Another reason for the slightly relaxed language is that we could have a feature that does 70% of a competitor's feature does, but that other 30% isn't valuable to our user base. We would still make the feature generally available without closing the gap to get to 1-1

---

*[docs/RFCS/beta_and_experimental_features.md, line 213 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBShjWlgwTVsDGFuya:-KwBShjWlgwTVsDGFuyb:btqel9i) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L213)):*
> ```Markdown
> 	to stable without disruption for early users.
> 
> ## Unresolved questions
> ```

I'd like to propose one additional note about positioning beta features: mainly that experimental features still pass our QA process and won't eat your data.  If we want a weaker guarantee that's fine, but whatever our stance is here we need to make that clear

---

*[docs/RFCS/beta_and_experimental_features.md, line 220 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8tYAqZlsmz-LjlA9:-KwBS7ltG4NOop3MmjIi:b-qef0mz) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L220)):*

I think a section is needed in this RFC describing exactly who is responsible for marking a feature as beta/experimental. In my opinion, the product manager should be responsible for making this designation as a part of the new definition of done work
</blockquote></details>

I agree in general that it's ultimately the PM, but I think anyone involved in the development (eng, design, docs) can raise their hand and say a feature should be experimental (along with rationale) as they may see something that the PM overlooked. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

awoods187:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 220 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8tYAqZlsmz-LjlA9:-KwBURNhbmd9KlNdMyON:b-h8n3at) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L220)):*
<details><summary><i>Previously, nstewart (Nate) wrote…</i></summary><blockquote>

I agree in general that it's ultimately the PM, but I think anyone involved in the development (eng, design, docs) can raise their hand and say a feature should be experimental (along with rationale) as they may see something that the PM overlooked. 
</blockquote></details>

Totally agree--didn't mean to dissuade anyone on the team from raising their hand. I think anyone can (and should) raise concerns, I just wanted someone to have clear responsibility to make this determination so that it's not forgotten. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

couchand:
Review status: 0 of 1 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjGOkG1VP2t-bcLBo:-KwBjGOlxf4bF-_LTtTJ:b-2fwmqt) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L38)):*
> ```Markdown
> - SQL contextual help;
> - the online documentation;
> - the admin UI ("web console");
> ```

is the term "web console" ever used?

---

*[docs/RFCS/beta_and_experimental_features.md, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjKdL4SI9NbCe30Kn:-KwBjKdL4SI9NbCe30Ko:b-49mujp) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L39)):*
> ```Markdown
> - the online documentation;
> - the admin UI ("web console");
> - log messages.
> ```

should log messages ever have a guarantee of compatibility?  by their nature they seem like something that can change without notice...

---

*[docs/RFCS/beta_and_experimental_features.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjaDn5ExRIWQpzFKg:-KwBjaDoeicw3tfoezGM:bdr5pnb) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L152)):*
> ```Markdown
> 
>   - otherwise, consider the feature as
> 	experimental, unless the RFC that introduces the new feature makes a
> ```

the formatting on all these second-level points is odd, particularly this one

---

*[docs/RFCS/beta_and_experimental_features.md, line 205 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjnlIMVNiPrOTqd4m:-KwBjnlIMVNiPrOTqd4n:bhjs3si) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L205)):*
> ```Markdown
> 	wrong expectation.
> 
> - use the word `experimental` in the *name* of the flags / settings instead
> ```

isn't this exactly what is proposed above? "The cluster setting should be named `experimental.name-of-the-feature`."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

dianasaur323:
*[docs/RFCS/beta_and_experimental_features.md, line 4 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCLIq8mc4PhJM63df2:-KwCLIq9UTg84jGAq-kA:b-pjoi9j) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L4)):*
> ```Markdown
> - Status: draft
> - Start Date: 2017-10-11
> - Authors: knz, Diana Hsieh
> ```

Ha, I think this is giving me way more credit than I deserve :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

docs/RFCS/beta_and_experimental_features.md
```diff
@@ -0,0 +1,220 @@
...
+to the UX elements explicitly:
+
+- for the log messages: TBD?
+- for command-line flags: add the marker `[EXPERIMENTAL]` in the help text (`--help`);
+- for cluster settings: add the marker `[EXPERIMENTAL]` in the setting description (column "Description" in `SHOW ALL CLUSTER SETTINGS`).
+- for session variables: add the marker `[EXPERIMENTAL]` in the variable description (column "Description" in `SHOW ALL`, also see below).
+- for local commands in the CLI shell: add the marker `[EXPERIMENTAL]` in the online help (output of `\h`)
+- for environment variables: add the word `EXPERIMENTAL` in the name of the environment variable.
+- for SQL contextual help: add the word `[EXPERIMENTAL]` in the description.
+- for the online documentation: be explicit that the feature is experimental.
```
jseldess: We can find a consistent way to do this, perhaps including `Experimental` in the page title, etc. We should probably also add a page explaining what this status means and listing all current experimental features.

docs/RFCS/beta_and_experimental_features.md
```diff
@@ -0,0 +1,220 @@
...
+- for the admin UI: some discreet UI element that marks the feature as experimental.
+- for SQL language extensions:
+
+  - use the word EXPERIMENTAL in the new syntax.
+  - hide the extension from docs.
```
jseldess: This may be excessive in some cases, especially if we want some users to kick the tires.

docs/RFCS/beta_and_experimental_features.md
```diff
@@ -0,0 +1,220 @@
...
+  of their description. Problems with this approach:
+
+  - if the feature evolves beyond the experimental phase to "stable"
+	without change, the client apps / deployments must still be
+	updated to use the new name. We want a path to promote a feature
+	to stable without disruption for early users.
+
+## Unresolved questions
+
+- Is it reasonable to keep the terms "Beta" and "Experimental"
```
jseldess: I find a bit odd that we use alpha and beta as release terms but beta and experimental here. I personally think alpha conveys that something is very "early".

dianasaur323:
Review status: 0 of 1 files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCMLbSrdGh0tQRuudo:-KwCMLbSrdGh0tQRuudp:bqxvih0) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L23)):*
> ```Markdown
> CockroachDB team reserves itself the right to change UX aspects
> without notice without running afoul of the expectation of backward
> compatibility.
> ```

I think we can also incorporate the concept of building "brand." So for features that don't "make data easy," it's better for us to clearly mark them as experimental so as not to hurt the perception of our brand.

---

*[docs/RFCS/beta_and_experimental_features.md, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCN90AG5zgEsWFSviq:-KwCN90AG5zgEsWFSvir:boyo9go) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L28)):*
> ```Markdown
> 
> In CockroachDB a single feature may pop up in the UX surface
> area in multiple locations, including but not limited to:
> ```

I think here, we might also want to include UX experiences that aren't as obvious - like UX that effects the performance of queries or the behavior of the database. Maybe this isn't the right place to put that. I think there are explicit UX locations and implicit UX locations.

---

*[docs/RFCS/beta_and_experimental_features.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjGOkG1VP2t-bcLBo:-KwCN2ik7RWMqRW376Jj:b-d6v4xm) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L38)):*

is the term "web console" ever used?
</blockquote></details>

I think admin UI might be good enough here, like Andrew says

---

*[docs/RFCS/beta_and_experimental_features.md, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjKdL4SI9NbCe30Kn:-KwCNNOZjNJTH6K2d6ZG:bk04vjs) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L39)):*

should log messages ever have a guarantee of compatibility?  by their nature they seem like something that can change without notice...
</blockquote></details>

We may need to guarantee compatibility at some point if we wanted to support some logging applications for metrics and alerting. perhaps pre-emptive to worry about that for now though.

---

*[docs/RFCS/beta_and_experimental_features.md, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCNgAWwvlhMCMdN12Q:-KwCNgAWwvlhMCMdN12R:bfr9qml) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L69)):*
> ```Markdown
> 
> More to the point, a SQL database is typically a product where users
> are likely to discover features via experimentation as opposed to
> ```

This is such a great point - documentation is simply not enough for things like this :+1: 

---

*[docs/RFCS/beta_and_experimental_features.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOAfd7RcnX-GtUvJE:-KwCOAfd7RcnX-GtUvJF:b-bbz1n1) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L83)):*
> ```Markdown
> When implementing new features first determine whether the feature
> should be considered beta/experimental at all. For this refer to the
> section below.
> ```

I do think sometimes we don't know whether a feature is experimental until we implement and find it lacking from a UX perspective

---

*[docs/RFCS/beta_and_experimental_features.md, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOKrhMrdSg8CiFO_j:-KwCOKrhMrdSg8CiFO_k:b-5mvg8v) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L88)):*
> ```Markdown
> to the UX elements explicitly:
> 
> - for the log messages: TBD?
> ```

Perhaps we don't have to worry about this. This is a bit low level.

---

*[docs/RFCS/beta_and_experimental_features.md, line 90 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOPkx6761EUKp8B6K:-KwCOPkx6761EUKp8B6L:byzs312) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L90)):*
> ```Markdown
> - for the log messages: TBD?
> - for command-line flags: add the marker `[EXPERIMENTAL]` in the help text (`--help`);
> - for cluster settings: add the marker `[EXPERIMENTAL]` in the setting description (column "Description" in `SHOW ALL CLUSTER SETTINGS`).
> ```

@mjibson named the import csv flag experimental.feature_name, so perhaps we should keep this syntax

---

*[docs/RFCS/beta_and_experimental_features.md, line 102 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCP3bd2iFMpitr2gel:-KwCP3bd2iFMpitr2gem:bnr0l4r) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L102)):*
> ```Markdown
>   - hide the extension from docs.
> 
> Gating the feature behind a cluster setting:
> ```

Could we actually move this above your list of places where EXPERIMENTAL should be? So, if possible, gate by cluster setting. In addition, do the following bullets. If gating isn't possible, do the following bullets.

---

*[docs/RFCS/beta_and_experimental_features.md, line 137 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBQTgf6PxJB9yKSkDj:-KwCPc7z2KLZwW32xwGE:b-9ek8s8) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L137)):*
<details><summary><i>Previously, nstewart (Nate) wrote…</i></summary><blockquote>

I don't think we should pin this to competing products. I think we should compare against our own quality standards.  So more like "If the UX of the feature falls short of our standards, treat the feature as experimental". Another reason for the slightly relaxed language is that we could have a feature that does 70% of a competitor's feature does, but that other 30% isn't valuable to our user base. We would still make the feature generally available without closing the gap to get to 1-1
</blockquote></details>

Agreed, I think feature parity will be hard for us

---

*[docs/RFCS/beta_and_experimental_features.md, line 143 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCPk-ZmxWXw46yzCLo:-KwCPk-_8-hqeGm2q6s4:bgiyav9) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L143)):*
> ```Markdown
>   that the differences with the competing product(s) are properly
>   documented in the RFC that introduces the new feature, and later in
>   online docs.
> ```

I think instead of this, we should say something like:

- If it does not meet the user's needs as specified in the feature's use cases. Or something along those lines. Forgive my inability to state that eloquently.

---

*[docs/RFCS/beta_and_experimental_features.md, line 145 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCQ0rC2SVjafIJK3Vc:-KwCQ0rDe853u6fCTfr_:bszuakh) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L145)):*
> ```Markdown
>   online docs.
> 
> - If it does perform very similarly, then:
> ```

Yeah, here again, perhaps we should focus on "meeting users needs in a way that meets our company-wide standards" versus feature parity

---

*[docs/RFCS/beta_and_experimental_features.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtW5_oexDbyKc_kR9:-KwCQaT2-TMxzC3OLl9Y:b-fc6d7e) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L193)):*

this is a very compelling argument to me for syntax over cluster settings
</blockquote></details>

Ah, this is quite compelling. Hm. I think it's overkill to do both. Unless you would want to give DBAs the power to prevent developers from using experimental features. We should decide on something here. So glad we are doing this RFC!

---

*[docs/RFCS/beta_and_experimental_features.md, line 211 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCQrcEKl35ZkBlO0pn:-KwCQrcEKl35ZkBlO0po:b-61zsi) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L211)):*
> ```Markdown
> 	without change, the client apps / deployments must still be
> 	updated to use the new name. We want a path to promote a feature
> 	to stable without disruption for early users.
> ```

Hm... that is a strike against putting EXPERIMENTAL in syntax that I think is pretty impactful from a upgrade / user experience

---

*[docs/RFCS/beta_and_experimental_features.md, line 218 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCR8Mmx4LTtSWKJUEc:-KwCR8Mmx4LTtSWKJUEd:b-2cz67u) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L218)):*
> ```Markdown
>   synonymous for now?
> 
> - What is the list of features that are already implemented but still
> ```

Top of mind for me are import CSV and temp storage (which I know isn't experimental, but probably should have been?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 36 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjGOkG1VP2t-bcLBo:-KwXHNIWGqY8YyKsnblx:bc7xv4q) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L38)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I think admin UI might be good enough here, like Andrew says
</blockquote></details>

This isn't really the place for this discussion, but even though we call it the "admin UI" internally, I'd prefer to shift to calling it the "web UI". Admins also use (and may prefer, where they have overlapping features) the CLI. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjKdL4SI9NbCe30Kn:-KwXIA1yB1oxQzJksGPT:b-br41vf) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L39)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

We may need to guarantee compatibility at some point if we wanted to support some logging applications for metrics and alerting. perhaps pre-emptive to worry about that for now though.
</blockquote></details>

-1 on making any guarantees at all about our log messages for the foreseeable future. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOAfd7RcnX-GtUvJE:-KwXJZeN95MbBrAwZ6pj:bk103no) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L83)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I do think sometimes we don't know whether a feature is experimental until we implement and find it lacking from a UX perspective
</blockquote></details>

In a sense, *everything* is experimental in an alpha. We should aim to identify which features will be experimental in a release before the beta freeze and mark them as such at that time. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 106 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXIh88mNVk-sU78wX2:-KwXIh88mNVk-sU78wX3:bawkr2f) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L106)):*
> ```Markdown
> If possible, make the feature subject to a cluster setting that
> enables it. This way a user that wants to run CockroachDB in
> "pristine" condition can disable the feature cluster-wide.
> ```

These settings default to false, right? No one's going to go in and manually disable a feature to get a "pristine" condition; it's about requiring people to opt in and accept the rough edges or potential future changes to the feature. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXJImIXVrZzjuUrkuN:-KwXJImIXVrZzjuUrkuO:b-7yfsy6) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L108)):*
> ```Markdown
> "pristine" condition can disable the feature cluster-wide.
> 
> The cluster setting should be named `experimental.name-of-the-feature`.
> ```

`experimental.name-of-the-feature.enabled`

---

*[docs/RFCS/beta_and_experimental_features.md, line 110 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXK2G5fm23_-_4Rabv:-KwXK2G5fm23_-_4Rabw:bgug98c) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L110)):*
> ```Markdown
> The cluster setting should be named `experimental.name-of-the-feature`.
> 
> For example we have `experimental.importcsv.enabled` at the time of
> ```

I would add:

" When an experimental feature becomes non-experimental, the setting that controlled it will remain  (forever? for N versions?), but it may become an error to attempt to set it to false. This allows cluster setup/testing scripts to enable experimental features that an application requires without breaking when the features lose their experimental status". 

---

*[docs/RFCS/beta_and_experimental_features.md, line 124 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXKxm6Z4u0qKF0w_35:-KwXKxm6Z4u0qKF0w_36:b-2b86op) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L124)):*
> ```Markdown
>   no way to obtain the previous behavior), treat the feature as
>   non-experimental. Of course during the design of the feature one
>   should strive to avoid such replacements.
> ```

I disagree with the last sentence. We should strive to make changes that are safe enough (and test them enough during the development cycle) that we don't need to expose them behind a flag. Experimental status should be relatively rare. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXLfyan8wzNrQf-E5u:-KwXLfyan8wzNrQf-E5v:b-7ineap) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L152)):*
> ```Markdown
> 
>   - otherwise, consider the feature as
> 	experimental, unless the RFC that introduces the new feature makes a
> ```

I strongly disagree with the experimental-by-default philosophy. An experimental phase (that spans a release - remember that in an alpha everything's experimental) is pretty expensive since we have to test with and without it, users who are interested in the feature have to deal with extra complexity, etc. Experimental features should be reserved for the relatively few projects that can't be completed in one release cycle. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtW5_oexDbyKc_kR9:-KwXNJO_AtByxkPEO0JR:bf837e7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L193)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Ah, this is quite compelling. Hm. I think it's overkill to do both. Unless you would want to give DBAs the power to prevent developers from using experimental features. We should decide on something here. So glad we are doing this RFC!
</blockquote></details>

Yeah, where possible (syntax, flag/env var names, etc) we should put the word "experimental" in the user's face. Cluster settings are only needed for things where otherwise the only way they'd know it's experimental is by reading the docs. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 211 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCQrcEKl35ZkBlO0pn:-KwXNapU_hFVDG7QMqlx:bycwhop) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L211)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Hm... that is a strike against putting EXPERIMENTAL in syntax that I think is pretty impactful from a upgrade / user experience
</blockquote></details>

If a feature is stabilized without further change, we should make the token `EXPERIMENTAL` optional (at least for a few releases). No reason to actively break applications that were using the experimental syntax. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 215 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAdV283YteeGt1OuUZ-r1-215:-KwXNyFcmo75_4sHDlMB:b-jomdy) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L215)):*
<details><summary><i>Previously, jseldess wrote…</i></summary><blockquote>

I find it a bit odd that we use alpha and beta as release terms but beta and experimental here. I personally think alpha conveys that something is very "early", though "experimental" is as explicit as it gets, which is also good. 
</blockquote></details>

I'd prefer "experimental" here and keep "beta" for the stage in the release cycle (where it refers to the whole product instead of a feature)

---

*[docs/RFCS/beta_and_experimental_features.md, line 218 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCR8Mmx4LTtSWKJUEc:-KwXO9N_8NZ1CvaA0cXE:bvgp4op) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L218)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Top of mind for me are import CSV and temp storage (which I know isn't experimental, but probably should have been?)
</blockquote></details>

Stats-based rebalancing is another. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

dianasaur323:
It would be great if we could merge this at some point, since we are about to make some decisions based on this RFC!

---

Review status: 0 of 1 files reviewed at latest revision, 37 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KyS3kNpGExKuSWYu9x4:-KyS3kNqweHRWbs1HaNY:bcevcxw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L97)):*
> ```Markdown
> - for the online documentation: be explicit that the feature is experimental.
> - for the admin UI: some discreet UI element that marks the feature as experimental.
> - for SQL language extensions:
> ```

We are about to use this RFC to gate point-in-time restore as experimental. It's a little unclear whether or not backup / restore falls under SQL Language extensions, but I'm interpreting that it does not so as to match how we've treated import csv. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-:-KyS3vCssUREjITu7rtQ:bb3agjv)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 37 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7EoRP1_PE12Hrx_s:-KyvV5NbBthem108y2xC:bt2vr1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L22)):*

repetition of without is awk
</blockquote></details>

Fixed.

---

*[docs/RFCS/beta_and_experimental_features.md, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCMLbSrdGh0tQRuudo:-KyvVHLXmoHdOdMbB96J:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L23)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I think we can also incorporate the concept of building "brand." So for features that don't "make data easy," it's better for us to clearly mark them as experimental so as not to hurt the perception of our brand.
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCN90AG5zgEsWFSviq:-KyvVPtCALgue5m8mcG5:b-uzj4e3) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L28)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I think here, we might also want to include UX experiences that aren't as obvious - like UX that effects the performance of queries or the behavior of the database. Maybe this isn't the right place to put that. I think there are explicit UX locations and implicit UX locations.
</blockquote></details>

This is the discussion about the difference between functional (surface area) and extra-functional (behavior/performance) aspects. So far this RFC was focusing on surface only. Would you prefer that we extend the scope of the RFC for extra-functional aspects too? That may need at least a separate section.

---

*[docs/RFCS/beta_and_experimental_features.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjGOkG1VP2t-bcLBo:-KyvVdf9wMbw5e9iJIiU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L38)):*

This isn't really the place for this discussion, but even though we call it the "admin UI" internally, I'd prefer to shift to calling it the "web UI". Admins also use (and may prefer, where they have overlapping features) the CLI. 
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjKdL4SI9NbCe30Kn:-KyvVsdbcwMrc_1-BsKW:b2scyuu) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L39)):*

-1 on making any guarantees at all about our log messages for the foreseeable future. 
</blockquote></details>

Clarified in paragraph below.

---

*[docs/RFCS/beta_and_experimental_features.md, line 71 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7foggbWhYttxHCdd:-KyvVy54KZL7G8rGlemv:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L71)):*

nit --remove "a" and add a " , "
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 75 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7oLkNh-V2RGCCNBd:-KyvW-HSOBJD5epAs6Du:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L75)):*

s/ guidelines 
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 81 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7snFLvJxyWyyBt4y:-KyvW1Ya5N8BmT9v4Z4a:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L81)):*

new features, first
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOAfd7RcnX-GtUvJE:-KyvW4_2vWZsEhsCzJzs:bcji08b) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L83)):*

In a sense, *everything* is experimental in an alpha. We should aim to identify which features will be experimental in a release before the beta freeze and mark them as such at that time. 
</blockquote></details>

I don't understand the feedback here -- how should I modify the text?

---

*[docs/RFCS/beta_and_experimental_features.md, line 85 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB7xtvMTR7T7Ha47ff:-KyvWBvqqS_XCMqzssIp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L85)):*

awk phrasing of to display
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOKrhMrdSg8CiFO_j:-KyvWCOmPn9K_qe3OhFW:b-mr6opz) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L88)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Perhaps we don't have to worry about this. This is a bit low level.
</blockquote></details>

Excluded from RFC scope as per Ben's remark above.

---

*[docs/RFCS/beta_and_experimental_features.md, line 90 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOPkx6761EUKp8B6K:-KyvWIg6FR1pP5LtC4TI:blqkzzy) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L90)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

@mjibson named the import csv flag experimental.feature_name, so perhaps we should keep this syntax
</blockquote></details>

This is problematic - I think that the state of the feature (enabled/disabled) should be preserved when the feature transitions from experimental to non-experimental. Should we have 2 settings per feature (1 to enable when experimental, 1 to enable when non-experimental) or just 1enable/disable setting per feature, with a product-wide "Accept experimental features"? @dianasaur323 @bdarnell please advise.

---

*[docs/RFCS/beta_and_experimental_features.md, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAdV283YteeGt1OuUZ-r1-95:-KyvWfDZY7a1wVbrw5og:b3w9a3b) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L95)):*
<details><summary><i>Previously, jseldess wrote…</i></summary><blockquote>

We can find a consistent way to do this, perhaps including `Experimental` in the page title, etc. We should probably also add a page explaining what this status means and listing all current experimental features.
</blockquote></details>

Yes I agree. do you suggest to modify the RFC text in any way to reflect this idea?

---

*[docs/RFCS/beta_and_experimental_features.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KyS3kNpGExKuSWYu9x4:-KyvWoEYwqwfmYhhX0wS:bo6yro9) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L97)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

We are about to use this RFC to gate point-in-time restore as experimental. It's a little unclear whether or not backup / restore falls under SQL Language extensions, but I'm interpreting that it does not so as to match how we've treated import csv. 
</blockquote></details>

backup/restore is grandfathered as non-experimental now. New backup/restore options however could be marked as such.

---

*[docs/RFCS/beta_and_experimental_features.md, line 100 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAdV283YteeGt1OuUZ-r1-100:-KyvX0hKEO1DuxGFvCxF:b-hyvli9) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L100)):*
<details><summary><i>Previously, jseldess wrote…</i></summary><blockquote>

This may be excessive in some cases, especially if we want some users to kick the tires.
</blockquote></details>

Modified to restrict docs to the "dev" tree.

---

*[docs/RFCS/beta_and_experimental_features.md, line 102 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCP3bd2iFMpitr2gel:-KyvX5pR_wTtdQc9qCsF:b77nmne) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L102)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Could we actually move this above your list of places where EXPERIMENTAL should be? So, if possible, gate by cluster setting. In addition, do the following bullets. If gating isn't possible, do the following bullets.
</blockquote></details>

Added a forward reference above. Also added a reminder that there are 2 sorts of cluster settings. PTAL

---

*[docs/RFCS/beta_and_experimental_features.md, line 106 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXIh88mNVk-sU78wX2:-KyvY1_lXTuDkLiVC8fe:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L106)):*

These settings default to false, right? No one's going to go in and manually disable a feature to get a "pristine" condition; it's about requiring people to opt in and accept the rough edges or potential future changes to the feature. 
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXJImIXVrZzjuUrkuN:-KyvY5C0qmf4QpbAJmoR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L108)):*

`experimental.name-of-the-feature.enabled`
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 110 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXK2G5fm23_-_4Rabv:-KyvYK1F8fxqUZsMpKxF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L110)):*

I would add:

" When an experimental feature becomes non-experimental, the setting that controlled it will remain  (forever? for N versions?), but it may become an error to attempt to set it to false. This allows cluster setup/testing scripts to enable experimental features that an application requires without breaking when the features lose their experimental status". 
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 124 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXKxm6Z4u0qKF0w_35:-KyvYcAPmIS4ezhZFv2y:b-5jp7rd) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L124)):*

I disagree with the last sentence. We should strive to make changes that are safe enough (and test them enough during the development cycle) that we don't need to expose them behind a flag. Experimental status should be relatively rare. 
</blockquote></details>

Removed the sentence.

---

*[docs/RFCS/beta_and_experimental_features.md, line 136 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8J0jWa6j9Nqtaxev:-KyvYTkHyk5cs9AmAZg7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L136)):*

e.g.,
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 137 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBQTgf6PxJB9yKSkDj:-KyvZa6sRqbpNBGfblXr:baq5riw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L137)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Agreed, I think feature parity will be hard for us
</blockquote></details>

Clarified this part in the light of the compatibility discussion. PTAL.

---

*[docs/RFCS/beta_and_experimental_features.md, line 143 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCPk-ZmxWXw46yzCLo:-KyvZzFs-IhCwxcefW7h:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L143)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I think instead of this, we should say something like:

- If it does not meet the user's needs as specified in the feature's use cases. Or something along those lines. Forgive my inability to state that eloquently.
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 145 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCQ0rC2SVjafIJK3Vc:-Kyv_5d-pBR3VoSksDYQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L145)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

Yeah, here again, perhaps we should focus on "meeting users needs in a way that meets our company-wide standards" versus feature parity
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjaDn5ExRIWQpzFKg:-Kyv_C-bZW9AxiGrXN2k:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L152)):*

the formatting on all these second-level points is odd, particularly this one
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXLfyan8wzNrQf-E5u:-Kyv_dGGdr4nFDU_IKir:b-p9rr8t) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L152)):*

I strongly disagree with the experimental-by-default philosophy. An experimental phase (that spans a release - remember that in an alpha everything's experimental) is pretty expensive since we have to test with and without it, users who are interested in the feature have to deal with extra complexity, etc. Experimental features should be reserved for the relatively few projects that can't be completed in one release cycle. 
</blockquote></details>

Added a clarification about cockroachdb-specific vs. feature parity. PTAL. Note that the RFC can also force the non-experimental nature if needed, but I think the RFC should do this so as to force the discussion to happen.

---

*[docs/RFCS/beta_and_experimental_features.md, line 159 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8UEahNmySYEskqm-:-Kyv_r_I2dOQM3wwIw77:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L159)):*

RFC,
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtW5_oexDbyKc_kR9:-Kyva36ZXN9WQCyvtoBn:b963668) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L193)):*

Yeah, where possible (syntax, flag/env var names, etc) we should put the word "experimental" in the user's face. Cluster settings are only needed for things where otherwise the only way they'd know it's experimental is by reading the docs. 
</blockquote></details>

There's an issue with "EXPERIMENTAL" in the syntax however: whenever we transition away from it we have to keep the keyword in the grammar to support apps that were built while the feature was experimental.

I'm considering a session variable instead. Session (not cluster-wide) so that the things are enabled per-app, and each new dev has to learn the experimental nature of the feature.

---

*[docs/RFCS/beta_and_experimental_features.md, line 205 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjnlIMVNiPrOTqd4m:-KyvaR4YXZJZCfxasbyi:b-pmv6n2) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L205)):*

isn't this exactly what is proposed above? "The cluster setting should be named `experimental.name-of-the-feature`."
</blockquote></details>

I was confused. The section about cluster settings clarifies that there are 2 sorts of cluster settings.

---

*[docs/RFCS/beta_and_experimental_features.md, line 211 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCQrcEKl35ZkBlO0pn:-KyvaX2G3NBXOfsCyo_3:b7rztcf) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L211)):*

If a feature is stabilized without further change, we should make the token `EXPERIMENTAL` optional (at least for a few releases). No reason to actively break applications that were using the experimental syntax. 
</blockquote></details>

I think I'm still uncomfortable about that. What about the session flag as suggested above?

---

*[docs/RFCS/beta_and_experimental_features.md, line 213 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBShjWlgwTVsDGFuya:-KyvadgTfr_QqxU01slE:b-69jk3m) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L213)):*
<details><summary><i>Previously, nstewart (Nate) wrote…</i></summary><blockquote>

I'd like to propose one additional note about positioning beta features: mainly that experimental features still pass our QA process and won't eat your data.  If we want a weaker guarantee that's fine, but whatever our stance is here we need to make that clear
</blockquote></details>

I'm not sure I understand. @nstewart can you clarify what is the concern here? What should I write down?

---

*[docs/RFCS/beta_and_experimental_features.md, line 215 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAdV283YteeGt1OuUZ-r1-215:-KyvatNdJ6vIhsB4P6RT:byym7b4) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L215)):*

I'd prefer "experimental" here and keep "beta" for the stage in the release cycle (where it refers to the whole product instead of a feature)
</blockquote></details>

I have removed the word "beta" throughout.

---

*[docs/RFCS/beta_and_experimental_features.md, line 215 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtbCLsEKgWtSslax9:-Kyvb3CsgvOP3j5GeiKH:beztrku) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L215)):*

so far, we've been using experimental to mean both "beta" like IMPORT CSV and "internal only" like https://github.com/cockroachdb/cockroach/pull/19140. the latter is a particularly illustrative example in that we intend to release sql syntax for setting zone configs at some point, but this is _definitely not_ the syntax we'd use for it when we do
</blockquote></details>

I added a sentence to put internal features out of scope at the beginning.

---

*[docs/RFCS/beta_and_experimental_features.md, line 218 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCR8Mmx4LTtSWKJUEc:-KyvbDToTEpSo9qAhrM3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L218)):*

Stats-based rebalancing is another. 
</blockquote></details>

Done.

---

*[docs/RFCS/beta_and_experimental_features.md, line 220 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwB8tYAqZlsmz-LjlA9:-KyvbGGQsB3j-ECxlPDb:bcw4l1k) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L220)):*

Totally agree--didn't mean to dissuade anyone on the team from raising their hand. I think anyone can (and should) raise concerns, I just wanted someone to have clear responsibility to make this determination so that it's not forgotten. 
</blockquote></details>

Ultimately the QA team should raise flags if a feature wasn't marked experimental but should be, according to the criteria set forth here. However, any person in the production chain (from design to QA) can elevate from non-experimental to experimental upon own initiative, as long as they justify so.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

dianasaur323:
Review status: 0 of 1 files reviewed at latest revision, 37 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/beta_and_experimental_features.md, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCN90AG5zgEsWFSviq:-KywjHWJHYr0ZA_bR14B:b-2qsgn3) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L28)):*

This is the discussion about the difference between functional (surface area) and extra-functional (behavior/performance) aspects. So far this RFC was focusing on surface only. Would you prefer that we extend the scope of the RFC for extra-functional aspects too? That may need at least a separate section.
</blockquote></details>

Hmm... let's not expand the scope of this. Based on how I used the RFC last time, it already has everything we need :)

---

*[docs/RFCS/beta_and_experimental_features.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOAfd7RcnX-GtUvJE:-KywjVuLSwXtJTU4YKar:b-x5nq9f) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L83)):*

I don't understand the feedback here -- how should I modify the text?
</blockquote></details>

I think the specific nit is around the timing of deciding whether or not something should be experimental. You have it as the first step when implementing a feature, which feels too early given that more information will come in throughout the development process

---

*[docs/RFCS/beta_and_experimental_features.md, line 90 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOPkx6761EUKp8B6K:-KywjloLc--d2TzMYkXn:bsid1tm) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L90)):*

This is problematic - I think that the state of the feature (enabled/disabled) should be preserved when the feature transitions from experimental to non-experimental. Should we have 2 settings per feature (1 to enable when experimental, 1 to enable when non-experimental) or just 1enable/disable setting per feature, with a product-wide "Accept experimental features"? @dianasaur323 @bdarnell please advise.
</blockquote></details>

I think it should be typical that when the feature moves out of experimental, it is actually turned on by default. Having a different setting sounds fine for now. I don't think we need to do a product-wide toggle at this point?

---

*[docs/RFCS/beta_and_experimental_features.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KyS3kNpGExKuSWYu9x4:-KywjzNwdLtPgjoGRGw6:b-o44dl6) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L97)):*

backup/restore is grandfathered as non-experimental now. New backup/restore options however could be marked as such.
</blockquote></details>

Well, we have a new backup / restore feature that allows point-in-time restore, and it will be experimental for now. 

---

*[docs/RFCS/beta_and_experimental_features.md, line 102 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCP3bd2iFMpitr2gel:-KywkeesWTpJgDP0llDb:b-in1nxy) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L102)):*

Added a forward reference above. Also added a reminder that there are 2 sorts of cluster settings. PTAL
</blockquote></details>

I can't remember what it used to look like, but it looks fine now?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 33 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/beta_and_experimental_features.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCOAfd7RcnX-GtUvJE:-Kz-On-3URh_KclX5ipZ:b-ngnslo) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L83)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I think the specific nit is around the timing of deciding whether or not something should be experimental. You have it as the first step when implementing a feature, which feels too early given that more information will come in throughout the development process
</blockquote></details>

Yeah, exactly. We can decide whether a new feature warrants the "experimental" caveat once we've built and QA'd it (as long as we do this before the freeze). 

---

*[docs/RFCS/beta_and_experimental_features.md, line 145 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwCQ0rC2SVjafIJK3Vc:-Kz-USd4My64jMawIU9B:b-v7xrmt) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L145)):*

Done.
</blockquote></details>

I think we may want to replace this whole decision tree with "does it meet user's needs in a way that meets our company-wide standards" (including by reference the policies around compatibility with other products, etc). I don't think spelling out a more specific process here is helpful. 

Furthermore, framing the decision in this way also tells us when a feature becomes non-experimental: when it has improved to meet our standards. Otherwise, it's unclear: if a feature is experimental because of the earlier rule that "the part of its UX that aims to correspond to an equivalent feature does not mirror 1-1 the surface aspects (functional) of a competing product", when does it ever leave experimental status? 

---

*[docs/RFCS/beta_and_experimental_features.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwXLfyan8wzNrQf-E5u:-Kz-QbF_EYGJzPgcF1PG:b-xdt2e3) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L152)):*

Added a clarification about cockroachdb-specific vs. feature parity. PTAL. Note that the RFC can also force the non-experimental nature if needed, but I think the RFC should do this so as to force the discussion to happen.
</blockquote></details>

It's still weighted too heavily towards experimental status for my taste. The decision tree to this point in the document is "If the feature has user-facing changes", "If the feature aims for compatibility with another product", "if it does match user expectations", "if the change is not "small"". If it matches user expectations, we can stop there: the feature doesn't need to be experimental. 

The RFC should argue that the feature will meet user expectations, or that any deviation from those expectations is justified. The RFC should only address experimental status if, for some reason, a feature that meets user expectations should nevertheless go through an experimental period, or conversely, if a feature that doesn't meet user expectations should be non-experimental (for example, the partitioning syntax now proposed in the RFC doesn't match that used by postgres, but I think we should still aim to release it as non-experimental). 

---

*[docs/RFCS/beta_and_experimental_features.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwAtW5_oexDbyKc_kR9:-Kz-SgiANJK8ynbEtLGt:b-a3nopz) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L193)):*

There's an issue with "EXPERIMENTAL" in the syntax however: whenever we transition away from it we have to keep the keyword in the grammar to support apps that were built while the feature was experimental.

I'm considering a session variable instead. Session (not cluster-wide) so that the things are enabled per-app, and each new dev has to learn the experimental nature of the feature.
</blockquote></details>

I think if keeping the `EXPERIMENTAL` token as optional is too much of a burden, I'd rather just do cluster settings. Having session variables for some things and cluster settings for others is confusing. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

couchand:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 32 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/beta_and_experimental_features.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-KwBjGOkG1VP2t-bcLBo:-Kz0AENOIe2-iyMyf0CT:b-tikxeb) ([raw file](https://github.com/cockroachdb/cockroach/blob/d2b4c4740093421f0ff01bfa8bb451762b2f8390/docs/RFCS/beta_and_experimental_features.md#L38)):*

Done.
</blockquote></details>

Huh, that seems like a change of terminology that should really be done everywhere at once.  Just changing it here seems to leave things in an inconsistent state.  I'm not sure that I disagree with a name clarification, but I don't think we should have different terms in different docs.

And again, is the term "web console" ever used?  I've never seen that myself, but maybe I've missed it.

---

*[docs/RFCS/beta_and_experimental_features.md, line 210 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175#-Kz0BYzJktEse0kJc94I:-Kz0BYzJktEse0kJc94J:b-6j9kx9) ([raw file](https://github.com/cockroachdb/cockroach/blob/2185f477213c981cbea0bc04714172b4c7314abb/docs/RFCS/beta_and_experimental_features.md#L210)):*
> ```Markdown
> ## Drawbacks
> 
> None known.
> ```

Here's one: it's more effort.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19175)*

knz:
@dianasaur323 do you want to schedule some time together and close the loop on this RFC? I would gladly help you propagate the consensus you have gathered into the rfc document.

knz:
Subsumed by https://github.com/cockroachdb/docs/pull/4326

PR #19176, opened 2017-10-11
sqlccl: support RESTORE DATABASE

dt:


benesch:
:raised_hands: Nice! A few questions and nits below.

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/backup_test.go, line 1927 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBDlZ_n38MnZCbzUgW:-KwBDlZa8JtLIKS188XJ:b-diymd) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/backup_test.go#L1927)):*
> ```Go
> 		) {
> 			t.Fatal(err)
> 		}
> ```

I'd love to also see a test that does something complicated like

```
RESTORE DATABASE d1, DATABASE d2, d3.*, d4.foo
```

---

*[pkg/ccl/sqlccl/restore.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwB9wAtfqQwuBY3p_s4:-KwB9wAugZvklvY-6yts:b-jnahr7) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L116)):*
> ```Go
> 
> 	if len(restoreDBNames) > 0 && renaming {
> 		return nil, errors.Errorf("cannot use %q option when restoring who databases", restoreOptIntoDB)
> ```

"who" -> "whole"

---

*[pkg/ccl/sqlccl/restore.go, line 166 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBB473rmnaH2zhPZNd:-KwBB473rmnaH2zhPZNe:b4w9pct) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L166)):*
> ```Go
> 		// Check that any DBs being restored do _not_ exist.
> 		for name := range restoreDBNames {
> 			existingDatabaseID, err := txn.Get(ctx, sqlbase.MakeNameMetadataKey(0, name))
> ```

`keys.RootNamespaceID` instead of 0, or expose `sql.databaseKey` and use `sql.DatabaseKey{name}.Key()`. Or file a TODO/issue to do a proper audit; I see another example that hardcodes 0 below.

---

*[pkg/ccl/sqlccl/restore.go, line 261 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBAmC8LR_JR8v7onBd:-KwBAmC9IQ7xyWzX6f4P:b-yfgpt5) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L261)):*
> ```Go
> 	// primitive are non-transactional), but this does mean if something fails
> 	// during restore we've "leaked" the IDs, in that the generator will have
> 	// been incremented.
> ```

Can you move these two paragraphs up above

```
for _, db := range restoreDBs {
```

(and change the first sentence to "allocate new IDs for each table and database")

---

*[pkg/ccl/sqlccl/restore.go, line 714 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBCCYQWTyACRX77eNb:-KwBCCYQWTyACRX77eNc:b-vzmrua) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L714)):*
> ```Go
> 		for _, desc := range databases {
> 			wroteDBs[desc.ID] = struct{}{}
> 			log.Infof(ctx, "Writing DB %q to ID %v", desc.Name, desc.ID)
> ```

debugging or intentional?

---

*[pkg/ccl/sqlccl/restore.go, line 730 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBCZhHmdG-WjJopqv0:-KwBCZhHmdG-WjJopqv1:b6u67z) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L730)):*
> ```Go
> 				// TODO(dt): Make this more configurable.
> 				table.Privileges = parentDB.GetPrivileges()
> 			}
> ```

Hmm, seems like we need to handle privileges. What if we're creating a new database in a new cluster with a different set of users?

I vote we adjust the privileges to only allow the restoring user to read/write.

---

*[pkg/ccl/sqlccl/restore.go, line 794 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBD04jxonS8Hs5ZK6T:-KwBD04jxonS8Hs5ZK6U:bgg17rp) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L794)):*
> ```Go
> 		if dbDesc := desc.GetDatabase(); dbDesc != nil {
> 			rewrite, ok := tableRewrites[dbDesc.ID]
> 			if ok {
> ```

nit: you could collapse these two lines

---

*[pkg/ccl/sqlccl/targets.go, line 92 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwB9NOclCxGkVkdUTZS:-KwB9NOclCxGkVkdUTZT:b1cntjj) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/targets.go#L92)):*
> ```Go
> 				if _, ok := explicitlyNamedDBs[normalizedDBName]; ok {
> 					retDBs = append(retDBs, dbDesc)
> 				}
> ```

This would make more sense to me if you moved it up one level of nesting. I know `hasKey(explicitlyNamedDBs, normalizedDBName) ⇒ hasKey(starByDatabase, normalizedDBName)`, but I think it's better not to have to think about that.

---

*[pkg/ccl/sqlccl/targets.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBEUk58MisCNFVn-Qt:-KwBEUk58MisCNFVn-Qu:b-l3b5d8) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/targets.go#L132)):*
> ```Go
> 			return nil, nil, errors.Errorf(`database "%s" does not exist`, dbName)
> 		}
> 	}
> ```

Would be great if you added a comment here to the effect of "no need to verify explicitlyNamedDBs, because it's always a subset of starByDatabase"

---

*[pkg/ccl/sqlccl/targets_test.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwB7iyJnbfL0aJIEp-v:-KwB7iyJnbfL0aJIEp-w:b-1nn1ds) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/targets_test.go#L98)):*
> ```Go
> 			targets := stmt.(*parser.Grant).Targets
> 
> 			matched, _, err := descriptorsMatchingTargets(test.sessionDatabase, descriptors, targets)
> ```

This test probably deserves more love.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-:-KwBE-JO4YfC0xqwnHSi:b-p2tr78)*

dt:

---

*[pkg/ccl/sqlccl/backup_test.go, line 1927 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBDlZ_n38MnZCbzUgW:-KwBF05V-TV-FuvxJunO:bc9qggh) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/backup_test.go#L1927)):*

I'd love to also see a test that does something complicated like

```
RESTORE DATABASE d1, DATABASE d2, d3.*, d4.foo
```
</blockquote></details>

You can't currently mix them, since currently `targets` is defined as _either_ `DATABASE name_list` or `table_pattern_list`.

---

*[pkg/ccl/sqlccl/restore.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwB9wAtfqQwuBY3p_s4:-KwBG3HI8AnRzPWdk5mB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L116)):*

"who" -> "whole"
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/restore.go, line 166 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBB473rmnaH2zhPZNd:-KwBFQRfVKlarRLmT9fX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L166)):*

`keys.RootNamespaceID` instead of 0, or expose `sql.databaseKey` and use `sql.DatabaseKey{name}.Key()`. Or file a TODO/issue to do a proper audit; I see another example that hardcodes 0 below.
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/restore.go, line 261 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBAmC8LR_JR8v7onBd:-KwBG205QVP7JKUE9aW5:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L261)):*

Can you move these two paragraphs up above

```
for _, db := range restoreDBs {
```

(and change the first sentence to "allocate new IDs for each table and database")
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/restore.go, line 714 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBCCYQWTyACRX77eNb:-KwBFbL7H7iJ2wWTXs7Z:byy1q1m) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L714)):*

debugging or intentional?
</blockquote></details>

Whoops, thanks.
Done.

---

*[pkg/ccl/sqlccl/restore.go, line 794 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBD04jxonS8Hs5ZK6T:-KwBFnVXnbRVJrBH0Y3j:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L794)):*

nit: you could collapse these two lines
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176)*

dt:
Review status: 3 of 5 files reviewed at latest revision, 10 unresolved discussions.

---

*[pkg/ccl/sqlccl/restore.go, line 166 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBB473rmnaH2zhPZNd:-KwBIrdPHrY0eWs6bdpw:bcgnkny) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L166)):*

Done.
</blockquote></details>

keys.RootNamespaceID is sufficiently widespread that i'm not too worried about encoding that logic here vs reusing databaseKey's logic -- IF someone decides they want to change it, they'll have to do that audit anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176)*

benesch:
Reviewed 2 of 2 files at r2.

---

*[pkg/ccl/sqlccl/backup_test.go, line 1927 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBDlZ_n38MnZCbzUgW:-KwBKd6MeIDDu_q4Mlmv:b8lu9sh) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/backup_test.go#L1927)):*

You can't currently mix them, since currently `targets` is defined as _either_ `DATABASE name_list` or `table_pattern_list`.
</blockquote></details>

Well that makes it easy!

---

*[pkg/ccl/sqlccl/restore.go, line 166 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBB473rmnaH2zhPZNd:-KwBKh24CIko0QnYiI_f:b4c9lyr) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L166)):*

keys.RootNamespaceID is sufficiently widespread that i'm not too worried about encoding that logic here vs reusing databaseKey's logic -- IF someone decides they want to change it, they'll have to do that audit anyway.
</blockquote></details>

:thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176)*

dt:

---

*[pkg/ccl/sqlccl/restore.go, line 730 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBCZhHmdG-WjJopqv0:-KwBLZdzxqxWpfidmcx_:b9n2jfa) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L730)):*

Hmm, seems like we need to handle privileges. What if we're creating a new database in a new cluster with a different set of users?

I vote we adjust the privileges to only allow the restoring user to read/write.
</blockquote></details>

Ideally we want `BACKUP DATABASE foo` and `RESTORE DATABASE foo` to Just Work as much as is possible. Recreating the db itself for you is one step towards that, and I think restoring privileges is another -- if we don't restore the grants as they were when backed up, we haven't really achieved our goal of removing to the backed up state.

On the other hand, we have no idea if the users to which privs are granted even exist, so maybe that doesn't matter. We might want to add an explicit restore users command that restores the backed up users and their privileges. This might also be easier if we a) kept grants on the user, rather than on the table b) had roles.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176)*

dt:
Review status: 2 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/ccl/sqlccl/restore.go, line 730 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBCZhHmdG-WjJopqv0:-KwBTCAbcI_pgqtGQMW_:b-6le34e) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L730)):*

Ideally we want `BACKUP DATABASE foo` and `RESTORE DATABASE foo` to Just Work as much as is possible. Recreating the db itself for you is one step towards that, and I think restoring privileges is another -- if we don't restore the grants as they were when backed up, we haven't really achieved our goal of removing to the backed up state.

On the other hand, we have no idea if the users to which privs are granted even exist, so maybe that doesn't matter. We might want to add an explicit restore users command that restores the backed up users and their privileges. This might also be easier if we a) kept grants on the user, rather than on the table b) had roles.
</blockquote></details>

okay, just blanking privs with the default for now, will revisit in a TODO.

---

*[pkg/ccl/sqlccl/targets.go, line 92 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwB9NOclCxGkVkdUTZS:-KwBQRonsvSWCuxhq4-b:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/targets.go#L92)):*

This would make more sense to me if you moved it up one level of nesting. I know `hasKey(explicitlyNamedDBs, normalizedDBName) ⇒ hasKey(starByDatabase, normalizedDBName)`, but I think it's better not to have to think about that.
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/targets.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBEUk58MisCNFVn-Qt:-KwBQSA5jWFx-VGtSMIw:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/targets.go#L132)):*

Would be great if you added a comment here to the effect of "no need to verify explicitlyNamedDBs, because it's always a subset of starByDatabase"
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/targets_test.go, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwB7iyJnbfL0aJIEp-v:-KwBTAeYoJkjv301bEWh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/targets_test.go#L98)):*

This test probably deserves more love.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176)*

benesch:
:lgtm:

---

Reviewed 3 of 3 files at r3.

---

*[pkg/ccl/sqlccl/restore.go, line 730 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-KwBCZhHmdG-WjJopqv0:-KwBVJho98ixmN_RnDm6:b3pdlem) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd98a25b6561096c186b6f8cd117587df781a283/pkg/ccl/sqlccl/restore.go#L730)):*

okay, just blanking privs with the default for now, will revisit in a TODO.
</blockquote></details>

:thumbsup: yeah, I think we need a better story around actually restoring the `system.users` table we include in every backup before we do anything smarter with privileges.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19176#-:-KwBVY4bZrSHUwLh9Eq3:bnfp4nl)*

PR #19178, opened 2017-10-11
docs/generated: temporarily disable automatic diagram generation

dt:
disabling it until we work out a better way for anyone to generate them, as the webservice results apparently no long match our local jar's output.

### DIFF
--- a/docs/generated/sql/diagrams/doc.go
+++ b/docs/generated/sql/diagrams/doc.go
@@ -14,4 +14,5 @@
 
 package diagrams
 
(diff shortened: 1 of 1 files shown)


dt:
huh, looks like the CLA bot and the reviewable comments got lost in the gh outage.

benesch (approved):
:(

dt:
@benesch I plan to bring it back shortly once it gets a little more polish, just want to make sure it doesn't get in anyone's way with spurious build failures in the meantime.

benesch:
Totally! Just ensaddened by the amount of pain these diagrams cause.

PR #19182, opened 2017-10-11
sql: revert the default behavior of DROP DATABASE to CASCADE

### LINKED ISSUE #19145: CDB 1.1, Drop Database with Restrict by default is breaking migration tools
With CDB 1.1, the ` DROP DATABASE` which was using `CASCADE` by default, now use `RESTRICT` so you must use `DROP DATABASE xxx CASCADE` to effectively drop a database that is not empty.

I understand the intention behind this change but at the same time, this breaks pretty much all the migration tools out there...

With Elixir / Ecto, I have no easy way to customize the statement that is being used to drop the database.

I suspect the same will happen with any other migration tools and I cant event submit a patch to Ecto so that we use cascade explicitly because it's not supported by Postgresql itself. The only option would be to fork the Postgresql adaptor and maintain it :(

I remember I think that @bdarnell voiced to have the same behaviour both in the CLI and with a connection directly but I wonder if it would not be better to only use `RESTRICT` by default in the CLI as to avoid any manual mistake but keep the `CASCADE` default otherwise. That might be a middle ground that is providing additional value and safety and at the same time, does not break migration tools.

knz:
... because that's what DROP DATABASE does in postgres, and schema
migration tools expect it to work this way.

However, in order to avoid unpleasant surprises by user, make
DROP DATABASE *without* behavior specifier also error out
when the session variable `sql_safe_updates` is set. This is
the case e.g. in interactive shells by default. For example:

```
root@:26257/> drop database t;
pq: rejected: DROP DATABASE on non-empty database without explicit CASCADE (sql_safe_updates = true)
```

Fixes #19145.

cc @tlvenn @bdarnell

### DIFF
--- a/pkg/sql/drop.go
+++ b/pkg/sql/drop.go
@@ -71,10 +71,20 @@ func (p *planner) DropDatabase(ctx context.Context, n *parser.DropDatabase) (pla
 		return nil, err
 	}
 
-	if len(tbNames) > 0 && n.DropBehavior != parser.DropCascade {
-		return nil, pgerror.NewErrorf(pgerror.CodeDependentObjectsStillExistError,
-			"database %q is not empty and CASCADE was not specified",
-			parser.ErrString(parser.Name(dbDesc.Name)))
+	if len(tbNames) > 0 {
+		switch n.DropBehavior {
+		case parser.DropRestrict:
+			return nil, pgerror.NewErrorf(pgerror.CodeDependentObjectsStillExistError,
+				"database %q is not empty and RESTRICT was specified",
+				parser.ErrString(parser.Name(dbDesc.Name)))
+		case parser.DropDefault:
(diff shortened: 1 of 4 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19182#-:-KwCZ7J9dW_JmqtqyiSM:bnfp4nl)*

jordanlewis:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19182#-:-KwCbuNE9w-eoI2WkFH5:bnfp4nl)*

knz:
TFYRs!

PR #19183, opened 2017-10-11
sql: Remove TODO about checking that indexes are unique

lgo:
We do not want to check for duplicate indexes as they have a use case,
as mentioned in #19166. In short, they can be used for non-partitioned
zoned indexes. There may also be cases where duplicated indexes are made
by ORMs (this happens for constraints,
https://github.com/cockroachdb/sequelize-cockroachdb/issues/18).


danhhz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19183#-:-KwBTFVuNZqHBr3K-YBE:bnfp4nl)*

PR #17877, opened 2017-08-23
sqlccl: add jobs records for csv import

dt:
no progress tracking or resuability stuff yet, just informational records to start with

PR #19185, opened 2017-10-11
style guide: suggest inline comments for literals passed as function args

andreimatei:
- suggest commenting inline literals, and suggest a format for it
- standardize the style when such comments are used for bool constant
- give guidance on avoiding bool parameters

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -56,6 +56,89 @@ short and related arguments (e.g. `start, end int64`) should either go on the sa
 or the type should be repeated on each line -- no argument should appear by itself
 on a line with no type (confusing and brittle when edited).
 
+### Inline comments for arguments to function calls
+
+A code reader encountering a function call should be able to intuit what all
+the arguments to the call represent. Whenever it wouldn't be otherwise clear
+what the value used as an argument represents (for example, from the variable's
+name if a variable is used or from the type name if a struct literal is used),
+consider annotating it with an inline comment specifying the respective
+parameter's name. Particularly, consider doing this for literals of "basic"
+types (boolean, numeric, string types, whether the type is predeclared or not)
+and for `nil` identifiers, as they are frequently not suggestive enough of what
+they represent.
+
+For example:
+
+```
+intentsToEvalResult(externalIntents, args, false /* alwaysReturn */)
+
+monitor := mon.MakeMonitor(
+  "in-mem temp storage",
+  mon.MemoryResource,
+  nil,             /* curCount */
+  nil,             /* maxHist */
+  1024*1024,       /* increment */
+  maxSizeBytes/10, /* noteworthy */
+)
+```
+
+Note: For `bool` constants, like for all literals, the comment should indicate
+the name of the parameter and does not depend on the argument value.
+*Do not* put a bang in the comment when commenting the `false` constant. Also,
+do not adapt the comment to negate the name of the parameter. For example:
+
+```
+func endTxn(commit bool){}
+
+OK:     endTxn(false /* commit */)
+NOT OK: endTxn(false /* !commit */)
+NOT OK: endTxn(false /* abort */)
+// If you want to add an explanation to an argument, a suggested style is to
+// include both the param name and the explanation with a dash between them:
+OK:     endTxn(false /* commit - we abort as we concluded above that we can't commit */)
+```
+
+#### Try to avoid `bool` parameters
+
+`bool` arguments to functions are often dubious things, as they hint to code that
+essentially reads like:
+
+```
+func doSomething(shouldDoX bool) {
+  if shouldDoX {
+    doX()
+  } else {
+    doY()
+  }
+}
+```
+
+This is not always the case. However, in cases where that is a fair assessment
+of the situation, consider whether the `doSomething` function should exist at
+all.  
+In cases where the `bool` in question, along with other arguments, acts as a
+"knob" to the function consider replacing it with some type of "configuration"
+struct (for examples, see [Dave Cheney's treatment of the
+topic](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)).
+In situations where there's a single `bool` param or the situation is less
+clear-cut, consider replacing the `bool` in question with an enum. For example:
+
+```
+type EndTxnAction bool
+
+const (
+  Commit EndTxnAction = false
+  Abort = true
+)
+
+func endTxn(action EndTxnAction) {}
+```
+is better than
+```
+func endTxn(commit bool) {}
+```
+
 ### fmt Verbs
 
 Prefer the most specific verb for your use. In other words, prefer to avoid %v


andreimatei:
I've reached a breaking point with both asking people to comment
literals passed as arguments to functions, and being told what the
comment form for bool constants should be, so I think it's quite
important we standardize (and mandates some of) these comments.

The biggest goal of the proposed addition is basically mandating that endTxn(false) be typed as endTxn(false /* commit */).

I hope that this goal is not controversial. Assuming that proves to be true, I'm interested in bikeshedding the details and see if anybody wants to contribute anything.

STYLE.md
```diff
@@ -56,6 +56,73 @@ short and related arguments (e.g. `start, end int64`) should either go on the sa
...
+function arguments have to be accompanied by an inline comment containing the
+name of the parameter, to aid code readability. Similarly, the `nil` identifier
+needs to be commented when used in this way.
+For example:
+
+```
+x := intentsToEvalResult(externalIntents, args, false /* alwaysReturn */)
+```
+
+Literals used as return values for function with multiple named results have to
```
tamird: s/function/functions/, s/have to/must/

STYLE.md
```diff
@@ -56,6 +56,73 @@ short and related arguments (e.g. `start, end int64`) should either go on the sa
...
+func endTxn(commit bool){}
+
+OK:     endTxn(false /* commit */)
+NOT OK: endTxn(false /* !commit */)
+NOT OK: endTxn(false /* abort */)
+```
+
+Exceptions:
+1. `nil` errors shouldn't be commented when they are the last return value.
+2. `nils` and other zero vals and sentinels shouldn't be commented in return
```
tamird: s/\`nils\`/\`nil\`s/

tamird (commented):
Seems fine, but I wouldn't do this without providing the necessary tooling to enforce (and the requisite boiling the ocean).

Another thing to consider is that this can be done entirely by tooling - VSCode does this for you when it renders the function call.

I don't have an opinion on this proposal, just did a proof-reading pass.

dt:
I'd argue that 9 times out of 10, adding a named constant in the package containing the function being called is more readable than using the literal value everywhere with a comment explaining what it is.
e.g. instead of `foo.Bar(baz, nil /* bazFetcher */)` add a `var DefaultBazFetcher` to foo and then use `foo.Bar(baz, foo.DefaultBazFetcher)`

andreimatei:
> Seems fine, but I wouldn't do this without providing the necessary tooling to enforce (and the requisite boiling the ocean).

I personally wouldn't know how to add the linting. But I don't think that should be a prerequisite for this kind of change. Even without 100% coverage, if putting it in the style guide tips the scale and causing more people to do it (and more reviewers to complain when other don't do it), the world will be better.
The only boiling of the ocean I plan on doing if for bools commented as `foo(false /* !param */).

> Another thing to consider is that this can be done entirely by tooling - VSCode does this for you when it renders the function call.

This is an argument I've heard once before and one of the goals here is to invalidate it. The fact that one editor does something smart when displaying code does not mean that others (and github, where I also read code) do it.

andreimatei:
> I'd argue that 9 times out of 10, adding a named constant in the package containing the function being called is more readable than using the literal value everywhere with a comment explaining what it is.

Yes, absolutely. I wanted to give some guidance to that effect but forgot. Will do. However, I don't want to (and also don't know how I would even if I wanted to) mandate that, and so the current proposal also stands.

danhhz:
I'm :-1: on this. Overall it just feels like too much work for too little benefit and it's just asking for rot when the name of the parameter gets changed.

@dt's proposal eliminates the possibility of rot, so I'd be more willing to accept something like that, but I still don't feel this needs to be mandated.

andreimatei:
> Overall it just feels like too much work for too little benefit and it's just asking for rot when the name of the parameter gets changed.

Even if the names of arguments change, a call-site with the old argument name is still better in my opinion than a call-site with a naked literal. If the semantics of the argument changed, than the call-sites must have also changed, so the comments would naturally be updated.

andreimatei:
@dt I've added some guidance for replacing `bool` args with enums.
I didn't do anything more general because I personally don't have any clear guidance to write; I think a lot of times passing `nil` as a sentinel is convenient and good enough. If you want to phrase something more general, about default values or even more general than that, tied to this PR or not, I'd be interested to read it.

dt:
to clarify, I wasn't suggesting that one must always use a full new enum type, or limiting my comment to bools -- I was just pointing out that named constants (of any type) are more robust than literals with inline comment labels (e.g. the compiler checks them for one).

petermattis:

---

*[STYLE.md, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwBr0mGfuhlwF7xYJzL:-KwBr0mGfuhlwF7xYJzM:b4k5xql) ([raw file](https://github.com/cockroachdb/cockroach/blob/4ae0c60a31cc8e4959459ba36d255892b1ca51c6/STYLE.md#L62)):*
> ```Markdown
> 
> Literals of "basic" types (predeclared boolean, numeric, string types) used as
> function arguments have to be accompanied by an inline comment containing the
> ```

I hesitate to require this style when we have so many violations in our code base. I'd be fine if this said `arguments should be accompanied`. You'd have to adjust the rest of the wording to use `should` instead of requiring this style.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185)*

andreimatei:

---

*[STYLE.md, line 71 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwBeqdRQZ61QY2fHwix-r1-71:-KwBrmDIgUhOT_o9ia3z:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b8f35b3412bdcf85d0a829602b68c740e2d898c4/STYLE.md#L71)):*

s/function/functions/, s/have to/must/
</blockquote></details>

Done.

---

*[STYLE.md, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwBeqdRQZ61QY2fHwix-r1-95:-KwBrmh6t7vISLDqm1PG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b8f35b3412bdcf85d0a829602b68c740e2d898c4/STYLE.md#L95)):*

s/\`nils\`/\`nil\`s/
</blockquote></details>

Done.

---

*[STYLE.md, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwBr0mGfuhlwF7xYJzL:-KwBrvvpWqpVn4JM3bTy:b-1iqijv) ([raw file](https://github.com/cockroachdb/cockroach/blob/4ae0c60a31cc8e4959459ba36d255892b1ca51c6/STYLE.md#L62)):*

I hesitate to require this style when we have so many violations in our code base. I'd be fine if this said `arguments should be accompanied`. You'd have to adjust the rest of the wording to use `should` instead of requiring this style.
</blockquote></details>

I'd be fine with this if "should" is understood as defined as in IETF RFCs:
```
This word, or the adjective "RECOMMENDED", mean that there
may exist valid reasons in particular circumstances to ignore a
particular item, but the full implications must be understood and
carefully weighed before choosing a different course.
```
and so this would continue to mean that new code has to be written with this rule in mind, literals are by default expected to have the comment, and reviewers are supposed to complain when the rule is violated without a "valid reason" existing.

Is that what you had in mind?
I guess I should have never used the construct "have to". I should have used "should" or "must", as Tamir pointed out below. And out of the two, I also believe that "must" is the wrong one - very few style rules are absolute, as the definition of that word requires. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185)*

jordanlewis:
I don't have a horse in this race. Personally I prefer not to have the inline comments, since my editor annotates the parameter names in those cases, but I understand that not all editors and viewers do that.  +1 on David's suggestion to add enums whenever possible. I think this is a reasonable amendment, especially if the purpose here is to have some standard documentation to refer to for newcomers rather than a 100% must-do stylistic rule.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwBtbXHKcYVgd4V_1DE:be4s0ip)*

andreimatei:
@dt I understand, but I still don't know how to phrase that well enough for this doc. But if you do / want to try...
Btw, the "compiler checks" part is not as true as one would think. For example, the following compiles:

```
type X int

func foo(x X) {}

func main() {
	foo(1)
}
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwBtWLW_ViNhGKOfnTN:bkm1e83)*

danhhz:
I just reread this more closely and it's way broader in scope than I thought from my original scan. With the exception of "Try to avoid `bool` parameters" I'm strongly :-1: on all of this. I'm not convinced of the readability improvements (having all these inline comments everywhere is going to make the code much _harder_ to read IMO). It's also a tremendous maintenance burden.

Perhaps if these were presented as a list of tools that might situationally help an engineer write clearer code, than I'd be happier with it. But even "recommendations" or "should" is going too far

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwBtLK1wEDDJB2Qh9sm:bf60bvi)*

dt:
@andreimatei i just meant that inline comments aren't checked by the compiler -- they can rot, be misspelled, etc. If you choose to define a constant `const DoFoo = true` and then use `pkg.DoFoo` instead of `true /* foo */`, if someone refactors `pkg` and changes the signatures, they'll also update your callsite.

Anyway, I like enums but they're sometimes heavier than justified. I like named constants more than literals with inline comment labels. Which approach makes sense though seems like it depends on the context -- sometimes a single bool or int arg is obvious without any of this. I think I agree with @danhhz that this might make more sense as "some patterns that may improve readability" as opposed to hard and fast rules.

andreimatei:
@danhhz can you please explain a bit more what you're thinking, please? How is `foo(false /* arg */`) *harder* to read than `foo(false)`? The latter is utterly *unreadable*.
You say for the second time that it's a "maintenance burden", but you didn't respond to my argument that no maintenance is required. Even if the comments rot, they're still better than no comment (except if the argument semantics changed, but then the call-sites must also change anyway).

@jordanlewis 
> I think this is a reasonable amendment, especially if the purpose here is to have some standard documentation to refer to for newcomers rather than a 100% must-do stylistic rule.

My intention is to make it a 100% must-do unless you have a reason not to :) Choosing to not do it would be an explicit choice, on one (or a related set of) cases.
Like, I would like it to be a rule, not something that you read once when you join the company and promptly forget.
Nobody's gonna say anything if both you and the reviewer missed one (particularly since there's no shamecube any more), but that's not the desired outcome.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwBuvrNtzSg1lT02FNG:ba7jpb5)*

andreimatei:
> I think I agree with @danhhz that this might make more sense as "some patterns that may improve readability" as opposed to hard and fast rules.

Well, if it were to be phrased as guidance, I'd need some examples about when it's not necessary, and I'm hard pressed to come up with a single one (perhaps other than `fibonacci(10)`). Since I think it's almost always necessary and, moreover, necessary to the point where I want to complain in almost every instance where it's not done, I think that makes it a desirable rule.
If you give me some examples where it's overkill and I buy them, then I'd definitely either add more exemptions or relax the language.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwBx2emjybt-EZvTgvA:bqdvvw)*

jordanlewis:
Here are some cases where the rule seems like overkill:

`foo.SetSomeRetryField(30)`
`fmt.Println("hi")`
`NewDString("andrei")`
`NewUndefinedDatabaseError("blah")`
`Compare(x, 100)`
`stream.Peek(5)`

I'll note that the main good examples I can think of are single-argument functions.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwByVf_mnyOxW5j4sH_:b-c0ykjq)*

andreimatei:
Thanks @jordanlewis. I've added those and others to the list of exemptions. Please take a look and let me know if you think it's sufficient.
I'm still reticent to weaken the general language because I don't know how to do it exactly and I'm afraid that the whole thing would not be taken as a "rule" any more. But if others feel strongly and would then support the change, I'm opened to figuring something out. Also, if someone else wants to propose specific weaker language, I'd be cool with it as long as the things that bother me remain clearly illegal (e.g. `endTxnArgs(txn, false)` and
```
		monitor = mon.MakeMonitor(
			"temp disk storage",
			mon.DiskResource,
			nil,
			nil,
			64*1024*1024,
			64*1024*1024/10,
		)
```
need to be illegal).

arjunravinarayan:
I'm a strong 👎 on this. You should use an editor that does this for you. The tooling exists that solves this problem locally. You're just choosing not to use them. Right now those of us with sane setups are punished with code that look like this:

screenshot 2017-10-11 18 26 11

I'm fine with @dt's proposal. Bool constants are probably not worth keeping anyway, unless the function name just makes it blindingly obvious (e.g. `setFooAutoRetryMode(false)`). 

I understand that this is a passion of yours so I'm not going to fight it, and I'll just live with the consequences of whatever comes out of this bikeshed, but I want to make my position clear that I think you're tilting at windmills.

andreimatei:
> I'm a strong 👎 on this. You should use an editor that does this for you. The tooling exists that solves this problem locally. You're just choosing not to use them.

Arjun, if your editor could run code formatters for you when you load a file (hell, maybe it can), would you argue that we should have no formatting rules and that I should use that editor? The fact that our code can be successfully edited on a typewriter is a great thing; do you really want to argue against it? Also, like I said before, I think it's fair to say that Github is a common base for all of us when it comes to reading code, and it doesn't do what your editor does. May I propose a different solution to the punishment: if this rule was in effect, you could disable that feature of your editor and you'd be better of for it - as it would also remove the tooltips from the many literals that are exempt from this rule and where they're unnecessary; every time I look at the screen with one of these fancy editors I'm in fact bothered by the littering of tooltips in places where they're not necessary.
If you do believe you have a point, would you then argue that the inline comments should *never* be  written (and mandate that)? Because, in my opinion, it's gotta be one or the other - either we think they're good or we think they're bad.

> I'm fine with @dt's proposal. Bool constants are probably not worth keeping anyway, unless the function name just makes it blindingly obvious (e.g. setFooAutoRetryMode(false)).

What is @dt's proposal that you're fine with, exactly? @dt has said that named constants are frequently a good idea. Great! If someone has a coherent phrasing for that that we can put in as *guidance*, then I'm all for it. But short of prohibiting all `nil`s and basic literals that this proposals wants commented (which I don't think @dt was proposing, and I don't think is feasible or even a good idea), the issue that this proposal is trying to address remains. I don't see how saying "let's do @dt's thing, instead" follows. I may be wrong somehow, though.

knz:
I'm mildly in favor of (some of) these recommendations, based on the argument that it will greatly help during reviews. I have had the pain that Andrei has shared here myself before and I would like that we start doing something about it.

I do however wish that Andrei would come with an extension to (say) crlfmt or some other tool that would help the user by suggesting the comment additions where they are missing, and only introduce this style guide change *after* we have a tool that can lint/enforce it.

---

Reviewed 1 of 1 files at r3.

---

*[STYLE.md, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwErc1lAqS3mr-UwL8d:-KwErc1miUa1oen0zfPZ:biflxmp) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L61)):*
> ```Markdown
> ### Inline comments for literals
> 
> Literals of "basic" types (predeclared boolean, numeric, string types) used as
> ```

s/Literals of ... used as/Literals used as/

All literals should be subject to this rule, regardless of type.

---

*[STYLE.md, line 64 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwErusSoD12RhA9zaiF:-KwErusTILFepYdE4IQi:b-c8dyki) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L64)):*
> ```Markdown
> function arguments should be accompanied by an inline comment containing the
> name of the parameter, to aid code readability. Similarly, the `nil` identifier
> needs to be commented when used in this way.
> ```

As well as anonymous closures.

---

*[STYLE.md, line 76 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEs1zY2MRe8xeQou43:-KwEs1zZ65mYqNVVxwsc:bbb89p8) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L76)):*
> ```Markdown
> ```
> func countVals() (numVals int, numIntents int) {
>   return 5 /* numVals */, 1 /* numIntents */
> ```

I like the argument rule but this is just overkill. If the function is long enough that you have to scroll to find its prototype, then the function is too long and should be split into sub-functions instead. This rule here is merely hiding the real problem.

---

*[STYLE.md, line 83 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEsLQIpNUx2TzTbK0E:-KwEsLQJRHgcfcNwZPPl:bsdt3uy) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L83)):*
> ```Markdown
> name of the argument / result and does not depend on the argument value. *Do
> not* put a bang in the comment when commenting the `false` constant. Also, do
> not adapt the comment to negate the name of the parameter. For example:
> ```

Please add a forward reference here to the paragraph below where you explain that an additional explanation is also acceptable.

---

*[STYLE.md, line 84 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEt0VAOyw2RwrSH8P6:-KwEt0VAOyw2RwrSH8P7:b-w6low) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L84)):*
> ```Markdown
> not* put a bang in the comment when commenting the `false` constant. Also, do
> not adapt the comment to negate the name of the parameter. For example:
> ```
> ```

Space missing between paragraph and example.

---

*[STYLE.md, line 92 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEt3E5kGTKWuVrk57a:-KwEt3E6gqFhIfT2jdHL:b3wjkaz) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L92)):*
> ```Markdown
> ```
> 
> Exceptions:
> ```

Space missing between "Exceptions:" and enumeration list.

---

*[STYLE.md, line 106 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEsZ2p3cfMNbfakQ8F:-KwEsZ2p3cfMNbfakQ8G:bhyzl6t) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L106)):*
> ```Markdown
>     errors.Wrap(err, "my message")
>     testutils.IsError(err, "job paused")
>     util.Pluralize(5)
> ```

Also: The format argument to all formatting functions.

---

*[STYLE.md, line 126 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEsoKkHej09mizZjoP:-KwEsoKkHej09mizZjoQ:b4ewgi2) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L126)):*
> ```Markdown
>     ```
> 
> 8. Literal format strings containing format verbs (i.e. `%s,%#v, etc`) are
> ```

please: 
```
(i.e.  `%s`, `%#v`, etc)
```

---

*[STYLE.md, line 130 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEslHtdc4UTvXvGkWA:-KwEslHuvihnB0JuJKMU:brux47a) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L130)):*
> ```Markdown
> 
>     ```
>     log.Infof(ctx, "reading key:%d", key)
> ```

please add a space between `:` and `%d` in the example!

---

*[STYLE.md, line 135 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEt7Dg3853YNcUA-h0:-KwEt7Dg3853YNcUA-h1:bco5blh) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L135)):*
> ```Markdown
> 
> Although  not required, consider using inline comments with names similar in
> spirit to the rules above whenever they help readability. In particular:
> ```

Space missing between "In particular:" and enumeration list.

---

*[STYLE.md, line 154 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtCFO4MSdiBIaXP57:-KwEtCFO4MSdiBIaXP58:b-w6low) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L154)):*
> ```Markdown
> way that makes sense at a particular site. When placed inline, the format for
> the explanation is `<literal> /* <param name> - explanation */`. For example:
> ```
> ```

Space missing between paragraph and example.

---

*[STYLE.md, line 162 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtQrZuxQwYBZ89LbV:-KwEtQrZuxQwYBZ89LbW:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L162)):*
> ```Markdown
> `bool` arguments to functions are often dubious things, as they hint to code that
> essentially reads like:
> ```
> ```

ditto

---

*[STYLE.md, line 175 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtaHWKVnlm0c4GhC4:-KwEtaHWKVnlm0c4GhC5:b-cpb6v9) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L175)):*
> ```Markdown
> However, in cases where that is a fair assessment of the situation, consider
> whether the `doSomething` function should exist at all. In situations where the
> situation is less clear-cut, consider replacing the `bool` in question with an
> ```

Before you introduce this rule please look at the generated assembly for a couple examples. I have a hunch that Go does not yet optimize two-valued enums as well as booleans. It would be a shame if this recommendation would make our code slower overall.

---

*[STYLE.md, line 177 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEt_YmZFVy12fTtgsx:-KwEt_Yn4W8U7_Qdq3Rr:b-w6low) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L177)):*
> ```Markdown
> situation is less clear-cut, consider replacing the `bool` in question with an
> enum. For example:
> ```
> ```

Space missing between paragraph and example.

---

*[STYLE.md, line 188 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtZRJ4egKDVJQV6RU:-KwEtZRJ4egKDVJQV6RV:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L188)):*
> ```Markdown
> ```
> is better than
> ```
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwEtoAzxvMSVFupkAfT:b-ycj504)*

tschottdorf:
I'm also :thumbsdown: on the (too constraining) specifics in this proposal. I don't necessarily buy @arjunravinarayan's editor argument since PRs don't support any of that, but I can't be in favor of anything that turns contributors into human code generators. On the other hand, if you can automate whatever you'd like to see by extending `crlfmt`, then good chance I won't mind, but that's definitely not possible with your rules below.

In the absence of that, I'm in favor of adding general (soft) guidance against methods that take various knobs, along with model solutions (options struct, named const) that reviewers can link to (after all, the bike shed discussion is often what we want to avoid), and I think for myself I've decided that I'll never write another method invocation that needs an arg-inline comment (famous last words). Obviating these in the first place is better than creating conventions around them in my opinion.

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwFSqDrjwDQkYOV_T8O:b-rwxkta)*

andreimatei:
> for myself I've decided that I'll never write another method invocation that needs an arg-inline comment 

Are you saying that you won't ever write a function taking a nil-able argument and pass the literal `nil`? Or are you saying that in the case where you will, the `nil` would still not need a comment, for some reason?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwFXp5EHSsA00agsPZE:b-wt3p6)*

andreimatei:
> I'm also 👎 on the (too constraining) specifics in this proposal.

This must mean that the proposal is forcing you to put comments on lines where you wouldn't want to see them in code. Can you give an example? I only want this proposal to force you to write comments on lines where it's not controversial that they're needed (definitely don't wanna err on the "too constraining" side), so I'd tweak it until no counter-example can be found (this doesn't apply to the people on the "they're never needed because of my editor" camp).

tschottdorf:
> Are you saying that you won't ever write a function taking a nil-able argument and pass the literal nil? Or are you saying that in the case where you will, the nil would still not need a comment, for some reason?

I was promising to hold myself to a standard in which method signatures are straightforward enough to not require inline comments, and in cases where that's not true and I can't reasonably change it, to use a const or an options struct.

> This must mean that the proposal is forcing you to put comments on lines where you wouldn't want to see them in code.

My point is that I don't want to force people to write inline comments matching a variable name when that isn't auto-generated (and then linted). It's not possible to automate your rules (at least not reasonably so?), so it's a non-starter for me. And then there's the fact that I'm not convinced there's a significant cost/effort ratio here, and so my motivation to invest time and energy to make the proposal something I could support is limited.

andreimatei:
> My point is that I don't want to force people to write inline comments matching a variable name when that isn't auto-generated

I'm honestly trying to understand this stance, so please humor me. If it's not the case that the proposal would force *you* to *read* comments that should be there, it also can't be that the proposal would force *you* to *write* comments that shouldn't be there. So then, the concern must be that the proposal is forcing some other poor schmuck to write comments that, while OK if written, were not, in your opinion, strictly necessary and thus they have spent time writing them that they might have spent better elsewhere. If that's homing in on the issue, can you give an example of a line of code with such a "not strictly necessary" comment that this proposal is mandating? Are *all* these name comments, in your opinion, "not strictly necessary"? If that's the case, is it because a) you know all the function signatures in the codebase by heart or b) because it's easy enough to always look up the signature of function you don't know when you're browsing through and reading code or c) something else?

dt:
I'm 👎 on requiring inline comments, and on _requiring_ pretty much any of these suggestions, or even, for that matter, presenting them as "SHOULD" suggestions. Readability is heavily dependent on context and part of what we consider when we write (and review) code. Sometimes a bool is fine and perfectly obvious what it means, sometimes an enum or at least named constants (with values of type bool even) are appropriate.

andreimatei:
> I'm 👎 on requiring inline comments, and on requiring pretty much any of these suggestions, or even, for that matter, presenting them as "SHOULD" suggestions.

OK, looks like I'll have to drop this as it didn't capture the hearts and minds, but it breaks my heart. @dt, can you please make a small effort to try to convince your open-minded colleague, as a courtesy to me? Saying generalities like "readability is heavily dependent on context" doesn't help me; by that argument, we'd have no readability rules, and yet we do. How is a `nil` in the middle of another 100 arguments not generally deserving a comment? Or, let me put it another way: would you not agree that there are lines of code with such `nil`s (say) that are always bad, regardless of their context? If you do agree with that, do you not believe that a rule can be written to capture them? Or, if you do believe that a rule could exist, would you look more favorably at the present proposal if it said that the inline comments are required only if no other comment around the call site exists that explains the literal in question (that part of your comment I think is valid and I like).

Separately, would anybody's mind be changed if the text said that the comments for literals are required by "default", but can be dropped, besides the many exceptions, whenever the writer makes a conscious decision and has a particular reason for doing so (other than laziness or unawareness of the default rule)? This was always the intent, but I didn't find good language for it.

For what it's worth, I didn't send this out lightly; I've sent it out after months of observing that we have a problem, of frustration that I have to make the same comment on reviews over and over (in cases where I did believe that readability was being hurt), and of agonizing in anticipation of this conversation.
The straw that broke my back was a conversation about competing styles of commenting `false` arguments: `name` vs `!name` vs `<negation of name>`. That is just dreadful.

dt:
I'm not sure how much I can add to what's been said above, but yes to answer one of your questions, I think this seems like its one of those aspects of readability that just appears to be hard to capture in a "rule", at least one more specific than "make it obvious to the reader what the parameters to a function mean / are doing".

andreimatei:
> I think this seems like its one of those aspects of readability that just appears to be hard to capture in a "rule"

Well, but you say this in the context of me claiming to present precisely such a rule. I claim to have done the "hard" work! So, what gives? I recognize that it's not the every mathematician's duty to find the problem with every nut's P!=NP proof, but still, have you actually considered this text? Would it produce any unwarranted burden? So far Jordan has been the only one who produced some counter-examples which, as he's admitted, were quite particular cases which can easily be exempted.

You've also generally stayed silent on whether you think we have any problem that deserves addressing in any way, or not. If you tell me "I haven't encountered a problem", than we can discuss that, or I can drop the whole thing. But so far nobody has referred to that, except Raphael who has agreed that there is a problem.

couchand:
I am so 👍👍  on this my thumbs hurt.  There is hardly a bigger impediment to reading unfamiliar code than mysterious and magical primitive arguments to functions defined far off (particularly booleans 😱).  There are two potential solutions to this: get rid of them completely (not practical) or (start to) explain them when they occur.  Leaving them mysterious and magical (a.k.a. the status quo) is not a solution.

Everybody keep in mind this is a change to the *style guide*, a document that should be assumed to read "do this always, unless it doesn't make sense to" (maybe it should say that explicitly, too).

None of the arguments presented here against the change really hold water.  Let me consider them each in turn:

- "I wouldn't do this without providing the necessary tooling to enforce" - @tamird 

As reasonably pointed out in numerous comments here, creating a completely hard and fast rule on this is not really possible, it is necessarily a judgement call.  That doesn't make it unimportant to do (as some have claimed... that's a crazy argument, we do lots of things that require human judgement, and that's a Good Thing), it just means that a human has to be in the loop, we can't automate it away.

- "it just feels like too much work for too little benefit" - @danhhz 

This view requires a limited perspective of the work and benefit.  If the original author of the code doesn't write this sort of annotation (or any other style that makes the intent clear), they are *imposing* work on future readers of their code (c.f. tragedy of the commons).  The original author isn't the primary intended beneficiary, at least not at the time of writing.

- "just asking for rot when the name of the parameter gets changed" - @danhhz 

If the meaning of the parameter changes, too, then the usage needs to change and so someone will be editing that line anyway and it won't rot.  If the meaning didn't change, the old comment is probably still okay and better than nothing, as @andreimatei already pointed out.

- "I hesitate to require this style when we have so many violations in our code base" - @petermattis 

That would be an argument against ever improving style guidelines in a way that isn't automatable, and clearly that's not a way to make progress.  Backfilling the annotations in code that isn't being touched isn't valuable, but anyone submitting a PR annotating a file they were researching should be applauded.

- "my editor annotates the parameter names in those cases" - @jordanlewis 

"not all editors and viewers do that." - @jordanlewis 

- "I'm not convinced of the readability improvements (having all these inline comments everywhere is going to make the code much harder to read IMO)" - @danhhz 

I suppose that depends on exactly what you mean by "readability".  I'm inclined to say the only useful definition would include actually understanding the meaning of the code.  If there are mysterious arguments without the meaning being obvious, the text may be easy enough to get through, but it's still not easy to understand.

- "You should use an editor that does this for you" - @arjunravinarayan 

If anything, this argument proves the value of this sort of annotation.  ICYMI, "not all editors and viewers do that." - @jordanlewis 

- "help the user by suggesting the comment additions where they are missing" - @knz 

This is actually really interesting, because it would be helpful, and is simple enough to actually build.  But this seems like a nice-to-have, not a blocker.

- "I can't be in favor of anything that turns contributors into human code generators" - @tschottdorf 

Anything?  How about a Go for loop? &lt;/snark&gt;

petermattis:
>> "I hesitate to require this style when we have so many violations in our code base" - @petermattis
>
>That would be an argument against ever improving style guidelines in a way that isn't automatable, and clearly that's not a way to make progress. Backfilling the annotations in code that isn't being touched isn't valuable, but anyone submitting a PR annotating a file they were researching should be applauded.

I did temper my statement by saying "require this style". I actually like this style of commenting literals myself and will encourage it in code reviews. It just feels wrong to mandate it without fixing up the existing code base and/or adding a lint for detecting violations. I don't think either of those hurdles is large given the existing lint and code checking infrastructure. It also seems reasonably clear to me that this proposal won't go forward without some automated checking. (Though it may not go forward even then).

RaduBerinde:
I agree this generally makes the code more readable. I think working on a tool for it is a bit of overkill. I would add it as a soft recommendation though (without all the details and exceptions, it should be a "use your best judgement" sort of thing). I think it's a good idea for reviewers to suggest this wherever it truly helps readability.

andreimatei:
Please give this updated proposal one last shot, and if the current form also fails, I'll finally shut up.
I've deleted everything and replaced the main contentious part with basically one sentence with extremely lax language. It's no longer a "rule" in any way, so now I'm basically hoping that this will be more broadly liked and then that people will follow it through their unencumbered free will.
The proposal no longer says anything about return statements and blank identifiers (i.e. `_, _ = foo()`).

The proposal now serves three purposes:
1.  Encourage inline comments for unclear arguments, particularly for basic literals, bool constants and nil identifiers.
2. Standardize a style for `false /* <param name> */` comments, when those comments exist.
3. Give guidance on sometimes avoiding `bool` arguments.

How do y'all feel about this?

P.S. @couchand, appreciate your support 😍

arjunravinarayan:
I am now :+1: on this. We should take the inputs of newer members of the codebase more seriously, and @couchand's support makes me rethink. I also buy the counterargument that my use of an editor plugin to achieve this means that I'm really a supporter of this feature all along. I also like @andreimatei's relaxation of the requirements, and new guidelines. I do not think the fact that the existing codebase doesn't adhere to this is that significant a constraint. Thank you @andreimatei.

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwLqDFtNo6xAdsmIf5e:b-43jn60)*

arjunravinarayan:
(By which I mean people who are newer to seeing our codebase experience the pain points more sharply and with fresh eyes, so their feedback is extra valuable in the context of updating the style guide)

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwLr642XiWmKamE0B5C:b8nnkmg)*

tschottdorf:
Thanks for revamping this, @andreimatei. I'm also :thumbsup: on this one as it captures your main point of creating an unambiguous rule for annotating plus suggests when to move away from simple annotations. (In contrast to the previous version which had, I find, an overwhelmingly precise (and yet ambiguous) set of rules that was complex even to parse).

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[STYLE.md, line 77 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwLz4D3iok9KvhXI365:-KwLz4D3iok9KvhXI366:b-t1iwlz) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L77)):*
> ```Markdown
> monitor := mon.MakeMonitor(
>   "in-mem temp storage",
>   mon.MemoryResource,
> ```

Arguably though, in this case an options struct sounds appropriate (but yes, kind of hard to make the case that it should be *mandatory*. I would suggest this as an example where one should consider it though).

---

*[STYLE.md, line 132 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwLzP2H5mKNpTHs_ju-:-KwLzP2IjdcQoKzCP2aH:b-7o379i) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L132)):*
> ```Markdown
> is better than
> ```
> func endTxn(commit bool) {}
> ```

In an ideal world, this would be `commitTxn()` and `abortTxn()` but then you're left to either have the bool-variant around still, or duplicating code, so probably not something we'd want to mention here as an alternative.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwLzwiETBUN_ABwY0_e:b8m6ewa)*

RaduBerinde:
LGTM

knz:
:lgtm:

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwQHc3ZHkYohwxpij52:bnfp4nl)*

tschottdorf:
BTW, I think https://mtlynch.io/human-code-reviews-1/#settle-style-arguments-with-a-style-guide will resonate strongly with you @andreimatei -- your initial approach was to basically specify the solution completely and I agree with that in spirit, except that in such a situation it would have to be enforced by a program and I didn't see that happening, and was worried that the resulting manual overhead would be too much. With your revised RFC, I hope we'll still avoid arguments because you can say "but look what the style guide says".

andreimatei:
That article does resonate with me. But the rest of what you're saying, although a bunch of people seemingly feel the same way, does not follow in my opinion.
I don't understand the "must be enforced by a program" part one bit. First of all, I don't think that what I want here _can_ be enforced by a program, as it was always supposed to require a degree of discernment between the specific, but very common, exceptions and the general case. More importantly, why is programatic enforcement a necessity? And btw, are we talking about a linter _enforcing_ the rule, or a linter _changing the code_ to conform to the rule? If it's the former, then the overhead you speak would remain (the latter has its own problems). Also, if we believe something helps readability, it helps readability whether or not a program enforces is. It also helps readability with every incremental application - even if some writers and some reviewers ignore it, as long as some behavior is changed some of the time, we've won something.

> With your revised RFC, I hope we'll still avoid arguments because you can say "but look what the style guide says".

I hope the same. But note that, technically, the style guide won't say anything after this change. If a reviewer refers to the style guide, and the author says "Neah, don't want to. It's readable enough. If I wanted to I would have done it in the first place so please stop bothering me.", they'd be within their rights. That's why I wanted a rule according to which some lines of code would clearly be _illegal_ - e.g. `foo(arg1, arg2, ... arg14,    nil,   arg15,...)`.

tschottdorf:
> I don't understand the "must be enforced by a program" part one bit. First of all, I don't think that what I want here can be enforced by a program

That's one of my argument. If a program can't, then we have a human do the job. Human time is pretty expensive, so I proposed reduction in scope to do less work (i.e. address main problem, be somewhat lenient in the grey area).

> More importantly, why is programmatic enforcement a necessity?

See above. Human time is expensive. We spend too much of it on nits already (including on occasions myself, because I'm too OCD to just let it slide).

> And btw, are we talking about a linter enforcing the rule, or a linter changing the code to conform to the rule? If it's the former, then the overhead you speak would remain (the latter has its own problems).

I'd be willing to settle for either, though the latter would be preferable. For the former, some overhead remains (wasting a CI run, having to fix something, etc) but all of the in-review overhead goes away. I don't have to point out that you used tabs here instead of spaces, you don't have to give me an ack when you fixed it, I don't have to check that it is now fixed.

> Also, if we believe something helps readability, it helps readability whether or not a program enforces is. It also helps readability with every incremental application - even if some writers and some reviewers ignore it, as long as some behavior is changed some of the time, we've won something.

Yes, and the current proposal has my blessing. I don't see the need to enumerate and judge all the possible infractions, something which you've admitted is impossible.

> But note that, technically, the style guide won't say anything after this change. If a reviewer refers to the style guide, and the author says "Neah, don't want to. It's readable enough. If I wanted to I would have done it in the first place so please stop bothering me.", they'd be within their rights.
That's why I wanted a rule according to which some lines of code would clearly be illegal - e.g. foo(arg1, arg2, ... arg14, nil, arg15,...).

A large amount of code we write is based judgment (not rules) and we all mix and match our style to what we like (which in turn is often drawn from what we see or get suggested). Having something in the style guide is a strong hint to not deviate without an equally strong reason. In the scenario you describe, I don't think the problem is with the rules, it's with the strong opinions involved on both sides, and, as an aside, for more reasonable examples often the sensible option is to *let it go*. Or, in the words of the blog post,

> If your style guide doesn’t specify a convention about a particular issue, it’s generally not worth arguing about.

That said, once we notice infractions that recur and we may want to add them, in which case, yeah, let's do it. But I'm not so sure it'll happen.

bdarnell:
:lgtm: (and :thumbsdown: on a linter for this)

---

Review status: all files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[STYLE.md, line 111 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwXQalsRE0zqxCPc07O:-KwXQalsRE0zqxCPc07P:bz5i9sy) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L111)):*
> ```Markdown
> ```
> 
> This is not always the case; e.g. `bool` arguments to constructors are fine.
> ```

I don't see why constructors are any different. (for example, `NewIterator(prefix bool)` would probably be better as `NewIterator()` and `NewPrefixIterator()`)

---

*[STYLE.md, line 115 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwXQNTEW6S5RoLM6B4r:-KwXQNTEW6S5RoLM6B4s:brbcy2d) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L115)):*
> ```Markdown
> whether the `doSomething` function should exist at all.  
> In cases where the `bool` in question acts as a "knob" to the function,
> consider replacing it with some type of "configuration" struct (for examples,
> ```

A configuration struct seems like a useful replacement for functions taking many arguments, but I don't see how it applies to individual bools. I would remove this sentence (unless you can explain what I'm missing). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwXRwWy9t0IMv_OTfkY:b-w4vk1l)*

dt:
I'm also more 👍  on this wording.

I might add that if you're adding a function where you expect callers may pass something like `nil` or `false` or `0` to indicate a particular behavior, providing names for those values in your package (const  can be a nice way to have call-sites document that behavior. The `enum-instead-of-bool` is essentially a special case of this general pattern, except with limited applicability and compiler enforcement of the callers using it.

An example of this in action: @benesch included nil values named `jobs.WithoutCancel` and a `jobs.Noop` that can be passed as the the `cancelFn` or `progressFn` to various `jobs` functions, and guides callers to use them rather than literals: https://github.com/cockroachdb/cockroach/pull/17877#issuecomment-324472915 

As an added bonus, unlike inline comments, these names won't rot when the function args are renamed.

nvanbenschoten:
:lgtm:

---

Review status: all files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[STYLE.md, line 85 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-Kw_m9LHWgdde6vOAQyr:-Kw_m9LHWgdde6vOAQys:bs6q81b) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L85)):*
> ```Markdown
> ```
> 
> Note: For `bool` constants, like for all literals, the comment should indicate
> ```

:100: on this rule. Are you going to go back and fix all existing violations after this? The `/* !commit */` style violations should be easy to find. The other type will be harder.

---

*[STYLE.md, line 115 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwXQNTEW6S5RoLM6B4r:-Kw_mbgKuF5tPeoGa7au:b32v57) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L115)):*

A configuration struct seems like a useful replacement for functions taking many arguments, but I don't see how it applies to individual bools. I would remove this sentence (unless you can explain what I'm missing). 
</blockquote></details>

+1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-Kw_mtzO4gTBrAlYwj9_:bnfp4nl)*

danhhz:
@andreimatei thank you for your perseverance through such a long (and violent) review and for being so motivated about the hygiene of our codebase. I know it's not as black and white as you'd have liked, but I think this softened version is a net win for codebase style. LGTM

benesch:
P.S. You may want to rebase this to pick up a fix for the flaky `TestSystemZoneConfigs`.

andreimatei:
@dt I've added a paragraph on named sentinels; please see if you like it.

---

Review status: all files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[STYLE.md, line 83 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEsLQIpNUx2TzTbK0E:-KwaF3cT-vL3FL-ngu7H:bmesv1r) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L83)):*

Please add a forward reference here to the paragraph below where you explain that an additional explanation is also acceptable.
</blockquote></details>

the paragraph below is gone, but I added an example with an explanation here

---

*[STYLE.md, line 84 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEt0VAOyw2RwrSH8P6:-KwaFC9eHJ1kCy4P_8CC:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L84)):*

Space missing between paragraph and example.
</blockquote></details>

Done.

---

*[STYLE.md, line 162 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtQrZuxQwYBZ89LbV:-KwaFIoSw-a6C_gO6izV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L162)):*

ditto
</blockquote></details>

Done.

---

*[STYLE.md, line 175 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtaHWKVnlm0c4GhC4:-KwaG0SWrg6jg72-IUvB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L175)):*

Before you introduce this rule please look at the generated assembly for a couple examples. I have a hunch that Go does not yet optimize two-valued enums as well as booleans. It would be a shame if this recommendation would make our code slower overall.
</blockquote></details>

Done.

---

*[STYLE.md, line 177 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEt_YmZFVy12fTtgsx:-KwaG12_yiFxKE-LT5Z7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L177)):*

Space missing between paragraph and example.
</blockquote></details>

Done.

---

*[STYLE.md, line 188 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwEtZRJ4egKDVJQV6RU:-KwaG1VLZzFPmLDEM5VB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/863a48bd89e9ec134fc39d3c5d3f4f29e56a88cc/STYLE.md#L188)):*

ditto
</blockquote></details>

Done.

---

*[STYLE.md, line 77 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwLz4D3iok9KvhXI365:-KwaE0gRKhu8Pv_K0SeY:b-1407w8) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L77)):*

Arguably though, in this case an options struct sounds appropriate (but yes, kind of hard to make the case that it should be *mandatory*. I would suggest this as an example where one should consider it though).
</blockquote></details>

I agree, but I don't know what to do about it exactly. I don't particularly want to expand the section on options structs at this time...

---

*[STYLE.md, line 85 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-Kw_m9LHWgdde6vOAQyr:-KwaED1KVeW0vrcENiTP:bk5qqea) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L85)):*

:100: on this rule. Are you going to go back and fix all existing violations after this? The `/* !commit */` style violations should be easy to find. The other type will be harder.
</blockquote></details>

yes, I'm going to fix all `/* !commit */`.

---

*[STYLE.md, line 111 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwXQalsRE0zqxCPc07O:-KwaFWI7mkiEV86nrD_E:b6s3c04) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L111)):*

I don't see why constructors are any different. (for example, `NewIterator(prefix bool)` would probably be better as `NewIterator()` and `NewPrefixIterator()`)
</blockquote></details>

removed the mention to ctors

---

*[STYLE.md, line 115 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwXQNTEW6S5RoLM6B4r:-KwaFxJIkrEEO91z0-Nu:bypywow) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L115)):*

+1
</blockquote></details>

rephrased. Please see if you like it.

---

*[STYLE.md, line 132 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwLzP2H5mKNpTHs_ju-:-KwaG5B9juQY2BL7VN9b:bve2kb) ([raw file](https://github.com/cockroachdb/cockroach/blob/024b784337bce0992a72695f73a62017d3b770d2/STYLE.md#L132)):*

In an ideal world, this would be `commitTxn()` and `abortTxn()` but then you're left to either have the bool-variant around still, or duplicating code, so probably not something we'd want to mention here as an alternative.
</blockquote></details>

yes. This paragraph only says that one is better than the other; doesn't make an absolute judgement :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwaLn2KypwM2HNW7fTx:b7t991y)*

dt:
:LGTM: and :plause: for the dedication and patience to see this one though!

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 23 unresolved discussions, some commit checks pending.

---

*[STYLE.md, line 142 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwaN3_g1aXG4CI-KFQX:-KwaN3_g1aXG4CI-KFQY:bped0nv) ([raw file](https://github.com/cockroachdb/cockroach/blob/2dedd9107b56f250fd69d46d09601daad06f5be0/STYLE.md#L142)):*
> ```Markdown
> ```
> 
> #### Try to give names to sentinel values
> ```

I generally dislike creating named aliases for nil or zero values (except for `false`). Using a literal nil or zero-valued struct literal is quite common in our codebase and it's rarely confusing; if there is a named constant I have to wonder whether that value is distinct from nil or not. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185)*

andreimatei:
Review status: 0 of 1 files reviewed at latest revision, 23 unresolved discussions, some commit checks pending.

---

*[STYLE.md, line 142 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwaN3_g1aXG4CI-KFQX:-KwaOJhT5BEqvCEGvEYT:b-ad9lgl) ([raw file](https://github.com/cockroachdb/cockroach/blob/2dedd9107b56f250fd69d46d09601daad06f5be0/STYLE.md#L142)):*

I generally dislike creating named aliases for nil or zero values (except for `false`). Using a literal nil or zero-valued struct literal is quite common in our codebase and it's rarely confusing; if there is a named constant I have to wonder whether that value is distinct from nil or not. 
</blockquote></details>

yes, I'm also ambivalent on this for `nil`. If I were to defend this, for `nil`, I'd say that you shouldn't have to wonder if the constant is different from `nil`, as the function should not tell you that `nil` is permitted (so, in that respect, I don't like the comment on this `Job.Progressed()` which makes a big deal out of the `nil` value.

For ints - `0` or `-1`, it has bothered me in the past that we use both of them as special values at times. At least giving a name to it would prevent me from seeing the value, so I'm more for it.
For structs, I don't know what to say.

@dt ? Others?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185)*

dt:
I'm 👍 on this with or without the wording on sentinels, so if removing it makes it less controversial, I wouldn't let it prolong this CL.

As far as the merits of the named sentinels, I generally do not assume that it is safe to just pass `nil`s to functions unless otherwise stated, so I would just use provided sentinel instead. Like @andreimatei, I'd hope that whether it is different than `nil` wouldn't be an issue since no one would be using / considering `nil` anyway.

Again though, if this is controversial, I'm more than fine just nuking this section in the interest of getting the broader recommendation merged.

andreimatei:
Thanks all. I'm quite happy with the outcome; nice doing business with you.

---

Review status: 0 of 1 files reviewed at latest revision, 23 unresolved discussions.

---

*[STYLE.md, line 142 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-KwaN3_g1aXG4CI-KFQX:-Kwlfl3rq97BdXVlD-6S:bnevc65) ([raw file](https://github.com/cockroachdb/cockroach/blob/2dedd9107b56f250fd69d46d09601daad06f5be0/STYLE.md#L142)):*

yes, I'm also ambivalent on this for `nil`. If I were to defend this, for `nil`, I'd say that you shouldn't have to wonder if the constant is different from `nil`, as the function should not tell you that `nil` is permitted (so, in that respect, I don't like the comment on this `Job.Progressed()` which makes a big deal out of the `nil` value.

For ints - `0` or `-1`, it has bothered me in the past that we use both of them as special values at times. At least giving a name to it would prevent me from seeing the value, so I'm more for it.
For structs, I don't know what to say.

@dt ? Others?
</blockquote></details>

I've removed this because a) it looks like it's not universally liked and b) I don't think this small section I had written gave justice to the topic; it wasn't very good.
Let's skin this cat (?) another time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19185#-:-KwlgtWXZgmLHL_R6H2f:b-y9253p)*

PR #19186, opened 2017-10-11
ui: reducer and redux state file cleanup.

couchand:
This is a few little renames/export changes to the reducers to try to make everything nice and consistent.  I'm happy to bikeshed the details, getting them consistent and not using a default export are the goals.

I'm not a big fan of `apiReducersReducer`, but left it as-is for consistency.  I'm sure there's a better name for that one (maybe `cachedDataReducer` to match the state key?).

### DIFF
--- a/pkg/ui/src/redux/alerts.ts
+++ b/pkg/ui/src/redux/alerts.ts
@@ -11,7 +11,7 @@ import { ThunkAction } from "redux-thunk";
 
 import { LocalSetting } from "./localsettings";
 import {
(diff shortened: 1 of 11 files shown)


mrtracy:
LGTM. I'm dealing with what is now "apiReducerReducers" as part of network refactoring.

---

Review status: 0 of 12 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19186#-:-KwC1jUVO93iIWJeqnBm:b-h0vqzb)*

PR #19191, opened 2017-10-11
config: rip out zone config migrations

benesch:
@bdarnell (and maybe @tschottdorf?) is this reasonable to do?

-----

Expressive zone configs (i.e., our current zone config architecture)
have been in place since beta-20161013. I think it's safe to assume that
all ancient zone configs have been migrated to the new format, so we can
rip out the migration code paths. The migration was never very good,
anyway; the old format supported per-replica attributes but the new
format only supported per-zone constraints, so attributes for all but
the first replica were dropped on the floor during the migration.


bdarnell:
:lgtm:

Since this is pre-1.0 and the failure mode is reasonable (a loosening of constraints on zone configs) I think it's fine to rip this out even if there is some chance that a cluster could have made it to today with old-style configs (and that seems very unlikely although I'm not sure offhand how exactly getZoneConfig is used). 

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19191#-:-KwXxUcpaafbncYY9bAK:b-ffi8yv)*

a-robinson (approved):
I'm 👍 on this. cc @BramGruneir

BramGruneir:
👍 On this!  Thanks for doing it.

benesch:
Woohoo, thanks for the reviews!

PR #19192, opened 2017-10-11
config: fix comment on ZoneConfig proto

benesch:
`ZoneConfigHuman` is not and never was. (It was excised from the PR that
added the comment during code review.) Also take the opportunity to
clarify usage of the word "range."


a-robinson (approved):
Heh, thanks. I might contend that it "applies to one or more ranges" rather than "a span of keys", but that's not necessarily worth the extra teamcity time.

benesch:
Good point! Done.

PR #19194, opened 2017-10-11
storage: optimization to Store.Send and Replica.Send pathways

spencerkimball:
Store:
- Skip the post-execution `Clock.Update()` call if the timestamp was not forwarded
  from the original `BatchRequest.Timestamp`.
- Utilize the built-in check in `Clock.Update()` for a timestamp
  exceeding max offset from physical now by adding new `Clock.UpdateAndCheckMaxOffset()`.
- `defer func()` in `Store.Send` was expensive, so changed the structure by using
  a helper function instead.
- Provide a fast path in Transaction.UpdateObservedTimestamp().
- Remove expensive log statement.
- Remove `Store.verifyKeys` and augment `keys.Range` to verify the same items. `keys.Range` is already called in multiple locations in `Replica`.

Replica:
- Re-order the `IsReadOnly` and `IsWrite` calls.
- Invoke `checkBatchRequest` with `isReadOnly` variable to avoid locking mutex.
- Move cast of `pErr.GetDetail()` to `RaftGroupDeletedError` into `pErr != nil` block.

Added an `RSpan` to the `BatchRequest.Header` to avoid recomputing `keys.Range`
multiple times in the process of executing a batch through `Store.Send` and
`Replica.Send`.


petermattis:
So many separate changes bundled together here, that I'm frankly scared about reviewing and signing off on this as I'm sure something will slip by. Is it possible to break this into separate more digestible PRs?

---

Review status: 0 of 24 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/storage/replica.go, line 1905 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwKwpRCAan-tpoIwSOh:-KwKwpRCAan-tpoIwSOi:b-ja1v5x) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/storage/replica.go#L1905)):*
> ```Go
> 	// are excluded.
> 	if pErr == nil && retry == proposalNoRetry && ec.ba.ReadConsistency != roachpb.INCONSISTENT {
> 		span, err := keys.Range(ec.ba, false)
> ```

@andreimatei Do you have anything to say here?

---

*[pkg/storage/store.go, line 2385 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwKvjZ6yjp8Yzr_GK5g:-KwKvjZ6yjp8Yzr_GK5h:b9xry71) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/storage/store.go#L2385)):*
> ```Go
> 			return nil, roachpb.NewError(err)
> 		}
> 		ba.RSpan = rs
> ```

So you're setting `Header.RSpan` on the server side to avoid recalculating it later? If yes, I don't think we should be adding a field to `Header`. Doing so expands the encoded size of `Header` even when `RSpan` isn't set. Instead, I think we should be create some sort of `RequestState` struct that we pass down through the stack (instead of passing `BatchRequest`). I realize this is more work, but I think we should avoid adding fields to protos that are never sent over the network. 

Also, this seems good to pull out into a separate PR.

Update: looking further in the PR I realize just how onerous this request is. Yet I maintain we should figure out a cleaner way to do it rather than adding `Header.RSpan`. Perhaps file an issue. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-:-KwKuzP_K7MzPgPafgGA:b1j0uo1)*

RaduBerinde:
*[pkg/storage/replica.go, line 1905 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwKwpRCAan-tpoIwSOh:-KwLJpp49aBGl0ILBHFI:bmi8ppe) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/storage/replica.go#L1905)):*

@andreimatei Do you have anything to say here?
</blockquote></details>

:100: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194)*

andreimatei:
Review status: 0 of 24 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/keys/keys.go, line 677 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwLKvATxHI4aV-RAFox:-KwLKvATxHI4aV-RAFoy:b-yjce15) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/keys/keys.go#L677)):*
> ```Go
> // Range returns a key range encompassing all the keys in the Batch.
> func Range(
> 	ba roachpb.BatchRequest, enforceLocalGlobalMix bool,
> ```

please document the new `enforceLocalGlobalMix`. Also, since this is a very common interface, I strongly suggest you replace the `bool` with an enum so that it has proper names. 

But actually, it strikes me that a param specifying whether some assertion should be checked or not is a weird thing (again, particularly for such a prominent interface). Without claiming to understand too much about it, I suspect this argument shouldn't exist. I think it'd be better if you made `Range()` do the assertion always, and add a new `RawRange()` or smth that doesn't (if you ever need to not do it). They both can call a private `rangeImpl()` that takes this knob.

More discussion on all these suggestions in https://github.com/cockroachdb/cockroach/pull/19185

---

*[pkg/storage/replica.go, line 1905 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwKwpRCAan-tpoIwSOh:-KwLKAZrGIgkYwvzw9JA:b-84488q) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/storage/replica.go#L1905)):*

:100: 
</blockquote></details>

Peter is :trollface:ing

If the arg stays (although I hope it won't, see comment on the function), please comment the surprising arg inline everywhere: `/* enforceGlobalLocalMix */`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194)*

tschottdorf:
Yeah, I'd also greatly prefer to see these changes submitted individually, also to see if any of them move the benchmarks.

---

Review status: 0 of 24 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-:-KwLfxwSBDpqfLMU3Blb:bbim727)*

andreimatei:
@spencerkimball , for all the log statements you've eliminated or made conditional, please consider using the new `ExpensiveLogEnabled()` method. It should probably always replace `if log.V(2) {...}`
https://github.com/cockroachdb/cockroach/blob/751a182b64008fbd75ded7441fe3a003e5aa2a8e/pkg/util/log/log.go#L183

---

Review status: 0 of 24 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-:-KwMQxRR2zabD6-9p-L6:b-smwla0)*

spencerkimball:
Yes I've split things up in separate PRs. I'm going to abandon this jumbo PR.

---

Review status: 0 of 24 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/storage/store.go, line 2385 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwKvjZ6yjp8Yzr_GK5g:-KxAMzXaUvEv8L-gObsF:b-woernf) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/storage/store.go#L2385)):*

So you're setting `Header.RSpan` on the server side to avoid recalculating it later? If yes, I don't think we should be adding a field to `Header`. Doing so expands the encoded size of `Header` even when `RSpan` isn't set. Instead, I think we should be create some sort of `RequestState` struct that we pass down through the stack (instead of passing `BatchRequest`). I realize this is more work, but I think we should avoid adding fields to protos that are never sent over the network. 

Also, this seems good to pull out into a separate PR.

Update: looking further in the PR I realize just how onerous this request is. Yet I maintain we should figure out a cleaner way to do it rather than adding `Header.RSpan`. Perhaps file an issue. 
</blockquote></details>

Actually the idea was to specifically make use of the fact that this is computed in the `DistSender`, and to allow that work to propagate to the Store via the network. So I think it is appropriate to make this part of the `BatchRequest.Header`. I'm going to send a new PR incorporating feedback here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-:-KxAMmbC59aEBv7fW2LX:b6r16c6)*

petermattis:
Review status: 0 of 24 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/storage/store.go, line 2385 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194#-KwKvjZ6yjp8Yzr_GK5g:-KxAdkKmxRMZy6UUk2Jb:bx3ffdi) ([raw file](https://github.com/cockroachdb/cockroach/blob/489240d18a57a0bafcaa36ae075b7af8f7774a9b/pkg/storage/store.go#L2385)):*

Actually the idea was to specifically make use of the fact that this is computed in the `DistSender`, and to allow that work to propagate to the Store via the network. So I think it is appropriate to make this part of the `BatchRequest.Header`. I'm going to send a new PR incorporating feedback here.
</blockquote></details>

Ok. Keep in mind how this will work in mixed-version clusters. Also, copying the key in the headers again increases the number of bytes sent over the network.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19194)*

PR #19200, opened 2017-10-12
kv: implement dist sender optimizations

spencerkimball:
- Fast path at start of `divideAndSendBatchToRanges` for case where
  all the key span is contained within a single range.
- Avoid batch truncation in fast path in `sendPartialBatch`.
- Fast path in `grpcTransport` for case of a single replica - send
  syncrhonously instead of via goroutine.
- Remove `DistSender.sendRPC`'s `context.WithCancel` as cancellation should
  properly be done in the transport, which is the component which understands
  if cancellation is necessary. gRPCTransport already has a cancellation
  mechanism, so we were doing this twice unnecessarily. In the case of the
  dist sender fast path (added to this commit) which prevents launching a
  goroutine in the event of a single replica, this means the transport can
  avoid starting any cancelable context.
- Remove unnecessary transaction clone from grpcTransport for local
  server case because after code audit, determined that there are no
  cases where `ba.Txn` is modified at `Store.Send` and below without
  first doing a similar shallow copy of `ba.Txn`.

I removed the shallow clone of the batch's transaction.  This is no
longer needed, based on a code audit. There is nowhere in the Store ->
Replica -> Raft evaluation path where we modify the batch's Txn
without first making a shallow copy of it.

### DIFF
--- a/pkg/kv/dist_sender.go
+++ b/pkg/kv/dist_sender.go
@@ -377,11 +377,6 @@ func (ds *DistSender) sendRPC(
 	// RangeNotFoundErrors.
 	ba.RangeID = rangeID
 
-	// A given RPC may generate retries to multiple replicas, but as soon as we
-	// get a response from one we want to cancel those other RPCs.
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	tracing.AnnotateTrace()
 	defer tracing.AnnotateTrace()
 
@@ -652,6 +647,28 @@ type response struct {
 func (ds *DistSender) divideAndSendBatchToRanges(
 	ctx context.Context, ba roachpb.BatchRequest, rs roachpb.RSpan, batchIdx int,
 ) (br *roachpb.BatchResponse, pErr *roachpb.Error) {
+	// Get initial seek key depending on direction of iteration.
+	var scanDir ScanDirection
+	var seekKey roachpb.RKey
+	if !ba.IsReverse() {
+		scanDir = Ascending
+		seekKey = rs.Key
+	} else {
+		scanDir = Descending
+		seekKey = rs.EndKey
+	}
+	ri := NewRangeIterator(ds)
+	ri.Seek(ctx, seekKey, scanDir)
+	if !ri.Valid() {
+		return nil, ri.Error()
+	}
+	// Take the fast path if this batch fits within a single range.
+	if !ri.NeedAnother(rs) {
+		ba.SetNewRequest()
+		resp := ds.sendPartialBatch(ctx, ba, rs, ri.Desc(), ri.Token(), batchIdx, false /* needsTruncate */)
+		return resp.reply, resp.pErr
+	}
+
 	// Make an empty slice of responses which will be populated with responses
 	// as they come in via Combine().
 	br = &roachpb.BatchResponse{
@@ -661,7 +678,6 @@ func (ds *DistSender) divideAndSendBatchToRanges(
 	// implicated in the span (rs) and combines them into a single
 	// BatchResponse when finished.
 	var responseChs []chan response
-	var seekKey roachpb.RKey
 	var couldHaveSkippedResponses bool
 	defer func() {
 		if r := recover(); r != nil {
@@ -702,18 +718,7 @@ func (ds *DistSender) divideAndSendBatchToRanges(
 		}
 	}()
 
-	// Get initial seek key depending on direction of iteration.
-	var scanDir ScanDirection
-	if !ba.IsReverse() {
-		scanDir = Ascending
-		seekKey = rs.Key
-	} else {
-		scanDir = Descending
-		seekKey = rs.EndKey
-	}
-	// Send the request to one range per iteration.
-	ri := NewRangeIterator(ds)
-	for ri.Seek(ctx, seekKey, scanDir); ri.Valid(); ri.Seek(ctx, seekKey, scanDir) {
+	for ; ri.Valid(); ri.Seek(ctx, seekKey, scanDir) {
 		// Increase the sequence counter only once before sending RPCs to
 		// the ranges involved in this chunk of the batch (as opposed to
 		// for each RPC individually). On RPC errors, there's no guarantee
@@ -801,7 +806,7 @@ func (ds *DistSender) divideAndSendBatchToRanges(
 		} else {
 			// Send synchronously if there is no parallel capacity left, there's a
 			// max results limit, or this is the final request in the span.
-			resp := ds.sendPartialBatch(ctx, ba, rs, ri.Desc(), ri.Token(), batchIdx)
+			resp := ds.sendPartialBatch(ctx, ba, rs, ri.Desc(), ri.Token(), batchIdx, true /* needsTruncate */)
 			responseCh <- resp
 			if resp.pErr != nil {
 				return
@@ -868,7 +873,7 @@ func (ds *DistSender) sendPartialBatchAsync(
 		ds.asyncSenderSem, false, /* !wait */
 		func(ctx context.Context) {
 			atomic.AddInt32(&ds.asyncSenderCount, 1)
-			responseCh <- ds.sendPartialBatch(ctx, ba, rs, desc, evictToken, batchIdx)
+			responseCh <- ds.sendPartialBatch(ctx, ba, rs, desc, evictToken, batchIdx, true /* needsTruncate */)
 		},
 	); err != nil {
 		return false
@@ -884,7 +889,9 @@ func (ds *DistSender) sendPartialBatchAsync(
 // replicas, we backoff and retry by refetching the range
 // descriptor. If the underlying range seems to have split, we
 // recursively invoke divideAndSendBatchToRanges to re-enumerate the
-// ranges in the span and resend to each.
+// ranges in the span and resend to each. If needsTruncate is true,
+// the supplied batch and span must be truncated to the supplied range
+// descriptor.
 func (ds *DistSender) sendPartialBatch(
 	ctx context.Context,
 	ba roachpb.BatchRequest,
@@ -892,6 +899,7 @@ func (ds *DistSender) sendPartialBatch(
 	desc *roachpb.RangeDescriptor,
 	evictToken *EvictionToken,
 	batchIdx int,
+	needsTruncate bool,
 ) response {
 	if batchIdx == 1 {
 		ds.metrics.PartialBatchCount.Inc(2) // account for first batch
@@ -900,23 +908,27 @@ func (ds *DistSender) sendPartialBatch(
 	}
 	var reply *roachpb.BatchResponse
 	var pErr *roachpb.Error
+	var err error
+	var positions []int
 
 	isReverse := ba.IsReverse()
 
-	// Truncate the request to range descriptor.
-	intersected, err := rs.Intersect(desc)
-	if err != nil {
-		return response{pErr: roachpb.NewError(err)}
-	}
-	truncBA, positions, err := truncate(ba, intersected)
-	if len(positions) == 0 && err == nil {
-		// This shouldn't happen in the wild, but some tests exercise it.
-		return response{
-			pErr: roachpb.NewErrorf("truncation resulted in empty batch on %s: %s", intersected, ba),
+	if needsTruncate {
+		// Truncate the request to range descriptor.
+		rs, err = rs.Intersect(desc)
+		if err != nil {
+			return response{pErr: roachpb.NewError(err)}
 		}
-	}
-	if err != nil {
-		return response{pErr: roachpb.NewError(err)}
+		ba, positions, err = truncate(ba, rs)
+		if len(positions) == 0 && err == nil {
+			// This shouldn't happen in the wild, but some tests exercise it.
+			return response{
+				pErr: roachpb.NewErrorf("truncation resulted in empty batch on %s: %s", rs, ba),
+			}
+		}
+		if err != nil {
+			return response{pErr: roachpb.NewError(err)}
+		}
 	}
 
 	// Start a retry loop for sending the batch to the range.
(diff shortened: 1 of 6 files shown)


pkg/kv/transport.go
```diff
@@ -199,56 +207,47 @@ func (gt *grpcTransport) SendNext(ctx context.Context, done chan<- BatchCall) {
...
+	log.VEventf(ctx, 2, "sending request to %s", client.remoteAddr)
+	reply, err := client.client.Batch(ctx, &client.args)
+	if reply != nil {
+		for i := range reply.Responses {
+			if err := reply.Responses[i].GetInner().Verify(client.args.Requests[i].GetInner()); err != nil {
+				log.Error(ctx, err)
+			}
+		}
+	}
+	return BatchCall{Reply: reply, Err: err}
```
tamird: why not preserve the multiple return here? you lose tooling support for error checking without much of a benefit.

pkg/sql/ambiguous_commit_test.go
```diff
@@ -92,7 +92,9 @@ func TestAmbiguousCommit(t *testing.T) {
 						sendNext: func(ctx context.Context, done chan<- kv.BatchCall) {
 							if ambiguousSuccess {
 								interceptDone := make(chan kv.BatchCall)
-								transport.SendNext(ctx, interceptDone)
+								go func() {
```
tamird: no need for the extra function here, this is equivalent to `go transport.SendNext(ctx, interceptDone)`

pkg/kv/dist_sender.go
```diff
@@ -900,23 +906,32 @@ func (ds *DistSender) sendPartialBatch(
...
-		return response{pErr: roachpb.NewError(err)}
-	}
-	truncBA, positions, err := truncate(ba, intersected)
-	if len(positions) == 0 && err == nil {
-		// This shouldn't happen in the wild, but some tests exercise it.
-		return response{
-			pErr: roachpb.NewErrorf("truncation resulted in empty batch on %s: %s", intersected, ba),
+	if needsTruncate {
+		// Truncate the request to range descriptor.
+		intersected, err = rs.Intersect(desc)
```
tamird: can this overwrite `rs` instead of assigning to `intersected`? seems like it'd reduce the possibility of doing the wrong thing. `ba` and `truncBA` seem to demonstrate the same oddity.

tamird (commented):
Drive-by; feel free to ignore.

nvanbenschoten:
Reviewed 6 of 6 files at r1.

---

*[pkg/kv/dist_sender.go, line 900 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwI5_d8RJ4ilOMnMGYv:-KwI5_d8RJ4ilOMnMGYw:b86qfq4) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/dist_sender.go#L900)):*
> ```Go
> 	evictToken *EvictionToken,
> 	batchIdx int,
> 	needsTruncate bool,
> ```

Could you add a comment describing what `needsTruncate` means?

---

*[pkg/kv/dist_sender.go, line 918 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwDBmw1v8Os9G3prIju-r1-918:-KwI67m1h_I23NXFybSf:bthczj4) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/dist_sender.go#L918)):*

can this overwrite `rs` instead of assigning to `intersected`? seems like it'd reduce the possibility of doing the wrong thing. `ba` and `truncBA` seem to demonstrate the same oddity.
</blockquote></details>

Yeah, I think getting rid of `intersected` and `truncBA` would be cleaner. Just reassign to the previous values `if needsTruncate`.

---

*[pkg/kv/transport.go, line 192 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwI73K5dMm6BkgbT4OL:-KwI73K5dMm6BkgbT4OM:b194gmc) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/transport.go#L192)):*
> ```Go
> 	// on context or launch in a goroutine.
> 	if len(gt.orderedClients) == 1 {
> 		done <- gt.send(ctx, client)
> ```

We no longer have the `NotLeaseHolderErrors` check here. Is it not possible for a single replica to temporarily lose an expiration-based lease if it gets slowed down for long enough? In that case, wouldn't we want to retry?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

petermattis:

---

*[pkg/kv/transport.go, line 217 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwKyLfkWE5hizslqcUm:-KwKyLfkWE5hizslqcUn:b-oyx877) ([raw file](https://github.com/cockroachdb/cockroach/blob/359532ca24405996b21d0a8ab6e6c5549dde60b6/pkg/kv/transport.go#L217)):*
> ```Go
> 					clonedTxn := origTxn.Clone()
> 					client.args.Txn = &clonedTxn
> 				}
> ```

This might no longer be necessary, but we should pay attention to test race failures to make sure. 

---

*[pkg/kv/transport.go, line 220 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwKxjoRBxPi-9BCH9NZ:-KwKxjoRBxPi-9BCH9N_:b-m8o8er) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/transport.go#L220)):*
> ```Go
> 			}
> 		}
> 		gt.setState(client.args.Replica, false /* pending */, retryable)
> ```

Curious why this code wasn't moved to `grpcTransport.send`. Doing so would address @nvanbenschoten's comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

bdarnell:
This generally looks good, but optimization PRs should nearly always be accompanied by benchmark numbers. And separate commits would also be helpful (both for reviewing and for future bisecting).

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-:-KwY39taWvWa2HebmISy:b-bj5tpb)*

spencerkimball:
Performance on single node 99% read-percent KV load generator improved with this PR by ~10%.

BEFORE
```
   60.0s        0         568667         9477.5      3.4      2.8     10.0     18.9     92.3

BenchmarkLoadgenKV/readPercent=99/splits=0/concurrency=32/duration=1m0s	  568667	    105513 ns/op

   60.0s        0         563939         9398.6      3.4      2.8     10.0     19.9    184.5

BenchmarkLoadgenKV/readPercent=99/splits=0/concurrency=32/duration=1m0s	  563939	    106399 ns/op

   60.0s        0         573619         9560.0      3.3      2.8     10.0     18.9    121.6

BenchmarkLoadgenKV/readPercent=99/splits=0/concurrency=32/duration=1m0s	  573620	    104602 ns/op
```

AFTER
```
   60.0s        0         634057        10567.3      3.0      2.5      8.9     16.8    113.2

BenchmarkLoadgenKV/readPercent=99/splits=0/concurrency=32/duration=1m0s	  634057	     94631 ns/op

   60.0s        0         629878        10497.7      3.0      2.5      8.9     17.8    159.4

BenchmarkLoadgenKV/readPercent=99/splits=0/concurrency=32/duration=1m0s	  629879	     95258 ns/op

   60.0s        0         616276        10270.6      3.1      2.6      8.9     17.8    113.2

BenchmarkLoadgenKV/readPercent=99/splits=0/concurrency=32/duration=1m0s	  616276	     97365 ns/op
```

spencerkimball:
Review status: 2 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/kv/dist_sender.go, line 900 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwI5_d8RJ4ilOMnMGYv:-Kx8cPHKktz9HRKj-JbP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/dist_sender.go#L900)):*

Could you add a comment describing what `needsTruncate` means?
</blockquote></details>

Done.

---

*[pkg/kv/dist_sender.go, line 918 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwDBmw1v8Os9G3prIju-r1-918:-Kx8dBmCtz8LxerWK58E:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/dist_sender.go#L918)):*

Yeah, I think getting rid of `intersected` and `truncBA` would be cleaner. Just reassign to the previous values `if needsTruncate`.
</blockquote></details>

Done.

---

*[pkg/kv/transport.go, line 192 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwI73K5dMm6BkgbT4OL:-Kx92e1RY5vycqVn2USr:b-x61wkg) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/transport.go#L192)):*

We no longer have the `NotLeaseHolderErrors` check here. Is it not possible for a single replica to temporarily lose an expiration-based lease if it gets slowed down for long enough? In that case, wouldn't we want to retry?
</blockquote></details>

OK, I moved the `NotLeaseHolderError` check into `grpcTransport.send`

---

*[pkg/kv/transport.go, line 217 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwKyLfkWE5hizslqcUm:-Kx93sMLmCfgIumz0MZn:b32w0x) ([raw file](https://github.com/cockroachdb/cockroach/blob/359532ca24405996b21d0a8ab6e6c5549dde60b6/pkg/kv/transport.go#L217)):*

This might no longer be necessary, but we should pay attention to test race failures to make sure. 
</blockquote></details>

OK

---

*[pkg/kv/transport.go, line 220 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwKxjoRBxPi-9BCH9NZ:-Kx92jliNocWBzznT7bK:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/transport.go#L220)):*

Curious why this code wasn't moved to `grpcTransport.send`. Doing so would address @nvanbenschoten's comment above.
</blockquote></details>

Done.

---

*[pkg/kv/transport.go, line 248 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwDBmw1v8Os9G3prIjx-r1-248:-Kx93Yq4PAndEkELUIoW:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/kv/transport.go#L248)):*

why not preserve the multiple return here? you lose tooling support for error checking without much of a benefit.
</blockquote></details>

Done.

---

*[pkg/sql/ambiguous_commit_test.go, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-KwDBmw1v8Os9G3prIjz-r1-95:-Kx93cBOkewn8WvQoYu3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e797510985ac138e4972a51a0e67e00d6cf23e7d/pkg/sql/ambiguous_commit_test.go#L95)):*

no need for the extra function here, this is equivalent to `go transport.SendNext(ctx, interceptDone)`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

tamird:
Reviewed 4 of 4 files at r2.

---

*[pkg/kv/transport.go, line 203 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-Kx9S8MqdlPReQ00hUtT:-Kx9S8MqdlPReQ00hUtU:b8q2k1o) ([raw file](https://github.com/cockroachdb/cockroach/blob/70eeefde44af0b5597a19f0e3343122daaa38594/pkg/kv/transport.go#L203)):*
> ```Go
> 		defer gt.closeWG.Done()
> 		gt.opts.metrics.SentCount.Inc(1)
> 		reply, err := func() (*roachpb.BatchResponse, error) {
> ```

does removing this function improve performance? keeping it has the benefit of not introducing named returns and the deferred error handling.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

tamird:
Reviewed 2 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

spencerkimball:

---

*[pkg/kv/transport.go, line 203 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-Kx9S8MqdlPReQ00hUtT:-Kx9TzwcPMZXpHjOWco-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/70eeefde44af0b5597a19f0e3343122daaa38594/pkg/kv/transport.go#L203)):*

does removing this function improve performance? keeping it has the benefit of not introducing named returns and the deferred error handling.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

tamird:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200)*

petermattis:
:lgtm: though what is wrong with your laptop? This is what I see on mine (without this change):

```
   60.0s        0        1454971        24248.9      1.3      0.9      4.2      6.8     75.5
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-:-Kx9WRd-wnfiSWSUXfCs:bqv1jk0)*

spencerkimball:
I know...not sure why my laptop is so slow.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19200#-:-Kx9agttRrMp6dABMBS5:bfxkn2)*

Issue #19214, opened 2017-10-12
release: v1.1.0


lego:
Candidate SHA: 8b86503
Deployment status: `adriatic`, `omega`

[Release process checklist](https://github.com/cockroachdb/cockroach/wiki/Release-Process)

Prep date: `N/A`
Skipped as v1.1.0-rc1 (#18903) is being promoted and was already prepped.

Release date: `10/11`
- [X] [Check cluster status](https://github.com/cockroachdb/cockroach/wiki/Release-Process#check-cluster)
- [X] [Tag release](https://github.com/cockroachdb/cockroach/wiki/Release-Process#tag-release)
- [x] [Trigger release build](https://github.com/cockroachdb/cockroach/wiki/Release-Process#trigger-build)
- [x] [Check binaries](https://github.com/cockroachdb/cockroach/wiki/Release-Process#check-binaries)

- For stable releases only:
  - [x] Update the [Homebrew Formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cockroach.rb) (PR in-flight: https://github.com/Homebrew/homebrew-core/pull/19374)
  - [x] Update the image tag in our [Kubernetes configurations](https://github.com/cockroachdb/cockroach/wiki/Release-Process#kubernetes-configurations)
  - [x] Update the version label in our [Red Hat Atomic Docker image](https://github.com/cockroachdb/cockroach/blob/master/build/deploy/rhel/Dockerfile). For major/minor releases, also update the documentation link in the [image's help file](https://github.com/cockroachdb/cockroach/blob/master/build/deploy/rhel/help.1)
  - [x] Announce version to the [registration cluster](https://github.com/cockroachlabs/production#publishing-a-new-release-version)

- [x] Update docs
- [x] External communications for release
- [x] File issue for next release and assign to release coordinator

lego:
Binaries built and docs are up. Filed issues for updating the homebrew formula (Homebrew/homebrew-core#19374), Kubernetes tags (https://github.com/kubernetes/kubernetes/pull/53814, https://github.com/kubernetes/charts/pull/2475) as well as our own Kubernetes and RHEL configs (https://github.com/cockroachdb/cockroach/pull/19218).

PR #19218, opened 2017-10-12
build,cloud: Bump Kubernetes/RHEL Docker image to 1.1.0

lgo:
#19214

### DIFF
--- a/build/deploy/rhel/Dockerfile
+++ b/build/deploy/rhel/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Alex Robinson <alex@cockroachlabs.com>
 
 LABEL name="cockroachdb/cockroach" \
       vendor="Cockroach Labs" \
-      version="1.0.6" \
+      version="1.1.0" \
       release="1" \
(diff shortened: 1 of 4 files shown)


a-robinson (approved):
Published: https://access.redhat.com/containers/#/registry.connect.redhat.com/cockroachdb/cockroach

a-robinson:
Is there a reason that you closed this instead of merging it? We do typically want to persist these updates to the repo.

lego:
o.O

Ah, I went to rebase because the base was failing due to links changing post-release and seem to have reset it instead. My bad!

a-robinson:
Would you mind pulling in the flag changes from #18485 while you're at it? Either that or I'll merge that one after this, but we need to start specifying the `--cache` and `--max-sql-memory` flags to avoid limiting our memory usage too much.

lego:
Pulled them in! Will merge when build passes.

lego:
🤔 should this also be cherry-picked to 1.1? I don't know how these configs are typically used or referenced, but I reckon if anyone looks at the 1.1 branch it would be nice to see these versions.

a-robinson:
Yeah, that would be reasonable, for the Kubernetes configs at least.

PR #17535, opened 2017-08-09
RFC: Change Data Capture

rjnn:
This RFC proposes a mechanism for subscribing to changes to a table or
database. It will support the Apache Kafka publisher API. It will make
use of a Replica level subsystem "ChangeFeed", the design of which is
described in #16838.

Issue #16593, opened 2017-06-19
read from a follower with timestamp bound


tbg:
I ended up thinking about this tonight due to a related problem, so here are
some notes. The difficulty is making this zone configurable. Might've missed
something.

## Goals

- assuming a read timestamp far enough in the "past", (usually) be able to read
  from any replica. (think: analytics, time travel queries, backups, queries
  that can't or don't need to pay the latency to a far-away lease holder).
- configurable on the level of zone configs (i.e. table)

## Sketch of implementation

Add a field `max_write_age` to the zone configs (a value of zero behaves like
MaxUint64). The idea is that the timestamp caches of the affected ranges have
a low watermark that does not trail `(now-max_write_age)`. Note that this
effectively limits how long transactions can write to approximately
`max_write_age`. In turn, when running a read-only transaction, once the
current HLC timestamp has passed `read_timestamp + max_write_age + max_offset`,
any replica can serve reads.

1. add a field `max_write_age` to the lease proto.
1. whenever a lease is proposed, `max_write_age` is populated with the value
the proposer believes is current.
1. lease extensions must not alter `max_write_age`. If a lease holder realizes
that the ZoneConfig's `max_write_age` has changed, it must request a new lease
(in practice, it only has to do this in case `max_write_age` increases) and let
the old one expire (or transfer its lease away). There is room for optimization
here: the replica could extend the lease with the new `max_write_age`, but all
members must enforce the smaller `max_write_age`s for as long as the "old"
version is not expired.
1. Make DistSender aware of `max_write_age`. When considering a read-only
`BatchRequest` with a timestamp eligible for a follower-served read, consider
followers, prioritizing those in close proximity.
1. A follower which receives a read-only batch first checks if the current
lease is active (not whether it holds the lease itself). If not, it behaves as
it would today (requests the lease). Otherwise, if it is not the lease holder,
it checks if the batch timestamp is eligible for a follower-served read based
on the information in the lease and the current timestamp. If so, it serves it
(it does not need to update the timestamp cache).
1. on writes that violate `now - max_write_age < write_ts`, behave as if there
were a timestamp cache entry at `now`.

An interesting observation is that this can also be modified to allow serving
read queries when Raft completely breaks down (think all inter-DC connections
fail): a replica can always serve what is "safe" based on the last known lease.
There is much more work to do to get these replicas to agree on a timestamp,
though. The resulting syntax could be something along the lines of

`SELECT (...) AS OF SYSTEM TIME STALE`

[...]

PR #19222, opened 2017-10-12
RFCs: Max safe timestamp for follower reads and change feeds

a-robinson:
Mostly just a more verbose transcription of the notes I took during
Nathan and I's discussion yesterday. Still a WIP compared to the
expected format for RFCs.

Really just a companion document to https://github.com/cockroachdb/cockroach/pull/17535 and https://github.com/cockroachdb/cockroach/pull/16838, but sending out for feedback.

@nvanbenschoten @tschottdorf @bdarnell @petermattis

### DIFF
--- a/docs/RFCS/00000000_max_safe_timestamp.md
+++ b/docs/RFCS/00000000_max_safe_timestamp.md
@@ -1,0 +1,300 @@
+- Feature Name: Max safe timestamp
+- Status: draft
+- Start Date: 2017-10-11
+- Authors: Alex Robinson, Nathan VanBenschoten
+- RFC PR: (PR # after acceptance of initial draft)
+- Cockroach Issue: [#2656](https://github.com/cockroachdb/cockroach/issues/2656),
+                   [#6130](https://github.com/cockroachdb/cockroach/issues/6130),
+                   [#9712](https://github.com/cockroachdb/cockroach/issues/9712),
+                   [#16593](https://github.com/cockroachdb/cockroach/issues/16593),
+                   [#16838](https://github.com/cockroachdb/cockroach/pull/16838),
+                   [#17535](https://github.com/cockroachdb/cockroach/pull/17535)
+
+# Summary
+
+Propose mechanisms for tracking what has come to be known as the "max safe
+timestamp" for a range, which is the maximum timestamp at which a range
+guarantees no more writes will be allowed to occur. While the thinking to this
+point has been that follower reads and change feeds (i.e. CDC) could share an
+approach to tracking the max safe timestamp, we believe that the different needs
+of the two features makes partially decoupling them advantageous.
+
+# Motivation
+
+The motivation is best understood by reading through the RFCs for
+[CDC](https://github.com/cockroachdb/cockroach/pull/17535) and [Change
+Feeds](https://github.com/cockroachdb/cockroach/pull/16838) as well as the
+[issue for follower
+reads](https://github.com/cockroachdb/cockroach/issues/16593). To summarize,
+non-leaseholder replicas need to know the range of timestamps at which it is
+safe to serve reads without coordinating with the leaseholder, and change feeds
+require what are known as "close notifications", which indicate no new writes
+will be allowed to happen at timestamps earlier than some "closed" timestamp.
+
+# Guide-level explanation
+
+TODO: Initial version of this RFC is just for reaching consensus on rough ideas
+
+# Reference-level explanation
+
+TODO: Initial version of this RFC is just for reaching consensus on rough ideas
+
+## Detailed design
+
+### Goals
+
+* Enable follower reads and change feeds.
+* Don't affect the performance of clusters that aren't using the above features.
+* Don't affect the performance of ranges that aren't using the above features.
+* Minimize the performance impact on writes to ranges with active follower reads
+  or changes feeds. A small performance hit caused by increased resource usage
+  is to be expected, but there should be no blocking or artificial delays
+  introduced.
+* Allow users to customize the trade-off between allowing older writes and
+  allowing more recent follower reads / close notifications.
+
+### Follower reads
+
+The basic approach to implementing follower reads was outlined [in a comment on
+the
+issue](https://github.com/cockroachdb/cockroach/issues/16593#issuecomment-309549461).
+The high-level breakdown of how it works is:
+
+* All replicas track the highest write timestamp they have appended to their
+  raft log, which we'll call the `max_write_timestamp`.
+* The leaseholder will also track the `max_proposed_write_timestamp`.
+* The leaseholder promises not to propose a write with timestamp less than
+  `max_proposed_write_timestamp - max_write_age`, where `max_write_age` is a
+  configurable setting that will limit how old writes can be. The default
+  value for `max_write_age` will be in the tens of seconds, but we expect that
+  setting it down to just a few seconds will be common if we can support it.
+  * Making this safe will require taking these timestamps into account when
+    proposing or re-proposing commands.
+  * Making this safe may also require the leaseholder (in rare cases) to propose
+    empty commands to make sure a write which has been proposed but has not yet
+    committed will fail the `LeaseAppliedIndex` / `MaxLeaseIndex` check if they
+    do eventually commit. This would be to prevent a write from getting
+    arbitrarily delayed between the leaseholder proposing the command and the
+    command actually making it into the raft log. A long enough delay before the
+    command makes it into the raft log could cause a violation of the
+    `max_proposed_write_timestamp` promise, for example if the delay is longer
+    than the range's configured `max_write_age`. Proposing empty commands would
+    increase the raft log's index such that the delayed command would fail its
+    `LeaseAppliedIndex` check, preventing it from applying with a timestamp
+    older than `max_write_timestamp - max_write_age`.
+* Followers can serve reads at times less than their `max_write_timestamp -
+  max_write_age`.
+  * This may mean the some followers can serve reads that others
+    can't, but all such reads will be correct as long as the leaseholder doesn't
+    break its promise about not proposing old writes.
+  * Intents encountered by the followers can be resolved as necessary. This will
+    add some additional latency to such reads, but maintains correctness. Any
+    intents left by transactions that are still pending should have their
+    transactions aborted at this point.
+* If a follower receives a stale read request that it can't serve, it has to
+  forward it to the leaseholder via a NotLeaseHolderError.
+* To make all of this usable, the `DistSender` will have to be instrumented to
+  decide which replica is closest, decide whether a follower may be able to
+  handle the request, and forward such requests to the leaseholder if the
+  follower can't.
+* The cluster's `maxOffset` doesn't need to be considered. During steady-state
+  operation under a single range lease, all timestamps for writes are getting
+  verified by the same leaseholder. When the lease changes hands, the new
+  leaseholder can base its new `max_proposed_write_timestamp` on when the
+  previous lease expired, as described in more detail below.
+
+There are still a couple big questions left.
+
+As long as new writes are happening on the range, all the replicas will
+naturally have their `max_write_timestamp` kept up-to-date by the timestamps on
+the newly proposed writes. However, if no write has been proposed for a while,
+the followers' `max_write_timestamp` will fall far behind the actual current
+time, limiting the usefulness of follower reads in inactive ranges. This
+suggests that we need periodic "heartbeats" of some sort to keep followers
+up-to-date with respect to which timestamps are safe to read. There are two
+options here:
+
+1. Have the leaseholder periodically propose an empty command to raft with the
+   current timestamp if no other writes have come through lately.
+2. Use a higher-level mechanism, such as attaching a new field to raft
+   heartbeats or piggybacking on quiescence to convey this information.
+
+Option 2 is lower overhead but riskier, since there's no guarantee that the new
+leaseholder after a lease change will know about the increased
+`max_write_timestamp`, but it seems as though a new leaseholder could just be
+extra cautious when it takes over the lease. The previous leaseholder couldn't
+have proposed a write with a timestamp greater than its lease expiration, so new
+leaseholders can start out by setting `max_proposed_write_timestamp` to the
+previous lease's expiration timestamp.
+
+The next bigger question is whether this extra work to keep
+`max_write_timestamp` up to date should always be enabled by default. There are
+a few options here:
+
+1. Always enabled. Leaseholders always have to do the work of keeping
+   `max_write_timestamp` up-to-date on followers.
+2. Only enabled when a user explicitly sets a `max_write_age` via zone config.
+3. Only enabled when a leaseholder has reason to believe that followers are
+   receiving read requests at old timestamps.
+4. A combination of 2 and 3.
+
+Option 1 is a bit of a non-starter due to the extra work required, since we
+shouldn't be slowing down users' clusters for features that they aren't using.
+It also would prevent ranges from [quiescing](20160824_quiesce_ranges.md).
+Option 2 is good because it gets explicit user consent, but bad because it means
+that users have to know about the feature before they're able to benefit from
+it. We'll at least want an option to configure `max_write_age`, though, since
+there could be workloads out there with really long-running writes that work
+today but wouldn't work with a small default `max_write_age`.  Heuristics aren't
+fun, but option 3 seems best as long as we can reasonably decide when to start,
+when to stop, and how frequently to update `max_write_timestamp` in the absence
+of writes. Instrumenting the `DistSender` to indicate that a stale read was
+rejected by a follower would be a good way of notifying the leaseholder to
+start, but deciding when to stop may just be a manner of timing out every so
+often.
+
+There are more questions that could be asked about follower reads (e.g. defining
+a new syntax for stale reads at a system-chosen timestamp), but those can be
+deferred to a follow-up PR or a later extension to this one.
+
+### Change feeds
+
+Unlike for follower reads, change feeds / CDC need to care very dearly about
+unresolved intents. A change feed can't publish a write to any listeners until
+it knows that the write's transaction has committed, meaning that it publishes
+new writes as the intents are resolved.
+
+Follower reads can simply resolve an intent when it encounters one, whereas
+change feeds need to make promises to clients in the form of "close
+notifications" that no new writes will come in at an old timestamps. The
+leaseholder can avoid proposing new writes at old timestamps using the same
+`max_proposed_write_timestamp - max_write_age` approach as for follower reads,
+but any existing intents on the range are also potential writes. An unresolved
+intent could become a valid write that the leaseholder doesn't learn about until
+any arbitrary later time unless it proactively tracks intents and resolves them.
+
+Here's a summary of our thoughts:
+
+* We can reuse the `max_write_timestamp` tracking from follower reads, but more
+  logic is needed to ensure no old intents are left unresolved.
+* We have to track unresolved intents somewhere - scanning the entire range to
+  check for intents every time we want to send a close notification is going to
+  be prohibitively expensive unless close notifications can be very infrequent.
+  * Tracking intents  will still be expensive, so we should only do so for
+    ranges that have an active change feed listening to them.
+* Where should we track intents?
+  * In-memory on all replicas: makes failover easy if a change feed needs to
+    switch which one it's talking to (e.g. because a node went down), but will
+    cost us a lot of memory usage. Benchmarking may demonstrate this is
+    feasible, but until then it seems too expensive.
+  * On-disk: we considered some linked list approaches where pointers are stored
+    as part of each intent to track the intents in a range, but couldn't come up
+    with a scheme that would be reasonably efficient to keep up-to-date.  If
+    possible, this would get around the need to track things in memory, but
+    doesn't seem very promising from an implementation perspective.
+  * In-memory on only the replica the change feed is attached to: minimizes
+    resources consumed (no disk I/O, just memory on one node), but failover to a
+    different replica will require work to reconstruct the in-memory data
+    structure. This appears to be the way to go (more detail below).
+* On lease transfer, the change feeds RFC already proposes a form of ["resume
+  tokens"](https://github.com/cockroachdb/cockroach/pull/16838/files#diff-6cc9e2a4b26d4ee01d624c7015fc5220R242).
+  These are meant to allow a new leaseholder to resume the change feed by
+  finding the corresponding point in its raft log and catching up by reading
+  from there. We believe the same can be done for intent-tracking -- if the
+  resume token also includes a reference to the point in the raft log
+  corresponding to the last "close notification", we know that there are no
+  intents remaining from before that point. The new leaseholder can scan on from
+  that point to reconstruct the in-memory intent-tracking data structure.
+  * In the case that the raft log has been truncated, a full range scan will be
+    required, as is already the case for the original resume token idea.
+* Given the above information, the replica serving the change feed (presumably
+  the leaseholder, but I do think we could support followers as well) can emit a
+  value as soon as its intent is resolved, and can emit a close notification for
+  any time less than `max_write_timestamp - max_write_age` and earlier than the
+  oldest remaining intent.
+  * The replica serving the change feed also has the right to proactively
+    resolve any unresolved intents older than `max_write_timestamp -
+    max_write_age` in order to enable the sending of a close notification.
+* Like for follower reads, we won't enable the requisite tracking of intents
+  unless there is a change feed attached to the range. If there isn't, there is
+  no reason to incur the cost of doing so.
+
+Thus, the max-safe timestamp needed by change feeds is largely additive on top
+of the max-safe timestamp used by follower reads, but with the optional
+optimization of not needing to send `max_write_timestamp` updates to followers
+if change feeds only attach to the leaseholder.
+
+## Drawbacks
+
+TODO: Initial version of this RFC is just for reaching consensus on rough ideas
+
+## Rationale and Alternatives
+
+### Use RocksDB time-bound iterators to find recent intents on-demand
+
+One alternative to consider for the tracking of intents is to rely on RocksDB's
+`TimeBoundIterator` to only scan over recently-written SSTables. This feature
+allows you to provide a range of MVCC timestamps and only read through the
+SSTables containing entries at those timestamps. This would allow us to just
+maintain the timestamp of the oldest intent and periodically do a
+`TimeBoundIterator` scan over all MVCC keys written after that timestamp,
+updating what is considered the oldest intent appropriately. This would
+alleviate the need for extra in-memory tracking and for any sort of coordination
+required to update the in-memory data structure on writes.
+
+The downside, though, is that it necessarily operates on all the ranges in a
+store, even if change feeds are only active on a small subset of the ranges.
+Thus, it makes the performance impact of change feeds scale with the writes on
+the entire cluster rather than with the writes on the tables or ranges being
+subscribed to, which could be problematic for many deployments. Tuning how
+frequently we ran such scans would also be tricky, given the need to balance the
+timeliness of close notifications with the cost of all the iteration.
+
+### Replicate the TimestampCache through Raft
+
+This alternative is only half serious, but it serves as a nice generalization of
+"max safe timestamp" that provides more insight into the concept. Instead of
+proposing dummy write commands to update the `max_write_timestamp` through Raft,
+the leaseholder could instead periodically bump its `TimestampCache` low water
+mark and send the entire structure through Raft. Followers could then maintain a
+copy of this cache themselves and serve any read requests that they observe
+where the timestamp is equal to or less than the minimum timestamp for all spans
+in the request within the `TimestampCache`.
+
+The benefit of this approach over replicating `max_write_timestamp` is that
+follower reads would not always need to trail the current time by more than
+`max_write_age`. If the `TimestampCache` is already updated for a read over a
+certain span on the leaseholder, it is guaranteed that no writes will take place
+under this timestamp within this span. This means that for certain spans of keys
+that already have an updated timestamp cache value on the leaseholder, followers
+could read locally at that updated time once the update is replicated through
+Raft. This would allow them to read locally for timestamps much closer to the
+present time.
+
+We can equate "max safe timestamp" to the low water mark of the `TimestampCache`
+in this alternative. The two ideas are analogous if we imagine that the low
+water mark of the cache is set to `max_write_timestamp-max_write_age` and that
+the cache maintains no timestamp intervals other than this low water mark.
+
+This has an interesting extension, where followers could proactively request
+`TimestampCache` updates, on-demand. They could then wait for the replicated
+cache update so they could field reads for certain spans locally. While this
+would still incur a round-trip, it could be used to read up-to-date information
+from followers when it's apparent that the read will create a very large result
+set and shipping this result over the network will be much more expensive than
+shipping the bookkeeping required to allow a follower read. The origins of this
+idea came from benefit 4 in the original comment of [this forum
+post](https://forum.cockroachlabs.com/t/why-do-we-keep-read-commands-in-the-command-queue/360).
+
+Or course, shipping the `TimestampCache` through Raft would probably be too
+expensive to be feasible.
+
+## Unresolved questions
+
+- We could use some eyes on the proposal/re-proposal question. Re-proposals got
+  brought up as a major roadblock on the original follower read issue, but we
+  aren't sure why the logic there can't just be changed to consider timestamps
+  when deciding whether to re-propose. Also, the method of proposing empty
+  commands to time out delayed proposals could benefit from review.
+- The user interface for follower reads still needs to be thought out, which can
+  either be done in a separate RFC or a follow-on here.
--- a/docs/RFCS/00000000_max_safe_timestamp.md
+++ b/docs/RFCS/00000000_max_safe_timestamp.md
@@ -289,6 +289,46 @@ post](https://forum.cockroachlabs.com/t/why-do-we-keep-read-commands-in-the-comm
 Or course, shipping the `TimestampCache` through Raft would probably be too
 expensive to be feasible.
 
+### Replicate the TimestampCache through Raft
+
+This alternative is only half serious, but it serves as a nice generalization of
+"max safe timestamp" that provides more insight into the concept. Instead of
+proposing dummy write commands to update the `max_write_timestamp` through Raft,
+the leaseholder could instead periodically bump its `TimestampCache` low water
+mark and send all `TimestampCache` entries from its store that overlap with its
+key span through Raft. Followers could then maintain a copy of this cache
+themselves and serve any read requests that they receive where the timestamp is
+equal to or less than the minimum timestamp for all spans in the request within
+the `TimestampCache`.
+
+The benefit of this approach over replicating `max_write_timestamp` is that
+follower reads would not always need to trail the current time by more than
+`max_write_age`. If the `TimestampCache` is already updated for a read over a
+certain span on the leaseholder, it is guaranteed that no writes will take place
+under this timestamp within this span. This means that for certain spans of keys
+that already have an updated timestamp cache value on the leaseholder, followers
+could read locally at that updated time once the update is replicated through
+Raft. This would allow them to read locally for timestamps much closer to the
+present time.
+
+We can equate "max safe timestamp" to the low water mark of the `TimestampCache`
+in this alternative. The two ideas are analogous if we imagine that the low
+water mark of the cache is set to `max_write_timestamp-max_write_age` and that
+the cache maintains no timestamp intervals other than this low water mark.
+
+This has an interesting extension, where followers could proactively request
+`TimestampCache` updates, on-demand. They could then wait for the replicated
+cache update so they could field reads for certain spans locally. While this
+would still incur a round-trip, it could be used to read up-to-date information
+from followers when it's apparent that the read will create a very large result
+set and shipping this result over the network will be much more expensive than
+shipping the bookkeeping required to allow a follower read. The origins of this
+idea came from benefit 4 in the original comment of [this forum
+post](https://forum.cockroachlabs.com/t/why-do-we-keep-read-commands-in-the-command-queue/360).
+
+Or course, shipping the `TimestampCache` through Raft would probably be too
+expensive to be feasible.
+
 ## Unresolved questions
 
 - We could use some eyes on the proposal/re-proposal question. Re-proposals got


danhhz:
so excited to see this get started!

---

Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGu1sSEtGYY42DRe-v:-KwGu1sSEtGYY42DRe-w:bkxrqm9) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L31)):*
> ```Markdown
> non-leaseholder replicas need to know the range of timestamps at which it is
> safe to serve reads without coordinating with the leaseholder, and change feeds
> require what are known as "close notifications", which indicate no new writes
> ```

when i talked to alex/nathan about this yesterday, i was a bit confused about what "no new writes" meant. as i recall from the explanation, it means even intents won't be resolved on that replica (and why this is important), which is not what i could have guessed from this phrasing. this could be that i'm not up on my core terminology, but making it more explicit couldn't hurt

update: ah, i see this is explained in more detail below. maybe move a bit of that context up here

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGvS3K9S82sG1Xf028:-KwGvS3K9S82sG1Xf029:b-13v7fc) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L49)):*
> ```Markdown
> the
> issue](https://github.com/cockroachdb/cockroach/issues/16593#issuecomment-309549461).
> The high-level breakdown of how it works is:
> ```

can you also include how max clock offset plays into the below?

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGuvrYFJi4GgpY3CCX:-KwGuvrYFJi4GgpY3CCY:b-b6kgjl) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L55)):*
> ```Markdown
> * The leaseholder will also track the `max_proposed_write_timestamp`.
> * The leaseholder promises not to propose a write with timestamp less than
>   `max_proposed_write_timestamp - max_write_age`, where `max_write_age` is a
> ```

what do we imagine to be the range of typical values for max_write_age?

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 151 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGwRajZxEMWK2kuGTt:-KwGwRajZxEMWK2kuGTu:b-xvp5os) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L151)):*
> ```Markdown
> * We can reuse the `max_write_timestamp` tracking from follower reads, but more
>   logic is needed to ensure no old intents are left unresolved.
> * We have to track unresolved intents somewhere - scanning the entire range to
> ```

not sure it would be enough to make any of these decisions differently, but I just want to make sure y'all know about `engine.NewTimeBoundIterator` which is given a range of our mvcc timestamps and uses that to skip any sstables that don't contain those times (we store the mvcc timestamps contained in an sstable's metadata every time we write one). these intents would be very new, which maximizes the chance that you'd be scanning only small sstables

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-:-KwGtzV6c9CDUDNxdvFP:bgi9zsv)*

nvanbenschoten:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGu1sSEtGYY42DRe-v:-KwMT6qhn8F2vKN50QJs:b-31vcv5) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L31)):*

when i talked to alex/nathan about this yesterday, i was a bit confused about what "no new writes" meant. as i recall from the explanation, it means even intents won't be resolved on that replica (and why this is important), which is not what i could have guessed from this phrasing. this could be that i'm not up on my core terminology, but making it more explicit couldn't hurt

update: ah, i see this is explained in more detail below. maybe move a bit of that context up here
</blockquote></details>

Yeah, that's the important distinction that we're trying to draw here. "No new writes" in the sense of follower reads means that no transactions can be committed with timestamps lower than some timestamp threshold. This means that if a scan occurs at the timestamp threshold, the only intents that can be seen will be those for already committed transactions (which can be cleaned up by the scan, if necessary). Since a follower won't be updating the `TimestampCache`, this is necessary to provide serializable reads on followers.

Closing out a timestamp as is needed by CDC is a superset of this requirement that also requires that all intents for transactions, even those already committed, must also be resolved by that timestamp threshold. This means that if a scan occurs at the timestamp threshold, it will never see any intents.

The reason we think these should be decoupled is that the latter requires significantly more bookkeeping to guarantee. Based on past discussions with Alex and Toby, I don't think it's possible to provide the second guarantee efficiently without retaining some kind of in-memory or persistent data structure that tracks pending intents. However, as the rest of this RFC describes, the first guarantee can be provided without nearly as much bookkeeping. In fact, other than the dummy writes (which if you squint are lease renewals), the first guarantee can be provided without any real performance concern.

One of our focuses throughout this discussion was to avoid paying for what isn't being used. If a range isn't using follower reads, it shouldn't pay the cost of these dummy writes. Likewise, if a range isn't monitoring a change feed, it shouldn't pay for the cost of tracking intents. By splitting `max_safe_timestamp` and `max_safe_intent_timestamp` (or whatever it's called) we can provide follower reads without paying the full cost required by CDC.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGuvrYFJi4GgpY3CCX:-KwMTLGgNPM9xJe3Fob5:bcwc3vq) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L55)):*

what do we imagine to be the range of typical values for max_write_age?
</blockquote></details>

It really depends on the use cases we want to support. 10 seconds has been thrown around in the past, but I think that was just a holdover from arbitrary limitations we previously placed on the `TimestampCache`. Making this easily configurable is important because I could easily see this ranging from seconds to hours, depending on the workload of an application and the properties it's desiring from Cockroach. 

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMUCXsZw0X8Yw9Wvlx:-KwMUCXsZw0X8Yw9Wvly:bxxg6fq) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L73)):*
> ```Markdown
>     add some additional latency to such reads, but maintains correctness. Any
>     intents left by transactions that are still pending could reasonably have
>     their transactions aborted at this point.
> ```

If the transaction associated with these intents has a timestamp below `max_write_timestamp - max_write_age` then they necessarily will have to be aborted (or at least pushed to a later timestamp).

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMZ_k7yPSoFajd9031:-KwMZ_k7yPSoFajd9032:b-22h0np) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L83)):*
> ```Markdown
> There are still a couple big questions left.
> 
> If no write has been proposed for a while, the followers' `max_write_timestamp`
> ```

Make a note of why this isn't an issue if the range is seeing any writes somewhere.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMYnd1uVKh7L8Y5Oyc:-KwMYnd1uVKh7L8Y5Oyd:b-8awlg9) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L99)):*
> ```Markdown
> extra cautious when it takes over the lease. For example, the previous
> leaseholder couldn't have proposed a write with a timestamp greater than
> `time.Now() + maxOffset`, so new leaseholders could start out by setting
> ```

nit: `clock.Now()` since we're dealing with `HLC timestamps`.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 106 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMZWhT4QyNiWRPOmDe:-KwMZWhT4QyNiWRPOmDf:b-rptw51) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L106)):*
> ```Markdown
> a few options here:
> 
> 1. Always enabled. Leaseholders always have to do the work of keeping
> ```

nit: These numbers got messed up.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMZKfepzLLy4DSVpNr:-KwMZKfepzLLy4DSVpNs:b79w0b5) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L114)):*
> ```Markdown
> 
> Option 1 is a bit of a non-starter due to the extra work required, since we
> shouldn't be slowing down users' clusters for features that they aren't using.
> ```

This also prevents ranges from quiescing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222)*

a-robinson:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGvS3K9S82sG1Xf028:-Kw_cowNqu2HhKQq6ceV:b-yjuo3) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L49)):*

can you also include how max clock offset plays into the below?
</blockquote></details>

I don't think it needs to play into the below, actually. At least not as far as follower reads are concerned. All timestamps are going through a single leaseholder except when the lease changes hands, so as long as we consider `maxOffset` as described below when a new lease is acquired, the leaseholder can just use HLC timestamps directly when comparing incoming writes with its `max_proposed_write_timestamp` or when updating its `max_proposed_write_timestamp`.

I've added a new bullet point about this. I'd love to be corrected now, though, if you can think of something I haven't.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGuvrYFJi4GgpY3CCX:-Kw_WBalYDmDShj8wOBe:b-8b9u11) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L55)):*

It really depends on the use cases we want to support. 10 seconds has been thrown around in the past, but I think that was just a holdover from arbitrary limitations we previously placed on the `TimestampCache`. Making this easily configurable is important because I could easily see this ranging from seconds to hours, depending on the workload of an application and the properties it's desiring from Cockroach. 
</blockquote></details>

I'd expect we'll put it in the low tens of seconds by default. Added a little note to the text.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMUCXsZw0X8Yw9Wvlx:-Kw_WTJNVyCls_YoiaYd:b-nu7g18) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L73)):*

If the transaction associated with these intents has a timestamp below `max_write_timestamp - max_write_age` then they necessarily will have to be aborted (or at least pushed to a later timestamp).
</blockquote></details>

Improved the wording.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMZ_k7yPSoFajd9031:-Kw_Xq19vuWzqxRTcT8e:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L83)):*

Make a note of why this isn't an issue if the range is seeing any writes somewhere.
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMYnd1uVKh7L8Y5Oyc:-Kw_XtTHyYaUtMJUFIOR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L99)):*

nit: `clock.Now()` since we're dealing with `HLC timestamps`.
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 106 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMZWhT4QyNiWRPOmDe:-Kw_XzuapgJMUoMGIyj8:bk80ksh) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L106)):*

nit: These numbers got messed up.
</blockquote></details>

Did they? They still rendered fine. I've cleaned them up so they look better in the non-rendered version, though.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMZKfepzLLy4DSVpNr:-Kw_YP16ErDjRiXNdtmt:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L114)):*

This also prevents ranges from quiescing.
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 151 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGwRajZxEMWK2kuGTt:-Kw_YR6JAbEbpHzB8kgD:bk4phvy) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L151)):*

not sure it would be enough to make any of these decisions differently, but I just want to make sure y'all know about `engine.NewTimeBoundIterator` which is given a range of our mvcc timestamps and uses that to skip any sstables that don't contain those times (we store the mvcc timestamps contained in an sstable's metadata every time we write one). these intents would be very new, which maximizes the chance that you'd be scanning only small sstables
</blockquote></details>

Yeah, thanks for pointing that out. I had actually been wondering about whether something like that was possible when thinking things through. It operates on all of the store's data, though, whereas change feeds may only be operating on a small fraction of a store's ranges. It'd also be tough to tune how frequently we ran it, needing to balance timeliness of close notifications with the cost of running it.

Added to the alternatives section.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222)*

arjunravinarayan:
Excited as well!

---

Reviewed 1 of 1 files at r2.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGvS3K9S82sG1Xf028:-Kw_qRIyuR4s9Cb1eI0B:by47aas) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L49)):*

I don't think it needs to play into the below, actually. At least not as far as follower reads are concerned. All timestamps are going through a single leaseholder except when the lease changes hands, so as long as we consider `maxOffset` as described below when a new lease is acquired, the leaseholder can just use HLC timestamps directly when comparing incoming writes with its `max_proposed_write_timestamp` or when updating its `max_proposed_write_timestamp`.

I've added a new bullet point about this. I'd love to be corrected now, though, if you can think of something I haven't.
</blockquote></details>

I agree with Alex that it doesn't come into play.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGuvrYFJi4GgpY3CCX:-Kw_nref6xaWLwLWDA8c:b-w3f75x) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L55)):*

I'd expect we'll put it in the low tens of seconds by default. Added a little note to the text.
</blockquote></details>

I could imagine this being lowered to low single digit seconds if you're using materialized views heavily, so the design shouldn't be so heavyweight as to preclude that setting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-:-Kw_qoch3L8RdU0y7fQ1:baw5de8)*

bdarnell:

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGu1sSEtGYY42DRe-v:-KwbTW9c-3yVWMOa-MRN:b-x1r0v5) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L31)):*
> the only intents that can be seen will be those for already committed transactions

"already finalized transactions", right? The txns can be committed or aborted as long as they're not pending. 

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGuvrYFJi4GgpY3CCX:-KwbSp2OwfHsVJyu4OvD:byzx9pl) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L55)):*

I could imagine this being lowered to low single digit seconds if you're using materialized views heavily, so the design shouldn't be so heavyweight as to preclude that setting.
</blockquote></details>

Or if you're using follower reads, not just materialized views. I think setting this down to a few seconds will be common (if we can support it)

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 51 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbTAtrO-D9M-NlJYvP:-KwbTAtrO-D9M-NlJYvQ:blhbgak) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L51)):*
> ```Markdown
> The high-level breakdown of how it works is:
> 
> * All replicas track the highest write timestamp they have observed, which
> ```

Does "observed" mean "appended to the raft log" or "applied"?

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 63 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbSNeTBsHfsK_12zIx:-KwbSNeTBsHfsK_12zIy:by8eu2f) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L63)):*
> ```Markdown
>     to make sure that writes which have been proposed but have not yet committed
>     will fail the `LeaseAppliedIndex` / `MaxLeaseIndex` check if they do
>     eventually commit. This would be to prevent a write from lingering for too
> ```

What do you mean by "lingering"? The write timestamp won't change, so is this referring to real time? 

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 74 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbTgIuq_tUPGBufzCU:-KwbTgIuq_tUPGBufzCV:b-o64wnk) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L74)):*
> ```Markdown
>     add some additional latency to such reads, but maintains correctness. Any
>     intents left by transactions that are still pending should have their
>     transactions aborted at this point.
> ```

How are the transactions aborted? The transaction record may live on another range; how is this synchronization managed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222)*

docs/RFCS/00000000_max_safe_timestamp.md
```diff
@@ -0,0 +1,241 @@
...
+naturally have their `max_write_timestamp` kept up-to-date by the timestamps on
+the newly proposed writes. However, if no write has been proposed for a while,
+the followers' `max_write_timestamp` will fall far behind the actual current
+time, limiting the usefulness of follower reads in inactive ranges. This
+suggests that we need periodic "heartbeats" of some sort to keep followers
+up-to-date with respect to which timestamps are safe to read. There are two
+options here:
+
+1. Have the leaseholder periodically propose an empty command to raft with the
+   current timestamp if no other writes have come through lately.
```
vivekmenezes: There could be another option to have the leaseholder periodically propose an empty command to raft with a future timestamp and disallow writes to that range until the future timestamp. This is like giving out a lease to the followers until a certain timestamp to allow reads at the current timestamp from followers at the expense of slowing down writes considerably

nvanbenschoten:
After having a pretty interesting discussion with @andreimatei and @bdarnell about [this forum post](https://forum.cockroachlabs.com/t/why-do-we-keep-read-commands-in-the-command-queue/360), I decided to add a new alternative. The idea is only half serious but provides a nice generalization of `max safe timestamp` idea that I think gives some valuable insight. It suggests that we could actually just ship the entire `TimestampCache` to replicas through Raft and these replicas could serve any reads locally that have a timestamp that is equal to or less than the minimum timestamp of all overlapping spans in the `TimestampCache`.

tschottdorf:
Didn't give the CDC portion a close read yet because I had some basic questions about the first part.

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMUCXsZw0X8Yw9Wvlx:-KwkNKipTZDd-nqbywkY:b-5l7nvn) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L73)):*

Improved the wording.
</blockquote></details>

Nathan, why do they *have* to be aborted? I agree that they should (because we don't want to wait). If they *have* to we'd have a problem because the intent may as well have be committed.

Side note: when serving reads from many followers, this will likely mean multiple (x #nodes) racing intent resolutions (we deduplicate somewhat in intent resolver, but only on each node). May not be a huge problem, but should definitely try to avoid intents there in the first place.

An automatically managed backoff would be interesting. Whenever an intent is hit, the lease holder adjusts (doubles?) the gap to the write frontier.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 63 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbSNeTBsHfsK_12zIx:-KwkJ1E6ctI7vltcyMq9:b-62s353) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L63)):*

What do you mean by "lingering"? The write timestamp won't change, so is this referring to real time? 
</blockquote></details>

Unrelated, but this might be a better alternative than waiting for the reproposal (or triggering a reproposal of potentially many old in-flight things) in general. Currently we repropose wildly, but we could send a "canary" which is potentially cheaper. We also don't currently actively invalidate commands once we've observed a higher lease applied index (i.e. if we proposed 3 a while ago but then we see 4 apply, we don't eagerly notify 3 but we probably should (and can gate the check on observing a gap). Might be worth filing, though hopefully we don't need it in practice.

Related: I'm not sure I understand how this mechanism is supposed to work:

1. leaseholder proposes a write at timestamp 10
1. write gets dropped but it will be reproposed randomly later
1. but there's a steady stream of writes that gets through just fine
1. around 9 seconds later, the leaseholder gets really nervous and tries to propose a new command to "cancel" the inflight
1. that proposal also gets stuck so nothing happens. Leaseholder can chose: stall all writes, or risk stale reads. Let's do stale reads because example.
1. at second 10,  the followers happily start dishing out "stale" reads
1. the write finally commits, but basically in violation of the (hypothetical) timestamp cache on the followers

So essentially correctness seems to hinge on being able to control how much time passes between proposal and application, *or* to stop accepting writes.

That seems obviously broken so I must be misunderstanding something fundamental. I pictured that the leaseholder would piggyback a safe timestamp on its proposals (or just send empty proposals when it needs to) that basically carry `max_safe_timestamp`. Followers would then allow reads below that timestamp. (And nothing mandates that we'd even need to eat the Raft overhead - it could just be a new RPC). Would be nice if the mechanism here worked because it has a tiny bit less overhead, but having to stall writes seems super dangerous, and would probably happen especially if we want to get really close behind the write frontier (which would be a huge plus I imagine). The "auto-bumping" mechanism also seems at odd with the CDC use case for which the close notification may end up lagging.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 98 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGrGKbcSNr4IwuoFKc-r2-98:-KwkOV2Q4RnFmaRV45xn:bd9jfyg) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L98)):*

There could be another option to have the leaseholder periodically propose an empty command to raft with a future timestamp and disallow writes to that range until the future timestamp. This is like giving out a lease to the followers until a certain timestamp to allow reads at the current timestamp from followers at the expense of slowing down writes considerably
</blockquote></details>

I don't think we should ever have this mechanism interfere with write traffic. In fact, that should be one of the goals stated in the RFC.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 84 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkO9NbT08Gp98Cuq9o:-KwkO9NbT08Gp98Cuq9p:bju1w6h) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L84)):*
> ```Markdown
>   operation under a single range lease, because all timestamps for writes are
>   getting verified by the same leaseholder. `maxOffset` only needs to come in
>   play when the lease for a range changes hands, as described below.
> ```

Just pointing out that you should try to consider (and perhaps accommodate, if possible) the case `maxOffset = infinity` (clockless reads).

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 108 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkOkyiF0kGFzUQWN1o:-KwkOkyiF0kGFzUQWN1p:b-biymlh) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L108)):*
> ```Markdown
> leaseholder couldn't have proposed a write with a timestamp greater than
> `clock.Now() + maxOffset`, so new leaseholders could start out by setting
> `max_proposed_write_timestamp` to `clock.Now() + maxOffset`.
> ```

It seems that  you should be able to remove `maxOffset` from the equation. You know the previous lease (well, at least at some point in time you knew it) which has an expiration, so just use that as the floor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-:-KwkQ9-0XxjXyKQdGURp:b-izg7rx)*

andreimatei:

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 238 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkYf_on6-0HACuAPUe:-KwkYf_pXdw4aFqNswQ6:b-h5g25e) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L238)):*
> ```Markdown
> "max safe timestamp" that provides more insight into the concept. Instead of
> proposing dummy write commands to update the `max_write_timestamp` through Raft,
> the leaseholder could instead periodically bump its `TimestampCache` low water
> ```

Alex just pointed out to me that the tscache is no longer a per-range structure - it's per store. So we should qualify here that we'd be sending the tscache entries overlaping the range in question.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 267 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkZ46xeyMpkoGH04BL:-KwkZ46xeyMpkoGH04BM:b50cog4) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L267)):*
> ```Markdown
> shipping the bookkeeping required to allow a follower read. The origins of this
> idea came from benefit 4 in the original comment of [this forum
> post](https://forum.cockroachlabs.com/t/why-do-we-keep-read-commands-in-the-command-queue/360).
> ```

We've discussed more than written here, and I think it'd be cool to write all of it up: a generalization to of the generalization is that followers' request don't always (usually?) need to wait for a Raft proposal. The follower needs the leaseholder's tscache to be updated, and it needs to make sure that it is up to date _enough_ with the Raft log to serve the read. So, it could tell the leaseholder what its current raft index is; the leaseholder's tscache could be augmented with a raft index number for each element -> meaning what index a follower needs to be able to serve reads that fall within that tscache entry. Now, when the leaseholder gets a request from a follower, it will:
1. check its tscache, update it if necessary, and observe that the follower is already up to date enough to serve the read
2. if the follower is not up to date enough, then it needs to tell the follower what index it needs to wait for. This will be the current index if the tscache had to be bumped, otherwise it's the raft index indicated by the tscache

If the requests from followers go through the command queue (to synchronize with in-flight writes), then I think no raft proposal is necessary.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222)*

tschottdorf:

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 63 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbSNeTBsHfsK_12zIx:-KwlvcaFr0dO70Qo2giz:b8dm7xt) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L63)):*

Unrelated, but this might be a better alternative than waiting for the reproposal (or triggering a reproposal of potentially many old in-flight things) in general. Currently we repropose wildly, but we could send a "canary" which is potentially cheaper. We also don't currently actively invalidate commands once we've observed a higher lease applied index (i.e. if we proposed 3 a while ago but then we see 4 apply, we don't eagerly notify 3 but we probably should (and can gate the check on observing a gap). Might be worth filing, though hopefully we don't need it in practice.

Related: I'm not sure I understand how this mechanism is supposed to work:

1. leaseholder proposes a write at timestamp 10
1. write gets dropped but it will be reproposed randomly later
1. but there's a steady stream of writes that gets through just fine
1. around 9 seconds later, the leaseholder gets really nervous and tries to propose a new command to "cancel" the inflight
1. that proposal also gets stuck so nothing happens. Leaseholder can chose: stall all writes, or risk stale reads. Let's do stale reads because example.
1. at second 10,  the followers happily start dishing out "stale" reads
1. the write finally commits, but basically in violation of the (hypothetical) timestamp cache on the followers

So essentially correctness seems to hinge on being able to control how much time passes between proposal and application, *or* to stop accepting writes.

That seems obviously broken so I must be misunderstanding something fundamental. I pictured that the leaseholder would piggyback a safe timestamp on its proposals (or just send empty proposals when it needs to) that basically carry `max_safe_timestamp`. Followers would then allow reads below that timestamp. (And nothing mandates that we'd even need to eat the Raft overhead - it could just be a new RPC). Would be nice if the mechanism here worked because it has a tiny bit less overhead, but having to stall writes seems super dangerous, and would probably happen especially if we want to get really close behind the write frontier (which would be a huge plus I imagine). The "auto-bumping" mechanism also seems at odd with the CDC use case for which the close notification may end up lagging.
</blockquote></details>

Ah, silly me. The steady stream of writes would immediately surpass the straggler's lease applied index. I'm still not sure this mechanism is easy to use for the CDC use case, but I understand how the basic version here checks out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222)*

a-robinson:
TFTRs!

---

Review status: 0 of 1 files reviewed at latest revision, 18 unresolved discussions.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGu1sSEtGYY42DRe-v:-KwotqAAuIV7WzzHfg6Z:bgh8ybm) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L31)):*

> the only intents that can be seen will be those for already committed transactions

"already finalized transactions", right? The txns can be committed or aborted as long as they're not pending. 
</blockquote></details>

It seems possible to run into an intent for a transaction that hasn't been committed or aborted, but the transaction would be abort-able at that point.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwMUCXsZw0X8Yw9Wvlx:-KxUAZtAAS-WlBU3Zvks:bk3fnzv) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L73)):*

Nathan, why do they *have* to be aborted? I agree that they should (because we don't want to wait). If they *have* to we'd have a problem because the intent may as well have be committed.

Side note: when serving reads from many followers, this will likely mean multiple (x #nodes) racing intent resolutions (we deduplicate somewhat in intent resolver, but only on each node). May not be a huge problem, but should definitely try to avoid intents there in the first place.

An automatically managed backoff would be interesting. Whenever an intent is hit, the lease holder adjusts (doubles?) the gap to the write frontier.
</blockquote></details>

Right, I don't believe that they'd necessarily have to be aborted.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 51 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbTAtrO-D9M-NlJYvP:-KwouL0ihgJki9t8vxfd:b1dg894) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L51)):*

Does "observed" mean "appended to the raft log" or "applied"?
</blockquote></details>

Great question. I think that appended to the raft log is sufficient.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 63 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbSNeTBsHfsK_12zIx:-KxdyUOUqJShcAbSwaBr:bey24tc) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L63)):*

Ah, silly me. The steady stream of writes would immediately surpass the straggler's lease applied index. I'm still not sure this mechanism is easy to use for the CDC use case, but I understand how the basic version here checks out.
</blockquote></details>

Yeah, "lingering" was referring to real time. I've tried to clarify this, since it was definitely a bit terse.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 74 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbTgIuq_tUPGBufzCU:-KxUAkCxxUPzY40HAHvI:b-cpjgf1) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L74)):*

How are the transactions aborted? The transaction record may live on another range; how is this synchronization managed?
</blockquote></details>

Why would any synchronization be needed beyond the normal synchronous intent resolution that gets done for reads?

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 98 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGrGKbcSNr4IwuoFKc-r2-98:-KxUBuMiNbdlSQKU1wAi:b-4609n5) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L98)):*

I don't think we should ever have this mechanism interfere with write traffic. In fact, that should be one of the goals stated in the RFC.
</blockquote></details>

Yeah, I don't know when someone would want to make that extreme of a tradeoff. Added a short list of goals, including not interfering with writes.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 84 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkO9NbT08Gp98Cuq9o:-KxUnQc1560BJh_kluA_:b-icbf59) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L84)):*

Just pointing out that you should try to consider (and perhaps accommodate, if possible) the case `maxOffset = infinity` (clockless reads).
</blockquote></details>

I'm a bit out of touch with our clockless reads. Do you mind explaining how clockless mode would come into play given that we're basing things on lease expiration timestamps?

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 108 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkOkyiF0kGFzUQWN1o:-KxUX3f3YqasJaQGfejU:b-as3bnk) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L108)):*

It seems that  you should be able to remove `maxOffset` from the equation. You know the previous lease (well, at least at some point in time you knew it) which has an expiration, so just use that as the floor.
</blockquote></details>

Good point, changed.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 238 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwkYf_on6-0HACuAPUe:-Kxe2iPdo06DpseREFdp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L238)):*

Alex just pointed out to me that the tscache is no longer a per-range structure - it's per store. So we should qualify here that we'd be sending the tscache entries overlaping the range in question.
</blockquote></details>

Done.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 269 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-Kxe2wl88bNHjhBSnYBi:-Kxe2wl88bNHjhBSnYBj:ba1u2wc) ([raw file](https://github.com/cockroachdb/cockroach/blob/95aa53e418b3ba291636b1a1ad286046bad6e83e/docs/RFCS/00000000_max_safe_timestamp.md#L269)):*
> ```Markdown
> post](https://forum.cockroachlabs.com/t/why-do-we-keep-read-commands-in-the-command-queue/360).
> 
> Or course, shipping the `TimestampCache` through Raft would probably be too
> ```

Sending it through raft probably would be due to all the extra disk writes that would be involved, but sending this stuff through a separate RPC might not be totally crazy. If you'd like to do an experiment to estimate how much bandwidth/serialization cost there'd be, we could consider this more seriously.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-:-Kxe3W2bTQVfYIlghTP4:bz5mair)*

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwGu1sSEtGYY42DRe-v:-KxiAp57UwF-BzkeNYhO:b7djedr) ([raw file](https://github.com/cockroachdb/cockroach/blob/189efe9d2841abc6e1168ebf17d850d685d4da78/docs/RFCS/00000000_max_safe_timestamp.md#L31)):*

It seems possible to run into an intent for a transaction that hasn't been committed or aborted, but the transaction would be abort-able at that point.
</blockquote></details>

I think it has to be aborted, not just abortable. We have to know that the transaction will never commit, and I don't think we can know that without actually aborting it (as long as it's pending, the commit could be in flight in the raft log). 

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 63 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbSNeTBsHfsK_12zIx:-KxiGjgOOKYRDl93TpnY:bszlc0n) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L63)):*

Yeah, "lingering" was referring to real time. I've tried to clarify this, since it was definitely a bit terse.
</blockquote></details>

If a real command gets delayed going into the raft log, what's stopping these dummy commands from doing the same thing? I think we have to rely solely on timestamps that have appeared in the raft log. 

The "heartbeats" described below to keep `max_write_timestamp` from falling too far behind have the side effect of disqualifying any "lingering" commands, but this is an optimization to advance the timestamps rather than a safety requirement. 

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 74 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KwbTgIuq_tUPGBufzCU:-KxiI0qh_1dCs9Do_Bx9:b-uyzi6h) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5a9d5510bfef180a8985a9ee7978d00e92af2f3/docs/RFCS/00000000_max_safe_timestamp.md#L74)):*

Why would any synchronization be needed beyond the normal synchronous intent resolution that gets done for reads?
</blockquote></details>

Ah, so you're talking about the normal post-read intent resolution. That's fine, but it means that this "max safe timestamp" is no longer a guarantee that local reads will be successful - you may have to push the transaction remotely, and after pushing you can't tell when the local replica has caught up with the intent resolution so you have to read remotely as well.

I was assuming that we wouldn't be advancing the follower read timestamp until all transactions were finalized and all intents resolved (since that's what CDC will require). We can do this simpler version for follower reads although leaving unresolved intents may mean it performs worse than expected. 

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 65 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KxiE0yl8-FSF8MRazm5:-KxiE0yl8-FSF8MRazm6:brgmr8x) ([raw file](https://github.com/cockroachdb/cockroach/blob/e6bd937933f97acfded882e2d617dcc95d2eebea/docs/RFCS/00000000_max_safe_timestamp.md#L65)):*
> ```Markdown
> * All replicas track the highest write timestamp they have appended to their
>   raft log, which we'll call the `max_write_timestamp`.
> * The leaseholder will also track the `max_proposed_write_timestamp`.
> ```

It looks like we require that `max_proposed_write_timestamp` is monotonic across lease holders. How is this guaranteed? (advancing `max_write_timestamp` on log append instead of apply seems to make this trickier)  It's likely that the existing lease mechanisms solve this but we need a more complete analysis of how we guarantee that the start time of a new lease will not allow the max safe timestamp to move backwards (do we rely on the lease duration being shorter than max_write_age?)

---

*[docs/RFCS/00000000_max_safe_timestamp.md, line 128 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222#-KxiKFUT3wELov2ivCjL:-KxiKFUT3wELov2ivCjM:bwzshec) ([raw file](https://github.com/cockroachdb/cockroach/blob/e6bd937933f97acfded882e2d617dcc95d2eebea/docs/RFCS/00000000_max_safe_timestamp.md#L128)):*
> ```Markdown
> have proposed a write with a timestamp greater than its lease expiration, so new
> leaseholders can start out by setting `max_proposed_write_timestamp` to the
> previous lease's expiration timestamp.
> ```

What does this mean for epoch-based leases? Leases have start times; can we use that instead? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19222)*

PR #19224, opened 2017-10-12
ui: update Replicas graph title to Replica Quiescence

dianasaur323:
Addresses https://github.com/cockroachdb/cockroach/issues/19012


couchand:
LGTM.

As an aside, you might consider making changes in your personal fork of the repo rather than directly in cockroachdb/cockroach, so they don't end up in the main branch list.  Of course, if you remember to delete them as soon as the PR merges it's not that big of a deal, but it's nice to keep this repo clean.

PR #19225, opened 2017-10-12
release-1.1: cluster: properly introduce cluster version v1.1

tbg:
We forgot to introduce a proper version for v1.1.0. As a result, clusters
running our v1.1 release identify as v1.0-3, which in turn means that the
upgrade instructions do not work:

https://www.cockroachlabs.com/docs/v1.1/upgrade-cockroach-version.html

(1.1 needs to be replaced by 1.0-3).

We've already released v1.1.0 so unless we decide that we want to yank the relase
(don't think it's necessary) this will only become active in v1.1.1 (before
the release of which we should add another version, v1.1.1). We should think
about linting this or at least making its verification a part of the release
checklist, too.

cc @bdarnell @jseldess

### DIFF
--- a/pkg/settings/cluster/cluster_version.go
+++ b/pkg/settings/cluster/cluster_version.go
@@ -24,7 +24,7 @@ var (
 	BinaryMinimumSupportedVersion = VersionBase
 
 	// BinaryServerVersion is the version of this binary.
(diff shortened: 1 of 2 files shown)


a-robinson (approved):
:( LGTM

We'll also need a similar change on master so that `1.1` gets recognized by future releases.

PR #19227, opened 2017-10-12
sqlccl: pass needed spans to prev backup checks

dt:
Previously we validated previous backups were well-formed and ordered.
We did not, however, pass the spans which needed to be covered, the way we do
when running the matching check in RESTORE, meaning that we'd accept a set of
previous backups that didn't actually cover the spans being backed up.

Re-ordering the steps in the backup plan slightly computes the matching tables
and spans before validating the prior backups (and setting the start time), thus
catching that case where the set of tables (and thus the spans for which we need
complete history in order to restore) has changed.

Another potential approach would be to automatically change the startTime for
the spans for which we are missing history, effectiely de-incrementalizing those
tables. This opens up significant additional complexity though. A simple error
at BACKUP time should at least indicate there is an issue right away, rather
than letting an operator believe they are making usable BACKUPs that cannot
actually be RESTOREd.

### DIFF
--- a/pkg/ccl/sqlccl/backup.go
+++ b/pkg/ccl/sqlccl/backup.go
@@ -119,9 +119,9 @@ func readBackupDescriptor(
 }
 
 // ValidatePreviousBackups checks that the timestamps of previous backups are
-// consistent. The most recently backed-up time is returned.
+// consistent and covers `spans`. The most recently backed-up time is returned.
 func ValidatePreviousBackups(
-	ctx context.Context, uris []string, settings *cluster.Settings,
+	ctx context.Context, uris []string, settings *cluster.Settings, spans []roachpb.Span,
 ) (hlc.Timestamp, error) {
 	if len(uris) == 0 || len(uris) == 1 && uris[0] == "" {
 		// Full backup.
@@ -131,7 +131,7 @@ func ValidatePreviousBackups(
 	for i, uri := range uris {
 		desc, err := ReadBackupDescriptorFromURI(ctx, uri, settings)
 		if err != nil {
-			return hlc.Timestamp{}, err
+			return hlc.Timestamp{}, errors.Wrapf(err, "failed to read backup from %q", uri)
 		}
 		backups[i] = desc
 	}
@@ -140,8 +140,8 @@ func ValidatePreviousBackups(
 	// timestamps to validate the previous backups that this one is incremental
 	// from.
 	lowWaterMark := keys.MinKey
-	_, endTime, err := makeImportSpans(nil, backups, lowWaterMark)
-	return endTime, err
+	_, endTime, err := makeImportSpans(spans, backups, lowWaterMark)
+	return endTime, errors.Wrap(err, "invalid previous backups (a new full backup may be required if a table has been created, dropped or truncated)")
 }
 
 func allSQLDescriptors(ctx context.Context, txn *client.Txn) ([]sqlbase.Descriptor, error) {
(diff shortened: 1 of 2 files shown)


dt:
This will probably be worth a cherrypick into 1.1.1.

mjibson:

---

*[pkg/ccl/sqlccl/backup.go, line 134 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227#-KwHbDG3lYag0kztYWbl:-KwHbDG3lYag0kztYWbm:b-gv1scn) ([raw file](https://github.com/cockroachdb/cockroach/blob/c146878adf96a302433f2e24f66a986372d8e69e/pkg/ccl/sqlccl/backup.go#L134)):*
> ```Go
> 		desc, err := ReadBackupDescriptorFromURI(ctx, uri, settings)
> 		if err != nil {
> 			return hlc.Timestamp{}, errors.Wrapf(err, "failed to read backup from %q", uri)
> ```

is it safe to print the uri here? have access keys been stripped out already?

---

*[pkg/ccl/sqlccl/backup.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227#-KwHbL7qoLxZnBlln41A:-KwHbL7qoLxZnBlln41B:b-olqhvh) ([raw file](https://github.com/cockroachdb/cockroach/blob/c146878adf96a302433f2e24f66a986372d8e69e/pkg/ccl/sqlccl/backup.go#L144)):*
> ```Go
> 	lowWaterMark := keys.MinKey
> 	_, endTime, err := makeImportSpans(spans, backups, lowWaterMark)
> 	return endTime, errors.Wrap(err, "invalid previous backups (a new full backup may be required if a table has been created, dropped or truncated)")
> ```

may be -> is?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227)*

dt:

---

*[pkg/ccl/sqlccl/backup.go, line 134 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227#-KwHbDG3lYag0kztYWbl:-KwHf7KEkf_jiRFECdpu:bsm2ugb) ([raw file](https://github.com/cockroachdb/cockroach/blob/c146878adf96a302433f2e24f66a986372d8e69e/pkg/ccl/sqlccl/backup.go#L134)):*

is it safe to print the uri here? have access keys been stripped out already?
</blockquote></details>

It's the URI in the query, and is going right back to the client that supplied it, so shouldn't ' be a problem?

---

*[pkg/ccl/sqlccl/backup.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227#-KwHbL7qoLxZnBlln41A:-KwHfi8JOFjoPdaDRceV:by9zet2) ([raw file](https://github.com/cockroachdb/cockroach/blob/c146878adf96a302433f2e24f66a986372d8e69e/pkg/ccl/sqlccl/backup.go#L144)):*

may be -> is?
</blockquote></details>

It could just be an ordering issue or missing uri though so I didn't want to assert it as absolutely required.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227)*

mjibson:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227#-:-KwHpNli-9C6N0NStZ7R:bnfp4nl)*

vivekmenezes:
can we merge this and get it on the release branch? Thanks

dt:
@vivekmenezes yep, just want to make sure @danhhz gets a chance to look at it first.

cuongdo:
@dt: @danhhz may be out for a bit, so I'm not sure when he'll get to look at this

danhhz:
:lgtm:

Let's make sure this gets added to the docs

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19227#-:-Kw_qS62aDvaBEUJwwVf:boeju0b)*

dt:
@danhhz the current `known-limitations` mentions this: https://github.com/cockroachdb/docs/pull/1990/files#diff-0cc21edfecfe7ed2ea6e689943ee42caR13

@jseldess (for routing): We can update the known limitation(and the BACKUP docs) to say that as of 1.1.1, if any tables have been created, dropped or truncated, an incremental `BACKUP` will fail and a full backup is required (prior to 1.1.1, the backup would appear to work but, as the current note says, a `RESTORE` would be impossible).

jseldess:
Thanks, @dt. Opened a docs issue. I'll make sure it gets documented for the 1.1.1 release.

PR #19228, opened 2017-10-12
sql: introduce MultiRowFetcher for one-pass row fetches for multiple (interleaved) tables

richardwu:
This PR introduces `MultiRowFetcher` which generalizees `RowFetcher` to
respond to and return rows from multiple tables from the same set of
spans.

`MultiRowFetcher` is initialized with a slice of `MultiRowFetcherArgs`
which encapsulates the arguments for one table in `RowFetcher`.

One key point is that the `reverse bool` argument that exists in
`RowFetcher` can only be enabled on a per-fetcher basis: one cannot scan
in arbitrary directions on a per-table basis (since row kvfetcher works
by scanning KV pairs sequentially).

`MultiRowFetcher` is necessary in order to fetch interleaved tables
efficiently in one-pass. See the RFC for interleaved tables here
https://github.com/cockroachdb/cockroach/pull/19028.

There are no tests yet since I wanted to get some comments on this approach. `MultiRowFetcher`
is backwards compatible with `RowFetcher`: only the `Init` parameters and the response from `NextRow/NextRowDecoded` differs slightly.

cc: @vivekmenezes

### DIFF
--- a/pkg/sql/sqlbase/multirowfetcher.go
+++ b/pkg/sql/sqlbase/multirowfetcher.go
@@ -1,0 +1,879 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package sqlbase
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+
+	"github.com/cockroachdb/cockroach/pkg/internal/client"
+	"github.com/cockroachdb/cockroach/pkg/keys"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
+	"github.com/cockroachdb/cockroach/pkg/sql/parser"
+	"github.com/cockroachdb/cockroach/pkg/util"
+	"github.com/cockroachdb/cockroach/pkg/util/encoding"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
+)
+
+type tableInfo struct {
+	// -- Fields initialized once --
+
+	desc             *TableDescriptor
+	index            *IndexDescriptor
+	isSecondaryIndex bool
+	indexColumnDirs  []encoding.Direction
+	// equivSignature is an equivalence class for each unique table-index
+	// pair. It allows us to check if an index key belongs to a given
+	// table-index.
+	equivSignature []byte
+
+	// The table columns to use for fetching, possibly including ones currently in
+	// schema changes.
+	cols []ColumnDescriptor
+
+	// The set of ColumnIDs that are required.
+	neededCols util.FastIntSet
+
+	// Map used to get the index for columns in cols.
+	colIdxMap map[ColumnID]int
+
+	// One value per column that is part of the key; each value is a column
+	// index (into cols).
+	indexColIdx []int
+
+	// -- Fields updated during a scan --
+
+	keyValTypes []ColumnType
+	extraTypes  []ColumnType
+	keyVals     []EncDatum
+	extraVals   []EncDatum
+	row         EncDatumRow
+	decodedRow  parser.Datums
+}
+
+// RowResponse is returned when NextRow is invoked. It contains the
+// EncDatumRow as well as the table and index descriptors for the corresponding
+// table/index.
+type RowResponse struct {
+	Row   EncDatumRow
+	Desc  *TableDescriptor
+	Index *IndexDescriptor
+}
+
+// DecodedRowResponse returns the decoded Row in terms of Datums as well as the
+// table and index descriptors for the corresponding table/index.
+type DecodedRowResponse struct {
+	Datums parser.Datums
+	Desc   *TableDescriptor
+	Index  *IndexDescriptor
+}
+
+// MultiRowFetcherTableArgs are the arguments passed to MultiRowFetcher.Init
+// for a given table that includes descriptors and row information
+type MultiRowFetcherTableArgs struct {
+	Desc             *TableDescriptor
+	Index            *IndexDescriptor
+	ColIdxMap        map[ColumnID]int
+	IsSecondaryIndex bool
+	Cols             []ColumnDescriptor
+	ValNeededForCol  []bool
+}
+
+// MultiRowFetcher handles fetching kvs and forming table rows for an
+// arbitrary number of tables.
+// MultiRowFetcher should be used in lieu of RowFetcher if for a given
+// set of spans, rows from multiple tables are interleaved (i.e. interleaved
+// tables).
+// Usage:
+//   var mrf MultiRowFetcher
+//   err := mrf.Init(..)
+//   // Handle err
+//   err := mrf.StartScan(..)
+//   // Handle err
+//   for {
+//      res, err := mrf.NextRow()
+//      // Handle err
+//      if res.row == nil {
+//         // Done
+//         break
+//      }
+//      // Process res.row
+//   }
+type MultiRowFetcher struct {
+	// tables is a slice of all the tables and their descriptors for which
+	// rows are returned.
+	tables []tableInfo
+
+	// allEquivSignatures is a map used for checking if an equivalence
+	// signature belongs to any table or table's ancestor. It also maps the
+	// string representation of every table's and every table's ancestors'
+	// signature to the table's index in 'tables' for lookup during decoding.
+	// If 2+ tables share the same ancestor signature, allEquivSignatures
+	// will map the signature to the largest 'tables' index.
+	// The full signature for a given table in 'tables' will always map to
+	// its own index in 'tables'.
+	allEquivSignatures map[string]int
+
+	// reverse denotes whether or not the spans should be read in reverse
+	// or not when StartScan is invoked.
+	reverse bool
+
+	// maxKeysPerRow memoizes the maximum number of keys per row
+	// out of all the tables. This is used to calculate the kvFetcher's
+	// firstBatchLimit.
+	maxKeysPerRow int
+
+	// True if the index key must be decoded.
+	// If there is more than one table, the index key must always be decoded.
+	// This is only false if there are no needed columns and the (single)
+	// table has no interleave children.
+	mustDecodeIndexKey bool
+
+	// returnRangeInfo, if set, causes the underlying kvFetcher to return
+	// information about the ranges descriptors/leases uses in servicing the
+	// requests. This has some cost, so it's only enabled by DistSQL when this
+	// info is actually useful for correcting the plan (e.g. not for the PK-side
+	// of an index-join).
+	// If set, GetRangeInfo() can be used to retrieve the accumulated info.
+	returnRangeInfo bool
+
+	// traceKV indicates whether or not session tracing is enabled. It is set
+	// when beginning a new scan.
+	traceKV bool
+
+	// -- Fields updated during a scan --
+
+	kvFetcher      kvFetcher
+	indexKey       []byte // the index key of the current row
+	prettyValueBuf *bytes.Buffer
+
+	rowReadyTable *tableInfo // the table for which a row was fully decoded and ready for output
+	currentTable  *tableInfo // the most recent table for which a key was decoded
+	keySigBuf     []byte     // buffer for the index key's signature
+	keyRestBuf    []byte     // buffer for the rest of the index key that is not part of the signature
+
+	// The current key/value, unless kvEnd is true.
+	kv                roachpb.KeyValue
+	keyRemainingBytes []byte
+	kvEnd             bool
+
+	// Buffered allocation of decoded datums.
+	alloc *DatumAlloc
+}
+
+// Init sets up a MultiRowFetcher for a given table and index. If we are using a
+// non-primary index, valNeededForCol can only be true for the columns in the
+// index.
+func (mrf *MultiRowFetcher) Init(
+	tables []MultiRowFetcherTableArgs, reverse bool, returnRangeInfo bool, alloc *DatumAlloc,
+) error {
+	if len(tables) == 0 {
+		panic("no tables to fetch from")
+	}
+
+	mrf.reverse = reverse
+	mrf.returnRangeInfo = returnRangeInfo
+	mrf.alloc = alloc
+	mrf.allEquivSignatures = make(map[string]int, len(tables))
+
+	// We must always decode the index key if we need to distinguish between
+	// rows from more than one table.
+	mrf.mustDecodeIndexKey = len(tables) >= 2
+
+	mrf.tables = make([]tableInfo, 0, len(tables))
+	for tableIdx, tableArgs := range tables {
+		table := tableInfo{
+			desc:             tableArgs.Desc,
+			colIdxMap:        tableArgs.ColIdxMap,
+			index:            tableArgs.Index,
+			isSecondaryIndex: tableArgs.IsSecondaryIndex,
+			cols:             tableArgs.Cols,
+			row:              make([]EncDatum, len(tableArgs.Cols)),
+			decodedRow:       make([]parser.Datum, len(tableArgs.Cols)),
+		}
+
+		// We produce references to every signature's reference.
+		equivSignatures, err := TableEquivSignatures(table.desc, table.index)
+		if err != nil {
+			return err
+		}
+		for i, sig := range equivSignatures {
+			// We always map the table's equivalence signature (last
+			// 'sig' in 'equivSignatures') to its tableIdx.
+			// This allows us to overwrite previous "ancestor
+			// signatures" (see below).
+			if i == len(equivSignatures)-1 {
+				mrf.allEquivSignatures[string(sig)] = tableIdx
+				break
+			}
+			// Map each table's ancestors' signatures to -1 so
+			// we know during ReadIndexKey if the parsed index
+			// key belongs to ancestor or one of our tables.
+			// We must check if the signature has already been set
+			// since it's possible for a later 'table' to have an
+			// ancestor that is a previous 'table', and we do not
+			// want to overwrite the previous table's tableIdx.
+			if _, exists := mrf.allEquivSignatures[string(sig)]; !exists {
+				mrf.allEquivSignatures[string(sig)] = -1
+			}
+		}
+		// The last signature is the given table's equivalence signature.
+		table.equivSignature = equivSignatures[len(equivSignatures)-1]
+
+		for i, v := range tableArgs.ValNeededForCol {
+			if !v {
+				continue
+			}
+			// The i-th column is required. Search the colIdxMap to find the
+			// corresponding ColumnID.
+			for col, idx := range table.colIdxMap {
+				if idx == i {
+					table.neededCols.Add(int(col))
+					break
+				}
+			}
+		}
+
+		// If there is more than one table, we have to decode the index key to
+		// figure out which table the row belongs to.
+		// If there are interleaves, we need to read the index key in order to
+		// determine whether this row is actually part of the index we're scanning.
+		// If we need to return any values from the row, we also have to read the
+		// index key to either get those values directly or determine the row's
+		// column family id to map the row values to their columns.
+		// Otherwise, we can completely avoid decoding the index key.
+		// TODO(jordan): Relax this restriction. Ideally we could skip doing key
+		// reading work if we need values from outside of the key, but not from
+		// inside of the key.
+		if !mrf.mustDecodeIndexKey && (!table.neededCols.Empty() || len(table.index.InterleavedBy) > 0 || len(table.index.Interleave.Ancestors) > 0) {
+			mrf.mustDecodeIndexKey = true
+		}
+
+		var indexColumnIDs []ColumnID
+		indexColumnIDs, table.indexColumnDirs = table.index.FullColumnIDs()
+
+		table.indexColIdx = make([]int, len(indexColumnIDs))
+		for i, id := range indexColumnIDs {
+			table.indexColIdx[i] = table.colIdxMap[id]
+		}
+
+		if table.isSecondaryIndex {
+			for i := range table.cols {
+				if table.neededCols.Contains(int(table.cols[i].ID)) && !table.index.ContainsColumnID(table.cols[i].ID) {
+					return fmt.Errorf("requested column %s not in index", table.cols[i].Name)
+				}
+			}
+		}
+
+		// Prepare our index key vals slice.
+		table.keyValTypes, err = GetColumnTypes(table.desc, indexColumnIDs)
+		if err != nil {
+			return err
+		}
+		table.keyVals = make([]EncDatum, len(indexColumnIDs))
+
+		if hasExtraCols(&table) {
+			// Unique secondary indexes have a value that is the
+			// primary index key.
+			// Primary indexes only contain ascendingly-encoded
+			// values. If this ever changes, we'll probably have to
+			// figure out the directions here too.
+			table.extraTypes, err = GetColumnTypes(table.desc, table.index.ExtraColumnIDs)
+			table.extraVals = make([]EncDatum, len(table.index.ExtraColumnIDs))
+			if err != nil {
+				return err
+			}
+		}
+
+		// Keep track of the maximum keys per row to accommodate a
+		// limitHint when StartScan is invoked.
+		if keysPerRow := table.desc.KeysPerRow(table.index.ID); keysPerRow > mrf.maxKeysPerRow {
+			mrf.maxKeysPerRow = keysPerRow
+		}
+
+		mrf.tables = append(mrf.tables, table)
+	}
+
+	if len(tables) == 1 {
+		// If there is more than one table, currentTable will be
+		// updated every time NextKey is invoked and rowReadyTable
+		// will be updated when a row is fully decoded.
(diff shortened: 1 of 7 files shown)


richardwu:
So I've been griping about how one would check for which table a given decoded key belongs to. The above implementation currently iterates through each table passed into `MultIRowFetcher` and tries to decode it for that table (extracts `TableID` and `IndexID` and checks for equivalence).

This iteration happens for every key, which may be especially bad if there are many tables (more than a couple) passed into `MultiRowFetcher`). One idea I thought of is we pre-process each table's row prefix (including all ancestor prefixes) during `Init`. We then extract the entire key prefix (with only its ancestors' table IDs, index IDs, and interleave sentinels) once for each key (this logic would be similar to [this part of `DecodeIndexKey`] (https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/sqlbase/table.go#L749#L771)). We then lookup the table with said prefix, and continue decoding the rest of the `KeyVals`. 

In fact, I wrote something similar in a `[table.go] DecodeMultiIndexKey` function which would actually shortcut to the terminating ancestor in [this part of `DecodeIndexKey`](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/sqlbase/table.go#L754#L756): it would do a lookup for the table with the `ancestor.TableID` and `ancestor.IndexID` via a map). This accomplishes the shortcutting as described above but re-uses most of the code from `DecodeIndexKey`.

Another optimization point is the ordering of the tables in the `tables []MultiRowFetcherArgs` parameter of `MultiRowFetcher.Init`. That is given `tables = {<tableA>, <tableB>}` and `<tableB>` has significantly more keys in the scan than `<tableA>`, we will more often successfully decode a key on the 2nd iteration. On the contrary, if we ordered `tables = {<tableB>, <tableA>}`, then we successfully decode a key on the 1st iteration more often. Intuitively, children rows are often orders of magnitude more numerous than their parent rows, so this scenario is relevant where `<tableA> = parent` and `<tableB> = child`. Furthermore, we expect keys for a given row or even for a `child` table to be sequentially scanned, so perhaps it's better to cache the index to start the iteration of `tables` after each successful key decoding (e.g. if `<tableB>` key is decoded, we start the next key decoding with `<tableB>`). 

Let me know your thoughts on the current solution, bearing in mind that `len(tables) = 2` for this first implementation of interleaved tables.

RaduBerinde:
Hi Richard,

Sorry I didn't get to look at this yet, I will try to do it soon. In the meantime I'm CCing @danhhz who worked on the original interleaving code and encodings, he may be best to answer your questions above (I will think about it as well).

RaduBerinde:
Yes, checking the prefix is what I would go with. Technically all we need to do is to check against the longest prefix and figure out which table it is based on how much of it matched (we can precalculate how much of the prefix matches for each table).

E.g. if we have a hierarchy of three tables:

```
/table1/<k1>/<interleavedTableSentinel>/k2/2/<interleavedTableSentinel>/k3/
            ^first table matches to here    ^second table matches to here
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-Kw_XUYy5QOeXLwFRPgd:b3fqpq0)*

richardwu:
I introduced the concept of "equivalence prefixes", which basically defines the equivalence class for a table by its entire prefix.

For example, the hierarchy
```
parent
  child
     grandchild
```
will produce the following equivalence prefixes for each table
```
parent:          /<parent table id>/<parent index id>
child:             /<parent table id>/<parent index id>/#/<child table id>/<child index id> 
grandchild:   /<parent table id>/<parent index id>/#/<child table id>/<child index id>/#/<grandchild table id>/<grandchild index id>
```

Some an index from a `child` KV pair which might look something like
```
/50/1/20/#/51/1/5/0
```
with `parent_id = 20` and `id=5` will produce the equivalence prefix
```
/50/1/#/51/1
```

The lookup for the table with the equivalent prefix is currently done through slice iteration (this is more efficient than a hash map for `n=2`). We can later refactor this to hash on the equivalence prefix and use a `map` instead once we use this for a significant # of tables.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-KwaxrhUvCjUvJyjkCg9:b-fy1770)*

RaduBerinde:
The equivalence prefixes stuff makes sense to me. I haven't yet looked over the multi-row-fetcher code.

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/table.go, line 2213 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KweRCjOBE9ojxfED6X1:-KweRCjOBE9ojxfED6X2:b-chsyw9) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf5bea44a04efc02a4852dd0058d612a27593c5/pkg/sql/sqlbase/table.go#L2213)):*
> ```Go
> //	<child@secondary>
> // are represented as (in normal encoding format)
> //    <parent@primary>: /<parent table id>/<parent's primary index id>
> ```

This should explain that we are stripping out the values at each level. Maybe give an example that from the key
`/<parent table id>/<parent index id>/<val1>/<val2>/#/<child table id>/<child index id>/<val3>/<val4>`
we extract the prefix
`/<parent table id>/<parent index id>/#/<child table id>/<child index id>`

Also this is not quite a prefix, maybe we need a different name. Maybe "fingerprint"?

---

*[pkg/sql/sqlbase/table.go, line 2217 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KweRp8GiZh90Ny_L7v_:-KweRp8GiZh90Ny_L7va:b-wlu374) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf5bea44a04efc02a4852dd0058d612a27593c5/pkg/sql/sqlbase/table.go#L2217)):*
> ```Go
> // Equivalence prefixes allow us to associate an index key with its table
> // without having to invoke DecodeIndexKey multiple times.
> func ExtractIndexKeyEquivPrefix(key []byte) (prefix []byte, err error) {
> ```

Ideally we wouldn't allocate stuff for each row; the multi-row-fetcher could pass in a buffer that can be reused

---

*[pkg/sql/sqlbase/table.go, line 2257 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KweSBcO2eUz2-PpFKC2:-KweSBcO2eUz2-PpFKC3:bcn2nz5) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf5bea44a04efc02a4852dd0058d612a27593c5/pkg/sql/sqlbase/table.go#L2257)):*
> ```Go
> 			// We insert an interleave sentinel and continue extracting the
> 			// descendant's IDs.
> 			if _, err := out.Write(encoding.EncodeNotNullDescending(scratch[:0])); err != nil {
> ```

Why aren't we encoding directly into the key? We should use a `[]byte` instead of a `bytes.Buffer` and do `key = EncodeNotNullDesc(key)`

Also, NYC but it would be nice if we had a wrapper `sql.EncodeInterleaveSentinel` around `encoding.EncodeNotNullDescending` (the call is confusing when you read the code).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-KweTGm20Ry_66D3NjCR:b84929a)*

richardwu:
PTAL, cleaned up the `bytes.Buffer` stuff and scratch space.

---

---

*[pkg/sql/sqlbase/table.go, line 2213 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KweRCjOBE9ojxfED6X1:-KwgizDGuXvmSsKEe4D5:brayhit) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf5bea44a04efc02a4852dd0058d612a27593c5/pkg/sql/sqlbase/table.go#L2213)):*

This should explain that we are stripping out the values at each level. Maybe give an example that from the key
`/<parent table id>/<parent index id>/<val1>/<val2>/#/<child table id>/<child index id>/<val3>/<val4>`
we extract the prefix
`/<parent table id>/<parent index id>/#/<child table id>/<child index id>`

Also this is not quite a prefix, maybe we need a different name. Maybe "fingerprint"?
</blockquote></details>

Updated the function comment and I renamed it to "signature" if that's alright.

---

*[pkg/sql/sqlbase/table.go, line 2217 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KweRp8GiZh90Ny_L7v_:-KwgXuD5Di9YdHilmY_1:bxnmy90) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf5bea44a04efc02a4852dd0058d612a27593c5/pkg/sql/sqlbase/table.go#L2217)):*

Ideally we wouldn't allocate stuff for each row; the multi-row-fetcher could pass in a buffer that can be reused
</blockquote></details>

Oops, meant to refactor this. Scratch space isn't necessary now that we're not using `bytes.Buffer`. Good catch!

---

*[pkg/sql/sqlbase/table.go, line 2257 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KweSBcO2eUz2-PpFKC2:-KwgmDl_ioMp8EORrLql:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bf5bea44a04efc02a4852dd0058d612a27593c5/pkg/sql/sqlbase/table.go#L2257)):*

Why aren't we encoding directly into the key? We should use a `[]byte` instead of a `bytes.Buffer` and do `key = EncodeNotNullDesc(key)`

Also, NYC but it would be nice if we had a wrapper `sql.EncodeInterleaveSentinel` around `encoding.EncodeNotNullDescending` (the call is confusing when you read the code).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-Kwguaeopu1MtLYvWg72:b-odsmfi)*

richardwu:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/sqlbase/table.go, line 2250 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kwh-hc6I6ot8YPc5eRT:-Kwh-hc79W7OLUlXh0Ky:b9syosh) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b9c060a96fb92f2010727e60d61b380e6b06816/pkg/sql/sqlbase/table.go#L2250)):*
> ```Go
> 			}
> 			// Append column values within index key to return.
> 			colVals = append(colVals, key[:len]...)
> ```

I realized that we can in fact memoize all the column values we come across instead of having to iterate through `DecodeIndexKey` again. All bytes we come across are appended to a buffer that will be returned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228)*

richardwu:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/table.go, line 2255 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kwkp9j2zH_Pogtigw13:-Kwkp9j2zH_Pogtigw14:b56z9vp) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/table.go#L2255)):*
> ```Go
> 			// The current ancestor is not one of the specified ancestors.
> 			// We can short-circuit and return false.
> 			if validAncestorSignatures != nil && !validAncestorSignatures[string(signature)] {
> ```

I added another optimization when the signature for an index key is being decoded: if passed a set of ancestor signatures, it will short-circuit if the current ancestor signature __of the index key__ isn't in the set.

This allows us to practically short-circuit interleaved keys that are a deeper hierarchy level than the tables we're scanning for. For example
```
/parent
   /parent/#/child
      /parent/#/child/#/grandchild/#/.../(grand)*child
   /parent/#/child
```
If we are scanning for `parent` and `child` rows and we are checking for the signature of the index key for `grandgrandchild`, it will short-circuit at the `grandchild` prefix instead of continuing until it reaches the end of the key.

For many keys, this should help out a bit and actually aligns us with how [`DecodeIndexKey` does it](https://github.com/cockroachdb/cockroach/blob/542195d9c1ed2b4b506a0039b1b95cdb1bb4ab1a/pkg/sql/sqlbase/table.go#L763). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228)*

RaduBerinde:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 431 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kx8Y5Y8hampW9sZWVdU:-Kx8Y5Y8hampW9sZWVdV:bbgc1sr) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/multirowfetcher.go#L431)):*
> ```Go
> 	// TODO(richardwu): Once MultiRowFetcher is used for more than two
> 	// tables, this logic would be cleaner and perhaps more efficient if
> 	// a map is used instead of a slice for mrf.tables.
> ```

We are already looking it up in a map `allEquivSignatures`.. Why look it up again? I think we should just have that as a `map[string]int` where the int is the table id (`0` to `len(mrf.tables)-1`) and have `IndexKeyEquivSignature` return the table id.

---

*[pkg/sql/sqlbase/table.go, line 2220 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kx8X55ktalmeEknVu-2:-Kx8X55ktalmeEknVu-3:b-er8udo) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/table.go#L2220)):*
> ```Go
> // Equivalence signatures allow us to associate an index key with its table
> // without having to invoke DecodeIndexKey multiple times.
> // If validAncestorSignatures is not nil, then IndexKeyEquivSignature will
> ```

What's the use case for `validAncestorSignatures=nil`? Just testing?

---

*[pkg/sql/sqlbase/table.go, line 2235 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kx8WXTJbcR_fqyACj5m:-Kx8WXTKy3esQ0szdtIo:brgu537) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/table.go#L2235)):*
> ```Go
> ) (signature []byte, rest []byte, success bool, err error) {
> 	var isSentinel bool
> 	signature = make([]byte, 0, len(key))
> ```

We should try to avoid per-row allocations as much as possible. I think we can reuse the same buffers across rows. (Can be in a separate PR though if you prefer)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228)*

richardwu:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 431 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kx8Y5Y8hampW9sZWVdU:-Kx97PzJM82KuE3HZaJV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/multirowfetcher.go#L431)):*

We are already looking it up in a map `allEquivSignatures`.. Why look it up again? I think we should just have that as a `map[string]int` where the int is the table id (`0` to `len(mrf.tables)-1`) and have `IndexKeyEquivSignature` return the table id.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/table.go, line 2220 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kx8X55ktalmeEknVu-2:-Kx9Kl8X8lc7s3Dn013x:b-26rh17) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/table.go#L2220)):*

What's the use case for `validAncestorSignatures=nil`? Just testing?
</blockquote></details>

Yup it was originally for testing. In hind sight, this is not good practice and I've updated the tests to pass a valid `validEquivSignatures` (renamed it from `validAncestorSignatures` since it includes the key's table's signature too) map.

---

*[pkg/sql/sqlbase/table.go, line 2235 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-Kx8WXTJbcR_fqyACj5m:-Kx9L06AZeNxh62b6NpX:b-7ibr6h) ([raw file](https://github.com/cockroachdb/cockroach/blob/350d36b3987ff7e559130a10ab96fb1405e96938/pkg/sql/sqlbase/table.go#L2235)):*

We should try to avoid per-row allocations as much as possible. I think we can reuse the same buffers across rows. (Can be in a separate PR though if you prefer)
</blockquote></details>

Refactored this so `MultiRowFetcher` passes down a buffer. Good catch! 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228)*

RaduBerinde:
:100: :lgtm:   Great work!

---

Reviewed 1 of 4 files at r3, 4 of 4 files at r5.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 476 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxA8_ePDb9Ujvzw6a2O:-KxA8_ePDb9Ujvzw6a2P:b-d7ca8t) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b352b91c98bfe65ca145ced8e27aed5a946d6ad/pkg/sql/sqlbase/multirowfetcher.go#L476)):*
> ```Go
> 	}
> 
> 	tableIdx, found := mrf.allEquivSignatures[string(keyEquivSignature)]
> ```

this could be returned by IndeKeyEquivSignature so we don't look it up again.

---

*[pkg/sql/sqlbase/table.go, line 2256 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxA82kMoVIZBfus9ki8:-KxA82kMoVIZBfus9ki9:bjbltlj) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b352b91c98bfe65ca145ced8e27aed5a946d6ad/pkg/sql/sqlbase/table.go#L2256)):*
> ```Go
> func IndexKeyEquivSignature(
> 	key []byte, validEquivSignatures map[string]int, signature []byte, rest []byte,
> ) ([]byte, []byte, bool, error) {
> ```

[nit] these should have names (or at least documented in the comment)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-KxA96s5MjryHyuCX7DS:ba8tqvf)*

richardwu:
I'm going to work on a few "unit" tests for `MultiRowFetcher` and I'll ping with the updated revision.

---

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 476 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxA8_ePDb9Ujvzw6a2O:-KxANLp4N6ZCcp5HT0xq:b-fdpa7k) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b352b91c98bfe65ca145ced8e27aed5a946d6ad/pkg/sql/sqlbase/multirowfetcher.go#L476)):*

this could be returned by IndeKeyEquivSignature so we don't look it up again.
</blockquote></details>

Done. Now that we're returning the table index directly, it begs the question whether it's worthwhile to return the signature of the key itself (it's not being used in `MultiRowFetcher`). For completeness as an exported method in `sqlbase/table.go`, it would make sense for it to also return the signature. A criticism against this is only `MultiRowFetcher` calls this function, so maybe we should just move the `EquivSignature` methods to `MultiRowFetcher` (and un-export them).

Thoughts?

---

*[pkg/sql/sqlbase/table.go, line 2256 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxA82kMoVIZBfus9ki8:-KxANMqCKWBsO_x_r-Hs:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b352b91c98bfe65ca145ced8e27aed5a946d6ad/pkg/sql/sqlbase/table.go#L2256)):*

[nit] these should have names (or at least documented in the comment)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-KxAO2hOUk6N6H4zDnV7:b1p5mk8)*

RaduBerinde:
Review status: 2 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 476 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxA8_ePDb9Ujvzw6a2O:-KxDUTcrB06qnM5fcPbJ:b-qyr78z) ([raw file](https://github.com/cockroachdb/cockroach/blob/2b352b91c98bfe65ca145ced8e27aed5a946d6ad/pkg/sql/sqlbase/multirowfetcher.go#L476)):*

Done. Now that we're returning the table index directly, it begs the question whether it's worthwhile to return the signature of the key itself (it's not being used in `MultiRowFetcher`). For completeness as an exported method in `sqlbase/table.go`, it would make sense for it to also return the signature. A criticism against this is only `MultiRowFetcher` calls this function, so maybe we should just move the `EquivSignature` methods to `MultiRowFetcher` (and un-export them).

Thoughts?
</blockquote></details>

The way it works now, the function only returns something useful if the signature is in the map, so returning the signature is not very useful (the caller knows the signature based on the ID). I would keep it in `sqlbase`, we can always extend/generalize it later if we ever need to.

---

*[pkg/sql/sqlbase/table.go, line 2262 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxDTmXRKsV-G1M0-llk:-KxDTmXRKsV-G1M0-lll:b-bxb4ue) ([raw file](https://github.com/cockroachdb/cockroach/blob/b96f405bf08ed07e643dcd553940b9fb42768e6d/pkg/sql/sqlbase/table.go#L2262)):*
> ```Go
> 	key []byte, validEquivSignatures map[string]int, keySigBuf []byte, keyRestBuf []byte,
> ) (signature []byte, rest []byte, success bool, tableIdx int, rr error) {
> 	signature = signature[:0]
> ```

This now needs to be`= keySigBuf[:0]` and `= keyRestBuf[:0]`. You can also keep the old names and use other names for the return values (`signatureResult restResult`), and never use those names in the code.

Also [nit] but the "key" in these names is not helpful, maybe just `signatureBuf`, `restBuf`.

---

*[pkg/sql/sqlbase/table.go, line 2313 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxDUDvV44WvgtO96EDB:-KxDUDvV44WvgtO96EDC:b-ib4k84) ([raw file](https://github.com/cockroachdb/cockroach/blob/b96f405bf08ed07e643dcd553940b9fb42768e6d/pkg/sql/sqlbase/table.go#L2313)):*
> ```Go
> 		} else {
> 			// The key has been fully decomposed.
> 			break
> ```

[nit] clearer if we just do `if !isSentinel { return ... }`. Then `recentTableIdx` can be scoped to the inside of the loop (`recentTableIdx, found := ..`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228)*

richardwu:
Added some solid unit tests, namely for the following cases:
1. `RowFetcher` 1 table fetches with the primary index
2. Fetches from secondary indexes (unique/non-unique with NULL values)
3. A power set of fetches from interleaved tables. Specifically for the interleaved hierarchy:
   ```
   parent1
   parent2
      child1
         grandchild1
           grandgrandchild1
      child2
      grandgrandchild1@ggc1_unique_idx     <-- interleaved on parent2_id (p2)
   parent3
   ```
   there are `2^8 - 1 = 255` test cases to query every non-empty subset of tables.

One takeaway: `MergeSpans` is absolutely necessary during planning to de-duplicate the spans being fed into `MultiRowFetcher` in the case of reading from multiple interleaved tables.

Ready for another look!

RaduBerinde:
LGTM  Nice tests!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-:-KxmcYRgInA0O0fne8cq:b-e3ofee)*

richardwu:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/table.go, line 2262 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxDTmXRKsV-G1M0-llk:-KxnYAKmJQKeaJ1zCdRB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b96f405bf08ed07e643dcd553940b9fb42768e6d/pkg/sql/sqlbase/table.go#L2262)):*

This now needs to be`= keySigBuf[:0]` and `= keyRestBuf[:0]`. You can also keep the old names and use other names for the return values (`signatureResult restResult`), and never use those names in the code.

Also [nit] but the "key" in these names is not helpful, maybe just `signatureBuf`, `restBuf`.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/table.go, line 2313 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228#-KxDUDvV44WvgtO96EDB:-KxnYSGxV7A_Y7HnH40t:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b96f405bf08ed07e643dcd553940b9fb42768e6d/pkg/sql/sqlbase/table.go#L2313)):*

[nit] clearer if we just do `if !isSentinel { return ... }`. Then `recentTableIdx` can be scoped to the inside of the loop (`recentTableIdx, found := ..`).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19228)*

PR #19230, opened 2017-10-13
util/protoutil: fix marshaling/unmarshal of non-protoutil.Message protos

### LINKED ISSUE #19195: rpc: Attempting to marshal runtime.errorBody
Repeating this sequence on various commits near the tip of master:

- Start a one-node cluster with `cockroach start --insecure --store <unique-store-name>`.
- Load admin UI, which immediately reports "Connection to CockroachDB node lost."
- Network requests for static resources are served successfully.
- API requests for data return a 500 error code with the body `{error:"failed to marshal error message"}`.
- Log inspection shows the error `I171011 21:08:12.590952 330 vendor/google.golang.org/grpc/grpclog/grpclog.go:128  Failed to marshal error message "error:\"grpc: the connection is unavailable\" code:14 ": unexpected type *runtime.errorBody does not implement protoutil.Message`

This doesn't happen every time, but repeating this sequence a half dozen times or so, one of them does.

A similar issue hit me a few weeks ago, except:

- The status code was 503.
- The response body was a protobuf-encoded "grpc: the connection is unavailable"
- Restarting my computer seemed to fix the issue.

This time, restarting seems to have no effect.

Once I observed the issue resolve itself after waiting a few minutes.  I have been unable to reproduce that particular aspect.

petermattis:
grpc-gateway uses internal protos that implement proto.Message but not
protoutil.Message. The marshaling and unmarshaling done to support
grpc-gateway doesn't require the same hooks, so it is safe to just use
proto.{Message,Marshal,Unmarshal}.

Fixes #19195

### DIFF
--- a/build/style_test.go
+++ b/build/style_test.go
@@ -387,6 +387,7 @@ func TestStyle(t *testing.T) {
 			"--",
 			"*.go",
 			":!util/protoutil/marshal.go",
+			":!util/protoutil/marshaler.go",
 			":!settings/settings_test.go",
 		)
 		if err != nil {
@@ -425,6 +426,7 @@ func TestStyle(t *testing.T) {
 			"*.go",
 			":!*.pb.go",
 			":!util/protoutil/marshal.go",
+			":!util/protoutil/marshaler.go",
 		)
 		if err != nil {
 			t.Fatal(err)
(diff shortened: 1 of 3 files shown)


couchand:
LGTM, and in a dozen `cockroach start`s I didn't see the connection issue come up.

PR #19242, opened 2017-10-13
sql: don't run db.Txn when getting cached db descs

jordanlewis:
Previously, when fetching the database ID from the database cache
(something that happens once per SQL transaction), we would always run
`db.Txn` regardless of whether the database ID was already in the cache
or not. `db.Txn` is fairly heavy-weight and shouldn't be invoked unless
necessary.

Now, we pass a function that invokes `db.Txn` into the cache, permitting
it to only make a new transaction when the database descriptor isn't
present.

I noticed this because we were spending about 5% of the whole execStmt
flow running `db.Txn`s pointlessly, since the vast majority of the time
we have the database descriptor cached. That 5% was attributed to
planning, since the database and table resolution happens during
planning. After this change, planning takes up 35% instead of 40% of the
execStmt flow.

### DIFF
--- a/pkg/sql/database.go
+++ b/pkg/sql/database.go
@@ -244,7 +244,10 @@ func getAllDatabaseDescs(
 // getDatabaseDesc returns the database descriptor given its name
 // if it exists in the cache, otherwise falls back to KV operations.
 func (dc *databaseCache) getDatabaseDesc(
-	ctx context.Context, txn *client.Txn, vt VirtualTabler, name string,
+	ctx context.Context,
+	txnRunner func(context.Context, func(context.Context, *client.Txn) error) error,
+	vt VirtualTabler,
+	name string,
 ) (*sqlbase.DatabaseDescriptor, error) {
 	// Lookup the database in the cache first, falling back to the KV store if it
 	// isn't present. The cache might cause the usage of a recently renamed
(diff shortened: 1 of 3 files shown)


petermattis:
:lgtm:

---

---

*[pkg/sql/table.go, line 321 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19242#-KwLrbeAHN18jEVgCM-U:-KwLrbeAHN18jEVgCM-V:bigh1xi) ([raw file](https://github.com/cockroachdb/cockroach/blob/81f60b404002acd16bf9c1c7ea40e70c6aabea8f/pkg/sql/table.go#L321)):*
> ```Go
> 		// Resolve the database from the database cache when the transaction
> 		// hasn't modified the database.
> 		log.VEventf(ctx, 2, "dbID=0")
> ```

Did you intend to leave this event?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19242#-:-KwLrgzEZmKQcVNekZ2-:bnfp4nl)*

jordanlewis:

---

*[pkg/sql/table.go, line 321 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19242#-KwLrbeAHN18jEVgCM-U:-KwMM1uCbwJ_lVKSrHP-:bmm6v4s) ([raw file](https://github.com/cockroachdb/cockroach/blob/81f60b404002acd16bf9c1c7ea40e70c6aabea8f/pkg/sql/table.go#L321)):*

Did you intend to leave this event?
</blockquote></details>

No sir. Thanks for the catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19242)*

PR #19247, opened 2017-10-13
builder: cache builds with ccache

benesch:
Teach the builder to cache intermediate compilation artifacts on the
host machine. This *vastly* speeds up repeated compilation of the C
dependencies and should yield a nice reduction in build time on
TeamCity.

The cache is limited to 5GB by default.


tschottdorf:
Is this a bit scary because it'll also have our release builds use ccache (?)? As far as I understand ccache is pretty good at nuking the cache when it needs to, but it's not *perfect*?

That said, I'd be perfectly happy to use `ccache` whenever building a PR's tests (maybe not on `master` to avoid rot?)

benesch:
Release builds specify their compiler toolchain explicitly and so won't go
through ccache :D

tschottdorf:
Cool, that's good to hear. Maybe sprinkle an appropriate assertion somewhere to make sure we don't throw this under the bus accidentally in the future? The speedup from this change must be huge so I'm all for it.

tschottdorf:
btw, the CI failure tells me that you should rebalance.

benesch:
Yeah, totally hear you. I switched to making it opt-in: you'll need to specify `COCKROACH_BUILDER_CCACHE=1` to use . If it seems to be working well, we can look into enabling it automatically on developer machines. For now, though, this commit tells teamcity-* scripts turn it on when building PRs.

benesch:
Ok, CI is happy. PTAL!

tbg (approved):
The ccache car setting looks like it wants DRY, but don't let that hold you back if that's not trivial. Could you source a script or would that not work?
How much time does it end up saving, btw?

tschottdorf:
PS codacy is unhappy.

benesch:
Yeah, you're right that it's time to start DRYing these teamcity-*.sh scripts. Extracted a helper that's sourced as necessary.

I'm expecting this to save about 1.5m-2m on every build. I'm also planning to bake a warmed ccache into the TeamCity agent, so builds on a cold agent should no longer be several minutes slower.

tbg (approved):
Sorry, thought I'd already approved this! Make sure your builder tag isn't stale, though.

benesch:
👍 no worries! I'm waiting until I have the bandwidth to debug the inevitable fallout from this.

benesch:
Ok, it's finally time! In this goes. TFTR!

PR #19266, opened 2017-10-16
storage: avoid miscounting scan requests in TestStoreScanIntents

benesch:
Previously, in TestScoreScanIntents, if the race detector slowed things
down sufficiently, a periodic gossip would be triggered before the test
completed. This gossip would generate scan requests that were included
in the test's bookkeeping, causing spurious failures.

Filter out scan requests that aren't generated by the test by using a
unique prefix for the test's keys. Enabling the DisablePeriodicGossip
testing knob would have the same effect, but this approach is more
robust against future features that might generate background scan
requests.


tbg (approved):
Thanks 👍

PR #19268, opened 2017-10-16
storage: prevent DisableOnePhaseCommit from wedging a cluster

benesch:
If one phase commit is disabled, node liveness is unable to update any
livenesses. This effectively halts all transactions from proceeding once
the epoch-based lease system is bootstrapped. Since TestAsOfRetry
disables one phase commit, if it's run with a slow enough race detector,
epoch-based leases will be up and running before it completes and it
will hang forever.

Prevent the DisableOnePhaseCommit testing knob from applying to
BatchRequests that set the Require1PC flag, like node liveness. Tests
that use the knob still get the desired effect of preventing one phase
commits on the BatchRequests they generate, as those obviously do not
set the Require1PC flag.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -4830,7 +4830,7 @@ func (r *Replica) evaluateTxnWriteBatch(
 	ms := enginepb.MVCCStats{}
 	// If not transactional or there are indications that the batch's txn will
 	// require restart or retry, execute as normal.
-	if !r.store.TestingKnobs().DisableOnePhaseCommits && isOnePhaseCommit(ba) {
+	if isOnePhaseCommit(ba, r.store.TestingKnobs()) {
 		arg, _ := ba.GetArg(roachpb.EndTransaction)
 		etArg := arg.(*roachpb.EndTransactionRequest)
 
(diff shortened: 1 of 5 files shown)


pkg/storage/replica.go
```diff
@@ -4830,7 +4830,8 @@ func (r *Replica) evaluateTxnWriteBatch(
 	ms := enginepb.MVCCStats{}
 	// If not transactional or there are indications that the batch's txn will
 	// require restart or retry, execute as normal.
-	if !r.store.TestingKnobs().DisableOnePhaseCommits && isOnePhaseCommit(ba) {
+	allowed1PC, required1PC := onePhaseCommitStatus(ba)
```
tschottdorf: It's odd to have to think about the combination `(false, true)`. Perhaps pass `DisableOnePhaseCommits` into `onePhaseCommitStatus` and then return a single bool which indicates whether to use one phase commit? And consider renaming: `DisableOptionalOnePhaseCommits`.

tbg (approved):
Basically LGTM and thanks for fixing this, though I do have one comment that you can take or leave.

benesch:
Much better. PTAL.

tschottdorf:
Wonderful, thanks! :100: 

---

Reviewed 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19268#-:-Kw_ZmkNRHJkvVJbAOdO:b-o3js16)*

benesch:
You bet. TFTR!

PR #19272, opened 2017-10-16
cli: move temp dirs cleanup to beginning of cli.runStart and add log message for temp dir

richardwu:
This PR moves the cleanup function to the beginning of cli.runStart to
clear up abandoned temp directory early on during node startup in case
the disk is full.

During server startup, if a temporary directory is created for temp
storage, it will output the temporary directory path above store[0].

For example, the output for
```
cockroach start --insecure --store=mystore
```
where a temporary directory is nested under `./mystore` produces the
following output
```
CockroachDB node starting at 2017-10-18 23:12:57.761605387 +0000 UTC (took 0.6s)
build:      CCL v1.2-alpha.20170901-828-gf8c68c39c-dirty @ 2017/10/18 23:04:51 (go1.9)
admin:      http://Richards-MBP:8080
sql:        postgresql://root@Richards-MBP:26257?application_name=cockroach&sslmode=disable
logs:       /Users/richardwu/go/src/github.com/cockroachdb/cockroach/mystore/logs
temp dir:   /Users/richardwu/go/src/github.com/cockroachdb/cockroach/mystore/cockroach-temp993448186
store[0]:   path=/Users/richardwu/go/src/github.com/cockroachdb/cockroach/mystore
status:     restarted pre-existing node
clusterID:  1c439e43-c1da-40a3-a851-6f39fbc20ee9
nodeID:     1
```

If no temporary directory is created and temp storage is in memory,
there will be no output for temp dir.

This PR also refactors where the temporary directory is created: instead
of creating the temporary directory when `server.NewServer` is invoked,
a temporary directory is created during `cli.runStart`.


andreimatei:
The 2nd example you list in the commit msg is not ideal - we'd like the tmp dir to also be listed, but it isn't. Would that be hard to do? Can't we create the temp dir before we generate that message?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-:-KwaNrxgDB9ua4hOg_iP:bew64bn)*

richardwu:
I believe this was discussed in [the PR that introduced the temp directory flag](https://github.com/cockroachdb/cockroach/pull/18544): it would be the responsibility of `server`, right before initializing the `TempEngine`, to create the temporary directory and not have `config` or `cli` create any temporary directories.

I guess the real question is what do we want to surface: the parent directory the temporary directory will be created in (what's passed into `--temp-dir`), the actual temporary directory, or perhaps even both?

The first option: maybe we should rename the output tag as `temp root dir` or `temp parent dir` instead of `temp dir` (current).

The second option: we could have an additional output after the temporary directory is initialized. This output would be after the CLI output which makes sense.

Thoughts?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-:-KwaONusVsZk3Ta25lDA:b-wj3y65)*

andreimatei:
If I remember correctly, my stance on the creation of the dir was that it shouldn't be done by a `Config` (and I stand by that); beyond that, I don't know if I've ever had an opinion on who should do it. It has bothered me that `TempStorageConfig` has both a `Path` and a `ParentDir` but I didn't want to think about what to do and so I let it go at the time.
It'd seem fine to me for `cli` (and tests and whoever else creates a TempStorageConfig`) to create the temp dir (and then pass the responsibility to delete it to the `Server`).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-:-KwaTx-cM-pS02MdZURC:bnqn7o6)*

richardwu:
I've removed `ParentDir` and instead introduced an ephemeral `tempParentDir` variable in the `cli` flag. It is used to initialize `cli.serverCfg` with the path of a temporary directory it creates in `initTempStorageConfig` during `cli.runStart`.

`cli` is now responsible of creating this temporary directory and passing it on to server, which has the responsibility of removing this directory.

Ready for another review!

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-:-Kwln1ZwF6AJ1xjJUKmK:b59i65t)*

andreimatei:
Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/base/config.go, line 429 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpY5F0tGP5LeNyy4JN:-KwpY5F0tGP5LeNyy4JO:bku9qjj) ([raw file](https://github.com/cockroachdb/cockroach/blob/e3956adb972f9ee81a4153c4cb509a3ececea964/pkg/base/config.go#L429)):*
> ```Go
> // parentDir will be defaulted to the store's path.
> func TempStorageConfigFromEnv(
> 	ctx context.Context, firstStore StoreSpec, parentDir string, maxSizeBytes int64,
> ```

parentDir doesn't really seem to be used in this function. What's the deal?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272)*

richardwu:
Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/base/config.go, line 429 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpY5F0tGP5LeNyy4JN:-KwpZYoenzVIcqiH4vxI:bf5hs1n) ([raw file](https://github.com/cockroachdb/cockroach/blob/e3956adb972f9ee81a4153c4cb509a3ececea964/pkg/base/config.go#L429)):*

parentDir doesn't really seem to be used in this function. What's the deal?
</blockquote></details>

It's used for deciding whether the temp storage will be in memory or not.

Alternatively, we could let the client of this function pass in `inMem bool`, but I think that logic should remain in this function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272)*

andreimatei:
LGTM but the new util file could benefit from unit tests

---

Review status: 0 of 10 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/cli/flags.go, line 43 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpcvxKBVDKESK-zClW:-KwpcvxKBVDKESK-zClX:bi50kx4) ([raw file](https://github.com/cockroachdb/cockroach/blob/328d18c76fba938686709d52ce5dc4c097ec1a97/pkg/cli/flags.go#L43)):*
> ```Go
> var serverHTTPHost, serverHTTPPort string
> var clientConnHost, clientConnPort string
> var tempParentDir string
> ```

nit: tempDir, to match the flag

---

*[pkg/cli/start.go, line 453 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpcLoLyV3sMCVR7NmM:-KwpcLoLyV3sMCVR7NmN:bm0wn7g) ([raw file](https://github.com/cockroachdb/cockroach/blob/328d18c76fba938686709d52ce5dc4c097ec1a97/pkg/cli/start.go#L453)):*
> ```Go
> 	// We need to clean up abandoned temporary directories from the
> 	// temporary directory record file before creating any new temporary
> 	// directories.  Record file can only exist if the first store is
> ```

double space

---

*[pkg/cli/start.go, line 546 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpdRlLYFYfz60W5Ope:-KwpdRlLYFYfz60W5Opf:b-dh4uim) ([raw file](https://github.com/cockroachdb/cockroach/blob/328d18c76fba938686709d52ce5dc4c097ec1a97/pkg/cli/start.go#L546)):*
> ```Go
> 		if err := func() error {
> 			var err error
> 			if err = serverCfg.InitNode(); err != nil {
> ```

why did you expand the scope of this `err`?
Why couldn't `var err error` stay where it was?

---

*[pkg/cli/start.go, line 405 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9HYLlZrg2CjHpFiRR:-Kx9HYLlZrg2CjHpFiRS:b-5nalj5) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L405)):*
> ```Go
> 
> func initTempStorageConfig(ctx context.Context) error {
> 	firstStore := serverCfg.Stores.Specs[0]
> ```

please pass `serverCfg` (or some part of it) as a param to this function

---

*[pkg/cli/start.go, line 407 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9Hl4AmCiNIzbwLNva:-Kx9Hl4AmCiNIzbwLNvb:ba789qs) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L407)):*
> ```Go
> 	firstStore := serverCfg.Stores.Specs[0]
> 	var recordPath string
> 	if firstStore.Path != "" {
> ```

nit: use `InMemory`. Besides being more readable, you use that below.

---

*[pkg/cli/start.go, line 450 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9I1XRVnZgDnVWBT8F:-Kx9I1XRVnZgDnVWBT8G:buzw36) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L450)):*
> ```Go
> 	}
> 
> 	// Re-initialize TempStorageConfig based on cli flag-initialize values
> ```

flag - initialize

---

*[pkg/cli/start.go, line 459 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9IJioT1X8iMdmkk9N:-Kx9IJioT1X8iMdmkk9O:b-rqijyq) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L459)):*
> ```Go
> 	)
> 
> 	// Default temp parent directory to first store path if the temp
> ```

nit: the use of "default" here is wrong - you're _setting_ `tempParentDir` to something; it will not be overwritten further.

---

*[pkg/server/server.go, line 306 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9H354Vl_L6-zCorgS:-Kx9H354Vl_L6-zCorgT:b-fdiic6) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/server/server.go#L306)):*
> ```Go
> 		firstStore := cfg.Stores.Specs[0]
> 		var err error
> 		if firstStore.Path == "" {
> ```

was there a reason to change this line? I find the old condition clearer.

---

*[pkg/util/temp_dir.go, line 52 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9FBdJ_Jf5Xx-Q7Eca:-Kx9FBdJ_Jf5Xx-Q7Ecb:bvu3ju4) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/util/temp_dir.go#L52)):*
> ```Go
> 		return err
> 	}
> 	return f.Sync()
> ```

do you need `Sync` given that you `Close`?

---

*[pkg/util/temp_dir.go, line 60 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9FKy0mDLVSBqmxeo_:-Kx9FKy0mDLVSBqmxeoa:bkilpgy) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/util/temp_dir.go#L60)):*
> ```Go
> // up abandoned temporary directories.
> // It should also be invoked when a newly created temporary directory is no
> // longer needed and needs to be removed and removed from the record file.
> ```

removed and removed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-:-Kx9IqRBGR3iHlryBx7P:b4l1o7s)*

richardwu:
Added some tests!

---

Review status: 0 of 11 files reviewed at latest revision, 11 unresolved discussions.

---

*[pkg/cli/flags.go, line 43 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpcvxKBVDKESK-zClW:-Kx9jikqUVMAvDKgr2KR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/328d18c76fba938686709d52ce5dc4c097ec1a97/pkg/cli/flags.go#L43)):*

nit: tempDir, to match the flag
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 453 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpcLoLyV3sMCVR7NmM:-Kx9m9LL4ZvXTIp5fGHE:bs7wup0) ([raw file](https://github.com/cockroachdb/cockroach/blob/328d18c76fba938686709d52ce5dc4c097ec1a97/pkg/cli/start.go#L453)):*

double space
</blockquote></details>

Done.k

---

*[pkg/cli/start.go, line 546 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-KwpdRlLYFYfz60W5Ope:-Kx9npIJpg81AEXJgetU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/328d18c76fba938686709d52ce5dc4c097ec1a97/pkg/cli/start.go#L546)):*

why did you expand the scope of this `err`?
Why couldn't `var err error` stay where it was?
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 405 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9HYLlZrg2CjHpFiRR:-Kx9lP24w69hPHLqH39K:b-s5h0e1) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L405)):*

please pass `serverCfg` (or some part of it) as a param to this function
</blockquote></details>

Good point: I made `initTempStorageConfig` take in the first store's spec as a parameter return the new `TempStorageConfig` and assigned it in `runStart` to `serverCfg.TempStorageConfig`. Much cleaner IMO.

---

*[pkg/cli/start.go, line 407 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9Hl4AmCiNIzbwLNva:-Kx9lk2OxJo4hQ7Mqptk:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L407)):*

nit: use `InMemory`. Besides being more readable, you use that below.
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 450 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9I1XRVnZgDnVWBT8F:-Kx9lkJFKQaSn2bjXn_8:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L450)):*

flag - initialize
</blockquote></details>

Done.

---

*[pkg/cli/start.go, line 459 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9IJioT1X8iMdmkk9N:-Kx9lrB8dwq5bOukfnw6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/cli/start.go#L459)):*

nit: the use of "default" here is wrong - you're _setting_ `tempParentDir` to something; it will not be overwritten further.
</blockquote></details>

Done.

---

*[pkg/server/server.go, line 306 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9H354Vl_L6-zCorgS:-Kx9mpn5V7k_qfSMt0e2:btf4wza) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/server/server.go#L306)):*

was there a reason to change this line? I find the old condition clearer.
</blockquote></details>

Changed this back.

---

*[pkg/util/temp_dir.go, line 52 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9FBdJ_Jf5Xx-Q7Eca:-Kx9nHYBEIB64ZvMUoPX:b-c3tnjw) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/util/temp_dir.go#L52)):*

do you need `Sync` given that you `Close`?
</blockquote></details>

Good point. 

---

*[pkg/util/temp_dir.go, line 60 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-Kx9FKy0mDLVSBqmxeo_:-Kx9nOeEY6BTOVfbQbPb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/20194bd978a257089a3d89f94eaf150b3a71fc33/pkg/util/temp_dir.go#L60)):*

removed and removed
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19272#-:-Kx9ziLSI_GcMbq2aw9-:b51dhyf)*

PR #19279, opened 2017-10-16
storage: less clobbering and better readability in NodeLiveness

tbg:
See #18219 for context. Needs testing, but I wanted to get this out for review as
I'm hoping that it can go into 1.1.2. or 1.1.3.

----

Investigated the decommissioning problems some users have faced, which are essentially
that sometimes `NodeLiveness` gets stuck with a liveness record that doesn't have
`Decommissioning == true` even though `./cockroach debug gossip-values` indicates that
Gossip did receive the correct record. This appears to be happening mostly (always?)
when the target node is dead.

Unfortunately, I couldn't come up with a conclusive scenario since the users presumably
initiated these commands long after the target node was dead (and so `IncrementEpoch`
*should* not be directed at this node, though perhaps it is?) but I decided to simplify
and harden the code against some of the known issues. Precisely,

1. Update the stored liveness when setting the decommissioning status. This alone would
   prevent the decommissioning command from getting stuck (since the wrong state would be
   expunged).
2. Pass all code which updates any cached liveness through a method which decides whether
   to update, namely when the epoch or expiration are newer (or, if neither has
   changed, update if there are any other changes). This should do the "right
   thing" for Epoch/Expiration, but may clobber for `Draining/Decommissioning`,
   though that problem should have gotten less frequent in this change since we
   previously didn't use this criterion for more than breaking ties.
3. Remove `(NodeLiveness).mu.self`. It was odd to have both that and an entry in `mu.nodes`,
   and the original motivation (avoiding late gossip updates to "wind back" the
   state) does not seem to apply any more. This change is logically disconnected from the
   initial issue, but clarifying the code to make it easier to reason about seemed like a
   clear win.

I'm mildly nervous that something here might introduce a silly bug, but if it did and we
don't discover it in review, it's a signal that we should introduce a sequence number as
described in #18219.

### DIFF
--- a/pkg/storage/node_liveness.go
+++ b/pkg/storage/node_liveness.go
@@ -125,7 +125,6 @@ type NodeLiveness struct {
 
 	mu struct {
 		syncutil.Mutex
-		self              Liveness
 		callbacks         []IsLiveCallback
 		nodes             map[roachpb.NodeID]Liveness
 		heartbeatCallback HeartbeatCallback
@@ -260,7 +259,7 @@ func (nl *NodeLiveness) setDrainingInternal(
 	}
 	newLiveness.Draining = drain
 	if err := nl.updateLiveness(ctx, &newLiveness, liveness, func(actual Liveness) error {
-		nl.setSelf(actual)
+		nl.maybeUpdate(actual)
 		if actual.Draining == newLiveness.Draining {
 			return errNodeDrainingSet
 		}
@@ -271,7 +270,7 @@ func (nl *NodeLiveness) setDrainingInternal(
 		}
 		return err
 	}
-	nl.setSelf(newLiveness)
+	nl.maybeUpdate(newLiveness)
 	return nil
 }
 
@@ -455,7 +454,7 @@ func (nl *NodeLiveness) heartbeatInternal(
 	}
 	if err := nl.updateLiveness(ctx, &newLiveness, liveness, func(actual Liveness) error {
 		// Update liveness to actual value on mismatch.
-		nl.setSelf(actual)
+		nl.maybeUpdate(actual)
 		// If the actual liveness is different than expected, but is
 		// considered live, treat the heartbeat as a success. This can
 		// happen when the periodic heartbeater races with a concurrent
@@ -475,7 +474,7 @@ func (nl *NodeLiveness) heartbeatInternal(
 	}
 
 	log.VEventf(ctx, 1, "heartbeat %+v", newLiveness.Expiration)
-	nl.setSelf(newLiveness)
+	nl.maybeUpdate(newLiveness)
 	nl.metrics.HeartbeatSuccesses.Inc(1)
 	return nil
 }
@@ -490,19 +489,6 @@ func (nl *NodeLiveness) Self() (*Liveness, error) {
 	return nl.getLivenessLocked(nl.gossip.NodeID.Get())
 }
 
-func (nl *NodeLiveness) setSelf(liveness Liveness) {
-	nl.mu.Lock()
-	// NB: shouldReplaceLiveness should always be true
-	// since setSelf is only ever called for non-overlapping
-	// liveness updated serialized through a single semaphore.
-	//
-	// We call it for symmetry with the nodes map.
-	if shouldReplaceLiveness(nl.mu.self, liveness) {
-		nl.mu.self = liveness
-	}
-	nl.mu.Unlock()
-}
-
 // GetIsLiveMap returns a map of nodeID to boolean liveness status of
 // each node.
 func (nl *NodeLiveness) GetIsLiveMap() map[roachpb.NodeID]bool {
@@ -512,11 +498,7 @@ func (nl *NodeLiveness) GetIsLiveMap() map[roachpb.NodeID]bool {
 	now := nl.clock.Now()
 	maxOffset := nl.clock.MaxOffset()
 	for nID, l := range nl.mu.nodes {
-		if nID == nl.mu.self.NodeID {
-			lMap[nID] = nl.mu.self.IsLive(now, maxOffset)
-		} else {
-			lMap[nID] = l.IsLive(now, maxOffset)
-		}
+		lMap[nID] = l.IsLive(now, maxOffset)
 	}
 	return lMap
 }
@@ -528,11 +510,7 @@ func (nl *NodeLiveness) GetLivenesses() []Liveness {
 	defer nl.mu.Unlock()
 	livenesses := make([]Liveness, 0, len(nl.mu.nodes))
 	for _, l := range nl.mu.nodes {
-		if l.NodeID == nl.mu.self.NodeID {
-			livenesses = append(livenesses, nl.mu.self)
-		} else {
-			livenesses = append(livenesses, l)
-		}
+		livenesses = append(livenesses, l)
 	}
 	return livenesses
 }
@@ -547,10 +525,6 @@ func (nl *NodeLiveness) GetLiveness(nodeID roachpb.NodeID) (*Liveness, error) {
 }
 
 func (nl *NodeLiveness) getLivenessLocked(nodeID roachpb.NodeID) (*Liveness, error) {
-	if nodeID != 0 && nodeID == nl.mu.self.NodeID {
-		copySelf := nl.mu.self
-		return &copySelf, nil
-	}
 	if l, ok := nl.mu.nodes[nodeID]; ok {
 		return &l, nil
 	}
@@ -583,22 +557,8 @@ func (nl *NodeLiveness) IncrementEpoch(ctx context.Context, liveness *Liveness) 
 	}
 	newLiveness := *liveness
 	newLiveness.Epoch++
-	maybeUpdate := func(l Liveness) {
-		nl.mu.Lock()
-		defer nl.mu.Unlock()
-
-		if nodeID := nl.gossip.NodeID.Get(); nodeID == l.NodeID {
-			if shouldReplaceLiveness(nl.mu.self, l) {
-				nl.mu.self = l
-			}
-		} else {
-			if shouldReplaceLiveness(nl.mu.nodes[l.NodeID], l) {
-				nl.mu.nodes[l.NodeID] = l
-			}
-		}
-	}
 	if err := nl.updateLiveness(ctx, &newLiveness, liveness, func(actual Liveness) error {
-		defer maybeUpdate(actual)
+		defer nl.maybeUpdate(actual)
 		if actual.Epoch > liveness.Epoch {
 			return errEpochAlreadyIncremented
 		} else if actual.Epoch < liveness.Epoch {
@@ -614,7 +574,7 @@ func (nl *NodeLiveness) IncrementEpoch(ctx context.Context, liveness *Liveness) 
 
 	log.VEventf(ctx, 1, "incremented node %d liveness epoch to %d",
 		newLiveness.NodeID, newLiveness.Epoch)
-	maybeUpdate(newLiveness)
+	nl.maybeUpdate(newLiveness)
 	nl.metrics.EpochIncrements.Inc(1)
 	return nil
 }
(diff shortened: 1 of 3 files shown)


bdarnell:
:lgtm: 

The fourth commit is more refactoring than I'd like to do in a patch release. OTOH, if we don't backport it then we'll be left with divergence between 1.1.2 and master (and we'll have to make sure we test the 1.1.2 version without the backport before it is released)

---

---

*[pkg/storage/node_liveness.go, line 732 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwfYdJCteUk_7gxJama:-KwfYdJCteUk_7gxJamb:b-f3wxpd) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4bc322352cf76e2792ca2c785c3cb10939afbe2/pkg/storage/node_liveness.go#L732)):*
> ```Go
> 		// decommissioning field changed.
> 		//
> 		// This has false positives (in which case we're clobbering the
> ```

Doesn't it also have false negatives when the draining/decomissioning flag is updated at the same time as expiration or epoch?

---

*[pkg/storage/node_liveness.go, line 779 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-Kwf_-yZKJT0yWI0nFAV:-Kwf_-yZKJT0yWI0nFAW:bh2wdy4) ([raw file](https://github.com/cockroachdb/cockroach/blob/3285ec8f47f3a9efdfd04e1faf078c926720f15c/pkg/storage/node_liveness.go#L779)):*
> ```Go
> 	}
> 	if err != nil {
> 		log.Warningf(ctx, "looking up own liveness: %s", err)
> ```

Return here instead of continuing. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-:-Kwf_VjW5miTxJmUmdac:b-h5gq1i)*

a-robinson:
:lgtm:

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4.

---

*[pkg/storage/node_liveness.go, line 741 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-Kwewg9cwdfY_JbUJX2p:-Kwewg9cwdfY_JbUJX2q:btl0i65) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4bc322352cf76e2792ca2c785c3cb10939afbe2/pkg/storage/node_liveness.go#L741)):*
> ```Go
> 
> 	// Accept an update if it moves either expiration or epoch forward.
> 	return old.Expiration.Less(new.Expiration) || old.Epoch < new.Epoch
> ```

The serialization of writes to the KV layer might prevent this from ever being an issue, but what happens if an update increases one of these and decreases the other? I'd feel a little safer if the logic was something like `return (old.Expiration.Less(new.Expiration) && old.Epoch <= new.Epoch) || (old.Epoch < new.Epoch && (old.Expiration.Less(new.Expiration) || old.Expiration.Equals(new.Expiration)))`, or if it consistently chose one type of update to prefer in the conflicting case -- e.g. prefer the version with the higher epoch over the one with the higher expiration

---

*[pkg/storage/node_liveness.go, line 774 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwfdK3NwBZomzWxZjSY:-KwfdK3NwBZomzWxZjSZ:byf1jl0) ([raw file](https://github.com/cockroachdb/cockroach/blob/3285ec8f47f3a9efdfd04e1faf078c926720f15c/pkg/storage/node_liveness.go#L774)):*
> ```Go
> 	}
> 
> 	self, err := nl.Self()
> ```

Minor, but we could avoid an extra lock-unlock if we just took the lock up here and called `nl.getLivenessLocked(selfID)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-:-KwfdSP_vBpWGnabXfcj:bnfp4nl)*

nvanbenschoten:
:lgtm:

---

---

*[pkg/storage/node_liveness.go, line 497 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwgLLaRUMHIE00OA09v:-KwgLLaRUMHIE00OA09w:b-7fe8m1) ([raw file](https://github.com/cockroachdb/cockroach/blob/22147f9d81f0f126a014afaf39d2b1d08b6882de/pkg/storage/node_liveness.go#L497)):*
> ```Go
> 	// NB: shouldReplaceLiveness should always be true
> 	// since setSelf is only ever called for non-overlapping
> 	// liveness updated serialized through a single semaphore.
> ```

If it should always be true, should we panic if it's not instead of silently ignoring the probable logic error?

EDIT: never mind. If we keep commit 4 then this all goes away.

---

*[pkg/storage/node_liveness.go, line 774 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwfdK3NwBZomzWxZjSY:-KwgM66JY7m6pWXwN7zO:bqf3ww) ([raw file](https://github.com/cockroachdb/cockroach/blob/3285ec8f47f3a9efdfd04e1faf078c926720f15c/pkg/storage/node_liveness.go#L774)):*

Minor, but we could avoid an extra lock-unlock if we just took the lock up here and called `nl.getLivenessLocked(selfID)`
</blockquote></details>

I wouldn't say minor, since your suggestion reduces the chance of weird races here. A single critical section would also improve clarity.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-:-KwgMuM3gVwMo8dW2u_-:bnfp4nl)*

tschottdorf:
@bdarnell there's always the option of withholding the last commit, but I'd prefer to have this exercised rather sooner than later (we could also omit it in 1.1 initially if we remember to add it in a later patch, or when conflicts arise). Your call.

Added a unit test, will merge later today if things look good.

---

Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/storage/node_liveness.go, line 732 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwfYdJCteUk_7gxJama:-Kwg3m_7hDOOgIhpIVSr:b-i8k9fu) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4bc322352cf76e2792ca2c785c3cb10939afbe2/pkg/storage/node_liveness.go#L732)):*

Doesn't it also have false negatives when the draining/decomissioning flag is updated at the same time as expiration or epoch?
</blockquote></details>

Not sure I understand, if both are updated at the same time, the decommissioning/draining update doesn't matter because we update anyway. Are you worried about something like setting decommissioning and immediately after bumping the epoch, and then seeing the gossip updates out of order? That would be fine too, because the epoch bump result would also have decommissioning set (since everything gets serialized through CPuts).

---

*[pkg/storage/node_liveness.go, line 741 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-Kwewg9cwdfY_JbUJX2p:-Kwg6QyPsB6_Y3guE2Ba:b-ame6zm) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4bc322352cf76e2792ca2c785c3cb10939afbe2/pkg/storage/node_liveness.go#L741)):*

The serialization of writes to the KV layer might prevent this from ever being an issue, but what happens if an update increases one of these and decreases the other? I'd feel a little safer if the logic was something like `return (old.Expiration.Less(new.Expiration) && old.Epoch <= new.Epoch) || (old.Epoch < new.Epoch && (old.Expiration.Less(new.Expiration) || old.Expiration.Equals(new.Expiration)))`, or if it consistently chose one type of update to prefer in the conflicting case -- e.g. prefer the version with the higher epoch over the one with the higher expiration
</blockquote></details>

That's a good point. In practice this shouldn't happen (Epoch never goes down, and Expiration, too, though I'm less convinced there if we're in clockless mode and things are *really* set up to fail). Preferring the one with higher Epoch now.

---

*[pkg/storage/node_liveness.go, line 774 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwfdK3NwBZomzWxZjSY:-KwiXseRpbZeOdxui-uh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3285ec8f47f3a9efdfd04e1faf078c926720f15c/pkg/storage/node_liveness.go#L774)):*

I wouldn't say minor, since your suggestion reduces the chance of weird races here. A single critical section would also improve clarity.
</blockquote></details>

Done.

---

*[pkg/storage/node_liveness.go, line 779 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-Kwf_-yZKJT0yWI0nFAV:-KwiXy0SIw7WCztSGjNg:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3285ec8f47f3a9efdfd04e1faf078c926720f15c/pkg/storage/node_liveness.go#L779)):*

Return here instead of continuing. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-:-Kwg3Vwqr0Yasvjyl1A6:b-avgksq)*

a-robinson:
:lgtm:

---

Reviewed 1 of 2 files at r5, 1 of 2 files at r8, 1 of 2 files at r9.
Review status: 2 of 3 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/storage/node_liveness.go, line 792 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwjXTRAIsO3hVjI_UKY:-KwjXTRAIsO3hVjI_UKZ:b-1f1zge) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0c63834d888e2ecb0ec8e1b736721680737d942/pkg/storage/node_liveness.go#L792)):*
> ```Go
> 		return 0
> 	}
> 	if err != nil {
> ```

Why check the error *after* using the `self` value? It looks like a mistake

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-:-KwjXcQvDVVTr6YdTaM2:bnfp4nl)*

tschottdorf:
Review status: 2 of 3 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/storage/node_liveness.go, line 792 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwjXTRAIsO3hVjI_UKY:-Kwk6h9Zw5ZdpQzZmLtQ:b1kqolf) ([raw file](https://github.com/cockroachdb/cockroach/blob/e0c63834d888e2ecb0ec8e1b736721680737d942/pkg/storage/node_liveness.go#L792)):*

Why check the error *after* using the `self` value? It looks like a mistake
</blockquote></details>

Definitely a mistake, at least one of the two of us is awake.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279)*

bdarnell:
I'm fine including all the commits in 1.1.2.

---

Review status: 1 of 3 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/storage/node_liveness.go, line 732 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-KwfYdJCteUk_7gxJama:-Kwkkxy9p_mwfEe1lqW0:b-eafitl) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4bc322352cf76e2792ca2c785c3cb10939afbe2/pkg/storage/node_liveness.go#L732)):*

Not sure I understand, if both are updated at the same time, the decommissioning/draining update doesn't matter because we update anyway. Are you worried about something like setting decommissioning and immediately after bumping the epoch, and then seeing the gossip updates out of order? That would be fine too, because the epoch bump result would also have decommissioning set (since everything gets serialized through CPuts).
</blockquote></details>

Yeah, I was worried about out-of-order gossip updates. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19279#-:-Kwkl4ka7c9Y1Wwrceft:bskhxas)*

tschottdorf:
> I'm fine including all the commits in 1.1.2.

Ack, thanks. I'll wait for a few days since I introduced some flakiness that I just fixed.

PR #19127, opened 2017-10-09
sql: new built-in generator pg_get_keywords

knz:
Requested in https://github.com/cockroachdb/cockroach/issues/10291#issuecomment-334985732.

This PR includes some other useful refactors and cleanups.

PR #19280, opened 2017-10-16
cherrypick-1.1: sql/parser: lowercase keyword arguments upfront

knz:
Performance pick from #19127.

cc @cockroachdb/release


knz:
Huh, I didn't understand that. I'm ok either way.

jordanlewis:
cc @cockroachdb/release - any guidance?

bdarnell:
We generally do not cherry-pick performance improvements into a release. We sometimes make exceptions for especially severe performance problems that can effectively cause outages (for example, 1.0.3 included a number of changes to garbage collection that could be called performance improvements but were motivated by the fact that this was causing outages for a customer)

In this case, the keyword names are public (via the `pg_get_keywords()` function), so it's possible people are relying on this detail and we shouldn't break it in a patch release)

knz:
This patch is fully invisible for users -- the pg_get_keywords is not present in 1.1 and this cherry-pick doesn't change that.

Issue #19202, opened 2017-10-12
stability: networking layer should budget file descriptors


a-robinson:
Version: CockroachDB CCL v1.1-beta.20170921 (linux amd64, built 2017/09/21 15:48:53, go1.8.3)

When https://github.com/cockroachdb/cockroach/issues/19171 (a file descriptor leak) was first encountered, the symptoms were that the process was using much more CPU than normal (200%), hadn't written out any logs for a day, and was completely unresponsive to all network requests on both port. There was no clue anywhere about what was wrong.

It took `strace`-ing to figure out what was actually happening, but users shouldn't be expected to figure this out to understand what's going on with their process if they somehow manage to get into this state.

Judging by the stack trace @mjibson copied into #19171, it's likely that most goroutines were blocked on trying to log, meaning that the goroutine with the logging mutex was presumably blocked trying to open a new log file. I believe @mjibson could provide the rest of the dump if necessary.

Some other sort of behavior would be preferable here, if possible. We obviously don't have a ton of options, but may be able to log to an existing log file that we have open if we fail to open a new one, or if worse comes to worse just crash. @mberhault, do you have any thoughts?

mberhault:
Ever-increase open FDs is the kind of thing we can't do too much about (other than fixing the bugs causing them).
That said, we should at the very least monitor it (`sys_fd_open` vs `sys_fd_softlimit`) and alert in prometheus. AdminUI/Log warnings when approaching the limit might be a good idea too.

mberhault:
I added a testing alert at 80% used.

bdarnell:
> it's likely that most goroutines were blocked on trying to log, meaning that the goroutine with the logging mutex was presumably blocked trying to open a new log file.

It doesn't retry opening a log file, does it? If it fails to open a log file it should probably just panic instead of retry (and certainly not retry while holding a singleton lock). 

> Ever-increase open FDs is the kind of thing we can't do too much about (other than fixing the bugs causing them).

There's no perfect solution, but giving each component of the system an FD budget and attempting to stay within it can help. We already do this for rocksdb. It's harder to do this for network connections than for files on disk, but we can stop accepting new connections on one or both ports when FD usage is above a threshold (or accept new connections and close old (ideally idle) ones, depending on what we want to prioritize).

a-robinson:
> It doesn't retry opening a log file, does it? If it fails to open a log file it should probably just panic instead of retry (and certainly not retry while holding a singleton lock).

[...]

PR #19287, opened 2017-10-16
cherrypick-1.1: util/log: don't panic

benesch:
Cherry-pick of #17871. Fixes, most likely, #19202.

cc @cockroachdb/release

----

Previously, log.outputLogEntry could panic while holding the log mutex.
This would deadlock any goroutine that logged while recovering from the
panic, which is approximately all of the recover routines. Most
annoyingly, the crash reporter would deadlock, swallowing the cause of
the panic.

Avoid panicking while holding the log mutex and use l.exit instead,
which exists for this very purpose. In the process, enforce the
invariant that l.mu is held when l.exit is called. (The previous
behavior was, in fact, incorrect, as l.flushAll should not be called
without holding l.mu.)

Also add a Tcl test to ensure this doesn't break in the future.

### DIFF
--- a/pkg/cli/interactive_tests/test_missing_log_output.tcl
+++ b/pkg/cli/interactive_tests/test_missing_log_output.tcl
@@ -18,6 +18,21 @@ eexpect "255"
 eexpect ":/# "
 end_test
 
+start_test "Check that a broken stderr prints a message to the log files."
+send "$argv start -s=path=logs/db --insecure --logtostderr --verbosity=1 2>&1 | cat\r"
+eexpect "CockroachDB node starting"
+system "killall cat"
+eexpect ":/# "
+system "grep -F 'log: exiting because of error: write /dev/stderr: broken pipe' logs/db/logs/cockroach.log"
+end_test
+
+start_test "Check that a broken log file prints a message to stderr."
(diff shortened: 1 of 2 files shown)


benesch:
Needed to apply the following patch to get tests to pass:

```diff
diff --git a/pkg/cli/interactive_tests/test_missing_log_output.tcl b/pkg/cli/interactive_tests/test_missing_log_output.tcl
index 58e219737..f1c391c85 100644
--- a/pkg/cli/interactive_tests/test_missing_log_output.tcl
+++ b/pkg/cli/interactive_tests/test_missing_log_output.tcl
@@ -27,7 +27,7 @@ system "grep -F 'log: exiting because of error: write /dev/stderr: broken pipe'
 end_test
 
 start_test "Check that a broken log file prints a message to stderr."
-send "$argv start -s=path=/dev/null --insecure --logtostderr\r"
+send "$argv start -s=path=logs/db --log-dir=/dev/null --insecure --logtostderr\r"
 eexpect "log: exiting because of error: log: cannot create log: open"
 eexpect "not a directory"
 eexpect ":/# "
```

This is a more sensible test, anyway. I'll send that patch on master.

benesch:
Never mind; @andreimatei already made the equivalent fix on master.

bdarnell (approved):

PR #19289, opened 2017-10-16
log: remove workaround for os.Stat failure on Windows pre-go1.9

benesch:


BramGruneir:
Have you tested this on windows?

BramGruneir:
LGTM if you have.

benesch:
Yup, finally found the time to verify. It works! 🎉 TFTR

PR #19278, opened 2017-10-16
sql: remove unnecessary DatumAlloc parameter from index key methods

richardwu:
DatumAllocs were previously used for `DecodeKeyVals` but now we pass
in slices of Datums.

This affected `DecodeIndexKey(Prefix)?` and `ExtractIndexKey`.

PR #19246, opened 2017-10-13
sqlbase: remove ColumnType from EncDatum

RaduBerinde:
#### distsqlrun: require types for RowBuffer

Curretly the RowBuffer (used mainly in tests) relies on the types in the
`EncDatum`s (unless it is an empty buffer). To prepare for a removing the type
information from `EncDatum`, requires the types in all cases.

#### sqlbase: remove ColumnType from EncDatum

We carry around the entire `ColumnType` with every `EncDatum`. Initially
`ColumnType` was just a 32-bit value (it used memory that is otherwise lost to
padding). But in time this type grew much bigger.

This change modifies the EncDatum interface to require the type information to
be passed in. We leave the type as a deprecated field and assert that the type
matches. In a subsequent commit, this type will be removed.

In addition to the immediate performance improvement, these changes should help
facilitate switching to a more efficient representation of rows or row groups
(which would definitely require the type information to be passed in).

#### sqlbase: remove the deprecated type from EncDatum

Infrastructure benchmarks for this group of changes:

```
name                        old time/op    new time/op    delta
Infrastructure/n1/r1-4         186µs ± 4%     187µs ±10%     ~       (p=0.959 n=8+8)
Infrastructure/n1/r100-4       395µs ±29%     345µs ±12%  -12.51%    (p=0.019 n=9+9)
Infrastructure/n1/r10000-4    16.9ms ±24%    14.1ms ±44%     ~     (p=0.052 n=10+10)
Infrastructure/n3/r1-4         598µs ± 2%     580µs ± 5%   -3.02%   (p=0.043 n=7+10)
Infrastructure/n3/r100-4      1.30ms ± 6%    1.15ms ± 3%  -12.09%   (p=0.000 n=9+10)
Infrastructure/n3/r10000-4    56.9ms ±14%    45.6ms ± 4%  -19.79%   (p=0.000 n=10+9)

name                        old alloc/op   new alloc/op   delta
Infrastructure/n1/r1-4        37.3kB ± 0%    33.0kB ± 0%  -11.63%   (p=0.000 n=10+9)
Infrastructure/n1/r100-4       187kB ± 0%     123kB ± 0%  -34.51%  (p=0.000 n=10+10)
Infrastructure/n1/r10000-4    15.8MB ± 0%     9.6MB ± 0%  -39.25%  (p=0.000 n=10+10)
Infrastructure/n3/r1-4         129kB ± 2%     117kB ± 2%   -9.18%  (p=0.000 n=10+10)
[...]

PR #19293, opened 2017-10-16
Revert "sql: remove unnecessary DatumAlloc parameter from index key methods"

RaduBerinde:
Reverts cockroachdb/cockroach#19278

We actually need the `DatumAlloc` in `ExtractIndexKey` [here](https://github.com/cockroachdb/cockroach/blob/407057955ddbf182e2ac12a7fbf40f6afd4e6922/pkg/sql/sqlbase/table.go#L901); we currently incorrectly allocate a new one. PR #19246 fixes this.

### DIFF
--- a/pkg/server/settingsworker.go
+++ b/pkg/server/settingsworker.go
@@ -46,7 +46,7 @@ func (s *Server) refreshSettings() {
 		// First we need to decode the setting name field from the index key.
 		{
 			nameRow := []sqlbase.EncDatum{{Type: tbl.Columns[0].Type}}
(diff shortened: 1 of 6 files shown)


richardwu:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19293#-:-KwawsPA5sXArYKHZpjO:bnfp4nl)*

PR #19297, opened 2017-10-16
sql/parser: Add builtin SQL function date_trunc()

rytaft:
Add support for the date_trunc() function, similar to the PostgreSQL
function which truncates dates and timestamps to the given
precision.  It should return the input date/timestamp with all fields
that are less significant than the selected one set to zero (or one,
for day and month).


knz:
The code appears correct to me, but I would recommend trying to make it slightly more compact; some suggestions below.

---

Reviewed 3 of 3 files at r1.

---

*[pkg/sql/logictest/testdata/logic_test/builtin_function, line 1163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwduEL0SUVmVqOKC1cG:-KwduEL0SUVmVqOKC1cH:bty12of) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/logictest/testdata/logic_test/builtin_function#L1163)):*
> ```
> 100801
> 
> query T
> ```

To condense the test, to make it both easier to read and faster to execute, I would recommend a SQL query:

```sql
SELECT date_trunc(prec, ts)::string,
       date_trunc(prec, ts) = date_trunc(prec || 's', ts)
  FROM (VALUES ('2016-02-10 19:46:33.306157519'::timestamp)) AS ts(ts),
        unnest(ARRAY['millisecond', 'microsecond', 'second', 'minute', 'hour', ...]) AS prec;
```

---

*[pkg/sql/parser/builtins.go, line 1398 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwdsjPwAtmF3rEqNRG1:-KwdsjPxW8_WRVSqA4qQ:b-hbhhhv) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/parser/builtins.go#L1398)):*
> ```Go
> 	},
> 
> 	"date_trunc": {
> ```

We've recently started becoming more thorough with linking to the pg docs for the functions we reproduce in CockroachDB. I would therefore suggest you add the relevant link as a comment here: https://www.postgresql.org/docs/10/static/functions-datetime.html#functions-datetime-trunc

(Although I also acknowledge we've not yet backfilled the links for the other functions that were implemented before we started finding this important...)

---

*[pkg/sql/parser/builtins.go, line 1412 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwdtBoZvhTBIzFaQoOm:-KwdtBoZvhTBIzFaQoOn:bgrg3zz) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/parser/builtins.go#L1412)):*
> ```Go
> 				"significant than `element` to zero (or one, for day and month)\n\n" +
> 				"Compatible elements: year, quarter, month, week, hour, minute, second,\n" +
> 				"millisecond, microsecond",
> ```

Period missing at the end of the sentence: "... millisecond, microsecond."

here and below.

---

*[pkg/sql/parser/builtins.go, line 3086 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-Kwdt_dWCgmEyzHiN2I8:-Kwdt_dWCgmEyzHiN2I9:b-4ahpym) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/parser/builtins.go#L3086)):*
> ```Go
> 	switch timeSpan {
> 	case "year", "years":
> 		toTime := time.Date(year, monthTrunc, dayTrunc, hourTrunc, minTrunc, secTrunc, nsecTrunc, loc)
> ```

The two lines "toTime = time.Date(...); return MakeDTimestampTZ(...)" repeat themselves over and over in this switch.

I believe you can apply DRY a bit more here and at least extract the call to MakeDTimestampTZ below, and probably the call to time.Date() as well, recognizing that most of the switch case only change a subset of the arguments.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-:-KwdvkaM6dFKMZ25kUb5:bgojzx7)*

dt:
:lgtm: mod @knz's comments on DRYing a little in the implementation.

I'd probably move the logic test cases to the `datetime` file (that's where `extract` tests are) rather than the `builtin_functions` file.

---

---

*[pkg/sql/logictest/testdata/logic_test/builtin_function, line 1163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwduEL0SUVmVqOKC1cG:-KwejHbJAOdo3-j18Zog:bv3nlo8) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/logictest/testdata/logic_test/builtin_function#L1163)):*

To condense the test, to make it both easier to read and faster to execute, I would recommend a SQL query:

```sql
SELECT date_trunc(prec, ts)::string,
       date_trunc(prec, ts) = date_trunc(prec || 's', ts)
  FROM (VALUES ('2016-02-10 19:46:33.306157519'::timestamp)) AS ts(ts),
        unnest(ARRAY['millisecond', 'microsecond', 'second', 'minute', 'hour', ...]) AS prec;
```
</blockquote></details>

While in general I like DRYing, I also like tests that are simple, and, importantly, *obviously* correct. This suggestion is certainly concise, but it took me a little bit of effort to follow what was going on with the values clause, array+unpack, etc. Personally I think I have a (weak) preference for something slightly less clever and more obvious. Putting all the cases in a single statement also makes it a little less obvious which one fails or potentially easier to overlook an incorrect value in the middle of a long line -- again, not a big deal, but just a little more effort on the human's part that my personal inclination would be to let a computer do.

Anyway, these are very weak preferences on my part, so I'm fine with it as-is (it matches the repeated literal in the `extract` tests), or as @knz suggested.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-:-KwejsFR3iLARRvsJndq:bcgp0pc)*

knz:

---

*[pkg/sql/logictest/testdata/logic_test/builtin_function, line 1163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwduEL0SUVmVqOKC1cG:-KwerWqFwP1YrvKxAcaE:b3fjj4o) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/logictest/testdata/logic_test/builtin_function#L1163)):*

While in general I like DRYing, I also like tests that are simple, and, importantly, *obviously* correct. This suggestion is certainly concise, but it took me a little bit of effort to follow what was going on with the values clause, array+unpack, etc. Personally I think I have a (weak) preference for something slightly less clever and more obvious. Putting all the cases in a single statement also makes it a little less obvious which one fails or potentially easier to overlook an incorrect value in the middle of a long line -- again, not a big deal, but just a little more effort on the human's part that my personal inclination would be to let a computer do.

Anyway, these are very weak preferences on my part, so I'm fine with it as-is (it matches the repeated literal in the `extract` tests), or as @knz suggested.
</blockquote></details>

An alternative is to create a small table with the various precision strings upfront (create table + insert .. values), and still express the test as a join.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

rytaft:
Thanks! It seems there are `extract` tests in both files.  Should I consolidate all the `extract` and `date_trunc` tests in the `datetime` file?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-:-KwevbyFAT8lGXFdGJ5S:b9z2b5l)*

dt:
ah, whoops, yeah, I'd move it all to `datetime`

rytaft:

---

*[pkg/sql/parser/builtins.go, line 1412 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwdtBoZvhTBIzFaQoOm:-KwfGNJAqv-ITU0-cOa9:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/parser/builtins.go#L1412)):*

Period missing at the end of the sentence: "... millisecond, microsecond."

here and below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

knz:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

rytaft:

---

*[pkg/sql/parser/builtins.go, line 1398 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwdsjPwAtmF3rEqNRG1:-KwfQf2M9Ujyg3imeVhP:b18viun) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/parser/builtins.go#L1398)):*

We've recently started becoming more thorough with linking to the pg docs for the functions we reproduce in CockroachDB. I would therefore suggest you add the relevant link as a comment here: https://www.postgresql.org/docs/10/static/functions-datetime.html#functions-datetime-trunc

(Although I also acknowledge we've not yet backfilled the links for the other functions that were implemented before we started finding this important...)
</blockquote></details>

I've added the link (similar to the comment for `encode`).  Let me know if you had something else in mind for the comment.  Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

rytaft:
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/builtins.go, line 3086 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-Kwdt_dWCgmEyzHiN2I8:-KwfaItxyWtCJ5mgeRUY:bquj2jf) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/parser/builtins.go#L3086)):*

The two lines "toTime = time.Date(...); return MakeDTimestampTZ(...)" repeat themselves over and over in this switch.

I believe you can apply DRY a bit more here and at least extract the call to MakeDTimestampTZ below, and probably the call to time.Date() as well, recognizing that most of the switch case only change a subset of the arguments.
</blockquote></details>

I've moved those two lines below the switch block.  Ready for you to take another look.  Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

rytaft:
I've moved all the tests to `datetime` and consolidated them into a few select queries similar to @knz's suggestion

---

Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-:-Kwgy4q4mVokt4Eg3hNp:b-rd96z6)*

rytaft:
Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/builtin_function, line 1163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwduEL0SUVmVqOKC1cG:-KwgyWAUd9wsRJ3Q-MZG:b-ecz75e) ([raw file](https://github.com/cockroachdb/cockroach/blob/fbdd2286956fc05eb0c339d945b87deefeab50e7/pkg/sql/logictest/testdata/logic_test/builtin_function#L1163)):*

An alternative is to create a small table with the various precision strings upfront (create table + insert .. values), and still express the test as a join.
</blockquote></details>

I created a table with input and output data for both `extract` and `date_trunc`, and added a few select statements for the tests.  I kept the queries very simple (no joins) so that it's easy to determine exactly what went wrong if there is an error.  Let me know if you still see ways to condense it more... Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

knz:
:+1:

---

Reviewed 3 of 3 files at r3.

---

*[pkg/sql/logictest/testdata/logic_test/datetime, line 365 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwigoixOSLD7FEoX85h:-KwigoixOSLD7FEoX85i:buk5uhe) ([raw file](https://github.com/cockroachdb/cockroach/blob/0abbba9d9a9e494ee8a8c923c1a7e777f4b1145c/pkg/sql/logictest/testdata/logic_test/datetime#L365)):*
> ```
>   element STRING,
>   input   TIMESTAMP WITH TIME ZONE,
>   output1 INT,
> ```

Output column names should be descriptive -- `extract_result`, `date_trunc_result`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-:-KwihG0-vCfNpS3sVdbl:bb6mt9b)*

dt (approved):

rytaft:

---

*[pkg/sql/logictest/testdata/logic_test/datetime, line 365 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-KwigoixOSLD7FEoX85h:-Kwjuhvb9aOzovfvJp6j:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0abbba9d9a9e494ee8a8c923c1a7e777f4b1145c/pkg/sql/logictest/testdata/logic_test/datetime#L365)):*

Output column names should be descriptive -- `extract_result`, `date_trunc_result`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297)*

knz:
:lgtm:

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19297#-:-KwjxJ_Inq6VqSn9_oD6:bnfp4nl)*

PR #19300, opened 2017-10-17
kv, roachpb: introduce a "best effort" option for scans

andreimatei:
The option is defined to scan until the end of a range. The idea is that
continuing would incur considerable cost, since it will likely involve
an RPC to a different node.
Uses cases include "soft limits" in SQL - situations where we estimate a
limit on how much we'd like to read but we're not really sure about it
(fewer data may suffice), range descriptor prefetching, and possibly
upcoming scans done for statistics-gathering.
There's also an option to scan until a minimum number of keys is read -
e.g. until one key is read. This would be useful for descriptor
(pre)fetching and also for SQL which would otherwise scan repeatedly to
skip empty ranges.

Future work might involve a finer-grained option for discriminating
between future local versus remote ranges (as the cost of continuing is
different between the two), and an option for stopping a scan when an
intent is encountered (as resolving the intent has a considerable cost).


andreimatei:
cc @RaduBerinde

RaduBerinde:
It's of limited use in SQL if it only supports a single scan request. Is it very hard to support multiple scans? The current code handles batch limits fine with multiple scans..

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-:-KweYEtTVt14yo7DcQ_f:bnf6f61)*

andreimatei:
You're thinking of scans for multiple index spans with a global limit?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-:-Kwez3gPZ83bWjJ93Y3l:byopc77)*

RaduBerinde:
Yes, e.g. `SELECT * FROM t WHERE k1 IN [1,5,10] LIMIT 10`

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-:-KwfPopP4XMwanfls5rK:b7w8u45)*

bdarnell:
The `INCONSISTENT` read option turned out to not be very useful because it was too broadly defined. When enabled it allowed reading from non-leaseholders, reading behind intents, bypassing the command queue, etc. 

This "best effort" flag is implement as a single-range limit, but it will block as long as it takes on that range (waiting on intent resolution, etc). Is that what we want, or do we want something that can also give up early when there are intents in the way? The latter may be useful for implementing something like the SKIP_LOCKED option, but that mode wouldn't want to restrict itself to a single range. I think that instead of a vague "best effort" flag that could be extended in various directions, it's better to be specific about what conditions cause the request to bail out (these flags can then be combined in various ways). 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-:-KwgENzWcDUAnrAqw2RG:b-7dfv7m)*

andreimatei:
I've extended this to make it work for multiple scans in a batch and also I've made it be an explicit range boundary option. Options for handling intents may come next.
PTAL.

---

Review status: 0 of 11 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-:-Kx-AH9YEAWtbdP8KZjh:b-5iy1xg)*

bdarnell:
:lgtm:

---

Reviewed 11 of 11 files at r2.

---

*[pkg/internal/client/batch.go, line 270 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx8uyT9NT2LmtJ74gLe:-Kx8uyTAwL2nBofK7Ukh:b6waivl) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/internal/client/batch.go#L270)):*
> ```Go
> 				result.ResumeSpan = *reply.Header().ResumeSpan
> 				if reply.Header().ResumeDetails == nil {
> 					log.Fatalf(ctx, "reply has ResumeSpan but no ResumeDetails: %s", result)
> ```

Won't this be a problem if the client is used to talk to an older server? Or is this code path only hit for replies generated locally?

---

*[pkg/kv/batch.go, line 105 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx9-0wI9JDAVVrbxqdM:-Kx9-0wI9JDAVVrbxqdN:b-wjaf45) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/kv/batch.go#L105)):*
> ```Go
> 	truncBA := ba
> 	truncBA.Requests = nil
> 	// Remove any ScanOptions, in case any were set. ScanOptions currently only
> ```

But you want to extend this to include options that will need to be handled on the server (skipping intents). Let's just let it pass through here instead of making a special case we'll have to undo later. 

---

*[pkg/roachpb/api.proto, line 82 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx90UCfmOj5eKMaO_FK:-Kx90UCfmOj5eKMaO_FL:b-28hiva) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L82)):*
> ```protobuf
>   // This field appeared in v1.2. Responses from storage coming from older
>   // servers will not contain it, but the DistSender always fills it in.
>   optional ResumeDetails resume_details = 7;
> ```

We might want to give the ResumeReason an `UNKNOWN` zero value and make this non-nullable to avoid an allocation and some pointer manipulations.  (What else might go in the ResumeDetails struct? Do we need this wrapper?)

---

*[pkg/roachpb/api.proto, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx91EhtW7rlp5_14rAk:-Kx91EhtW7rlp5_14rAl:b4u75ab) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L247)):*
> ```protobuf
> // If ScanOptions is present on a batch, the batch can only be made up of scan
> // requests (i.e.  {Reverse,}ScanReques), except for
> // {Begin,End}TransactionRequest which are still allowed. Moreover, the batch
> ```

It's not obvious to me that it's safe to combine these limited scans with `{Begin,End}Transaction`. Since we currently would never generate such a batch (BeginTransaction is sent on the first write, so it would never appear in a scan-only batch), we should probably disallow this mixture until we have a stronger need for it. 

---

*[pkg/roachpb/api.proto, line 255 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx91k-CRXKB32_KlVVg:-Kx91k-CRXKB32_KlVVh:bgy661l) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L255)):*
> ```protobuf
> 
>   // stop_at_range_boundary, if set, means that the scans will stop at the first
>   // range boundary after min_span_request_keys is specified (if it is
> ```

`s/min_span_request_keys/min_results/g`

---

*[pkg/roachpb/api.proto, line 300 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx90mNTrQ3zM-8ByR3W:-Kx90mNTrQ3zM-8ByR3X:b-ia6wgu) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L300)):*
> ```protobuf
>   optional bool return_intents = 3 [(gogoproto.nullable) = false];
> 
>   optional ScanOptions opt = 4 [(gogoproto.nullable) = false];
> ```

It doesn't look like the `{Reverse,}ScanRequest.opt` fields are used.

---

*[pkg/ui/yarn.lock, line 444 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx8wScwc8wFqi9ZkOAw:-Kx8wScwc8wFqi9ZkOAx:b-lnrfwk) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/ui/yarn.lock#L444)):*
> ```
>   resolved "https://registry.yarnpkg.com/async/-/async-0.9.2.tgz#aea74d5e61c1f899613bf64bda66d4c78f2fd17d"
> 
> async@^2.1.2, async@^2.3.0:
> ```

Not sure what happened here but I don't think you want this file in your commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-:-Kx98rkvC6cGajRzdRF2:bnfp4nl)*

andreimatei:

---

*[pkg/internal/client/batch.go, line 270 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx8uyT9NT2LmtJ74gLe:-Kx9gO3ccVIvJ5PL6IvO:b-reaihh) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/internal/client/batch.go#L270)):*

Won't this be a problem if the client is used to talk to an older server? Or is this code path only hit for replies generated locally?
</blockquote></details>

The `ResumeDetails` is filled by the DistSender, which is local to the client (except RemoteSender wierdness, but I believe @tschottdorf has all but eliminated that case).

---

*[pkg/kv/batch.go, line 105 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx9-0wI9JDAVVrbxqdM:-Kx9gohChM_7pvrV33aq:bxubtoc) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/kv/batch.go#L105)):*

But you want to extend this to include options that will need to be handled on the server (skipping intents). Let's just let it pass through here instead of making a special case we'll have to undo later. 
</blockquote></details>

The idea is that when we extend it, this will become more discriminant and only clear it when its valid to do so.
This was done as a suggestion from @nvanbenschoten meant to trigger an approving nod from byte-counting @petermattis.
I'm flexible on it, though.

---

*[pkg/roachpb/api.proto, line 82 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx90UCfmOj5eKMaO_FK:-Kx9wkrAIJtxA_30WqtL:bm1y6m7) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L82)):*

We might want to give the ResumeReason an `UNKNOWN` zero value and make this non-nullable to avoid an allocation and some pointer manipulations.  (What else might go in the ResumeDetails struct? Do we need this wrapper?)
</blockquote></details>

removed the wrapper

---

*[pkg/roachpb/api.proto, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx91EhtW7rlp5_14rAk:-Kx9iOepX3os6fCAsnRC:b-veudx6) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L247)):*

It's not obvious to me that it's safe to combine these limited scans with `{Begin,End}Transaction`. Since we currently would never generate such a batch (BeginTransaction is sent on the first write, so it would never appear in a scan-only batch), we should probably disallow this mixture until we have a stronger need for it. 
</blockquote></details>

Why do you suspect it may not be safe to combine these scans with begin/end (or, with which one of begin/end specifically you think it may not be safe)?
I've modeled that after what we currently allow in a batch with `max_span_request_keys`.

About the `begin` being sent with the first write - that's true when requests are batched and sent through the `client.Txn` interface, but is that the same as describing what the "server" is supposed to accept?

---

*[pkg/roachpb/api.proto, line 255 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx91k-CRXKB32_KlVVg:-Kx9kdSHYJalpuupDAe-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L255)):*

`s/min_span_request_keys/min_results/g`
</blockquote></details>

Done.

---

*[pkg/roachpb/api.proto, line 300 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx90mNTrQ3zM-8ByR3W:-Kx9kiq2Zo9wt2N8BfgF:bfwz83) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L300)):*

It doesn't look like the `{Reverse,}ScanRequest.opt` fields are used.
</blockquote></details>

good catch. This had moved to the header but I forgot to remove it from here.

---

*[pkg/ui/yarn.lock, line 444 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx8wScwc8wFqi9ZkOAw:-Kx9ptJDAhUh2qiqUfXw:bvxpdgj) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/ui/yarn.lock#L444)):*

Not sure what happened here but I don't think you want this file in your commit.
</blockquote></details>

Done. One of my laptops updated this file. I didn't know why but I saw that updates to it were often associated with rando proto updates, so I assumed it's something about my yarn version (and I had the latest yarn version on that laptop).
The current laptop doesn't seem to generate those updates...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300)*

bdarnell:
Reviewed 10 of 10 files at r4.

---

*[pkg/roachpb/api.proto, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx91EhtW7rlp5_14rAk:-KxAG5n2eV3axMMjTSUp:bvhvnh1) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L247)):*

Why do you suspect it may not be safe to combine these scans with begin/end (or, with which one of begin/end specifically you think it may not be safe)?
I've modeled that after what we currently allow in a batch with `max_span_request_keys`.

About the `begin` being sent with the first write - that's true when requests are batched and sent through the `client.Txn` interface, but is that the same as describing what the "server" is supposed to accept?
</blockquote></details>

I think letting them through for `max_span_request_keys` is a mistake too. My concern is mainly about the EndTransaction: when we reach one of these limits, we return *without error* and without executing the last requests in the batch. This seems to falsely imply that the transaction was committed.  Since no client sends batches that combine scans with begin/end txn, we might as well disallow this combination until we have a better justification and can define how a  scan that reaches its limit should interact with the transaction ops. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300)*

andreimatei:
Review status: 7 of 9 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/roachpb/api.proto, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx91EhtW7rlp5_14rAk:-KxAUICxoI_ajtnCdQTW:bkb3q59) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/roachpb/api.proto#L247)):*

I think letting them through for `max_span_request_keys` is a mistake too. My concern is mainly about the EndTransaction: when we reach one of these limits, we return *without error* and without executing the last requests in the batch. This seems to falsely imply that the transaction was committed.  Since no client sends batches that combine scans with begin/end txn, we might as well disallow this combination until we have a better justification and can define how a  scan that reaches its limit should interact with the transaction ops. 
</blockquote></details>

As we've discussed, the transaction does in fact commit because the DistSender splits the `EndTransaction` into its own "part" of the batch and runs it if no error is returned by previous parts. And `Begin/EndTransaction` appear in these batches when non-transactional batches that span ranges are wrapped by the TxnCoordSender.

---

*[pkg/kv/batch.go, line 105 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300#-Kx9-0wI9JDAVVrbxqdM:-KxF9I9aY9W-xSTRzUrv:b9tt6tu) ([raw file](https://github.com/cockroachdb/cockroach/blob/cacfd3af574474acb1d98e03b1b7081f8324fe31/pkg/kv/batch.go#L105)):*

The idea is that when we extend it, this will become more discriminant and only clear it when its valid to do so.
This was done as a suggestion from @nvanbenschoten meant to trigger an approving nod from byte-counting @petermattis.
I'm flexible on it, though.
</blockquote></details>

After all, I've removed this. It was incorrect to do it here, because a truncated batch might be recursively passed by the DistSender to be further truncated when the DistSender's idea of ranges changes. So it's unclear where this clearing of the field would be correct...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19300)*

PR #19306, opened 2017-10-17
cli: fix broken output of special characters in queries

### LINKED ISSUE #19307: sql: special characters in column names are not handled properly
Identified by @bdarnell in https://github.com/cockroachdb/cockroach/pull/18576#issuecomment-330672033 and https://github.com/cockroachdb/cockroach/pull/18576#issuecomment-336720926:

> Do CSV readers interpret \n in this way? My understanding is that the way to represent a cell containing a newline in a CSV is to enclose the value in quotes and then include the literal newline

This pertains to the following output by the csv and tsv formatters:

```sql
> \set display_format csv
> select e'ab\ncd' as "ef
gh";
```

result:

```
"ef\ngh"
"ab
cd"
```

i.e. the value part is properly formatted according to tsv/csv rules (value contains literal newline) but not the column name part.

### LINKED ISSUE #20848: cli: Title width mis-matches rest of results table width
I don't know what is causing it as it seems to be off by exactly 3 chars in this case.

For example,

```
root@:26257/test> select '::ff'::inet > '12.168.0.1'::inet;
+--------------------------------+
| '::ff'::INET > '12.168.0.1'::INET |
+--------------------------------+
| true                           |
+--------------------------------+
```

knz:
(This is the bug fix prefix to #18576. Pulling as separate PR since the other PR is likely to iterate longer.)

Fixes #19307.
Fixes #20848.

Prior to this patch, the CLI tools that display SQL rows (namely,
`sql`, `user`, `zone`) would make an effort to display anything with
the Unicode graphical property, and all whitespace, unchanged.

As this is the most common case, a bug slipped through: when a
character is neither graphical nor whitespace, it would erroneously
get processed two times for escape: a first one by the low-level value
formatter, and a second time by the table formatter.

For example:

```
> cockroach sql -e "select e'\x01' as x"
x
"""\x01"""
(1 row)
```

The triple quotes in the output were because first the ASCII 1 byte
would be escaped to the string `"\x01"` (first layer of quoting), then
the TSV formatter would introduce the additional necessary CSV escapes
by tripling them.

The error was that although the content of the string must be escaped
by the first-level formatter, that formatter should not introduce
boundary quotes -- the boundary quoting is under the responsibility of
the table formatter.

With this patch   :

```
> cockroach sql -e "select e'\x01' as x"
x
\x01
(1 row)
```

Note that although this patch addresses the most egregious
consequences of the original bug, the approach is arguably not yet
complete: we probably would like the SQL formatter to insert values
that are *equivalent* to those received from the server; this means
that *whether to quote or not should be decided by the formatter*.
Note that only the `raw` and `sql` formatters may be interested in
this change; the more common `pretty`, `tsv`, `csv` and `records`
formatter are fine as-is. A subsequent commit can introduce the
corresponding necessary plumbing.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -589,7 +589,7 @@
   branch = "master"
   name = "github.com/olekukonko/tablewriter"
   packages = ["."]
-  revision = "a7a4c189eb47ed33ce7b35f2880070a0c82a67d4"
+  revision = "b8a9be070da40449e501c3c4730a889e42d87a9e"
 
 [[projects]]
   name = "github.com/opencontainers/go-digest"
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -39,6 +39,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/security/securitytest"
 	"github.com/cockroachdb/cockroach/pkg/server"
 	"github.com/cockroachdb/cockroach/pkg/sql/lex"
+	"github.com/cockroachdb/cockroach/pkg/sql/sem/tree"
 	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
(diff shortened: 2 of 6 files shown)


bdarnell:

---

*[pkg/cli/sql_util.go, line 667 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-Kwjj-wtuVcMAYz69aeB:-Kwjj-wtuVcMAYz69aeC:btu33co) ([raw file](https://github.com/cockroachdb/cockroach/blob/c5fafe67a791726db77ac17fa99b8049359b0c08/pkg/cli/sql_util.go#L667)):*
> ```Go
> 		s := fmt.Sprintf("%+q", t)
> 		// Strip the start and final quotes. The surrounding display
> 		// format (e.g. CSV/TSV) will add its own quotes.
> ```

I think we still need a test where these quotes are added. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306)*

knz:

---

*[pkg/cli/sql_util.go, line 667 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-Kwjj-wtuVcMAYz69aeB:-L3OcImIKDlBE06u0E4O:bkyytn5) ([raw file](https://github.com/cockroachdb/cockroach/blob/c5fafe67a791726db77ac17fa99b8049359b0c08/pkg/cli/sql_util.go#L667)):*

I think we still need a test where these quotes are added. 
</blockquote></details>

I have added the missing tests. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306)*

knz:
Also added an extra commit to fix #19307.

bdarnell:
Reviewed 1 of 3 files at r2, 1 of 2 files at r3, 2 of 3 files at r4.
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/cli/format_table.go, line 223 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-L3QtUW95rMny1HSZ_4V:-L3QtUW95rMny1HSZ_4W:b-eo1s2n) ([raw file](https://github.com/cockroachdb/cockroach/blob/f243262c6fee6c94115315b1ba8bc26667fd9dd9/pkg/cli/format_table.go#L223)):*
> ```Go
> 		p.table.SetAutoWrapText(false)
> 		for i, c := range p.cols {
> 			p.cols[i] = strings.Replace(strings.Replace(c, "\n", `\n`, -1), "\t", `\t`, -1)
> ```

Escaping the output strings with `%q` only to (partially) unescape them here is silly (and incorrect - this needs to be more sophisticated to handle escaped backslashes). I think we need to pass the raw values further down and then escape them in a format-appropriate way. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306)*

knz:
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/cli/format_table.go, line 223 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-L3QtUW95rMny1HSZ_4V:-L3SzTy60HR0SuZRsMS3:b-vkn2pn) ([raw file](https://github.com/cockroachdb/cockroach/blob/f243262c6fee6c94115315b1ba8bc26667fd9dd9/pkg/cli/format_table.go#L223)):*

Escaping the output strings with `%q` only to (partially) unescape them here is silly (and incorrect - this needs to be more sophisticated to handle escaped backslashes). I think we need to pass the raw values further down and then escape them in a format-appropriate way. 
</blockquote></details>

This is the format-appropriate location :)

Anyway we're really working around a bug in the upstream table writer. Instead of this work around I went and fixed the bug upstream: https://github.com/olekukonko/tablewriter/pull/85

Will rebase when that merges.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306)*

bdarnell:
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/cli/format_table.go, line 223 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-L3QtUW95rMny1HSZ_4V:-L3TXBToCnV0cD9JQQb2:bx25sug) ([raw file](https://github.com/cockroachdb/cockroach/blob/f243262c6fee6c94115315b1ba8bc26667fd9dd9/pkg/cli/format_table.go#L223)):*

This is the format-appropriate location :)

Anyway we're really working around a bug in the upstream table writer. Instead of this work around I went and fixed the bug upstream: https://github.com/olekukonko/tablewriter/pull/85

Will rebase when that merges.
</blockquote></details>

Oh, I misread the distinction between `"\n"` and `` `\n` `` (I think it would be more clear to use `"\\n"` for the escaped form). Carry on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306)*

knz:
Review status: 1 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/cli/format_table.go, line 223 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-L3QtUW95rMny1HSZ_4V:-L4CwX78iN9GTWOamlE2:b-26rxmg) ([raw file](https://github.com/cockroachdb/cockroach/blob/f243262c6fee6c94115315b1ba8bc26667fd9dd9/pkg/cli/format_table.go#L223)):*

Oh, I misread the distinction between `"\n"` and `` `\n` `` (I think it would be more clear to use `"\\n"` for the escaped form). Carry on.
</blockquote></details>

Fixed via upgrade to fixed upstream. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306)*

bdarnell:
:lgtm:

---

Reviewed 1 of 2 files at r6, 5 of 5 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19306#-:-L4H9Gw3fCVps5yqjukr:bnfp4nl)*

knz:
TFYR!

PR #17988, opened 2017-08-29
scripts: add a change log generator script

knz:
This groups commits by author, in the style of the Linux CHANGES file.

For example:

```
% ./scripts/changelog.py v1.0.5 v1.0.4
Ben Darnell:
        #17654 182613 (+  63 -  20 ~  83/ 1) cherrypick: security: Disable 3DES cipher suites

Jordan Lewis:
        #17653 828a46 (+  23 -   4 ~  27/ 4) cherrypick: sql: two bugfixes to fks that use a partial index (2 commits)

kena:
        #17689 6255bb (+   7 -   2 ~   9/ 1) cherrypick: log: allow panic to propagate on crash reporting timeouts
        #17675 a725c2 (+  17 -   5 ~  22/ 2) cherrypick: log: ensure Fatal calls are reported even if logs don't go to stderr
        #17681 e73fc3 (+  50 -  39 ~  89/ 2) cherrypick: sql: disallow AS OF SYSTEM TIME 0, add more AS OF tests
        #17281 8d0dea (+  15 -   0 ~  15/ 3) cherrypick: sql: ensure that planNodes are Close()d on error after expansion

Tobias Schottdorf:
        #17385 63abc8 (+  18 -   3 ~  21/ 1) cherry-pick: base: bump DefaultSendNextTimeout from 500ms to 10m
```

PR #19308, opened 2017-10-17
scripts: migrate non-essential scripts to separate repo

knz:
Suggested by @bdarnell in https://github.com/cockroachdb/cockroach/pull/17988#issuecomment-336678825:

The script `pprof` is retired, as its documentation indicated it is
becoming obsolete as of Go 1.9 which is now current.

The script `travis_retry.sh` is retired, as we are not using Travis
any more.

The script `verify-release-binaries.sh` is retired, as it was replaced
by http://github.com/benesch/bincheck/.

The following scripts have been migrated, together with their history,
to the new repository at https://github.com/cockroachdb/scripts:

- `authors.sh`
- `bench`
- `crlog.vim`
- `crlogictest.vim`
- `git-blame-stats.sh`
- `log2html.sh`
- `purge-terraform.sh`
- `torch`


Makefile
```diff
@@ -342,11 +342,11 @@ $(ARCHIVE): $(ARCHIVE).tmp
 	gzip -c $< > $@
 
 # TODO(benesch): Make this recipe use `git ls-files --recurse-submodules`
-# instead of scripts/ls-files.sh once Git v2.11 is widely deployed.
+# instead of build/ls-files.sh once Git v2.11 is widely deployed.
 .INTERMEDIATE: $(ARCHIVE).tmp
 $(ARCHIVE).tmp: ARCHIVE_BASE = cockroach-$(shell cat .buildinfo/tag)
 $(ARCHIVE).tmp: .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
-	scripts/ls-files.sh | $(TAR) -cf $@ -T - $(TAR_XFORM_FLAG),^,$(ARCHIVE_BASE)/src/github.com/cockroachdb/cockroach/, $^
+	build/ls-files.sh | $(TAR) -cf $@ -T - $(TAR_XFORM_FLAG),^,$(ARCHIVE_BASE)/src/github.com/cockroachdb/cockroach/, $^
```
benesch: Thanks!

scripts/travis_retry.sh
```diff
@@ -1,49 +0,0 @@
-#!/usr/bin/env bash
```
benesch: This script is used by azworker and gceworker

benesch (commented):
Thanks for taking this on, @knz! I'd personally prefer if we kept

* crlog.vim
* crlogictest.vim
* log2html.sh
* purge-terraform.sh

as those are scripts that are tightly coupled to this repository. (Though on the last one I'm particularly biased.) The others are scripts that are generally useful, interact mostly with Git/Go, so might as well live elsewhere.

benesch:
Probably wait for @bdarnell's opinion before making the changes I suggested.

bdarnell:
We definitely need to keep travis_retry.sh since it's used by the `*worker` scripts. (it's cargo-culted from somewhere and should probably be renamed; we never used travis in this repo). The git authors/blame scripts definitely belong in a separate repo. I'm more ambivalent about the log formatters and `bench`, which are more cockroach-specific. If we're going to have a separate repo I'd probably move them there, though. 

Unfortunately, we still need `scripts/pprof`. The `https+insecure` mode still doesn't work in Go 1.9. It's not cockroach-specific, though, so I guess for the sake of consistency it belongs in the other repo.

bdarnell:
On a more general note, I'm not 100% on board with the idea of this scripts repo - my issue with these git-parsing scripts is not that they should live in a cockroach-owned location but not in the main repo, it's that I think they don't really belong in a cockroach-owned location at all. They're more personal - are any of these scripts used by more than one person?

dt:
I'm 👎 on separate repos (in general) -- if/when we want to move, reconfigure, etc things, it is just more that has to move, *manually*, in lockstep.

tschottdorf:
I'm fine nuking all of the (non-essential) scripts and having their owners host them in their own repos. The idea with this "scripts" folder was that things would ultimately be useful, but I'm not sure that they are.

dt:
strawman: we could make a `contrib` folder (either top level with a `scripts` folder, or just under scripts) for "someone finds this useful, so here it is, ymmv" stuff, potentially including some of these scripts which aren't referenced in any of our guides/docs/workflows.

jordanlewis:
FWIW, I think it's super convenient to have scripts like `scripts/bench` accessible from the cockroach working directory. I use that one all the time.

PR #19315, opened 2017-10-17
partitioning: PARTITION BY syntax + PartitioningDescriptor proto + validatePartitioning and tests

danhhz:
This adds the PARTITION BY syntax, the PartitioningDescriptor proto, some validation methods + tests, and hooks the syntax up to the descriptor for CREATE TABLE. Filling in the descriptors currently doesn't have any other effect on the system, but getting this in will allow @benesch and I to start splitting up the work.

Subpartitioning is included, but a few of the features in the RFC are left for another PR (DEFAULT for list partitioning, MAXVALUE for range partitioning, PARTITION LIKE, repartitioning, partitioning of indexes other than the primary).

@knz I'd love for your eyes on this, especially the `sql.y`changes and the `addPartitionedBy` method in create.go. There was a decent amount of cargo culting, so please do give it a thorough look. Also note that (as reflected in the latest update to the RFC) we've moved away from using `table_ref` to define partitions for now. We took some informal polls, and everyone we asked assumed it would stay up to date if the underlying query changed (like a view does), so we've punted on it until we figure out the UX issues.


benesch:
Reviewed 6 of 6 files at r1, 7 of 7 files at r2, 6 of 6 files at r3.

---

*[pkg/sql/create.go, line 1130 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfIFDvsgIg2hOn9dip:-KwfIFDvsgIg2hOn9diq:bjsnulh) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1130)):*
> ```Go
> 	if err != nil {
> 		return nil, errors.Wrap(err, typedExpr.String())
> 	}
> ```

@knz should look at this. I think type checking is supposed to happen in `planner.CreateTable`, not `createTableNode.Start`, but I might be making that up.

---

*[pkg/sql/create.go, line 1133 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfJHmei1l5xh7UvstH:-KwfJHmei1l5xh7UvstI:bu8dvuj) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1133)):*
> ```Go
> 	datums := tupleDatum.(*parser.DTuple)
> 	if len(datums.D) != len(cols) {
> 		return nil, errors.Errorf("partition has %d columns but %d values were supplied",
> ```

optional nit: `column(s)`, `value(s)`

---

*[pkg/sql/create.go, line 1165 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfGVBf-BKoYynN-9Kj:-KwfGVBf-BKoYynN-9Kk:bil7tj7) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1165)):*
> ```Go
> 	colMap := make(map[sqlbase.ColumnID]int)
> 	for i := 0; i < len(partBy.Fields); i++ {
> 		if colOffset+i > len(indexDesc.ColumnIDs) {
> ```

Is this not off-by-one? I feel like you want `>=` but maybe my mental model is wrong.

---

*[pkg/sql/create.go, line 1176 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfHgULcpMUlLUw4Ac8:-KwfHgULcpMUlLUw4Ac9:bhvcdvh) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1176)):*
> ```Go
> 			return errors.New("declared columns must be a prefix of index being partitioned")
> 		}
> 		colMap[col.ID] = i
> ```

Seems unused?

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 14 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfD-2ciEhoXDvJB8lI:-KwfD-2ciEhoXDvJB8lJ:b-59vhx5) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L14)):*
> ```
> )
> 
> statement error declared columns must be a prefix of index being partitioned
> ```

The equivalent interleaved error message is "interleaved columns must match parent". I find that to be less confusing wording, but I don't have a strong preference besides making them consistent.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 36 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfGOK4vO1OFOgtgr8C:-KwfGOK4vO1OFOgtgr8D:b-6bka7n) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L36)):*
> ```
>         PARTITION p1_1 VALUES (0)
>     )
> )
> ```

Can you add a test for when the number of partition columns is greater than the number of columns in the primary key? (Unless I missed it!)

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 105 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfDPZpjjwmlj7YMNRb:-KwfDPZpjjwmlj7YMNRc:b-d433sm) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L105)):*
> ```
> # TODO(dan): Partition names should follow the same foldings and uniqueness
> # rules as table names. Add some more tests to the `case_sensitive_names` logic
> # test.
> ```

I'd add the parse tests examples here as `statement ok`s.

---

*[pkg/sql/parser/create.go, line 643 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfAAZyb3QNpqJLuDXq:-KwfAAZyb3QNpqJLuDXr:b-tkt896) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/create.go#L643)):*
> ```Go
> type PartitionBy struct {
> 	Fields NameList
> 	List   []ListPartition
> ```

Only one of `List` or `Range` can be set, right? Perhaps leave a comment.

---

*[pkg/sql/parser/create.go, line 680 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfAM9s1dL2An_Pq3FZ:-KwfAM9s1dL2An_Pq3F_:b-jhdhmr) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/create.go#L680)):*
> ```Go
> func (node ListPartition) Format(buf *bytes.Buffer, f FmtFlags) {
> 	fmt.Fprintf(buf, `PARTITION %s VALUES `, node.Name)
> ```

I believe `Fprintf`ing like this is dangerous, because `node.Name` might need escaping. `Name` implements `NodeFormatter`, so you can call `FormatNode(..., &node.Name)` instead.

---

*[pkg/sql/parser/create.go, line 701 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfAdgrETLtovm_QmuC:-KwfAdgrETLtovm_QmuD:b-pi9bzy) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/create.go#L701)):*
> ```Go
> func (node RangePartition) Format(buf *bytes.Buffer, f FmtFlags) {
> 	fmt.Fprintf(buf, `PARTITION %s VALUES LESS THAN `, node.Name)
> ```

Ditto here re: name formatting.

---

*[pkg/sql/sqlbase/encoded_datum.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfK8XvgKTwxSi_Z1jR:-KwfK8XvgKTwxSi_Z1jS:bvtsefl) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/encoded_datum.go#L98)):*
> ```Go
> 	if len(buf) == 0 {
> 		return EncDatum{}, nil, errors.New("empty encoded value")
> 	}
> ```

Seems weird that `EncDatumFromEncoded` panics and this returns an error. I feel like either `EncDatumFromEncoded` should return an error instead of panicking, or this check should be moved into the partitioning code.

---

*[pkg/sql/sqlbase/structured.go, line 1250 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfNZpTrFJNVs57xwf9:-KwfNZpTrFJNVs57xwfA:b-gqyf4q) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1250)):*
> ```Go
> 			// decide to be less strict for some reason, then adjust the
> 			// bytes.Compare below that verifies the ValuesLessThan sortedness
> 			// to use EncDatumRow.Compare instead.
> ```

Since the column ID is going to be ignored, and using `encoding.NoColumnID` is just a space optimization, I have a weak preference for not validating it.

---

*[pkg/sql/sqlbase/structured.go, line 1293 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfOOTaDM9QC1a1KCvP:-KwfOOTaDM9QC1a1KCvQ:bexw4vy) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1293)):*
> ```Go
> 					if err != nil {
> 						return fmt.Errorf("decoding %s: %v", p.Name, err)
> 					}
> ```

Ah, interesting. This is just to verify that the encoding isn't corrupt?

---

*[pkg/sql/sqlbase/structured.go, line 1297 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfOjyPYlgdJUkhIcQ9:-KwfOjyPYlgdJUkhIcQA:btk5sfa) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1297)):*
> ```Go
> 						return fmt.Errorf("%v cannot be present in more than one partition", datums)
> 					}
> 					listValues[string(value)] = struct{}{}
> ```

Doesn't this need to use key encoding? `1.0:::decimal` and `1.000:::decimal` are not distinct primary keys and so you shouldn't be able to use them as distinct values in a list encoding.

---

*[pkg/sql/sqlbase/structured.go, line 1328 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfP-NHWbsteZNVzhor:-KwfP-NHWbsteZNVzhos:bffqwih) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1328)):*
> ```Go
> 
> 				rangeValues[string(p.ValuesLessThan)] = struct{}{}
> 				if bytes.Compare(lastValue, p.ValuesLessThan) >= 0 {
> ```

Ditto here—doesn't this need to compare the key encodings to ensure that range split points sort correctly as used in primary keys?

---

*[pkg/sql/sqlbase/structured.go, line 1344 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfLV2kuiDIBkgT7EIJ:-KwfLV2kuiDIBkgT7EIK:b-ngx02k) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1344)):*
> ```Go
> 
> 	return desc.ForeachNonDropIndex(func(idxDesc *IndexDescriptor) error {
> 		if idxDesc.Partitioning.NumColumns > 0 {
> ```

Since you have to do this in `check` anyway, might as well remove this `if`.

---

*[pkg/sql/sqlbase/structured_test.go, line 1053 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfPCbN1IUvsMWS-SXd:-KwfPCbN1IUvsMWS-SXe:bddbm1p) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured_test.go#L1053)):*
> ```Go
> 			t.Errorf(`%d: got "%v" expected "%v"`, i, err, test.err)
> 		}
> 	}
> ```

Nice tests! I think we should add some collation/decimal nonsense to test the key/value encoding stuff too.

---

*[pkg/util/protoutil/clone.go, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfC_RbTiHH3Wr-JYge:-KwfC_RbTiHH3Wr-JYgf:bbyzhys) ([raw file](https://github.com/cockroachdb/cockroach/blob/890407d29f04a67d208fd5f83f9ceec46ac1c5e3/pkg/util/protoutil/clone.go#L86)):*
> ```Go
> 		// placeholder in here and immediately overwite it after
> 		// typeIsOrContainsVerbotenImpl returns.
> 		types.known[key] = nil
> ```

Warning: haven't thought about this at all. Someone who knows this code should look at this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

benesch:

---

*[pkg/sql/parser/sql.y, line 2671 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfR-9ioeGFJPyQKIwT:-KwfR-9ioeGFJPyQKIwU:b-57wd0x) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/sql.y#L2671)):*
> ```Yacc
> 
> list_partitions:
>   PARTITION name VALUES list_partition_values ',' list_partitions
> ```

I *believe* we can use `unrestricted_name` here, since partition names can't appear anywhere in the grammar besides immediately after a `PARTITION` keyword. Ditto everywhere below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

danhhz:
Thanks for the review @benesch! RFAL

---

Review status: 2 of 19 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[pkg/sql/create.go, line 1133 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfJHmei1l5xh7UvstH:-KwffsTM8hMKudXrNwr6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1133)):*

optional nit: `column(s)`, `value(s)`
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1165 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfGVBf-BKoYynN-9Kj:-KwfgXX2QSBnOZ8NWDGP:b-yfjr8l) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1165)):*

Is this not off-by-one? I feel like you want `>=` but maybe my mental model is wrong.
</blockquote></details>

you're totally right. i think i lost this test from our brainstorm somehow. fixed and tested

there's another similar comparison in `validatePartitioning`, but I think that one is correctly just `>`. wanna double check it tho?

---

*[pkg/sql/create.go, line 1176 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfHgULcpMUlLUw4Ac8:-KwfhQpQnG7fx5vlKM9k:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/create.go#L1176)):*

Seems unused?
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 14 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfD-2ciEhoXDvJB8lI:-Kwfhdy4WhwbLSKeaClL:b-wwgafb) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L14)):*

The equivalent interleaved error message is "interleaved columns must match parent". I find that to be less confusing wording, but I don't have a strong preference besides making them consistent.
</blockquote></details>

Happy to change it but there's no parent in partitioning. I'm operating at like 60% right now, so wanna walk me through exactly what you'd like the error to be : - )?

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 36 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfGOK4vO1OFOgtgr8C:-Kwfhrh1nPYP4cz0Z1S6:bek5qel) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L36)):*

Can you add a test for when the number of partition columns is greater than the number of columns in the primary key? (Unless I missed it!)
</blockquote></details>

whoops. done

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 105 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfDPZpjjwmlj7YMNRb:-Kwfie3edWr0gm3w_3Vo:bb7f2mf) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L105)):*

I'd add the parse tests examples here as `statement ok`s.
</blockquote></details>

done

---

*[pkg/sql/parser/create.go, line 643 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfAAZyb3QNpqJLuDXq:-KwfinDMTnFM_apfWims:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/create.go#L643)):*

Only one of `List` or `Range` can be set, right? Perhaps leave a comment.
</blockquote></details>

Done.

---

*[pkg/sql/parser/create.go, line 680 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfAM9s1dL2An_Pq3FZ:-KwfioA1MB27YCq3l8Ff:b-66agvc) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/create.go#L680)):*

I believe `Fprintf`ing like this is dangerous, because `node.Name` might need escaping. `Name` implements `NodeFormatter`, so you can call `FormatNode(..., &node.Name)` instead.
</blockquote></details>

whoops. it was originally a string and i changed it last minute. done

---

*[pkg/sql/parser/create.go, line 701 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfAdgrETLtovm_QmuC:-Kwfj3tcO3Jh46mJx7AK:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/create.go#L701)):*

Ditto here re: name formatting.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 2671 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfR-9ioeGFJPyQKIwT:-KwfjSldpZVTlTachTCA:b-lljrjl) ([raw file](https://github.com/cockroachdb/cockroach/blob/65b45c61c45da07b981d2ef5fc7b3c15d73e61b4/pkg/sql/parser/sql.y#L2671)):*

I *believe* we can use `unrestricted_name` here, since partition names can't appear anywhere in the grammar besides immediately after a `PARTITION` keyword. Ditto everywhere below.
</blockquote></details>

you appear to be correct, sir

---

*[pkg/sql/sqlbase/encoded_datum.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfK8XvgKTwxSi_Z1jR:-Kwfjia74wUxy168D9AU:b-m10x9v) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/encoded_datum.go#L98)):*

Seems weird that `EncDatumFromEncoded` panics and this returns an error. I feel like either `EncDatumFromEncoded` should return an error instead of panicking, or this check should be moved into the partitioning code.
</blockquote></details>

I definitely think EncDatumFromEncoded should return an error instead of panicing, but I don't particularly want to bite off that refactor in this PR. Disagree that this check should be moved into partitioning code.

---

*[pkg/sql/sqlbase/structured.go, line 1250 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfNZpTrFJNVs57xwf9:-KwflucIDS4YdqMImro0:b-ocktcm) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1250)):*

Since the column ID is going to be ignored, and using `encoding.NoColumnID` is just a space optimization, I have a weak preference for not validating it.
</blockquote></details>

the fix for your key vs value encoding point below make the decision for us. removed.

---

*[pkg/sql/sqlbase/structured.go, line 1293 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfOOTaDM9QC1a1KCvP:-KwfljKZbIDHwHrbk854:b-4ysnst) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1293)):*

Ah, interesting. This is just to verify that the encoding isn't corrupt?
</blockquote></details>

yeah, i was just being paranoid. tho, your comment below means i have to parse them anyway now

---

*[pkg/sql/sqlbase/structured.go, line 1297 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfOjyPYlgdJUkhIcQ9:-Kwfp90wGSSv_aSwBlyx:b-y7is4k) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1297)):*

Doesn't this need to use key encoding? `1.0:::decimal` and `1.000:::decimal` are not distinct primary keys and so you shouldn't be able to use them as distinct values in a list encoding.
</blockquote></details>

great catch. in my mental model, they were distinct for the purposes of indexing, but obviously that's wrong

---

*[pkg/sql/sqlbase/structured.go, line 1328 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfP-NHWbsteZNVzhor:-KwfpRN0Gc4ZqqqM3D9x:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1328)):*

Ditto here—doesn't this need to compare the key encodings to ensure that range split points sort correctly as used in primary keys?
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.go, line 1344 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfLV2kuiDIBkgT7EIJ:-KwfpYdMZm1hHjDIMwtV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured.go#L1344)):*

Since you have to do this in `check` anyway, might as well remove this `if`.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured_test.go, line 1053 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfPCbN1IUvsMWS-SXd:-KwftwjOIgmFmOUv_rRP:bxicwgz) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/sqlbase/structured_test.go#L1053)):*

Nice tests! I think we should add some collation/decimal nonsense to test the key/value encoding stuff too.
</blockquote></details>

I added them in the logic test, since they'd just be raw encoded bytes here

---

*[pkg/util/protoutil/clone.go, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfC_RbTiHH3Wr-JYge:-KwfuWf8qGhMqudiyDlO:b-x3sh6f) ([raw file](https://github.com/cockroachdb/cockroach/blob/890407d29f04a67d208fd5f83f9ceec46ac1c5e3/pkg/util/protoutil/clone.go#L86)):*

Warning: haven't thought about this at all. Someone who knows this code should look at this.
</blockquote></details>

tamir is the author and the only one that seems to have made any substantive changes, so it might be time to study up : - ). it's not complicated

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-:-Kwfyt7khXS-kzZ55bNz:b-lvzbv8)*

benesch:

---

*[pkg/util/protoutil/clone.go, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfC_RbTiHH3Wr-JYge:-KwhbGXgJ9xj_AjNJDJ6:blkfhbf) ([raw file](https://github.com/cockroachdb/cockroach/blob/890407d29f04a67d208fd5f83f9ceec46ac1c5e3/pkg/util/protoutil/clone.go#L86)):*

Drat. I was hoping Peter/Tobi's recent work on `protoutil` had introduced the bug. I'm pretty burned out on code reviews today, but will happily look tomorrow.
</blockquote></details>

Ok, this LGTM! Just to be sure, though, could you add a test to clone_test.go that verifies that something recursive *and* uncloneable panics? I'm thinking something like:

```
type foo struct {
  f *foo
  a []struct{}
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

danhhz:
Review status: 1 of 22 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 14 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfD-2ciEhoXDvJB8lI:-Kwjj6pgx14UqbAnb-Cu:b-1ha4iy) ([raw file](https://github.com/cockroachdb/cockroach/blob/410dd8486effc0b4fc58f8d6739f734ed89b3bbd/pkg/sql/logictest/testdata/logic_test/partitioning#L14)):*

I guess it's just the "be a prefix of" wording that I find confusing. What about "declared partition columns must match index being partitioned"? 
</blockquote></details>

Done. Agree that subpartitioning confuses the prefix part

---

*[pkg/sql/sqlbase/structured.go, line 1259 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwg6XJL9QWhxucuaJl2:-KwjjNAOFaWUq0L9m7Rj:bclrt8l) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4be8f2dadb579912fab4a1b7fd8b557e5e7f097/pkg/sql/sqlbase/structured.go#L1259)):*

Heads up: 28fecfdc7 changed the signature of this function. (GitHub is complaining about conflict that would have pointed you to this, but thought I'd flag the commit because I just went digging myself.)
</blockquote></details>

Yup, I was waiting to address most of your comments so reviewable didn't get confused, but this one really isn't so bad.

---

*[pkg/util/protoutil/clone.go, line 86 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwfC_RbTiHH3Wr-JYge:-Kwjog4msCYCDNitSnyq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/890407d29f04a67d208fd5f83f9ceec46ac1c5e3/pkg/util/protoutil/clone.go#L86)):*

Ok, this LGTM! Just to be sure, though, could you add a test to clone_test.go that verifies that something recursive *and* uncloneable panics? I'm thinking something like:

```
type foo struct {
  f *foo
  a []struct{}
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

knz:
Please accept my apologies in my delay reviewing this PR. 
There's a lot of good work in there. 

I have only two noteworthy concerns remaining:

- some notes about the grammar, in comments below.

- the observation that the handling of expression in the PARTITION clause is performed only in MakeTableDesc. Doing so prevents the code from supporting placeholders in any way. I think (as Nikhil justly observed) this may be overly restrictive - compared to CHECK and DEFAULT expressions, which semantically must only depend on the rest of the schema, the  partition keys are independent from the schema and could come from the context, including placeholders. However it's not a big restriction -- the code, as is, simply doesn't support placeholders, and we can add that support later.

---

Reviewed 1 of 7 files at r5, 21 of 21 files at r7, 6 of 6 files at r8, 7 of 7 files at r9, 4 of 6 files at r10.
Review status: 20 of 22 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[pkg/sql/create.go, line 1119 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjswYRXlrBAAou8_7t:-KwjswYRXlrBAAou8_7u:b-xbhpdj) ([raw file](https://github.com/cockroachdb/cockroach/blob/a27e72d3659799a53c4c05f43a129313cde31753/pkg/sql/create.go#L1119)):*
> ```Go
> }
> 
> func valueEncodeTuple(
> ```


This reviewer would welcome a short comment that explains what this function does.

---

*[pkg/sql/create.go, line 1151 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjt-w7mztCZJDXzNpk:-Kwjt-w7mztCZJDXzNpl:b-rqxvo7) ([raw file](https://github.com/cockroachdb/cockroach/blob/a27e72d3659799a53c4c05f43a129313cde31753/pkg/sql/create.go#L1151)):*
> ```Go
> }
> 
> func addPartitionedBy(
> ```

ditto.

---

*[pkg/sql/parser/sql.y, line 2670 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqMRcT6fwr_75tjJ6:-KwjpzJPJaBKEJPUDWbv:b-sqm4dl) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2670)):*
> ```Yacc
>   }
> 
> list_partitions:
> ```

Lists in yacc can be left recursive.

This avoids re-creating the array at every iteration, i.e. use `append($1.listPartitions(), Partition{...})` instead.

If there's an ambiguity otherwise (I suspect you wrote it this way because there is an ambiguity), that might warrant updating your spec with extra punctuation to disambiguate. Experimentally I have found that this specific kind of grammar hurdle can also become a problem for the human understanding.

---

*[pkg/sql/parser/sql.y, line 2671 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjpo2Bc6vnA69HSxb8:-Kwjpo2CRxjITCvhCRaR:bohz4mg) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2671)):*
> ```Yacc
> 
> list_partitions:
>   PARTITION unrestricted_name VALUES list_partition_values ',' list_partitions
> ```

I'm a bit uncomfortable about requiring the PARTITION keyword again here, given that it is already mandatory above (PARTITION  BY...)
What else can there be in a list_partitions or range_partitions other than a partition?

---

*[pkg/sql/parser/sql.y, line 2706 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqwMMVFnLcHtqEBjJ:-KwjqwMMVFnLcHtqEBjK:b-anba20) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2706)):*
> ```Yacc
> 
> range_partitions:
>   PARTITION unrestricted_name VALUES LESS THAN ctext_row ',' range_partitions
> ```

ditto repetition of the keyword

Also the choice of phrase "LESS THAN" is a bit unfortunate too. The problem here is that the English word doesn't tell the reader of the SQL code whether the inequality is strict or not, and forces them to refer to the documentation. Any reason to use "LESS THAN" here instead of more simply the punctuation "<=" ?

---

*[pkg/sql/sqlbase/structured.go, line 1229 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjsL1tECHmSE_ZzS6C:-KwjsL1uu6fXvW_Dyvwn:b5chf6o) ([raw file](https://github.com/cockroachdb/cockroach/blob/a27e72d3659799a53c4c05f43a129313cde31753/pkg/sql/sqlbase/structured.go#L1229)):*
> ```Go
> }
> 
> func (desc *TableDescriptor) validatePartitioning() error {
> ```

There's a comment missing here that explains what this function does, at a high level.

The comment would help the reviewer (e.g. me) to compare the intent (in the comment) with the execution (the code), and verify that you actually implemented what you wanted.

---

*[pkg/sql/sqlbase/structured.proto, line 204 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjs3d3DQn8L4gmIGIh:-Kwjs3d3DQn8L4gmIGIi:bqnlrb5) ([raw file](https://github.com/cockroachdb/cockroach/blob/cf1c61251eab8084a88be4afc3bf0a3ff0f784a4/pkg/sql/sqlbase/structured.proto#L204)):*
> ```protobuf
>     // ValuesLessThan is exclusive upper bound of this range partition. It is
>     // encoded with the EncDatum value encoding.
>     optional bytes values_less_than = 2;
> ```

I would be much more comfortable as a reader of this code if this field was more simply called "upper_bound".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-:-KwjtP8m6YoKU5IqCFvM:b-3tvd1s)*

knz:
Although for good measure, please check with a pgwire test that you get an error if you try to use placeholders, as opposed to a panic :)

---

Review status: 20 of 22 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-:-KwjuIV5H1SiBrPIlQdV:bm3p4g0)*

danhhz:
Thanks for the review @knz!

As mentioned below, I'd like to leave discussion of specific syntax to the rfc if that's possible.

I would indeed like to leave placeholders for a followup. I agree that they should be supported and I'll file an issue to make sure this happens in 1.2. Getting this in now (even if it's not perfect) will help a lot in making it possible for @benesch and I to split up the development work. They were panicing but I was able to turn it into an error without going into the weeds too much

---

Review status: 3 of 22 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/sql/create.go, line 1119 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjswYRXlrBAAou8_7t:-KwkPGLXg8HJrizucrGm:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a27e72d3659799a53c4c05f43a129313cde31753/pkg/sql/create.go#L1119)):*

This reviewer would welcome a short comment that explains what this function does.
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1151 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjt-w7mztCZJDXzNpk:-KwkPGyUmcvyn-t6E7ll:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a27e72d3659799a53c4c05f43a129313cde31753/pkg/sql/create.go#L1151)):*

ditto.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 2670 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqMRcT6fwr_75tjJ6:-KwkS3jozqZ29AVI6wmt:b-56678x) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2670)):*

Lists in yacc can be left recursive.

This avoids re-creating the array at every iteration, i.e. use `append($1.listPartitions(), Partition{...})` instead.

If there's an ambiguity otherwise (I suspect you wrote it this way because there is an ambiguity), that might warrant updating your spec with extra punctuation to disambiguate. Experimentally I have found that this specific kind of grammar hurdle can also become a problem for the human understanding.
</blockquote></details>

I just tried this and indeed there is a shift/reduce conflict. Does "updating your spec with extra punctuation" mean throw some parens around it? If so, ditto comment below about leaving syntax change discussions for the rfc. Otherwise, can you be more specific?

---

*[pkg/sql/parser/sql.y, line 2671 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjpo2Bc6vnA69HSxb8:-KwkPxHaE_dwSyUjZ-GW:bin0625) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2671)):*

I'm a bit uncomfortable about requiring the PARTITION keyword again here, given that it is already mandatory above (PARTITION  BY...)
What else can there be in a list_partitions or range_partitions other than a partition?


</blockquote></details>

This matches what is currently presented in the rfc (https://github.com/cockroachdb/cockroach/pull/18683). I'd like to keep any discussion of the syntax there, if that's okay.

In the meantime, it would help @benesch and I a lot if we can merge this PR before that's all finalized. Can I convince you to let this in as-is if I promise to update the grammar in a followup PR to match whatever is in the RFC?

---

*[pkg/sql/parser/sql.y, line 2706 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqwMMVFnLcHtqEBjJ:-KwkRWb3ysoto-XQudWq:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2706)):*

ditto repetition of the keyword

Also the choice of phrase "LESS THAN" is a bit unfortunate too. The problem here is that the English word doesn't tell the reader of the SQL code whether the inequality is strict or not, and forces them to refer to the documentation. Any reason to use "LESS THAN" here instead of more simply the punctuation "<=" ?
</blockquote></details>

ditto

---

*[pkg/sql/sqlbase/structured.go, line 1229 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjsL1tECHmSE_ZzS6C:-KwkU98jmGu_MKX0N3A4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a27e72d3659799a53c4c05f43a129313cde31753/pkg/sql/sqlbase/structured.go#L1229)):*

There's a comment missing here that explains what this function does, at a high level.

The comment would help the reviewer (e.g. me) to compare the intent (in the comment) with the execution (the code), and verify that you actually implemented what you wanted.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.proto, line 204 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjs3d3DQn8L4gmIGIh:-KwkUakb8hrNw9Jr9cx2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/cf1c61251eab8084a88be4afc3bf0a3ff0f784a4/pkg/sql/sqlbase/structured.proto#L204)):*

I would be much more comfortable as a reader of this code if this field was more simply called "upper_bound".
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-:-KwkYGBr6mYcBjt10pRf:byh636v)*

knz:
Reviewed 3 of 19 files at r11, 6 of 6 files at r12, 7 of 7 files at r13, 6 of 6 files at r14.

---

*[pkg/sql/parser/sql.y, line 2670 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqMRcT6fwr_75tjJ6:-Kwnrcyrb7ETpZvOZ28X:b-u2bxgm) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2670)):*

I just tried this and indeed there is a shift/reduce conflict. Does "updating your spec with extra punctuation" mean throw some parens around it? If so, ditto comment below about leaving syntax change discussions for the rfc. Otherwise, can you be more specific?
</blockquote></details>

So yeah I tried what I meant, and I discovered the grammar is indeed problematic:

```yacc
list_partitions:
  list_partitions_item
| list_partitions ',' list_partitions_item { $$.val = append($1.listPartitions(), $2.listPartition()) }

list_partitions_item:
  PARTITION unrestricted_name VALUES list_partition_values opt_partition_by {...}
```

The issue is explained by yacc:
```
2585: shift/reduce conflict (shift 2589(0), red'n 485(1)) on ','
state 2585
        list_partitions_item:  PARTITION unrestricted_name VALUES list_partition_values.    (485)
        list_partitions_item:  PARTITION unrestricted_name VALUES list_partition_values.partition_by
        list_partition_values:  list_partition_values.',' ctext_row

        PARTITION  shift 2312
        ','  shift 2589
        .  reduce 485 (src line 2670)

        partition_by  goto 2588
```

In other words, look at the following example:

```sql
  PARTITION p1 VALUES (1), (2)  , (3), ...   -- more values
  PARTITION p1 VALUES (1), (2)  , PARTITION  -- more partition list items
                               ^ here
```

yacc sees a comma and has to decide whether to continue a list_partition_values,  or complete the current list_partition_item  and move on to another. However, it can only decide this really by observing the token *after* the comma. The problem is that LALR only has one token of lookahead.

Massaging the grammar like you did is not only inefficient in the action, it also makes the grammar fragile with further updates: for each additional optional item we may want to add in the future, we'll need to double (!) the number of rules to list them all explicitly side-by-side, to disambiguate.

To my eyes, I see this as a reason to tweak the proposed grammar to avoid the problem altogether. I'll make a proposal on the RFC.

---

*[pkg/sql/parser/sql.y, line 2671 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjpo2Bc6vnA69HSxb8:-KwniorLASsc9N8Yv7eR:b63m1ko) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2671)):*

This matches what is currently presented in the rfc (https://github.com/cockroachdb/cockroach/pull/18683). I'd like to keep any discussion of the syntax there, if that's okay.

In the meantime, it would help @benesch and I a lot if we can merge this PR before that's all finalized. Can I convince you to let this in as-is if I promise to update the grammar in a followup PR to match whatever is in the RFC?
</blockquote></details>

:+1:
However if/when/after we do that it would do good to the RFC to highlight there how the syntax is further motivated by grammar restrictions / tweaks.

---

*[pkg/sql/sqlbase/structured.go, line 1238 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwnwEDqACHF8-FgVwkd:-KwnwEDr-mr5pdohWn8P:bg4e8mv) ([raw file](https://github.com/cockroachdb/cockroach/blob/f57805c7eb33dab7d9ea9206097a1e2fcc849887/pkg/sql/sqlbase/structured.go#L1238)):*
> ```Go
> 
> 	a := &DatumAlloc{}
> 	translateValueEncodingToKey := func(
> ```

As we recently learned the Go compiler doesn't know how to extract this code into a static function, even though `a` which it captures is only scoped to `validatePartitioning`. 
As such I think it's better to extract this closure into its own separate function (with a comment!) and give it `a` as explicit argument.

---

*[pkg/sql/sqlbase/structured.go, line 1269 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwnwd1KRL6qvvYTO-AN:-Kwnwd1KRL6qvvYTO-AO:b-2wuxh2) ([raw file](https://github.com/cockroachdb/cockroach/blob/f57805c7eb33dab7d9ea9206097a1e2fcc849887/pkg/sql/sqlbase/structured.go#L1269)):*
> ```Go
> 
> 	var check func(idxDesc *IndexDescriptor, partDesc *PartitioningDescriptor, colOffset int) error
> 	check = func(idxDesc *IndexDescriptor, partDesc *PartitioningDescriptor, colOffset int) error {
> ```

ditto, with `partitionNames`. Also this closure is too large anyway to deserve being written in the body of the calling function -- it impairs readability.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

knz:

---

*[pkg/sql/parser/parse_test.go, line 147 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwo3tnzvciyRGn2c7ZY:-Kwo3to-YYzi0j5nnEJ3:b-3dhqjk) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa5397570e3bfb4c8882a73779066683c8950a25/pkg/sql/parser/parse_test.go#L147)):*
> ```Go
> 		{`CREATE TABLE IF NOT EXISTS a (b INT)`},
> 
> 		{`CREATE TABLE a (b INT) PARTITION BY LIST (b) (PARTITION p1 VALUES (1, DEFAULT), PARTITION p2 VALUES (2))`},
> ```

You're missing tests when there are more than 1 tuple after the VALUES keyword.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

knz:

---

*[pkg/sql/parser/sql.y, line 2670 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqMRcT6fwr_75tjJ6:-Kwo4d1YDbQsCknLL5qq:b2p1svy) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2670)):*

So yeah I tried what I meant, and I discovered the grammar is indeed problematic:

```yacc
list_partitions:
  list_partitions_item
| list_partitions ',' list_partitions_item { $$.val = append($1.listPartitions(), $2.listPartition()) }

list_partitions_item:
  PARTITION unrestricted_name VALUES list_partition_values opt_partition_by {...}
```

The issue is explained by yacc:
```
2585: shift/reduce conflict (shift 2589(0), red'n 485(1)) on ','
state 2585
        list_partitions_item:  PARTITION unrestricted_name VALUES list_partition_values.    (485)
        list_partitions_item:  PARTITION unrestricted_name VALUES list_partition_values.partition_by
        list_partition_values:  list_partition_values.',' ctext_row

        PARTITION  shift 2312
        ','  shift 2589
        .  reduce 485 (src line 2670)

        partition_by  goto 2588
```

In other words, look at the following example:

```sql
  PARTITION p1 VALUES (1), (2)  , (3), ...   -- more values
  PARTITION p1 VALUES (1), (2)  , PARTITION  -- more partition list items
                               ^ here
```

yacc sees a comma and has to decide whether to continue a list_partition_values,  or complete the current list_partition_item  and move on to another. However, it can only decide this really by observing the token *after* the comma. The problem is that LALR only has one token of lookahead.

Massaging the grammar like you did is not only inefficient in the action, it also makes the grammar fragile with further updates: for each additional optional item we may want to add in the future, we'll need to double (!) the number of rules to list them all explicitly side-by-side, to disambiguate.

To my eyes, I see this as a reason to tweak the proposed grammar to avoid the problem altogether. I'll make a proposal on the RFC.
</blockquote></details>

Okay here are two patches on top of yours:

- one where the tuples after VALUES have another delimiter than a comma (space in this example): https://github.com/knz/cockroach/commit/01af4830293fb629520cbdc2d704da7e2cdd1b49 - see discussion on RFC

- one where the scanner (!) is modified to look ahead of the comma to disambiguate: https://github.com/knz/cockroach/commit/a86a8cf149c8f088eb88bf93cc427177b50ce409 - I don't like this; lookahead code is hard to debug when it goes wrong.

---

*[pkg/sql/parser/sql.y, line 2671 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwjpo2Bc6vnA69HSxb8:-Kwo4SwunedbgAGw0cNQ:b-3cie6l) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2671)):*

:+1:
However if/when/after we do that it would do good to the RFC to highlight there how the syntax is further motivated by grammar restrictions / tweaks.
</blockquote></details>

I changed my mind on this. The RFC properly motivates the choice. Thanks for reminding me to look at the RFC :)

---

*[pkg/sql/parser/sql.y, line 2706 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqwMMVFnLcHtqEBjJ:-Kwo4_9ZidRaf02AtWHA:bqgea1n) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2706)):*

ditto
</blockquote></details>

Changed my mind here too. :+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

benesch:

---

*[pkg/util/protoutil/clone.proto, line 17 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwp1oUJMq7sMn7NIHWY:-Kwp1oUJMq7sMn7NIHWZ:b-cp8779) ([raw file](https://github.com/cockroachdb/cockroach/blob/dfd290b2d317d2a8d3c74fc2d061bde469397758/pkg/util/protoutil/clone.proto#L17)):*
> ```protobuf
> syntax = "proto3";
> package cockroach.util.protoutil;
> option go_package = "protoutil";
> ```

Thought: can this go in `protoutil_test`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315)*

danhhz:
Addressed all the non-syntax comments, which are being left for discussion in the rfc. @knz, is this okay to merge? I'd really like to get it in today

---

Review status: 8 of 22 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/parser/parse_test.go, line 147 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwo3tnzvciyRGn2c7ZY:-Kwp-0HINX-dHX4XxG7I:b-2jvje4) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa5397570e3bfb4c8882a73779066683c8950a25/pkg/sql/parser/parse_test.go#L147)):*

You're missing tests when there are more than 1 tuple after the VALUES keyword.
</blockquote></details>

Indeed I am! Done.

---

*[pkg/sql/parser/sql.y, line 2670 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwjqMRcT6fwr_75tjJ6:-Kwp-JzsNsFStkFSmaxD:b5xpx10) ([raw file](https://github.com/cockroachdb/cockroach/blob/72326d70be8e563332c800aad39b7f2d7f1aad12/pkg/sql/parser/sql.y#L2670)):*

Okay here are two patches on top of yours:

- one where the tuples after VALUES have another delimiter than a comma (space in this example): https://github.com/knz/cockroach/commit/01af4830293fb629520cbdc2d704da7e2cdd1b49 - see discussion on RFC

- one where the scanner (!) is modified to look ahead of the comma to disambiguate: https://github.com/knz/cockroach/commit/a86a8cf149c8f088eb88bf93cc427177b50ce409 - I don't like this; lookahead code is hard to debug when it goes wrong.


</blockquote></details>

Sounds good! I cross posted those two examples to the RFC

---

*[pkg/sql/sqlbase/structured.go, line 1238 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-KwnwEDqACHF8-FgVwkd:-Kwp10ykRpU-yxePyua8:b-cr6uag) ([raw file](https://github.com/cockroachdb/cockroach/blob/f57805c7eb33dab7d9ea9206097a1e2fcc849887/pkg/sql/sqlbase/structured.go#L1238)):*

As we recently learned the Go compiler doesn't know how to extract this code into a static function, even though `a` which it captures is only scoped to `validatePartitioning`. 
As such I think it's better to extract this closure into its own separate function (with a comment!) and give it `a` as explicit argument.
</blockquote></details>

I don't buy the performance argument here. There's all sorts of other allocations that we don't yet care about in this codepath. However you're making a readability argument below and that I'm willing to address. Extracted and commented

---

*[pkg/sql/sqlbase/structured.go, line 1269 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwnwd1KRL6qvvYTO-AN:-Kwp2FIUcamQ-DMKwOP7:bhwunie) ([raw file](https://github.com/cockroachdb/cockroach/blob/f57805c7eb33dab7d9ea9206097a1e2fcc849887/pkg/sql/sqlbase/structured.go#L1269)):*

ditto, with `partitionNames`. Also this closure is too large anyway to deserve being written in the body of the calling function -- it impairs readability.
</blockquote></details>

I personally find this _easier_ to read, but if you don't, that's good enough for me.

---

*[pkg/util/protoutil/clone.proto, line 17 at r11](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-Kwp1oUJMq7sMn7NIHWY:-Kwp4Tt1bQ7Moi2RPjdG:bio45nk) ([raw file](https://github.com/cockroachdb/cockroach/blob/dfd290b2d317d2a8d3c74fc2d061bde469397758/pkg/util/protoutil/clone.proto#L17)):*

Thought: can this go in `protoutil_test`?
</blockquote></details>

`found packages protoutil (clone.go) and protoutil_test (clone.pb.go)` : - /

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19315#-:-Kwp5YWI_ji-gmfXWcvt:b-aubki9)*

danhhz:
@knz It's seems like you're more or less on board with this (modulo syntax) and it'd be real nice to be done resolving merge conflicts on this (which also makes it hard to do that work in a branch), so I'm planning to merge on green. If you have any other comments, please do leave them and I'll address them in a follow up

knz:
So what do you propose to do with the syntax problem then?

danhhz:
Resolve it in the rfc and then update the code to match

knz:
Oh I see. Thanks! Carry on.

LGTM

danhhz:
Thanks! And please do let me know if there was anything else you wanted to see here. Happy to resolve those in followups too

PR #19319, opened 2017-10-17
build: Make sure to base RHEL Docker images off the latest base

a-robinson:
To make sure that we're always using the latest base image, particularly
when doing rebuilds to handle security issues in the base image.

### DIFF
--- a/build/deploy/rhel/Dockerfile
+++ b/build/deploy/rhel/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/rhel-atomic
+FROM registry.access.redhat.com/rhel-atomic:latest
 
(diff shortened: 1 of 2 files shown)


mjibson:
I can't see in the [dockerfile docs](https://docs.docker.com/engine/reference/builder/#from) where it says that using latest (or probably specifying a tag) will guarantee an uncached pull (which is what I think this PR is trying to achieve). Can you link me some docs about it?

benesch:
Agreed—are you sure you don't need `docker build --pull`? https://github.com/moby/moby/issues/4238#issuecomment-54322081

a-robinson:
Done, thanks.

madelynnblue (approved):

Issue #18500, opened 2017-09-15
sql+cli: dump should work with PKs named other than "primary"


jtescher:
It looks like the primary key constraint name has to be "primary" for sql dump to work, however schema migration tools often have their own naming conventions e.g. liquibase does "pk_[tablename]" so this could be a somewhat common situation.

Steps to reproduce:

```sql
CREATE DATABASE liquibase;
CREATE TABLE liquibase.episodes (id INT NOT NULL, season INT NULL, num INT NULL, title STRING NULL, stardate DECIMAL NULL, CONSTRAINT "pk_episodes" PRIMARY KEY (id), FAMILY "primary" (id, season, num), FAMILY fam_1_title (title), FAMILY fam_2_stardate (stardate));

CREATE DATABASE standard;
CREATE TABLE standard.episodes (id INT NOT NULL, season INT NULL, num INT NULL, title STRING NULL, stardate DECIMAL NULL, CONSTRAINT "primary" PRIMARY KEY (id), FAMILY "primary" (id, season, num), FAMILY fam_1_title (title), FAMILY fam_2_stardate (stardate));
```

Running `./cockroach dump liquibase --insecure` will now error with:

```
Error: pq: column name "rowid" not found
```

Even though `./cockroach dump standard --insecure` works as expected

(client and server version: v1.1-beta.20170907)

mjibson:
I wasn't able to reproduce this:

```
root@:26257/> CREATE DATABASE liquibase;
CREATE DATABASE

Time: 31.434091ms

root@:26257/> CREATE TABLE liquibase.episodes (id INT NOT NULL, season INT NULL, num INT NULL, title STRING NULL, stardate DECIMAL NULL, CONSTRAINT "pk_episodes" PRIMARY KEY (id), FAMILY "primary" (id, season, num), FAMILY fam_1_title (title), FAMILY fam_2_stardate (stardate));CREATE TABLE

Time: 59.355878ms

root@:26257/> 
root@:26257/> CREATE DATABASE standard;
CREATE DATABASE

Time: 64.314378ms

root@:26257/> CREATE TABLE standard.episodes (id INT NOT NULL, season INT NULL, num INT NULL, title STRING NULL, stardate DECIMAL NULL, CONSTRAINT "primary" PRIMARY KEY (id), FAMILY "primary" (id, season, num), FAMILY fam_1_title (title), FAMILY fam_2_stardate (stardate));
CREATE TABLE

Time: 40.491742ms
```

[...]

PR #19325, opened 2017-10-17
cli: test dump with non-default PK name

madelynnblue:
Although this test is already passing, it is being added to master
so we don't regress and so it can be backported to 1.1, where it
is failing.

See #18500

After this merges, this PR and #18472 will be cherry picked to 1.1.


knz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19325#-:-Kwir338yOa0Cg_5L6j5:bnfp4nl)*

PR #19326, opened 2017-10-17
 sqlccl: error if file:// is used on multi-node

### LINKED ISSUE #19158: backup: prevent mistaken use of `nodelocal`
Some feedback on `nodelocal` from external users and even internal users -- engineers and PMs who are familiar with using a distributed database -- have made it clear that `nodelocal` is confusing.

 In a distributed cluster, what does "write this thing to the local disk" even mean? Clearly no one is sure.

Example, from the QA pass on EXPORT:

> * Problems with providing nodelocal paths with different (possibly invalid?) numbers of slashes (confusing trying to map to/from the template of `[scheme]://[host]/[path]?[parameters]` in the  docs -- if everything goes into the external I/O dir, why is the starting slash even needed?):
> ```
> root@:26257/defaultdb> EXPORT INTO CSV 'nodelocal:' FROM SELECT * from startrek.quotes;
> pq: Local storage requested but path not provided
> root@:26257/defaultdb> EXPORT INTO CSV 'nodelocal:/tmp/csv1' FROM SELECT * from startrek.quotes;
> +----------+------+-------+
> | filename | rows | bytes |
> +----------+------+-------+
> | n1.0.csv |  200 | 20918 |
> +----------+------+-------+
> root@:26257/defaultdb> EXPORT INTO CSV 'nodelocal://tmp/csv1' FROM SELECT * from startrek.quotes;
> pq: nodelocal does not support hosts: nodelocal://tmp/csv1
> root@:26257/defaultdb> EXPORT INTO CSV 'nodelocal:///tmp/csv1' FROM SELECT * from startrek.quotes;
> +----------+------+-------+
> | filename | rows | bytes |
> +----------+------+-------+
> | n1.0.csv |  200 | 20918 |
> +----------+------+-------+
> ```
> * For nodelocal mode, it’s really unintuitive where the file goes.
>   * I think the docs need to more strongly call out what’s needed for NFS/nodelocal to make sense. Using it without setting the `--external-io-dir` flag is basically nonsensical even if all the nodes are on the same machine, since the resulting file location(s) can't be known in advance.
>   * I got pretty confused at first trying to find the output file, not realizing it was written into the leaseholder’s store directory (not the absolute filesystem path, or on gateway node, or on the sql shell’s node).
> * Parameters that aren’t recognized should be an error. Instead, they’re silently ignored
> ```
> root@:26257/defaultdb> EXPORT INTO CSV 'nodelocal:///tmp/csv1?AWS_ACCESS_KEY_ID=foo' FROM SELECT * from startrek.episodes;
> +----------+------+-------+
> | filename | rows | bytes |
> +----------+------+-------+
> | n3.0.csv |   79 |  3021 |
> +----------+------+-------+
> root@:26257/defaultdb> EXPORT INTO CSV 'nodelocal:///tmp/csv1?foo=bar' FROM SELECT * from startrek.episodes;
> +----------+------+-------+
> | filename | rows | bytes |
> +----------+------+-------+
> | n3.0.csv |   79 |  3021 |
> +----------+------+-------+
> ```

## Feature

Ensure that BACKUP destination is the same across all nodes (since non-cloud paths like `nodelocal://` or even `http://` may point at local paths).

## backstory
A recurring source of confusion is the local file storage for backups/imports/etc.

`nodelocal` is the simplest and easiest backup location in a single-node cluster.

In multi-node clusters though, if a given node writes a file to a path on its local disk, that isn't particularly useful to the other nodes if/when they need that file, unless that path happens to be backed by something like NFS or another network file system. So far though, this seems to be easy to overlook, and both internal testers and external customers have run into this.

After initial internal users hit it, we changed the scheme from `file://` (or no scheme at all, and just allowing e.g. `BACKUP foo.* to '/mnt/foo'`) to be `nodelocal`  in the hopes that it would make it a little more clear what it did, or at least that its unfamiliar name would give a user pause and lead them to look at the docs where we added callout box that said much or the above.

However it seems we have not adequately addressed the potential pitfall, since we are still encountering internal and external users hitting this. We can continue to make the docs warning more explicit, but some users may not read it (if they see an existing example and crib it, forget the warning, etc).

The scariest part of this is that a user might run a `BACKUP ... nodelocal://` that completes successfully and think their data is backed up and later lose data because their backup isn't usable.  Ideally we should make this mistake harder to make.

One option would be to return an error, rather than writing a (likely useless backup) if `nodelocal` is used on a multi-node cluster. This would catch that mistake right away and make it clear what's wrong (versus an opaque "missing file" error at restore), but breaks the legitimate (though maybe uncommon) use case of an actual network-backed filesystem. We could add a URI flag like `?fs=nfs` to skip the error, or we could add an alias for `nodelocal` like `nfs` that does the same.

dt:
This re-introduces common the file:// uri scheme for export storage.

It was previously renamed to “nodelocal” in an attempt to emphasize that it was local to each node an unusable in distributed clusters, but it seems that that was not sufficiently. Instead of relying simply on an odd name, this makes it an error to use it on multi-node clusters, at which point we may as well use the commonly accepted scheme “file://” (“nodelocal” is kept as an alias for compatibility).

In cases where a user really, truly wants to use local file:// paths in a multi-node cluster, e.g. if those “local” paths are actually NFS mounts, the alternative scheme “nfs://“ can be used instead, to skip the node-count check.

Also allow for a clean and simple “BACKUP foo TO /path/to/foo” scheme-less local-path URIs.

Fixes #19158.


benesch:
Reviewed 11 of 11 files at r1, 2 of 2 files at r2.

---

*[pkg/ccl/sqlccl/backup.go, line 638 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfgWMGY9YLC4aC76a7:-KwfgWMGY9YLC4aC76a8:b-thtub5) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a87462504736e815cb5f5f8d9d0f2d4411e0fc1/pkg/ccl/sqlccl/backup.go#L638)):*
> ```Go
> 		defer exportStore.Close()
> 
> 		nodeCount := clusterNodeCount(p.ExecCfg().Gossip)
> ```

Might as well move this down into the first `if` block to avoid fetching the node count if not using `file://` storage.

---

*[pkg/ccl/sqlccl/backup_test.go, line 245 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfgBYjssD5Uytx6MJf:-KwfgBYjssD5Uytx6MJg:b-uf49sr) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a87462504736e815cb5f5f8d9d0f2d4411e0fc1/pkg/ccl/sqlccl/backup_test.go#L245)):*
> ```Go
> 		_, dir, _, sqlDB, cleanupFn := backupRestoreTestSetup(t, singleNode, numAccounts, initNone)
> 		defer cleanupFn()
> 		dir = strings.TrimPrefix(dir, "nfs:/")
> ```

~optional nit: including the `/` weirds me out. Consider trimming `nfs:` or `nfs` instead, and adjusting the prefixes beneath accordingly.~

Never mind; I hadn't reviewed the following commit.

---

*[pkg/ccl/sqlccl/backup_test.go, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwfh-b12N_R9AIJoQu2:-Kwfh-b12N_R9AIJoQu3:b-x0m8i3) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4ab9d0f91be53940a8efea339683291020b279f/pkg/ccl/sqlccl/backup_test.go#L247)):*
> ```Go
> 		dir = strings.TrimPrefix(dir, "nfs:/")
> 
> 		for _, scheme := range []string{"", "file:/", "nodelocal:/", "nfs:/"} {
> ```

We should test or disallow relative paths. (I imagine they're confusing and not useful, since they're relative to wherever the heck the server's CWD is.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:

---

*[pkg/ccl/sqlccl/backup_test.go, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwfh-b12N_R9AIJoQu2:-KwfnQfclmWYQZEl3fkb:bra7c7q) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4ab9d0f91be53940a8efea339683291020b279f/pkg/ccl/sqlccl/backup_test.go#L247)):*

We should test or disallow relative paths. (I imagine they're confusing and not useful, since they're relative to wherever the heck the server's CWD is.)
</blockquote></details>

Note that I did some specific work to enable relative paths: https://github.com/cockroachdb/cockroach/commit/6a3b7a087f027c0d310f60a25fa84da6ac202bc2

However that was for when `cockroach load csv` was useful before IMPORT was working. I think that since we are now discouraging use of load csv I'm not opposed to removing cwd support. Or that support could be moved explicitly to the CLI command where it is expected.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
at @mjibson's request, I added another commit that pulled this check into the sqlccl helper that combines URI->conf and conf->ExportStore, and I switched all sqlccl callsites to use said helper. RFAL.

benesch:
Reviewed 4 of 4 files at r3.

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-KwfrlIeSN5R06lJBkr0:b-wkofd6) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*
> ```Go
> 		dataFiles = []string{fmt.Sprintf("%s.dat", table)}
> 	}
> 	createTable, err := readCreateTableFromStore(ctx, table, cluster.NoSettings, 0)
> ```

Might as well pull out a named constant here, too, since you did it everywhere else. Or extract a global named constant like `sqlccl.IrrelevantClusterSize`or something?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:

---

*[pkg/ccl/sqlccl/backup_test.go, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwfh-b12N_R9AIJoQu2:-Kwfs7ayC7qTm4WoKuLN:bzec8t2) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4ab9d0f91be53940a8efea339683291020b279f/pkg/ccl/sqlccl/backup_test.go#L247)):*

Note that I did some specific work to enable relative paths: https://github.com/cockroachdb/cockroach/commit/6a3b7a087f027c0d310f60a25fa84da6ac202bc2

However that was for when `cockroach load csv` was useful before IMPORT was working. I think that since we are now discouraging use of load csv I'm not opposed to removing cwd support. Or that support could be moved explicitly to the CLI command where it is expected.
</blockquote></details>

As a followup because I hadn't everything from before: I agree that the export storage should disallow relative paths. The CLI appears to already convert relative to absolute.

---

*[pkg/ccl/sqlccl/csv.go, line 273 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfteYVwH49mBNcipLW:-KwfteYVwH49mBNcipLX:bjgb7ya) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L273)):*
> ```Go
> 
> func readCreateTableFromStore(
> 	ctx context.Context, filename string, settings *cluster.Settings, clusterSize int,
> ```

Should we make a new struct type holding settings and cluster size? They are plumbed around together in a bunch of places now.

---

*[pkg/ccl/storageccl/export_storage.go, line 213 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfuM0fN4YBBRN1WaNh:-KwfuM0fN4YBBRN1WaNi:bhv42rb) ([raw file](https://github.com/cockroachdb/cockroach/blob/19f1914fdd72de5dc5f3f2790916d429f4fdeda9/pkg/ccl/storageccl/export_storage.go#L213)):*
> ```Go
> 		return "", err
> 	}
> 	return fmt.Sprintf("nodelocal://%s", path), nil
> ```

This should probably stay the same. This function should only be used during CLI operations, and so should never trigger the cluster size error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 273 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfteYVwH49mBNcipLW:-KwfvPfDxq2e2SfskvxF:bc5jmmk) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L273)):*

Should we make a new struct type holding settings and cluster size? They are plumbed around together in a bunch of places now.
</blockquote></details>

(This is definitely not a suggestion since it could be a bad idea. Just making sure we've thought about it.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 12 of 14 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/ccl/sqlccl/backup.go, line 638 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfgWMGY9YLC4aC76a7:-Kwg0l2waECU-e233Rr0:bjs7hg1) ([raw file](https://github.com/cockroachdb/cockroach/blob/2a87462504736e815cb5f5f8d9d0f2d4411e0fc1/pkg/ccl/sqlccl/backup.go#L638)):*

Might as well move this down into the first `if` block to avoid fetching the node count if not using `file://` storage.
</blockquote></details>

Ended up going the opposite direction and pulling the `if` into the helper (@mjibson wanted it to be impossible to forget) so it's no as easy to do conditionally anymore. Could pass `Gossip` to it instead I guess if we really care?

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-Kwg571Mlg_ymssAVqED:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*

Might as well pull out a named constant here, too, since you did it everywhere else. Or extract a global named constant like `sqlccl.IrrelevantClusterSize`or something?
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/csv.go, line 273 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfteYVwH49mBNcipLW:-Kwg474hz7FlgrV18Y3i:b4gbbvy) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L273)):*

(This is definitely not a suggestion since it could be a bad idea. Just making sure we've thought about it.)
</blockquote></details>

I'm not a huge fan of the struct -- it's just this package, and settings feels like a first-class arg to most functions to me.

---

*[pkg/ccl/storageccl/export_storage.go, line 213 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfuM0fN4YBBRN1WaNh:-Kwg4U0EYdBB8XU9lZSq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/19f1914fdd72de5dc5f3f2790916d429f4fdeda9/pkg/ccl/storageccl/export_storage.go#L213)):*

This should probably stay the same. This function should only be used during CLI operations, and so should never trigger the cluster size error.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 12 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/backup_test.go, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwfh-b12N_R9AIJoQu2:-Kwg5XEfGn2YOpTo0N_a:b580acg) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4ab9d0f91be53940a8efea339683291020b279f/pkg/ccl/sqlccl/backup_test.go#L247)):*

As a followup because I hadn't everything from before: I agree that the export storage should disallow relative paths. The CLI appears to already convert relative to absolute.
</blockquote></details>

Hm. I agree that you _probably_ want to use an abs path rather than guess at the server CWD, but do we really want to _mandate_ that? I guess?

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

knz:
Review status: 11 of 14 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-Kwiv58WHFaJa0id-wA7:b-y15b7g) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*
> ```Go
> 		return nil, err
> 	}
> 	if conf.Provider == roachpb.ExportStorageProvider_LocalFile && clusterSize > 1 && !conf.LocalFile.IsNFS {
> ```

Argh, I'm sorry I haven't seen this before. It seems unwise to me to use the acronym "NFS" if you mean something else. The problem with using the word "nfs" is that anyone who is *not* using NFS will dismiss the importance of this because they will feel it's not relevant to them.

I could suggest `shared://` instead, and emphasize in the docs / error message / commit message that "the path is visible from all the nodes, i.e. it is backed by a network filesystem mounted on the same location on every node", since you need both conditions (network fs + same mount point).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 11 of 14 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-KwjyauzIftUVZ5Jb6lQ:b1noc98) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

Argh, I'm sorry I haven't seen this before. It seems unwise to me to use the acronym "NFS" if you mean something else. The problem with using the word "nfs" is that anyone who is *not* using NFS will dismiss the importance of this because they will feel it's not relevant to them.

I could suggest `shared://` instead, and emphasize in the docs / error message / commit message that "the path is visible from all the nodes, i.e. it is backed by a network filesystem mounted on the same location on every node", since you need both conditions (network fs + same mount point).
</blockquote></details>

@knz so we _want_ everyone who is not using NFS (or samba) to dismiss this option. 

We included local file system access for two use-cases: single node testing (where "file://" is obvious and ok), and, much less likely, real nfs (or equivalent) deployments. 

We thought an odd name like "nodelocal://" would send a user to docs to find out what it did, and then see our warning that it only made sense if pointed to an NFS or similar network-backed path. However we're _still_ finding that internal and external users miss this and point it at non-NFS paths, so we wanted to make it even clearer. 

The majority of users using local files are probably in the first group -- playing with things on single node.  Given the complexities in deploying a real distributed FS, we'd like to think those in the second group, with real NFS deployments, are i.e. enterprises, and thus are more likely to read the docs before trying things.

So, overall, I'm not as concerned with someone who is using something other than NFS (e.g. afs or samba) dismissing this mistakenly, since that would probably just cause them to re-read the docs, rather than mistakenly produce useless backups and potentially open themselves to data loss.

All of that said, the `nfs://` scheme is commonly accepted (RFC 2224) for native, direct NFS access, so this is indeed a misuse of it, so we should pick something else.

I'd be fine with something really blunt, given that we only expect a few deployments will legitimately want this behavior. Maybe `network-backed-file://` or `file+sharedfs://`?  (schemes may contain ., + and -)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

knz:
Review status: 11 of 14 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-Kwk47ke-VV_OHB34uZe:bkb0vf7) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

@knz so we _want_ everyone who is not using NFS (or samba) to dismiss this option. 

We included local file system access for two use-cases: single node testing (where "file://" is obvious and ok), and, much less likely, real nfs (or equivalent) deployments. 

We thought an odd name like "nodelocal://" would send a user to docs to find out what it did, and then see our warning that it only made sense if pointed to an NFS or similar network-backed path. However we're _still_ finding that internal and external users miss this and point it at non-NFS paths, so we wanted to make it even clearer. 

The majority of users using local files are probably in the first group -- playing with things on single node.  Given the complexities in deploying a real distributed FS, we'd like to think those in the second group, with real NFS deployments, are i.e. enterprises, and thus are more likely to read the docs before trying things.

So, overall, I'm not as concerned with someone who is using something other than NFS (e.g. afs or samba) dismissing this mistakenly, since that would probably just cause them to re-read the docs, rather than mistakenly produce useless backups and potentially open themselves to data loss.

All of that said, the `nfs://` scheme is commonly accepted (RFC 2224) for native, direct NFS access, so this is indeed a misuse of it, so we should pick something else.

I'd be fine with something really blunt, given that we only expect a few deployments will legitimately want this behavior. Maybe `network-backed-file://` or `file+sharedfs://`?  (schemes may contain ., + and -)
</blockquote></details>

`network://`, `sharedfile://`, `shared://`, `global://`
Not too fond of extra punctuation (+, -) since I foresee some parsers won't like it.

Also I'd like to come back to this restriction "the mount point must be the same everywhere". I don't see it spelled out in the error message nor the commit message, but I think the restriction is actually there and should be documented.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 11 of 14 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-Kwk4iSr1GSPFgMbLIL-:bvvw5zd) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

`network://`, `sharedfile://`, `shared://`, `global://`
Not too fond of extra punctuation (+, -) since I foresee some parsers won't like it.

Also I'd like to come back to this restriction "the mount point must be the same everywhere". I don't see it spelled out in the error message nor the commit message, but I think the restriction is actually there and should be documented.

</blockquote></details>

`shared` or `sharedfile` or `global`  all sound too easy to me -- i.e. I could imagine something thinking that meant that that implied _we_ were handling the sharing for them. Given our first attempt with `nodelocal`, I think we should err on the side of explicit.

what about `networkfs://` ?

Also, the URI RFC says that that punctuation is allowed, and the important parser (the one we're using) handles it, but I guess if we can find something without punctuation that we like, then we might as well play it safe.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

knz:
Review status: 11 of 14 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-Kwk7oS8LX8O5L_8JHz8:bb2vyl8) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

`shared` or `sharedfile` or `global`  all sound too easy to me -- i.e. I could imagine something thinking that meant that that implied _we_ were handling the sharing for them. Given our first attempt with `nodelocal`, I think we should err on the side of explicit.

what about `networkfs://` ?

Also, the URI RFC says that that punctuation is allowed, and the important parser (the one we're using) handles it, but I guess if we can find something without punctuation that we like, then we might as well play it safe.

</blockquote></details>

`network+file://` or `network-file://` would work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 2 of 14 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-KwkCyYPef1KvKJixsBT:bubr6zw) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*
> network-file://

:thumbsup: Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

bdarnell:
Reviewed 2 of 4 files at r3, 12 of 12 files at r6.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-KwkYGML6sdRVF2NP3Tz:bwdm083) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

> network-file://

:thumbsup: Done.
</blockquote></details>

Standards-lawyer nit: Per RFC 3986,`://` in a URI indicates that the following component is a hostname (or more precisely an "authority"). So in the standard `file:` URI scheme (RFC 8089), the URI is written with three slashes (empty authority), or one slash (authority omitted, absolute path). I think no slashes might be legal for a relative path, but this is not clear; it's rare for relative paths to be used with an explicit URI scheme. 

---

*[pkg/ccl/sqlccl/backup_test.go, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwfh-b12N_R9AIJoQu2:-Kwk_0K6VDUez6n-JaiG:bd3bdk0) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4ab9d0f91be53940a8efea339683291020b279f/pkg/ccl/sqlccl/backup_test.go#L247)):*

Hm. I agree that you _probably_ want to use an abs path rather than guess at the server CWD, but do we really want to _mandate_ that? I guess?

Done.
</blockquote></details>

I think we should mandate absolute path syntax in the URI; relative paths in URIs are poorly defined. However, we don't have to evaluate the absolute path relative to the root of the server's filesystem. Admins should be able to (or maybe *required* to) point to an alternative directory to serve as the root of this "filesystem", instead of allowing access to any filesystem location that the server can access. 

I thought `nodelocal` was going to be test-only and disabled in production servers. I don't think it's appropriate to have access to arbitrary filesystem locations via the database without opt-in. (this can easily be a source of security vulnerabilities. [example](https://legalhackers.com/advisories/MySQL-Exploit-Remote-Root-Code-Execution-Privesc-CVE-2016-6662.html))

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-KwkckYhV1Rn8MYcY--J:bg0jfi3) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*

Done.
</blockquote></details>

This doesn't appear to have been done.

---

*[pkg/ccl/storageccl/export_storage.go, line 124 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwkcxneMG1j_ksS12UX:-KwkcxneMG1j_ksS12UY:bhl2u7y) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/storageccl/export_storage.go#L124)):*
> ```Go
> 		conf.LocalFile.Path = uri.Path
> 		conf.LocalFile.IsSharedFS = uri.Scheme == "network-file"
> 	case "":
> ```

I'm no longer convinced that having now 4 separate schemes for local file storage is a good idea. I will argue for the removal of "" and "file" so that there's just one way to do a thing instead of 3.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-Kwkib4UE7Si974vFahJ:b-2spxu3) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*

This doesn't appear to have been done.
</blockquote></details>

huh? it's called `clusterSize` now?

---

*[pkg/ccl/storageccl/export_storage.go, line 124 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwkcxneMG1j_ksS12UX:-Kwkiho1kkoKdJgz1Hb7:b-h4cel0) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/storageccl/export_storage.go#L124)):*

I'm no longer convinced that having now 4 separate schemes for local file storage is a good idea. I will argue for the removal of "" and "file" so that there's just one way to do a thing instead of 3.
</blockquote></details>

Eh, I think we should use the standard `file://`.

As far as too many synonyms: I think `nodelocal` is strange, and should go away if we have a more reliable solution than an odd name, like a error. Obviously we'll still accept it for compatibility, but we'll expunge it from docs, so I wouldn't count it as one of the ways.

I don't feel _that_ strongly about plain paths working, but I do think it's nice and elegant, e.g. `RESTORE DATABASE foo FROM /tmp/foo`. That said, `RESTORE DATABASE foo FROM file://tmp/foo` isn't _that_ much worse though, thus my lack of a strong opinion here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-KwlJMZB_O3psT9-A14N:bfu3iam) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*

huh? it's called `clusterSize` now?
</blockquote></details>

I thought the suggestion was to make a singleton, similar to cluster.NoSettings, instead of a bunch of consts that happened to be identical.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:

---

*[pkg/ccl/sqlccl/backup_test.go, line 247 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwfh-b12N_R9AIJoQu2:-KwlNr2bWtZ_HVbWp_fy:bt54nwn) ([raw file](https://github.com/cockroachdb/cockroach/blob/f4ab9d0f91be53940a8efea339683291020b279f/pkg/ccl/sqlccl/backup_test.go#L247)):*

I think we should mandate absolute path syntax in the URI; relative paths in URIs are poorly defined. However, we don't have to evaluate the absolute path relative to the root of the server's filesystem. Admins should be able to (or maybe *required* to) point to an alternative directory to serve as the root of this "filesystem", instead of allowing access to any filesystem location that the server can access. 

I thought `nodelocal` was going to be test-only and disabled in production servers. I don't think it's appropriate to have access to arbitrary filesystem locations via the database without opt-in. (this can easily be a source of security vulnerabilities. [example](https://legalhackers.com/advisories/MySQL-Exploit-Remote-Root-Code-Execution-Privesc-CVE-2016-6662.html))
</blockquote></details>

Hm, we should probably open a dedicated issue for this, rather than an inline reviewable thread #19357

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-KwlKtmUkkxjp1WPvwOp:b-pk3gat) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*

I thought the suggestion was to make a singleton, similar to cluster.NoSettings, instead of a bunch of consts that happened to be identical.
</blockquote></details>

ah, sorry, I went with the first suggestion "pull out a named constant here". will switch it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

knz:

---

*[pkg/ccl/sqlccl/backup_test.go, line 253 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwnf6naJ6okIcWh9eQx:-Kwnf6naJ6okIcWh9eQy:bitmxi7) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/sqlccl/backup_test.go#L253)):*
> ```Go
> 		_, dir, _, sqlDB, cleanupFn := backupRestoreTestSetup(t, singleNode, numAccounts, initNone)
> 		defer cleanupFn()
> 		dir = strings.TrimPrefix(dir, "network-file:/")
> ```

ditto what Ben said above: the proper URI syntax for things that don't have a hostname is `prefix:///path/to/thing`, with 3 forward slashes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:

---

*[pkg/ccl/sqlccl/backup_test.go, line 253 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwnf6naJ6okIcWh9eQx:-KwoTSxs1zZAU8wLBzSm:b-sy9dsf) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/sqlccl/backup_test.go#L253)):*

ditto what Ben said above: the proper URI syntax for things that don't have a hostname is `prefix:///path/to/thing`, with 3 forward slashes.
</blockquote></details>

Yep, that's what we've got here: `dir` here is defined in the setup helper as `"network-file://" + dir` and `dir` has a leading slash. Indeed, in `export_stroage.go`, we already have an assertion that that `host` field of the parsed URI is empty for file paths. I can strip an extra slash here to make that clearer though if you think it would help.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

knz:

---

*[pkg/ccl/sqlccl/backup_test.go, line 253 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwnf6naJ6okIcWh9eQx:-KwoWbafEirQd3f6MCoJ:b-7ejlc0) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/sqlccl/backup_test.go#L253)):*

Yep, that's what we've got here: `dir` here is defined in the setup helper as `"network-file://" + dir` and `dir` has a leading slash. Indeed, in `export_stroage.go`, we already have an assertion that that `host` field of the parsed URI is empty for file paths. I can strip an extra slash here to make that clearer though if you think it would help.
</blockquote></details>

Nah, it's all right. Carry on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-KwpUbAPXsdLRJzoG4Hb:buhw5w3) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

Standards-lawyer nit: Per RFC 3986,`://` in a URI indicates that the following component is a hostname (or more precisely an "authority"). So in the standard `file:` URI scheme (RFC 8089), the URI is written with three slashes (empty authority), or one slash (authority omitted, absolute path). I think no slashes might be legal for a relative path, but this is not clear; it's rare for relative paths to be used with an explicit URI scheme. 
</blockquote></details>

We explicitly check for and reject non-empty Host components of the parsed URI for file URIs, so I think we're already enforcing that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 10 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/csv.go, line 98 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwfrlIeSN5R06lJBkr-:-KwpbiZr7Nn4gkEhchDL:b-40l26q) ([raw file](https://github.com/cockroachdb/cockroach/blob/143c5f0d94353c10f9934dfbf9f024c4ac212c73/pkg/ccl/sqlccl/csv.go#L98)):*

ah, sorry, I went with the first suggestion "pull out a named constant here". will switch it. 
</blockquote></details>

okay, went ahead and switched to a constant, and while I was at it, just passing a *gossip so we can defer counting to inside the conditional where we actually use it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:
Review status: 10 of 14 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/cliccl/load.go, line 173 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwpoDm3fnjeAZc1amNk:-KwpoDm3fnjeAZc1amNl:bjdbqef) ([raw file](https://github.com/cockroachdb/cockroach/blob/d6e2cbb18a18edd876dc7200e3e35133f56e8e3b/pkg/ccl/cliccl/load.go#L173)):*
> ```Go
> 		}
> 	}
> 	desc, err := sqlccl.ReadBackupDescriptorFromURI(ctx, basepath, cluster.NoSettings, nil /* gossip */)
> ```

Can we export NoCluster? Just like cluster.NoSettings?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:
Review status: 10 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/storageccl/export_storage.go, line 124 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwkcxneMG1j_ksS12UX:-KwppJy8NB00i7dYFT2r:bkwfris) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/storageccl/export_storage.go#L124)):*

Eh, I think we should use the standard `file://`.

As far as too many synonyms: I think `nodelocal` is strange, and should go away if we have a more reliable solution than an odd name, like a error. Obviously we'll still accept it for compatibility, but we'll expunge it from docs, so I wouldn't count it as one of the ways.

I don't feel _that_ strongly about plain paths working, but I do think it's nice and elegant, e.g. `RESTORE DATABASE foo FROM /tmp/foo`. That said, `RESTORE DATABASE foo FROM file://tmp/foo` isn't _that_ much worse though, thus my lack of a strong opinion here.


</blockquote></details>

I would prefer exactly one supported and documented non-nsf scheme (`file` is fine with me, but I don't care), and the deprecated `nodelocal`. I would prefer to drop `""`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

bdarnell:
Review status: 11 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-Kwum1nkpWGINTA8LQzx:b-2p8xn7) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

We explicitly check for and reject non-empty Host components of the parsed URI for file URIs, so I think we're already enforcing that?
</blockquote></details>

I was just confused by the number of slashes in backup_test.go.

---

*[pkg/ccl/sqlccl/backup_test.go, line 253 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwnf6naJ6okIcWh9eQx:-Kwun0ZiE10h2Kzaz27X:b-8sfgi1) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/sqlccl/backup_test.go#L253)):*

Nah, it's all right. Carry on.
</blockquote></details>

So after trimming the prefix, `dir` starts with two slashes? That's weird. I'd either strip off two slashes or leave all three. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

dt:
Review status: 6 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/cliccl/load.go, line 173 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwpoDm3fnjeAZc1amNk:-Kwq0X4twjkIchryMnCj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d6e2cbb18a18edd876dc7200e3e35133f56e8e3b/pkg/ccl/cliccl/load.go#L173)):*

Can we export NoCluster? Just like cluster.NoSettings?
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/backup.go, line 77 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwiv58VgT_tRWddm-A9:-KwvAVB4VX0YqiKx6u2Y:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/256301eca5124d9bc6a6bd8df174ec539f30dea8/pkg/ccl/sqlccl/backup.go#L77)):*

I was just confused by the number of slashes in backup_test.go.
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/backup_test.go, line 253 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-Kwnf6naJ6okIcWh9eQx:-KwvAKaX1_ni8_1FkYLp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/sqlccl/backup_test.go#L253)):*

So after trimming the prefix, `dir` starts with two slashes? That's weird. I'd either strip off two slashes or leave all three. 
</blockquote></details>

Done.

---

*[pkg/ccl/storageccl/export_storage.go, line 124 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-KwkcxneMG1j_ksS12UX:-Kwq0VbNg98_je0W7WGn:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3f112a85c7942bceb155ed1be968552f9d72afb/pkg/ccl/storageccl/export_storage.go#L124)):*

I would prefer exactly one supported and documented non-nsf scheme (`file` is fine with me, but I don't care), and the deprecated `nodelocal`. I would prefer to drop `""`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326)*

mjibson:
:lgtm:

---

Review status: 6 of 14 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-:-Kx9LoUm_gWWqgjC6Qwi:bnfp4nl)*

knz:
:lgtm:

---

Reviewed 5 of 12 files at r6, 1 of 4 files at r7, 7 of 7 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19326#-:-Kxs4Xdu9KPZB-Mv8JGh:bnfp4nl)*

dt:
Holding off on this pending discussion and resolution on a plan in #19357.

dt:
I think our base dir restrictions made nodelocal a little less likely to be used incorrectly, so I'm not convinced this is still worth adding.

dianasaur323:
@Amruta-Ranade is there an issue to document this in 2.0? A user just asked about it on GitHub. https://github.com/cockroachdb/cockroach/issues/21136

Amruta-Ranade:
@dianasaur323 No, I don't think we have a doc issue for this one. Opening a doc issue now.

dianasaur323:
thank you!! @Amruta-Ranade

dt:
@Amruta-Ranade, this was closed without merging it, so nothing to do here w.r.t. docs.

dianasaur323:
@dt thanks for the clarification. I honestly can't remember what we ended up landing on anymore. Does that mean that nodelocal still works then?

Issue #17135, opened 2017-07-20
distsql: support outer merge joins


rampage644:
Recent change #17017 changed `joinerBase` ctor interface and only currently used `HashJoiner` was adapted for the change. `MergeJoiner` doesn't properly initializes base class values.

RaduBerinde:
Special support for the merged columns is no longer necessary, thanks to #19660. We just need to plan mere joins for outer joins as well.

lego:
Unfortunately, we can't just plan it! :(

Alter enabling the merge joiner to be planned, it seems to be incorrect on equalities where some columns may be null. Doing `l.k = r.k` behaves like `l.k = r.k OR (l.k IS NULL AND r.k IS NULL)` which is incorrect. For example:
```sql
# The row data
root@:26257> select * from order_test;
+------+------+
|  k   |  v   |
+------+------+
|    0 | NULL |
|    0 |    1 |
|    2 |    4 |
| NULL |    4 |
+------+------+
(4 rows)

root@:26257/test> set distsql=off;

root@:26257/test>
SELECT
  l.k, l.v, r.k, r.v
FROM
  (SELECT * FROM order_test ORDER BY k, v) l
FULL OUTER JOIN
  (SELECT * FROM order_test ORDER BY k, v) r
  ON l.k = r.k and l.v = r.v;
+------+------+------+------+
|  k   |  v   |  k   |  v   |
+------+------+------+------+
| NULL |    4 | NULL | NULL |
|    0 | NULL | NULL | NULL |
|    0 |    1 |    0 |    1 |
|    2 |    4 |    2 |    4 |
| NULL | NULL | NULL |    4 |
| NULL | NULL |    0 | NULL |
+------+------+------+------+
(6 rows)

Time: 1.291227ms

root@:26257/test> set distsql=always;

root@:26257/test>
SELECT
  l.k, l.v, r.k, r.v
FROM
  (SELECT * FROM order_test ORDER BY k, v) l
FULL OUTER JOIN
  (SELECT * FROM order_test ORDER BY k, v) r
  ON l.k = r.k and l.v = r.v;
+------+------+------+------+
|  k   |  v   |  k   |  v   |
+------+------+------+------+
| NULL |    4 | NULL |    4 |
|    0 | NULL |    0 | NULL |
|    0 |    1 |    0 |    1 |
|    2 |    4 |    2 |    4 |
+------+------+------+------+
(4 rows)

Time: 4.018594ms

```

[...]

PR #19327, opened 2017-10-17
RFC: SCRUB index and physical check implementation

lgo:
This RFC presents a method of running a check on a secondary index in
an efficient manner, as well as other physical checks which are closely
in implementation.

This check can be run through the SQL statement `SCRUB`, using the
options to specifically check indexes or check the physical data of the
database.

cc: @vivekmenezes

### DIFF
--- a/docs/RFCS/scrub_index_implementation.md
+++ b/docs/RFCS/scrub_index_implementation.md
@@ -1,0 +1,461 @@
+- Feature Name: SCRUB index and physical check implementation
+- Status: draft
+- Start Date: 2017-10-11
+- Authors: Joey Pereira
+- RFC PR: [#19327](https://github.com/cockroachdb/cockroach/pull/19327)
+- Cockroach Issue: [#10425](https://github.com/cockroachdb/cockroach/issues/10425) (Consistency checking), [#19221](https://github.com/cockroachdb/cockroach/issues/19221) (Using distSQL to do index checks)
+
+# Summary
+
+This RFC presents a method of running a check on a secondary index in an
+efficient manner, as well as other physical checks which are closely in
+implementation.
+
+This check can be run through the SQL statement `SCRUB`, using the
+options to specifically check indexes or check the physical data of the
+database.
+
+
+# Motivation
+
+See the [SCRUB command RFC] for motivation.
+
+In addition to the motivations for the check, this work may also benefit
+schema changes. The last unresolved question of the
+[online schema change RFC] mentions a method of distributing backfill
+work. While this is currently done through distSQL, there are problems
+where it's currently unable to backfill an index at a specific
+timestamp. This has become prevalent as it may be the source of issues
+such as [#18533].
+
+The methods for collecting index data introduced in this RFC are similar
+to that which is needed to create an alternative backfilling mechanism.
+
+[#18533]: https://github.com/cockroachdb/cockroach/issues/18533
+[SCRUB command RFC]: https://github.com/cockroachdb/cockroach/issues/18675
+[online schema change RFC]: https://github.com/cockroachdb/cockroach/blob/94b3f764ead019c529f14c7aab64724991f3ac67/docs/RFCS/20151014_online_schema_change.md
+
+
+# Guide-level explanation
+
+After implementing the methods proposed by this RFC, the `SCRUB` command
+will have the means to check for physical data errors and for index
+errors. In particular, the errors we are checking for are:
+- `missing_secondary_index`
+- `dangling_secondary_index`
+- `invalid_encoding`
+- `invalid_composite_encoding`
+- `noncanonical_encoding`
+
+The first two are index errors. Missing secondary indexes occur when an
+index entry is not correctly made or updated with the primary k/v.
+Conversely, a dangling index entries occur if the primary key referenced
+in the index k/v is not valid. This happens if the primary k/v being
+incorrectly deleted while the index k/v remains, in addition to the
+reasons for missing index entries.
+
+To detect missing index entries, we need to do a full scan of the
+primary k/v. Then for each primary k/v, we expect to find a
+corresponding index entry. If no index entry is found, we have a missing
+index entry. This does not capture any dangling index entries though, as
+we are only looking up index entries which have a valid primary key
+reference. In order to find dangling indexes, we need to do a similar
+process where we do a full scan of the secondary index k/v.
+
+As a result, a full outer join is a natural solution for this problem.
+Taking this approach, we limit the scans done to 2, as opposed to
+potentially scanning both the primary k/v and the secondary index k/v
+and doing expensive point-lookups to the other set, twice.
+
+Lastly, we want to propagate the errors to be the SQL results of
+`SCRUB`, including the information of:
+- Database and table involved
+- Error type
+- Primary key and secondary index value involved
+- Potentially the raw key value
+
+The following changes are introduced for in order to do the entirety of
+the index check:
+
+- Add logic plan creation (`scrubNode`[0] initialization) to construct a
+  logical plan for checking an index, then plan and run it through the
+  distSQL execution engine. This plan will accomplish the full scan and
+  join in one go.
+
+  The solution instead constructs multiple logical plans
+  upon execution of our original plan, and proceeds to run them.
+
+The latter three errors related to encoding happen due to errors in the
+row data or our functions manipulating the row data. These can be tested
+just by doing a full scan of the data to check (primary k/v and index
+entry k/v). Similarly to the above, we will also want the error results
+to be propagated up as the SQL results.
+
+The following changes are introduced for running physical checks:
+
+- Transform the `scrubNode`[0] a logical plan that fully encapsulates
+  the logic to scan relevant tables. This plan will run through distSQL,
+  similar to any regular plan.
+
+- Add a `RunCheck` enum to `scanNode` and plumb it through the
+  `TableReader` for `RowFetcher` initialization during distSQL planning.
+  This enum would only be meaningfully set during planning of `SCRUB`
+  when doing a physical check.
+
+- Modify the `RowFetcher` to conditionally run check code in existing
+  row fetching code paths, using the `RunCheck` indicator. For example,
+  we want to return an error instead of panic when NULL is encountered
+  on a non-NULL column.
+
+- Add a new iterator function to `RowFetcher`, `NextCheckError`, for
+  just scanning for errors. In here we can execute additional checking
+  code on a separate code path than the critical fetching one. This lets
+  us consume check errors instead of row data.
+
+- Add logic to `TableReader` to consume `NextCheckError` and handle the
+  difference between wanted rows and returned rows (check error row
+  type). This may potentially require a new `TableChecker`. With this
+  change, we are able to retrieve the error results desired.
+
+[0]: `scrubNode` is the root planning node when using the `SCRUB`
+     command.
+
+
+# Reference-level explanation
+
+The details of implementation are broken down into a few parts. NB:
+the optimizations section is just ideation. The goal of it is to discuss
+the feasibility and time-effort of the possibilities.
+
+1. Index checks and how they will be implemented.
+
+2. Physical checks and how they will be implemented.
+
+3. Generating and executing the logical plan in the distSQL execution engine
+
+4. Optimizations
+
+   4.a. Multiple simultaneous index checks.
+
+   4.b. Executing index and physical checks simultaneously.
+
+
+## 1. Index checks and how they will be implemented.
+
+Checking the index can be expressed through a rather simple logical SQL
+plan.  The final result will return all erroneous index entries -- both
+dangling references and missing entries. In short, this can be broken
+down into a few stages of processing:
+
+1) Scan through the primary k/v to fetch relevant index data[1] and
+   filter rows where the secondary index value is `NULL`.
+2) Scan through the secondary k/v to fetch all data stored in the index.
+3) Execute a natural full outer join on both streams to pair primary k/v
+   with their secondary index k/v. This filters any entry from both
+   streams that don't have a corresponding entry in the other stream.
+
+For example, if we have the following schema:
+
+```sql
+CREATE TABLE t1 (
+   pkey int primary key,
+   name int,
+   INDEX name_idx (id ASC),
+)
+```
+
+Then checking the indexes can be done through the logical plan
+represented by:
+
+```sql
+SELECT p.pkey, p.name, s.pkey, s.name
+FROM
+  t1@{FORCE_INDEX=primary,NO_INDEX_JOIN} as p
+FULL OUTER JOIN
+  t1@{FORCE_INDEX=name_idx,NO_INDEX_JOIN} as s
+ON
+  ((p.pkey IS NOT NULL and s.pkey IS NOT NULL and p.pkey = s.pkey) OR
+   (p.pkey IS NULL AND s.pkey IS NULL)) AND
+  ((p.name IS NOT NULL and s.name IS NOT NULL and p.name = s.name) OR
+   (p.name IS NULL AND s.name IS NULL)) AND
+WHERE
+  (p.pkey IS NULL AND p.name IS NULL) OR
+  (s.pkey IS NULL AND s.name IS NULL)
+```
+
+ In short, this query is:
+ 1) Scanning the primary index and the secondary index.
+ 2) Joining them on all of the secondary index columns and extra columns
+    that are equal.
+ 2) Joining them on all of the secondary index columns and extra columns
+    that are equivalent. This is a fairly verbose check due to
+    equivilancy properties when involving NULLs in order to be an
+    anti-join.
+ 3) Filtering to achieve an anti-join. The first line of the predicate
+    takes rows on the right for the anti-join. The second line of the
+    predicate takes rows on the left for the anti-join.
+
+ Because this is an anti-join, the results are as follows:
+ - If any primary index column on the left is NULL, that means the right
+   columns are present. This is because of the invariant that primary
+   index columns are never null.
+ - Otherwise, the left columns are present.
+
+As for the overall performance of this operation, if N is the number of
+primary index k/v rows on a node, and M is the number of secondary index
+k/v rows on a node. Then each node will scan exactly N+M rows and stream
+them to a joiner.
+
+Because the inputs will not be sorted similarily and will be fairly
+large, this join will ideally be a hash join. In this scenario, the
+memory used on each node will be at most the total data divided by the
+joiners, and temp storage will be used if necessary.
+
+[1]: Relevant index data includes the columns, extra columns, and store
+     columns of the index.
+
+## 2. Physical checks and their implementation
+
+The **physical table** is the underlying k/v behind tables. This is
+where we find primary and secondary indexes (as distinct keys),
+groupings of certain data (column families), how index data is organized
+(index storing conditions), and the encoding of the data.
+
+Most of the checks that involve the physical table are encoding and
+related data validity. This data is already partially checked during
+scanning in `RowFetcher`. Because of this, adding checks is a matter of
+a few additive changes, and handling the error to instead return row
+results.
+
+The checks we can implement in `RowFetcher` include:
+- Tables using [FAMILY][] have the data organized into families
+  correctly
+- Invalid data encodings
+- Non-canonical encodings (i.e. data round-trips)
+- [Composite encodings][] are valid
+- Key encoding correctly reflects the ordering of the decoded values
+- `NOT NULL` constraint violation (not physical table, but checked at
+  this layer)
+
+The underlying scanning is in the `RowFetcher`. Inside the current
+`RowFetcher` the method `processKv` is responsible for encoding specific
+work as it iterates over the k/v, which is where the checks will be
+added, including:
+- Check composite encodings are correct and round-trip
+- Check key encodings correctly reflect ordering of decoded values
+- Check the KV pair is well-formed (everything is present, no extra
+  data, checksum matches)
+- For primary keys, verify that the column family 0 (sentinel) is
+  present
+
+We can also throttle the speed of the check by modifying the `limitHint`
+provided to `KVFetcher`, and sleeping between batches.
+
+Thus, to run the check we just need to do a scan over the whole table,
+and the returned values will be the `CheckError` as found in the
+[SCRUB interface RFC]. Ideally, this is executed through the distSQL, as
+there will be very few results relative to the data scanned.
+
+Invalid data encoding (and others) already return an `error` while
+scanning. We want to capture errors in `RowFetcher` that come from any
+external function calls and wrap them in a `scrub.Error` struct
+with an error code annotated.
+
+Those error structs will be captured further upstream inside the
+TableReader and translated into the row results. In the existing
+scanning code path, this same error struct will simply be unwrapped.
+
+For example, consider this excerpt from `RowFetcher.processValueSingle`:
+
+```go
+value, err := UnmarshalColumnValue(rf.alloc, typ, kv.Value)
+if err != nil {
+  return "", "", err
+}
+// Becomes
+value, err := UnmarshalColumnValue(rf.alloc, typ, kv.Value)
+if err != nil {
+  return "", "", scrub.NewError(
+    scrub.InvalidEncodingError,
+    err,
+  )
+}
+```
+
+Another chunk of the work will then be plumbing through the check error
+results out as the table schema we are scanning will differ from the
+result rows representing errors. To get around this we must manually
+construct the distSQL physical plan. This will also take some extra code
+to wrap the call from TableReader to RowFetcher where it will transform
+any caught errors.
+
+[SCRUB interface RFC]: https://github.com/cockroachdb/cockroach/pull/18675
+
+## 3. Generating and executing the logical plan in the distSQL execution engine
+
+During local engine execution, we will generate the necessary check
+plans, then execute the plans in sequence on the distSQL execution
+engine. These plans will scan the data in a distributed fashion and the
+only data leaving the nodes are any error results. This strategy is
+similar to schema change index backfilling.
+
+The advantage of this over the alternative method [3], creating the full
+logical plan during query planning, is that we can easily run the checks
+sequentially. We can block waiting for the distSQL results of each check
+before running the next check. This also pulls the work of aggregating
+results from the distSQL execution engine to simpler Go code that
+accumulates the results.
+
+
+## 4. Optimizations
+
+Note that this section is largely just the ideas behind optimizations
+and are not required to be resolved. The goal of this section is to
+gauge feasibility, and whether or not it's worth it to go down those
+routes.
+
+
+### 4.a. Multiple simultaneous index checks
+
+A challenge here is producing both check errors and the necessary data
+to run index checks simultaneously. It is not currently known if
+multiple different schemas can be streamed out of one table reader.
+
+In the `RowFetcher`, this may look like a `NextRowOrCheckError` iterator
+that may return either row data or check error data, and we emit those
+to the corresponding outputs.
+
+A workaround may be to have an intermediary processor for each mirrored
+stream which filters out only the columns desired for each consumer --
+either a specific index check (a joiner), or the final result errors (in
+final aggregation).
+
+Alternatively, a crazy idea may modify the RowFetcher or TableReader to
+produce two separate schemas, branching off the idea of MultiRowFetchers
+made for the adjustments for interleaved tables.
+
+
+### 4.b. Executing index and physical checks simultaneously
+
+While checking multiple indexes sequentially is discussed above, but a
+problem is that a naive solution involves doing `N` scans of the primary
+k/vs for checking `N` indexes. The following approaches are both
+non-trivial changes, so this idea is largely suggested as an
+optimization for future work.
+
+The extra scans can be avoided if we scan all the necessary data for all
+indexes being check in one table reader, then use a mirror router to
+mirror the data to multiple join consumers. Unfortunately, this would
+require making one of the changes:
+
+- Manual surgery of a distSQL physical plan to independently plan each
+  index check, then combine them with:
+
+   - Slicing out the primary k/v scan in each, replacing it with a
+     single mirrored scan that fetches all the data.
+
+   - Remove each plan's final aggregator, and have them all stream to a
+     single new final aggregator.
+
+- Or, adding the mechanisms to be able to do a `UNION` of all the
+  results and then elide redundant table readers based on processors'
+  column requirements. This is the same final aggregation `UNION`
+  mentioned in section 3.a. but we also need to solve the problem of
+  scanning the same k/v data multiple times.
+
+
+# Rationale and Alternatives
+
+## [1] Use a `JOIN READER` aggregator for index checking
+
+Instead of using a `JOINER` aggregator, we could use a `JOIN READER`.
+How this would work is that instead of the 3 stages introduced in 1.a,
+the process for checking would involve 2 steps:
+1) Scan the primary k/v for a table.
+2) Use a `JOIN READER` aggregator that does point-lookups against the
+   secondary index.
+
+This has two major problems:
+- Producing lots of RPCs for the secondary index entries, potentially
+  being a performance problem.
+- Does not check the secondary index for dangling entries. This requires
+  a second scan which does the reverse point-lookup.
+
+
+## [2] Do point-lookups directly during RowFetching for index checking
+
+This is by far the easiest method to do as it only involves:
+1) Collect the required secondary index data. The first code path that
+   accesses required data is in `RowFetcher`.
+2) Do point-lookups for the secondary index entry, and compare.
+
+Step 1 is ideally done in the `RowFetcher` as this is also where we can
+avoid passing decoded data further upstream, when we only need check
+errors.
+
+Step 2 can be optimized by batching the k/v lookups, which would be a
+trivial mechanism to add inside `RowFetcher`. In this case, having an
+alternative `NextRow` called `NextCheckError` makes more sense to not
+add this code to the critical code path.
+
+A [proof of concept] has been done where this does not batch lookups or
+propagate errors to the client. Errors encountered are just logged.
+
+This is very similar to the alternative proposal [1]. This approach has
+the same problems.
+
+[proof of concept]: https://github.com/lego/cockroach/blob/0f022a8e0ef97d8233ccc8aa31466090ef3dee7e/pkg/sql/sqlbase/rowfetcher.go#L651-L673
+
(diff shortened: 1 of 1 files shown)


knz:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwir7mNc8S1IiKVWmd-:-Kwir7mNc8S1IiKVWmd0:bgy8rr3) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L1)):*
> ```Markdown
> - Feature Name: SCRUB index and physical check implementation
> ```

As per the README the file name doesn't get a date prefix until the PR is merged.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwirUh8utlWKbg_sEYp:-KwirUh8utlWKbg_sEYq:b-pmqgrr) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L32)):*
> ```Markdown
> 
> Using the results from this RFC, an alternative backfilling mechanism
> can be designed a re-constructed index is fed into a new `IndexWriter`
> ```

Structure of the sentence is unclear.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 46 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwisCuuRK3TNfLzkyRW:-KwisCuuRK3TNfLzkyRX:bbzplxm) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L46)):*
> ```Markdown
> # Guide-level explanation
> 
> The following changes are introduced for running index checks:
> ```

Before you explain the changes, make a overarching story about how it would work. For example detail what happens from the point a scrub is issued to the point it completes, and how the pieces interact. The pieces that need to be built need to be justified by the story.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 48 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwircEIDufNg3dnqRRo:-KwircEIDufNg3dnqRRp:b698i6f) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L48)):*
> ```Markdown
> The following changes are introduced for running index checks:
> 
> - Add logic to `scrubTable`[0] to construct a logical plan for checking
> ```

Remind the reader what scrubTable is.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 51 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwiryqRKg_kUpymAPMx:-KwiryqRKg_kUpymAPMy:b-hhckwt) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L51)):*
> ```Markdown
>   an index, then plan and run it through the distSQL execution engine.
> 
> - Add support for a full join in `MergeJoiner`.
> ```

Spell in full: "full outer join". 

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwisd0tYx2-H1oPexAr:-Kwisd0tYx2-H1oPexAs:bv2ayy1) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L115)):*
> ```Markdown
> 1) Scan through the primary k/v to fetch relevant index data[1].
> 2) Scan through the secondary k/v to fetch all data stored in the index.
> 3) Execute a natural full join on both streams to pair primary k/v with
> ```

It sounds elegant! 

Unfortunately I don't think you can get away with this RFC without discussing the algorithmic costs of this solution. The RFC should outline (even not in detail) what are the time and memory/storage costs of the proposed solution, and on which nodes these costs are paid.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 156 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwitFQUf-K-8e3T3xTl:-KwitFQUf-K-8e3T3xTm:b6v57p0) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L156)):*
> ```Markdown
>   Ordering: name
> 
> JOIN join
> ```

The join node needs to be annotated to indicate it's an outer join.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 185 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwitJvhqxMINhGQ73vR:-KwitJvhqxMINhGQ73vS:b-4xubkj) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L185)):*
> ```Markdown
> 
> [2]: It appears that support for `FULL` joins is not supported by
>      `MergeJoiner`. This would have to be an additional change.
> ```

I would like the RFC to outline how a full outer join would work in distsql.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 219 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwitg7wrkR5_p3To35F:-Kwitg7wrkR5_p3To35G:b-e8zfak) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L219)):*
> ```Markdown
> TODO: Is there a correct way to map errors and to their type, even
> though they are free-form text? An alternative might be to plumb a
> struct that contains the `error` and the error type (and maybe other
> ```

Yeah separate error "codes" or identifiers seem useful / necessary here.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwits-iEpSZYZrLCfc8:-Kwits-iEpSZYZrLCfc9:b-mswu79) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L272)):*
> ```Markdown
> 
> This may be easier to implement as we don't need to be concerned with
> creating a new distSQLPlanner, it will be limited in a few ways. If any
> ```

Why do you not need to create a distsqlplanner? 

It seems to me that you'll need one, because the plan needs to be deployed to the nodes that already have a replica of the data.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

knz:
@RaduBerinde can you check whether what this RFC needs of distsql is actually feasible?

lego:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwir7mNc8S1IiKVWmd-:-KwjhjMErKlbFUz3ckbD:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L1)):*

As per the README the file name doesn't get a date prefix until the PR is merged.
</blockquote></details>

Done.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwirUh8utlWKbg_sEYp:-Kwjh30ngBkESGdDdHH3:b-lkbnkm) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L32)):*

Structure of the sentence is unclear.
</blockquote></details>

Rewrote it a bit (in short, the idea/methods used to "collect the index" are common for an alternative way of backfilling indexes)

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 46 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwisCuuRK3TNfLzkyRW:-KwkawLXyv2GnMHaScSr:b-r20y3l) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L46)):*

Before you explain the changes, make a overarching story about how it would work. For example detail what happens from the point a scrub is issued to the point it completes, and how the pieces interact. The pieces that need to be built need to be justified by the story.
</blockquote></details>

Good point :). Added a lot more information to the guide section.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 48 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwircEIDufNg3dnqRRo:-Kwk7xACokjHxosumstY:bau54uj) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L48)):*

Remind the reader what scrubTable is.
</blockquote></details>

Thanks for nothing this. I've changed the footnote to clarify this.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 51 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwiryqRKg_kUpymAPMx:-Kwk81ySsG68ntnr9yUi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L51)):*

Spell in full: "full outer join". 
</blockquote></details>

Done.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwisd0tYx2-H1oPexAr:-KwkgpYyl1onRGajWAz_:b-tf01t9) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L115)):*

It sounds elegant! 

Unfortunately I don't think you can get away with this RFC without discussing the algorithmic costs of this solution. The RFC should outline (even not in detail) what are the time and memory/storage costs of the proposed solution, and on which nodes these costs are paid.

</blockquote></details>

Good point! This has led me to realize hash joining may actually be preferred.

I've added my assessment of costs at the end of this section, although I'll admit my knowledge is limited for the operations going on or what to focus on.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 156 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwitFQUf-K-8e3T3xTl:-Kwk85bUWkaboMDqAWm8:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L156)):*

The join node needs to be annotated to indicate it's an outer join.
</blockquote></details>

Done.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 185 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KwitJvhqxMINhGQ73vR:-KwkAOqMBGkG1ciWICRi:b-i3v2t9) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L185)):*

I would like the RFC to outline how a full outer join would work in distsql.
</blockquote></details>

I ended up adding a reference to an issue on this because full outer joining is partially implemented for merge join. 

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 219 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwitg7wrkR5_p3To35F:-KwkhJVPtiSW6LvKGlT-:b4iozoq) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L219)):*

Yeah separate error "codes" or identifiers seem useful / necessary here.
</blockquote></details>

Are you suggesting an approach where the error strings have codes, such as
```go
errors.Errorf("single entry value with no default column id")
// becomes
errors.Wrap(errors.Errorf("single entry value with no default column id"), "E101")
```
To clean this up we could use a function to wrap `errors.Wrap` to
identify all errors that are annotated for `SCRUB` purposes, and use a
string enum in place of `"E101"`. Then when we want to translate these
errors to our check error type we can look for the error code prefix.

Alternatively, we can wrap errors in a struct such as
```go
type ScrubError struct {
  Err error
  Code int
}

errors.Errorf("single entry value with no default column id")
// becomes
ScrubError{
  Err: errors.Errorf("single entry value with no default column id"),
  Code: BAD_COLUMN_ID,
}
```

In both cases, I think we only need to add error annotations within
`RowFetcher`. Any errors that we get while calling external functions I
believe fall under an error. e.g. if we get an error from
`UnmarshalColumnValue` it will always be an `INVALID_ENCODING` error.

---

*[docs/RFCS/20171011_scrub_index_implementation.md, line 272 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwits-iEpSZYZrLCfc8:-KwkAdQBQ7BgNYGboebs:bxjcubk) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/20171011_scrub_index_implementation.md#L272)):*

Why do you not need to create a distsqlplanner? 

It seems to me that you'll need one, because the plan needs to be deployed to the nodes that already have a replica of the data.
</blockquote></details>

The idea behind this approach is the `scrubNode` plan node is transformed into a plan captures the entirety of the check logic, where the resulting plan won't even contain a `scrubNode`. As a result, nothing special needs to be done and the plan is executed as any other plan is within `executor.go`.

This differs from 3.b. where there won't be any (or minimal) transformations made to `scrubNode` and instead logic is embedded into `scrubNode.Start` where it will generate and run manually crafted distSQL plans.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

knz:
Reviewed 2 of 2 files at r2.

---

*[docs/RFCS/scrub_index_implementation.md, line 219 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwitg7wrkR5_p3To35F:-KwoGbO59HbCJQuzyxVZ:b-gl3trb) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/scrub_index_implementation.md#L219)):*

Are you suggesting an approach where the error strings have codes, such as
```go
errors.Errorf("single entry value with no default column id")
// becomes
errors.Wrap(errors.Errorf("single entry value with no default column id"), "E101")
```
To clean this up we could use a function to wrap `errors.Wrap` to
identify all errors that are annotated for `SCRUB` purposes, and use a
string enum in place of `"E101"`. Then when we want to translate these
errors to our check error type we can look for the error code prefix.

Alternatively, we can wrap errors in a struct such as
```go
type ScrubError struct {
  Err error
  Code int
}

errors.Errorf("single entry value with no default column id")
// becomes
ScrubError{
  Err: errors.Errorf("single entry value with no default column id"),
  Code: BAD_COLUMN_ID,
}
```

In both cases, I think we only need to add error annotations within
`RowFetcher`. Any errors that we get while calling external functions I
believe fall under an error. e.g. if we get an error from
`UnmarshalColumnValue` it will always be an `INVALID_ENCODING` error.
</blockquote></details>

The latter with a struct is the way to go. I would recommend you consider using `pgerror.Error` for this purpose.

---

*[docs/RFCS/scrub_index_implementation.md, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwo7qKiMHKh14XobpYc:-Kwo7qKjjFnYeDGjBFvU:b-uy8sr8) ([raw file](https://github.com/cockroachdb/cockroach/blob/8ab45b8dc87d8d8991801e1f7b6d5534514c5e17/docs/RFCS/scrub_index_implementation.md#L90)):*
> ```Markdown
>   during execution of our original plan, and proceeds to run them.
> 
> - If we follow the first solition, the entire logical plan needs to
> ```

solution*

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

lego:

---

*[docs/RFCS/scrub_index_implementation.md, line 219 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwitg7wrkR5_p3To35F:-KwouOFM3ebchT0sr9ZB:b-jnmmv9) ([raw file](https://github.com/cockroachdb/cockroach/blob/48ceb9818a0e43b6524a83819df6e251d896e415/docs/RFCS/scrub_index_implementation.md#L219)):*

The latter with a struct is the way to go. I would recommend you consider using `pgerror.Error` for this purpose.
</blockquote></details>

:+1:. Added details to the doc.

---

*[docs/RFCS/scrub_index_implementation.md, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-Kwo7qKiMHKh14XobpYc:-Kwoudu4bpANMb-VnWz5:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8ab45b8dc87d8d8991801e1f7b6d5534514c5e17/docs/RFCS/scrub_index_implementation.md#L90)):*

solution*
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

knz:
LGTM. However I think you should get a review by @RaduBerinde too.

---

Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-:-KxEYwcNmM9IKxmRvl8W:b-tkuyvj)*

RaduBerinde:
Sorry for the late response. This looks good to me overall!

---

---

*[docs/RFCS/scrub_index_implementation.md, line 178 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF7-brFi2yqQKJK14H:-KxF7-brFi2yqQKJK14I:bdw9sxy) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L178)):*
> ```Markdown
> ```
> 
> Then we can express our intents as the logical plan represented by:
> ```

let's not use "intents" :)

---

*[docs/RFCS/scrub_index_implementation.md, line 186 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF6v3YMqylih1gO4z3:-KxF6v3YMqylih1gO4z4:b96otdv) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L186)):*
> ```Markdown
>   t1@name_idx as s
> WHERE
>   p.pkey IS NULL AND s.name IS NULL
> ```

I don't get this condition.. The key from one side is NULL and the name from the other side is NULL? Seems like it should be `(p.pkey IS NULL AND p.name IS NULL) OR (s.pkey IS NULL AND s.name IS NULL)`

BTW We will implement more join types that are useful for various queries; one type is "anti-join" and I think it's exactly what we need here. Once we have it, hopefully the (not-yet-existing) optimizer can figure out that's what this query needs and uses that (or we can introduce an internal "anti join" statement)

---

*[docs/RFCS/scrub_index_implementation.md, line 237 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF8B0hMt1BGCjV8VS7:-KxF8B0hMt1BGCjV8VS8:bsiwf3w) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L237)):*
> ```Markdown
> exactly N+M rows and stream them to a joiner.
> 
> If we do a hash join, then we can also distribute the join processor
> ```

The distsql planning code currently always plans distributed hash joins (each node that has a stream has a HashJoiner). The HashJoiner uses temp storage if necessary.

---

*[docs/RFCS/scrub_index_implementation.md, line 261 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF8_tk3cvincDi3mb3:-KxF8_tk3cvincDi3mb4:b-j8yo00) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L261)):*
> ```Markdown
>      [#17135](https://github.com/cockroachdb/cockroach/issues/17135)
> 
> [distSQL plan]: https://cockroachdb.github.io/distsqlplan/decode.html?eJyskjFr-zAQxff_pxC3ZPjfENuZBAVNbVOCU5x0Kqao1pEKHMlIMqSEfPdiaUhjkrSBjne633t3D-3BWEWl3JIH_goZ1Aidsw15b93QSgNztQM-RdCm68PQrhEa6wj4HoIOLQGHtXxvqSKpyAGCoiB1G0U7p7fSfYqQAcKyD5yJDEUO9QHB9uGo6IPcEPDsgH_gqtWbVrtfm-a3mK6sC2M_kf2_KF5cFD9q9sY6RY7UiWw9kD-NnNnwUfqPJ6vNeMtlyUTG7pgoAOFet4EcZ-XLYgEIFRk1lJMU3YRzvlpX8_IBh-yYyDFOIhMFMjFL1cWbZ7cEWpHvrPE0vv2s8nQ4mNSGUoDe9q6hZ2ebaJPKZeRiQ5EP6bVIxdzEp_jNvsPZVTg_gadjOL_BOR_DxVV4NnKuD_--AgAA___vMDTp
> ```

This plan is not very helpful.. What use is the sorter before HashJoin? Also, what's with `Filter: NULL`?

I don't think we need to go into this much detail into the specifics of how we run this query - we probably want to rely on the the existing planning code (and later optimizer) to take care of that.

---

*[docs/RFCS/scrub_index_implementation.md, line 402 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF5xcg9LDaxOXVaHMc:-KxF5xchyJT2RSc5Qx_P:b-exyqpt) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L402)):*
> ```Markdown
> Using query execution to the bulk of the logic involves a similar
> strategy as schema change index backfilling. Within local execution, we
> need to initialize a `distSQLPlanner` to plan and run a distSQL plan. It
> ```

Why do we need a new `distSQLPlanner`? It was meant to be a singleton, we should use the one in executor (schema changes should use that one too).

This approach seems more straightforward compared to using a new plan node.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-:-KxFAK9lSpOCURSQlpS6:bd9dgrv)*

lego:

---

*[docs/RFCS/scrub_index_implementation.md, line 178 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF7-brFi2yqQKJK14H:-KxInNqF2ukTzAJPkTGO:b4b5c01) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L178)):*

let's not use "intents" :)
</blockquote></details>

Not the intent of that word :P. Done.

---

*[docs/RFCS/scrub_index_implementation.md, line 186 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF6v3YMqylih1gO4z3:-KxInZg9AoebfAHw4SDj:b-oqcv3h) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L186)):*

I don't get this condition.. The key from one side is NULL and the name from the other side is NULL? Seems like it should be `(p.pkey IS NULL AND p.name IS NULL) OR (s.pkey IS NULL AND s.name IS NULL)`

BTW We will implement more join types that are useful for various queries; one type is "anti-join" and I think it's exactly what we need here. Once we have it, hopefully the (not-yet-existing) optimizer can figure out that's what this query needs and uses that (or we can introduce an internal "anti join" statement)
</blockquote></details>

Ah, yep, the condition is incorrect.

---

*[docs/RFCS/scrub_index_implementation.md, line 237 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF8B0hMt1BGCjV8VS7:-KxIsMywtoqqSESRh0Lv:b-ucfknh) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L237)):*

The distsql planning code currently always plans distributed hash joins (each node that has a stream has a HashJoiner). The HashJoiner uses temp storage if necessary.
</blockquote></details>

Thanks!

---

*[docs/RFCS/scrub_index_implementation.md, line 261 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF8_tk3cvincDi3mb3:-KxIqa_4bpKI24HjrV62:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L261)):*

This plan is not very helpful.. What use is the sorter before HashJoin? Also, what's with `Filter: NULL`?

I don't think we need to go into this much detail into the specifics of how we run this query - we probably want to rely on the the existing planning code (and later optimizer) to take care of that.
</blockquote></details>

:+1:

---

*[docs/RFCS/scrub_index_implementation.md, line 402 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KxF5xcg9LDaxOXVaHMc:-KxIpPDYtKSIiYAIS55_:bqg6rs1) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b71c4c948ff131152643452033ccc8d0054c11b/docs/RFCS/scrub_index_implementation.md#L402)):*

Why do we need a new `distSQLPlanner`? It was meant to be a singleton, we should use the one in executor (schema changes should use that one too).

This approach seems more straightforward compared to using a new plan node.
</blockquote></details>

Gotcha. I assumed that was not the case because of the schema changes.

I also didn't notice the executor passes it through to`planner.session.distSQLPlanner`! This is great, as it's now pretty easy to proceed with this knowing that :).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

RaduBerinde:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-:-KxJ784_c46KDKtfJe9G:bnfp4nl)*

lego:
Sorry about letting the dust settle on this one. I largely left it open while working on the implementation to still figure out some things before demoting some parts of the RFC to alternatives.

If :thumbsup: I'll stamp the draft and merge it in. Currently the implementation PR is open at https://github.com/cockroachdb/cockroach/pull/19961, so there will be a few updates to reflect the final direction once that PR settles.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-:-KyqhrZ-cx-SUR9cfCx3:b-xvfj5j)*

RaduBerinde:
*[docs/RFCS/scrub_index_implementation.md, line 216 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrKNKiwZFV4p5fo7Ph:-KyrKNKj1kiK0LO7TCoD:b99l9o) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a03897bdd043a7f4004bc96158b5dc24a8c15de/docs/RFCS/scrub_index_implementation.md#L216)):*
> ```Markdown
> If we instead do a merge join, then each node scanning data will also
> need to sort the secondary k/v rows before they are streamed to the
> merge join processor. The merge join is also monolithic and can't be
> ```

merge joins are distributed too, in exactly the same way we distribute hash joins

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

RaduBerinde:
*[docs/RFCS/scrub_index_implementation.md, line 314 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrL5frsB49xLw-gBXj:-KyrL5frsB49xLw-gBXk:b-xq90vw) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a03897bdd043a7f4004bc96158b5dc24a8c15de/docs/RFCS/scrub_index_implementation.md#L314)):*
> ```Markdown
> There are two approaches for how execution can be done.
> ```

1 is missing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

lego:

---

*[docs/RFCS/scrub_index_implementation.md, line 216 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrKNKiwZFV4p5fo7Ph:-KyrOscILeewNC-NFWJh:b7i5nqp) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a03897bdd043a7f4004bc96158b5dc24a8c15de/docs/RFCS/scrub_index_implementation.md#L216)):*

merge joins are distributed too, in exactly the same way we distribute hash joins
</blockquote></details>

TIL! I assumed not. The whole route by hash thing did not seem intuitive for a sorted input and especially mergejoiner. But it just clicked that this is only for distributing the values and they will still be processed in order and still distributed in the case of mergejoiner! Whoa, awesome!

---

*[docs/RFCS/scrub_index_implementation.md, line 314 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrL5frsB49xLw-gBXj:-KyrOoMwDNBTMq5bbTKX:b11l6ol) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a03897bdd043a7f4004bc96158b5dc24a8c15de/docs/RFCS/scrub_index_implementation.md#L314)):*

1 is missing
</blockquote></details>

Oops, fixed. 1 was moved to alternatives but I forgot to fix this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

RaduBerinde:
*[docs/RFCS/scrub_index_implementation.md, line 216 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrKNKiwZFV4p5fo7Ph:-KyrSLucl4Zep1gKhgjF:bargtnp) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a03897bdd043a7f4004bc96158b5dc24a8c15de/docs/RFCS/scrub_index_implementation.md#L216)):*

TIL! I assumed not. The whole route by hash thing did not seem intuitive for a sorted input and especially mergejoiner. But it just clicked that this is only for distributing the values and they will still be processed in order and still distributed in the case of mergejoiner! Whoa, awesome!
</blockquote></details>

Yup, exactly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

RaduBerinde:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/scrub_index_implementation.md, line 296 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrSTYPLTEK0luNTIlq:-KyrSTYPLTEK0luNTIlr:bd6eqhp) ([raw file](https://github.com/cockroachdb/cockroach/blob/a96ef9c093d8828c1a49c64ba806ed1bfdf20b45/docs/RFCS/scrub_index_implementation.md#L296)):*
> ```Markdown
> ## 3. Generating and executing the logical plan in the distSQL execution engine
> 
> During local engine execution we will generate the the necessary check
> ```

the the

---

*[docs/RFCS/scrub_index_implementation.md, line 298 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrSVylAf6VCAnfIZEI:-KyrSVylAf6VCAnfIZEJ:bidpc92) ([raw file](https://github.com/cockroachdb/cockroach/blob/a96ef9c093d8828c1a49c64ba806ed1bfdf20b45/docs/RFCS/scrub_index_implementation.md#L298)):*
> ```Markdown
> During local engine execution we will generate the the necessary check
> plans, then execute the plans in sequence on the distSQL execution
> engine. These plans will be the ones such that will scan the data in the
> ```

These plans will scan the data

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327)*

lego:
I should have stated this more formally yesterday, but this RFC is moving into the FCP until Thursday.

(I will admit it is more lightweight of an RFC so I didn't quite consider that, as this is mostly a proposal of how to go about this which is already underway)

---

---

*[docs/RFCS/scrub_index_implementation.md, line 296 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrSTYPLTEK0luNTIlq:-KyvAaidEK_A1zytteQQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a96ef9c093d8828c1a49c64ba806ed1bfdf20b45/docs/RFCS/scrub_index_implementation.md#L296)):*

the the
</blockquote></details>

Done.

---

*[docs/RFCS/scrub_index_implementation.md, line 298 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-KyrSVylAf6VCAnfIZEI:-KyvAazNOJJC6nFcq4OU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a96ef9c093d8828c1a49c64ba806ed1bfdf20b45/docs/RFCS/scrub_index_implementation.md#L298)):*

These plans will scan the data
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19327#-:-KyvAvLnKLGfcCauzcIK:b-yokmm4)*

PR #19339, opened 2017-10-18
engine: Fix splitting to work on ranges with a very large first row

### LINKED ISSUE #19296: storage: ranges not being properly split after drastically reducing max range size
I had been gradually decreasing my cluster's max range size while trying to reproduce https://github.com/cockroachdb/cockroach/issues/19236, going from 64MiB to 8Mib to 4Mib to 2Mib before trying to drop down to 64Kib. Splits happened as expected all the way down to 2Mib, leaving the cluster with about 850 ranges, but didn't happen for the last drop down to 64Kib. I can see from the admin UI that the split queue processed a bunch of replicas, but the number of ranges in the cluster didn't actually change, meaning no splits were actually carried out.

1. This is basically impossible to debug without modifying the binary, unless I'm missing something. There aren't even any verbose tracing messages in the split queue.
2. I don't see why splits wouldn't happen, given that the ranges I've looked at are much larger than 64KiB (most appear to be around 450KB), and there are multiple logical keys in each range (given the load generator involved in #19236, there are `26^3`, or `17576`, keys).

a-robinson:
Previously, if the target split point was contained within the first SQL
row of a range, MVCCFindSplitKey would return a key within that first
row, which would prevent a split from happening because we only split on
row boundaries (and don't want to split on the first key in a range).
Now, we'll continue iterating until we find the first key that isn't
part of the first row in such ranges.

Fixes #19296 

Thanks to @jordanlewis for helping me figure out how to create keys to test this with.

### DIFF
--- a/pkg/storage/engine/mvcc_test.go
+++ b/pkg/storage/engine/mvcc_test.go
@@ -37,6 +37,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/zerofields"
 	"github.com/cockroachdb/cockroach/pkg/util/caller"
+	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
@@ -3129,11 +3130,30 @@ func TestFindSplitKey(t *testing.T) {
 // they avoid splits through invalid key ranges.
 func TestFindValidSplitKeys(t *testing.T) {
 	defer leaktest.AfterTest(t)()
+
+	// Manually creates rows corresponding to the schema:
+	// CREATE TABLE t (id STRING PRIMARY KEY, col INT)
+	encodeTableKey := func(rowVal string, colFam uint32) roachpb.Key {
+		tableKey := keys.MakeTablePrefix(keys.MaxReservedDescID + 1)
+		rowKey := roachpb.Key(encoding.EncodeVarintAscending(append([]byte(nil), tableKey...), 1))
+		rowKey = encoding.EncodeStringAscending(encoding.EncodeVarintAscending(rowKey, 1), rowVal)
+		colKey := keys.MakeFamilyKey(append([]byte(nil), rowKey...), colFam)
+		return colKey
+	}
+	splitKeyFromTableKey := func(tableKey roachpb.Key) roachpb.Key {
+		splitKey, err := keys.EnsureSafeSplitKey(tableKey)
+		if err != nil {
+			t.Fatal(err)
+		}
+		return splitKey
+	}
+
 	rangeID := roachpb.RangeID(1)
 	testCases := []struct {
-		keys     []roachpb.Key
-		expSplit roachpb.Key
-		expError bool
+		keys       []roachpb.Key
+		rangeStart roachpb.Key // optional
+		expSplit   roachpb.Key
+		expError   bool
 	}{
 		// All m1 cannot be split.
 		{
@@ -3212,6 +3232,53 @@ func TestFindValidSplitKeys(t *testing.T) {
 			expSplit: nil,
 			expError: false,
 		},
+		// Some example table data. Make sure we don't split in the middle of a row
+		// or return the start key of the range.
+		{
+			keys: []roachpb.Key{
+				encodeTableKey("a", 1),
+				encodeTableKey("a", 2),
+				encodeTableKey("a", 3),
+				encodeTableKey("a", 4),
+				encodeTableKey("a", 5),
+				encodeTableKey("b", 1),
+				encodeTableKey("c", 1),
+			},
+			rangeStart: splitKeyFromTableKey(encodeTableKey("a", 1)),
+			expSplit:   splitKeyFromTableKey(encodeTableKey("b", 1)),
+			expError:   false,
+		},
+		// More example table data. Make sure ranges at the start of a table can
+		// be split properly - this checks that the minSplitKey logic doesn't
+		// break for such ranges.
+		{
+			keys: []roachpb.Key{
+				encodeTableKey("a", 1),
+				encodeTableKey("b", 1),
+				encodeTableKey("c", 1),
+				encodeTableKey("d", 1),
+			},
+			rangeStart: keys.MakeTablePrefix(keys.MaxReservedDescID + 1),
+			expSplit:   splitKeyFromTableKey(encodeTableKey("c", 1)),
+			expError:   false,
+		},
+		// More example table data. Make sure ranges at the start of a table can
+		// be split properly (even if "properly" means creating an empty LHS,
+		// splitting here will at least allow the resulting RHS to split again).
+		{
+			keys: []roachpb.Key{
+				encodeTableKey("a", 1),
+				encodeTableKey("a", 2),
+				encodeTableKey("a", 3),
+				encodeTableKey("a", 4),
+				encodeTableKey("a", 5),
+				encodeTableKey("b", 1),
+				encodeTableKey("c", 1),
+			},
+			rangeStart: keys.MakeTablePrefix(keys.MaxReservedDescID + 1),
+			expSplit:   splitKeyFromTableKey(encodeTableKey("a", 1)),
+			expError:   false,
+		},
 	}
 
 	for i, test := range testCases {
@@ -3231,6 +3298,9 @@ func TestFindValidSplitKeys(t *testing.T) {
 				t.Fatal(err)
 			}
 			rangeStart := test.keys[0]
+			if len(test.rangeStart) > 0 {
+				rangeStart = test.rangeStart
+			}
 			rangeEnd := test.keys[len(test.keys)-1].Next()
 			rangeStartAddr, err := keys.Addr(rangeStart)
 			if err != nil {
--- a/c-deps/libroach/db.cc
+++ b/c-deps/libroach/db.cc
@@ -2420,12 +2420,13 @@ bool MVCCIsValidSplitKey(DBSlice key, bool allow_meta2_splits) {
   return IsValidSplitKey(ToSlice(key), allow_meta2_splits);
 }
 
-DBStatus MVCCFindSplitKey(DBIterator* iter, DBKey start, DBKey end, 
+DBStatus MVCCFindSplitKey(DBIterator* iter, DBKey start, DBKey end, DBKey min_split,
                           int64_t target_size, bool allow_meta2_splits, DBString* split_key) {
   auto iter_rep = iter->rep.get();
   const std::string start_key = EncodeKey(start);
   iter_rep->Seek(start_key);
   const std::string end_key = EncodeKey(end);
+  const rocksdb::Slice min_split_key = ToSlice(min_split.key);
 
   int64_t size_so_far = 0;
   std::string best_split_key = start_key;
@@ -2444,7 +2445,7 @@ DBStatus MVCCFindSplitKey(DBIterator* iter, DBKey start, DBKey end,
     }
 
     ++n;
-    const bool valid = n > 1 && IsValidSplitKey(decoded_key, allow_meta2_splits);
+    const bool valid = n > 1 && IsValidSplitKey(decoded_key, allow_meta2_splits) && decoded_key.compare(min_split_key) >= 0;
     int64_t diff = target_size - size_so_far;
     if (diff < 0) {
       diff = -diff;
@@ -2453,7 +2454,10 @@ DBStatus MVCCFindSplitKey(DBIterator* iter, DBKey start, DBKey end,
       best_split_key = decoded_key.ToString();
       best_split_diff = diff;
     }
-    if (diff > best_split_diff) {
+    // If diff is increasing, that means we've passed the ideal split point and
+    // should return the first key that we can. Note that best_split_key may
+    // still be empty if we haven't reached min_split_key yet.
+    if (diff > best_split_diff && !best_split_key.empty()) {
       break;
     }
 
--- a/c-deps/libroach/include/libroach.h
+++ b/c-deps/libroach/include/libroach.h
@@ -222,7 +222,7 @@ typedef struct {
 MVCCStatsResult MVCCComputeStats(DBIterator* iter, DBKey start, DBKey end, int64_t now_nanos);
 
 bool MVCCIsValidSplitKey(DBSlice key, bool allow_meta2_splits);
-DBStatus MVCCFindSplitKey(DBIterator* iter, DBKey start, DBKey end, 
+DBStatus MVCCFindSplitKey(DBIterator* iter, DBKey start, DBKey end, DBKey min_split,
                           int64_t target_size, bool allow_meta2_splits, DBString* split_key);
 
 // DBStatsResult contains various runtime stats for RocksDB.
--- a/pkg/storage/engine/engine.go
+++ b/pkg/storage/engine/engine.go
@@ -100,7 +100,7 @@ type Iterator interface {
 	// allowMeta2Splits is false.
 	//
 	// TODO: remove allowMeta2Splits in version 1.3.
-	FindSplitKey(start, end MVCCKey, targetSize int64, allowMeta2Splits bool) (MVCCKey, error)
+	FindSplitKey(start, end, minSplitKey MVCCKey, targetSize int64, allowMeta2Splits bool) (MVCCKey, error)
 }
 
 // Reader is the read interface to an engine's data.
--- a/pkg/storage/engine/mvcc.go
+++ b/pkg/storage/engine/mvcc.go
@@ -2284,9 +2284,21 @@ func MVCCFindSplitKey(
 	it := engine.NewIterator(false /* prefix */)
 	defer it.Close()
 
+	minSplitKey := key
+	// If this is a table, we can only split at row boundaries.
+	if remainder, _, err := keys.DecodeTablePrefix(roachpb.Key(key)); err == nil {
+		// If this is the first range containing a table, its start key won't
+		// actually contain any row information, just the table ID. We don't want
+		// to restrict splits on such tables, since key.PrefixEnd will just be the
+		// end of the table span.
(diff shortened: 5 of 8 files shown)


petermattis:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[c-deps/libroach/db.cc, line 2462 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-Kwke18dPVAWvT3pkHHc:-Kwke18dPVAWvT3pkHHd:b-z7cy2z) ([raw file](https://github.com/cockroachdb/cockroach/blob/63ccaf20bb82e8d6746f803f16e297eaa5666b70/c-deps/libroach/db.cc#L2462)):*
> ```cpp
>       // minimum split key, keep going until we find the first key at
>       // least as large as min_split.
>       if (kComparator.Compare(best_split_key, ToSlice(min_split.key)) >= 0) {
> ```

`kComparator` is used for comparing undecoded keys. I'm surprised this is working. Or am I missing something?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

a-robinson:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[c-deps/libroach/db.cc, line 2462 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-Kwke18dPVAWvT3pkHHc:-Kwkekp6PgMfaqNKEhnu:bqvzh84) ([raw file](https://github.com/cockroachdb/cockroach/blob/63ccaf20bb82e8d6746f803f16e297eaa5666b70/c-deps/libroach/db.cc#L2462)):*

`kComparator` is used for comparing undecoded keys. I'm surprised this is working. Or am I missing something?
</blockquote></details>

I'm glad you commented on this. I believe it's working by [falling back to the direct slice comparison](https://github.com/cockroachdb/cockroach/blob/master/c-deps/libroach/db.cc#L490), but it's clearly a bit silly to rely on that.

Changed to just using a direct comparison.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

bdarnell:
Reviewed 6 of 7 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[c-deps/libroach/db.cc, line 2467 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwkpI2rr-kbSUCz5RPU:-KwkpI2rr-kbSUCz5RPV:b-ual2cj) ([raw file](https://github.com/cockroachdb/cockroach/blob/b10d2905cec805c6e4a65748ac1365f2cdccb912/c-deps/libroach/db.cc#L2467)):*
> ```cpp
>       }
>       if (valid && decoded_key.compare(min_split_key) >= 0) {
>         best_split_key = decoded_key.ToString();
> ```

The structure of this code (with two different comparisons to min_split key, and two assignments `best_split_key = decoded_key.ToString()`) looks weird. Can we refactor to make this less repetitive?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

a-robinson:

---

*[c-deps/libroach/db.cc, line 2467 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwkpI2rr-kbSUCz5RPU:-KwoqdsEzO89k6ZqHmYW:btp0hz3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b10d2905cec805c6e4a65748ac1365f2cdccb912/c-deps/libroach/db.cc#L2467)):*

The structure of this code (with two different comparisons to min_split key, and two assignments `best_split_key = decoded_key.ToString()`) looks weird. Can we refactor to make this less repetitive?
</blockquote></details>

I can remove the second comparison and assignment, but I wouldn't say it's clearer:

```
    const bool valid = n > 1 && IsValidSplitKey(decoded_key, allow_meta2_splits);
    int64_t diff = target_size - size_so_far;
    if (diff < 0) {
      diff = -diff;
    }
    // If diff is increasing, that means we've passed the ideal split point and
    // should return the first key that we can.
    if (diff > best_split_diff) {
      // Make sure the key being returned is valid. If it's smaller than the
      // minimum split key, keep going until we find the first key at
      // least as large as min_split_key.
      if (best_split_key.compare(min_split_key) >= 0) {
        break;
      }
    }
    if (valid && ((diff < best_split_diff) || (diff > best_split_diff && decoded_key.compare(min_split_key) >= 0))) {
      best_split_key = decoded_key.ToString();
      best_split_diff = diff;
    }
```

I'm not sure we can break it down any more than that without completely rethinking things or changing behavior. Do you see anything?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

pkg/storage/engine/mvcc_test.go
```diff
@@ -3222,6 +3287,11 @@ func TestFindValidSplitKeys(t *testing.T) {
 			ms := &enginepb.MVCCStats{}
 			val := roachpb.MakeValueFromString(strings.Repeat("X", 10))
 			for _, k := range test.keys {
+				if i == 7 {
```
danhhz: why is 7 special cased? deserves a comment if this is intentional

a-robinson: Heh, that definitely wasn't supposed to be left in. Thanks!

petermattis:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[c-deps/libroach/db.cc, line 2429 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpHG8u1M6yOZL-jQYb:-KwpHG8vse45otpuY--A:b2qtshu) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a386a682f9e18cb065ff76ee93dafac7ba8081b/c-deps/libroach/db.cc#L2429)):*
> ```cpp
>   iter_rep->Seek(start_key);
>   const std::string end_key = EncodeKey(end);
>   const std::string min_split_key = ToString(min_split.key);
> ```

I think you could change this to:

```
const rocksdb::Slice min_split_key = ToSlice(min_split.key);
```

Almost certainly doesn't matter for performance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

a-robinson:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[c-deps/libroach/db.cc, line 2429 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpHG8u1M6yOZL-jQYb:-KwpNfwiIhjdQf-bWPG3:b-mf2f5f) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a386a682f9e18cb065ff76ee93dafac7ba8081b/c-deps/libroach/db.cc#L2429)):*

I think you could change this to:

```
const rocksdb::Slice min_split_key = ToSlice(min_split.key);
```

Almost certainly doesn't matter for performance.
</blockquote></details>

It looks pretty free given https://github.com/facebook/rocksdb/blob/31d3e4181099bd92ca5c1be0ab8b7f51765fa558/include/rocksdb/slice.h#L241 and https://github.com/facebook/rocksdb/blob/31d3e4181099bd92ca5c1be0ab8b7f51765fa558/include/rocksdb/slice.h#L42. We can't change it because currently we're comparing `min_split_key`to `decoded_key`, which is a string and thus doesn't have a direct `compare` operation on a `rocksdb::Slice`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

nvanbenschoten:
Reviewed 4 of 7 files at r1, 2 of 2 files at r4.
Review status: 7 of 8 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[c-deps/libroach/db.cc, line 2467 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwkpI2rr-kbSUCz5RPU:-KwpzCbnJDYljYZKRyk6:bpdi798) ([raw file](https://github.com/cockroachdb/cockroach/blob/b10d2905cec805c6e4a65748ac1365f2cdccb912/c-deps/libroach/db.cc#L2467)):*

I can remove the second comparison and assignment, but I wouldn't say it's clearer:

```
    const bool valid = n > 1 && IsValidSplitKey(decoded_key, allow_meta2_splits);
    int64_t diff = target_size - size_so_far;
    if (diff < 0) {
      diff = -diff;
    }
    // If diff is increasing, that means we've passed the ideal split point and
    // should return the first key that we can.
    if (diff > best_split_diff) {
      // Make sure the key being returned is valid. If it's smaller than the
      // minimum split key, keep going until we find the first key at
      // least as large as min_split_key.
      if (best_split_key.compare(min_split_key) >= 0) {
        break;
      }
    }
    if (valid && ((diff < best_split_diff) || (diff > best_split_diff && decoded_key.compare(min_split_key) >= 0))) {
      best_split_key = decoded_key.ToString();
      best_split_diff = diff;
    }
```

I'm not sure we can break it down any more than that without completely rethinking things or changing behavior. Do you see anything?
</blockquote></details>

If you pulled the `diff < best_split_diff` into a `diff_shrinking` variable, I'd find your second approach cleaner. Something like:
```
int64_t diff = target_size - size_so_far;
if (diff < 0) {
    diff = -diff;
}
bool diff_shrinking = diff < best_split_diff;
// If diff is increasing or staying the same, that means we've passed the ideal
// split point and should return the first key that we can.
if (!diff_shrinking) {
    // Make sure the key being returned is valid. If it's smaller than the
    // minimum split key, keep going until we find the first key at
    // least as large as min_split_key.
    if (best_split_key.compare(min_split_key) >= 0) {
        break;
    }
}
if (valid && (diff_shrinking|| decoded_key.compare(min_split_key) >= 0)) {
    best_split_key = decoded_key.ToString();
    best_split_diff = diff;
}
```

---

*[pkg/storage/engine/mvcc.go, line 2294 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-Kwpvg5wUmEJHXNfhivV:-Kwpvg5wUmEJHXNfhivW:b-dc4zsp) ([raw file](https://github.com/cockroachdb/cockroach/blob/472a7ea1d2bccb2f8b601b2fbafe0b37a0aeb363/pkg/storage/engine/mvcc.go#L2294)):*
> ```Go
> 		// to restrict splits on such tables, since key.PrefixEnd will just be the
> 		// end of the table span.
> 		if len(remainder) > 0 {
> ```

`len(remainder) == 0` is what we fall into in the third new test case, right? Or is it the second and third?

---

*[pkg/storage/engine/mvcc_test.go, line 3154 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpvQn0oaZVyl9HT0kJ:-KwpvQn1PqIzQ-N6-8qf:b-d6wy6e) ([raw file](https://github.com/cockroachdb/cockroach/blob/472a7ea1d2bccb2f8b601b2fbafe0b37a0aeb363/pkg/storage/engine/mvcc_test.go#L3154)):*
> ```Go
> 	testCases := []struct {
> 		keys       []roachpb.Key
> 		rangeStart roachpb.Key
> ```

Let's add a comment that this isn't required.

---

*[pkg/storage/engine/mvcc_test.go, line 3265 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpvHba_nrr6jDKLjsc:-KwpvHbblaTKYsEEwPeI:bi9pqg2) ([raw file](https://github.com/cockroachdb/cockroach/blob/472a7ea1d2bccb2f8b601b2fbafe0b37a0aeb363/pkg/storage/engine/mvcc_test.go#L3265)):*
> ```Go
> 		// More example table data. Make sure ranges at the start of a table can
> 		// be split properly (even if "properly" means creating an empty LHS,
> 		// splitting here will at least allow the resulting RHS to split again).
> ```

Could you describe here why this is treated differently than the case before it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

bdarnell:
Reviewed 1 of 2 files at r4, 1 of 1 files at r5.

---

*[c-deps/libroach/db.cc, line 2467 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwkpI2rr-kbSUCz5RPU:-Kwul6ZITrrrrA3_Ogyn:b-qjjgg4) ([raw file](https://github.com/cockroachdb/cockroach/blob/b10d2905cec805c6e4a65748ac1365f2cdccb912/c-deps/libroach/db.cc#L2467)):*

If you pulled the `diff < best_split_diff` into a `diff_shrinking` variable, I'd find your second approach cleaner. Something like:
```
int64_t diff = target_size - size_so_far;
if (diff < 0) {
    diff = -diff;
}
bool diff_shrinking = diff < best_split_diff;
// If diff is increasing or staying the same, that means we've passed the ideal
// split point and should return the first key that we can.
if (!diff_shrinking) {
    // Make sure the key being returned is valid. If it's smaller than the
    // minimum split key, keep going until we find the first key at
    // least as large as min_split_key.
    if (best_split_key.compare(min_split_key) >= 0) {
        break;
    }
}
if (valid && (diff_shrinking|| decoded_key.compare(min_split_key) >= 0)) {
    best_split_key = decoded_key.ToString();
    best_split_diff = diff;
}
```
</blockquote></details>

Can `min_split_key` simply be part of the `valid` condition? It's not clear to me why it's being handled separately. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

a-robinson:
Review status: 5 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[c-deps/libroach/db.cc, line 2467 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwkpI2rr-kbSUCz5RPU:-KwupkUelFHvPkeA0Zws:b20eml3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b10d2905cec805c6e4a65748ac1365f2cdccb912/c-deps/libroach/db.cc#L2467)):*
> Can min_split_key simply be part of the valid condition? It's not clear to me why it's being handled separately.

I was worried about always doing the string comparison in code that was apparently performance-sensitive enough to justify moving it into C++, but if the `MVCCFindSplitKey_RocksDB` benchmark is representative it actually doesn't make a difference, so I've switched over to that. Thanks!

---

*[c-deps/libroach/db.cc, line 2429 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpHG8u1M6yOZL-jQYb:-KwutXf72JofYJypuY_8:b6anb6) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a386a682f9e18cb065ff76ee93dafac7ba8081b/c-deps/libroach/db.cc#L2429)):*

It looks pretty free given https://github.com/facebook/rocksdb/blob/31d3e4181099bd92ca5c1be0ab8b7f51765fa558/include/rocksdb/slice.h#L241 and https://github.com/facebook/rocksdb/blob/31d3e4181099bd92ca5c1be0ab8b7f51765fa558/include/rocksdb/slice.h#L42. We can't change it because currently we're comparing `min_split_key`to `decoded_key`, which is a string and thus doesn't have a direct `compare` operation on a `rocksdb::Slice`.
</blockquote></details>

I was able to change this now, after simplifying the code below a bit.

---

*[pkg/storage/engine/mvcc.go, line 2294 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-Kwpvg5wUmEJHXNfhivV:-Kwu-hoSWZngqrZ4KAbB:bq9rno6) ([raw file](https://github.com/cockroachdb/cockroach/blob/472a7ea1d2bccb2f8b601b2fbafe0b37a0aeb363/pkg/storage/engine/mvcc.go#L2294)):*

`len(remainder) == 0` is what we fall into in the third new test case, right? Or is it the second and third?
</blockquote></details>

Both the second and third. Any time the range's start key is a table prefix (e.g. `/Table/50`) instead of a row prefix (e.g. `/Table/50/1/1/"a"`).

---

*[pkg/storage/engine/mvcc_test.go, line 3154 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpvQn0oaZVyl9HT0kJ:-Kwu07OHPJ8dxgjwgX4e:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/472a7ea1d2bccb2f8b601b2fbafe0b37a0aeb363/pkg/storage/engine/mvcc_test.go#L3154)):*

Let's add a comment that this isn't required.
</blockquote></details>

Done.

---

*[pkg/storage/engine/mvcc_test.go, line 3265 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-KwpvHba_nrr6jDKLjsc:-Kwu0dc8GIOwY_O9iH-B:b-cwvxd2) ([raw file](https://github.com/cockroachdb/cockroach/blob/472a7ea1d2bccb2f8b601b2fbafe0b37a0aeb363/pkg/storage/engine/mvcc_test.go#L3265)):*

Could you describe here why this is treated differently than the case before it?
</blockquote></details>

The prior case perhaps be removed in favor of this one, since they both make sure we don't construct a `minSplitKey` that rules out all keys in the range (the concern being that `/Table/50`.PrefixNext() might be greater than all rows in the range). This one just adds on the wrinkle that we *are* allowed to split on the first row of this kind of range. I've updated the comment on the prior case to try to clarify.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339)*

bdarnell:
:lgtm:

---

Reviewed 1 of 3 files at r6, 1 of 1 files at r7.
Review status: 7 of 8 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-:-KwuuS3_Pk5rsmt-RZDN:bnfp4nl)*

petermattis:
:lgtm:

---

Review status: 7 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19339#-:-Kww6mqSm55sWPww1p5t:bnfp4nl)*

PR #19351, opened 2017-10-18
server: avoid generating 2MB of stack traces during `go test -v`

benesch:


tbg (approved):
Thanks! 💯

tschottdorf:
cc @andreimatei I think you filed an issue about this that I now don't seem to find.

andreimatei:
I remember my complain was about some change to some testing infrastructure that you were doing at the time, Tobi, that had as an effect that when it fired, it'd make it really hard to see which test was the one encountering the problem. But I can't remember what it was...

benesch:
Ok, merging! TFTR

PR #19353, opened 2017-10-18
storage: mark replicas added to replica GC queue as destroyed

m-schneider:
Before when adding replicas to the GC queue, they weren't fully considered
destroyed. This lead to redirectOnOrAcquireLease waiting for the
replica to be GCed before returning a NotLeaseHolderError.

Now redirectOnOrAcquireLease will respond faster and anything depending
on that such as an RPC will not hang.

### DIFF
--- a/pkg/storage/queue.go
+++ b/pkg/storage/queue.go
@@ -180,6 +180,9 @@ type queueConfig struct {
 	// want to try to replicate a range until we know which zone it is in and
 	// therefore how many replicas are required).
 	acceptsUnsplitRanges bool
+	// processDestroyedReplicas controls whether or not we want to process replicas
+	// that have been destroyed but not GCed.
+	processDestroyedReplicas bool
 	// processTimeout is the timeout for processing a replica.
 	processTimeout time.Duration
 	// successes is a counter of replicas processed successfully.
@@ -599,11 +602,13 @@ func (bq *baseQueue) processReplica(
 		return errors.New("cannot process uninitialized replica")
 	}
 
-	if err := repl.IsDestroyed(); err != nil {
-		if log.V(3) {
-			log.Infof(queueCtx, "replica destroyed (%s); skipping", err)
+	if reason, err := repl.IsDestroyed(); err != nil {
+		if !bq.queueConfig.processDestroyedReplicas || reason != destroyReasonRemovalPending {
+			if log.V(3) {
+				log.Infof(queueCtx, "replica destroyed (%s); skipping", err)
+			}
+			return nil
 		}
-		return nil
 	}
 
 	// If the queue requires a replica to have the range lease in
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -216,6 +216,40 @@ func (d *atomicDescString) String() string {
 	return *(*string)(atomic.LoadPointer(&d.strPtr))
 }
 
+// DestroyReason indicates if a replica is alive, destroyed, corrupted or pending destruction.
+type DestroyReason int
+
+const (
+	// The replica is alive.
+	destroyReasonAlive DestroyReason = iota
+	// The replica has been corrupted.
+	destroyReasonCorrupted
+	// The replica has been marked for GC, but hasn't been GCed yet.
+	destroyReasonRemovalPending
+	// The replica has been GCed.
+	destroyReasonRemoved
+)
+
+type destroyStatus struct {
+	reason DestroyReason
+	err    error
+}
+
+// IsAlive returns true when a replica is alive.
+func (s destroyStatus) IsAlive() bool {
+	return s.reason == destroyReasonAlive
+}
+
+// RemovedOrCorrupt returns true if a replica has either been removed or is corrupted.
+func (s destroyStatus) RemovedOrCorrupt() bool {
+	return (s.reason == destroyReasonCorrupted) || (s.reason == destroyReasonRemoved)
+}
+
+func (s *destroyStatus) Set(err error, reason DestroyReason) {
+	s.err = err
+	s.reason = reason
+}
+
 // A Replica is a contiguous keyspace with writes managed via an
 // instance of the Raft consensus algorithm. Many ranges may exist
 // in a store and they are unlikely to be contiguous. Ranges are
@@ -288,13 +322,9 @@ type Replica struct {
 	mu struct {
 		// Protects all fields in the mu struct.
 		syncutil.RWMutex
-		// Has the replica been destroyed.
-		destroyed error
-		// Corrupted persistently (across process restarts) indicates whether the
-		// replica has been corrupted.
-		//
-		// TODO(tschottdorf): remove/refactor this field.
-		corrupted bool
+		// The destroyed status of a replica indicating if it's alive, corrupt,
+		// scheduled for destruction or has been GCed.
+		destroyStatus
 		// Is the range quiescent? Quiescent ranges are not Tick()'d and unquiesce
 		// whenever a Raft operation is performed.
 		quiescent bool
@@ -484,7 +514,7 @@ var _ KeyRange = &Replica{}
 func (r *Replica) withRaftGroupLocked(
 	shouldCampaignOnCreation bool, f func(r *raft.RawNode) (unquiesceAndWakeLeader bool, _ error),
 ) error {
-	if r.mu.destroyed != nil {
+	if r.mu.destroyStatus.RemovedOrCorrupt() {
 		// Silently ignore all operations on destroyed replicas. We can't return an
 		// error here as all errors returned from this method are considered fatal.
 		return nil
@@ -688,8 +718,11 @@ func (r *Replica) initRaftMuLockedReplicaMuLocked(
 	if err != nil {
 		return err
 	}
-	r.mu.destroyed = pErr.GetDetail()
-	r.mu.corrupted = r.mu.destroyed != nil
+	if r.mu.destroyStatus.RemovedOrCorrupt() {
+		if err := pErr.GetDetail(); err != nil {
+			r.mu.destroyStatus.Set(err, destroyReasonRemoved)
+		}
+	}
 
 	if replicaID == 0 {
 		repDesc, ok := desc.GetReplicaDescriptor(r.store.StoreID())
@@ -851,6 +884,14 @@ func (r *Replica) setReplicaIDRaftMuLockedMuLocked(replicaID roachpb.ReplicaID) 
 	if r.mu.replicaID > replicaID {
 		return errors.Errorf("replicaID cannot move backwards from %d to %d", r.mu.replicaID, replicaID)
 	}
+
+	if r.mu.destroyStatus.reason == destroyReasonRemovalPending {
+		// An earlier incarnation of this replica was removed, but apparently it has been re-added
+		// now, so reset the status.
+		r.mu.destroyStatus.err = nil
+		r.mu.destroyStatus.reason = destroyReasonAlive
+	}
+
 	// if r.mu.replicaID != 0 {
 	// 	// TODO(bdarnell): clean up previous raftGroup (update peers)
 	// }
@@ -968,7 +1009,7 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 	// replica can get destroyed is an option, alternatively we can clear
 	// our leader status and close the proposalQuota whenever the replica is
 	// destroyed.
-	if r.mu.destroyed != nil {
+	if r.mu.destroyStatus.RemovedOrCorrupt() {
 		if r.mu.proposalQuota != nil {
 			r.mu.proposalQuota.close()
 		}
@@ -1098,8 +1139,9 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 // getEstimatedBehindCountRLocked returns an estimate of how far this replica is
 // behind. A return value of 0 indicates that the replica is up to date.
 func (r *Replica) getEstimatedBehindCountRLocked(raftStatus *raft.Status) int64 {
-	if !r.isInitializedRLocked() || r.mu.replicaID == 0 || r.mu.destroyed != nil {
-		// The range is either not fully initialized or has been destroyed.
+	if !r.isInitializedRLocked() || r.mu.replicaID == 0 || !r.mu.destroyStatus.IsAlive() {
+		// The range is either not fully initialized, has been destroyed, or is slated for removal
+		// (and thus likely ignored by its former Raft peers).
 		return 0
 	}
 	if r.mu.quiescent {
@@ -1151,11 +1193,12 @@ func (r *Replica) IsFirstRange() bool {
 	return r.RangeID == 1
 }
 
-// IsDestroyed returns a non-nil error if the replica has been destroyed.
-func (r *Replica) IsDestroyed() error {
+// IsDestroyed returns a non-nil error if the replica has been destroyed
+// and the reason if it has.
+func (r *Replica) IsDestroyed() (DestroyReason, error) {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	return r.mu.destroyed
+	return r.mu.destroyStatus.reason, r.mu.destroyStatus.err
 }
 
 // GetLease returns the lease and, if available, the proposed next lease.
@@ -1769,7 +1812,6 @@ func (r *Replica) Send(
 	ctx context.Context, ba roachpb.BatchRequest,
 ) (*roachpb.BatchResponse, *roachpb.Error) {
 	var br *roachpb.BatchResponse
-
 	if r.leaseholderStats != nil && ba.Header.GatewayNodeID != 0 {
 		r.leaseholderStats.record(ba.Header.GatewayNodeID)
 	}
@@ -2501,7 +2543,7 @@ func (r *Replica) executeReadOnlyBatch(
 		endCmds.done(br, pErr, proposalNoRetry)
 	}()
 
-	if err := r.IsDestroyed(); err != nil {
+	if _, err := r.IsDestroyed(); err != nil {
 		return nil, roachpb.NewError(err)
 	}
 
@@ -2937,9 +2979,13 @@ func (r *Replica) propose(
 	spans *SpanSet,
 ) (chan proposalResult, func() bool, func(), *roachpb.Error) {
 	noop := func() {}
-	if err := r.IsDestroyed(); err != nil {
-		return nil, nil, noop, roachpb.NewError(err)
+
+	r.mu.Lock()
+	if !r.mu.destroyStatus.IsAlive() {
+		r.mu.Unlock()
+		return nil, nil, noop, roachpb.NewError(r.mu.destroyStatus.err)
 	}
+	r.mu.Unlock()
 
 	rSpan, err := keys.Range(ba)
 	if err != nil {
@@ -3034,12 +3080,12 @@ func (r *Replica) propose(
 		r.mu.Unlock()
 	}
 
-	// NB: We need to check Replica.mu.destroyed again in case the Replica has
+	// NB: We need to check Replica.mu.destroyStatus again in case the Replica has
 	// been destroyed between the initial check at the beginning of this method
 	// and the acquisition of Replica.mu. Failure to do so will leave pending
 	// proposals that never get cleared.
-	if err := r.mu.destroyed; err != nil {
-		return nil, nil, undoQuotaAcquisition, roachpb.NewError(err)
+	if !r.mu.destroyStatus.IsAlive() {
+		return nil, nil, undoQuotaAcquisition, roachpb.NewError(r.mu.destroyStatus.err)
 	}
 
 	repDesc, err := r.getReplicaDescriptorRLocked()
@@ -4693,7 +4739,7 @@ func (r *Replica) acquireSplitLock(
 			// then presumably it was alive for some reason other than a concurrent
 			// split and shouldn't be destroyed.
 			rightRng.mu.Lock()
-			rightRng.mu.destroyed = errors.Errorf("%s: failed to initialize", rightRng)
+			rightRng.mu.destroyStatus.Set(errors.Errorf("%s: failed to initialize", rightRng), destroyReasonRemoved)
 			rightRng.mu.Unlock()
 			r.store.mu.Lock()
 			r.store.mu.replicas.Delete(int64(rightRng.RangeID))
@@ -5547,8 +5593,7 @@ func (r *Replica) maybeSetCorrupt(ctx context.Context, pErr *roachpb.Error) *roa
 
 		log.Errorf(ctx, "stalling replica due to: %s", cErr.ErrorMsg)
 		cErr.Processed = true
-		r.mu.destroyed = cErr
-		r.mu.corrupted = true
+		r.mu.destroyStatus.Set(cErr, destroyReasonCorrupted)
 		pErr = roachpb.NewError(cErr)
 
 		// Try to persist the destroyed error message. If the underlying store is
--- a/pkg/storage/replica_gc_queue.go
+++ b/pkg/storage/replica_gc_queue.go
@@ -92,14 +92,15 @@ func newReplicaGCQueue(store *Store, db *client.DB, gossip *gossip.Gossip) *repl
 	rgcq.baseQueue = newBaseQueue(
 		"replicaGC", rgcq, store, gossip,
 		queueConfig{
-			maxSize:              defaultQueueMaxSize,
-			needsLease:           false,
-			needsSystemConfig:    false,
-			acceptsUnsplitRanges: true,
-			successes:            store.metrics.ReplicaGCQueueSuccesses,
-			failures:             store.metrics.ReplicaGCQueueFailures,
-			pending:              store.metrics.ReplicaGCQueuePending,
-			processingNanos:      store.metrics.ReplicaGCQueueProcessingNanos,
+			maxSize:                  defaultQueueMaxSize,
+			needsLease:               false,
+			needsSystemConfig:        false,
+			acceptsUnsplitRanges:     true,
+			processDestroyedReplicas: true,
+			successes:                store.metrics.ReplicaGCQueueSuccesses,
+			failures:                 store.metrics.ReplicaGCQueueFailures,
+			pending:                  store.metrics.ReplicaGCQueuePending,
+			processingNanos:          store.metrics.ReplicaGCQueueProcessingNanos,
 		},
 	)
 	return rgcq
--- a/pkg/storage/replica_test.go
+++ b/pkg/storage/replica_test.go
@@ -6102,8 +6102,8 @@ func TestReplicaCorruption(t *testing.T) {
 	r := tc.store.LookupReplica(rkey, rkey)
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	if r.mu.destroyed.Error() != pErr.GetDetail().Error() {
-		t.Fatalf("expected r.mu.destroyed == pErr.GetDetail(), instead %q != %q", r.mu.destroyed, pErr.GetDetail())
+	if r.mu.destroyStatus.err.Error() != pErr.GetDetail().Error() {
+		t.Fatalf("expected r.mu.destroyed == pErr.GetDetail(), instead %q != %q", r.mu.destroyStatus, pErr.GetDetail())
 	}
 
 	// Verify destroyed error was persisted.
@@ -6111,8 +6111,8 @@ func TestReplicaCorruption(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if r.mu.destroyed.Error() != pErr.GetDetail().Error() {
-		t.Fatalf("expected r.mu.destroyed == pErr.GetDetail(), instead %q != %q", r.mu.destroyed, pErr.GetDetail())
+	if r.mu.destroyStatus.err.Error() != pErr.GetDetail().Error() {
+		t.Fatalf("expected r.mu.destroyed == pErr.GetDetail(), instead %q != %q", r.mu.destroyStatus, pErr.GetDetail())
 	}
 
 	// TODO(bdarnell): when maybeSetCorrupt is finished verify that future commands fail too.
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -311,10 +311,10 @@ func (rs *storeReplicaVisitor) Visit(visitor func(*Replica) bool) {
 		// can still happen with this code.
 		rs.visited++
 		repl.mu.RLock()
-		destroyed := repl.mu.destroyed
+		destroyed := repl.mu.destroyStatus
 		initialized := repl.isInitializedRLocked()
 		repl.mu.RUnlock()
-		if initialized && destroyed == nil && !visitor(repl) {
+		if initialized && (destroyed.IsAlive() || destroyed.reason == destroyReasonRemovalPending) && !visitor(repl) {
 			break
 		}
 	}
@@ -2155,7 +2155,6 @@ func (s *Store) RemoveReplica(
 			<-s.snapshotApplySem
 		}()
 	}
-
 	rep.raftMu.Lock()
 	defer rep.raftMu.Unlock()
 	return s.removeReplicaImpl(ctx, rep, consistentDesc, destroy)
@@ -2215,7 +2214,7 @@ func (s *Store) removeReplicaImpl(
 	rep.mu.Lock()
 	rep.cancelPendingCommandsLocked()
 	rep.mu.internalRaftGroup = nil
-	rep.mu.destroyed = roachpb.NewRangeNotFoundError(rep.RangeID)
+	rep.mu.destroyStatus.Set(roachpb.NewRangeNotFoundError(rep.RangeID), destroyReasonRemoved)
 	rep.mu.Unlock()
 	rep.readOnlyCmdMu.Unlock()
 
@@ -2239,6 +2238,7 @@ func (s *Store) removeReplicaImpl(
 	// TODO(peter): Could release s.mu.Lock() here.
 	s.maybeGossipOnCapacityChange(ctx, rangeChangeEvent)
 	s.scanner.RemoveReplica(rep)
+
 	return nil
 }
 
@@ -2390,7 +2390,7 @@ func (s *Store) deadReplicas() roachpb.StoreDeadReplicas {
 	s.mu.replicas.Range(func(k int64, v unsafe.Pointer) bool {
 		r := (*Replica)(v)
 		r.mu.RLock()
-		corrupted := r.mu.corrupted
+		corrupted := r.mu.destroyStatus.reason == destroyReasonCorrupted
 		desc := r.mu.state.Desc
 		r.mu.RUnlock()
 		replicaDesc, ok := desc.GetReplicaDescriptor(s.Ident.StoreID)
@@ -2576,7 +2576,6 @@ func (s *Store) Send(
 		if br, pErr = s.maybeWaitInPushTxnQueue(ctx, &ba, repl); br != nil || pErr != nil {
 			return br, pErr
 		}
-
 		br, pErr = repl.Send(ctx, ba)
 		if pErr == nil {
 			return br, nil
@@ -2834,7 +2833,6 @@ func (s *Store) HandleSnapshot(
 			if header.RaftMessageRequest.ToReplica.ReplicaID == 0 {
 				inSnap.snapType = snapTypePreemptive
 			}
-
 			if err := s.processRaftSnapshotRequest(ctx, &header.RaftMessageRequest, inSnap); err != nil {
 				return sendSnapError(errors.Wrap(err.GoError(), "failed to apply snapshot"))
 			}
@@ -3288,6 +3286,16 @@ func (s *Store) HandleRaftResponse(ctx context.Context, resp *RaftMessageRespons
 			if err != nil {
 				log.Errorf(ctx, "unable to add to replica GC queue: %s", err)
 			} else if added {
+				repl.mu.Lock()
+				// The replica will be garbage collected soon (we are sure
+				// since our replicaID is definitely too old), but in the meantime we
+				// already want to bounce all traffic from it. Note that the replica
+				// could be re-added with a higher replicaID, in which this error is
+				// cleared in setReplicaIDRaftMuLockedMuLocked.
+				if repl.mu.destroyStatus.IsAlive() {
+					repl.mu.destroyStatus.Set(roachpb.NewRangeNotFoundError(repl.RangeID), destroyReasonRemovalPending)
+				}
+				repl.mu.Unlock()
 				log.Infof(ctx, "added to replica GC queue (peer suggestion)")
 			}
 		case *roachpb.StoreNotFoundError:
@@ -3914,14 +3922,15 @@ func (s *Store) tryGetOrCreateReplica(
 
 		repl.raftMu.Lock()
 		repl.mu.RLock()
-		destroyed, corrupted := repl.mu.destroyed, repl.mu.corrupted
+		destroyed := repl.mu.destroyStatus
 		repl.mu.RUnlock()
-		if destroyed != nil {
+		if destroyed.reason == destroyReasonRemoved {
 			repl.raftMu.Unlock()
-			if corrupted {
-				return nil, false, destroyed
-			}
 			return nil, false, errRetry
+		}
+		if destroyed.reason == destroyReasonCorrupted {
+			repl.raftMu.Unlock()
+			return nil, false, destroyed.err
 		}
 		repl.mu.Lock()
 		if err := repl.setReplicaIDRaftMuLockedMuLocked(replicaID); err != nil {
@@ -3983,7 +3992,7 @@ func (s *Store) tryGetOrCreateReplica(
 	if err := repl.initRaftMuLockedReplicaMuLocked(desc, s.Clock(), replicaID); err != nil {
 		// Mark the replica as destroyed and remove it from the replicas maps to
 		// ensure nobody tries to use it
-		repl.mu.destroyed = errors.Wrapf(err, "%s: failed to initialize", repl)
+		repl.mu.destroyStatus.Set(errors.Wrapf(err, "%s: failed to initialize", repl), destroyReasonRemoved)
 		repl.mu.Unlock()
 		s.mu.Lock()
 		s.mu.replicas.Delete(int64(rangeID))
--- a/pkg/storage/store_test.go
+++ b/pkg/storage/store_test.go
@@ -471,7 +471,7 @@ func TestStoreRemoveReplicaDestroy(t *testing.T) {
 	}
 
 	repl1.mu.Lock()
-	expErr := roachpb.NewError(repl1.mu.destroyed)
+	expErr := roachpb.NewError(repl1.mu.destroyStatus.err)
 	lease := *repl1.mu.state.Lease
 	repl1.mu.Unlock()
 


nvanbenschoten:
Reviewed 5 of 5 files at r1.

---

*[pkg/storage/queue.go, line 602 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwpruH6xw950cOU4L8o:-KwpruH6xw950cOU4L8p:b-hzj9ha) ([raw file](https://github.com/cockroachdb/cockroach/blob/06aaba95b99a3ec03d90257ee81750dfb5000459/pkg/storage/queue.go#L602)):*
> ```Go
> 	}
> 
> 	if err, pending := repl.IsDestroyed(); err != nil && pending == false {
> ```

nit: `!pending`. Also, in conjunction with the comment on the `pending` flag, can we add one here to discuss why `pending` is taken into consideration?

---

*[pkg/storage/replica.go, line 290 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwpsCKLuD1GxvIR31yN:-KwpsCKLuD1GxvIR31yO:bhycvxi) ([raw file](https://github.com/cockroachdb/cockroach/blob/06aaba95b99a3ec03d90257ee81750dfb5000459/pkg/storage/replica.go#L290)):*
> ```Go
> 		destroyed struct {
> 			destroyedErr error
> 			pending      bool
> ```

This would be easier to understand if it had a comment describing when a `Replica's` destroyed state is considered pending. 

Also, it doesn't look like we ever set this back to `false`. Without being able to visualize the lifetime of this flag, I'm not easily able to convince myself that it's necessary.

---

*[pkg/storage/replica.go, line 1147 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwproRHNQOQkCbENaHF:-KwproRHNQOQkCbENaHG:b-b9o8tn) ([raw file](https://github.com/cockroachdb/cockroach/blob/06aaba95b99a3ec03d90257ee81750dfb5000459/pkg/storage/replica.go#L1147)):*
> ```Go
> }
> 
> // IsDestroyed returns a non-nil error if the replica has been destroyed.
> ```

We should comment about the other return variable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

m-schneider:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/storage/queue.go, line 602 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwpruH6xw950cOU4L8o:-KwqBNt1_I7Szk8PVRvR:bdcv6kv) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/queue.go#L602)):*

nit: `!pending`. Also, in conjunction with the comment on the `pending` flag, can we add one here to discuss why `pending` is taken into consideration?
</blockquote></details>

Good point!

---

*[pkg/storage/replica.go, line 290 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwpsCKLuD1GxvIR31yN:-Kwq9-XAx-rlwSOQdMAc:b-hkgd7q) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L290)):*

This would be easier to understand if it had a comment describing when a `Replica's` destroyed state is considered pending. 

Also, it doesn't look like we ever set this back to `false`. Without being able to visualize the lifetime of this flag, I'm not easily able to convince myself that it's necessary.
</blockquote></details>

The flag is necessary to avoid breaking a couple of things. queue.go would completely ignore a "destroyed" replica for any queue so it would never actually get GCed. Without the specific check on pending in withRaftGroupLocked the replica from the replica descriptor so the RangeLookupRequest in func (rgcq *replicaGCQueue) process would fail because the replica wouldn't be a current member of the range. 

---

*[pkg/storage/replica.go, line 1147 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwproRHNQOQkCbENaHF:-KwqD1aQPq_hBPihrlpd:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L1147)):*

We should comment about the other return variable.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

bdarnell:
Most of the uses of `IsDestroyed` do `err != nil && !pending`, so a replica in the pending state is not counted as destroyed in most cases (where is the place where we break out of the hanging lease request? I'm having trouble finding it). I think it might be better to make this a separate flag, distinct from the existing destroyed state, so it can be checked only in the place(s) where it matters. 

---

Reviewed 5 of 5 files at r2.

---

*[pkg/storage/queue.go, line 602 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqRMo7jt6cBDH661eL:-KwqRMo7jt6cBDH661eM:b-im537) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/queue.go#L602)):*
> ```Go
> 	}
> 
> 	if err, pending := repl.IsDestroyed(); err != nil && !pending {
> ```

Most queues don't want to process destroyed replicas - only the replica GC queue does. I think it might be better to have a special case here (a flag in queueConfig that is only set for replicaGCQueue)  so that we don't send destroyed replicas to the other queues.

---

*[pkg/storage/replica.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kwp58UVFv_8jBiQh26W:-Kwp58UVFv_8jBiQh26X:bb8o5z4) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L289)):*
> ```Go
> 		// Has the replica been destroyed.
> 		destroyed struct {
> 			destroyedErr error
> ```

Call this `err` to avoid the repetition of `destroyed.destroyedErr`. 

---

*[pkg/storage/replica.go, line 688 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kwp6eu1M-Mj91lGfT1j:-Kwp6eu1M-Mj91lGfT1k:bpahkck) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L688)):*
> ```Go
> 		return err
> 	}
> 	r.mu.destroyed.destroyedErr = pErr.GetDetail()
> ```

I think `pending` should probably be set here, to allow the replica GC queue to attempt to clean up these corrupted replicas. This would be an additional behavior change, but it seems like the right thing to do. The only problem is that the whole corrupted-replica mechanism is very under-tested and 

---

*[pkg/storage/replica.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqW-3pFh1gfJ6mF9wZ:-KwqW-3pFh1gfJ6mF9w_:b-gscqh5) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L690)):*
> ```Go
> 		return err
> 	}
> 	if !r.mu.destroyed.pending {
> ```

This is an "init" method; when would `r.mu.destroyed.pending` be true here?

---

*[pkg/storage/replica.go, line 1287 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqWJ3hNoL_NsH7in9Q:-KwqWJ3hNoL_NsH7in9R:b-z95w4u) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L1287)):*
> ```Go
> 			case LeaseState_ERROR:
> 				// Lease state couldn't be determined.
> 				log.Infof(ctx, "lease state couldn't be determined")
> ```

Did you mean to leave this change in?

---

*[pkg/storage/store.go, line 2148 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqWtuwMrs7TMSbJnCL:-KwqWtuwMrs7TMSbJnCM:bd9ntp4) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/store.go#L2148)):*
> ```Go
> 		select {
> 		case s.snapshotApplySem <- struct{}{}:
> 			log.Infof(ctx, "SEMAPHORE AQUIRED")
> ```

Don't forget to remove these messages before merging.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KwqXPpZhhlMRT_zk079:b-8nii2d)*

tschottdorf:
@m-schneider the error (copied for context below) during tests are because in `withRaftGroupLocked`, you take into account the `pending` flag:

```go
	if r.mu.destroyed.destroyedErr != nil && !r.mu.destroyed.pending {
		// Silently ignore all operations on destroyed replicas. We can't return an
		// error here as all errors returned from this method are considered fatal.
		return nil
	}
```

When the replica was *actually* destroyed, the code didn't actually *reset* the pending flag to `false`, and so the above snippet would recreate the in-memory Raft group even though its on-disk state had been removed (something that usually causes an explosion). After fixing that, the same problem occurred, though a little less often, and so I looked for another place that set `pending` to `true`. Of course there's only one, so that place also has to check whether the replica has already been gc'ed (it's convenient to instead avoid doing anything if there's already a `destroyedErr`).

With that change made, `make stress PKG=./pkg/storage/ TESTS=TestReplicateAfterRemoveAndSplit` runs for at least a couple hundred iters (conserving battery, so haven't checked further, but it failed pretty much immediately before).

Diff below:

```diff
diff --git a/pkg/storage/replica.go b/pkg/storage/replica.go
index f3c001cda..5a6677312 100644
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -507,6 +507,7 @@ func (r *Replica) withRaftGroupLocked(
 	ctx := r.AnnotateCtx(context.TODO())
 
 	if r.mu.internalRaftGroup == nil {
+		log.Infof(context.Background(), "NEWRAFT: %+v", r.mu.destroyed)
 		raftGroup, err := raft.NewRawNode(newRaftConfig(
 			raft.Storage((*replicaRaftStorage)(r)),
 			uint64(r.mu.replicaID),
diff --git a/pkg/storage/store.go b/pkg/storage/store.go
index 11b868ef8..77e904479 100644
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -2217,6 +2217,7 @@ func (s *Store) removeReplicaImpl(
 	rep.cancelPendingCommandsLocked()
 	rep.mu.internalRaftGroup = nil
 	rep.mu.destroyed.destroyedErr = roachpb.NewRangeNotFoundError(rep.RangeID)
+	rep.mu.destroyed.pending = false // set this before actually destroying
 	rep.mu.Unlock()
 	rep.readOnlyCmdMu.Unlock()
 
@@ -3291,8 +3292,12 @@ func (s *Store) HandleRaftResponse(ctx context.Context, resp *RaftMessageRespons
 			if err != nil {
 				log.Errorf(ctx, "unable to add to replica GC queue: %s", err)
 			} else if added {
-				repl.mu.destroyed.destroyedErr = roachpb.NewRangeNotFoundError(repl.RangeID)
-				repl.mu.destroyed.pending = true
+				repl.mu.Lock()
+				if repl.mu.destroyed.destroyedErr == nil {
+					repl.mu.destroyed.destroyedErr = roachpb.NewRangeNotFoundError(repl.RangeID)
+					repl.mu.destroyed.pending = true
+				}
+				repl.mu.Unlock()
 				log.Infof(ctx, "added to replica GC queue (peer suggestion)")
 			}
 		case *roachpb.StoreNotFoundError:

```

and here's the crash (without the diff):

```
E171024 11:17:16.490175 535 util/log/crash_reporting.go:119  [s3,r1/3:/M{in-ax}] a panic has occurred!
panic: applied(19) is out of range [prevApplied(10), committed(10)] [recovered]
	panic: applied(19) is out of range [prevApplied(10), committed(10)]

goroutine 535 [running]:
github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).Recover(0xc420c10480, 0x6575d00, 0xc420d60390)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:183 +0x131
panic(0x564e360, 0xc4202de4a0)
	/usr/local/Cellar/go/1.9/libexec/src/runtime/panic.go:491 +0x283
github.com/cockroachdb/cockroach/pkg/storage.(*raftLogger).Panicf(0xc4202de470, 0x5939d8c, 0x3c, 0xc420cfa150, 0x3, 0x3)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/pkg/storage/raft.go:115 +0x1dc
github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft.(*raftLog).appliedTo(0xc420d44d90, 0x13)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft/log.go:202 +0x174
github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft.newRaft(0xc420db75d0, 0x476be4b)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft/raft.go:323 +0xc22
github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft.NewRawNode(0xc420db75d0, 0x0, 0x0, 0x0, 0xc420cc5680, 0xc4209485b0, 0x4014de8)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/vendor/github.com/coreos/etcd/raft/rawnode.go:79 +0x71
github.com/cockroachdb/cockroach/pkg/storage.(*Replica).withRaftGroupLocked(0xc420e26000, 0x0, 0x59aae40, 0xc4202d2d10, 0x100000001)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/pkg/storage/replica.go:510 +0x2e3
github.com/cockroachdb/cockroach/pkg/storage.(*Replica).maybeInitializeRaftGroup(0xc420e26000, 0x6575d00, 0xc420afa5d0)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/pkg/storage/replica.go:1760 +0x185
github.com/cockroachdb/cockroach/pkg/storage.(*Replica).Send(0xc420e26000, 0x6575d00, 0xc420afa5d0, 0x6b49d27d, 0xf2, 0x0, 0x0, 0x1, 0x0, 0x0, ...)
	/Users/tschottdorf/go/src/github.com/cockroachdb/cockroach/pkg/storage/replica.go:1784 +0x136
```

---

---

*[pkg/storage/replica.go, line 688 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kwp6eu1M-Mj91lGfT1j:-KxD4Tctmaow7fS1WSj6:b-4qgcwi) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L688)):*

I think `pending` should probably be set here, to allow the replica GC queue to attempt to clean up these corrupted replicas. This would be an additional behavior change, but it seems like the right thing to do. The only problem is that the whole corrupted-replica mechanism is very under-tested and 
</blockquote></details>

... and basically garbage. I think the only reason for keeping it at the moment is that it manages to persist the corruption so that you can't just restart the node and keep going. We're definitely not in a place where this corruption is handled in any way other than exploding. (Even worse, we sometimes keep running and it doesn't end well).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KxD5nG3-f8xMtKJ81EP:b-r34p0o)*

tschottdorf:
re: @bdarnell's comments, I tend to agree that it's a bit awkward. How's the following:

```
type destroyReason int

type destroyStatus struct {
    reason destroyReason
    err error
}

func (s destroyStatus) IsAlive() bool {
  return s.reason == destroyReasonAlive
}

const (
  destroyReasonAlive destroyReason = iota // `err` must be nil (Go has no enums :crying_cat_face: )
  destroyReasonCorrupted                  // supersedes the `corrupted` bool
  destroyReasonRemovalPending             // replaces `pending==true`
  destroyReasonRemoved                    // replaces `pending==false`
)

type replicaMu struct {
// ...
  destroyed destroyStatus
}
```

Besides, this is mostly a WIP to see if it actually translates into better chaos behavior, so if we're reasonably certain that it does what it should, I'd rather test drive it first and then get into the bike shedding.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KxD8t_q7hHgH4LnAJ1d:b-qqb1g4)*

tschottdorf:

---

*[pkg/storage/queue.go, line 602 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqRMo7jt6cBDH661eL:-KxDBiP-MMZ02H6bxk8f:bhr0w2p) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/queue.go#L602)):*

Most queues don't want to process destroyed replicas - only the replica GC queue does. I think it might be better to have a special case here (a flag in queueConfig that is only set for replicaGCQueue)  so that we don't send destroyed replicas to the other queues.
</blockquote></details>

This means adding a field `processNonAlive` (name :bike: :house_with_garden: ) to `queueConfig` and using the field here. (All queues except the replica gc queue set it true, of course). Note that this would also mean that previously GC'ed replicas may be GC'ed again. It'd be good to avoid this from actually happening (i.e. the replicaGCQueue's process method would check the status and only actually GC if it's alive or marked for deletion but not previously GC'ed).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

m-schneider:
That looks much cleaner, I'll refactor and rerun the tests.

---

Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/storage/queue.go, line 602 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqRMo7jt6cBDH661eL:-KxJvhvAiVg5hx9oL9zg:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/queue.go#L602)):*

This means adding a field `processNonAlive` (name :bike: :house_with_garden: ) to `queueConfig` and using the field here. (All queues except the replica gc queue set it true, of course). Note that this would also mean that previously GC'ed replicas may be GC'ed again. It'd be good to avoid this from actually happening (i.e. the replicaGCQueue's process method would check the status and only actually GC if it's alive or marked for deletion but not previously GC'ed).
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 289 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kwp58UVFv_8jBiQh26W:-KxJvjiib2qSvvs7PeQ5:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L289)):*

Call this `err` to avoid the repetition of `destroyed.destroyedErr`. 
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 688 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kwp6eu1M-Mj91lGfT1j:-KxJw0aahlv6-7dVSpYP:b-c1a2eh) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1f626a81b006dee0fa2e158b19cc8ea3f724049/pkg/storage/replica.go#L688)):*

... and basically garbage. I think the only reason for keeping it at the moment is that it manages to persist the corruption so that you can't just restart the node and keep going. We're definitely not in a place where this corruption is handled in any way other than exploding. (Even worse, we sometimes keep running and it doesn't end well).
</blockquote></details>

I'm going to run some tests on a real cluster first, but then I'll set pending to true  and see what happens!

---

*[pkg/storage/replica.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqW-3pFh1gfJ6mF9wZ:-KxJvsUT25eMK0kd3evo:bwqrvcp) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L690)):*

This is an "init" method; when would `r.mu.destroyed.pending` be true here?
</blockquote></details>

Got rid of it!

---

*[pkg/storage/replica.go, line 1287 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqWJ3hNoL_NsH7in9Q:-KxJwaqnRZAUrqjV6FUy:b-87j1r7) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L1287)):*

Did you mean to leave this change in?
</blockquote></details>

Gone!

---

*[pkg/storage/store.go, line 2148 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqWtuwMrs7TMSbJnCL:-KxIyMBSktDOSKjOKZF1:b-k7mx83) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/store.go#L2148)):*

Don't forget to remove these messages before merging.
</blockquote></details>

Deleted!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KxIxKaiPMaArZPy9GgO:bsfxu5c)*

m-schneider:
After migrating the experiment to an ephemeral cluster, this change seems to fix the drop we've been seeing. Here's a graph of a node in an ephemeral cluster before the change was applied.

screenshot-2017-10-30 cockroach console 1

You can see the dips when the restarting node rejoins the cluster.

Here's the cluster with this PR:

screenshot-2017-10-31 cockroach console

As you can see there are no dips when the restarted node comes back online at the 10, 30 and 50 minute marks.

nvanbenschoten:
@m-schneider That improvement is great, nice job!

---

Reviewed 1 of 6 files at r3, 5 of 5 files at r4.

---

*[pkg/storage/queue.go, line 195 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxnwv9WbYZ67fAPWTsL:-Kxnwv9WbYZ67fAPWTsM:b-gjcqf) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/queue.go#L195)):*
> ```Go
> 	// purgatory is a gauge measuring current replica count in purgatory.
> 	purgatory *metric.Gauge
> 	// whether or not we want to process replicas that have been destroyed but not GCed.
> ```

nit: let's move up next to the other bool options (needsSystemConfig, acceptsUnsplitRanges, ...) and begin the comment with the name of the variable.

---

*[pkg/storage/replica.go, line 219 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxo-HdO58c-chU0guYR:-Kxo-HdO58c-chU0guYS:b-eeov36) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L219)):*
> ```Go
> }
> 
> // The reason that a replica has been marked as destroyed.
> ```

I think the linter's going to yell about this comment.

---

*[pkg/storage/replica.go, line 235 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxnxgrq3ARd7Vsq4bi_:-Kxnxgrq3ARd7Vsq4bia:b-rfwfas) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L235)):*
> ```Go
> 	// Corrupted persistently (across process restarts) indicates whether the
> 	// replica has been corrupted.
> 	destroyReasonCorrupted
> ```

We can actually fail to persist the corruption error but we'll still be in this state (see `maybeSetCorrupt`). Because of that, this comment is a little misleading.

---

*[pkg/storage/replica.go, line 237 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnxSe7psHAW8DPqYoq:-KxnxSe7psHAW8DPqYor:b-93z8w2) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L237)):*
> ```Go
> 	destroyReasonCorrupted
> 	destroyReasonRemovalPending // replaces `pending==true`
> 	destroyReasonRemoved        // replaces `pending==false`
> ```

These comments won't mean much after this PR. Let's add explanations to each variant.

---

*[pkg/storage/replica.go, line 312 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxnz7zpTwc3bqVfH4-5:-Kxnz7zpTwc3bqVfH4-6:b-v5rupx) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L312)):*
> ```Go
> 		// Protects all fields in the mu struct.
> 		syncutil.RWMutex
> 		// Has the replica been destroyed. Pending will be set when a replica is
> ```

Stale comment.

---

*[pkg/storage/replica.go, line 709 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyUV7Q_Qzvw3VySIP:-KxnyUV7Q_Qzvw3VySIQ:b-4pro9m) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L709)):*
> ```Go
> 	}
> 	if r.mu.destroyed.reason != destroyReasonRemovalPending {
> 		r.mu.destroyed.err = pErr.GetDetail()
> ```

Is it possible for `pErr != nil && pErr.GetDetail() == nil`? If not, we should check if `pErr` is non-null and if not, set `r.mu.destroyed.err` and `r.mu.destroyed.reason` together. That would make this easier to understand.

---

*[pkg/storage/replica.go, line 1176 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyLqm62ED0OI2PrCD:-KxnyLqm62ED0OI2PrCE:bn40rft) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L1176)):*
> ```Go
> 
> // IsDestroyed returns a non-nil error if the replica has been destroyed
> // and the value of pending. If pending is true the replica has been added
> ```

Stale comment about `pending`

---

*[pkg/storage/replica.go, line 2964 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyC-C6bSSo8Rgh0wW:-KxnyC-C6bSSo8Rgh0wX:b-5pzsy1) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L2964)):*
> ```Go
> 
> 	r.mu.Lock()
> 	if err := r.mu.destroyed.err; err != nil && !r.mu.destroyed.IsAlive() {
> ```

Same question as below. Seems like the error checks are no longer necessary. 

---

*[pkg/storage/replica.go, line 3067 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxny-5os8IVV_rBY3jX:-Kxny-5os8IVV_rBY3jY:b-l4u51p) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L3067)):*
> ```Go
> 	// and the acquisition of Replica.mu. Failure to do so will leave pending
> 	// proposals that never get cleared.
> 	if err := r.mu.destroyed.err; err != nil && r.mu.destroyed.reason == destroyReasonRemoved {
> ```

Do we need to check if `err` is null? Will `reason == destroyReasonRemoved` ever be true when err is not set?

---

*[pkg/storage/replica_gc_queue.go, line 103 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzFe6BxUR5yXgjS4L:-KxnzFe6BxUR5yXgjS4M:bigj1de) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica_gc_queue.go#L103)):*
> ```Go
> 			pending:                  store.metrics.ReplicaGCQueuePending,
> 			processingNanos:          store.metrics.ReplicaGCQueueProcessingNanos,
> 			processDestroyedReplicas: true,
> ```

nit: we should move this up next to the other bool options.

---

*[pkg/storage/store.go, line 314 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzWP8cI02f6Tvr-Zn:-KxnzWP8cI02f6Tvr-Zo:b-sqp5ia) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L314)):*
> ```Go
> 		rs.visited++
> 		repl.mu.RLock()
> 		destroyed := repl.mu.destroyed
> ```

Consider renaming this now that its type has changed.

---

*[pkg/storage/store.go, line 317 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzOwz5gk-b-fLRWZE:-KxnzOwz5gk-b-fLRWZF:b-j18db7) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L317)):*
> destroyed.IsAlive() || destroyed.reason == destroyReasonRemovalPending

Does this pair of states warrant a name and a corresponding method? `IsNotRemoved` or something?

---

*[pkg/storage/store.go, line 3293 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzlAsRbiKoNuAdFDo:-KxnzlAsRbiKoNuAdFDp:bgybjmq) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L3293)):*
> ```Go
> 				if repl.mu.destroyed.IsAlive() {
> 					repl.mu.destroyed.err = roachpb.NewRangeNotFoundError(repl.RangeID)
> 					repl.mu.destroyed.reason = destroyReasonRemovalPending
> ```

Since we're setting these two together everywhere now, it probably makes sense to encapsulate setting them into a method so that we don't miss setting one or the other somewhere. What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-Kxo-_TmBynD8SiO7-tB:b-u7f9vn)*

m-schneider:
Review status: 2 of 6 files reviewed at latest revision, 19 unresolved discussions, all commit checks successful.

---

*[pkg/storage/queue.go, line 195 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxnwv9WbYZ67fAPWTsL:-Kxo7mE7qCBUcW9TEeLQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/queue.go#L195)):*

nit: let's move up next to the other bool options (needsSystemConfig, acceptsUnsplitRanges, ...) and begin the comment with the name of the variable.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 219 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxo-HdO58c-chU0guYR:-Kxscx6c1baWOOGv46y7:bt2vr0y) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L219)):*

I think the linter's going to yell about this comment.
</blockquote></details>

Fixed!

---

*[pkg/storage/replica.go, line 235 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxnxgrq3ARd7Vsq4bi_:-KxoA-SWNedv8efI-gTy:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L235)):*

We can actually fail to persist the corruption error but we'll still be in this state (see `maybeSetCorrupt`). Because of that, this comment is a little misleading.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 237 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnxSe7psHAW8DPqYoq:-Kxo8z0-WQ--e43M5bkU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L237)):*

These comments won't mean much after this PR. Let's add explanations to each variant.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 312 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxnz7zpTwc3bqVfH4-5:-Kxo8zkK4JXwhewJieHx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L312)):*

Stale comment.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 709 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyUV7Q_Qzvw3VySIP:-KxsHD7ZcwPh3QQVOANS:bg1u2xw) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L709)):*

Is it possible for `pErr != nil && pErr.GetDetail() == nil`? If not, we should check if `pErr` is non-null and if not, set `r.mu.destroyed.err` and `r.mu.destroyed.reason` together. That would make this easier to understand.
</blockquote></details>

Added a function to set the values together.

---

*[pkg/storage/replica.go, line 1176 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyLqm62ED0OI2PrCD:-Kxo95XOfjhzWKLmwtSk:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L1176)):*

Stale comment about `pending`
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 2964 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyC-C6bSSo8Rgh0wW:-KxsczutHcuapWGcxk45:b6z6k5) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L2964)):*

Same question as below. Seems like the error checks are no longer necessary. 
</blockquote></details>

Got rid of them.

---

*[pkg/storage/replica.go, line 3067 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-Kxny-5os8IVV_rBY3jX:-Kxsd21skvftQOVvg_Rv:bfsvh3e) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L3067)):*

Do we need to check if `err` is null? Will `reason == destroyReasonRemoved` ever be true when err is not set?
</blockquote></details>

Got rid of the error check.

---

*[pkg/storage/replica_gc_queue.go, line 103 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzFe6BxUR5yXgjS4L:-Kxo9GOanDtKCKLBme7S:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica_gc_queue.go#L103)):*

nit: we should move this up next to the other bool options.
</blockquote></details>

Done.

---

*[pkg/storage/store.go, line 314 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzWP8cI02f6Tvr-Zn:-KxsdGmaj6nN3OT-jxzg:b-qjwzdz) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L314)):*

Consider renaming this now that its type has changed.
</blockquote></details>

What do you think is a good name?

---

*[pkg/storage/store.go, line 317 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzOwz5gk-b-fLRWZE:-Kxsd7Wht9tJIoMnmrFU:b-rf9zst) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L317)):*

> destroyed.IsAlive() || destroyed.reason == destroyReasonRemovalPending

Does this pair of states warrant a name and a corresponding method? `IsNotRemoved` or something?
</blockquote></details>

It's only used once, so I don't know if that's enough for adding a function.

---

*[pkg/storage/store.go, line 3293 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzlAsRbiKoNuAdFDo:-KxsG7uLWP1YLprTSV8r:b-4wzgh4) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L3293)):*

Since we're setting these two together everywhere now, it probably makes sense to encapsulate setting them into a method so that we don't miss setting one or the other somewhere. What do you think?
</blockquote></details>

Sounds good, I added a function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

nvanbenschoten:
This is looking a lot better. Just nits at this point.

---

Reviewed 4 of 4 files at r5.

---

*[pkg/storage/replica.go, line 1176 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyLqm62ED0OI2PrCD:-KyH0MZE-QkNVpFCJFYC:bn1gqqc) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L1176)):*

Done.
</blockquote></details>

Hmm, I don't see a change.

---

*[pkg/storage/replica.go, line 231 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH0jSxTmYJ0yxrvB_G:-KyH0jSxTmYJ0yxrvB_H:b-sdhw4x) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L231)):*
> ```Go
> }
> 
> func (s *destroyStatus) SetDestroyed(err error, reason DestroyReason) {
> ```

nit: this stutters. Consider calling this simply `Set`

---

*[pkg/storage/replica.go, line 239 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH-Sd-hH154Pb1x1SF:-KyH-Sd-hH154Pb1x1SG:b-3375u5) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L239)):*
> ```Go
> 	// The Replica is alive.
> 	destroyReasonAlive DestroyReason = iota
> 	// Indicates whether the replica has been corrupted.
> ```

nit: `The replica has been corrupted` to remain constant with the other comments.

---

*[pkg/storage/replica.go, line 716 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH0-Yu5GZSp-ngQ8uT:-KyH0-Yu5GZSp-ngQ8uU:b-v3ejhl) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L716)):*
> ```Go
> 	}
> 	if r.mu.destroyed.reason != destroyReasonRemovalPending {
> 		if pErr.GetDetail() != nil {
> ```

nit: the following cuts down on a bit of repetition
```
if err := pErr.GetDetail(); err != nil {
    r.mu.destroyed.SetDestroyed(err, destroyReasonRemoved)
}
```

---

*[pkg/storage/replica.go, line 3073 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH0_8X-r07CYVEr9op:-KyH0_8X-r07CYVEr9oq:b-maf2ty) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L3073)):*
> ```Go
> 	// and the acquisition of Replica.mu. Failure to do so will leave pending
> 	// proposals that never get cleared.
> 	if r.mu.destroyed.reason == destroyReasonRemoved {
> ```

Does this need to be performed under lock?

---

*[pkg/storage/store.go, line 314 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzWP8cI02f6Tvr-Zn:-KyH0vhUQWk9z5NSJXvO:beyw1th) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L314)):*

What do you think is a good name?
</blockquote></details>

Should we be calling this `destroyedStatus` everywhere?

---

*[pkg/storage/store.go, line 317 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzOwz5gk-b-fLRWZE:-KyH0tobN1Rsew4SB_hD:bb73djm) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L317)):*

It's only used once, so I don't know if that's enough for adding a function.
</blockquote></details>

SGTM

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KyH16E9vSwjOWepfEFZ:b5llctn)*

bdarnell:
Reviewed 1 of 6 files at r3, 1 of 5 files at r4, 4 of 4 files at r5.

---

*[pkg/storage/replica.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqW-3pFh1gfJ6mF9wZ:-KyISysiLbw1st1Jr8Kr:b-rbbhau) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L690)):*

Got rid of it!
</blockquote></details>

Looks like it's back in a slightly different form? 

---

*[pkg/storage/replica.go, line 236 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyIRc_3XP1G2HgPfnLL:-KyIRc_3XP1G2HgPfnLM:b85nvbv) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L236)):*
> ```Go
> }
> 
> const (
> ```

Nit: I'd put these constants immediately after the `type DestroyReason int` line.

---

*[pkg/storage/replica.go, line 511 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyIS2dFv8es6l2gk798:-KyIS2dFv8es6l2gk799:bfgxwil) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L511)):*
> ```Go
> 	shouldCampaignOnCreation bool, f func(r *raft.RawNode) (unquiesceAndWakeLeader bool, _ error),
> ) error {
> 	if r.mu.destroyed.reason == destroyReasonRemoved {
> ```

This looks like a change in behavior - previously, corrupted replicas had a non-nil `destroyed` too. I think anything that was `destroyed != nil` should now be `destroyed.IsAlive()`. Any time we're not using `IsAlive` probably deserves a comment. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

m-schneider:
Review status: 2 of 6 files reviewed at latest revision, 12 unresolved discussions.

---

*[pkg/storage/replica.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqW-3pFh1gfJ6mF9wZ:-KyNH8I38yXJw7ymJ_CW:bjugx3r) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L690)):*

Looks like it's back in a slightly different form? 
</blockquote></details>

Does that work better?

---

*[pkg/storage/replica.go, line 1176 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnyLqm62ED0OI2PrCD:-KyMsPDWZukVwn5UFZ8G:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/replica.go#L1176)):*

Hmm, I don't see a change.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 231 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH0jSxTmYJ0yxrvB_G:-KyMsW9v5BAcSyYeSYMp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L231)):*

nit: this stutters. Consider calling this simply `Set`
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 236 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyIRc_3XP1G2HgPfnLL:-KyMsXQkYGjdxfi-Pq12:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L236)):*

Nit: I'd put these constants immediately after the `type DestroyReason int` line.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 239 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH-Sd-hH154Pb1x1SF:-KyMsX3lJ4GO1dtGBzO2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L239)):*

nit: `The replica has been corrupted` to remain constant with the other comments.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 511 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyIS2dFv8es6l2gk798:-KyNGUzoMRO-9jhp2nV8:bb3lih9) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L511)):*

This looks like a change in behavior - previously, corrupted replicas had a non-nil `destroyed` too. I think anything that was `destroyed != nil` should now be `destroyed.IsAlive()`. Any time we're not using `IsAlive` probably deserves a comment. 
</blockquote></details>

I added a named function, but IsAlive isn't preserving old behavior because destroyed can get set earlier now. Leaving it as IsAlive breaks some tests like Example_rebalancing.

---

*[pkg/storage/replica.go, line 716 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH0-Yu5GZSp-ngQ8uT:-KyNHMVVCjcC8EZIvJSC:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L716)):*

nit: the following cuts down on a bit of repetition
```
if err := pErr.GetDetail(); err != nil {
    r.mu.destroyed.SetDestroyed(err, destroyReasonRemoved)
}
```
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 3073 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyH0_8X-r07CYVEr9op:-KyNHXbpaX5dK89JGH8e:b-vfft99) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L3073)):*

Does this need to be performed under lock?
</blockquote></details>

It's locked above.

---

*[pkg/storage/store.go, line 314 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KxnzWP8cI02f6Tvr-Zn:-KyRRJQwoFwzzaigUVPU:bv77htt) ([raw file](https://github.com/cockroachdb/cockroach/blob/89a63a12abe53498f7e61c4e012b15ab848d8e5b/pkg/storage/store.go#L314)):*
> destroyedStatus
Makes sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

bdarnell:
:lgtm:

---

Reviewed 4 of 4 files at r6.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[pkg/storage/replica.go, line 690 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KwqW-3pFh1gfJ6mF9wZ:-KyRfQkVtm0QBlnuzyDj:b-ao5fzs) ([raw file](https://github.com/cockroachdb/cockroach/blob/786e399d1939b608926a6ef091eadf80e6b6ca6d/pkg/storage/replica.go#L690)):*

Does that work better?
</blockquote></details>

Yep, makes sense now. 

---

*[pkg/storage/replica.go, line 511 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyIS2dFv8es6l2gk798:-KyRf2eBxYkbunAHFZzH:b-e04k6h) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L511)):*

I added a named function, but IsAlive isn't preserving old behavior because destroyed can get set earlier now. Leaving it as IsAlive breaks some tests like Example_rebalancing.
</blockquote></details>

Hmm, that's kind of surprising. I'd have thought that even with destroyed being set earlier, it wasn't happening until after we no longer need the raft group. Anyway, this is fine if you uncomment it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KyRfZ47eQ5fuese8PR_:bnfp4nl)*

nvanbenschoten:
:lgtm:

---

---

*[pkg/storage/replica.go, line 223 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyRkEJVjfJF-h2LVUy-:-KyRkEJVjfJF-h2LVUy0:bkfihgf) ([raw file](https://github.com/cockroachdb/cockroach/blob/d92e360c676d487a8cc9f272029f4759bcd71db4/pkg/storage/replica.go#L223)):*
> ```Go
> 
> const (
> 	// The Replica is alive.
> ```

nit: Replica is capital here, lowercase elsewhere.

---

*[pkg/storage/replica.go, line 517 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyRkKkhC8ig27qEDiMd:-KyRkKki6cF5xPM5hbSf:bljwhnl) ([raw file](https://github.com/cockroachdb/cockroach/blob/d92e360c676d487a8cc9f272029f4759bcd71db4/pkg/storage/replica.go#L517)):*
> ```Go
> 	shouldCampaignOnCreation bool, f func(r *raft.RawNode) (unquiesceAndWakeLeader bool, _ error),
> ) error {
> 	//if r.mu.destroyed.RemovedOrCorrupt() {
> ```

Did you mean to comment this out?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KyRkTHNOhxfYfXdfz7Q:bnfp4nl)*

tschottdorf:
I never see `reason` be reset. Maybe I'm missing something, but it's my understanding that a rapid remove-readd should first mark the replica as will-deleted but then upgrades it with a higher replicaID. Wouldn't the replica be in an undesired state at that point or is something resetting `reason` that I've missed?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KyRtg3jNFMD3R2V6ekD:b-ssnb7r)*

m-schneider:
Where was it reset before? I looked through a version of the code without my change and couldn't find where the error was set to null.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KySSL8R7nkQIrGgwhJu:bipvnzg)*

tschottdorf:
It wasn't reset before, but I also don't think it was ever set before when it wasn't permanently broken (corrupted) or definitely being GC'ed. You're setting the error earlier and, I think, making it necessary to reset the error when the replicaID changes (at least if the error is of the right type, you wouldn't want to wipe a corruption error).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KyST5tFYV1qn8wl-s6N:b-7f2n16)*

m-schneider:
Got it, I actually check for slightly different statuses of the destroyed reason in different functions so things like proposals get use a weaker form of liveness to short circuit while other functions make sure that it's either been GCed or is corrupt.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KySUXtGPLyi1aMqbHby:b-pq9ccb)*

m-schneider:
Review status: 5 of 6 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/storage/replica.go, line 511 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyIS2dFv8es6l2gk798:-KySk-ubszj0SeESkrU6:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/91b86db2c5ff5951c6b5e252ee7a89151c57ffed/pkg/storage/replica.go#L511)):*

Hmm, that's kind of surprising. I'd have thought that even with destroyed being set earlier, it wasn't happening until after we no longer need the raft group. Anyway, this is fine if you uncomment it. 
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 223 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyRkEJVjfJF-h2LVUy-:-KySjygTksicz8p9jogl:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d92e360c676d487a8cc9f272029f4759bcd71db4/pkg/storage/replica.go#L223)):*

nit: Replica is capital here, lowercase elsewhere.
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 517 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyRkKkhC8ig27qEDiMd:-KySk0DPy403p3uCFm12:baiyqox) ([raw file](https://github.com/cockroachdb/cockroach/blob/d92e360c676d487a8cc9f272029f4759bcd71db4/pkg/storage/replica.go#L517)):*

Did you mean to comment this out?
</blockquote></details>

Uncommented!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

tschottdorf:
:lgtm: with some comments (including the replica reset). I think you can safely remove the `[WIP]` from the PR title and commit messages :-)

---

Reviewed 2 of 4 files at r5, 3 of 4 files at r6, 1 of 1 files at r7.

---

*[pkg/storage/replica.go, line 1004 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWrl6tBZT_0cWNBgTs:-KyWrl6tBZT_0cWNBgTt:b-nd2yqa) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L1004)):*
> ```Go
> 	// our leader status and close the proposalQuota whenever the replica is
> 	// destroyed.
> 	if r.mu.destroyStatus.reason == destroyReasonRemoved {
> ```

Should this be `RemovedOrCorrupt()` to preserve the old behavior?

---

*[pkg/storage/replica.go, line 1135 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWs4vZCVZpQFce2afa:-KyWs4vZCVZpQFce2afb:btd50zg) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L1135)):*
Comment is slightly out of date. How about

> // The range is either not fully initialized, has been destroyed, or is slated for removal (and thus likely ignored by its former Raft peers).

---

*[pkg/storage/replica.go, line 3074 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWsmp0-5dmQN5k4mjN:-KyWsmp0-5dmQN5k4mjO:bv35de7) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L3074)):*
> ```Go
> 	}
> 
> 	// NB: We need to check Replica.mu.destroyed again in case the Replica has
> ```

`destroyStatus`

---

*[pkg/storage/replica.go, line 3078 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWsrurQFP5WbSm9uwW:-KyWsrurQFP5WbSm9uwX:b-kujm6u) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L3078)):*
> ```Go
> 	// and the acquisition of Replica.mu. Failure to do so will leave pending
> 	// proposals that never get cleared.
> 	if r.mu.destroyStatus.reason == destroyReasonRemoved {
> ```

this should mirror the initial check with `!IsAlive`.

---

*[pkg/storage/store.go, line 3290 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWtGKtTt4afowjqwTB:-KyWtGKtTt4afowjqwTC:bxcpy7l) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/store.go#L3290)):*
This could use a comment:

> // The replica will be garbage collected soon (we are sure since our replicaID is definitely too old), but in the meantime we already want to bounce all traffic from it. Note that the replica could be re-added with a higher replicaID, in which this error is cleared in `setReplicaIDRaftMuLockedMuLocked`.

This is also not done yet, but I think it needs to happen. In the method above, in the case in which the replicaID increases, we need code like

```go
if repl.mu.destroyStatus.reason == destroyReasonRemovalPending {
  // An earlier incarnation of this replica was removed, but apparently it has been re-added
  // now, so reset the status.
  repl.mu.destroyStatus.err = nil
  repl.mu.destroyStatus.reason = destroyReasonAlive
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-:-KyWumJmkoJC8gQvslOg:bcifuj5)*

tschottdorf:
Reviewed 2 of 2 files at r8.

---

*[pkg/storage/store.go, line 3293 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyXJsqpRC6g2KFy2QIE:-KyXJsqq-o8uYH9BzfGc:b2hdfcr) ([raw file](https://github.com/cockroachdb/cockroach/blob/c07f3b059377bebc37fbcfb89d273c650d74a63e/pkg/storage/store.go#L3293)):*
> ```Go
> 				// since our replicaID is definitely too old), but in the meantime we
> 				// already want to bounce all traffic from it. Note that the replica
> 				// could be re-added with a higher replicaID, in which this error is
> ```

aw, I missed a word here (`in which case`), but it's not worth re-running CI.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

tschottdorf:
LG! Let's 🛍 this 🏓

m-schneider:

---

*[pkg/storage/replica.go, line 1004 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWrl6tBZT_0cWNBgTs:-KyXEloIdNupgznlY95f:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L1004)):*

Should this be `RemovedOrCorrupt()` to preserve the old behavior?
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 1135 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWs4vZCVZpQFce2afa:-KyXEnAh_TxY2wBPQitv:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L1135)):*

Comment is slightly out of date. How about

> // The range is either not fully initialized, has been destroyed, or is slated for removal (and thus likely ignored by its former Raft peers).
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 3074 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWsmp0-5dmQN5k4mjN:-KyXEsoiNo4KG81kkHUQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L3074)):*

`destroyStatus`
</blockquote></details>

Done.

---

*[pkg/storage/replica.go, line 3078 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWsrurQFP5WbSm9uwW:-KyXEqKkNH4rdcsPvJfg:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/replica.go#L3078)):*

this should mirror the initial check with `!IsAlive`.
</blockquote></details>

Done.

---

*[pkg/storage/store.go, line 3290 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353#-KyWtGKtTt4afowjqwTB:-KyXEqyv1kwJ_t6bM5TV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/72fc130a5fab2eb17fda676d29f00833d786396b/pkg/storage/store.go#L3290)):*

This could use a comment:

> // The replica will be garbage collected soon (we are sure since our replicaID is definitely too old), but in the meantime we already want to bounce all traffic from it. Note that the replica could be re-added with a higher replicaID, in which this error is cleared in `setReplicaIDRaftMuLockedMuLocked`.

This is also not done yet, but I think it needs to happen. In the method above, in the case in which the replicaID increases, we need code like

```go
if repl.mu.destroyStatus.reason == destroyReasonRemovalPending {
  // An earlier incarnation of this replica was removed, but apparently it has been re-added
  // now, so reset the status.
  repl.mu.destroyStatus.err = nil
  repl.mu.destroyStatus.reason = destroyReasonAlive
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19353)*

PR #19358, opened 2017-10-19
ui: use targeted polyfill instead of babel-polyfill

benesch:
Fixes the build failure in #19063. Opening a PR to verify that TeamCity is happy. @mrtracy feel free to do with this what you wish!


benesch:
This was squashed into #19063.

Issue #19345, opened 2017-10-18
storage: reconsider command queue RangeGroup implementation


asubiotto:
Using a load generator with a contentious workload, I saw a fair amount of time spent in `CommandQueue.getPrereqs`:
<img width="781" alt="screen shot 2017-10-17 at 3 50 04 pm" src="https://user-images.githubusercontent.com/10560359/31728330-0569fb76-b3fa-11e7-8504-c200e13383d4.png">

Switching the underlying implementation of the range groups from a `RangeTree` to a `RangeList` saw an improvement in the percentage of time spent in `getPrereqs`:
<img width="456" alt="screen shot 2017-10-17 at 5 09 17 pm" src="https://user-images.githubusercontent.com/10560359/31728465-6681c0ba-b3fa-11e7-9c91-4544578098e8.png">

The `CommandQueue` microbenchmarks show a marked improvement using the `RangeList` instead of the `RangeTree`:
```
benchstat rangetree rangelist
name                                          old time/op  new time/op  delta
CommandQueueGetPrereqsAllReadOnly/size=1-8     112ns ± 2%   109ns ± 1%   -3.25%  (p=0.000 n=20+20)
CommandQueueGetPrereqsAllReadOnly/size=4-8     112ns ± 3%   109ns ± 2%   -2.91%  (p=0.000 n=17+19)
CommandQueueGetPrereqsAllReadOnly/size=16-8    129ns ±43%   108ns ± 1%  -16.60%  (p=0.001 n=17+15)
CommandQueueGetPrereqsAllReadOnly/size=64-8    107ns ± 2%   106ns ± 3%   -1.31%  (p=0.007 n=20+20)
CommandQueueGetPrereqsAllReadOnly/size=128-8   109ns ± 2%   109ns ± 3%     ~     (p=0.960 n=19+20)
CommandQueueGetPrereqsAllReadOnly/size=256-8   111ns ± 1%   105ns ± 3%   -5.98%  (p=0.000 n=20+20)
CommandQueueReadWriteMix/readsPerWrite=1-8    9.06ms ± 2%  6.60ms ± 6%  -27.18%  (p=0.000 n=20+20)
CommandQueueReadWriteMix/readsPerWrite=4-8    5.15ms ± 1%  4.29ms ± 4%  -16.60%  (p=0.000 n=20+19)
[...]

PR #19361, opened 2017-10-19
util/interval: bucket multiple Ranges in single rangeList node

nvb:
Related to #19345.

Previously, the linked list implementation version of `RangeGroup` would
store a single `Range` in each list `Element`. This was pretty heavy on
allocations because it required a new list `Element` every time we
inserted a `Range` that couldn't be merged with existing ranges. Is also
required us to compare against every `Range` in the group when searching,
which was a big deal because searching for an overlapping `Range` is the
first step of almost all `RangeGroup` operations.

The implementation now buckets a constant number of `Ranges` in each
linked list node. This has two benefits. First, it reduces allocations
by a factor of `rangeListNodeBucketSize` because each list node can now
store up to `rangeListNodeBucketSize` ranges. Second, it allows us to
scan through the `rangeList` faster when searching for a certain range
by comparing only against the largest `Range` in each node as we scan
forward, until we find one that is larger than our desired `Range`. This
technique is similar in spirit to a two-layer skip list.

At the time of testing, a `rangeListNodeBucketSize` of `8` seemed to strike
the best balance between fewer list nodes and larger list nodes based on
benchmarking results.

```
RangeList-4    1.30µs ± 3%    0.84µs ± 2%  -35.11%  (p=0.000 n=9+10)

RangeList-4      444B ± 0%       12B ± 0%  -97.30%  (p=0.000 n=10+10)

RangeList-4      21.0 ± 0%      12.0 ± 0%  -42.86%  (p=0.000 n=10+10)
```

The first commit is much smaller than what the diff is claiming..

### DIFF
--- a/pkg/util/interval/interval.go
+++ b/pkg/util/interval/interval.go
@@ -59,6 +59,10 @@ type inclusiveOverlapper struct{}
 
 // Overlap checks where a and b overlap in the inclusive way.
 func (overlapper inclusiveOverlapper) Overlap(a Range, b Range) bool {
+	return overlapsInclusive(a, b)
+}
+
+func overlapsInclusive(a Range, b Range) bool {
 	return a.Start.Compare(b.End) <= 0 && b.Start.Compare(a.End) <= 0
 }
 
@@ -70,6 +74,10 @@ type exclusiveOverlapper struct{}
 
 // Overlap checks where a and b overlap in the exclusive way.
 func (overlapper exclusiveOverlapper) Overlap(a Range, b Range) bool {
+	return overlapsExclusive(a, b)
+}
+
+func overlapsExclusive(a Range, b Range) bool {
 	return a.Start.Compare(b.End) < 0 && b.Start.Compare(a.End) < 0
 }
 
--- a/pkg/util/interval/range_group.go
+++ b/pkg/util/interval/range_group.go
@@ -81,19 +81,161 @@ type RangeGroupIterator interface {
 	Next() (Range, bool)
 }
 
-// rangeList is an implementation of a RangeGroup using a linked
-// list to sequentially order non-overlapping ranges.
+const rangeListNodeBucketSize = 8
+
+type rangeListNode struct {
+	slots [rangeListNodeBucketSize]Range // ordered, non-overlapping
+	len   int
+}
+
+func newRangeListNodeWithRange(r Range) *rangeListNode {
+	var n rangeListNode
+	n.push(r)
+	return &n
+}
+
+func (n *rangeListNode) push(r Range) {
+	n.slots[n.len] = r
+	n.len++
+}
+
+func (n *rangeListNode) full() bool {
+	return n.len == len(n.slots)
+}
+
+func (n *rangeListNode) min() Comparable {
+	return n.slots[0].Start
+}
+
+func (n *rangeListNode) max() Comparable {
+	return n.slots[n.len-1].End
+}
+
+// findIdx finds the upper-bound slot index that the provided range should fit
+// in the rangeListNode. It also returns whether the slot is currently occupied
+// by an overlapping range.
+func (n *rangeListNode) findIdx(r Range, inclusive bool) (int, bool) {
+	overlapFn := overlapsExclusive
+	passedCmp := 0
+	if inclusive {
+		overlapFn = overlapsInclusive
+		passedCmp = -1
+	}
+	for i, nr := range n.slots[:n.len] {
+		switch {
+		case overlapFn(nr, r):
+			return i, true
+		case r.End.Compare(nr.Start) <= passedCmp:
+			// Past where overlapping ranges would be.
+			return i, false
+		}
+	}
+	return n.len, false
+}
+
+// rangeList is an implementation of a RangeGroup using a bucketted linked list
+// to sequentially order non-overlapping ranges.
 //
 // rangeList is not safe for concurrent use by multiple goroutines.
 type rangeList struct {
-	ll list.List
+	ll  list.List
+	len int
 }
 
 // NewRangeList constructs a linked-list backed RangeGroup.
 func NewRangeList() RangeGroup {
 	var rl rangeList
 	rl.ll.Init()
+	rl.ll.PushFront(&rangeListNode{})
 	return &rl
+}
+
+// findNode returns the upper-bound node that the range would be bucketted in,
+// along with that node's previous element. It also returns whether the range
+// overlaps with the bounds of the node.
+func (rl *rangeList) findNode(r Range, inclusive bool) (prev, cur *list.Element, inCur bool) {
+	if err := rangeError(r); err != nil {
+		panic(err)
+	}
+	reachedCmp := 1
+	passedCmp := 0
+	if inclusive {
+		reachedCmp = 0
+		passedCmp = -1
+	}
+	for e := rl.ll.Front(); e != nil; e = e.Next() {
+		n := e.Value.(*rangeListNode)
+		if n.len == 0 {
+			// The node is empty. This must be the last node in the list.
+			return prev, e, false
+		}
+		// Check if the range starts at a value less than (or equal to, for
+		// inclusive) the maximum value in this node. This is what we want.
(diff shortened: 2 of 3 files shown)


asubiotto:
Thanks for the work on this. This looks good to me but I would definitely get the review of somebody more familiar with this code.

What `concurrency` did you run `tpcc` with? IIRC the CPU percent with`concurrency=32` was pretty elevated without this change.

Regarding using `RangeList` vs `RangeTree`, the intuitive thing to do would be to use a `RangeTree` but if I'm not mistaken, the big problem there is that the amount of allocations we do doesn't justify the search speedup, especially since the command queue doesn't grow to be that big (although this is just the workloads we have tested with) + a tendency to decrease the number of ranges by merging. I would switch to using a `RangeList` and then have a closer look at `RangeTree` and see what we could do to improve it. cc @jordanlewis who has spent some time improving allocations in the `RangeTree`.

---

Reviewed 1 of 1 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/util/interval/range_group.go, line 336 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-Kx8l5YqSht2LTYaTVIA:-Kx8l5YqSht2LTYaTVIB:btwqb47) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L336)):*
> ```Go
> 				// If we broke, that means index j and up are still needed in
> 				// the current node. Shift these over.
> 				copy(mergeN.slots[mergeStart:mergeStart+origLen-j], mergeN.slots[j:origLen])
> ```

I think you could put this into the `else` statement and remove the `broke` variable. 

---

*[pkg/util/interval/range_group.go, line 362 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-Kx8mxKi3mX4t5p8uM5G:-Kx8mxKi3mX4t5p8uM5H:baswjxq) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L362)):*
> ```Go
> 	if prevN != nil {
> 		// There is a previous node and it is full. If it was not, we would have
> 		// fell in one of the two cases above.
> ```

I think the second sentence is unnecessary here; I would just remove it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-:-Kx8oAXkOg9zlUq5OJJH:b-cb1jv9)*

jordanlewis:
:lgtm: for the first 2 commits, not sure about changing the default right now but I think you could get the first 2 in soon!

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[pkg/util/interval/range_group.go, line 119 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-KxDhw5YFjx_vcPxy2n5:-KxDhw5YFjx_vcPxy2n6:biygo9c) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L119)):*
> ```Go
> func (n *rangeListNode) findIdx(r Range, inclusive bool) (int, bool) {
> 	for i, nr := range n.slots[:n.len] {
> 		if inclusive {
> ```

I think this segment could be DRY'd a little. The only difference with `inclusive` is the `r.End.Compare` check, right? What about:

```
maxForCompare := 0
if inclusive {
   maxForCompare = -1
}

for i, nr := range n.slots[:n.len] {
  switch {
  case InclusiveOverlapper.Overlap(np, r):
    return i, true
  case r.End.Compare(nr.Start) <= maxForCompare
    return i, false
}
return n.len, false
```

It might be mildly more efficient as well since there's less comparison in the loop, but maybe it's pointless with branch prediction.

---

*[pkg/util/interval/range_group.go, line 173 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-KxDjSms4SM0SeJt67_2:-KxDjSms4SM0SeJt67_3:b-81u9fo) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L173)):*
> ```Go
> 			// Check if the range starts at a value less than or equal to the
> 			// maximum value in this node. This is what we want.
> 			if n.max().Compare(r.Start) >= 0 {
> ```

More DRYing? I guess you made a conscious choice to not do that but I feel like these repeated stanzas are a little hard to read since you need to stare at them to see if there's more differences than just > vs >=.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-:-KxDmpXyxpgtD6TRnTpo:b-g3wdzz)*

nvanbenschoten:
Thanks for the reviews! I'll pull the third commit into a different PR that we can experiment with.

---

---

*[pkg/util/interval/range_group.go, line 119 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-KxDhw5YFjx_vcPxy2n5:-KxJzMACQoPZk4WrgnWz:b-r8rns4) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L119)):*

I think this segment could be DRY'd a little. The only difference with `inclusive` is the `r.End.Compare` check, right? What about:

```
maxForCompare := 0
if inclusive {
   maxForCompare = -1
}

for i, nr := range n.slots[:n.len] {
  switch {
  case InclusiveOverlapper.Overlap(np, r):
    return i, true
  case r.End.Compare(nr.Start) <= maxForCompare
    return i, false
}
return n.len, false
```

It might be mildly more efficient as well since there's less comparison in the loop, but maybe it's pointless with branch prediction.
</blockquote></details>

I wrote this differently at first (like your suggested structure) but used closures to deal with the `< 0` vs `<= 0` decision. I then felt bad about the closure overhead and ended up here. The `maxForCompare` idea is great! Done.

---

*[pkg/util/interval/range_group.go, line 173 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-KxDjSms4SM0SeJt67_2:-KxK-56eu6anJRHkwbHn:b-ishv4w) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L173)):*

More DRYing? I guess you made a conscious choice to not do that but I feel like these repeated stanzas are a little hard to read since you need to stare at them to see if there's more differences than just > vs >=.
</blockquote></details>

Again, I was trying to change the comparator at first and threw that out to settle on this. Changing the operand is blowing my mind!

---

*[pkg/util/interval/range_group.go, line 336 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-Kx8l5YqSht2LTYaTVIA:-KxK2pgiq0-D5kVexS4r:b-es5eje) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L336)):*

I think you could put this into the `else` statement and remove the `broke` variable. 
</blockquote></details>

Yep, you're completely right. Done.

---

*[pkg/util/interval/range_group.go, line 362 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-Kx8mxKi3mX4t5p8uM5G:-KxK1mydC6pvppuPj4HH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/840874986bd2fc5d8185e6ecad658a536fd5a6f8/pkg/util/interval/range_group.go#L362)):*

I think the second sentence is unnecessary here; I would just remove it.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19361#-:-KxKVJFUShIf8AFWHdIi:b8auo3w)*

PR #19364, opened 2017-10-19
cli/sql: help users who enter a single '?' further

knz:
With the advent of JSON in CockroachDB, the single '?' token has
become a JSON operator. Since CockroachDB also supports contextual
help, the contextual help token has become '??'. However users may not
know this.

This patch ensures that if a user ends a line with a single '?' token,
a note is printed to the screen before the user is invited to complete
the statement (as the final ? may be part of a JSON expression
continued in the next line).  For example:

```
root@:26257/> select 1 ?
Note: a single '?' is a JSON operator. If you want contextual help, use '??'.
           -> 2;
pq: unsupported comparison operator: <int> ? <int>
root@:26257/>
```

cc @jseldess

### DIFF
--- a/pkg/cli/interactive_tests/test_contextual_help.tcl
+++ b/pkg/cli/interactive_tests/test_contextual_help.tcl
@@ -62,6 +62,18 @@ eexpect "See also"
 eexpect root@
 end_test
 
+start_test "Check that a useful reminder is given if the user mistakenly uses a single ?."
+send "select ?\r"
+eexpect "Note:"
+eexpect JSON
+eexpect "use '??'"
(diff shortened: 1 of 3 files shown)


justinj:
Code LGTM, I think CI is saying that there were some other functions that were using `isEndOfStatement` that need to be changed to its new signature.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19364#-:-Kwose-mYLqYvqXYUCW_:b-f77zcy)*

jseldess:
Thanks, @knz!

PR #19220, opened 2017-10-12
rfcs: data structures for SQL logical planning

knz:
cc @albler; Alberto, could you review this too and provide feedback?

Note this is merely an outline to set the scene, I think this RFC is not meant to decide all the details, however it should be sufficient to ensure that multiple people can work on their own prototype and use the same terminology and guiding principles.

PR #19366, opened 2017-10-19
rfcs: complementary RFC to inventory the plan properties

knz:
Note: I intend to merge this PR into #19135 together with #19220.

### DIFF
--- a/docs/RFCS/sql_plan_properties.md
+++ b/docs/RFCS/sql_plan_properties.md
@@ -1,0 +1,909 @@
+- Feature Name: SQL plan properties
+- Status: draft
+- Start Date: 2017-10-15
+- Authors: Peter Mattis, Radu Berinde, knz
+- RFC PR: (PR # after acceptance of initial draft)
+- Cockroach Issue: (one or more # from the issue tracker)
+
+# Summary
+
+This RFC proposes an *inventory of logical and physical plan properties*
+that can be useful during logical planning. Reminder: the properties
+are annotations next to logical and physical (sub-)plans that are
+used during plan transformations, mainly for the purpose of
+optimization.
+
+As such, this RFC is complementary to:
+
+- [RFC: SQL Logical planning](https://github.com/cockroachdb/cockroach/pull/19135)
+- [RFC: Data structures for logical planning](https://github.com/cockroachdb/cockroach/pull/19220)
+
+and focuses on the definition of "properties", with the other RFCs
+providing context on how they are to be used / integrated.
+
+# Motivation
+
+The situation we have to contend with is that 1) the set of properties
+that are useful is (and will be) changing over time, in tandem with
+the development of new optimizations 2) new understanding gained over
+time will lead to tweaking ever slightly the technical or formal
+definitions of properties. It is thus hard to propose a fixed plan, in
+the traditional shape of an RFC, of a set of properties ahead of time
+and commit to that plan ahead of time.
+
+What will not change however, is the need to share *the same words
+across the team to designate the same things over time*, whether or
+not the "things" in question are currently in use, not in use any
+more, or not in use yet.
+
+The main purpose of this RFC is thus not to provide implementation
+details nor an overarching integration story. Instead, **its main
+purpose is to give stable names to definitions.**
+
+A secondary purpose is to provide some context to these definitions,
+so that the contributors who end up working with them **gain an
+understanding of their properties more easily than by just reading the
+code where these properties and their uses are being prototyped.**
+
+A third purpose is to establish a baseline when the implementation
+will need verification / review. By having the description
+of the properties in prose, a reviewer can then compare the code
+to the specification of the property to ascertain that the
+code is indeed matching the intent.
+
+# Guide-level explanation
+
+## Summary
+
+| Name                    | Type      | Category  | Origin    | Kind       | Level | Used by (example)         | Synthetic notation     |
+|-------------------------|-----------|-----------|-----------|------------|-------|---------------------------|------------------------|
+| injectivity             | intset    | L, sem    | derived   | scalar     | opt   | TBD                       | `x INJECTIVE{...}`     |
+| monotonicity            | cf. below | L, sem    | derived   | scalar     | opt   | TBD                       | `x MONOTONIC(...)`     |
+| constness               | bool      | L, sem    | derived   | scalar     | opt   | TBD                       |                        |
+| known value             | int       | L, sem    | derived   | scalar     | opt   | predicate push-down       | `x = v`                |
+| nullability             | intset    | L, sem    | derived   | common     | opt   | TBD                       | `x NULL` / `NOT NULL`  |
+| constant columns        | intset    | L, sem    | derived   | common     | opt   | TBD                       | `x CONST`              |
+| variable dependencies   | intset    | L, x-func | derived   | common     | mand. | decorrelation             |                        |
+| result columns          | intset    | L, sem    | opt req.  | relational | mand. | equivalence comparisons   |                        |
+| used columns            | intset    | L, x-func | derived   | relational | opt   | physical planning         |                        |
+| required columns        | intset    | L, x-func | derived   | relational | opt   | join transformations      |                        |
+| unique sets             | intsetset | L, sem    | derived   | relational | opt   | TBD                       | `UNIQUE {...}`         |
+| key sets                | intsetset | L, sem    | derived   | relational | opt   | TBD                       | `KEY {...}`            |
+| equivalency groups      | intsetset | L, sem    | derived   | relational | opt   | TBD                       | `EQ {...}`             |
+| range constraints       | cf. below | L, sem    | derived   | rel(comm?) | opt   | TBD                       | `CHECK ...`            |
+| req. presentation order | []int     | L, x-func | opt req.  | relational | opt   | physical planning         | `REQUIRE PRESENT(...)` |
+| required ordering       | cf. below | L, sem    | opt req.  | relational | opt   | ordering transformations  | `REQUIRE ORDER BY(...)` |
+| required algorithm      | enum      | L, x-func | opt req.  | relational | opt   | physical planning, search |                        |
+| required scan index     | cf. below | L, x-func | opt req.  | relational | opt   | physical planning, search |                        |
+| apparent presentation   | []int     | P, x-func | derived   | relational | mand. | physical planning         | `PRESENTING(...)`      |
+| apparent orderings      | cf. below | P, sem    | derived   | relational | opt   | TBD                       | `ORDERED BY(...)`      |
+| algorithm               | enum      | P, x-func | derived.  | relational | mand. | physical planning         |                        |
+| scan index              | string    | P, x-func | derived   | relational | mand. | physical planning         |                        |
+| spans for scans         | []span    | P, sem    | derived   | relational | opt   | TBD                       |                        |
+
+Note: in the literature "functional dependencies" is the group formed
+by variable dependencies, equivalency groups and range constraints.
+
+We distinguish the *category*:
+
+- *logical* (L) properties which define equivalency classes, and are
+  shared among all m-expressions in the class.
+- *physical* (P) properties which can be different per m-expression in a
+  class.
+
+We distinguish two *sub-categories*:
+
+- *semantic* (sem) properties which describe the relationships between input
+  and output values of (sub-)queries.
+- *extra-functional* (x-func) properties that do not carry information about
+  the input/output values.
+
+We distinguish the *origin*:
+
+- *optionally required* properties which can be specified in the
+  query.  For example, presentation order. Optionally required
+  properties are necessarily logical -- when specified, they must not
+  disappear or be replaced in alternatives within an equivalency
+  class. We use **verbs** in their synthetic notation, if any,
+  e.g. ORDER BY / PRESENT.
+
+- purely *derived* properties that are computed during query analysis
+  and never specified explicitly, for example needed columns.  Can be
+  either logical or physical.  All physical properties are necessarily
+  derived (possibly from optionally required logical properties).  We
+  use **adjectives** / **gerunds** in their synthetic notation, if
+  any, e.g. ORDERED BY / PRESENTING.
+
+(Mismatches between required and derived properties cause the
+introduction of "enforcer nodes". See the planning RFC for more information.)
+
+We distinguish the *level* (term to be refined!):
+
+- *mandatory* properties must be present in the expression. A
+  mandatory property that is not also specified explicitly in the
+  query (if it's optionally requried) must be derived for every
+  expression in the final plan.
+
+- *optional* properties can be omitted. The absence of the property is
+  not equivalent to the presence of any value for the property, however,
+  i.e. an expression where the property is omitted is in a different
+  class than all the expressions that have the property.
+
+We distinguish the *kind*:
+
+- *relational* properties that are useful for relational expressions.
+- *scalar* properties that are useful for scalar expressions.
+
+Properties can be both relational and scalar; we'll call them *common*
+in contexts where a separate word is needed.
+
+## Injectivity
+
+- Type: intset
+- Category: logical, semantic
+- Origin: derived
+- Kind: scalar
+- Level:  optional
+
+An injective function is a function that preserves distinctness: it
+never maps distinct elements of its domain to the same element of its
+codomain.
+
+If an expression is injective, then if it is known that the operand is
+UNIQUE, then the result is also UNIQUE.
+
+For example, the function `exp(x) = e^x` is injective.
+
+A multi-argument function can be injective on one argument or a set
+thereof. For example the `tuple` function is injective on the tuple of
+its arguments; the `pow` function is injective on its first argument
+and second argument separately if the other argument is constant.
+
+Example synthetic notation:
+
+```
+-- query:
+> select exp(k) as a, log(k) as b, pow(k, 3) as c, pow(3, k) as d, (v, k) as e FROM kv
+
+-- properties for the result:
+PROPERTIES(
+  a INJECTIVE{k}
+  b INJECTIVE{k}
+  c INJECTIVE{k},
+  d INJECTIVE{k},
+  e INJECTIVE{k, v}
+)
+```
+
+## Monotonicity
+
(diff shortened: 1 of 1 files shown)


petermattis:
Glad to see this inventory getting done. Some of the properties you listed I would not consider properties, but metadata or implementation data on the expressions. I suppose that depends on what our definition of properties is. I'm defining properties as common metadata that is maintained on a per expression basis. Individual expressions will also have additional metadata. For example, an `index scan` operation needs to know the index it is scanning, but parent expressions only need to know the ordering that produces.

---

Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

*[docs/RFCS/sql_plan_properties.md, line 82 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-KwtyWHHX2PqUIOkEwTQ:-KwtyWHHX2PqUIOkEwTR:b690eu3) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L82)):*
> ```Markdown
> | algorithm               | enum      | P, x-func | derived.  | relational | mand. | physical planning         |                        |
> | scan index              | string    | P, x-func | derived   | relational | mand. | physical planning         |                        |
> | spans for scans         | []span    | P, sem    | derived   | relational | opt   | TBD                       |                        |
> ```

I think this table should be in an appendix as I find it obfuscates rather than clarifies. In its place, I'd put a list of the logical and physical properties that are hyperlinked to their sections. 

---

*[docs/RFCS/sql_plan_properties.md, line 90 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwtynw4alqLBFK01eqB:-Kwtynw4alqLBFK01eqC:brbqg6n) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L90)):*
> ```Markdown
> 
> - *logical* (L) properties which define equivalency classes, and are
>   shared among all m-expressions in the class.
> ```

I'd like to leave m-expressions and equivalency classes out of the definition of logical and physical properties. As an alternative:

*logical* properties are those that can be derived from the extended relational algebra. For example, NULL-ability of columns and keys, weak keys and foreign keys.

*physical* properties are those that can be derived from the specific implementation of relational algebra expressions. The sole physical property is ordering which is not a consideration in the extended relational algebra. (`ORDER BY` sits outside of the extended relational algebra).

---

*[docs/RFCS/sql_plan_properties.md, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu-HYqkBbl-IN-GbPi:-Kwu-HYun-4ESJF7Rio5:bdkerlw) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L94)):*
> ```Markdown
>   class.
> 
> We distinguish two *sub-categories*:
> ```

This felt overly categorical on my first read through the document. While I don't mind the categorization being done, placing the categorization up front got in my way of reading about the properties. Perhaps move to a terminology section at the end of the document.

---

*[docs/RFCS/sql_plan_properties.md, line 101 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu-PCkBSBQ0oKwCmhL:-Kwu-PCkBSBQ0oKwCmhM:b6fzoxf) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L101)):*
> ```Markdown
>   the input/output values.
> 
> We distinguish the *origin*:
> ```

Similar to the above, this felt overly categorical on my first read through the document.

---

*[docs/RFCS/sql_plan_properties.md, line 138 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu-ZRoaauLIPjT3xRt:-Kwu-ZRoaauLIPjT3xRu:b-wzgrx) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L138)):*
> ```Markdown
> 
> Properties can be both relational and scalar; we'll call them *common*
> in contexts where a separate word is needed.
> ```

I feel like this section is "burying the lede" which means that you're keeping me from the good stuff. I know you prefer to have all of these definitions upfront, but I think the average reader would be find with them in a terminology section at the end.

---

*[docs/RFCS/sql_plan_properties.md, line 351 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu18mRXJq8cJDM-RkT:-Kwu18mRXJq8cJDM-RkU:b8zqla5) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L351)):*
> ```Markdown
> (No synthetic notation)
> 
> Example with memo:
> ```

I think using `memo` for explanation here and elsewhere in this document is less clear than if you showed the relational algebra expressions in a nested tree structure.

---

*[docs/RFCS/sql_plan_properties.md, line 590 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu1cPhV6-DxktSj62W:-Kwu1cPhV6-DxktSj62X:bgos4dd) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L590)):*
> ```Markdown
> ```
> 
> ## Range constraints
> ```

I've been calling this column value constraints. `Range` is a bit of an overloaded term in Cockroach.

---

*[docs/RFCS/sql_plan_properties.md, line 843 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0eNgggDwspiFWtEX:-Kwu0eNgggDwspiFWtEY:blgri1o) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L843)):*
> ```Markdown
> ```
> 
> ## (Actual) algorithm
> ```

I don't think this is a property, but instead metadata on the expression. 

---

*[docs/RFCS/sql_plan_properties.md, line 860 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0cPHQYi-TjzbMmzP:-Kwu0cPHQYi-TjzbMmzQ:b-ltbpol) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L860)):*
> ```Markdown
> relational expression in the plan.
> 
> ## Scan index
> ```

I don't think this is a property. When enumerating the alternative ways of scanning a table, we would expand all of the possible indexes. This seems like metadata about the index scan rather than a property that needs to be propagated to other expressions.

---

*[docs/RFCS/sql_plan_properties.md, line 876 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0dXFFxKKIUwp5xhg:-Kwu0dXFFxKKIUwp5xhh:b-lcqvau) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L876)):*
> ```Markdown
> query plan.
> 
> ## Spans for scans
> ```

Similar to `Scan index`, I don't think this is a property, but metadata on the `index scan` expression.

---

*[docs/RFCS/sql_plan_properties.md, line 897 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu06_omC6rXEth91ym:-Kwu06_omC6rXEth91yn:b-ndc6e7) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L897)):*
> ```Markdown
> - column labels. Labels disappear from the expressions during
>   the prep phase, and only remains at the top level, as
>   an attribute of the results for the entire query.
> ```

While it isn't necessary to maintain labels for correctness, doing so can aid debug-ability. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-:-Kwu2YlwFqYX6FbW7cBR:bb22yxp)*

knz:
I am moving this document over to #19135. I will process your last few comments there.

knz:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/sql_plan_properties.md, line 82 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-KwtyWHHX2PqUIOkEwTQ:-KwymTk3ynnPZ9XLPQoD:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L82)):*

I think this table should be in an appendix as I find it obfuscates rather than clarifies. In its place, I'd put a list of the logical and physical properties that are hyperlinked to their sections. 
</blockquote></details>

Done.

---

*[docs/RFCS/sql_plan_properties.md, line 90 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwtynw4alqLBFK01eqB:-KwyheLiyJb9T8v6DFs9:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L90)):*

I'd like to leave m-expressions and equivalency classes out of the definition of logical and physical properties. As an alternative:

*logical* properties are those that can be derived from the extended relational algebra. For example, NULL-ability of columns and keys, weak keys and foreign keys.

*physical* properties are those that can be derived from the specific implementation of relational algebra expressions. The sole physical property is ordering which is not a consideration in the extended relational algebra. (`ORDER BY` sits outside of the extended relational algebra).
</blockquote></details>

Done.

---

*[docs/RFCS/sql_plan_properties.md, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu-HYqkBbl-IN-GbPi:-KwyhfUAKk2B0RJ1MRJt:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L94)):*

This felt overly categorical on my first read through the document. While I don't mind the categorization being done, placing the categorization up front got in my way of reading about the properties. Perhaps move to a terminology section at the end of the document.
</blockquote></details>

Done.

---

*[docs/RFCS/sql_plan_properties.md, line 101 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu-PCkBSBQ0oKwCmhL:-Kwyhgyji4Tns7hFZQdt:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L101)):*

Similar to the above, this felt overly categorical on my first read through the document.
</blockquote></details>

Done.

---

*[docs/RFCS/sql_plan_properties.md, line 138 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu-ZRoaauLIPjT3xRt:-KwyhhPUmezJVmHGwFgJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L138)):*

I feel like this section is "burying the lede" which means that you're keeping me from the good stuff. I know you prefer to have all of these definitions upfront, but I think the average reader would be find with them in a terminology section at the end.
</blockquote></details>

Done.

---

*[docs/RFCS/sql_plan_properties.md, line 351 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu18mRXJq8cJDM-RkT:-KwyhkdIkDd2R71agI1m:b-byn2sy) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L351)):*

I think using `memo` for explanation here and elsewhere in this document is less clear than if you showed the relational algebra expressions in a nested tree structure.
</blockquote></details>

I added the expression tree, for clarity. I am keeping the memo because we should not lose track of the fact that properties are carried by classes, not expression nodes.

---

*[docs/RFCS/sql_plan_properties.md, line 590 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu1cPhV6-DxktSj62W:-Kwyi2Bine0n5qiCFHaG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L590)):*

I've been calling this column value constraints. `Range` is a bit of an overloaded term in Cockroach.
</blockquote></details>

Done.

---

*[docs/RFCS/sql_plan_properties.md, line 843 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0eNgggDwspiFWtEX:-KwyiApX4P3VPuPJIGXV:bbdo1ey) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L843)):*

I don't think this is a property, but instead metadata on the expression. 
</blockquote></details>

I do not see what is this distinction you are making between physical properties and expression metadata.

Two m-expressions that have the same operator and same operands can only differ in their physical properties. If the only thing that differs is the algorithm, then necessarily the algorithm must be a physical property.

---

*[docs/RFCS/sql_plan_properties.md, line 860 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0cPHQYi-TjzbMmzP:-KwyiLhLALGplLJrGPVZ:bshwmcf) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L860)):*

I don't think this is a property. When enumerating the alternative ways of scanning a table, we would expand all of the possible indexes. This seems like metadata about the index scan rather than a property that needs to be propagated to other expressions.
</blockquote></details>

"We would expand all of the possible indexes" == "Generate all the possible scan expressions, differing from each other in which index they use"

Two m-expressions that have the same operator and same operands can only differ in their physical properties. If the only thing that differs is the index name, then necessarily the index name must be a physical property.

---

*[docs/RFCS/sql_plan_properties.md, line 876 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0dXFFxKKIUwp5xhg:-Kwyihocp7XweGltlaDg:b-b3e8vm) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L876)):*

Similar to `Scan index`, I don't think this is a property, but metadata on the `index scan` expression.
</blockquote></details>

Two m-expressions that have the same operator and same operands can only differ in their physical properties. If the only thing that differs is the list of spans, then necessarily the list of spans must be a physical property.

---

*[docs/RFCS/sql_plan_properties.md, line 897 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu06_omC6rXEth91ym:-KwyilX_eBEFT1haMxU9:b-ir904y) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L897)):*

While it isn't necessary to maintain labels for correctness, doing so can aid debug-ability. 
</blockquote></details>

Agreed. Added a note.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/sql_plan_properties.md, line 843 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0eNgggDwspiFWtEX:-Kwz-UnqJg-xGTR9YIYA:b-ls1sg8) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L843)):*

I do not see what is this distinction you are making between physical properties and expression metadata.

Two m-expressions that have the same operator and same operands can only differ in their physical properties. If the only thing that differs is the algorithm, then necessarily the algorithm must be a physical property.
</blockquote></details>

The distinction I'm making is that a property is something we'd want to maintain on every expression while something like the algorithm is private to the expression. My guess is that you're looking at "property" in a general sense, while I'm trying to restrict it to attributes that are common across either logical or physical nodes. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[docs/RFCS/sql_plan_properties.md, line 843 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366#-Kwu0eNgggDwspiFWtEX:-KwzJuwrLStVytL3i5nW:bigkf4i) ([raw file](https://github.com/cockroachdb/cockroach/blob/e4a9d487cb5d997771d935012ed313e55ccfb82d/docs/RFCS/sql_plan_properties.md#L843)):*

The distinction I'm making is that a property is something we'd want to maintain on every expression while something like the algorithm is private to the expression. My guess is that you're looking at "property" in a general sense, while I'm trying to restrict it to attributes that are common across either logical or physical nodes. 
</blockquote></details>

Ah now I understand what you mean. I will try to integrate this idea into the definition.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19366)*

PR #13428, opened 2017-02-06
sql: perform index backfill using distsql

vivekmenezes:
One major area of testing that needs to be covered is a test that brings up a cluster with a few nodes, creates a table with range leases spread across the cluster, and runs an index backfill. I'd like to add a test and a benchmark that does this. This seems like a test/framework that would be good for distsql in general.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach/13428)

PR #19372, opened 2017-10-19
sql: remove unnecessary call to SetSystemConfig in index truncation

lgo:
It was originally needed because the ResumeSpan was added to the TableDescriptor. With the
move of ResumeSpan to the Jobs table, it is no longer needed.

cc: @vivekmenezes

### DIFF
--- a/pkg/sql/backfill.go
+++ b/pkg/sql/backfill.go
@@ -290,11 +290,6 @@ func (sc *SchemaChanger) truncateIndexes(
 					defer sc.testingKnobs.RunAfterBackfillChunk()
 				}
 
-				// TODO(vivek): See comment in backfillIndexesChunk.
-				if err := txn.SetSystemConfigTrigger(); err != nil {
-					return err
-				}
-
 				tc := &TableCollection{leaseMgr: sc.leaseMgr}
 				defer tc.releaseTables(ctx)
 				tableDesc, err := sc.getTableVersion(ctx, txn, tc, version)


lego:
Either we should remove it, or fix the TODO as the keyword it is referencing no longer exists.

vivekmenezes:
My understanding is that SetSystemConfig may not be needed any more because the transaction is no longer modifying the table descriptor or system config. Whenever looking at a line of code it's always good to run git blame to see the PR that introduced it.

lego:
The PR which removed the referenced comment is https://github.com/cockroachdb/cockroach/pull/13428. It states
```go
// TODO(vivek): We need to set the system-config trigger before doing any
// transaction writes. This is also called by
// SchemaChanger.maybeWriteResumeSpan, but it is too late at that point. Do
// we even need to be setting the system-config trigger here. We're
// changing the TableDescriptor.Mutations.ResumeSpan, but it isn't clear to
// me that we need to gossip the updated TableDescriptor.
txn.SetSystemConfigTrigger()
```

I'm not too familiar with the schema changer code, but I don't see any modifications of the table descriptor, and it seems that the `TableDescriptor.Mutations.ResumeSpan` referenced in the comment no longer exists -- I would guess this is now `JobDetails.SchemaChangeDetails.ResumeSpan`.

It seems like we can indeed remove it. Would you agree?

vivekmenezes:
Yes, please do remove the call and change the commit message to reflect the new change:

sql: remove unnecessary call to SetSystemConfig in index truncation

It was originally needed because the ResumeSpan was added to the TableDescriptor. With the
move of ResumeSpan to the Jobs table, it is no longer needed.

lego:
Great! Done :)

vivekmenezes:
LGTM

PR #19374, opened 2017-10-19
build: nuke ignored files in teamcity-assert-clean

jordanlewis:


tschottdorf:
I can't prove to myself that this is doing the right thing, so my preference would be to unconditionally nuke. That, or we should add supple commentary! :)

tbg (approved):

benesch (approved):

jordanlewis:
PTAL @tschottdorf

tbg (commented):
Wanna pull the cord on doing the right thing for today, and just nuke unconditionally? This is a huge PITA for everyone involved.

We can do cleanup in separate PR.

PR #19379, opened 2017-10-19
sql: add builtin function for random UUID generation 

### LINKED ISSUE #19368: sql: implement gen_random_uuid builtin
Some ORMs ([ActiveRecord](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/12) being the salient example) use `gen_random_uuid()` as the default column value for `UUID` columns by default. Add this builtin. It can probably be implemented identically to our pre-existing UUID generation builtin.

https://www.postgresql.org/docs/9.6/static/pgcrypto.html

dsaveliev:
Closes #19368


pkg/sql/logictest/testdata/logic_test/builtin_function
```diff
@@ -1388,6 +1388,13 @@ select ln(4.0786335175292462e+34::decimal)
 ----
 79.693655171940461633
 
+query IB
+SELECT LENGTH(GEN_RANDOM_UUID()),
+       GEN_RANDOM_UUID() = GEN_RANDOM_UUID()
+----
+16
+false
+
```
justinj: nit: I know it doesn't match the one right below here, but I think this test is short enough that the query and expected results can all go on one line.

pkg/sql/parser/builtins.go
```diff
@@ -366,6 +366,18 @@ var Builtins = map[string][]Builtin{
 		},
 	},
 
+	"gen_random_uuid": {
+		Builtin{
+			Types:      ArgTypes{},
+			ReturnType: fixedReturnType(TypeBytes),
```
justinj: Our current `uuid_v4` returns `BYTES` for backwards-compatibility reasons (from before we supported `UUID` columns), it would be really great if newer versions could match Postgres here by returning a `TypeUUID` instead!

In addition, we mark some functions as `impure` so that our planning code knows how to deal with them. Since this function is impure, could you add the `impure` flag? (see [here](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/parser/builtins.go#L2339-L2348)). And while we're at it we can tag it as a `categoryIDGeneration` (which is used for our docs) as well.

justinj (commented):
Hi @dsaveliev! Thanks so much for the PR! It looks really good - I just added a few comments 😄

pkg/sql/parser/builtins.go
dsaveliev: Thanks for comments! 
You are right about the type - I was confused by the nearby UUID functions.
It should be better by now.

pkg/sql/logictest/testdata/logic_test/builtin_function
dsaveliev: Done.

pkg/sql/parser/builtins.go
```diff
@@ -366,6 +366,20 @@ var Builtins = map[string][]Builtin{
...
+		Builtin{
+			Types:      ArgTypes{},
+			ReturnType: fixedReturnType(TypeUUID),
+			category:   categoryIDGeneration,
+			impure:     true,
+			fn: func(_ *EvalContext, _ Datums) (Datum, error) {
+				uv := uuid.MakeV4()
+				return NewDUuid(DUuid{uv}), nil
+			},
+			Info: "Generates a random UUID and returns its byte string representation.",
```
justinj: Sorry to ask for another change, but could we also update this description, since it doesn't return the byte string any more?

dsaveliev: Oh, shame on me, just a minute.

dsaveliev: I explicitly specified the return type to avoid confusion with BYTES type.

justinj:
One more small thing - we have a doc generator that needs to be run manually (I forgot to mention this before) or else our CI will complain that it wasn't updated, you can fix this by running
```
make generate PKG=./docs/...
```
and committing the result.

dsaveliev:
I updated the docs. Thank you for your patience! 😊

justinj:
No, thank you for the contribution @dsaveliev! 😄

Issue #16585, opened 2017-06-18
storage: sporadic context cancellation after 0.0s of attempting command RequestLease


tschottdorf:
current master, `./cockroach start --insecure --logtostderr` (new single node):

```
E170617 22:55:26.441058 152 storage/queue.go:739  [n1,replicate] 10 replicas failing with "0 of 0 stores with all attributes matching []; likely not enough nodes in cluster"
I170617 22:55:36.415164 14 server/status/runtime.go:225  [n1] runtime stats: 72 MiB RSS, 134 goroutines, 14 MiB/11 MiB/34 MiB GO alloc/idle/total, 19 MiB/35 MiB CGO alloc/total, 62.69cgo/sec, 0.01/0.00 %(u/s)time, 0.00 %gc (1x)
W170617 22:55:45.928204 20037 storage/replica.go:2524  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
I170617 22:55:46.417136 14 server/status/runtime.go:225  [n1] runtime stats: 72 MiB RSS, 134 goroutines, 20 MiB/5.6 MiB/34 MiB GO alloc/idle/total, 19 MiB/35 MiB CGO alloc/total, 79.39cgo/sec, 0.01/0.00 %(u/s)time, 0.00 %gc (0x)
W170617 22:55:50.430619 20031 storage/replica.go:2524  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
I170617 22:55:56.415088 14 server/status/runtime.go:225  [n1] runtime stats: 72 MiB RSS, 134 goroutines, 16 MiB/8.7 MiB/34 MiB GO alloc/idle/total, 19 MiB/35 MiB CGO alloc/total, 62.41cgo/sec, 0.01/0.00 %(u/s)time, 0.00 %gc (1x)
W170617 22:55:56.432489 20053 storage/replica.go:2524  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
W170617 22:56:03.927933 20044 storage/replica.go:2524  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
[...]

PR #19380, opened 2017-10-19
storage: Avoid proposing commands if context is canceled

### LINKED ISSUE #16911: storage: context cancellation after 0.0s of attempting command RequestLease
While doing various performance tests on `denim`, I noticed these log messages:

```
W170706 20:27:35.037503 4599635 storage/replica.go:2544  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.1s of attempting command RequestLease [/System/"",/Min)
W170706 20:27:39.449651 4611708 storage/replica.go:2544  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
W170706 20:27:44.706887 4626744 storage/replica.go:2544  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
W170706 20:27:49.189390 4641731 storage/replica.go:2544  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
W170706 20:27:49.210479 4641773 storage/replica.go:2544  [n1,s1,r2/1:/System/{-tsd}] context cancellation after 0.0s of attempting command RequestLease [/System/"",/Min)
```

The messages are always for `r1` or `r2`. And this seems to be very reproducible. I've seen it occur every time I've looked over the past 3 test runs.

a-robinson:
This seems like a good practice in general, but was motivated mainly by
the "context cancellation after 0.0s of attempting command RequestLease"
that pop up way more often than it seems like they should. This will
still log them, but will avoid the ambiguous errors.

Also, this means that #17014 apparently didn't completely fix #16911,
since I've been seeing these errors while running against master.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -2880,6 +2880,14 @@ func (r *Replica) propose(
 		return nil, nil, noop, roachpb.NewError(err)
 	}
 
+	// Checking the context just before proposing can help avoid ambiguous errors.
+	if err := ctx.Err(); err != nil {
+		errStr := fmt.Sprintf("%s before proposing: %s", err, ba.Summary())
+		log.Warning(ctx, errStr)
+		log.ErrEvent(ctx, errStr)
+		return nil, nil, noop, roachpb.NewError(err)
+	}
+
 	// Must check that the request is in bounds at proposal time in
 	// addition to application time because some evaluation functions
 	// (especially EndTransaction with SplitTrigger) fail (with a


tschottdorf:
You're seeing those on an idle cluster? My fix was specific to async lease renewals.

tbg (approved):

a-robinson:
Not idle, a very loaded cluster. So loaded that the node holding the lease for r1 is constantly using around 80% of its 8 cores.

a-robinson:
If you want access, it's running on GCE and I'm happy to give it to you.

tschottdorf:
Ok. #16911 was about an idle local cluster. I think we're seeing context cancellations in your case because contexts actually time out, for example because intent resolutions have spent time in the intent resolver's semaphore. It's question how we want to handle this in general -- once you load a cluster sufficiently, contexts will expire. Other than making sure the system doesn't have obvious bottlenecks, one thing we'll likely want to do is to just log less.

a-robinson:
I assume the issue is coming from the `expired` code path, so I guess it's just a matter of tracking down why contexts along that code path would be getting canceled or deciding that it's fine to use a detached context for such lease requests.

a-robinson:
> #16911 was about an idle local cluster.

It says it was about "While doing various performance tests on denim", not on an idle cluster :)

tschottdorf:
Ah, I was thinking of its sibling https://github.com/cockroachdb/cockroach/issues/16585. I'll reopen the issue you referenced. Optimistic that these context cancellations are expected (either when held-up intent resolutions are put in charge of extending a lease, though renewal should do it in that case) or when a replica has moved but wasn't GC'ed yet (which is what @m-schneider is looking at these days).

tschottdorf:
Oh, #16911 mentions that the messages are always about range 1. What ranges are you seeing them on?

a-robinson:
I'm also seeing them only on range 1.

It definitely isn't due to "when a replica has moved but wasn't GC'ed yet", since it's a 3 node cluster where replicas aren't being moved. I agree that it isn't particularly shocking to see them, though, given the amount of load on the cluster. I'm also seeing some other commands time out on other ranges, they just always time out in the command queue instead of after proposing.

tschottdorf:
I'll take a look in #16911.

PR #19381, opened 2017-10-19
build: use baked-in deps for java acceptance tests

justinj:
In addition, remove the `target` directory after testing so we don't
leave files sitting around. We could do `mvn clean` but that requires a
plugin I didn't add to the pom.xml and I don't want to eat another
dockerhub run. I'll add it the next time I need to modify the Dockerfile
(which I suspect will be soon).


jordanlewis (approved):
Plz delete the extra dependencies

PR #19386, opened 2017-10-19
storage: deflake TestNodeHeartbeatCallback

### LINKED ISSUE #19362: storage: TestNodeHeartbeatCallback failed under stress
SHA: https://github.com/cockroachdb/cockroach/commits/2daa6f3e541b34603f5702bcbed0286f28b0552a

Parameters:
```
TAGS=deadlock
GOFLAGS=
```

Stress build found a failed test: https://teamcity.cockroachdb.com/viewLog.html?buildId=382320&tab=buildLog

```
I171019 08:56:50.289097 2430563 storage/store.go:1212  [s1] [n1,s1]: failed initial metrics computation: [n1,s1]: system config not yet available
I171019 08:56:50.289205 2430563 gossip/gossip.go:331  [n1] NodeDescriptor set to node_id:1 address:<network_field:"tcp" address_field:"127.0.0.1:53470" > attrs:<> locality:<> ServerVersion:<major_val:0 minor_val:0 patch:0 unstable:0 > 
W171019 08:56:50.297743 2430563 gossip/gossip.go:1277  [n2] no incoming or outgoing connections
I171019 08:56:50.298283 2430563 storage/store.go:1212  [s2] [n2,s2]: failed initial metrics computation: [n2,s2]: system config not yet available
I171019 08:56:50.298381 2431843 gossip/client.go:129  [n2] started gossip client to 127.0.0.1:53470
I171019 08:56:50.298401 2430563 gossip/gossip.go:331  [n2] NodeDescriptor set to node_id:2 address:<network_field:"tcp" address_field:"127.0.0.1:46730" > attrs:<> locality:<> ServerVersion:<major_val:0 minor_val:0 patch:0 unstable:0 > 
W171019 08:56:50.372419 2430563 gossip/gossip.go:1277  [n3] no incoming or outgoing connections
I171019 08:56:50.372968 2432660 gossip/client.go:129  [n3] started gossip client to 127.0.0.1:53470
I171019 08:56:50.373200 2430563 storage/store.go:1212  [s3] [n3,s3]: failed initial metrics computation: [n3,s3]: system config not yet available
I171019 08:56:50.373328 2430563 gossip/gossip.go:331  [n3] NodeDescriptor set to node_id:3 address:<network_field:"tcp" address_field:"127.0.0.1:38217" > attrs:<> locality:<> ServerVersion:<major_val:0 minor_val:0 patch:0 unstable:0 > 
I171019 08:56:50.378278 2433158 util/stop/stopper.go:476  quiescing; tasks left:
1      [async] storage.multiTestContextKVTransport: calling next replica
I171019 08:56:50.381261 2432518 vendor/google.golang.org/grpc/grpclog/grpclog.go:75  transport: http2Server.HandleStreams failed to read frame: read tcp 127.0.0.1:38217->127.0.0.1:43936: use of closed network connection
I171019 08:56:50.381332 2431728 vendor/google.golang.org/grpc/grpclog/grpclog.go:75  transport: http2Server.HandleStreams failed to read frame: read tcp 127.0.0.1:46730->127.0.0.1:58027: use of closed network connection
I171019 08:56:50.381395 2428764 vendor/google.golang.org/grpc/grpclog/grpclog.go:75  transport: http2Server.HandleStreams failed to read frame: read tcp 127.0.0.1:53470->127.0.0.1:40461: use of closed network connection
	node_liveness_test.go:231: store 2 last uptime = 0; wanted 123
```

tbg:
The node could become live by observing the Gossip update (triggered by its
heartbeat write) before getting to the point in `heartbeatInternal` at which
it would invoke the heartbeat callback. As a result, one location in this
test needs to be prepared to wait for a split second.

Flaky in the hundreds before, now not flaky with 2k+ iterations.

Fixes #19362.

### DIFF
--- a/pkg/storage/client_test.go
+++ b/pkg/storage/client_test.go
@@ -816,21 +816,24 @@ func (m *multiTestContext) addStore(idx int) {
 	}
 	store.WaitForInit()
 
+	ran := struct {
+		sync.Once
+		ch chan struct{}
+	}{
+		ch: make(chan struct{}),
+	}
 	m.nodeLivenesses[idx].StartHeartbeat(ctx, stopper, func(ctx context.Context) {
 		now := m.clock.Now()
 		if err := store.WriteLastUpTimestamp(ctx, now); err != nil {
 			log.Warning(ctx, err)
 		}
+		ran.Do(func() {
+			close(ran.ch)
+		})
 	})
-	// Wait until we see the first heartbeat.
-	testutils.SucceedsSoon(m.t, func() error {
-		if live, err := m.nodeLivenesses[idx].IsLive(nodeID); err != nil {
-			return err
-		} else if !live {
-			return errors.Errorf("node %d not yet live", nodeID)
-		}
-		return nil
-	})
+	// Wait until we see the first heartbeat by waiting for the callback (which
+	// fires *after* the node becomes live).
+	<-ran.ch
 }
 
 func (m *multiTestContext) nodeDesc(nodeID roachpb.NodeID) *roachpb.NodeDescriptor {
--- a/pkg/storage/node_liveness_test.go
+++ b/pkg/storage/node_liveness_test.go
@@ -244,6 +244,9 @@ func TestNodeHeartbeatCallback(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
+	// NB: since the heartbeat callback is invoked synchronously in
+	// `Heartbeat()` which this goroutine invoked, we don't need to wrap this in
+	// a retry.
 	if err := verifyUptimes(); err != nil {
 		t.Fatal(err)
 	}


nvanbenschoten:

---

*[pkg/storage/node_liveness_test.go, line 230 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386#-KwqHUigXwlljRwhwBwl:-KwqHUihCpWRf_PQ7w0w:b-rsdy6n) ([raw file](https://github.com/cockroachdb/cockroach/blob/24d30541aea1d6ec80683a42c80e5f9f2f748746/pkg/storage/node_liveness_test.go#L230)):*
> ```Go
> 	}
> 
> 	// We have to retry here since the node could have become live via its own
> ```

The effect of the node becoming live via its own Gossip update is that `mtc.Start` returns, right? So the race is between `mtc.Start` returning so we can call `verifyUptimes` and the heartbeat callback firing?

If so, can we change `mtc.addStore` to only return when the callback has been called? We have direct access to it since we pass the callback into `StartHeartbeat`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386)*

tschottdorf:

---

*[pkg/storage/node_liveness_test.go, line 230 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386#-KwqHUigXwlljRwhwBwl:-KwqQJ8hau9--YKeI_bC:b-u6is21) ([raw file](https://github.com/cockroachdb/cockroach/blob/24d30541aea1d6ec80683a42c80e5f9f2f748746/pkg/storage/node_liveness_test.go#L230)):*

The effect of the node becoming live via its own Gossip update is that `mtc.Start` returns, right? So the race is between `mtc.Start` returning so we can call `verifyUptimes` and the heartbeat callback firing?

If so, can we change `mtc.addStore` to only return when the callback has been called? We have direct access to it since we pass the callback into `StartHeartbeat`.
</blockquote></details>

Not sure I follow -- the nodes become live because they heartbeat their record in the background (as a running server does). The detail that the callback isn't atomic with the heartbeat is inconsequential outside of this test, and it seems preferable to document it in the test rather than hiding the fact somewhere else. Perhaps I'm missing your point?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386)*

nvanbenschoten:

---

*[pkg/storage/node_liveness_test.go, line 230 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386#-KwqHUigXwlljRwhwBwl:-KwqVE7qOoa-aQrcWBJN:b-i104y8) ([raw file](https://github.com/cockroachdb/cockroach/blob/24d30541aea1d6ec80683a42c80e5f9f2f748746/pkg/storage/node_liveness_test.go#L230)):*

Not sure I follow -- the nodes become live because they heartbeat their record in the background (as a running server does). The detail that the callback isn't atomic with the heartbeat is inconsequential outside of this test, and it seems preferable to document it in the test rather than hiding the fact somewhere else. Perhaps I'm missing your point?
</blockquote></details>

I guess what I'm getting at is that in `mtc.addStore` we currently "Wait until we see the first heartbeat" by polling `m.nodeLivenesses[idx].IsLive(nodeID)`. Once we see this, we begin to race with the callback, which causes this flake. We could instead wait until we see the first call to the heartbeat callback (which we set there) and avoid this race. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386)*

tschottdorf:

---

*[pkg/storage/node_liveness_test.go, line 230 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386#-KwqHUigXwlljRwhwBwl:-KwsoxzZBpi7-izSqGqf:b-mp42hu) ([raw file](https://github.com/cockroachdb/cockroach/blob/24d30541aea1d6ec80683a42c80e5f9f2f748746/pkg/storage/node_liveness_test.go#L230)):*

I guess what I'm getting at is that in `mtc.addStore` we currently "Wait until we see the first heartbeat" by polling `m.nodeLivenesses[idx].IsLive(nodeID)`. Once we see this, we begin to race with the callback, which causes this flake. We could instead wait until we see the first call to the heartbeat callback (which we set there) and avoid this race. 
</blockquote></details>

Ah, I see. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19386)*

tschottdorf:
@nvanbenschoten I'll merge this because it's an annoying flake, but please double-check that I'm doing what you had in mind.

nvanbenschoten:
LGTM, yeah that's exactly what I meant, thanks!

PR #19390, opened 2017-10-19
sql: Fix DROP INDEX IF EXISTS failure with 'idx' syntax

lgo:
Previously, if you don't use the `table@idx` syntax but instead use
`idx` then `DROP INDEX IF EXISTS` behaves like `DROP INDEX`, and it
throws an error if the index doesn't exist.

---
WIP note that the tests are expected to fail due to the unresolved comments below.

### DIFF
--- a/pkg/sql/drop.go
+++ b/pkg/sql/drop.go
@@ -242,8 +242,11 @@ type fullIndexName struct {
 //   Notes: postgres allows only the index owner to DROP an index.
 //          mysql requires the INDEX privilege on the table.
 func (p *planner) DropIndex(ctx context.Context, n *parser.DropIndex) (planNode, error) {
-	idxNames := make([]fullIndexName, len(n.IndexList))
-	for i, index := range n.IndexList {
+	// Keep a track of the indexes that exist to check. When the IF EXISTS
+	// options are provided, we will simply not include any indexes that
+	// don't exist and continue execution.
+	idxNames := make([]fullIndexName, 0, len(n.IndexList))
+	for _, index := range n.IndexList {
 		tn, err := p.expandIndexName(ctx, index, false /* requireTable */)
 		if err != nil {
 			return nil, err
@@ -251,8 +254,8 @@ func (p *planner) DropIndex(ctx context.Context, n *parser.DropIndex) (planNode,
 			// Only index names of the form "idx" throw an error here if they
 			// don't exist.
 			if n.IfExists {
-				// Noop.
-				return &zeroNode{}, nil
+				// Skip this index and don't return an error.
+				continue
 			}
 			// Index does not exist, but we want it to error out.
 			return nil, fmt.Errorf("index %q not found", index.Index)
@@ -267,8 +270,7 @@ func (p *planner) DropIndex(ctx context.Context, n *parser.DropIndex) (planNode,
 			return nil, err
 		}
 
-		idxNames[i].tn = tn
-		idxNames[i].idxName = index.Index
+		idxNames = append(idxNames, fullIndexName{tn: tn, idxName: index.Index})
 	}
 	return &dropIndexNode{n: n, idxNames: idxNames}, nil
 }
--- a/pkg/sql/drop.go
+++ b/pkg/sql/drop.go
@@ -244,9 +244,18 @@ type fullIndexName struct {
 func (p *planner) DropIndex(ctx context.Context, n *parser.DropIndex) (planNode, error) {
 	idxNames := make([]fullIndexName, len(n.IndexList))
 	for i, index := range n.IndexList {
-		tn, err := p.expandIndexName(ctx, index)
+		tn, err := p.expandIndexName(ctx, index, false /* requireTable */)
 		if err != nil {
 			return nil, err
+		} else if tn == nil {
+			// Only index names of the form "idx" throw an error here if they
+			// don't exist.
+			if n.IfExists {
+				// Noop.
+				return &zeroNode{}, nil
+			}
+			// Index does not exist, but we want it to error out.
+			return nil, fmt.Errorf("index %q not found", index.Index)
 		}
 
 		tableDesc, err := MustGetTableDesc(ctx, p.txn, p.getVirtualTabler(), tn, true /*allowAdding*/)
@@ -299,6 +308,8 @@ func (p *planner) dropIndexByName(
 ) error {
 	idx, dropped, err := tableDesc.FindIndexByName(string(idxName))
 	if err != nil {
+		// Only index names of the form "table@idx" throw an error here if they
+		// don't exist.
 		if ifExists {
 			// Noop.
 			return nil
--- a/pkg/sql/logictest/testdata/logic_test/drop_index
+++ b/pkg/sql/logictest/testdata/logic_test/drop_index
@@ -57,8 +57,10 @@ users  foo      false   2    id      ASC        false    true
 users  bar      true    1    id      ASC        false    false
 users  bar      true    2    name    ASC        false    false
 
+# Also test that dropping with a non-existing index still drops 'foo'.
+
 statement ok
-DROP INDEX IF EXISTS users@foo
+DROP INDEX IF EXISTS users@foo, users@zap
 
 query TTBITTBB colnames
 SHOW INDEXES FROM users
@@ -178,8 +180,10 @@ users
 statement ok
 CREATE INDEX baz ON users (name)
 
+# Also test that dropping with a non-existing index still drops 'baz'.
+
 statement ok
-DROP INDEX IF EXISTS baz
+DROP INDEX IF EXISTS baz, zap
 
 query TTBITTBB colnames
 SHOW INDEXES FROM users
@@ -187,7 +191,7 @@ SHOW INDEXES FROM users
 Table  Name     Unique  Seq  Column  Direction  Storing  Implicit
 users  primary  true    1    id      ASC        false    false
 
-# Test that it still succeeds without an index.
+# Test that it still succeeds when an index does not exist.
 
 statement ok
 DROP INDEX IF EXISTS baz
--- a/pkg/sql/logictest/testdata/logic_test/drop_index
+++ b/pkg/sql/logictest/testdata/logic_test/drop_index
@@ -19,6 +19,9 @@ CREATE TABLE othertable (
 statement error index name "baw" is ambiguous
 DROP INDEX baw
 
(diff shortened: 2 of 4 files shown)


lego:

---

*[pkg/sql/drop.go, line 254 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqWwO-Xe5lTe4F4jsm:-KwqWwO-Xe5lTe4F4jsn:bawp6g0) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L254)):*
> ```Go
> 			// Only indexes of the form "idx" throw an error here if they
> 			// don't exist.
> 			if n.IfExists {
> ```

I believe we want to be a bit selective with the here error here.

There's a handful of things that return errors. The error we want to let slide is `"index %q not in any of the tables"`. There's also `"index name %q is ambiguous ..."` which we do not want to allow. What's a good way to handle those different cases? `pgerror.Error`? Do we want to use our own custom code then?

Likewise for the below case, only letting the index does not exist become a non-error noop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390)*

lego:

---

*[pkg/sql/logictest/testdata/logic_test/drop_index, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqXkieLLorbQR6Y16C:-KwqXkieLLorbQR6Y16D:bybmywv) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/logictest/testdata/logic_test/drop_index#L44)):*
> ```
> users  bar      true    2    name    ASC        false    false
> 
> statement error index "zap" does not exist
> ```

This "spurious" change was because I spent some time scratching my head, reading "ffo" as "foo". This is to save the next soul who would make that mistake. :bath:ing those tests, eh?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390)*

knz:
Thanks for catching this.

---

Reviewed 2 of 2 files at r1.

---

*[pkg/sql/drop.go, line 252 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqZULq3FpkeH_ECSqc:-KwqZULrBcBIFtHBJEJV:b8vskbi) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L252)):*
> ```Go
> 		tn, err := p.expandIndexName(ctx, index)
> 		if err != nil {
> 			// Only indexes of the form "idx" throw an error here if they
> ```

"index" != "index name"

here and below.

---

*[pkg/sql/drop.go, line 254 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqWwO-Xe5lTe4F4jsm:-KwqZRzf6-hi7IgNOZxv:b-5vm5ol) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L254)):*

I believe we want to be a bit selective with the here error here.

There's a handful of things that return errors. The error we want to let slide is `"index %q not in any of the tables"`. There's also `"index name %q is ambiguous ..."` which we do not want to allow. What's a good way to handle those different cases? `pgerror.Error`? Do we want to use our own custom code then?

Likewise for the below case, only letting the index does not exist become a non-error noop.
</blockquote></details>

What does pg do?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-:-KwqZYbs1XMuyTLWs7gI:b-x3gtb4)*

lego:

---

*[pkg/sql/drop.go, line 254 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqWwO-Xe5lTe4F4jsm:-KwqaJo4hD6UF2mqojBr:bp3jjya) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L254)):*

What does pg do?
</blockquote></details>

Nothing that we can learn from :P.

```
postgres=# create index name on t1 (name);
CREATE INDEX
postgres=# create index name on t2 (name);
ERROR:  relation "name" already exists
```

I just realized we only check for duplicate index names over within the scope of a table, not the database. Yet we can reference indexes without a table name. Welp.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390)*

lego:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/sql/drop.go, line 252 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqZULq3FpkeH_ECSqc:-KwueY8B3qxfWw2ul-7J:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L252)):*

"index" != "index name"

here and below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390)*

knz:
Reviewed 1 of 1 files at r2.

---

*[pkg/sql/drop.go, line 254 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqWwO-Xe5lTe4F4jsm:-Kx7bIPyQ8ejIXkIBrl-:b6g7y0p) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L254)):*

Nothing that we can learn from :P.

```
postgres=# create index name on t1 (name);
CREATE INDEX
postgres=# create index name on t2 (name);
ERROR:  relation "name" already exists
```

I just realized we only check for duplicate index names over within the scope of a table, not the database. Yet we can reference indexes without a table name. Welp.
</blockquote></details>

Ok I looked at the code closer and I concur something more involved needs to be done.

1) you can extend the prototype of `expandIndexName` and also `findTableContainingIndex`. They both need to take a parameter `requireTable bool`, to gate the error return on "index not found": if false then the function can return a nil TableName to indicate the index was not found, with no error. An error must be returned on ambiguous names in all cases.

2) that said you see that DROP INDEX is applied for multiple indexes. For each of them, `findTableContainingIndex` loops over all table descriptors over again. This is ... inefficient. I would recommend extending `findTableContainingIndex` to work over multiple index names, not just one, then use that in DropIndex directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390)*

lego:
Aha, had a mistake with returning a `zeroNode` if an idex isn't found. `DROP INDEX IF EXISTS ...` will be a no-op if only one index doesn't exist (specifically while not using `tbl@idx` syntax). This is fixed with a second commit.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/drop.go, line 254 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-KwqWwO-Xe5lTe4F4jsm:-Kx8bMafOGedDYMjcSqO:btkae6c) ([raw file](https://github.com/cockroachdb/cockroach/blob/1a0abf9374dadb0bba8147260c907cb2cbefb022/pkg/sql/drop.go#L254)):*

Ok I looked at the code closer and I concur something more involved needs to be done.

1) you can extend the prototype of `expandIndexName` and also `findTableContainingIndex`. They both need to take a parameter `requireTable bool`, to gate the error return on "index not found": if false then the function can return a nil TableName to indicate the index was not found, with no error. An error must be returned on ambiguous names in all cases.

2) that said you see that DROP INDEX is applied for multiple indexes. For each of them, `findTableContainingIndex` loops over all table descriptors over again. This is ... inefficient. I would recommend extending `findTableContainingIndex` to work over multiple index names, not just one, then use that in DropIndex directly.

</blockquote></details>

Done 1.

Now for 2, both `findTableContainingIndex` and `TableDescriptor.FindIndexByName` need to be adjusted for this, and the latter has a handful of call sites. `FindIndexByName` can be modified and named `FindIndexesByName`, and a wrapper func can be made for the single name call.

Do we need to do this? The only time this the call is done with multiple index names is `DROP INDEX`. Right now this call is `O(n^2*m)` (n <- indexes, m <- tables) and the difference afterward is reducing the `n*m` call (`findTableContainingIndex`) to only happen once for each database we're dropping indexes for, as opposed to for each index.

While it wouldn't be difficult to change `findTableContainingIndex` to handle an array of indexes, we run into two problems:
- We want to skip indexes with an explicit table name.
- We have to group each index name without explicit table names by their database name and do a separate `findTableContainingIndex` for each one.

So, I'm not so gung-ho on replacing this because of the added complexity. Feel free to make the judgment call for if we should do it still, as I don't think I know where the line is drawn here for complexity vs. improvement, just pointing out these things before running into it after :).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-:-Kx8oSWv1QDXn67bQBvj:bgaipx3)*

lego:
bump @knz

knz:
:lgtm: but please consider the comment below

---

Reviewed 2 of 3 files at r3, 2 of 2 files at r4.

---

*[pkg/sql/drop.go, line 249 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-Kxm5rMmADWeQWjPvJ20:-Kxm5rMnlqVunLZy-NBI:b-8qpmkf) ([raw file](https://github.com/cockroachdb/cockroach/blob/eae429b81a060635cbc3f069a5d936b44f7d6dc2/pkg/sql/drop.go#L249)):*
> ```Go
> 	// don't exist and continue execution.
> 	idxCount := 0
> 	idxNames := make([]fullIndexName, len(n.IndexList))
> ```

Easier on the eye: `idxNames := make([]fullIndexName, 0, len(n.IndexList))`

then use `append` below and drop `idxCount`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19390#-:-Kxm62Idlmr43ZbSafGO:b-la83q6)*

lego:
TFTR!

PR #19392, opened 2017-10-19
acceptance/cluster: don't run containers as root

benesch:
Root containers leak root files and directories everywhere, which brick
TeamCity agents unless one is very, very careful. Run the containers as
the invoking user instead.

This requires replacing all volumes with bind mounts, as volumes are
always root-owned and thus unwriteable by the new container user. Bind
mounts, by contrast, inherit the permissions of the file or directory on
the host. (Under the hood, volumes are just automatically-managed bind
mounts to somewhere in /var/lib/docker/, so there's no performance cost
to this.)

### DIFF
--- a/pkg/acceptance/cluster/docker.go
+++ b/pkg/acceptance/cluster/docker.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"os/user"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -151,6 +152,13 @@ func pullImage(
 	return nil
 }
 
+// splitBindSpec splits a Docker bind specification into its host and container
+// paths.
+func splitBindSpec(bind string) (hostPath string, containerPath string) {
+	s := strings.SplitN(bind, ":", 2)
+	return s[0], s[1]
+}
+
 // createContainer creates a new container using the specified
 // options. Per the docker API, the created container is not running
 // and must be started explicitly. Note that the passed-in hostConfig
@@ -167,6 +175,27 @@ func createContainer(
 	// Disable DNS search under the host machine's domain. This can
 	// catch upstream wildcard DNS matching and result in odd behavior.
 	hostConfig.DNSSearch = []string{"."}
+
+	// Run the container as the current user to avoid creating root-owned files
+	// and directories from within the container.
+	user, err := user.Current()
+	if err != nil {
+		maybePanic(err)
+	}
+	containerConfig.User = fmt.Sprintf("%s:%s", user.Uid, user.Gid)
+
+	// Additionally ensure that the host side of every bind exists. Otherwise, the
+	// Docker daemon will create the host directory as root before running the
+	// container.
+	for _, bind := range hostConfig.Binds {
+		hostPath, _ := splitBindSpec(bind)
+		if _, err := os.Stat(hostPath); os.IsNotExist(err) {
+			maybePanic(os.MkdirAll(hostPath, 0755))
+		} else {
+			maybePanic(err)
+		}
+	}
+
 	resp, err := l.client.ContainerCreate(ctx, &containerConfig, &hostConfig, nil, containerName)
 	if err != nil {
 		return nil, err
@@ -262,7 +291,7 @@ func (c *Container) Wait(ctx context.Context, condition container.WaitCondition)
 	case err := <-errCh:
 		return err
 	case waitOKBody := <-waitOKBodyCh:
-		outputLog := filepath.Join(c.cluster.logDir, "console-output.log")
+		outputLog := filepath.Join(c.cluster.volumesDir, "logs", "console-output.log")
 		cmdLog, err := os.Create(outputLog)
 		if err != nil {
 			return err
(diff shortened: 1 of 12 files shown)


pkg/acceptance/cluster/docker.go
```diff
@@ -262,7 +291,7 @@ func (c *Container) Wait(ctx context.Context, condition container.WaitCondition)
 	case err := <-errCh:
 		return err
 	case waitOKBody := <-waitOKBodyCh:
-		outputLog := filepath.Join(c.cluster.logDir, "console-output.log")
+		outputLog := filepath.Join(c.cluster.volumesDir, "logs", "console-output.log")
```
tschottdorf: Does this mean we'll henceforth also put the data directories into the artifacts? Maybe that's ok, but explain it in the commit message.

pkg/acceptance/cluster/dockercluster.go
```diff
@@ -177,32 +177,30 @@ func CreateDocker(
...
-	// logs.
-	var uniqueLogDir string
-	logDirRemovable := false
-	if logDir != "" {
+
+	if volumeDir == "" {
+		volumeDir, err = ioutil.TempDir("", fmt.Sprintf("cockroach-acceptance-%s", clusterIDS))
+		maybePanic(err)
+	} else {
+		volumeDir = fmt.Sprintf("%s-%s", volumeDir, clusterIDS)
```
tschottdorf: So you don't actually use the `volumeDir` that was supplied? Perhaps just use `filepath.Join(volumeDir, clusterIDS)`.

pkg/acceptance/cluster/dockercluster.go
```diff
@@ -177,32 +177,30 @@ func CreateDocker(
...
 		config:    cfg,
 		stopper:   stopper,
 		// TODO(tschottdorf): deadlocks will occur if these channels fill up.
-		events:          make(chan Event, 1000),
-		expectedEvents:  make(chan Event, 1000),
-		logDir:          uniqueLogDir,
-		logDirRemovable: logDirRemovable,
+		events:         make(chan Event, 1000),
+		expectedEvents: make(chan Event, 1000),
+		volumesDir:     volumeDir,
```
tschottdorf: Feels like `volumeDir` should be called `volumesDir`.

tschottdorf: (or vice versa)

pkg/acceptance/cluster/dockercluster.go
```diff
@@ -916,23 +896,19 @@ func (l *DockerCluster) ExecCLI(ctx context.Context, i int, cmd []string) (strin
...
+	volumes, err := ioutil.ReadDir(l.volumesDir)
+	if err != nil {
+		log.Warning(ctx, err)
+		return
+	}
+	for _, v := range volumes {
+		if preserveLogs && v.Name() == "logs" {
+			continue
+		}
+		if err := os.RemoveAll(filepath.Join(l.volumesDir, v.Name())); err != nil {
```
tschottdorf: Makes me nervous, because I tend to run with `-l .` which I think might translate to `volumesDir = $GOPATH/src/github.com/cockroachdb/cockroach/acceptance` and you'd just maul my checkout.

tschottdorf: Guess in practice it's fine because `volumesDir` is either a tempdir or my originally specified directory with the clusterID joined, right? Perhaps add a comment on `volumesDir`.

tbg (approved):
The previous code has had a lot of iterations to work out random kinks, but not running as root is great and overall you've also simplified things a bit. Thanks!

pkg/acceptance/cluster/dockercluster.go
tschottdorf: Mind changing it and if something goes wrong, you get to point the finger at me 👈 ?

pkg/acceptance/cluster/dockercluster.go
tschottdorf: Is this really done?

tbg (approved):

benesch:
Whoops, sorry, didn't push my changes because there's one last hurdle to making this work properly. Specifically, a bunch of tests fail when not run as root because they expect a valid /etc/passwd entry for the user. I have a not horrible solution to this problem in general (https://github.com/benesch/autouseradd), but I have a side quest PR I'm going to send your way first.

benesch:
Side quest: 19915

pkg/acceptance/cluster/dockercluster.go
benesch: Works for me. Done!

benesch:
Whew, it's finally green! @tschottdorf want to take one last look at the `autouseradd` stuff I added?

tschottdorf:
:lgtm:!

---

Reviewed 1 of 3 files at r1, 11 of 11 files at r2.

---

*[pkg/acceptance/util_docker.go, line 41 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19392#-KyWqScZO-QyxrpslZh5:-KyWqSc_D7Wau0fWmNrh:b74ws31) ([raw file](https://github.com/cockroachdb/cockroach/blob/dd4259b84547ecfc3e1780e66f7c8b3a5e1f858a/pkg/acceptance/util_docker.go#L41)):*
> ```Go
> 			"PGSSLKEY=/certs/node.key",
> 		},
> 		Entrypoint: []string{"autouseradd", "-u", "roach", "-C", "/home/roach", "--"},
> ```

:heartbeat: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19392#-:-KyWqdHpqUNoklNFob9I:bg7igkw)*

benesch:
Woohoo! TFTRs!

PR #19395, opened 2017-10-19
build: move {metacheck,unconvert} into style_test

tamird:
Avoids shelling out; runs everything in the test process and avoids `unconvert`'s extra parsing of the AST.


pkg/cli/interactive_tests/test_missing_log_output.tcl
```diff
@@ -76,7 +76,7 @@ start_test "Test that quit does not emit unwanted logging output"
 # Unwanted: between the point the command starts until it
 # either prints the final ok message or fails with some error
 # (e.g. due to no definite answer from the server).
-send "echo marker; $argv quit 2>&1 | grep '^\\(\[IWEF\]\[0-9\]\\)' \r"
+send "echo marker; $argv quit 2>&1 | grep -vE '^\[IWEF\]\[0-9\]+ grpc: '\r"
```
tschottdorf: Not good enough, I think! We need to match

> W171020 12:24:14.884065 21 vendor/google.golang.org/grpc/clientconn.go:1034  grpc: addrConn.resetTransport failed to create client transport:

So I'm missing a `.*` somewhere.

tamird: Good point! Fixed in #19402.

tamird:
In my testing, this reduces `make lint` from 86s to 56s. (CI failed on an acceptance timeout that seems unrelated to this PR). `make lintshort` is unaffected.

tamird:
Ah, looks like @a-robinson kindly rebuilt this.

tschottdorf:
This looks pretty good to me and the speedup is nice. It could use a few more inline comments (i.e. what's the megachecker struct, etc) but :lgtm:  and thanks!

---

Reviewed 1 of 1 files at r1, 28 of 28 files at r6, 3 of 3 files at r7, 1 of 1 files at r8, 26 of 26 files at r9, 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19395#-:-KwvTbUOeM-bQx8SpL5B:bpx9uha)*

tamird:
Happy to add whatever comments you'd like to see; added one for the struct, but lmk if you want to see others - I'll do that in a follow-up.

Gonna need a rebuild to get past these new flakes though!

tamird:
Thank you, kind stranger!

PR #19396, opened 2017-10-19
sql: introduce weak keys

### LINKED ISSUE #19343: sql: incorrect key property for unique indexes with nullable columns
The current code assumes that a UNIQUE index implies distinct values. But that's not quite true because of NULL columns - a unique index operates under the assumption that NULL != NULL so it allows multiple NULL values.

One place where we use the key information is to elide DISTINCT:

```sql
create table kv (k INT PRIMARY KEY, v INT, UNIQUE INDEX idx(v));

root@:26257/test> select distinct * from (select v from kv);
+------+
|  v   |
+------+
|    1 |
|    2 |
| NULL |
+------+
(3 rows)

root@:26257/test> select distinct * from (select v from kv@idx);
+------+
|  v   |
+------+
| NULL |
| NULL |
| NULL |
|    1 |
|    2 |
+------+
(5 rows)
```

We incorrectly elide the distinctNode:

```
root@:26257/test> explain select distinct * from (select v from kv);
+-------+----------+-------+-------------+
| Level |   Type   | Field | Description |
+-------+----------+-------+-------------+
|     0 | distinct |       |             |
|     1 | render   |       |             |
|     2 | scan     |       |             |
|     2 |          | table | kv@primary  |
|     2 |          | spans | ALL         |
+-------+----------+-------+-------------+

root@:26257/test> explain select distinct * from (select v from kv@idx);
+-------+--------+-------+-------------+
| Level |  Type  | Field | Description |
+-------+--------+-------+-------------+
|     0 | render |       |             |
|     1 | scan   |       |             |
|     1 |        | table | kv@idx      |
|     1 |        | spans | ALL         |
+-------+--------+-------+-------------+
```

RaduBerinde:
The current `physicalProps` has a problem - what we consider keys may not actually
be keys if we have NULLs on any of the respective columns. This is because e.g.
even a unique index allows multiple NULL values. See #19343 for an example.

To fix this we change keys to be "weak keys"; a weak key only means that it
*could* be a key, if we have no NULLs on the respective columns. In order for
weak keys to be useful, we also introduce the set of "not null" columns as a
property. This allows us to easily know if a weak key is actually a key.

We can do better on detecting not-null columns, e.g. for filters; this will be
done in a follow-up change.

Fixes #19343.


petermattis:
Review status: 0 of 30 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/sql/physical_props.go, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-Kwr9xLjjeFpg_-Sp3Ju:-Kwr9xLjjeFpg_-Sp3Jv:b-ht08bb) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/physical_props.go#L65)):*
> ```Go
> // A set of columns S forms a "weak key" if, after projection onto S, no two
> // rows are equal and no rows contain NULL values. A UNIQUE index on a table is
> // a candidate key and possibly a key if all of the columns are NOT NULL. A weak
> ```

`s/candidate/weak/g`, though some Googling suggests that `candidate key` is actually the more accurate term: https://stackoverflow.com/questions/34303878/what-is-the-candidate-key-in-sql-server

The wikipedia definition is quite a bit more confusing: https://en.wikipedia.org/wiki/Candidate_key

I'm not sure where I got the term "weak key". I sort of prefer it, but I think it would be better to stick with standard(-ish) terminology.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396)*

RaduBerinde:
*[pkg/sql/physical_props.go, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-Kwr9xLjjeFpg_-Sp3Ju:-KwrDkiSjw-ycb5G6iev:b2hyvn1) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/physical_props.go#L65)):*

`s/candidate/weak/g`, though some Googling suggests that `candidate key` is actually the more accurate term: https://stackoverflow.com/questions/34303878/what-is-the-candidate-key-in-sql-server

The wikipedia definition is quite a bit more confusing: https://en.wikipedia.org/wiki/Candidate_key

I'm not sure where I got the term "weak key". I sort of prefer it, but I think it would be better to stick with standard(-ish) terminology.
</blockquote></details>

Hmm. The definition on wikipedia is saying that "candidate key" is a key which is minimal (not a superset of another key). This is the case here, so are these.. "candidate weak keys"? haha

After googling some more, I found other places where "candidate key" is defined as a key (not a weak key). E.g. http://www.crazyforcode.com/keys-rdbms/ http://www.java67.com/2012/10/difference-between-primary-key-vs-candidate-keys-sql-database.html

At this point, I'd rather go with "weak" to avoid confusion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396)*

knz:
:lgtm:

Is 1.1 affected?

---

Reviewed 30 of 30 files at r1.

---

*[pkg/sql/physical_props.go, line 120 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-KwtoFzOCjWXoJg6Ynfp:-KwtoFzOCjWXoJg6Ynfq:b-xgs4r7) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/physical_props.go#L120)):*
> ```Go
> 	// columns for which we know we have only non-NULL values. For groups of
> 	// equivalent columns, only the group representative can be in this set.
> 	// This set contains all the constant columns.
> ```

A column constrained to be only NULL is constant, but you probably don't want it here.

---

*[pkg/sql/plan_phyical_props.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-Kwtp1SwjlWrolkkLjrT:-Kwtp1SxZhRNknQm-jIH:bc1f2fc) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/plan_phyical_props.go#L1)):*
> ```Go
> // Copyright 2017 The Cockroach Authors.
> ```

I haven't seen it before, there's a typo in the file name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-:-KwtpjEzLHjP8c5WoDBX:bb0t2l6)*

knz:
Regarding the constant column: right now the code only places columns in the constant set as a result of `=` comparisons, so that necessarily excludes all-null columns. However the explanatory comment should express this clearly.

RaduBerinde:
TFTR! No, this key stuff was added recently. We did have a limited version where an ordering could be marked as "key", so there may be a subtle issue with orderings and NULL values.

---

Review status: 28 of 30 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/physical_props.go, line 120 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-KwtoFzOCjWXoJg6Ynfp:-KwtrQa2k0rszT3gHHkF:b6r17t6) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/physical_props.go#L120)):*

A column constrained to be only NULL is constant, but you probably don't want it here.
</blockquote></details>

Hm, good catch. Clarified that constant columns cannot be NULL.

---

*[pkg/sql/plan_phyical_props.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-Kwtp1SwjlWrolkkLjrT:-KwtriSB0HEpBjq0bjSg:b-zaav8m) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/plan_phyical_props.go#L1)):*

I haven't seen it before, there's a typo in the file name.
</blockquote></details>

haha good catch. Renamed (in a separate commit).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-:-KwtrnwmtnctGxvF2BtU:bq56uii)*

justinj:
Review status: 28 of 30 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/physical_props.go, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-KwtwOSy5Ujb4yzyYnrh:-KwtwOSy5Ujb4yzyYnri:bpkypor) ([raw file](https://github.com/cockroachdb/cockroach/blob/89fc7e8f54d52b816b41267a212ee5fc199cc020/pkg/sql/physical_props.go#L61)):*
> ```Go
> //
> // A set of columns S forms a "key" if no two rows are equal when projected on
> // S. A requirement for a column set to be a key is for no columns in the set to
> ```

Feel free to ignore me, but I initially found this description a little confusing, I first read it as
```
Definition: A set of columns S forms a "key" if no two rows
are equal when projected on S. 
If S is a key, then no column in S is NULL-able.
```
which contradicts the statement that comes after, and I think it's meant to be read as
```
A set of columns S forms a "key" if
1. no two rows are equal when projected on S and
2. no column in S is NULL-able.
```
is that correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396)*

RaduBerinde:
Review status: 28 of 30 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/physical_props.go, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-KwtwOSy5Ujb4yzyYnrh:-Kwu-VwMtzKBikhQXazD:b261rfv) ([raw file](https://github.com/cockroachdb/cockroach/blob/89fc7e8f54d52b816b41267a212ee5fc199cc020/pkg/sql/physical_props.go#L61)):*

Feel free to ignore me, but I initially found this description a little confusing, I first read it as
```
Definition: A set of columns S forms a "key" if no two rows
are equal when projected on S. 
If S is a key, then no column in S is NULL-able.
```
which contradicts the statement that comes after, and I think it's meant to be read as
```
A set of columns S forms a "key" if
1. no two rows are equal when projected on S and
2. no column in S is NULL-able.
```
is that correct?
</blockquote></details>

You are correct. The phrasing was indeed confusing and there were some mistakes. Fixed, PTAL

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396)*

justinj:
Reviewed 1 of 1 files at r2.
Review status: 28 of 30 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/physical_props.go, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-KwtwOSy5Ujb4yzyYnrh:-Kwu1AP3bhmM7Dk81-uj:b-6yc0nv) ([raw file](https://github.com/cockroachdb/cockroach/blob/89fc7e8f54d52b816b41267a212ee5fc199cc020/pkg/sql/physical_props.go#L61)):*

You are correct. The phrasing was indeed confusing and there were some mistakes. Fixed, PTAL
</blockquote></details>

:thumbsup: thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396)*

petermattis:
Review status: 28 of 30 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/physical_props.go, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396#-Kwr9xLjjeFpg_-Sp3Ju:-Kwu3Xf7BS1tssEtB2eF:bn1wz0t) ([raw file](https://github.com/cockroachdb/cockroach/blob/4c4ee9b6bbd48982f55429c5d1de03e4b9de25e2/pkg/sql/physical_props.go#L65)):*

Hmm. The definition on wikipedia is saying that "candidate key" is a key which is minimal (not a superset of another key). This is the case here, so are these.. "candidate weak keys"? haha

After googling some more, I found other places where "candidate key" is defined as a key (not a weak key). E.g. http://www.crazyforcode.com/keys-rdbms/ http://www.java67.com/2012/10/difference-between-primary-key-vs-candidate-keys-sql-database.html

At this point, I'd rather go with "weak" to avoid confusion.
</blockquote></details>

:thumbsup:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19396)*

RaduBerinde:
> We did have a limited version where an ordering could be marked as "key", so there may be a subtle issue with orderings and NULL values.

Indeed there is a bug here in 1.1 as well, I sent out a PR for 1.1.

PR #19400, opened 2017-10-20
cherrypick-1.1: cli: simplify and improve dump performance

### LINKED ISSUE #18500: sql+cli: dump should work with PKs named other than "primary"

madelynnblue:
Cherrypick #18472 and #19325.

#18472 was initially not included in 1.1 because it was thought to only be a performance increase, and the release cycle was late enough that it wasn't worth the risk to include it. However #18500 showed that the 1.1 branch had a known bug, so now it may be important enough to include here.

We could have alternatively made a from-scratch change to just fix the bug in 1.1, but I think it's better to cherrypick a more thought out and tested solution instead.

Our other option is to not accept the risk of the code changes here, and document the bug as a known limitation. However I think the change is safe enough to include, and comes with a hefty performance improvement to boot.

Fixes #18500

### DIFF
--- a/pkg/cli/dump.go
+++ b/pkg/cli/dump.go
@@ -24,8 +24,10 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
+
 	"github.com/cockroachdb/cockroach/pkg/sql/parser"
-	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19400#-:-KwuqH4oeKHtdF8_XkJd:bnfp4nl)*

PR #19405, opened 2017-10-20
sql: Add server_version_num session variable

jimlar:
I've come across an ORM that is using this session vaiable to detect
postgres version. Seems ok to support this since server_version
is already present. This can make life easier when porting an app from
postgres to cockroachdb.

Not sure about the type of the var being string type, but I can see
other vars that are integers in postgres that are represented as strings
here (max_index_keys for example).

The specific ORM I've come across is "dat" for Go: https://github.com/mgutz/dat

My guess is that other ORMs/frameworks would use this variable too to check
postgres version > some version.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/explain
+++ b/pkg/sql/logictest/testdata/logic_test/explain
@@ -145,7 +145,7 @@ EXPLAIN SHOW DATABASE
 0  render  ·     ·
 1  filter  ·     ·
 2  values  ·     ·
-2  ·       size  2 columns, 20 rows
+2  ·       size  2 columns, 21 rows
 
 query ITTT
 EXPLAIN SHOW TIME ZONE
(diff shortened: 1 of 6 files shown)


knz:
Hello! Thank you for your contribution!
The code change looks good.

Can you tell us more which ORM motivates this change? It would be best if you could spell it out in the commit message, so that a person who reads the history of the code can get the context for the change.

Thank you again!

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19405#-:-KwtqsI60HVJebAg94Xr:br5cpyt)*

knz:
Thank you!

:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19405#-:-KwtwASbR9fL3JnNhyxY:b3oh6o9)*

PR #19415, opened 2017-10-20
sql: fix repeated escapes in array parsing

justinj:
Previously, we wouldn't parse properly arrays containing strings with
multiple escapes in a row. This commit fixes that.

I continue to have issues with this code, unfortunately. I've lost
confidence that what I'm doing here with these test cases is an
appropriate way of testing this kind of thing (at least as far as
getting coverage goes), but I'm at a loss for a better way that won't be
really heavyweight. If we were to generate strings, how do we do it in a
way that is specific to the kind of escaping this function deals with
(which is non-standard) but would also unearth new problems?

Going to cherry-pick for 1.1.2.

### DIFF
--- a/pkg/sql/parser/parse_array.go
+++ b/pkg/sql/parser/parse_array.go
@@ -47,9 +47,8 @@ var isElementChar = func(r rune) bool {
 func (p *parseState) gobbleString(isTerminatingChar func(ch byte) bool) (out string, err error) {
 	var result bytes.Buffer
 	start := 0
-	i := -1
-	for i < len(p.s) && (i < 0 || !isTerminatingChar(p.s[i])) {
-		i++
+	i := 0
+	for i < len(p.s) && !isTerminatingChar(p.s[i]) {
 		// In these strings, we just encode directly the character following a
 		// '\', even if it would normally be an escape sequence.
 		if i < len(p.s) && p.s[i] == '\\' {
@@ -60,6 +59,8 @@ func (p *parseState) gobbleString(isTerminatingChar func(ch byte) bool) (out str
 				i++
 			}
 			start = i
+		} else {
+			i++
 		}
 	}
 	if i >= len(p.s) {
--- a/pkg/sql/parser/parse_array_test.go
+++ b/pkg/sql/parser/parse_array_test.go
@@ -14,7 +14,11 @@
 
 package parser
 
-import "testing"
+import (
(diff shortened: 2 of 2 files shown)


knz:
Do we have (can you make) a pretty-printer which formats these strings like the parser wants them?
If you have such a thing, then you can generate random sequences of bytes (or printable characters with a configurable % of non-printable characters), then check that they round trip via the pretty-printer and your parser here.

---

Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/parser/parse_array_test.go, line 51 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19415#-Kx7R0tkoMF13uWrBidy:-Kx7R0tkoMF13uWrBidz:bkxm78w) ([raw file](https://github.com/cockroachdb/cockroach/blob/1e6b77a2ad1fd750129cf11e1b918557855b2823/pkg/sql/parser/parse_array_test.go#L51)):*
> ```Go
> 		{`{" hello "}`, stringColTypeString, Datums{NewDString(" hello ")}},
> 		{`{"hel,lo"}`, stringColTypeString, Datums{NewDString("hel,lo")}},
> 		{`{"hel\"lo"}`, stringColTypeString, Datums{NewDString("hel\"lo")}},
> ```

To clarify what is happening in all these tests, please enclose the string on the right also with the back ticks (to drop one level of escaping).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19415#-:-Kx7RK7Rplo2Q5g9v9YF:b-6299sn)*

knz:
Reviewed 1 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19415)*

justinj:
OK, I tried it and it worked really well (I needed to write a special printer because arrays are odd) - it catches all the problems that we had before and it doesn't see any with the current implementation. I'll do this a lot more for future parsers.

knz:
:lgtm: well done!

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19415#-:-Kx93xfqtoyd92SRt4PN:b5nrhdo)*

PR #19416, opened 2017-10-20
sql: Add grammar for referential integrity constraints

BramGruneir:
Add in grammars and actions for:
* ON DELETE NO ACTION
* ON UPDATE NO ACTION
* ON DELETE RESTRICT
* ON UPDATE RESTRICT

One of the key points in the PR is that there is now a difference between RESTRICT and NO ACTION. NO ACTION is the default action and will not print on a SHOW CREATE TABLE. RESTRICT however will show up. However, both at this point are functionally the same. This is designed to set up the ability to add in DEFERRED at a later date.

This PR unlocks work on cascading referential integrity constraints.


knz:

---

*[pkg/sql/create.go, line 889 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7UPPPiS3uyxD49C0X:-Kx7UPPQr1B73dKeXxiA:b-2u3lqr) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/create.go#L889)):*
> ```Go
> 		SharedPrefixLen: int32(len(srcCols)),
> 		OnDelete:        sqlbase.ForeignKeyReference_Action(d.Actions.Delete),
> 		OnUpdate:        sqlbase.ForeignKeyReference_Action(d.Actions.Update),
> ```

Add a check below, so that if the user enters something else than NoAction for either of those, an "unsupported" error is produced.

We don't want users of alpha releases to get the wrong impression until the feature is completed.

---

*[pkg/sql/parser/create.go, line 547 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7SnG9EwoGnjLBCagx:-Kx7SnG9EwoGnjLBCagy:b5fscxe) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/create.go#L547)):*
> ```Go
> )
> 
> var referenceAction = [...]string{
> ```

s/referenceAction/referenceActionName/

---

*[pkg/sql/parser/sql.y, line 364 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7TCxy5lxybpJHhWN2:-Kx7TCxy5lxybpJHhWN3:bte69m9) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L364)):*
> ```Yacc
>     return u.val.(ReferenceAction)
> }
> func (u *sqlSymUnion) referenceActions() *ReferenceActions {
> ```

No need for the pointer type here.

---

*[pkg/sql/parser/sql.y, line 801 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7TGLEUZbNmi6ZBUtn:-Kx7TGLEUZbNmi6ZBUto:bcj35ou) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L801)):*
> ```Yacc
> %type <ColumnQualification> col_qualification_elem
> %type <empty> key_match
> %type <*ReferenceActions> reference_actions
> ```

`%type <ReferenceActions>` is OK here (it's a small type).

---

*[pkg/sql/parser/sql.y, line 2565 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7UFp2aTAEfB36xfCU:-Kx7UFp2aTAEfB36xfCV:bkg29nh) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L2565)):*
> ```Yacc
> // Table constraints:
> //    PRIMARY KEY ( <colnames...> )
> //    FOREIGN KEY ( <colnames...> ) REFERENCES <tablename> [( <colnames...> )]
> ```

This needs to be extended.

---

*[pkg/sql/parser/sql.y, line 2572 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7UHWfIA75laIw607C:-Kx7UHWgPzHQ-sxGGlD9:b8eka3b) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L2572)):*
> ```Yacc
> //   [CONSTRAINT <constraintname>] {NULL | NOT NULL | UNIQUE | PRIMARY KEY | CHECK (<expr>) | DEFAULT <expr>}
> //   FAMILY <familyname>, CREATE [IF NOT EXISTS] FAMILY [<familyname>]
> //   REFERENCES <tablename> [( <colnames...> )]
> ```

This too.

---

*[pkg/sql/sqlbase/structured.proto, line 96 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7TVKBtLcVX4ffGG_4:-Kx7TVKBtLcVX4ffGG_5:b-dsa3in) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/sqlbase/structured.proto#L96)):*
> ```protobuf
> message ForeignKeyReference {
>   enum Action {
>     // Note that this enum must exactly match sql/parser/create.go's
> ```

"magic matching" by means of comments and human willingness is dangerous. We haven't required this before.

It wouldn't hurt too much (IMHO) to have a `[...]int{ parser.NoAction: ForeignKeyReference_Action_NO_ACTION, ...}` around here and use it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416)*

BramGruneir:
Review status: 4 of 15 files reviewed at latest revision, 7 unresolved discussions.

---

*[pkg/sql/create.go, line 889 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7UPPPiS3uyxD49C0X:-Kx8mz0HeIW3bQ1Wwq_q:bccrvqv) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/create.go#L889)):*

Add a check below, so that if the user enters something else than NoAction for either of those, an "unsupported" error is produced.

We don't want users of alpha releases to get the wrong impression until the feature is completed.
</blockquote></details>

Sure.  
Both NoAction and Restrict are now allowed.

---

*[pkg/sql/parser/create.go, line 547 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7SnG9EwoGnjLBCagx:-Kx8qmN7PJJ4nQn4esDX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/create.go#L547)):*

s/referenceAction/referenceActionName/
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 364 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7TCxy5lxybpJHhWN2:-Kx8spgCiC_ijNbGHjUJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L364)):*

No need for the pointer type here.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 801 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7TGLEUZbNmi6ZBUtn:-Kx8suHpJomWRWlFRvPV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L801)):*

`%type <ReferenceActions>` is OK here (it's a small type).
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 2565 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7UFp2aTAEfB36xfCU:-Kx8tRC3IhcBgOwLDDGX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L2565)):*

This needs to be extended.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 2572 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7UHWfIA75laIw607C:-Kx8tWV6FZ2DDLx0j005:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/parser/sql.y#L2572)):*

This too.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.proto, line 96 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx7TVKBtLcVX4ffGG_4:-Kx96rsF8ctpZR9MqbDF:blypl4y) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/sqlbase/structured.proto#L96)):*

"magic matching" by means of comments and human willingness is dangerous. We haven't required this before.

It wouldn't hurt too much (IMHO) to have a `[...]int{ parser.NoAction: ForeignKeyReference_Action_NO_ACTION, ...}` around here and use it.
</blockquote></details>

I was going to write a parser.  This is much better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416)*

knz:
Very nice! You got the hang of this sql thing on the first try :)

---

Reviewed 11 of 11 files at r2.

---

*[pkg/sql/create.go, line 886 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9hCqzzqAvGMtcRaVU:-Kx9hCqzzqAvGMtcRaVV:bty5oef) ([raw file](https://github.com/cockroachdb/cockroach/blob/871118672e097eaeca62cf927a7f4e04d051249d/pkg/sql/create.go#L886)):*
> ```Go
> 		d.Actions.Delete != parser.Restrict {
> 		feature := fmt.Sprintf("ON DELETE %s", d.Actions.Delete)
> 		return pgerror.Unimplemented(feature, "unsupported "+feature)
> ```

1) I would recommend you add a test for this check, as in the comment below.

2) place the "unsupported " prefix inside the `Sprintf` call to save one string concatenation.

---

*[pkg/sql/logictest/testdata/logic_test/fk, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9hidLTziBNjZuV6BX:-Kx9hidLTziBNjZuV6BY:bzads8g) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/logictest/testdata/logic_test/fk#L69)):*
> ```
> 
> statement error pq: columns cannot be used by multiple foreign key constraints
> ALTER TABLE orders ADD FOREIGN KEY (product) REFERENCES products ON UPDATE NO ACTION
> ```

Perhaps add a new test here to check the other actions are still unsupported.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-Kx9hp7-sQkcRGUj7EDe:b6oyibg)*

BramGruneir:
Review status: 14 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/create.go, line 886 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9hCqzzqAvGMtcRaVU:-Kx9t2gIIKH8w4uyrxy0:b-v0kzpi) ([raw file](https://github.com/cockroachdb/cockroach/blob/871118672e097eaeca62cf927a7f4e04d051249d/pkg/sql/create.go#L886)):*

1) I would recommend you add a test for this check, as in the comment below.

2) place the "unsupported " prefix inside the `Sprintf` call to save one string concatenation.
</blockquote></details>

1 - So this final check is not really needed, since the grammar right now will reject the statements and there are already tests for all cases.  But I could remove that from the sql.y and make this be the main check.  What are you thoughts?

2 - done.  

---

*[pkg/sql/logictest/testdata/logic_test/fk, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9hidLTziBNjZuV6BX:-Kx9sijQrExNuz8Lw73D:b-pak02l) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/logictest/testdata/logic_test/fk#L69)):*

Perhaps add a new test here to check the other actions are still unsupported.
</blockquote></details>

There already are below.  What other type of check would you like me to add?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416)*

jordanlewis:
Are there any migration concerns in this PR?

The default used to be `RESTRICT`. Now it's `NO ACTION`. I know that these two actions are the same. Will previously-created descriptors with unspecified actions get properly handed by the new code?

---

Reviewed 4 of 13 files at r1, 10 of 11 files at r2, 1 of 1 files at r3.

---

*[pkg/sql/parser/sql.y, line 3009 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9vrdcMHMUn-Cke-0y:-Kx9vrdcMHMUn-Cke-0z:b6vbxb4) ([raw file](https://github.com/cockroachdb/cockroach/blob/f9189965d2b8d9da9f2a4e79ff44980dd582c5f2/pkg/sql/parser/sql.y#L3009)):*
> ```Yacc
>   }
> // RESTRICT is currently the default and only supported ON DELETE/UPDATE
> // behavior and thus needs no special handling.
> ```

This comment seems stale - RESTRICT is no longer the default, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-Kx9yNRNFBjrKOYplOP9:bp52mqr)*

jordanlewis:

---

*[pkg/sql/parser/sql.y, line 3714 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9zKj65Jy-Hg9F9Oov:-Kx9zKj65Jy-Hg9F9Oow:bj04fuw) ([raw file](https://github.com/cockroachdb/cockroach/blob/f9189965d2b8d9da9f2a4e79ff44980dd582c5f2/pkg/sql/parser/sql.y#L3714)):*
> ```Yacc
> // creating shift/reduce conflicts.
> //
> //      The locking clause (FOR UPDATE etc) may be be` or after
> ```

Stray comment change I think

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416)*

knz:
Interesting comment by Jordan. Perhaps "No action" is just a syntax thing -- to be desugared to RESTRICT automatically in sql/create.go, and not present at all in the sqlbase code (and the proto code)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-Kx9zaLBDpCxKTb7WcMH:bnuyrec)*

knz:
Review status: 13 of 15 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/logictest/testdata/logic_test/fk, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9hidLTziBNjZuV6BX:-KxAIZY0ew0adaovToQe:b-so0kxa) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/logictest/testdata/logic_test/fk#L69)):*

There already are below.  What other type of check would you like me to add?
</blockquote></details>

Ok now I know what was confusing me.

With your patch we now have two places where we check for unsupported syntax: both in the parser (rejects CASCADE etc) and create.go (rejects currently nothing: the parser only lets NO ACTION and RESTRICT through). 

What I had in mind is to let _all_ the syntax through (with suitable tests in parse_test.go), and have *all* the supported checks in create.go. this way we have only 1 code path that checks for supported features, and the logic test can exercise the code path in create.go.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416)*

BramGruneir:
Review status: 6 of 12 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/logictest/testdata/logic_test/fk, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-Kx9hidLTziBNjZuV6BX:-KxDwwbfc6bNZIEVCwq3:b-3om433) ([raw file](https://github.com/cockroachdb/cockroach/blob/1093b7508a65e5f8847dacebd701327d6ed83443/pkg/sql/logictest/testdata/logic_test/fk#L69)):*

Ok now I know what was confusing me.

With your patch we now have two places where we check for unsupported syntax: both in the parser (rejects CASCADE etc) and create.go (rejects currently nothing: the parser only lets NO ACTION and RESTRICT through). 

What I had in mind is to let _all_ the syntax through (with suitable tests in parse_test.go), and have *all* the supported checks in create.go. this way we have only 1 code path that checks for supported features, and the logic test can exercise the code path in create.go.

</blockquote></details>

Works for me!
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416)*

dt:
just to second what bram said: `no action` and `restrict` behave differently when we implement `deferrable`. Ideally we should have been rejecting `RESTRICT` if we weren't capturing it in table descriptors, since to match postgres, the default if unspecified was supposed to be `NO ACTION`. I think it's slightly unfortunate we've been accepting explicit `RESTRICT` and not distinguishing it in the descriptor from unspecified, but I'm not _that_ worried about it.

If we do want to migrate though, it would probably be now, while we can tell them apart, not when we add deferrable, but I don't really know that it is worth it.

---

Review status: 6 of 12 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-KxDxjj8KsNxruAecc5Q:buypzsi)*

knz:
:lgtm: Well done.

---

Reviewed 9 of 9 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-KxE2wMqvwXg4LITtGOZ:b-wehp93)*

jordanlewis:
After in-person discussion I am still mildly worried about migrations concerns.

The worst-case scenario is that someone upgrades their database directly from 1.1 to a version that has `DEFERRABLE` constraints implemented, and they have an old table that was created with `RESTRICT`.

At that point, what can happen? Since `NO ACTION` is deferrable, but `RESTRICT` isn't, user transactions can fail where they shouldn't. For example:

```
CREATE TABLE t (id INT PRIMARY KEY)

CREATE TABLE u (
  id INT PRIMARY KEY, t_id INT,
  CONSTRAINT fk FOREIGN KEY (t_id)
    REFERENCES t(id)
    ON DELETE RESTRICT
);
```

<user upgrades to CockroachDB 1.2 which includes `DEFERRABLE` constraints>

```
BEGIN;
SET CONSTRAINTS ALL DEFERRED;
INSERT INTO u VALUES (1, 1);
INSERT INTO t VALUES (1);
END;
```

The above transaction will succeed, but it should have failed. This is somewhat of a contrived example, to be sure, but I don't think we can just paper this problem over. At the least we'll need to make a note in the release notes that users that need `RESTRICT` behavior must perform the proper `ALTER TABLE` command to restore it post-upgrade.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-KxE8JVUsUBqpw9E88C7:ba8bqnd)*

dt:
Well, we could the proto enum default `RESTRICT` instead, so that tables created <1.2, with an ignored `RESTRICT` modifier, will correctly behave as `RESTRICT` when we add `deferrable`. Downside to that though is that tables created per 1.2 that didn't specify anything will also look like they were explicitly marked as `RESTRICT`, while after 1.2 we'll default to `NO ACTION`. Fortunately, we didn't even accept `NO ACTION` before, so we don't have to worry about anyone who provided it explicitly.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-KxEEPJUxExDkYafuwfV:b-lkla3n)*

BramGruneir:
Ok, after some in person discussion, we're going to stick to the behaviour in this PR and we will send out an advisory release note stipulating that all tables created prior to this change, even if they had specified `RESTRICT` will not be set to `NO ACTION` and that to have a table with restrict will require re-creating the foreign key reference.  I'll add this to the new release notes section in the commit message.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-KxELFZrdQDBXNrYWTLi:b-fj22lr)*

jordanlewis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19416#-:-KxEQv6BTpKricC2QipU:bnfp4nl)*

PR #19417, opened 2017-10-20
sql: extract not-null constraints from filters

RaduBerinde:
#### sql: infrastructure to infer not-null constraints

Adding functions that infer not-null variable constraints from filters. These
will be used to improve the not-null column info in physicalProps.

#### sql: extract not-null constraints from filters

Extract not-null column information from filter expressions (applies to scanNode
and filterNode).

Discussed in #19343.

### DIFF
--- a/pkg/sql/expr_filter.go
+++ b/pkg/sql/expr_filter.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/cockroachdb/cockroach/pkg/sql/parser"
+	"github.com/cockroachdb/cockroach/pkg/util"
 )
 
 // varConvertFunc is a callback that is used when splitting filtering expressions (see splitFilter).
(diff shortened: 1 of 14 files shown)


petermattis:
Smaller change that I expected. I'm not terribly familiar with this code at the moment, but the test changes :lgtm:.

---

Reviewed 2 of 2 files at r1, 2 of 2 files at r2.
Review status: 4 of 14 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/sql/expr_filter_test.go, line 188 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19417#-KwveZ11o8asomiFVH_H:-KwveZ12HBqKQoZg_fp0:bgt33u8) ([raw file](https://github.com/cockroachdb/cockroach/blob/035691225f4b7afd161b00d37793413a1b2a31d8/pkg/sql/expr_filter_test.go#L188)):*
> ```Go
> 		{`a > b`, []int{0, 1}},
> 		{`a = b`, []int{0, 1}},
> 		{`a + b`, []int{0, 1}},
> ```

This isn't a valid filter as it isn't returning a boolean. As a projection (i.e. render), it doesn't imply `a` or `b` is not NULL. Perhaps the expression should be `a + b > 1`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19417#-:-Kwvgu_kBQTuY_BRVd8I:b-wdwx9n)*

knz:
:lgtm:

---

Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 10 of 10 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19417#-:-KwyhAxGfZXGOJIa1W9i:bnfp4nl)*

RaduBerinde:
*[pkg/sql/expr_filter_test.go, line 188 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19417#-KwveZ11o8asomiFVH_H:-Kwz1_O2aGxVTHlWdaVY:b9rfyrb) ([raw file](https://github.com/cockroachdb/cockroach/blob/035691225f4b7afd161b00d37793413a1b2a31d8/pkg/sql/expr_filter_test.go#L188)):*

This isn't a valid filter as it isn't returning a boolean. As a projection (i.e. render), it doesn't imply `a` or `b` is not NULL. Perhaps the expression should be `a + b > 1`. 
</blockquote></details>

Good point. I added an assertion that the expression is bool (or null).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19417)*

Issue #16796, opened 2017-06-29
perf: timestampCache causes GC-related slowdown


petermattis:
Running a read-only workload on a single node cluster shows performance degradation from ~19k ops/sec to a steady state of ~13.5k ops/sec. Investigation eventually pointed to the `timestampCache`. Specifically, if `timestampCache.AddRequest` is disabled, steady state performance is ~18k ops/sec. The workload under investigation is:

```
kv --read-percent 100 --concurrency 16 --splits 100 --write-seq 100000
```

Interestingly, the manipulation of the `timestampCache.requests` btree by `AddRequest` doesn't seem to be the problem. If `AddRequest` is tweaked to add and immediately delete the request, performance is good.

Part of the problem seems to be the btree. Replacing the btree with a fixed size ring buffer of approximately the same size results in steady state performance of ~15k ops/sec. The ring buffer experiment isn't a drop in replacement. More work would be required to flesh out its functionality and that additional functionality might end up negating the modest improvement it is providing.

Another experiment was to zero the `cacheRequest` before inserting it. This brought steady state performance back up to ~17k ops/sec. 

My current suspicion is that the non-zeroed `cacheRequests` are causing additional GC pressure. Enabling `GODEBUG=gctrace=1` shows:

|   | GC CPU  |
| ------------- | ------------- |
| master | 13% |
| disabled timestampCache | 7% |
| zeroed cacheRequest | 8% |

Note that there are two fields in `cacheRequest` that zeroing affects with this workload: `cacheRequest.span` and `cacheRequest.reads`. I'm not sure what to do here. We need the spans in order to later expand the requests into the interval tree. It is surprising to me that merely holding on to these keys is causing additional GC CPU usage. Suggestions are welcome.

bdarnell:
Try zeroing the `span` and `reads` field independently. I would expect that zeroing `span` wouldn't make a difference for GC costs (since it just points to bytes), but zeroing `reads` would (since it's a slice containing pointers).

How much of the Go memory was used by the timestamp cache in these tests? It's usually the biggest single consumer of memory on the go side, in which case it might not be surprising that eliminating it cuts GC costs nearly in half. If that's the case, our options are limited. We could allocate a big slab of memory, copy the keys into it, and store indexes into that slab instead of the pointers in `roachpb.Span`. Or we could move the whole timestamp cache into C++.

petermattis:
|  | ops/sec | GC CPU |
| ---- | ---- | ---- |
[...]

PR #19418, opened 2017-10-20
[WIP] storage: clone Spans before adding them to tscache.Request and use sync.Pool

nvb:
Related to #16796.

This PR makes two related changes:
- it clones all Spans before adding them to a `tscache.Request`
- it adds `*tscache.Request` to a `sync.Pool`

Both of these were done to reduce GC time. Using GCTrace on my linux machine (since
it seems to be broken on Darwin) a read-only workload originally showed a consistent 
**7%** of time spent in GC. This was when running:
```
kv --read-percent 100 --concurrency 16 --splits 100 --write-seq 100000
```

After the first change, this time dropped to **3%**. After the second change, it dropped
again to **2%**. 

The second change is pretty straightforward. The first is more interesting, especially in
light of the experimentation done in #16796. If I individually deep copy the `Spans` by
adding a `Clone` method to `roachpb.Span`, I don't see the same effect as I do here. I
had originally thought that the old GC time was in-part due to scanning the corresponding
`BatchRequest` and `endCmds` structs, which were for some reason being referenced and
therefore unable to be GCed. This proves that that was not the case. This leads me to believe
that the GC cost reduction is due to the single slab of memory being allocated for all keys in
a `BatchRequest`. This theory is fairly unsubstantiated though. I'm curious if anyone has any
other ideas as to why we're seeing such a big difference.

In addition to the GC impact, the two commits also has a fairly large impact on `kv`
performance as a whole. The following are the results from three 1 minute `kv` runs
before and after the changes. As would be expected from a GC related change like this,
we see the impact on the tail latencies. Specifically, it reduces 99th percentile latency by
about **33%**. It also increases throughput by about **7%**.

### Before
```
   60.0s        0        1985392        33089.4      0.5      0.4      0.9      2.6     30.4

BenchmarkLoadgenKV/readPercent=100/splits=100/concurrency=16/duration=1m0s	 1985395	     30221 ns/op

   60.0s        0        1979390        32989.3      0.5      0.4      0.9      2.8     27.3

BenchmarkLoadgenKV/readPercent=100/splits=100/concurrency=16/duration=1m0s	 1979390	     30312 ns/op

   60.0s        0        1986851        33113.6      0.5      0.4      0.9      2.8     27.3

BenchmarkLoadgenKV/readPercent=100/splits=100/concurrency=16/duration=1m0s	 1986852	     30199 ns/op
```

### After
```
   60.0s        0        2121080        35350.8      0.5      0.4      0.9      1.8     23.1

BenchmarkLoadgenKV/readPercent=100/splits=100/concurrency=16/duration=1m0s	 2121080	     28287 ns/op

   60.0s        0        2128153        35468.7      0.5      0.4      0.8      1.8     27.3

BenchmarkLoadgenKV/readPercent=100/splits=100/concurrency=16/duration=1m0s	 2128153	     28193 ns/op

   60.0s        0        2123837        35396.8      0.5      0.4      0.8      1.8     26.2

BenchmarkLoadgenKV/readPercent=100/splits=100/concurrency=16/duration=1m0s	 2123847	     28251 ns/op
```

### DIFF
--- a/pkg/util/bufalloc/byte_allocator.go
+++ b/pkg/util/bufalloc/byte_allocator.go
@@ -14,6 +14,10 @@
 
 package bufalloc
 
+import (
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
+)
+
 // ByteAllocator provides chunk allocation of []byte, amortizing the overhead
 // of each allocation. Because the underlying storage for the slices is shared,
 // they should share a similar lifetime in order to avoid pinning large amounts
@@ -60,4 +64,53 @@ func (a ByteAllocator) Copy(src []byte, extraCap int) (ByteAllocator, []byte) {
 	a, alloc = a.Alloc(len(src), extraCap)
 	copy(alloc, src)
 	return a, alloc
+}
+
+// CopySpan performs a deep copy of the provided Span. The returned Span
+// should be treated as immutable.
+func (a ByteAllocator) CopySpan(src roachpb.Span) (ByteAllocator, roachpb.Span) {
+	if len(src.Key) == 0 {
+		return a, roachpb.Span{}
+	}
+	var dst roachpb.Span
+	extraCap := 0
+	if len(src.EndKey) > 0 {
+		a, dst.EndKey = a.Copy(src.EndKey, 0)
+	} else {
+		// In a few places including in the TimestampCache we translate
+		// [Key, nil) -> [Key, Key.Next()). If the start key has enough
+		// extra capacity, the Next call does not need to allocate and
+		// can instead share the same memory with the start key.
+		extraCap = 1
+	}
+	a, dst.Key = a.Copy(src.Key, extraCap)
+	return a, dst
+}
+
+// CopyRSpan performs a deep copy of the provided RSpan. The returned RSpan
+// should be treated as immutable.
+func (a ByteAllocator) CopyRSpan(src roachpb.RSpan) (ByteAllocator, roachpb.RSpan) {
+	var dst roachpb.Span
+	a, dst = a.CopySpan(src.AsRawSpanWithNoLocals())
+	return a, roachpb.RSpan{
+		Key:    roachpb.RKey(dst.Key),
+		EndKey: roachpb.RKey(dst.EndKey),
+	}
+}
+
+// SpanSize returns the amount of memory required to copy the provided Span.
+func SpanSize(src roachpb.Span) int {
+	if len(src.Key) == 0 {
+		return 0
+	}
+	endSize := 1
+	if len(src.EndKey) > 0 {
+		endSize = len(src.EndKey)
+	}
+	return len(src.Key) + endSize
+}
+
+// RSpanSize returns the amount of memory required to copy the provided RSpan.
+func RSpanSize(src roachpb.RSpan) int {
+	return SpanSize(src.AsRawSpanWithNoLocals())
 }
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -1930,10 +1930,9 @@ func (ec *endCmds) done(br *roachpb.BatchResponse, pErr *roachpb.Error, retry pr
 
 func makeTSCacheRequest(
 	ba *roachpb.BatchRequest, br *roachpb.BatchResponse, span roachpb.RSpan,
-) tscache.Request {
-	cr := tscache.Request{
-		Timestamp: ba.Timestamp,
-	}
+) *tscache.Request {
+	cr := tscache.NewRequest()
+	cr.Timestamp = ba.Timestamp
 	if ba.Txn != nil {
 		cr.TxnID = ba.Txn.ID
 	}
@@ -1988,10 +1987,15 @@ func makeTSCacheRequest(
 		}
 	}
 
-	if readCount > 0 {
+	// Allocate slices for the read and write spans. NewRequest will return a
+	// Read slice with some space already allocated, so we try to use this
+	// wherever possible.
+	if cap(cr.Reads) < readCount {
+		// Maybe we can use the pre-allocated span slice for Writes?
+		cr.Writes = cr.Reads
 		cr.Reads = make([]roachpb.Span, 0, readCount)
 	}
-	if writeCount > 0 {
+	if cap(cr.Writes) < writeCount {
 		cr.Writes = make([]roachpb.Span, 0, writeCount)
 	}
 
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -48,6 +48,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/storage/storagebase"
 	"github.com/cockroachdb/cockroach/pkg/storage/tscache"
 	"github.com/cockroachdb/cockroach/pkg/util"
+	"github.com/cockroachdb/cockroach/pkg/util/bufalloc"
 	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/humanizeutil"
@@ -1931,29 +1932,93 @@ func makeTSCacheRequest(
 	ba *roachpb.BatchRequest, br *roachpb.BatchResponse, span roachpb.RSpan,
 ) tscache.Request {
 	cr := tscache.Request{
-		Span:      span,
 		Timestamp: ba.Timestamp,
 	}
 	if ba.Txn != nil {
 		cr.TxnID = ba.Txn.ID
 	}
 
+	// Copy all keys necessary for the TimestampCache Request into a single byte
+	// slice. This has been shown to dramatically reduce the GC time spent
+	// scanning TimestampCache-related memory. Before doing so, add up the
+	// length of all keys we'll need to copy so we can allocate exactly the
+	// buffer size we'll need. While doing so, we can also count how large the
+	// read and write span slices need to be.
+	var readCount, writeCount int
+	keysSize := bufalloc.RSpanSize(span)
 	for i, union := range ba.Requests {
 		args := union.GetInner()
 		if roachpb.UpdatesTimestampCache(args) {
 			header := args.Header()
+
+			// Count key sizes and read or write requests. This should be kept
+			// in-sync with the type switch below.
+			switch args.(type) {
+			case *roachpb.DeleteRangeRequest:
+				writeCount++
+				keysSize += bufalloc.SpanSize(header)
+			case *roachpb.EndTransactionRequest:
+				keysSize += bufalloc.SpanSize(header)
+			case *roachpb.ScanRequest:
+				readCount++
+				keysSize += len(header.Key)
+				resp := br.Responses[i].GetInner().(*roachpb.ScanResponse)
+				if ba.Header.MaxSpanRequestKeys != 0 &&
+					ba.Header.MaxSpanRequestKeys == int64(len(resp.Rows)) {
+					// See below. +1 because we call Next on this key.
+					keysSize += len(resp.Rows[len(resp.Rows)-1].Key) + 1
+				} else {
+					keysSize += len(header.EndKey)
+				}
+			case *roachpb.ReverseScanRequest:
+				readCount++
+				keysSize += len(header.EndKey)
+				resp := br.Responses[i].GetInner().(*roachpb.ReverseScanResponse)
+				if ba.Header.MaxSpanRequestKeys != 0 &&
+					ba.Header.MaxSpanRequestKeys == int64(len(resp.Rows)) {
+					// See below.
+					keysSize += len(resp.Rows[len(resp.Rows)-1].Key)
+				} else {
+					keysSize += len(header.Key)
+				}
+			default:
+				readCount++
+				keysSize += bufalloc.SpanSize(header)
+			}
+		}
+	}
+
+	if readCount > 0 {
+		cr.Reads = make([]roachpb.Span, 0, readCount)
+	}
+	if writeCount > 0 {
+		cr.Writes = make([]roachpb.Span, 0, writeCount)
+	}
+
+	// Create a ByteAllocator with exactly the number of bytes we'll need for
+	// all keys in the CacheRequest. It should never need to re-alloc.
+	alloc := bufalloc.ByteAllocator(make([]byte, 0, keysSize))
+	alloc, cr.Span = alloc.CopyRSpan(span)
+
+	for i, union := range ba.Requests {
+		args := union.GetInner()
+		if roachpb.UpdatesTimestampCache(args) {
+			header := args.Header()
+			var headerCopy roachpb.Span
 			switch args.(type) {
 			case *roachpb.DeleteRangeRequest:
 				// DeleteRange adds to the write timestamp cache to prevent
 				// subsequent writes from rewriting history.
-				cr.Writes = append(cr.Writes, header)
+				alloc, headerCopy = alloc.CopySpan(header)
+				cr.Writes = append(cr.Writes, headerCopy)
 			case *roachpb.EndTransactionRequest:
 				// EndTransaction adds to the write timestamp cache to ensure replays
 				// create a transaction record with WriteTooOld set.
 				//
 				// Note that TimestampCache.ExpandRequests lazily creates the
 				// transaction key from the request key.
-				cr.Txn = roachpb.Span{Key: header.Key}
+				alloc, headerCopy = alloc.CopySpan(header)
+				cr.Txn = headerCopy
 			case *roachpb.ScanRequest:
 				resp := br.Responses[i].GetInner().(*roachpb.ScanResponse)
 				if ba.Header.MaxSpanRequestKeys != 0 &&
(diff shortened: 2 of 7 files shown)


pkg/util/bufalloc/byte_allocator.go
```diff
@@ -61,3 +65,26 @@ func (a ByteAllocator) Copy(src []byte, extraCap int) (ByteAllocator, []byte) {
...
+// CopySpan performs a deep copy of the provided Span.
+func (a ByteAllocator) CopySpan(src roachpb.Span) (ByteAllocator, roachpb.Span) {
+	var dst roachpb.Span
+	extraCap := 0
+	if len(src.EndKey) > 0 {
+		a, dst.EndKey = a.Copy(src.EndKey, 0)
+	} else {
+		extraCap = 1
+	}
+	a, dst.Key = a.Copy(src.Key, extraCap)
```
tschottdorf: Comment that this is a cheaper way to get `key.Next()` and `key` to share the same memory and is kosher because the returned `Span` is immutable. That latter assumption is worth mentioning in the method comment.

tschottdorf:
Nice! :lgtm: mod TODOs and @petermattis.

---

Reviewed 3 of 3 files at r1, 6 of 6 files at r2.

---

*[pkg/storage/replica.go, line 1941 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvnEkIpZTCgUpPGsSJ:-KwvnEkIpZTCgUpPGsSK:bq4l77b) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1941)):*
> ```Go
> 	}
> 
> 	var alloc bufalloc.ByteAllocator
> ```

Add commentary explaining why copying spans is the better approach.

---

*[pkg/storage/replica.go, line 1989 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvndJXGRA3hE7YypHe:-KwvndJXGRA3hE7YypHf:b-tgi9tp) ([raw file](https://github.com/cockroachdb/cockroach/blob/e50a14fb4fbd0e50319a9e7ce0243334d4ac1eb7/pkg/storage/replica.go#L1989)):*
> ```Go
> 					// the span.
> 
> 					// Make a copy of the key to avoid holding on to the large buffer the
> ```

This was a bad idea then or just no measurable difference?

---

*[pkg/storage/tscache/cache_impl.go, line 573 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-Kwvp5AmF5ehqUhS2929:-Kwvp5AmF5ehqUhS292A:bs3wdpu) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c2e547b457a3ae3a1797d190c675c83000c19d/pkg/storage/tscache/cache_impl.go#L573)):*
> ```Go
> 	// Expand the requests, inserting the spans into either the read or write
> 	// interval caches.
> 	for _, req := range reqs {
> ```

Does it make sense to zero out `reqs` while you're at it? The thinking being that `reqs` would otherwise continue to reference `req` and GC has to look at that.

```go
for i := range reqs {
  req, reqs[i] = reqs[i], nil
  // ...
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-:-Kwvppm_m3DQeVoT-qdX:bhrc17b)*

petermattis:
That's a nice win.

---

---

*[pkg/storage/replica.go, line 1941 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvnEkIpZTCgUpPGsSJ:-Kww5--gLReqQ8Oh5cEC:b-tdgutz) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1941)):*

Add commentary explaining why copying spans is the better approach.
</blockquote></details>

There is a serious downside to this approach: the minimum chunk size used by `bufalloc.ByteAllocator` is 512 bytes. 

As an alternative, perhaps we should make a first pass over the request gathering the spans and their total size and then allocate a buffer of exactly that size.

---

*[pkg/storage/replica.go, line 1953 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-Kww4jfDzgLImeDxRvLc:-Kww4jfDzgLImeDxRvLd:begmxm) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1953)):*
> ```Go
> 				// DeleteRange adds to the write timestamp cache to prevent
> 				// subsequent writes from rewriting history.
> 				alloc, headerCopy = alloc.CopySpan(header)
> ```

Was this actually necessary? Keys that come from the request should have been allocated separately. Or perhaps bundling them into big allocation buffers is a win for the GC. Curious to understand this better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-:-Kww5YB8QFwvmdVtGiqm:bqa7z7y)*

nvanbenschoten:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/storage/replica.go, line 1941 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvnEkIpZTCgUpPGsSJ:-KxAC08sQiBuPpOSZGKN:b-qu6n94) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1941)):*

There is a serious downside to this approach: the minimum chunk size used by `bufalloc.ByteAllocator` is 512 bytes. 

As an alternative, perhaps we should make a first pass over the request gathering the spans and their total size and then allocate a buffer of exactly that size.
</blockquote></details>

Good point. Done.

---

*[pkg/storage/replica.go, line 1953 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-Kww4jfDzgLImeDxRvLc:-KxAC6M08Tu5oSxOhuMP:buxo284) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1953)):*

Was this actually necessary? Keys that come from the request should have been allocated separately. Or perhaps bundling them into big allocation buffers is a win for the GC. Curious to understand this better.
</blockquote></details>

One of my experiments was simply cloning all `Spans` separately, but that didn't seem to provide the same improvement. This leads me to believe that it's the bundling of all keys into the same allocation buffer that provides the GC win we see here. To double check this, the following diff can be applied on top of this PR. It results in GC time more than doubling up to **5%**:

```
diff --git a/pkg/storage/replica.go b/pkg/storage/replica.go
index bfe64eca8..390fc6eea 100644
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -2001,7 +2001,7 @@ func makeTSCacheRequest(

        // Create a ByteAllocator with exactly the number of bytes we'll need for
        // all keys in the CacheRequest. It should never need to re-alloc.
-       alloc := bufalloc.ByteAllocator(make([]byte, keysSize))
+       alloc := bufalloc.ByteAllocator(nil)
        alloc, cr.Span = alloc.CopyRSpan(span)

        for i, union := range ba.Requests {
diff --git a/pkg/util/bufalloc/byte_allocator.go b/pkg/util/bufalloc/byte_allocator.go
index 6f016ef83..b0bad095c 100644
--- a/pkg/util/bufalloc/byte_allocator.go
+++ b/pkg/util/bufalloc/byte_allocator.go
@@ -75,7 +75,8 @@ func (a ByteAllocator) CopySpan(src roachpb.Span) (ByteAllocator, roachpb.Span)
        var dst roachpb.Span
        extraCap := 0
        if len(src.EndKey) > 0 {
-               a, dst.EndKey = a.Copy(src.EndKey, 0)
+               dst.EndKey = make([]byte, len(src.EndKey))
+               copy(dst.EndKey, src.EndKey)
        } else {
                // In a few places including in the TimestampCache we translate
                // [Key, nil) -> [Key, Key.Next()). If the start key has enough
@@ -83,7 +84,8 @@ func (a ByteAllocator) CopySpan(src roachpb.Span) (ByteAllocator, roachpb.Span)
                // can instead share the same memory with the start key.
                extraCap = 1
        }
-       a, dst.Key = a.Copy(src.Key, extraCap)
+       dst.Key = make([]byte, len(src.Key), len(src.Key)+extraCap)
+       copy(dst.Key, src.Key)
        return a, dst
 }
```

This also makes me pretty confident that the arena allocation alone in https://github.com/andy-kimball/arenaskl/blob/master/arena.go will be a big win (and may make a large part of this change unnecessary).

---

*[pkg/storage/replica.go, line 1989 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvndJXGRA3hE7YypHe:-KxADX_RdTY21oSZgV11:b2yf9eq) ([raw file](https://github.com/cockroachdb/cockroach/blob/e50a14fb4fbd0e50319a9e7ce0243334d4ac1eb7/pkg/storage/replica.go#L1989)):*

This was a bad idea then or just no measurable difference?
</blockquote></details>

We'll still make the copy in `CopySpan` below!

---

*[pkg/storage/tscache/cache_impl.go, line 573 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-Kwvp5AmF5ehqUhS2929:-KwvsdlLzKTRZRgW9GuN:b6dihjd) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c2e547b457a3ae3a1797d190c675c83000c19d/pkg/storage/tscache/cache_impl.go#L573)):*

Does it make sense to zero out `reqs` while you're at it? The thinking being that `reqs` would otherwise continue to reference `req` and GC has to look at that.

```go
for i := range reqs {
  req, reqs[i] = reqs[i], nil
  // ...
}
```
</blockquote></details>

Yeah, that's exactly what we want to do. `Request.release` already beat you to it though :wink: 

There's a bit of new logic in `Request.release`, so PTAL at that.

---

*[pkg/util/bufalloc/byte_allocator.go, line 78 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvV-W5uMGwJsvfiMOZ-r1-78:-KxADe5ruTx8ywgdtS9S:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/util/bufalloc/byte_allocator.go#L78)):*

Comment that this is a cheaper way to get `key.Next()` and `key` to share the same memory and is kosher because the returned `Span` is immutable. That latter assumption is worth mentioning in the method comment.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418)*

petermattis:
:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica.go, line 1941 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KwvnEkIpZTCgUpPGsSJ:-KxDdjKtiVzNz83k1aUA:breo44v) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1941)):*

Good point. Done.
</blockquote></details>

The two loops over the request are unfortunate, but I see that it allows you to size the read and write slice precisely. 

---

*[pkg/storage/replica.go, line 1953 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-Kww4jfDzgLImeDxRvLc:-KxDcv6UyknKUt3haRRW:b-ncy1k0) ([raw file](https://github.com/cockroachdb/cockroach/blob/9a95e3239a7b41a36aad2ce14583efc8daa383f9/pkg/storage/replica.go#L1953)):*

One of my experiments was simply cloning all `Spans` separately, but that didn't seem to provide the same improvement. This leads me to believe that it's the bundling of all keys into the same allocation buffer that provides the GC win we see here. To double check this, the following diff can be applied on top of this PR. It results in GC time more than doubling up to **5%**:

```
diff --git a/pkg/storage/replica.go b/pkg/storage/replica.go
index bfe64eca8..390fc6eea 100644
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -2001,7 +2001,7 @@ func makeTSCacheRequest(

        // Create a ByteAllocator with exactly the number of bytes we'll need for
        // all keys in the CacheRequest. It should never need to re-alloc.
-       alloc := bufalloc.ByteAllocator(make([]byte, keysSize))
+       alloc := bufalloc.ByteAllocator(nil)
        alloc, cr.Span = alloc.CopyRSpan(span)

        for i, union := range ba.Requests {
diff --git a/pkg/util/bufalloc/byte_allocator.go b/pkg/util/bufalloc/byte_allocator.go
index 6f016ef83..b0bad095c 100644
--- a/pkg/util/bufalloc/byte_allocator.go
+++ b/pkg/util/bufalloc/byte_allocator.go
@@ -75,7 +75,8 @@ func (a ByteAllocator) CopySpan(src roachpb.Span) (ByteAllocator, roachpb.Span)
        var dst roachpb.Span
        extraCap := 0
        if len(src.EndKey) > 0 {
-               a, dst.EndKey = a.Copy(src.EndKey, 0)
+               dst.EndKey = make([]byte, len(src.EndKey))
+               copy(dst.EndKey, src.EndKey)
        } else {
                // In a few places including in the TimestampCache we translate
                // [Key, nil) -> [Key, Key.Next()). If the start key has enough
@@ -83,7 +84,8 @@ func (a ByteAllocator) CopySpan(src roachpb.Span) (ByteAllocator, roachpb.Span)
                // can instead share the same memory with the start key.
                extraCap = 1
        }
-       a, dst.Key = a.Copy(src.Key, extraCap)
+       dst.Key = make([]byte, len(src.Key), len(src.Key)+extraCap)
+       copy(dst.Key, src.Key)
        return a, dst
 }
```

This also makes me pretty confident that the arena allocation alone in https://github.com/andy-kimball/arenaskl/blob/master/arena.go will be a big win (and may make a large part of this change unnecessary).
</blockquote></details>

Interesting. Thanks.

---

*[pkg/storage/tscache/request.go, line 92 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KxDdUFdNMEGSxtrDYqo:-KxDdUFdNMEGSxtrDYqp:b-87c9y8) ([raw file](https://github.com/cockroachdb/cockroach/blob/928ec15149d529cc1cbbcde3519eb2fae43781d4/pkg/storage/tscache/request.go#L92)):*
> ```Go
> 	if cap(cr.Reads) > requestAllocSpans {
> 		if cap(cr.Writes) > requestAllocSpans {
> 			// Throw away.
> ```

Can't you keep the request, but set `Request.reads = nil` and `Request.writes = nil`? Then in `NewRequest` you'd reinitialize `cr.Reads = crAlloc.spans[:0]`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-:-KxDdqqL5J0dlhz5SfOS:bnfp4nl)*

nvanbenschoten:
Review status: 0 of 7 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[pkg/storage/tscache/request.go, line 92 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KxDdUFdNMEGSxtrDYqo:-KxHSDbxj_pEL_BkiBPQ:b8ytfod) ([raw file](https://github.com/cockroachdb/cockroach/blob/928ec15149d529cc1cbbcde3519eb2fae43781d4/pkg/storage/tscache/request.go#L92)):*

Can't you keep the request, but set `Request.reads = nil` and `Request.writes = nil`? Then in `NewRequest` you'd reinitialize `cr.Reads = crAlloc.spans[:0]`. 
</blockquote></details>

The issue is that we don't have access to the `crAlloc` struct anywhere outside of the `sync.Pool` constructor. Two options would be:
1. don't worry about `cr.Reads` losing a reference to its collocated array, just truncate whatever slice it has at this point (what this change used to do)
2. maintain a pointer from `Requests` to `requestAlloc`

Neither of those sound like particularly good options. The first would result in fewer and fewer `Requests` in the pool being collocated with their `Reads` slice over time, resulting in fragmentation and wasted memory. The second seems like a waste of space and makes this optimization leak into other code that shouldn't have to worry about it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418)*

petermattis:
Review status: 0 of 7 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[pkg/storage/tscache/request.go, line 92 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418#-KxDdUFdNMEGSxtrDYqo:-KxITXlwxGrtBep3_tuC:b-2r12wq) ([raw file](https://github.com/cockroachdb/cockroach/blob/928ec15149d529cc1cbbcde3519eb2fae43781d4/pkg/storage/tscache/request.go#L92)):*

The issue is that we don't have access to the `crAlloc` struct anywhere outside of the `sync.Pool` constructor. Two options would be:
1. don't worry about `cr.Reads` losing a reference to its collocated array, just truncate whatever slice it has at this point (what this change used to do)
2. maintain a pointer from `Requests` to `requestAlloc`

Neither of those sound like particularly good options. The first would result in fewer and fewer `Requests` in the pool being collocated with their `Reads` slice over time, resulting in fragmentation and wasted memory. The second seems like a waste of space and makes this optimization leak into other code that shouldn't have to worry about it.
</blockquote></details>

Ah, I was thinking you could change `requestPool` to return a `requestAlloc`, but then you'd still need to hold on to the `requestAlloc` for when you put the request back in the pool. Carry on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19418)*

PR #19419, opened 2017-10-20
cherry-pick 1.1: sql: don't set isKey if columns are nullable

### LINKED ISSUE #19343: sql: incorrect key property for unique indexes with nullable columns

RaduBerinde:
We assume that the columns of an index form a "key" but that is only correct if
the columns are not null.

Fixes #19343.

This is not really a cherry-pick since this code has changed a lot; it's a fully reworked fix.

CC @cockroachdb/release

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/order_by
+++ b/pkg/sql/logictest/testdata/logic_test/order_by
@@ -627,15 +627,13 @@ INSERT INTO bar VALUES (0, NULL), (1, NULL)
 query ITTTTT
 EXPLAIN (METADATA) SELECT * FROM bar ORDER BY baz, id
 ----
-0  scan  ·      ·          (id, baz)  +baz,key
-0  ·     table  bar@i_bar  ·          ·
-0  ·     spans  ALL        ·          ·
+0  sort  ·      ·          (id, baz)  +baz,+id
+0  ·     order  +baz,+id   ·          ·
+1  scan  ·      ·          (id, baz)  +baz
+1  ·     table  bar@i_bar  ·          ·
+1  ·     spans  ALL        ·          ·
 
-# Here rowsort is needed because the ORDER BY clause does not guarantee any
-# relative ordering between rows where baz is NULL. As we see above, because
-# this is a unique index, the ordering `+baz,+id` is deemed equivalent to just
-# `+baz`.
-query IT rowsort
+query IT
 SELECT * FROM bar ORDER BY baz, id
 ----
 0 NULL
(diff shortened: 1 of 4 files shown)


pkg/sql/logictest/testdata/logic_test/order_by
```diff
@@ -627,15 +627,13 @@ INSERT INTO bar VALUES (0, NULL), (1, NULL)
 query ITTTTT
 EXPLAIN (METADATA) SELECT * FROM bar ORDER BY baz, id
 ----
-0  scan  ·      ·          (id, baz)  +baz,key
-0  ·     table  bar@i_bar  ·          ·
-0  ·     spans  ALL        ·          ·
-
-# Here rowsort is needed because the ORDER BY clause does not guarantee any
```
RaduBerinde: This comment was incorrect - we advertise that `ORDER BY` has "null first" semantics.

knz:
Thanks!

:lgtm:

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19419#-:-Kwyg3PuQBCxNebtx7W3:b-69ief5)*

PR #19420, opened 2017-10-20
sql: support DEFAULT and MAXVALUE in PARTITION definitions

danhhz:
DEFAULT is used in a list partitioning as a catchall for values not
matched by any other partition. To avoid ambiguity in a multi-column
partitioning, when DEFAULT is specified for a column, it must be
specified for every column "after" that. So `(1, DEFAULT, DEFAULT)` is
valid but (1, DEFAULT, 1) is not.

MAXVALUE is used in a range partitioning as a sentinel that sorts larger
than any data in the column. It has the same ambiguity problem as
DEFAULT and so the same restriction.

@knz this is a strawman for one way we could deal with DEFAULT and MAXVALUE in PARTITION BY. what are your thoughts? I could also reuse DefaultVal instead of introducing PartitionDefault


danhhz:
Forgot to mention, this is obviously quite unpolished, I'm just looking for comments on the overall approach first.

knz:
What you're doing here seems kosher to me. No design comments.

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-:-Kx7UygitU2RZRy8g4ez:b-gu73dq)*

benesch:
Approach looks good to me as well. Just two thoughts below.

---

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 208 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-Kx9RBmYM8IfLzK2QaH7:-Kx9RBmYM8IfLzK2QaH8:b-buqvkd) ([raw file](https://github.com/cockroachdb/cockroach/blob/47b0786f89dc47f9b357ce0c817bfc90f6beb52d/pkg/sql/logictest/testdata/logic_test/partitioning#L208)):*
> ```
>     PARTITION p2 VALUES (1, DEFAULT),
>     PARTITION p3 VALUES (2, 3),
>     PARTITION p4 VALUES (DEFAULT, DEFAULT)
> ```

Do you disallow `MAXVALUE` in list partitions? Seems like it'll be accepted as an alias for `DEFAULT`.

---

*[pkg/sql/parser/parse_test.go, line 148 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-Kx9Q11y0p53Hx9xYQy2:-Kx9Q11y0p53Hx9xYQy3:b-ukf0b1) ([raw file](https://github.com/cockroachdb/cockroach/blob/47b0786f89dc47f9b357ce0c817bfc90f6beb52d/pkg/sql/parser/parse_test.go#L148)):*
> ```Go
> 
> 		{`CREATE TABLE a (b INT) PARTITION BY LIST (b) (PARTITION p1 VALUES (1, DEFAULT), PARTITION p2 VALUES (2), (3))`},
> 		{`CREATE TABLE a (b INT) PARTITION BY RANGE (b) (PARTITION p1 VALUES LESS THAN (1), PARTITION p2 VALUES LESS THAN (2, 3), PARTITION p3 VALUES LESS THAN (MAXVALUE))`},
> ```

We should probably parse a DEFAULT in here too, no?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-:-Kx9RHkXSxgr__VZ-TQ1:b-o8j2f2)*

danhhz:
Okay, this should be ready to go

---

Review status: 9 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 208 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-Kx9RBmYM8IfLzK2QaH7:-KxA2zd0owcAtN0f99UV:b-mll4nk) ([raw file](https://github.com/cockroachdb/cockroach/blob/47b0786f89dc47f9b357ce0c817bfc90f6beb52d/pkg/sql/logictest/testdata/logic_test/partitioning#L208)):*

Do you disallow `MAXVALUE` in list partitions? Seems like it'll be accepted as an alias for `DEFAULT`.
</blockquote></details>

yeah, i just hadn't gotten to implementing it. it's disallowed now

---

*[pkg/sql/parser/parse_test.go, line 148 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-Kx9Q11y0p53Hx9xYQy2:-KxA379PiadaKlbpVaZA:b-p0l9ku) ([raw file](https://github.com/cockroachdb/cockroach/blob/47b0786f89dc47f9b357ce0c817bfc90f6beb52d/pkg/sql/parser/parse_test.go#L148)):*

We should probably parse a DEFAULT in here too, no?
</blockquote></details>

in PARTITION BY RANGE? (fyi, there's already on in the PARTITION BY LIST case. it worked before because we were using ctext_row in sql.y)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-:-KxA4ULOmS7B1_XcF3Hq:b-o11hh1)*

benesch:
:lgtm: but see my note about duplicate MAXVALUEs.

---

Reviewed 7 of 7 files at r2.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 17 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBnnC_xB3z56l8MoRN:-KxBnnC_xB3z56l8MoRO:b-u476bx) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L17)):*
> ```
> CREATE TABLE t (a INT, b INT, c INT) PARTITION BY LIST (a) (
>     PARTITION p1 VALUES (0)
> )
> ```

Heh, took me a while to figure out what was wrong with this one. Nice test.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 89 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBnwhgH8onoV8UN5F5:-KxBnwhgH8onoV8UN5F6:bv7j0uo) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L89)):*
> ```
> )
> 
> statement error cannot be present in more than one partition
> ```

These should probably say `DEFAULT cannot be present in more than one partition` to test that we're not saying `MAXVALUE` accidentally.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBo4CZ6ofmViX9uzze:-KxBo4CZ6ofmViX9uzzf:b-sctrut) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L101)):*
> ```
> )
> 
> statement error PARTITION p1: 1 not allowed after DEFAULT
> ```

This is kind of hard to parse. I think `non-DEFAULT value (1) not allowed after DEFAULT` or something along those lines would be better. Don't hold up this PR though.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 116 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBobj0b24z8h39E5Tv:-KxBobj0b24z8h39E5Tw:byjzlyx) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L116)):*
> ```
>     PARTITION p1 VALUES LESS THAN (MAXVALUE),
>     PARTITION p2 VALUES LESS THAN (1)
> )
> ```

See my note below about testing something like `(1, MAXVALUE)` and `(2, MAXVALUE)`.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 130 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBoIGYHgRYtBwsbJ6z:-KxBoIGYHgRYtBwsbJ7-:bc0nuyx) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L130)):*
> ```
> )
> 
> statement error PARTITION p1: 1 not allowed after MAXVALUE
> ```

Ditto here: I find this error confusing.

---

*[pkg/sql/parser/parse_test.go, line 148 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-Kx9Q11y0p53Hx9xYQy2:-KxBkAiT9mZXyF4_Kcv8:bx2t19f) ([raw file](https://github.com/cockroachdb/cockroach/blob/47b0786f89dc47f9b357ce0c817bfc90f6beb52d/pkg/sql/parser/parse_test.go#L148)):*

in PARTITION BY RANGE? (fyi, there's already on in the PARTITION BY LIST case. it worked before because we were using ctext_row in sql.y)
</blockquote></details>

Oh, whoops. I wasn't paying enough attention. Carry on.

---

*[pkg/sql/sqlbase/structured.go, line 1274 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBktrZa4BkCaTn_ADl:-KxBktr_HkB0u_uW_RL7:b-gff5f8) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/sqlbase/structured.go#L1274)):*
> ```Go
> 		// this encoding has to match the index direction (instead of always
> 		// being ascending).
> 		keyEncBuf, err = EncodeTableKey(keyEncBuf, datums[i], encoding.Ascending)
> ```

This seems easy enough to plumb down now via `indexDesc.ColumnDirections`? No need to hold up this PR; just wondering if there's some roadblock I'm not considering.

---

*[pkg/sql/sqlbase/structured.go, line 1382 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBnMKW08YcvpFPybo4:-KxBnMKW08YcvpFPybo5:b-iqapee) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/sqlbase/structured.go#L1382)):*
> ```Go
> 				// any possible value for this column.
> 				keyEncBuf = roachpb.Key(keyEncBuf).PrefixEnd()
> 			}
> ```

I'm not convinced that this works when partitioning by multiple range columns at once, e.g.

```
(1, DEFAULT)
(2, DEFAULT)
```

But don't take my word for it. It's late.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-:-KxBorfQ32nmskEmxwJL:b-rgnguz)*

benesch:
For posterity: @danhhz convinced me at lunch that `PrefixEnd` safely detects duplicate `MAXVALUE`s. Press onward!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-:-KxE9bqLb924qYYjEeB9:b-qw6yr0)*

danhhz:
@benesch, I'm still working through some merge conflicts, but this should be ready for another look

---

Review status: 13 of 16 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 17 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBnnC_xB3z56l8MoRN:-KxJ8_mPbJq7ARBEbgWy:b7xuroj) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L17)):*

Heh, took me a while to figure out what was wrong with this one. Nice test.
</blockquote></details>

Added a comment to make this easier on the next person

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 89 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBnwhgH8onoV8UN5F5:-KxJTGRJDW2YVTvBJnk5:bb6un1j) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L89)):*

These should probably say `DEFAULT cannot be present in more than one partition` to test that we're not saying `MAXVALUE` accidentally.
</blockquote></details>

Done

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBo4CZ6ofmViX9uzze:-KxJ9An8I5WRXZyO5UXb:b-6j5tlq) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L101)):*

This is kind of hard to parse. I think `non-DEFAULT value (1) not allowed after DEFAULT` or something along those lines would be better. Don't hold up this PR though.
</blockquote></details>

love it. done

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 116 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBobj0b24z8h39E5Tv:-KxJ9U2supDc3vjl7G6a:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L116)):*

See my note below about testing something like `(1, MAXVALUE)` and `(2, MAXVALUE)`.
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 130 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBoIGYHgRYtBwsbJ6z:-KxJ9UfGizH_9cTNq11H:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/logictest/testdata/logic_test/partitioning#L130)):*

Ditto here: I find this error confusing.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.go, line 1274 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBktrZa4BkCaTn_ADl:-KxJ9f9fMQKM6VzSLB7z:bsn5hmx) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/sqlbase/structured.go#L1274)):*

This seems easy enough to plumb down now via `indexDesc.ColumnDirections`? No need to hold up this PR; just wondering if there's some roadblock I'm not considering.
</blockquote></details>

There's no good way to test it right now and i think it deserves some careful tests.

---

*[pkg/sql/sqlbase/structured.go, line 1382 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxBnMKW08YcvpFPybo4:-KxJA9_LMYuaYWxtOgEh:b-88fbs) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ef018f5b0480f535f2383224a8b870135373930/pkg/sql/sqlbase/structured.go#L1382)):*

I'm not convinced that this works when partitioning by multiple range columns at once, e.g.

```
(1, DEFAULT)
(2, DEFAULT)
```

But don't take my word for it. It's late.
</blockquote></details>

Realized there's no good place right now to document what we worked out over lunch (though I definitely want to). Added a TODO to tighten up the semantics of what is responsible for constructing the keyspan for a partition and to document it then (and add lots of tests).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-:-KxJTG-7bra2p-3b3Q16:b-uwyp9k)*

benesch:
Reviewed 13 of 13 files at r3.

---

*[pkg/sql/sqlbase/structured.go, line 1412 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxK8XfaJQyiOQHR_HnU:-KxK8XfaJQyiOQHR_HnV:b-lq3rw6) ([raw file](https://github.com/cockroachdb/cockroach/blob/55e082c6dd3575753e7d8a6d9388a129c1aee9cc/pkg/sql/sqlbase/structured.go#L1412)):*
> ```Go
> 				if len(datums) != len(colIDs) {
> 					return fmt.Errorf("prefix (%s) cannot be present in more than one partition",
> 						printPartitioningPrefix(datums, "MAXVALUE"))
> ```

And perhaps this should be `parser.ErrString(append(datums, parser.DString("MAXVALUE")))`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420)*

benesch:

---

*[pkg/sql/sqlbase/structured.go, line 1412 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420#-KxK8XfaJQyiOQHR_HnU:-KxK9j5zd-K3Nv6ox2TA:b1hd0rx) ([raw file](https://github.com/cockroachdb/cockroach/blob/55e082c6dd3575753e7d8a6d9388a129c1aee9cc/pkg/sql/sqlbase/structured.go#L1412)):*

And perhaps this should be `parser.ErrString(append(datums, parser.DString("MAXVALUE")))`.
</blockquote></details>

Ugh, meant to delete this. I realized halfway through that DStrings get wrapped in quotes. The `printPartitioningPrefix` helper seems as elegant as it's going to get.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19420)*

danhhz:
Yeah, I don't love the helper, but it's the best I could come up with. Thanks for the reviews!

PR #19421, opened 2017-10-20
storage: update commentary on non-preemptible loops and GC

petermattis:
Non-preemptible loops prevent GC from starting which will cause the
whole process to lock up.

See #19404


a-robinson (approved):

tbg (approved):
Thanks!

nvb (approved):

PR #19348, opened 2017-10-18
storage: de-flake TestNodeIsLiveCallback

tbg:
This test exercised that liveness callbacks fired after a node's liveness
changed from false to true. However, the trigger was tied to Gossip updates
only, but the liveness also updated "eagerly" after the heartbeat.

Recent refactorings made it much more likely that the eager update took
precedence, in which case the Gossip handler would avoid updating a
second time, forgetting the callbacks as a consequence.

This was addressed by making callback invocation the duty of the code
that ingests livenesses.

The test was extremely flaky before, and passed several thousand iterations as
of this commit.

Fixes #19344.

PR #19422, opened 2017-10-21
cherrypick-1.1: less clobbering in node liveness

tbg:
Combination of #19386, #19215, #19279, #19348.

cc @cockroachdb/release

### DIFF
--- a/pkg/storage/client_test.go
+++ b/pkg/storage/client_test.go
@@ -816,21 +816,24 @@ func (m *multiTestContext) addStore(idx int) {
 	}
 	store.WaitForInit()
 
+	ran := struct {
+		sync.Once
+		ch chan struct{}
+	}{
+		ch: make(chan struct{}),
+	}
 	m.nodeLivenesses[idx].StartHeartbeat(ctx, stopper, func(ctx context.Context) {
 		now := m.clock.Now()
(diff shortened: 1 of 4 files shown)


bdarnell:
:lgtm:

This fixes the problem of dead nodes not being removed from the UI when decommissioned, right? Do we also expect other user-visible improvements or just fixing test flakiness?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19422#-:-Kx8sKv_nzitTBNlL0mQ:b-spvh8k)*

tschottdorf:
I am hoping that this fixes issues such as https://github.com/cockroachdb/cockroach/issues/18219 (the flake fixes here were actually only exposed/triggered by the new code). Unfortunately, I never got a conclusive sequence of events that would explain the problem (liveness not knowing that gossip thinks a peer is decommissioning), and our acceptance tests don't repro it (and I've tried varying them, too). At the same time, for the folks who had it happen to them, it usually happens on all nodes. But my difficulty in conjuring up a repro could just be a reflection of how the code was to reason about.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19422#-:-Kx8tR8zKD86J00MGruc:btoothe)*

PR #19424, opened 2017-10-21
sql: convert placeholders to leaf values

jordanlewis:
Previously, the `Placeholder` expression type was merely a marker that
was replaced during type checking with an actual Datum value by a tree
traversal.

This was inefficient - every query with placeholders would end up
getting path-copied during that replacement step. Additionally, it was
an obstacle to optimizations like plan caching that requires that
expression trees retain their meaning after type checking.

Now, a `Placeholder` is a fully fledged `Expr` that can be evaluated.
Evaluating a `Placeholder` returns the `Datum` that it points to in the
`PlaceholderInfo` map.

This transformation required a lot of plumbing, which makes up the bulk
of this patch. `EvalContext` had to be augmented to include a
`PlaceholderInfo` map so that the placeholder values can be known at
`Placeholder.Eval()` time. `EvalContext`s and `SemaContext`s also needed
to be plumbed to every place that needs to check for constant
expressions, since checking for constant expressions now requires that
placeholder values be known (placeholders without values are not
constant expressions; placeholders with values are).

**n.b.:**
after this patch alone, we still end up path copying all placeholders because of
`NormalizeVisitor`, which traverses expression trees and `Eval()`s all constant
subexpressions, which `Placeholder` now counts as.

I think the right solution here is to actually have `Placeholder` implement `Datum`
(and plumb `EvalContext` into all `Datum` methods). That way `Placeholder` can
simply forward its methods to its underlying `Datum` and we get the behavior that
we want (no path copying) without changing all of the index selection and analysis
code that expects `Datum` objects in specific parts of the tree.

### DIFF
--- a/pkg/sql/parser/normalize.go
+++ b/pkg/sql/parser/normalize.go
@@ -95,7 +95,7 @@ func (expr *UnaryExpr) normalize(v *normalizeVisitor) TypedExpr {
 		return val
 	case UnaryMinus:
 		// -0 -> 0 (except for float which has negative zero)
-		if val.ResolvedType() != TypeFloat && IsNumericZero(val) {
+		if val.ResolvedType() != TypeFloat && v.isNumericZero(val) {
 			return val
 		}
 		switch b := val.(type) {
@@ -134,25 +134,25 @@ func (expr *BinaryExpr) normalize(v *normalizeVisitor) TypedExpr {
 
 	switch expr.Operator {
 	case Plus:
-		if IsNumericZero(right) {
+		if v.isNumericZero(right) {
 			final, v.err = ReType(left, expectedType)
 			break
 		}
-		if IsNumericZero(left) {
+		if v.isNumericZero(left) {
 			final, v.err = ReType(right, expectedType)
 			break
 		}
 	case Minus:
-		if IsNumericZero(right) {
+		if v.isNumericZero(right) {
 			final, v.err = ReType(left, expectedType)
 			break
 		}
 	case Mult:
-		if IsNumericOne(right) {
+		if v.isNumericOne(right) {
 			final, v.err = ReType(left, expectedType)
 			break
 		}
-		if IsNumericOne(left) {
+		if v.isNumericOne(left) {
 			final, v.err = ReType(right, expectedType)
 			break
 		}
@@ -160,7 +160,7 @@ func (expr *BinaryExpr) normalize(v *normalizeVisitor) TypedExpr {
 		// because if the other operand is NULL during evaluation
 		// the result must be NULL.
 	case Div, FloorDiv:
-		if IsNumericOne(right) {
+		if v.isNumericOne(right) {
 			final, v.err = ReType(left, expectedType)
 			break
 		}
@@ -357,7 +357,7 @@ func (expr *ComparisonExpr) normalize(v *normalizeVisitor) TypedExpr {
 				expr.Left = left.Left
 				expr.Right = newRightExpr
 				expr.memoizeFn()
-				if !isVar(expr.Left) {
+				if !isVar(v.ctx, expr.Left) {
 					// Continue as long as the left side of the comparison is not a
 					// variable.
 					continue
@@ -411,7 +411,7 @@ func (expr *ComparisonExpr) normalize(v *normalizeVisitor) TypedExpr {
 				expr.Left = left.Right
 				expr.Right = newRightExpr
 				expr.memoizeFn()
-				if !isVar(expr.Left) {
+				if !isVar(v.ctx, expr.Left) {
 					// Continue as long as the left side of the comparison is not a
 					// variable.
 					continue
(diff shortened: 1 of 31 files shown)


knz:
n.b. normalization needs only occur once during prepare, and not any more during execute.  It's unfortunate that our current code organization invokes normalization again. I'm tempted to propose to make a separate patch that 1) disables normalization during plan construction 2) traverses the plan (via sql/walk.go) after it's built to normalize the expressions. This second step would only be done when the plan is executed, not when it's just prepared.

pkg/sql/parser/normalize.go
```diff
@@ -718,19 +724,25 @@ func (v *isConstVisitor) run(expr Expr) bool {
...
 
-func isVar(expr Expr) bool {
-	_, ok := expr.(VariableExpr)
-	return ok
+func isVar(evalCtx *EvalContext, expr Expr) bool {
+	switch expr.(type) {
+	case VariableExpr:
+		return true
+	case *Placeholder:
+		return evalCtx != nil && (!evalCtx.HasPlaceholders() || evalCtx.Placeholders.IsUnresolvedPlaceholder(expr))
```
justinj: If I understand this change, maybe it would make sense to rename this function something like `isNonConstant`? I know "variable" as an adjective implies that but it's easy to misread as the noun in this case (or even just a comment implying that "variable" in this context refers to the expression varying, rather than a `VariableExpr`).

justinj (commented):
This is a v good change 👍

jordanlewis:
@knz hmm I don't think that's true, since normalization involves constant subexpression evaluation. For example, take the query:

```
SELECT * FROM kv WHERE unindexed_col = (1 + $1) * (2 + $2)
```

In this case, constant subexpression evaluation during prepare does nothing, since there are no constant subexpressions to evaluate.

After placeholders are available, you really do want to run another pass of constant subexpression evaluation to prevent needing to re-evaluate the arithmetic on every row.

jordanlewis:
OK, totally reworked this - `Placeholder` is now a `Datum` implementation!

There's a lot going on in this patch. I tried my best to split it up into small commits. The last commit is where `Placeholder` is actually modified to be a leaf value - every commit before that are random bugfixes/modifications that had to happen to make everything work.

jordanlewis:
Well, you might want to hold off on this for a bit until I fix the tests.

jordanlewis:
OK, this is ready for a look. There was a syntax roundtripping issue that needed to be solved - see #19555.

knz:
I like where this is going. Nice work.

---

Reviewed 15 of 15 files at r2, 5 of 5 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 11 of 11 files at r6, 2 of 2 files at r7, 19 of 19 files at r8.

---

*[pkg/sql/explain_plan.go, line 91 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424#-KxmAbX3gQPIFlwy9iIY:-KxmAbX3gQPIFlwy9iIZ:b-g9rbvt) ([raw file](https://github.com/cockroachdb/cockroach/blob/98e03566ad3749093950baa0d4a89c557640aacc/pkg/sql/explain_plan.go#L91)):*
> ```Go
> 		parser.FmtSimple, explainer.showTypes, explainer.symbolicVars, explainer.qualifyNames,
> 	)
> 	if _, ok := origStmt.(*parser.Execute); ok {
> ```

Woah this is probably not the location where you want this to happen. What of `explain select * from [execute ...]` ?

---

*[pkg/sql/planner.go, line 167 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424#-KxmBDlp7KVJ3qW0gjkj:-KxmBDlqTvZE1Sr3dreO:b-pohdcw) ([raw file](https://github.com/cockroachdb/cockroach/blob/f209fa664fa166825f5028337a120589e6a8c246/pkg/sql/planner.go#L167)):*
> ```Go
> 	}
> 
> 	p.evalCtx.Placeholders = &p.semaCtx.Placeholders
> ```

See my comment elsewhere -- I'm not sure this needs to be reset so often.

---

*[pkg/sql/session.go, line 633 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424#-Kxm9Y2LXurj4XTNzN73:-Kxm9Y2MH-WKsL8sNcjR:b-q9g56k) ([raw file](https://github.com/cockroachdb/cockroach/blob/95e13d2781ada65a74fccb15271cbceaad67cc75/pkg/sql/session.go#L633)):*
> ```Go
> ) (parser.Datums, error) {
> 	placeholders := ep.p.semaCtx.Placeholders
> 	defer func() {
> ```

Unfortunately the Go compiler doesn't know how to extract this closure properly (i.e. it will force a heap allocation)
You can fix this with:
```go
defer func(prev parser.PlaceholderInfo) { ep.p.semaCtx.Placeholders = prev }(ep.p.semaCtx.Placeholders)
```

Also I may be mistaken but I think the various assignments to `ep.p.evalCtx.Placeholders` are not necessary, because the address of `ep.p.semaCtx.Placeholders` doesn't change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424#-:-KxmB1U3219pr5Ju7MTf:bpspjbg)*

jordanlewis:

---

*[pkg/sql/explain_plan.go, line 91 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424#-KxmAbX3gQPIFlwy9iIY:-KxodJ5VQf2ibNTX9b-s:bu0b4mf) ([raw file](https://github.com/cockroachdb/cockroach/blob/98e03566ad3749093950baa0d4a89c557640aacc/pkg/sql/explain_plan.go#L91)):*

Woah this is probably not the location where you want this to happen. What of `explain select * from [execute ...]` ?
</blockquote></details>

Oof you are right. Any suggestions? I'm not sure how to do this cleanly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424)*

knz:
Reviewed 1 of 31 files at r9, 1 of 1 files at r11, 10 of 11 files at r12, 19 of 19 files at r13.

---

*[pkg/sql/explain_plan.go, line 91 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424#-KxmAbX3gQPIFlwy9iIY:-KxrbcABEy3faFT0oI-o:b-v3x7r5) ([raw file](https://github.com/cockroachdb/cockroach/blob/98e03566ad3749093950baa0d4a89c557640aacc/pkg/sql/explain_plan.go#L91)):*

Oof you are right. Any suggestions? I'm not sure how to do this cleanly.
</blockquote></details>

yes, `newPlan` (`plan.go`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19424)*

jordanlewis:
Going to close this for the time being as most of this work is going to end up in different PRs. I'm still using this branch as a staging area and may open this PR again for review of the final piece that actually switches over to leaf placeholders.

Issue #19316, opened 2017-10-17
build: Lint against `if err != nil { return nil }`


bdarnell:
https://github.com/cockroachdb/cockroach/issues/19172#issuecomment-337061024 is the most recent occurrence of a bug we've seen before:

```
if err != nil {
    return nil
}
```

Most of the time we want to return an error if we see a non-nil error anywhere (usually `err` itself or a wrapper around it). Sometimes we do mean to swallow an error and return nil, but this is rare and should have a comment when it occurs. We should have a linter to detect this case (maybe contribute it upstream to golint/staticcheck/something).

More precisely, the linter should look for an if statement whose condition is comparing a value of type `error` against nil, then returning `nil` as a  return value of type `error`. If there is any comment inside the `if` block, it should be allowed.

tschottdorf:
As part of #19172 I was hoping to just bang this out, but now it looks like it needs more time than I imagined. Here's my (non-compiling) WIP:

```go
diff --git a/build/style_test.go b/build/style_test.go
index 9ce8eb3b1..bba14a420 100644
--- a/build/style_test.go
+++ b/build/style_test.go
@@ -1188,6 +1188,39 @@ func checkConvertFloatToUnsigned(j *lint.Job) {
 	})
 }
 
+func containsErrComparison(j *lint.Job, expr ast.Expr) {
+	binExpr, ok := expr.(*ast.BinaryExpr)
+	if !ok || binExpr.Op != token.NEQ {
+		return
+	}
+	for _, pair := range [2][2]ast.Expr{{binExpr.X, binExpr.Y}, {binExpr.Y, binExpr.X}} {
+		errExpr, nilExpr := pair[0], pair[1]
+		typ := j.Program.Info.TypeOf(errExpr)
+		pkg := j.Program.Info.ObjectOf(errExpr).Package()
+		obj, _, _ := types.LookupFieldOrMethod(typ, true, pkg, "Error")
+		if obj == nil {
+			continue
+		}
+		if nilExpr.String() != "nil" {
+			continue
+		}
+		j.Errorf(expr, "foo")
+	}
+}
+
+func checkReturnNilWithErr(j *lint.Job) {
+	forAllFiles(j, func(n ast.Node) bool {
+		ifStmt, ok := n.(*ast.IfStmt)
+		if !ok {
+			return true
+		}
+		if !containsErrComparison(j, ifStmt.Cond) {
+			return true
+		}
+		return true
+	})
+}
+
[...]

PR #19426, opened 2017-10-22
server: use net.Pipe instead of TCP HTTP<->gRPC connections

### LINKED ISSUE #10374: server: admin UI communicates on server's default hostname only
### Steps to repro
1. Create node certificates not including server's hostname
2. Start node using `--advertise-host=<node internal IP address>`
3. Access admin UI.

### Expected Results
Admin UI loads data from cluster.

### Actual Result
Admin UI does not load data from cluster because it attempts to communicate using the server's `hostname` value as the host name.

tamird:
Fixes #10374.
Closes #17145.

### DIFF
--- a/pkg/rpc/context.go
+++ b/pkg/rpc/context.go
@@ -262,6 +262,56 @@ func (ctx *Context) removeConn(key string, meta *connMeta) {
 	}
 }
 
+// GRPCDialOptions returns the minimal `grpc.DialOption`s necessary to connect
+// to a server created with `NewServer`.
+//
+// At the time of writing, this is being used for making net.Pipe-based
+// connections, so only those options that affect semantics are included. In
+// particular, performance tuning options are omitted. Decompression is
+// necessarily included to support compression-enabled servers, and compression
+// is included for symmetry. These choices are admittedly subjective.
+func (ctx *Context) GRPCDialOptions() ([]grpc.DialOption, error) {
+	var dialOpts []grpc.DialOption
+	if ctx.Insecure {
+		dialOpts = append(dialOpts, grpc.WithInsecure())
+	} else {
+		tlsConfig, err := ctx.GetClientTLSConfig()
+		if err != nil {
+			return nil, err
+		}
+		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+	}
+
+	// The limiting factor for lowering the max message size is the fact
+	// that a single large kv can be sent over the network in one message.
+	// Our maximum kv size is unlimited, so we need this to be very large.
+	//
+	// TODO(peter,tamird): need tests before lowering.
+	dialOpts = append(dialOpts, grpc.WithDefaultCallOptions(
+		grpc.MaxCallRecvMsgSize(math.MaxInt32),
+		grpc.MaxCallSendMsgSize(math.MaxInt32),
+	))
+
+	dialOpts = append(dialOpts, grpc.WithDecompressor(snappyDecompressor{}))
+	// Compression is enabled separately from decompression to allow staged
+	// rollout.
+	if ctx.rpcCompression {
+		dialOpts = append(dialOpts, grpc.WithCompressor(snappyCompressor{}))
+	}
+
+	if tracer := ctx.AmbientCtx.Tracer; tracer != nil {
+		// We use a SpanInclusionFunc to circumvent the interceptor's work when
+		// tracing is disabled. Otherwise, the interceptor causes an increase in
+		// the number of packets (even with an empty context!). See #17177.
+		interceptor := otgrpc.OpenTracingClientInterceptor(
+			tracer,
+			otgrpc.IncludingSpans(otgrpc.SpanInclusionFunc(spanInclusionFunc)),
+		)
+		dialOpts = append(dialOpts, grpc.WithUnaryInterceptor(interceptor))
+	}
+	return dialOpts, nil
+}
+
 // GRPCDial calls grpc.Dial with the options appropriate for the context.
 func (ctx *Context) GRPCDial(target string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
 	value, ok := ctx.conns.Load(target)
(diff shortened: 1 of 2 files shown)


tschottdorf:
:lgtm:. I have a suggestion for the second commit that I think would be nice, but your change is an improvement even if you decide not to implement it (assuming it isn't flawed in the first place).

---

---

*[pkg/server/server.go, line 897 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx2yhPDb9yEZq5gRqOl:-Kx2yhPDb9yEZq5gRqOm:b59l4v8) ([raw file](https://github.com/cockroachdb/cockroach/blob/54f5bb9dcd0432969f22d49a20598115a8b2806c/pkg/server/server.go#L897)):*
> ```Go
> 
> 		if err := initServer.awaitBootstrap(); err != nil {
> 			return err
> ```

@bdarnell another case of https://github.com/cockroachdb/cockroach/issues/19316.

---

*[pkg/server/server.go, line 1128 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx30yHHJ7HEO0y4Irw_:-Kx30yHHJ7HEO0y4Irwa:bsgn0wz) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fd2786a54d6536764a92cb302db2385c948779/pkg/server/server.go#L1128)):*
> ```Go
> 
> 	s.stopper.RunWorker(workersCtx, func(context.Context) {
> 		netutil.FatalIfUnexpected(s.grpc.Serve(&singleListener{
> ```

Right now, if for whatever reason `c2`  got closed, things would be toast. What if you changed `singleListener` to always return the same conn, and additionally you wrap `c2` in a noop close method? That way, even if a the `rpcContext` decided that the connection was dead (say there's a clock jump, or a 5s gc pause) and closed it, the connection could be reestablished.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-:-Kx31ZxQx009TXvYl4Rw:bb9k0fg)*

tschottdorf:

---

*[pkg/server/server.go, line 1128 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx30yHHJ7HEO0y4Irw_:-Kx31y7eIxU3vK9BxSrI:b-dmkggj) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fd2786a54d6536764a92cb302db2385c948779/pkg/server/server.go#L1128)):*

Right now, if for whatever reason `c2`  got closed, things would be toast. What if you changed `singleListener` to always return the same conn, and additionally you wrap `c2` in a noop close method? That way, even if a the `rpcContext` decided that the connection was dead (say there's a clock jump, or a 5s gc pause) and closed it, the connection could be reestablished.
</blockquote></details>

Hmm, I guess that wouldn't change anything because you pass `conn`  to grpc-gateway and have no interface for reconnections in the first place.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426)*

tschottdorf:
Nice!

---

Reviewed 2 of 2 files at r3.

---

*[pkg/rpc/context.go, line 266 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx3HyC-6Vp3CbfTXTds:-Kx3HyC0oKKXCc8gNq4u:b9gug6h) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8d93b7c166626cac8dc2c630c97356affecddfe/pkg/rpc/context.go#L266)):*
> ```Go
> 
> // GRPCDialOptions returns the minimal GRPC dial options appropriate for the
> // context.
> ```

It might be confusing for future readers to decide what the set of "minimal" options is. In fact, I'm not so sure myself since it's largely driven by this specific change. Seems that you specifically exclude performance settings and timeouts but include tracing and tls. Mentioning both of these is a good start.

---

*[pkg/server/server.go, line 1138 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx3IklLBVdtxTnyBQQn:-Kx3IklLBVdtxTnyBQQo:b7v8cdm) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8d93b7c166626cac8dc2c630c97356affecddfe/pkg/server/server.go#L1138)):*
> ```Go
> 	}
> 
> 	conn, err := grpc.DialContext(ctx, s.cfg.AdvertiseAddr, append(
> ```

Comment here why you eschew the standard way of getting a ClientConn. (Avoid unnecessary moving parts on this connection, in particular, avoid having it go unhealthy since there's no way to recover it).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-:-Kx3J4ugANSaZwJ10LNy:b-83sjoh)*

tamird:
TFTR! CI failures were due to missing decompression options (server would compress, client wouldn't know how to decompress).

---

---

*[pkg/rpc/context.go, line 266 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx3HyC-6Vp3CbfTXTds:-Kx3qqOeDcxfgnDSw8Qv:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8d93b7c166626cac8dc2c630c97356affecddfe/pkg/rpc/context.go#L266)):*

It might be confusing for future readers to decide what the set of "minimal" options is. In fact, I'm not so sure myself since it's largely driven by this specific change. Seems that you specifically exclude performance settings and timeouts but include tracing and tls. Mentioning both of these is a good start.
</blockquote></details>

Done.

---

*[pkg/server/server.go, line 1128 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx30yHHJ7HEO0y4Irw_:-Kx3qs6-ONyhc2EZEmdP:b-kx42fa) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fd2786a54d6536764a92cb302db2385c948779/pkg/server/server.go#L1128)):*

Hmm, I guess that wouldn't change anything because you pass `conn`  to grpc-gateway and have no interface for reconnections in the first place.
</blockquote></details>

Yeah, nothing to do here unless we made `singleListener` into a channel and then had the dialer method below call `net.Pipe` every time. I think I prefer to never allow more than one of these in-process connections now that there should be no reason for it ever to be closed.

---

*[pkg/server/server.go, line 1138 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-Kx3IklLBVdtxTnyBQQn:-Kx3r4w_69vNE_cJz819:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d8d93b7c166626cac8dc2c630c97356affecddfe/pkg/server/server.go#L1138)):*

Comment here why you eschew the standard way of getting a ClientConn. (Avoid unnecessary moving parts on this connection, in particular, avoid having it go unhealthy since there's no way to recover it).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19426#-:-Kx3r5opLearv6P9o8u3:b-qbym7v)*

a-robinson:
Nice, thanks @tamird!

PR #19429, opened 2017-10-22
cli/sql: bump the go-libedit dependency to fix signal handling on darwin

### LINKED ISSUE #19132: cli: cli crashes when I resize my terminal
Using iterm2 on OSX, whenever I resize my terminal while using the CLI, the thing crashes. It spews a ton of `fatal: morestack on g0`, and, after a few pathetic seconds, `Illegal instruction: 4`.

Assigning to @knz in the hope that his libedit transition will magically solve this.

knz:
The details of the story can be found here:

https://github.com/knz/go-libedit/commit/5ffa3044c44237e9fc907fd3fc17e4cf604a2f12
(failed attempt: https://github.com/knz/go-libedit/commit/3b0f69cb0f672aad6a49ee224c3a801ae576a6e9)

This patch also tweaks the handling of the prompts, since go-libedit
1.5 doesn't do Go callbacks any more.

Fixes #19132.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -456,8 +456,8 @@
 [[projects]]
   name = "github.com/knz/go-libedit"
   packages = [".","common","other","unix"]
-  revision = "7577af3c24e9e2f3298da1358c4fa5e2329fd304"
-  version = "v1.4"
+  revision = "aafda877029e573c3bc5e352d000a41076e3d973"
+  version = "v1.6.1"
 
 [[projects]]
   branch = "master"
--- a/pkg/cli/sql.go
+++ b/pkg/cli/sql.go
@@ -595,11 +595,6 @@ func (c *cliState) GetCompletions(_ string) []string {
 	return nil
 }
 
-// GetLeftPrompt implements the readline.LeftPromptGenerator interface.
-func (c *cliState) GetLeftPrompt() string {
-	return c.currentPrompt
-}
-
 func (c *cliState) doStart(nextState cliStateEnum) cliStateEnum {
 	// Common initialization.
 	c.partialLines = []string{}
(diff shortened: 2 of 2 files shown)


bdarnell:
:lgtm:

```
  The second option seems "safer" (and also has the benefit
  of allowing the user to customize which key combo triggers
  the behavior) BUT doesn't work in practice: when libedit
  receives *another* signal, especially CONT (e.g. after
  a Ctrl+Z suspend), it will restore the terminal configuration
  to default and Ctrl+C will resume its standard behavior
  (send SIGINT). Thus Ctrl+C after a suspend will break things.
```

This sounds like a bug that should be fixed in libedit (or am I missing something?)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19429#-:-Kx9E4ZLp103VFev4rJ3:b-qt9mhj)*

knz:
libedit does *not* handle ctrl+c as keyboard input; I was adding the code to do so in go-libedit by using the tcsetattr syscall. From libedit's perspective, 1) it has complete control over tcsetattr 2) the terminal should handle special signal-inducing combos.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19429#-:-Kx9FzdP0l73HVLEp4NN:b-vou3ej)*

knz:
Wow I'm super bummed with what TC just helped me realized: SIGINT may not be delivered on the same pthread as the go-libedit code, so using longjmp from the handler to the libedit code confuses the Go runtime (that can move the goroutine from one pthread to another behind its back). Of course since my small toy test was just using 1 goroutine all was well. 

Gosh, signals are hard. I'll try something else.

benesch:
Can you forward the signal via `pthread_kill` to the thread that called `sigsetjmp`?

knz:
> Can you forward the signal via pthread_kill to the thread that called sigsetjmp?

Interesting idea. I'll try next weekend. One more iteration! :)

knz:
mistake

knz:
I think I got it on second try. https://github.com/knz/go-libedit/commit/5ffa3044c44237e9fc907fd3fc17e4cf604a2f12 

yay

bdarnell:
LGTM

knz:
Thanks! :tada:

PR #19436, opened 2017-10-22
storage: harden node liveness mechanism

spencerkimball:
- Introduce a new gossip key for self-report node liveness info. Each node
  optimistically gossips node liveness in advance of heartbeats. Nodes
  receiving self-reported gossip trust them, as long as they're not the node's
  own self-reported value. In other words, nodes trust other nodes' self
  reported liveness values and give them the benefit of the doubt before
  attempting to increment their epochs, but will only trust the official liveness
  range's gossip of liveness values when considering whether it is itself live.

- Don't increment the epoch of another node if current node is not live. This
  just seems like good policy.

### DIFF
--- a/pkg/gossip/keys.go
+++ b/pkg/gossip/keys.go
@@ -56,6 +56,10 @@ const (
 	// KeyNodeLivenessPrefix is the key prefix for gossiping node liveness info.
 	KeyNodeLivenessPrefix = "liveness"
 
+	// KeyNodeLivenessSelfReportedPrefix is the key prefix for nodes
+	// optimistically gossiping their expected node liveness info.
+	KeyNodeLivenessSelfReportedPrefix = "self-liveness"
+
 	// KeySentinel is a key for gossip which must not expire or
 	// else the node considers itself partitioned and will retry with
 	// bootstrap hosts.  The sentinel is gossiped by the node that holds
@@ -121,6 +125,12 @@ func NodeIDFromKey(key string) (roachpb.NodeID, error) {
 // MakeNodeLivenessKey returns the gossip key for node liveness info.
 func MakeNodeLivenessKey(nodeID roachpb.NodeID) string {
 	return MakeKey(KeyNodeLivenessPrefix, nodeID.String())
+}
+
+// MakeNodeLivenessSelfReportedKey returns the gossip key for
+// self-reported node liveness info.
+func MakeNodeLivenessSelfReportedKey(nodeID roachpb.NodeID) string {
+	return MakeKey(KeyNodeLivenessSelfReportedPrefix, nodeID.String())
 }
 
 // MakeStoreKey returns the gossip key for the given store.
--- a/pkg/storage/node_liveness.go
+++ b/pkg/storage/node_liveness.go
@@ -20,6 +20,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 
 	"github.com/cockroachdb/cockroach/pkg/base"
@@ -120,6 +121,7 @@ type NodeLiveness struct {
 	db                *client.DB
 	gossip            *gossip.Gossip
 	livenessThreshold time.Duration
+	renewalDuration   time.Duration
 	heartbeatInterval time.Duration
 	pauseHeartbeat    atomic.Value // contains a bool
 	selfSem           chan struct{}
@@ -152,6 +154,7 @@ func NewNodeLiveness(
 		db:                db,
 		gossip:            g,
 		livenessThreshold: livenessThreshold,
+		renewalDuration:   renewalDuration,
 		heartbeatInterval: livenessThreshold - renewalDuration,
 		selfSem:           make(chan struct{}, 1),
 		otherSem:          make(chan struct{}, 1),
@@ -169,6 +172,9 @@ func NewNodeLiveness(
 
 	livenessRegex := gossip.MakePrefixPattern(gossip.KeyNodeLivenessPrefix)
 	nl.gossip.RegisterCallback(livenessRegex, nl.livenessGossipUpdate)
+
+	livenessSelfReportedRegex := gossip.MakePrefixPattern(gossip.KeyNodeLivenessSelfReportedPrefix)
+	nl.gossip.RegisterCallback(livenessSelfReportedRegex, nl.livenessSelfReportedGossipUpdate)
 
 	return nl
 }
@@ -338,17 +344,16 @@ func (nl *NodeLiveness) StartHeartbeat(
 	stopper.RunWorker(ctx, func(context.Context) {
 		ambient := nl.ambientCtx
 		ambient.AddLogTag("hb", nil)
-		ticker := time.NewTicker(nl.heartbeatInterval)
-		defer ticker.Stop()
-		incrementEpoch := true
+		var timer timeutil.Timer
+		defer timer.Stop()
+		firstHeartbeat := true
 		for {
 			if !nl.pauseHeartbeat.Load().(bool) {
 				func() {
 					// Give the context a timeout approximately as long as the time we
 					// have left before our liveness entry expires.
-					ctx, cancel := context.WithTimeout(context.Background(), nl.livenessThreshold-nl.heartbeatInterval)
-					ctx, sp := ambient.AnnotateCtxWithSpan(ctx, "heartbeat")
-					defer cancel()
+					var sp opentracing.Span
+					ctx, sp = ambient.AnnotateCtxWithSpan(context.Background(), "heartbeat")
 					defer sp.Finish()
 
 					// Retry heartbeat in the event the conditional put fails.
@@ -357,21 +362,49 @@ func (nl *NodeLiveness) StartHeartbeat(
 						if err != nil && err != ErrNoLivenessRecord {
 							log.Errorf(ctx, "unexpected error getting liveness: %v", err)
 						}
-						if err := nl.heartbeatInternal(ctx, liveness, incrementEpoch); err != nil {
-							if err == ErrEpochIncremented {
-								log.Infof(ctx, "%s; retrying", err)
-								continue
+						// Self-report this liveness via gossip optimistically. This prevents
+						// other nodes from incrementing our epoch in the event that the
+						// heartbeat is too slow due to excessive load on the liveness range.
+						if err == nil && !firstHeartbeat {
+							key := gossip.MakeNodeLivenessSelfReportedKey(liveness.NodeID)
+							if err := nl.gossip.AddInfoProto(key, liveness, 0); err != nil {
+								log.Warningf(ctx, "failed to gossip self-reported node liveness (%+v): %v", liveness, err)
 							}
-							log.Warningf(ctx, "failed node liveness heartbeat: %v", err)
+						}
+						// Create a context with timeout for the actual node liveness.
+						ctxWithTimeout, cancel := context.WithTimeout(ctx, nl.renewalDuration)
+						err = nl.heartbeatInternal(ctxWithTimeout, liveness, firstHeartbeat /* incrementEpoch */)
+						cancel()
+						if err != nil {
+							if err == ErrEpochIncremented || err == context.Canceled {
+								log.Infof(ctx, "%s; retrying immediately", err)
+								r.Reset()
+							}
+							log.Warningf(ctx, "failed node liveness heartbeat (retrying): %v", err)
+							continue
 						} else {
-							incrementEpoch = false // don't increment epoch after first heartbeat
+							firstHeartbeat = false // don't increment epoch after first heartbeat
 						}
 						break
 					}
 				}()
 			}
+
+			// Compute our next heartbeat based on current expiration.
+			heartbeatIn := nl.heartbeatInterval
+			if liveness, err := nl.Self(); err != nil {
+				log.Errorf(ctx, "unexpected error getting liveness: %v", err)
+			} else {
+				expiresIn := time.Duration(liveness.Expiration.WallTime - nl.clock.Now().WallTime)
+				if expiresIn < nl.livenessThreshold {
+					heartbeatIn = expiresIn - nl.renewalDuration
+				}
+			}
+
+			timer.Reset(heartbeatIn)
 			select {
-			case <-ticker.C:
+			case <-timer.C:
+				timer.Read = true
 			case <-nl.triggerHeartbeat:
 			case <-stopper.ShouldStop():
 				return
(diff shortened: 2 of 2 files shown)


spencerkimball:
TODO: unittests. This is more of an experiment for the folks running into problems with liveness failures during heavy restore load. (+cc @mjibson @danhhz @dt).

dt:
cc @a-robinson

nvanbenschoten:
The code here looks good and the idea seems sound, but I'm not familiar enough with the subtleties of node liveness to make a ruling on the efficacy and safety of this change.

---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/storage/node_liveness.go, line 357 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19436#-Kx9Z8YEUWhm4PD-Kywk:-Kx9Z8YEUWhm4PD-Kywl:b-eb7v1v) ([raw file](https://github.com/cockroachdb/cockroach/blob/825dcf3b0110b4774f1ffdf31bd777b44cd2e16d/pkg/storage/node_liveness.go#L357)):*
> ```Go
> 						// Self-report this liveness via gossip optimistically. This prevents
> 						// other nodes from incrementing our epoch in the event that the
> 						// heartbeat is too slow due to excessive load on the liveness range.
> ```

Make a note here about why we don't do this if `firstHeartbeat`.

---

*[pkg/storage/node_liveness.go, line 358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19436#-Kx9ZcXimThYPBSuNWnq:-Kx9ZcXimThYPBSuNWnr:bhwwlp) ([raw file](https://github.com/cockroachdb/cockroach/blob/825dcf3b0110b4774f1ffdf31bd777b44cd2e16d/pkg/storage/node_liveness.go#L358)):*
> ```Go
> 						// other nodes from incrementing our epoch in the event that the
> 						// heartbeat is too slow due to excessive load on the liveness range.
> 						if err == nil && !firstHeartbeat {
> ```

Won't `err` always equal `ErrNoLivenessRecord` if `firstHeartbeat` is true? If so, do we need to maintain `firstHeartbeat` at all?

In other words, could we get rid of `firstHeartbeat` and do:
```
liveness, err := nl.Self()
if err != nil && err != ErrNoLivenessRecord {
    log.Errorf(ctx, "unexpected error getting liveness: %v", err)
}
// Self-report this liveness via gossip optimistically. This prevents
// other nodes from incrementing our epoch in the event that the
// heartbeat is too slow due to excessive load on the liveness range.
if err == nil {
    key := gossip.MakeNodeLivenessSelfReportedKey(liveness.NodeID)
    if err := nl.gossip.AddInfoProto(key, liveness, 0); err != nil {
        log.Warningf(ctx, "failed to gossip self-reported node liveness (%+v): %v", liveness, err)
    }
}
if err := nl.heartbeatInternal(ctx, liveness, err == ErrNoLivenessRecord /* incrementEpoch */); err != nil {
    if err == ErrEpochIncremented {
        log.Infof(ctx, "%s; retrying", err)
        continue
    }
    log.Warningf(ctx, "failed node liveness heartbeat: %v", err)
}
break
```

---

*[pkg/storage/node_liveness.go, line 785 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19436#-Kx9emjJCw-zU9wf-WNM:-Kx9emjJCw-zU9wf-WNN:b31xhm0) ([raw file](https://github.com/cockroachdb/cockroach/blob/825dcf3b0110b4774f1ffdf31bd777b44cd2e16d/pkg/storage/node_liveness.go#L785)):*
> ```Go
> 	// benefit of the doubt. We must never do so with our own, which we
> 	// use as a guarantee of our ability to act based on epoch range
> 	// leases.
> ```

What could go wrong if we updated our own liveness record?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19436#-:-Kx9etat6qVc7XPvIkOC:b-eza92j)*

dt:
Sadly around 7 hours in to a TPC-H SF=100 lineitem IMPORT job, it looks like my 3 node, local ssd cluster running 825dcf has taken the ~usual turn for the worse: logs mentioning node liveness heartbeats failing, slow commits, etc, admin UI dead, can't read system tables. Eventually the OS seems to have OOM killed node 1. Initial I restarted it, but them remembered IMPORT isn't reusable so I just nuked it. 

[2.cockroach.stderr.txt](https://github.com/cockroachdb/cockroach/files/1413181/2.cockroach.stderr.txt)
[3.cockroach.stderr.txt](https://github.com/cockroachdb/cockroach/files/1413182/3.cockroach.stderr.txt)
[1.cockroach.stderr.txt](https://github.com/cockroachdb/cockroach/files/1413183/1.cockroach.stderr.txt)

spencerkimball:
That's too bad. I'm going to create a new PR to add logging which will helpfully shed light on what's happening in this restore case.

a-robinson:
@dt what are the repro steps for running an IMPORT like that?

dt:
I started a new cluster, then ran:
```
SET CLUSTER SETTING experimental.importcsv.enabled = true;
CREATE DATABASE tpch100;
SET DATABASE tpch100;
IMPORT TABLE lineitem
CREATE USING 'gs://cockroach-fixtures/tpch-csv/schema/lineitem.sql'
CSV DATA (
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.1',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.2',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.3',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.4',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.5',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.6',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.7',
    'gs://cockroach-fixtures/tpch-csv/sf-100/lineitem.tbl.8'
) WITH distributed, delimiter = '|', temp = 'gs://cockroach-tmp/csvtest';
```

These should be readable from GCE VMs in the cockroach orgs or by employees who gcloud auth logic. To write to the last one, it seems GCE VMs need to be created with the read/write scope, rather than the current readonly, or you can gcloud auth login on each of them. attempting to fix that now though.

(anyone who doesn't have access to those buckets: csv data is just tpc-h `dbgen`, sf 100, chunked 8 ways, and the temp is just anything with read and write perms)

pkg/storage/replica_range_lease.go
```diff
@@ -194,6 +194,14 @@ func (p *pendingLeaseRequest) requestLeaseAsync(
 					}
 				} else if status.Liveness.Epoch == *status.Lease.Epoch {
 					// If not owner, increment epoch if necessary to invalidate lease.
+					// However, we only do so in the event that the next leaseholder is
+					// considered live at this time. If not, there's no sense in
+					// incrementing the expired leaseholder's epoch.
+					if live, liveErr := repl.store.cfg.NodeLiveness.IsLive(nextLeaseHolder.NodeID); !live || liveErr != nil {
```
a-robinson: I like this idea, but it looks like you failed to actually skip incrementing the epoch if this condition is true. I'm rerunning a couple experiments and have fixed this logic up locally to actually return here.

spencerkimball:
Oh nice catch. The just desserts from not writing unittests. We might want
to consider rerunning the restore test with this fix. +@dt

a-robinson:
No need for @dt to do so, I'm currently running two tests on GCE using his instructions. One against master at 9c10b690e5f1e0d1f61c41eaaf9bf09ca8418a55 (as a control group) and one against 9c10b690e5f1e0d1f61c41eaaf9bf09ca8418a55 plus https://github.com/a-robinson/cockroach/commit/7ea7cbdc0543ca4282caa7ad39eebaac99a2a939 and https://github.com/a-robinson/cockroach/commit/a0df8d8166d6fdb4f30ab8c40b66476ff8506c49

dt:
@a-robinson oops, i neglected this point, but: if they're running concurrently, did you happen to think to point them at unique `temp` paths?

a-robinson:
Uh, nope. Should I restart one of them to use a different temp directory before they start writing to it?

dt:
@a-robinson hm. if you cared about the data they were producing, yes. As it is, one or both will probably fail at some point, but maybe not until they go to write the final descriptor? if they get to that point, they're probably doing better than usual

a-robinson:
The experiments on Friday didn't go so well. Node liveness still had problems, and the OOMs (#19635) made it so that I only even managed to get logs off 3 of the 6 machines (2 from the experimental cluster, 1 from the control group).

From what I saw, though, there aren't any small tweaks to the specifics of how our existing node liveness works that will be immune from breaking during one of these jobs. Not with slow disk writes like this popping up in the logs:

```
W171028 05:42:50.850398 131 storage/engine/rocksdb.go:1495  batch [1/51/0] commit took 5m50.353594272s (>500ms)
```

Something broader will be needed, whether that means not touching disk for liveness heartbeats or limiting the rate of bulk input. However, even if we get node liveness under control by not relying on the disk, writes this slow mean that an import job will effectively prevent interactive workloads from using the database while the import is happening.

a-robinson:
Also, @dt, is there a way in future experiments to skip all the pre-processing phase and just get straight to the point where the job starts writing to disk? The pre-processing phase took more than 9 hours, but the failures happened within minutes after the disk writing started.

petermattis:
@a-robinson Wow, that's a long commit. Did you see the discussion in https://github.com/cockroachdb/cockroach/pull/19618? Can we try tweaking the sysctl settings mentioned in that PR:

```
~ sudo sysctl vm.dirty_bytes=$[4 * 1024 * 1024]
~ sudo sysctl vm.dirty_background_bytes=$[1 * 1024 * 1024]
```

a-robinson:
Yeah, I'll try that out today.

a-robinson:
Liveness heartbeats still slowed down quite a bit with those settings on the VMs (enough that at least one heartbeat timed out before nodes started OOMing), but not quite as egregiously as on Friday. This was running against the same two additional commits as on Friday (https://github.com/cockroachdb/cockroach/pull/19436#issuecomment-340057144). It's certainly a good lead, even if it may not be the ultimate solution.

a-robinson:
I came back around to retry the sysctl experiment using VMs with more memory, inspired by the effect the sysctls had on the testing in https://github.com/cockroachdb/cockroach/issues/19699#issuecomment-342947639. With VMs big enough to avoid OOM'ing (I used the n1-highmem-8 machine type) and the sysctls set as recommended above, the SF-10 restore completed just fine. 99th percentile latencies sync latencies never even reached 450ms, and 99th percentile KV latencies topped out at 1.3s. As mentioned in #19699, modifying these sysctls does seem to be the best solution so far.

spencerkimball:
@a-robinson take a look at this. I saw some node liveness failures locally when creating huge transactions and I took another look at this PR with fresh eyes and realized there were a couple of big bugs in the logic. In particular:
- When the attempt to update the liveness failed with a context timeout (I believe this is the usual error when we have long disk syncs), we would not try again until waiting exactly another heartbeat interval – more than long enough for the liveness record to have its epoch incremented. This wasn't allowing us enough time to get back in there and re-gossip the self-reported liveness.
- We were just using a simple ticker for the heartbeat, when in fact the update to the node liveness can take several seconds, even if successful. This changes the ticker to a timer and does a better job of deciding when to update the liveness record again.

spencerkimball:
@dt, @a-robinson, @mjibson are there still big jobs we have that can cause spurious node liveness failures that we can easily test this on?

a-robinson:
> When the attempt to update the liveness failed with a context timeout (I believe this is the usual error when we have long disk syncs), we would not try again until waiting exactly another heartbeat interval – more than long enough for the liveness record to have its epoch incremented. This wasn't allowing us enough time to either try again to heartbeat our record before being preempted via an epoch increment, or to re-gossip the self-reported liveness (the whole point of this PR).

That was true when we were using the old `heartbeatInterval / 2` context deadline, but is that still true if we're setting the context timeout to when the liveness record expires, as we were roughly doing as of 6b0c5898db8dc6aa5c2f2485d541f7eff049e9bf and are doing more precisely after your change from `Ticker` to `Timer`? The context won't time out until the liveness record has expired, at which point the next `Tick` or `Timer` event would already be coming immediately.

> We were just using a simple ticker for the heartbeat, when in fact the update to the node liveness can take several seconds, even if successful. This changes the ticker to a timer and does a better job of deciding when to update the liveness record again.

I agree that the new version is a little more accurate, but I don't think the old version was far off. Whenever a tick happened, we would heartbeat with a new expiration time of `Now()+nl.livenessThreshold + maxOffset`. The time between the tick and `Now()` being called should be negligibly small, so the new expiration should be roughly `nl.heartbeatInterval` after the tick even if the heartbeat took a few seconds to succeed. Thus, when next tick happens, it should be about `nl.livenessThreshold-nl.heartbeatInterval` before the liveness record expires. Since `nl.livenessThreshold-nl.heartbeatInterval == nl.heartbeatInterval`, the next tick will come just as our deadline expires.

The new version would certainly be easier to reason about if we tweaked things so that `nl.heartbeatInterval` wasn't just half of `nl.livenessThreshold`, though. I'm fine with changing it over.

> @dt, @a-robinson, @mjibson are there still big jobs we have that can cause spurious node liveness failures that we can easily test this on?

None that I know of, but I still haven't been able to run big (tpc-h sf-100 or larger) distributed imports due to the memory issues they were having. You could always just revert https://github.com/cockroachdb/cockroach/pull/20352/files or https://github.com/cockroachdb/cockroach/pull/20442 and run https://github.com/cockroachdb/cockroach/pull/19436#issuecomment-339669076 if you're feeling really ambitious.

spencerkimball:
@a-robinson I agree on the point you're making about the ticker, but you're not taking into consideration the old retry loop and how we exit it on cancelation. Here's the common scenario:
- Node 1's ticker fires, initiating heartbeat attempt
- New context created with timeout for 4.5s in the future
- Retry loop starts, sends first heartbeat to liveness leader which is going very slowly
- Context cancelation occurs at 4.5s. Because this isn't an epoch incremented error, we break out of retry loop
- Wait in the select for another 4.5s for ticker
- At this point, other nodes are free to increment node 1's liveness record – liveness hasn't been heartbeat, and we haven't gossiped our self-reported liveness in a full 9s either! This process could continue indefinitely.

The context cancelations simply aren't being handled appropriately.

a-robinson:
Why are we waiting another 4.5 seconds for the ticker after the context cancellation? The ticker will tick immediately because it's been 4.5 seconds since its last tick, triggering an immediate retry.

The ticker ticks every 4.5 seconds based on when the last tick was, not based on when we start waiting for the tick. Take [this small example](https://play.golang.org/p/I9ns1VAyS-L) -- if we sleep for the tick duration after a tick, the next time we try to pull off the channel there's a tick there immediately, not after another tick duration.

spencerkimball:
For some reason I was thinking the heartbeat interval would be every 9s. You're right then – doesn't seem like there's much of a problem here to fix. Disregard this PR then. I'm going to close it.

PR #19437, opened 2017-10-23
ui: Add Query Manager Reducer

mrtracy:
Adds the Query Manager Reducer, which is designed to to hold metadata
state for asynchronous queries. For each query, this state includes:
+ Whether the query is currently running.
+ If the last attempt to run the query encountered an error.
+ The time at which the query was last attempted.

Currently, this metadata is being stored in the Admin UI right alongside
the data itself, and each query is responsible for managing its own
execution and metadata. This commit does not change this, but creates a
reducer that will be used to store it after further refactoring - the
next step will be to create a Saga that executes queries, using this
reducer to store query metadata.

### DIFF
--- a/pkg/ui/src/redux/queryManager/reducer.spec.ts
+++ b/pkg/ui/src/redux/queryManager/reducer.spec.ts
@@ -1,0 +1,114 @@
+import { assert } from "chai";
+import moment from "moment";
+
+import {
+  managedQueryReducer,
+  ManagedQueryState,
+  queryBegin,
+  queryComplete,
+  queryError,
+  queryManagerReducer,
+  QueryManagerState,
+} from "./reducer";
+
+describe("Query Manager State", function () {
+  describe("managed query reducer", function () {
+    const testMoment = moment();
+    const testError = new Error("err");
+    let state: ManagedQueryState;
+
+    beforeEach(function() {
+      state = managedQueryReducer(undefined, {} as any);
+    });
+
+    it("has the correct initial state", function () {
+      assert.deepEqual(state, new ManagedQueryState());
+    });
+
+    it("dispatches queryBegin correctly", function () {
+      // We expect "isRunning" to be true and all other fields to be null.
+      const expected = new ManagedQueryState();
+      expected.isRunning = true;
+      expected.lastError = null;
+      expected.completedAt = null;
+
+      state = managedQueryReducer(state, queryBegin("ID"));
+      assert.deepEqual(state, expected);
+    });
+
+    it("dispatches queryError correctly", function () {
+      // We expect "isRunning" to be false; both the error field and completedAt
+      // should be populated with the supplied information from the action.
+      const expected = new ManagedQueryState();
+      expected.isRunning = false;
+      expected.lastError = testError;
+      expected.completedAt = testMoment;
+
+      state = managedQueryReducer(state, queryBegin("ID"));
+      state = managedQueryReducer(state, queryError("ID", testError, testMoment));
+      assert.deepEqual(state, expected);
+    });
+
+    it("dispatches queryComplete correctly", function () {
+      // We expect "isRunning" to be false, completedAt to be populated, and
+      // the error field to be null.
+      const expected = new ManagedQueryState();
+      expected.isRunning = false;
+      expected.lastError = null;
+      expected.completedAt = testMoment;
+
+      state = managedQueryReducer(state, queryBegin("ID"));
+      state = managedQueryReducer(state, queryComplete("ID", testMoment));
+      assert.deepEqual(state, expected);
+    });
+
+    it("clears error on queryBegin", function () {
+      const expected = new ManagedQueryState();
+      expected.isRunning = true;
+      expected.lastError = null;
+      expected.completedAt = null;
+
+      state = managedQueryReducer(state, queryError("ID", testError, testMoment));
+      state = managedQueryReducer(state, queryBegin("ID"));
+      assert.deepEqual(state, expected);
+    });
+
+    it("ignores unrecognized actions", function () {
+      const origState = state;
+      state = managedQueryReducer(state, { type: "unsupported" } as any);
+      assert.equal(state, origState);
+    });
+  });
+
+  describe("query manager reducer", function () {
+    const testMoment = moment();
+    const testError = new Error("err");
+    let state: QueryManagerState;
+
+    beforeEach(function() {
+      state = queryManagerReducer(undefined, {} as any);
+    });
+
+    it("has the correct initial value", function () {
+      assert.deepEqual(state, {});
+    });
+
+    it("correctly dispatches based on ID", function () {
+      const expected = {
+        "1": managedQueryReducer(undefined, queryBegin("1")),
+        "2": managedQueryReducer(undefined, queryError("2", testError, testMoment)),
+        "3": managedQueryReducer(undefined, queryComplete("3", testMoment)),
+      };
+
+      state = queryManagerReducer(state, queryBegin("1"));
+      state = queryManagerReducer(state, queryBegin("2"));
+      state = queryManagerReducer(state, queryBegin("3"));
+      state = queryManagerReducer(state, queryError("2", testError, testMoment));
+      state = queryManagerReducer(state, queryError("3", testError, testMoment));
+      state = queryManagerReducer(state, queryBegin("3"));
+      state = queryManagerReducer(state, queryComplete("3", testMoment));
+
+      assert.deepEqual(state, expected);
+    });
+  });
+});
--- a/pkg/ui/src/redux/queryManager/reducer.ts
+++ b/pkg/ui/src/redux/queryManager/reducer.ts
@@ -1,0 +1,149 @@
+import moment from "moment";
+import { Action } from "redux";
+
+import nextState from "src/util/nextState";
+
+const QUERY_BEGIN = "cockroachui/queries/QUERY_BEGIN";
+const QUERY_ERROR = "cockroachui/queries/QUERY_ERROR";
+const QUERY_COMPLETE = "cockroachui/queries/QUERY_COMPLETE";
+
+interface QueryBeginAction extends Action {
+    type: typeof QUERY_BEGIN;
+    payload: {
+        id: string;
+    };
+}
+
+interface QueryErrorAction extends Action {
+    type: typeof QUERY_ERROR;
+    payload: {
+        id: string;
+        error: Error;
+        timestamp: moment.Moment;
+    };
+}
+
+interface QueryCompleteAction extends Action {
+    type: typeof QUERY_COMPLETE;
+    payload: {
+        id: string;
+        timestamp: moment.Moment;
+    };
+}
+
+type QueryAction = QueryBeginAction | QueryErrorAction | QueryCompleteAction;
+
+/**
+ * queryBegin is dispatched by the query manager whenever the query with the
+ * given ID has started execution.
+ */
+export function queryBegin(id: string): QueryBeginAction {
+    return {
+        type: QUERY_BEGIN,
+        payload: {
+            id,
+        },
+    };
+}
+
+/**
+ * queryError is dispatched by the query manager whenever the query with the
+ * given ID has stopped due to an error condition. This action contains the
+ * returned error, along with the timestamp when the error was received.
+ */
+export function queryError(id: string, error: Error, timestamp: moment.Moment): QueryErrorAction {
+    return {
+        type: QUERY_ERROR,
+        payload: {
+            id,
+            error,
+            timestamp,
+        },
+    };
+}
+
+/**
+ * queryComplete is dispatched by the query manager whenever the query with the
+ * given ID has completed successfully. It includes the timestamp when the query
+ * was completed.
+ */
+export function queryComplete(id: string, timestamp: moment.Moment): QueryCompleteAction {
+    return {
+        type: QUERY_COMPLETE,
+        payload: {
+            id,
+            timestamp,
+        },
+    };
+}
+
+/**
+ * ManagedQueryState maintains the current state for a single managed query.
+ */
+export class ManagedQueryState {
+    // True if this query is currently running asynchronously.
+    isRunning = false;
+    // If the previous attempt to run this query ended with an error, this field
+    // contains that error.
+    lastError: Error = null;
+    // Contains the timestamp when the query last compeleted, regardless of
+    // whether it succeeded or encountered an error.
+    completedAt: moment.Moment = null;
+}
+
+/**
+ * managedQueryReducer reduces actions for a single managed query.
+ */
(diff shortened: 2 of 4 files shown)


couchand:
My only real beef is that the action fields should be nested under `payload`.  Other than that, I left a number of questions/nits.

---

Reviewed 4 of 4 files at r1.

---

*[pkg/ui/src/redux/state.ts, line 13 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8epNvKAlYsYGCXCdv:-Kx8epNwPNVCcctYX111:b-wx1iun) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/state.ts#L13)):*
> ```XML
> import { timeWindowReducer, TimeWindowState } from "./timewindow";
> import { uiDataReducer, UIDataState } from "./uiData";
> import { queryManagerReducer, QueryManagerState } from "./queryManager/reducer";
> ```

alphabetize please?

---

*[pkg/ui/src/redux/queryManager/reducer.spec.ts, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8lGYlf1zfZ_YITFcq:-Kx8lGYlf1zfZ_YITFcr:bxdgvx) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.spec.ts#L28)):*
> ```XML
>     });
> 
>     it("dispatches queryBegin", function () {
> ```

it might be nice to say a few words about what's expected to happen in each case.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 14 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8fAZDs3Hh7WLNF085:-Kx8fAZEnOQD6yYaiWis:b-sj9e5s) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L14)):*
> ```XML
> }
> 
> interface QueryErrorAction extends Action {
> ```

just curious why the first action is a `class` and the second and third are `interface`s.    seems like `interface` is better (less runtime overhead?) but I'm not sure I know all the tradeoffs.  they should at least be consistent (unless there's some reason for them not to be, of course).

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 16 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8iFOn3TjO5K_fgCfJ:-Kx8iFOn3TjO5K_fgCfK:blxnq2k) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L16)):*
> ```XML
> interface QueryErrorAction extends Action {
>     type: typeof QUERY_ERROR;
>     id: string;
> ```

nest these fields under a `payload` key?

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 78 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8gHaTn0I1u88mVlLP:-Kx8gHaTn0I1u88mVlLQ:b-p7x70m) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L78)):*
> ```XML
>     // Contains the timestamp when the query last compeleted, regardless of
>     // whether it succeeded or encountered an error.
>     queriedAt: moment.Moment = null;
> ```

the name sort of sounds like it's when then query was initiated.  not sure i have a better idea or if it really matters, just thought i'd mention it.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8h04NHAQHUPS_0nIa:-Kx8h04OG05oO-7uTam3:b-3kuwoe) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L87)):*
> ```XML
>     state = new ManagedQueryState(), action: QueryAction,
> ): ManagedQueryState {
>     if (_.isNil(action)) {
> ```

this check shouldn't be necessary, right?  the `default` case below is enough?

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 93 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8hBYI1ufvUFU1WkqS:-Kx8hBYI1ufvUFU1WkqT:b-yfgerv) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L93)):*
> ```XML
>     switch (action.type) {
>         case QUERY_BEGIN:
>             state = _.clone(state);
> ```

spread?  or for the first two cases, since they don't maintain anything from the old state, maybe just
```
return {
  isRunning: true,
  lastError: null,
  queriedAt: null
}
```
?

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 128 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8i5rGyIGKau7VSmYz:-Kx8i5rH5Hw_M2T-WwFu:bc4tysj) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L128)):*
> ```XML
>     state: QueryManagerState = {}, action: QueryAction,
> ): QueryManagerState {
>     if (_.isNil(action)) {
> ```

same comment about this not being needed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-:-Kx8m58NGolqqskYnG0m:b6cqsuy)*

mrtracy:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/ui/src/redux/state.ts, line 13 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8epNvKAlYsYGCXCdv:-Kx99neLxySAxdVpcFsq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/state.ts#L13)):*

alphabetize please?
</blockquote></details>

Done.

---

*[pkg/ui/src/redux/queryManager/reducer.spec.ts, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8lGYlf1zfZ_YITFcq:-Kx9MlsztvHVHDI7UPdv:b-mjcpy3) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.spec.ts#L28)):*

it might be nice to say a few words about what's expected to happen in each case.
</blockquote></details>

Added some information to these "dispatches" tests; added in comment form, a bit too much information for the test titles.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 14 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8fAZDs3Hh7WLNF085:-Kx99oxbbE2ESCU5_iD8:b-gmby37) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L14)):*

just curious why the first action is a `class` and the second and third are `interface`s.    seems like `interface` is better (less runtime overhead?) but I'm not sure I know all the tradeoffs.  they should at least be consistent (unless there's some reason for them not to be, of course).
</blockquote></details>

Hah, no reason, just an artifact of development. Switched all of them to interface.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 16 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8iFOn3TjO5K_fgCfJ:-Kx99vGwKxGo8NgyYQAP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L16)):*

nest these fields under a `payload` key?
</blockquote></details>

Done.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 78 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8gHaTn0I1u88mVlLP:-Kx9LUpvHLtx0HwNJ6aZ:b-s1q7r4) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L78)):*

the name sort of sounds like it's when then query was initiated.  not sure i have a better idea or if it really matters, just thought i'd mention it.
</blockquote></details>

After thinking about it again, I agree; changed to "completedAt".

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8h04NHAQHUPS_0nIa:-Kx99zf6j6KOqMEkui-C:b-63sndt) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L87)):*

this check shouldn't be necessary, right?  the `default` case below is enough?
</blockquote></details>

Well, if the action passed is null or undefined then it will blow up at `switch(action.type)`, "cannot read property 'type' of undefined".

However, it appears that redux itself never passes "null" or "undefined" for the action; i'm wondering where I got the impression that this could happen.  Maybe only in our tests?

I'm going to remove this check and use `{}` as the initializing action in tests.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 93 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8hBYI1ufvUFU1WkqS:-Kx9CIUJ7CA6OZ3Du8RO:bl2nipa) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L93)):*

spread?  or for the first two cases, since they don't maintain anything from the old state, maybe just
```
return {
  isRunning: true,
  lastError: null,
  queriedAt: null
}
```
?
</blockquote></details>

I was using the spread operator at first, but I discovered something unfortunate: the type checking is less stringent.  Specifically, the spread operator will *not* check for superfluous fields in the same way that direct object initialization will.

That is, you could write the following:
```typescript
return {
   ...state,
   isRunning: true,
   error: null,
   queriedAt: null,
}
```

And the type system would happily let you do this, without letting you know that field `error` actually doesn't exist on the type of state and you made a typo.

What typescript is missing is an concept of explicit "Exact Types", which would disallow superfluous fields; typescript does enforce exact types when assigning an object literal to a typed variable, but it does not enforce this when assigning existing objects to a variable.  Spread operators are *not* considered to be object literals by typescript, so they do not disallow superfluous fields.

However, you have inspired me to create a utility function `nextState` which allows us to do type-safe assignment.

```typescript
return nextState(state, {
  isRunning: true,
  error: null,  // will not compile because this field does not exist
  queriedAt: null,
);
```

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 128 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8i5rGyIGKau7VSmYz:-Kx9LJvD56RzdlSOBKJ0:bo4btz7) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L128)):*

same comment about this not being needed?
</blockquote></details>

Addressed above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437)*

couchand:
:lgtm:

---

Reviewed 2 of 5 files at r2.
Review status: 2 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/ui/src/redux/queryManager/reducer.ts, line 93 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx8hBYI1ufvUFU1WkqS:-Kx9dLGC6ctZ0UBFINGD:bix2ftn) ([raw file](https://github.com/cockroachdb/cockroach/blob/68c53ca780f355a59d76f379e7ef3e009ca11f5f/pkg/ui/src/redux/queryManager/reducer.ts#L93)):*

I was using the spread operator at first, but I discovered something unfortunate: the type checking is less stringent.  Specifically, the spread operator will *not* check for superfluous fields in the same way that direct object initialization will.

That is, you could write the following:
```typescript
return {
   ...state,
   isRunning: true,
   error: null,
   queriedAt: null,
}
```

And the type system would happily let you do this, without letting you know that field `error` actually doesn't exist on the type of state and you made a typo.

What typescript is missing is an concept of explicit "Exact Types", which would disallow superfluous fields; typescript does enforce exact types when assigning an object literal to a typed variable, but it does not enforce this when assigning existing objects to a variable.  Spread operators are *not* considered to be object literals by typescript, so they do not disallow superfluous fields.

However, you have inspired me to create a utility function `nextState` which allows us to do type-safe assignment.

```typescript
return nextState(state, {
  isRunning: true,
  error: null,  // will not compile because this field does not exist
  queriedAt: null,
);
```
</blockquote></details>

:+1::+1: nice.  too bad about that spread, but this is a great workaround.  I actually seem to remember a colleague at my last gig complaining to me about this issue, but I didn't actually understand what he was saying until now.

---

*[pkg/ui/src/util/nextState.ts, line 22 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-Kx9dtKWlMAtT0xmUs4U:-Kx9dtKX1tilwTvBhXM7:bzypgo) ([raw file](https://github.com/cockroachdb/cockroach/blob/8ddf5fc39214e65cbd0f74913e5549919477cf1d/pkg/ui/src/util/nextState.ts#L22)):*
> ```XML
>  * overwriting fields that exist on the type of state. In the examples above,
>  * using the spread operator would compile even if "prop1" was not a field of
>  * state's type. This is an explicit design choice of typescript.
> ```

:frowning_face: boo typescript

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19437#-:-Kx9e7mQs-0H0Hotwk8D:bnfp4nl)*

PR #19441, opened 2017-10-23
storage: factor out replicaStateLoader

tbg:
This is one of the self-contained threads to unravel for #18784.

The resulting code allows the `StateLoader` struct to be moved into
its own package (but since moves rot quickly, I'm going to merge this
first and then execute the move separately).


petermattis:
:lgtm:

---

Review status: 0 of 11 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19441#-:-Kx8Y-VS0dQx-fszNAVn:bnfp4nl)*

PR #19446, opened 2017-10-23
util: Replace errors with pgerror for IPAddr parsing functions

lgo:
Previously, helper functions for parsing IPAddr datums did not generate
pgerrors.


lego:
Just replaced the other instances and added the file to the style lint :)

justinj:
LGTM! I would say that we shouldn't bother tagging internal errors not originating from the package, but I don't really care.

Issue #16004, opened 2017-05-17
storage: *errors.errorString: replica.go:1351 on-disk and in-memory state diverged:


tschottdorf:
This happened on a 1.0 binary, so it's neither #15819 nor the more recently #15935.

https://sentry.io/cockroach-labs/cockroachdb/issues/269875826/

```
*errors.errorString: replica.go:1351 on-disk and in-memory state diverged:
%s
  File "github.com/cockroachdb/cockroach/pkg/storage/replica.go", line 1351, in assertStateRLocked
  File "github.com/cockroachdb/cockroach/pkg/storage/replica.go", line 1336, in assertState
  File "github.com/cockroachdb/cockroach/pkg/storage/replica_proposal.go", line 792, in handleEvalResult
  File "github.com/cockroachdb/cockroach/pkg/storage/replica.go", line 3755, in processRaftCommand
  File "github.com/cockroachdb/cockroach/pkg/storage/replica.go", line 2886, in handleRaftReadyRaftMuLocked
...
(5 additional frame(s) were not displayed)

replica.go:1351 on-disk and in-memory state diverged:
%s
```

tamird:
This looks like a sentry misconfiguration. The actual SHA
is 0ba0af9ed345ecd645efb4743b8d3100a2a78955, which is #15819.

On Wed, May 17, 2017 at 4:06 PM, Tobias Schottdorf <notifications@github.com

tschottdorf:
Ah, well spotted. Any idea what the misconfiguration is?

petermattis:
@tamird Can you elaborate on why you think this is a sentry misconfiguration? I'm seeing the `rev` match the `v1.0` SHA. What are you looking at?

a-robinson:
Where are you seeing that? I do see a couple occurrences at that commit, but also one at 43aa1ed04cd817348921ac27e11e25725f634aa6 with release=v1.0

tschottdorf:
![image](https://cloud.githubusercontent.com/assets/5076964/26174229/ab3aa7c8-3b1c-11e7-9313-571191940ea5.png)

on the right hand side, around the middle.

tamird:
@a-robinson but 43aa1ed is not the 1.0 SHA! It's on the release branch, but it's not actually 1.0.

[...]

PR #19447, opened 2017-10-23
storage: report failed raft command application

tbg:
Previously, if `applyRaftCommand` returned an error, it would mark the replica
as corrupt but then go on and execute the side effects and potentially the
assertions in `assertState()`. These were then likely to fail and return a
misleading error, as likely seen in #16004.

Instead, cause a fatal error right when observing the error, and potentially
capture the root cause on sentry.io. The (perhaps too optimisic) expectation
is that after accounting these disk corruption/space errors, there will be
much fewer (possibly no) reports triggered by `assertState()`.

Touches #16004.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -4450,9 +4450,25 @@ func (r *Replica) processRaftCommand(
 		}
 
 		var delta enginepb.MVCCStats
-		delta, pErr = r.applyRaftCommand(
-			ctx, idKey, raftCmd.ReplicatedEvalResult, raftIndex, leaseIndex, writeBatch)
-		raftCmd.ReplicatedEvalResult.Delta = delta.ToNetworkStats()
+		{
+			var err error
+			delta, err = r.applyRaftCommand(
+				ctx, idKey, raftCmd.ReplicatedEvalResult, raftIndex, leaseIndex, writeBatch)
+			raftCmd.ReplicatedEvalResult.Delta = delta.ToNetworkStats()
+
+			// applyRaftCommand returned an error, which usually indicates
+			// either a serious logic bug in CockroachDB or a disk
+			// corruption/out-of-space issue. Make sure that these fail with
+			// descriptive message so that we can differentiate the root causes.
+			if err != nil {
+				log.Errorf(ctx, "unable to update the state machine: %s", err)
+				// Report the fatal error separately and only with the error, as that
+				// triggers an optimization for which we directly report the error to
+				// sentry (which in turn allows sentry to distinguish different error
+				// types).
+				log.Fatal(ctx, err)
+			}
+		}
 
 		if filter := r.store.cfg.TestingKnobs.TestingPostApplyFilter; pErr == nil && filter != nil {
 			pErr = filter(storagebase.ApplyFilterArgs{
@@ -4463,6 +4479,9 @@ func (r *Replica) processRaftCommand(
 			})
 		}
 
+		// calling maybeSetCorrupt here is mostly for tests and looks. The
+		// interesting errors originate in applyRaftCommand, and they are
+		// already handled above.
 		pErr = r.maybeSetCorrupt(ctx, pErr)
 		if pErr == nil {
 			pErr = forcedErr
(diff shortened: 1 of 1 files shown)


nvanbenschoten:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/storage/replica.go, line 4464 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447#-Kx8uiLW1uEbs1MXSBIs:-Kx8uiLW1uEbs1MXSBIt:b95gte1) ([raw file](https://github.com/cockroachdb/cockroach/blob/2613077f41fca134abe1945358b24126f125655f/pkg/storage/replica.go#L4464)):*
> ```Go
> 			// descriptive message so that we can differentiate the root causes.
> 			if err != nil {
> 				log.Errorf(ctx, "unable to update the state machine: %s", err)
> ```

Why add the `log.Error` before the `log.Fatal`? Is there a difference between them that I'm not aware of?

---

*[pkg/storage/replica.go, line 4481 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447#-Kx8uxxtq3LC52EW-1fT:-Kx8uxxtq3LC52EW-1fU:bvxhue4) ([raw file](https://github.com/cockroachdb/cockroach/blob/2613077f41fca134abe1945358b24126f125655f/pkg/storage/replica.go#L4481)):*
> ```Go
> 		// interesting errors originate in applyRaftCommand, and they are
> 		// already handled above.
> 		pErr = r.maybeSetCorrupt(ctx, pErr)
> ```

What's the deal with this? Can we delete it? Can it be moved into the `TestingPostApplyFilter` block?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447)*

tschottdorf:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/storage/replica.go, line 4464 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447#-Kx8uiLW1uEbs1MXSBIs:-Kx8wGzVzFZuKG5dZWPe:b-ur7cdn) ([raw file](https://github.com/cockroachdb/cockroach/blob/2613077f41fca134abe1945358b24126f125655f/pkg/storage/replica.go#L4464)):*
Oops, yeah, that's subtle. Added comment:

> // Report the fatal error separately and only with the error, as that
// triggers an optimization for which we directly report the error to
// sentry (which in turn allows sentry to distinguish different error
// types).

---

*[pkg/storage/replica.go, line 4481 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447#-Kx8uxxtq3LC52EW-1fT:-Kx8w0U8sIGG7S7SDdz7:b-ltrar5) ([raw file](https://github.com/cockroachdb/cockroach/blob/2613077f41fca134abe1945358b24126f125655f/pkg/storage/replica.go#L4481)):*

What's the deal with this? Can we delete it? Can it be moved into the `TestingPostApplyFilter` block?
</blockquote></details>

The replica corruption handling has been around for a while, it's pretty half-assed but has enough pull that I don't want to rip it out during cherry-picking season (especially since we'll want something like it at some point, but not quite yet, so it's not clear whether to keep or not). I'd rather leave it for now and definitely in this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447)*

bdarnell:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447#-:-Kx9ULRBymhrx92_nxiJ:bnfp4nl)*

nvanbenschoten:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19447#-:-Kx9Xd98ykPM37KrRhhB:bnfp4nl)*

PR #19448, opened 2017-10-23
storage: return error from split/merge lock acquisition

tbg:
See #19172.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -4383,7 +4383,22 @@ func (r *Replica) processRaftCommand(
 	} else {
 		log.Event(ctx, "applying command")
 
-		if splitMergeUnlock := r.maybeAcquireSplitMergeLock(ctx, raftCmd); splitMergeUnlock != nil {
+		if splitMergeUnlock, err := r.maybeAcquireSplitMergeLock(ctx, raftCmd); err != nil {
+			log.Eventf(ctx, "unable to acquire split lock: %s", err)
+			// Send a crash report because a former bug in the error handling might have
+			// been the root cause of #19172.
+			_ = r.store.stopper.RunAsyncTask(ctx, "crash report", func(ctx context.Context) {
+				log.SendCrashReport(
+					ctx,
+					&r.store.cfg.Settings.SV,
+					0, // depth
+					"while acquiring split lock: %s",
+					[]interface{}{err},
+				)
+			})
+
+			forcedErr = roachpb.NewError(err)
+		} else if splitMergeUnlock != nil {
 			// Close over raftCmd to capture its value at execution time; we clear
 			// ReplicatedEvalResult on certain errors.
 			defer func() {
@@ -4519,21 +4534,21 @@ func (r *Replica) processRaftCommand(
 // applying the command to perform any necessary cleanup.
 func (r *Replica) maybeAcquireSplitMergeLock(
 	ctx context.Context, raftCmd storagebase.RaftCommand,
-) func(storagebase.ReplicatedEvalResult) {
+) (func(storagebase.ReplicatedEvalResult), error) {
 	if split := raftCmd.ReplicatedEvalResult.Split; split != nil {
 		return r.acquireSplitLock(ctx, &split.SplitTrigger)
 	} else if merge := raftCmd.ReplicatedEvalResult.Merge; merge != nil {
 		return r.acquireMergeLock(&merge.MergeTrigger)
 	}
-	return nil
+	return nil, nil
 }
 
 func (r *Replica) acquireSplitLock(
 	ctx context.Context, split *roachpb.SplitTrigger,
-) func(storagebase.ReplicatedEvalResult) {
+) (func(storagebase.ReplicatedEvalResult), error) {
 	rightRng, created, err := r.store.getOrCreateReplica(ctx, split.RightDesc.RangeID, 0, nil)
 	if err != nil {
-		return nil
+		return nil, err
 	}
 
 	// It would be nice to assert that rightRng is not initialized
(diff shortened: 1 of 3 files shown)


nvanbenschoten:

---

*[pkg/storage/store.go, line 3878 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-Kx8tLwxcq3-FndmLybh:-Kx8tLwxcq3-FndmLybi:b-m9lntu) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8190f4daf9aaec039a68ad1cb1685b811e3f928/pkg/storage/store.go#L3878)):*
> ```Go
> 			//
> 			// Inspired by https://github.com/cockroachdb/cockroach/pull/19404.
> 			runtime.Gosched()
> ```

`Gosched` can hurt performance, especially in the tail latencies, so I wouldn't say there's no harm. That said, this shouldn't be very common so I doubt it will have an impact. Still, I'm not sure it's worth the risk if there's no reason to add this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448)*

tschottdorf:

---

*[pkg/storage/store.go, line 3878 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-Kx8tLwxcq3-FndmLybh:-Kx8uKFepvV3pCbl26cO:b-scwx4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8190f4daf9aaec039a68ad1cb1685b811e3f928/pkg/storage/store.go#L3878)):*

`Gosched` can hurt performance, especially in the tail latencies, so I wouldn't say there's no harm. That said, this shouldn't be very common so I doubt it will have an impact. Still, I'm not sure it's worth the risk if there's no reason to add this.
</blockquote></details>

Note that the `Gosched` is on the `err = errRetry` path, so I doubt it matters for performance. And, with a loop like this, I can't *prove* that there's no reason to add this, and unless we can't, I think we should generally add it to avoid this kind of freak deadlock.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448)*

petermattis:
Reviewed 1 of 1 files at r1.

---

*[pkg/storage/replica.go, line 4536 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-Kx8uONGgb5rjF-8pRVi:-Kx8uONGgb5rjF-8pRVj:b-8ttlle) ([raw file](https://github.com/cockroachdb/cockroach/blob/fefcc3b66c586b2e8b72fb4f772c6212f266e43c/pkg/storage/replica.go#L4536)):*
> ```Go
> 	rightRng, created, err := r.store.getOrCreateReplica(ctx, split.RightDesc.RangeID, 0, nil)
> 	if err != nil {
> 		log.Fatalf(ctx, "while acquiring split lock: %s", err)
> ```

Shouldn't you return the error to the caller in order to force error the Raft command? The code in `Store.tryGetOrCreateReplica` has enough error paths that I can't convince myself none of them will occur.

---

*[pkg/storage/store.go, line 3878 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-Kx8twc_G8CXF13T1XrO:-Kx8twc_G8CXF13T1XrP:b-6xgx27) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8190f4daf9aaec039a68ad1cb1685b811e3f928/pkg/storage/store.go#L3878)):*
> ```Go
> 			//
> 			// Inspired by https://github.com/cockroachdb/cockroach/pull/19404.
> 			runtime.Gosched()
> ```

Was this addition motivated by a real concern. `Store.tryGetOrCreateReplica` performs multiple lock operations which will definitely provide stop-the-world safe points for the GC.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448)*

tschottdorf:

---

*[pkg/storage/replica.go, line 4536 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-Kx8uONGgb5rjF-8pRVi:-Kx8yXoZpy7_ya7-oiV3:b6kfnye) ([raw file](https://github.com/cockroachdb/cockroach/blob/fefcc3b66c586b2e8b72fb4f772c6212f266e43c/pkg/storage/replica.go#L4536)):*

Shouldn't you return the error to the caller in order to force error the Raft command? The code in `Store.tryGetOrCreateReplica` has enough error paths that I can't convince myself none of them will occur.
</blockquote></details>

Good point. I definitely want to learn about the errors out there in the wild though, so I added reporting to sentry. PTAL!

---

*[pkg/storage/store.go, line 3878 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-Kx8twc_G8CXF13T1XrO:-Kx9-jydXb9aDfLYXTCk:bf660ad) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8190f4daf9aaec039a68ad1cb1685b811e3f928/pkg/storage/store.go#L3878)):*

Was this addition motivated by a real concern. `Store.tryGetOrCreateReplica` performs multiple lock operations which will definitely provide stop-the-world safe points for the GC.
</blockquote></details>

No, just paranoia given the recent deadlock in TestRefreshPendingCmds. You're right that the abundance of lock acquisition makes this too paranoid. Removed the commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448)*

bdarnell:
:lgtm:

---

Reviewed 2 of 4 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-:-Kx9W9cU6zQIByBTaDGa:bnfp4nl)*

nvanbenschoten:
LGTM, but the PR title is incorrect now.

petermattis:
:lgtm:

---

Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19448#-:-Kx9XTmEM3QMPE5djDK6:bnfp4nl)*

tschottdorf:
TFTRs, everyone! @nvanbenschoten changed the title.

nvanbenschoten:
I stumbled upon this PR while cleaning up #38954 and am now confused about why setting `forcedErr` to the result of `maybeAcquireSplitMergeLock` is correct. We don't expect to ever see an error here, but if we do, I don't think we can silently let it prevent one of the replicas from applying the split/merge. That's the kind of thing that causes replica inconsistencies. I think the only option we have is to fatal.

tbg:
Not sure what we were thinking there, `Fatal` looks like the only option.

PR #19456, opened 2017-10-23
ui: refactor updateLinkedGuidelines to put hover state in the store

couchand:
Rather than imperatively updating the linked guidelines on mouse move,
keep the hover state in the Redux store and declaratively update the
linked guidelines when rendering charts.

This is a refactor to support the upcoming work replacing NVD3 with
vanilla D3 charting.

### DIFF
--- a/pkg/ui/src/redux/hover.ts
+++ b/pkg/ui/src/redux/hover.ts
@@ -1,0 +1,63 @@
+/**
+ * Monitors the currently hovered chart and point in time.
+ */
+
+import moment from "moment";
+import { Action } from "redux";
+
+import { PayloadAction } from "src/interfaces/action";
+import { AdminUIState } from "src/redux/state";
+
+export const HOVER_ON = "cockroachui/hover/HOVER_ON";
+export const HOVER_OFF = "cockroachui/hover/HOVER_OFF";
+
+/**
+ * HoverInfo is conveys the current hover position to the state.
+ */
+export interface HoverInfo {
+  hoverChart: string;
+  hoverTime: moment.Moment;
+}
+
+export class HoverState {
+  // Are we currently hovering over a chart?
+  currentlyHovering = false;
+  // Which chart are we hovering over?
+  hoverChart: string;
+  // What point in time are we hovering over?
+  hoverTime: moment.Moment;
+}
+
+export function hoverReducer(state = new HoverState(), action: Action): HoverState {
+  switch (action.type) {
+    case HOVER_ON:
+      const { payload: hi } = action as PayloadAction<HoverInfo>;
+      return {
+        currentlyHovering: true,
+        hoverChart: hi.hoverChart,
+        hoverTime: hi.hoverTime,
+      };
+    case HOVER_OFF:
+      return new HoverState();
+    default:
+      return state;
+  }
+}
+
+export function hoverOn(hi: HoverInfo): PayloadAction<HoverInfo> {
+  return {
+    type: HOVER_ON,
+    payload: hi,
+  };
+}
+
+export function hoverOff(): Action {
+  return {
+    type: HOVER_OFF,
+  };
+}
+
+/**
+ * Are we currently hovering, and if so, which chart and when?
+ */
+export const hoverStateSelector = (state: AdminUIState) => state.hover;
--- a/pkg/ui/src/redux/state.ts
+++ b/pkg/ui/src/redux/state.ts
@@ -6,6 +6,7 @@ import createSagaMiddleware from "redux-saga";
 import thunk from "redux-thunk";
 
 import { apiReducersReducer, APIReducersState } from "./apiReducers";
+import { hoverReducer, HoverState } from "./hover";
 import { localSettingsReducer, LocalSettingsState } from "./localsettings";
 import { metricsReducer, MetricsState, queryMetricsSaga } from "./metrics";
 import { queryManagerReducer, QueryManagerState } from "./queryManager/reducer";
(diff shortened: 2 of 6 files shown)


mrtracy:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/ui/src/redux/hover.ts, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-KxJJm6xu71qNRSxwvrt:-KxJJm6xu71qNRSxwvru:b-l3wwkp) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad681114a712d9e88fbb9eb11041855acab01d9a/pkg/ui/src/redux/hover.ts#L32)):*
> ```XML
> 
>   constructor() {
>     this.currentlyHovering = false;
> ```

Since this is a class definition, you can initialize this inline above:

```typescript
export class HoverState {
  currentlyHovering = false;
```

and drop the constructor entirely.

---

*[pkg/ui/src/redux/hover.ts, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-KxJJTrhOA3MQROoqcxM:-KxJJTrhOA3MQROoqcxN:bxb4z94) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad681114a712d9e88fbb9eb11041855acab01d9a/pkg/ui/src/redux/hover.ts#L40)):*
> ```XML
>     case HOVER_ON:
>       const { payload: hi } = action as PayloadAction<HoverInfo>;
>       state = _.clone(state);
> ```

`nextState()` is checked in now, we should start using that everywhere.

---

*[pkg/ui/src/views/cluster/containers/nodeGraphs/index.tsx, line 186 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-KxJM17Kev2SdS7o4i3i:-KxJM17Kev2SdS7o4i3j:bkyoucp) ([raw file](https://github.com/cockroachdb/cockroach/blob/ad681114a712d9e88fbb9eb11041855acab01d9a/pkg/ui/src/views/cluster/containers/nodeGraphs/index.tsx#L186)):*
> ```XML
>       return (
>         <div key={key}>
>           <MetricsDataProvider id={key} hoverOn={hoverOn} hoverOff={hoverOff} hoverState={hoverState}>
> ```

This should not be on <etricsDataProvider; MDP is only responsible for getting data and passing it to its children, it has no concept of being rendered or whatnot.

Instead, do the same thing that MetricsDataProvider does to pass additional props: use cloneElement.

```Typescript
<MetricsDataProvider id={key}>
  { React.cloneElement(graph. { hoverOn, hoverOff, hoverState }) }
</MetricsDataProvider>
```

Then you won't need to modify MetricsDataProvider (or its associated interfaces).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456)*

couchand:
Taken care of the below comments.  Before merging I need to take care of the bisect TODO (which shouldn't take too long) and squash the commits.

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/ui/src/redux/hover.ts, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-KxJJm6xu71qNRSxwvrt:-KxJV_McRGRBvBITwOz2:b-7rl0jd) ([raw file](https://github.com/cockroachdb/cockroach/blob/56b47555afeded41b24119a256c28838f4896f9c/pkg/ui/src/redux/hover.ts#L32)):*

Since this is a class definition, you can initialize this inline above:

```typescript
export class HoverState {
  currentlyHovering = false;
```

and drop the constructor entirely.
</blockquote></details>

done.

---

*[pkg/ui/src/redux/hover.ts, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-KxJJTrhOA3MQROoqcxM:-KxJVaouhyINZJ41lMXI:b7zip1c) ([raw file](https://github.com/cockroachdb/cockroach/blob/56b47555afeded41b24119a256c28838f4896f9c/pkg/ui/src/redux/hover.ts#L40)):*

`nextState()` is checked in now, we should start using that everywhere.
</blockquote></details>

Looking at this again, since it completely replaces `state`, I just returned the new object instead.

---

*[pkg/ui/src/views/cluster/containers/nodeGraphs/index.tsx, line 186 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-KxJM17Kev2SdS7o4i3i:-KxJVqxA26TdK9-wxA2b:bjxwasb) ([raw file](https://github.com/cockroachdb/cockroach/blob/56b47555afeded41b24119a256c28838f4896f9c/pkg/ui/src/views/cluster/containers/nodeGraphs/index.tsx#L186)):*

This should not be on <etricsDataProvider; MDP is only responsible for getting data and passing it to its children, it has no concept of being rendered or whatnot.

Instead, do the same thing that MetricsDataProvider does to pass additional props: use cloneElement.

```Typescript
<MetricsDataProvider id={key}>
  { React.cloneElement(graph. { hoverOn, hoverOff, hoverState }) }
</MetricsDataProvider>
```

Then you won't need to modify MetricsDataProvider (or its associated interfaces).
</blockquote></details>

I didn't think `MetricsDataProvider` was the right place for this, but I didn't want to introduce a new component for it.  For some reason it didn't occur to me to just inline the usage like that.  Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19456#-:-KxJaZdy6Q3cUlovuzsu:blsrla0)*

couchand:
OK I think this is ready to merge, PTAL @mrtracy @vilterp

mrtracy (approved):

couchand:
TFTR!

PR #19458, opened 2017-10-23
roachpb: provide a fast path for updating observed timestamps

spencerkimball:
For single node requests generated via `loadgen/kv` and resent to
`Store.Send`, this improves performance by ~5%.

### DIFF
--- a/pkg/roachpb/data.go
+++ b/pkg/roachpb/data.go
@@ -938,6 +938,17 @@ func (t *Transaction) ResetObservedTimestamps() {
 // operations in the transaction. When multiple calls are made for a single
 // nodeID, the lowest timestamp prevails.
 func (t *Transaction) UpdateObservedTimestamp(nodeID NodeID, maxTS hlc.Timestamp) {
+	// Fast path optimization for either no observed timestamps or
+	// exactly one, for the same nodeID as we're updating.
+	if l := len(t.ObservedTimestamps); l == 0 {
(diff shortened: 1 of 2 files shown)


nvanbenschoten:
:lgtm: nice improvement!

---

Reviewed 2 of 2 files at r1.

---

*[pkg/roachpb/data.go, line 943 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19458#-Kx9g3r-qzNScZhRRQTh:-Kx9g3r-qzNScZhRRQTi:b110bio) ([raw file](https://github.com/cockroachdb/cockroach/blob/e7e71fffcd6dde94697acecee14d079b414c27e7/pkg/roachpb/data.go#L943)):*
> ```Go
> 	// Fast path optimization for either no observed timestamps or
> 	// exactly one, for the same nodeID as we're updating.
> 	if l := len(t.ObservedTimestamps); l == 0 {
> ```

nit: `switch len(t.ObservedTimestamps) { ...` would be cleaner.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19458#-:-Kx9h5rRISyM4DeZO2a4:b-j5h9d)*

petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19458#-:-Kx9ifIejDyrXZMiQIay:bnfp4nl)*

tbg (approved):

PR #19463, opened 2017-10-23
WIP: Add job control (cancel/pause/resume) to schema changes

### LINKED ISSUE #16018: sql: pause/unpause/cancel schema changes
We need a way to pause/unpause a schema change when it is hogging resources on a cluster.

It is possible that someone can use the API we provide to build an external scheduler that pauses/unpauses the schema change based on resource demands. Although we should really make pausing/unpausing internal and automatic.

We also need a mechanism to cancel a schema change. Some schema changes take days to execute (schema change of a 500M row table) and if someone inadvertently runs a schema change
thinking it will take minutes, they might like to cancel the schema change.

We have a jobs table that is shared between backup/restore schema change and other future job types. The SQL commands and the admin UI interface should be shared across these job types and can be operations on the common jobs table. Only the underlying implementation need be different. @benesch so he is aware.

lgo:
(moving the first commit to #19496)

This has a big WIP tag as this PR is to hash out the specifics of moving schema changers to jobs. Will close #16018 when complete.


pkg/ui/src/util/eventTypes.ts
```diff
@@ -29,6 +29,8 @@ export const DROP_VIEW = "drop_view";
 export const REVERSE_SCHEMA_CHANGE = "reverse_schema_change";
 // Recorded when a previously initiated schema change has completed.
 export const FINISH_SCHEMA_CHANGE = "finish_schema_change";
+// Recorded when a previously initiated schema change has completed.
```
couchand: copy-pasta

pkg/ui/src/util/eventTypes.ts
```diff
@@ -44,7 +46,8 @@ export const SET_CLUSTER_SETTING = "set_cluster_setting";
 export const nodeEvents = [NODE_JOIN, NODE_RESTART, NODE_DECOMMISSIONED, NODE_RECOMMISSIONED];
 export const databaseEvents = [CREATE_DATABASE, DROP_DATABASE];
 export const tableEvents = [CREATE_TABLE, DROP_TABLE, ALTER_TABLE, CREATE_INDEX,
-  DROP_INDEX, CREATE_VIEW, DROP_VIEW, REVERSE_SCHEMA_CHANGE, FINISH_SCHEMA_CHANGE];
+  DROP_INDEX, CREATE_VIEW, DROP_VIEW, REVERSE_SCHEMA_CHANGE, FINISH_SCHEMA_CHANGE,
+  FINISH_SCHEMA_CHANGE_ROLLBACK];
```
couchand: ugh, I don't like this formatting but w/e don't worry about it

lego: Mind that I haven't linted (and this may undoubtedly not even compile). -manyellsatcloud-

couchand: no I think we just have permissive linting for this case

vivekmenezes:
Can you please split this into two PRs. The first being,  Refactor the schema changer to use jobs

lego:
Sorry pre-maturely made this PR then realized a few suggestions weren't correct, so I was re-writting the plan for doing this (it's incoming)

Also I don't think there's a way to break up this change into multiple functioning PRs. Removing the schema change manager relies on schema change jobs being resumable. Removing the schema change lease also relies on schema change jobs being resumable (and happens for free). The rest is just about adding job control, and I don't think that can be done paritally for this.

lego:
Sorry, I realized we can add cancelation relatively simply but the rest of the work is chunked up with managing pause/resume.

lego:
(Ignore the code, it was WIP and I realized several flaws as I've written this plan.)

@vivekmenezes @benesch I'd like some eyes on this plan of how the schema changes will work with jobs. I'll outline cases that come up and how things differ between them.

## Creating a schema change and running it synchronously (simple case)
- A schema change is created, which is always done so during a transaction.
  - When adding the mutations to the table descriptor, we also create a new job for the schema change. This jobID is recorded into `TableDescriptor.mutationJobs.JobID`. (currently done)
  - Initialized jobs will be `pending`.
- After the transaction has finished, due to either rollback or commit, schema change execution begins.
  - We go through the schema changes in FIFO order, executing them until they finish.


## Creating a schema change, running it synchronously, and canceling midway
- After the transaction has finished, due to either rollback or commit, schema change execution begins.
  - We will go through the schema changes in FIFO order, executing them.
  - If a schema change is canceled, a `SchemaChangeControlError` will be returned[0]. 
  - In `SchemaChanger.exec` we catch this error and clean up the mutations.
  - We continue executing other schema changes in the transaction.


## Creating a schema change, running it synchronously, and pausing midway
- After the transaction has finished, due to either rollback or commit, schema change execution begins.
  - We will go through the schema changes in FIFO order, executing them.
  - If a schema change is paused, a `SchemaChangeControlError` will be returned[0]. 
  - In `SchemaChanger.exec` we catch this error and push the mutations to the end of the mutation queue.
  - Regular execution continues, where we go through other schema changes in FIFO order. We don't run the paused schema change. Even if it is later unpaused, the job registry handels this,.
  - NB: If any later schema changes depend on this one, they will fail. Furthermore, will this impact queries in the same transaction? (I don't think they will)

Alternatively, we can block the transaction completion until the job is unpaused. However if we also don't push back mutations, this implies that no schema changes may happen until we pause the job.


## Creating a schema change and quiting the transaction before schema changes complete
- After the transaction has finished, due to either rollback or commit, schema change execution begins.
  - We will go through the schema changes in FIFO order, executing them.
  - The session is closed, so our schema change is stopped mid-way, with the latest progress recorded in the job details.
  - Before exiting, we need to remove all the schema change jobs for the session from the local node's job registry to indicate the jobs are no longer running. cc: @benesch there is no way to do this currently. Thoughts on adding an interface to do this?

    ```go
    // SignalJobExecutionStopped removes the job from the local registry and releases the job lease.
    func (r *Registry) SignalJobExecutionStopped(ctx context.Context, jobID int64) error
    ```

    Furthermore, are we able to run this on session closure in the `executor`, or does the routine just exit silently?

A problem we run into is when jobs are resumed. They will be automatically started on nodes using the resume hook, but we'll run into two situations:
- The schema change is the next mutation. When the job is adopted by another node and resuming, it can continue until compeletion.
- The schema change is not the next mutation. When the job is adopted and called with the resume hook (even from `pending`), we need a way wait until the schema change is the next mutation in-line. A possible solution is:
  - We can keep sleeping until we are. Alternatively this routine can listen to gossip updates of the table descriptor to check when it's next. Notably, the job will appear as `running` during this. (Will this work? Do we need to combine it with a sleep loop like the SchemaChangeManager?)
  - NB: This means if we run 1000 schema changes, and close the session (i.e. background them), we will have 1000 jobs with active, channel blocked routines. This seems less than ideal... We can maybe pause the routines, but we may not want the user to see `paused`.

Another problem is that if a schema change is paused, and later cancelled, we need to still clean up mutations. cc: @benesch currently the `cancelFn` is only for liveness cancellations. I propose a `CleanupHook` to execute on the cancelation of a job that is not actively running.


### Alternative: only have one active schema change job per table
When the schema changes are resumed, we can take an approach where only one schema change may be in the `running` state per `TableDescriptor`. Instead of the the pending schema change jobs sleeping until they are next in-line, the finishing schema change job will start the next one when it completes. This requires a few changes to jobs:
- We will need an option to not have nodes adopt a `pending` job.
- We need the ability to "start" a pending job, e.g.
  ```go
  nextJobID := ...
  registry.BeginPendingJob(nextJobID)
  ```

  (Whether this happens via. `ResumeHook`, in the current routine, or as a separate hook is TBD. All work for the purposes fo this.)


[0]: `SchemaChangeControlError` will be used to identify if the error was job control (pause/cancel), and will be annotated as such.

lego:
Also some further things to think about:
- How do we cancel a destructive schema change? e.g. `DROP (INDEX|TABLE)`/`ALTER TABLE ... DROP COLUMN`. This might need some more logic to `jobs` for a hook like `CanCancel` :/

vivekmenezes:
Probably worth talking about this in person. Let's chat tomorrow.

benesch:
@vivekmenezes and @jordanlewis are the schema change experts, so I can only provide feedback on a few things:

> Before exiting, we need to remove all the schema change jobs for the session from the local node's job registry to indicate the jobs are no longer running. cc: @benesch there is no way to do this currently. Thoughts on adding an interface to do this?

This is actually what `job.FinishedWith` is for. You can teach it to handle a "context canceled" error, or whatever error you get when the session closes, the same way it currently handles `InvalidStatusError`: by unregistering the job.

> The schema change is not the next mutation. When the job is adopted and called with the resume hook (even from pending), we need a way wait until the schema change is the next mutation in-line.

I think you should just prevent the job from being adopted in this case. You could plumb a `canAdopt` callback through `Job.adopt`, which would validate that the job ID is next in line on the table descriptor.

> Another problem is that if a schema change is paused, and later cancelled, we need to still clean up mutations. cc: @benesch currently the cancelFn is only for liveness cancellations. I propose a CleanupHook to execute on the cancelation of a job that is not actively running.

Yerp, seems reasonable. CC @mjibson who's working on something similar for `RESTORE`. I'd advocate for CleanupHook to run unconditionally, regardless of whether the job is currently running or not. It also needs to run *in the transaction* where the job is marked as canceled. You either want to launch the rollback and mark the old job as canceled, or do neither; that atomicity is particularly important.

I'm also getting the sense that we could remove some of the mutation jobs FIFO pain by building dependent jobs. It wouldn't be that hard to give each job a `preceding_job_id` field.

On a meta-level, this is an incredibly hard task; I'd be careful when trying to estimate how long this will take.

lego:
Sidenote: should we split this discussion into another issue for tracking this discussion and changes to jobs? I didn't see an existing issue.

> This is actually what job.FinishedWith is for. You can teach it to handle a "context canceled" error, or whatever error you get when the session closes, the same way it currently handles InvalidStatusError: by unregistering the job.

Perfect! I'm a bit worried about how other uses of `FinishedWith` may be affected or if this behaviour is a bit too implicit.

> I think you should just prevent the job from being adopted in this case. You could plumb a canAdopt callback through Job.adopt, which would validate that the job ID is next in line on the table descriptor.

Unless I'm wrong, this would require a CanAdoptHook to configure this and then plumb it through.

> Yerp, seems reasonable. CC @mjibson who's working on something similar for RESTORE. I'd advocate for CleanupHook to run unconditionally, regardless of whether the job is currently running or not. It also needs to run in the transaction where the job is marked as canceled. You either want to launch the rollback and mark the old job as canceled, or do neither; that atomicity is particularly important.

:+1:.

> I'm also getting the sense that we could remove some of the mutation jobs FIFO pain by building dependent jobs. It wouldn't be that hard to give each job a preceding_job_id field.

While I agree it may simplify things it unfortunately doesn't solve the fact that some code paths are running the jobs synchronously. These jobs will also have preceding job. An alternative would be queuing all the schema change jobs in the background and blocking on their completion -- but there's no obvious way to do this.

I'll think about this one a bit more.

benesch:
> Perfect! I'm a bit worried about how other uses of FinishedWith may be affected or if this behaviour is a bit too implicit.

My guess is we're not handling the case of a client hangup in backup/restore properly, when we should be. So I think you'd actually improve behavior elsewhere. (Yes, the code overall is terrible, I'll grant you that.)

> Unless I'm wrong, this would require a CanAdoptHook to configure the callback for each job type, then plumb that one through.

Yup, that's exactly what I'm thinking. The resume hook can return a resumeFn and a canAdoptFn. Or a struct with two function callbacks, or a struct that implements an interface. Whatever's clean. This isn't a hot code path, fortunately.

> While I agree it may simplify things it unfortunately doesn't solve the fact that some code paths are running the jobs synchronously. These jobs will also have preceding job. An alternative would be queuing all the schema change jobs in the background and blocking on their completion -- but there's no obvious way to do this.

They're only synchronous as long as the node that started the jobs is alive, right? You can plumb a channel into the job if it has a SQL session attached, and nonblockingly signal the channel when the job completes.

> There's also an edge case when we start a rollback. The new rollback job will have to modify any jobs depending on the original schema change job to now depend on the rollback job.

Those dependent jobs should just be canceled, no?

lego:
> Those dependent jobs should just be canceled, no?

Dependant here loosely refers to the order of schema changes, not an actual schema dependency. In this situation, we’re just pushing a job onto the head of some chain of dependent jobs, as opposed every other case where we would want to add a new job to the end. I don't see it being a particular problem, but a weird quirk when thinking about how `preceding_job_id` would be used here.

benesch:
Ah, I see. Perhaps we need the concept of an order-only prerequisite and a must-succeed prerequisite.

lego:
Closing, as I won't get around to this change anymore :'( and the branch isn't really relevant now after the job refactors. I've made a summary of what's needed to finish this in the feature issue, https://github.com/cockroachdb/cockroach/issues/16018#issuecomment-351185206.

Issue #17819, opened 2017-08-22
storageccl: AddSSTable complains about mismatched end key


dt:
running tpch line item SF 10 import:
```
addsstable [/Table/59/1/11018023/3/0,/Table/59/1/11036803/2/0/NULL):
verifying sstable data: key /Table/59/1/11024007/2/0 not in request range
[/Table/59/1/11018023/3/0,/Table/59/1/11024007/2)
```

Generated/Imported using this runbook: https://docs.google.com/document/d/18Gls6c3dyUuoiUc1zP6QnBp52U7hWiviB54n0E_oO5s/edit

dt:
Some investigation so far: this is the verify call done by addSSTable, the start/end are computed by for the batch, not passed in by the csv code so, uh... scary?

Going to add a log of `batchEndKey` to `sstBatcher.Finish` and try to repro again.

danhhz:
I suspect this is going to be exceptionally hard to repro. Leaving the issue open but don't think there's anything to be done for 1.1

dt:
@mjibson just saw this again in #19399:
```
importing 324 ranges: addsstable [/Table/56/9/2/20/289690669019889664/0,/Table/56/9/3/12/289690135848779776/0/NULL): verifying sstable data: key /Table/56/9/2/63/289688124228730880/0 not in request range [/Table/56/9/2/20/289690669019889664/0,/Table/56/9/2/63/289688124228730880)
```

dt:
so reading though the code again, the bounds printed in the `addsstable` prefix of the message are the start and end from `evalImport`'s batcher, which are what get passed to the `AddSSTable` call, and I would expect to match what we then pass to `verifySSTable` in `evalAddSSTable`, where we print them into the error again. The fact that the end key has changed, and is smaller, really _looks_ like the request got split, [i.e. by `pkg/kv.truncate`](https://github.com/cockroachdb/cockroach/blob/master/pkg/kv/batch.go#L84).

dt:
@tschottdorf Does it actually make sense to split an AddSSTableRequest? off the top of my head, that seems wrong, unless process the sst to only include the updated range? IIRC, we don't do that though?

tschottdorf:
I think you're correct in that it only makes sense if we also clip the key range that is actually ingested. Alternatively, you could add capability to return errors from `truncate` and have the import machinery track the key ranges that are required through these errors. The former looks preferable: though it sends a bit more over the network, it seems more "straightforward".

[...]

PR #19465, opened 2017-10-23
storageccl: add a little detail to debug logs/errors

dt:
trying to track down #17819


tbg (approved):

dt:
added another which I think should be interesting, https://github.com/cockroachdb/cockroach/pull/19465/files#diff-ac8932565749dcc4723ee0df6a3ca145R293

madelynnblue (approved):

Issue #18525, opened 2017-09-16
Is the column "table_catalog" missing in the table "information_schema.table_constraints"?


Vincent-Pang:
According to the SQL spec ISO/IEC 9075:1999 (sorry for that I don't have any newer spec),
section "21.40 TABLE_CONSTRAINTS base table", the following columns are missing in cockroachdb:

TABLE_CATALOG
IS_DEFERRABLE
INITIALLY_DEFERRED

FYI:
The reason that I concern with the column "table_catalog" is because I am using an ORM lib [typeorm](https://github.com/typeorm/typeorm) with typescript. When it try to synchronize my schema classes with the db, it will perform sth like
```
SELECT xxx FROM information_schema.table_constraints WHERE table_catalog = yyy ....
```

Thanks.

jordanlewis:
Thanks for reporting this, I agree that we're missing these columns.

jordanlewis:
@vivekmenezes for triage, could be a good quick ramp up to learn about our virtual tables

bsnyder788:
I would like to take this one on @jordanlewis. Any pointers on this one?

jordanlewis:
@bsnyder788 very sorry for the delay. This dropped off my radar. You are more than welcome to pitch in here.

Check out `information_schema.go` for the definition.

bsnyder788:
thanks @jordanlewis, I will check it out!

catroot:
Also pls add some more sensitive columns:
`peewee.ProgrammingError: relation "information_schema.constraint_column_usage" does not exist`
is it possible to make an `information_schema` 1-to-1 like it done in postgres?

jordanlewis:
Hi @catroot, going to open a new issue for `constraint_column_usage`. This one should have been closed earlier.

Issue #9897, opened 2016-10-12
sql: support "DEFERRABLE INITIALLY DEFERRED" for constraints


alex:
Currently attempting to use this syntax results in a syntax error:

``` sql
ALTER TABLE "auth_permission" ADD CONSTRAINT "auth_permission_content_type_id_2f476e4b_fk_django_co" FOREIGN KEY ("content_type_id") REFERENCES "django_content_type" ("id") DEFERRABLE INITIALLY DEFERRED
```

It'd be great if it was supported; or failing that if it at least errored with an explicit "this feature will not be supported" message (as is done with `SAVEPOINT`)

dt:
I'll add them to the grammar and return a more helpful error message for now, until they get implemented.

Also, I'm just curious: is that a query that django generates out of the box (versus something hand-written)? I just ask because I'm wondering if this is something that we'll need to implement for django compatibility.

alex:
Yeah, it generates these when creating foreign keys.

dt:
Unassigned since I'm not actively working on this right now. In case anyone decides to try this, I think we might be able to accumulate checks to perform in the session, and then run them during COMMIT. How to represent them and how that interacts with distsql need to be thought out further though.

knz:
I think it is not incorrect as per sql specs to ignore the additional specifier and just perform the check immediately. I understand that `deferrable` means the engine *may* defer checking the constraint, but it doesn't need to. Is there a reason why we couldn't do this as a stop-gap?

dt:
Well, if the user says `INITIALLY DEFERRED`, silently ignoring that seems bad.

Also, you _need_ deferrable to work with circular references.

knz:
ah, I forgot about circular references. my bad.

dt:
Yeah, `deferrable` isn't so much for the engine as it is for a txn, where you don't care that every intermediate state validate (yay isolation), so allowing them not to allows otherwise impossible behaviors (e.g. circular FKs) or significant performance optimizations (e.g. batched validation).

akkinenirajesh:
I am willing to contribute to this issue. But new to our code base. Can someone give an highlevel plan which I will try to convert to code. 

Initial idea: Collect all the checks(Some struct) into some collection inside client.Txn if differable and/or differed. Then while commit run those checks, and throw errors.

dt:
@akkinenirajesh Thanks! I'm happy help with codebase/design/whatever!

[...]

PR #19466, opened 2017-10-23
sql: add missing columns to information_schema.table_constraints

dsaveliev:
Related to #18525 
It seems that the deferrable constraints are still not supported - #9897, 
so I put a couple of stubs in the `ConstraintDetail` in advance.

### DIFF
--- a/pkg/sql/information_schema.go
+++ b/pkg/sql/information_schema.go
@@ -410,9 +410,12 @@ CREATE TABLE information_schema.table_constraints (
 	CONSTRAINT_CATALOG STRING NOT NULL DEFAULT '',
 	CONSTRAINT_SCHEMA STRING NOT NULL DEFAULT '',
 	CONSTRAINT_NAME STRING NOT NULL DEFAULT '',
+	TABLE_CATALOG STRING NOT NULL DEFAULT '',
 	TABLE_SCHEMA STRING NOT NULL DEFAULT '',
 	TABLE_NAME STRING NOT NULL DEFAULT '',
-	CONSTRAINT_TYPE STRING NOT NULL DEFAULT ''
+	CONSTRAINT_TYPE STRING NOT NULL DEFAULT '',
+	IS_DEFERRABLE STRING NOT NULL DEFAULT '',
+	INITIALLY_DEFERRED STRING NOT NULL DEFAULT ''
 );`,
 	populate: func(ctx context.Context, p *planner, prefix string, addRow func(...parser.Datum) error) error {
 		return forEachTableDescWithTableLookup(ctx, p, prefix, func(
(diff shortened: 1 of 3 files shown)


pkg/sql/information_schema.go
```diff
@@ -430,9 +434,13 @@ CREATE TABLE information_schema.table_constraints (
 					defString,                         // constraint_catalog
 					parser.NewDString(db.Name),        // constraint_schema
 					dStringOrNull(name),               // constraint_name
+					defString,                         // table_catalog
 					parser.NewDString(db.Name),        // table_schema
 					parser.NewDString(table.Name),     // table_name
 					parser.NewDString(string(c.Kind)), // constraint_type
+					yesOrNoDatum(c.IsDeferrable),      // is_deferrable
+					yesOrNoDatum(c.InitiallyDeferred), // initially_deferred
```
justinj: You can just hardcode the argument here as false, there's a handful of places we do something similar when we need to expose info regarding features we don't support (also nit: extra line).

pkg/sql/sqlbase/table.go
```diff
@@ -2026,6 +2026,12 @@ type ConstraintDetail struct {
 	Details     string
 	Unvalidated bool
 
+	// This fields just a stubs for deferrable constraints logic.
+	// Looks like AlterTableAddConstraint command (as well as CreateTable) should
+	// store somewhere ValidationBehavior for further use in this structure.
+	IsDeferrable      bool
+	InitiallyDeferred bool
```
justinj: Likewise, I think it would be better not to have these around if they're not actually used. I think supporting `DEFERRABLE` constraints is a moderately large project.

justinj (commented):
Hey @dsaveliev, thanks for another contribution 😄

pkg/sql/sqlbase/table.go
benesch: Indeed. There's an RFC coming soon on the topic courtesy of @BramGruneir.

dsaveliev: Ok, stubs removed.

pkg/sql/information_schema.go
dsaveliev: Fixed.

pkg/sql/information_schema.go
```diff
@@ -410,9 +410,13 @@ CREATE TABLE information_schema.table_constraints (
...
 	CONSTRAINT_SCHEMA STRING NOT NULL DEFAULT '',
 	CONSTRAINT_NAME STRING NOT NULL DEFAULT '',
+	TABLE_CATALOG STRING NOT NULL DEFAULT '',
 	TABLE_SCHEMA STRING NOT NULL DEFAULT '',
 	TABLE_NAME STRING NOT NULL DEFAULT '',
-	CONSTRAINT_TYPE STRING NOT NULL DEFAULT ''
+	CONSTRAINT_TYPE STRING NOT NULL DEFAULT '',
+	IS_DEFERRABLE STRING NOT NULL DEFAULT '',
+	INITIALLY_DEFERRED STRING NOT NULL DEFAULT ''
+
```
justinj: Would you also mind removing this extra line? Sorry for not noticing the first time around.

dsaveliev: Oh, sure. Fixed.

justinj:
Thanks again!

PR #19467, opened 2017-10-23
storage: avoid unnecessary clock updates and access to clock mutex

spencerkimball:
Because `Clock.Update` already has code in it to verify that the clock isn't
being updated to a time more recent than the current physical wall clock time
plus the max offset, add a new method, `Clock.UpdateAndCheckMaxOffset` to
combine the update and check into a single lock.

Also, in `Store.Send`, only update the clock in the event that the
`BatchResponse.Timestamp` has been forwarded from `BatchRequest.Timestamp`.

For single node requests generated via loadgen/kv and resent to
Store.Send, this improves performance by ~7.5%.

### DIFF
--- a/pkg/util/hlc/hlc.go
+++ b/pkg/util/hlc/hlc.go
@@ -19,6 +19,7 @@
 package hlc
 
 import (
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


bdarnell:
:lgtm:

I'm amazed the effect is that large. 

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19467#-:-KxAE9MtewJw5eWeVLwh:b6vwusk)*

spencerkimball:
Take that with a grain of salt. That's 3x 60s measurements on my laptop. Also, this is testing performance from `Store.Send`, not an external load generator.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19467#-:-KxAEOwQTWj3k27v8XvB:b-9ghr51)*

pkg/util/hlc/hlc.go
```diff
@@ -181,36 +182,46 @@ func (c *Clock) PhysicalTime() time.Time {
...
-// the remote timestamp was rejected due to clock offset,
-// in which case the timestamp of the clock will not have been
-// altered.
-// To timestamp events of local origin, use Now instead.
+// Update takes a hybrid timestamp, usually originating from an event
+// received from another member of a distributed system. The clock is
+// updated and the hybrid timestamp associated to the receipt of the
+// event returned. If the remote timestamp exceeds the wall clock
+// time by more than the maximum clock offset, the update is rejected
+// and a warning is logged.
```
tschottdorf: This comment is incorrect (it was incorrect before, though).

spencerkimball: Will send an update.

Issue #19288, opened 2017-10-16
perf: critical path analysis for DistSQL


arjunravinarayan:
Perform a critical path analysis for DistSQL, similar to [the single key read critical path analysis](https://spencerkimball.github.io/critical-path/), focusing first on the simplest of DistSQL queries (`SELECT COUNT(*) FROM table`), to examine where the greatest performance bottlenecks are.

This will help triage performance work on making DistSQL faster.

arjunravinarayan:
I'm working with a `COUNT(*)` query on the `lineitem` table from the
TPC-H scalefactor=10 dataset. Query run on a single node (a
`roachprod` worker with a local-ssd attached). The goal is to see
where the bottlenecks are in the system to running this query
faster. The query:

`SELECT COUNT(*) FROM lineitem`.

First step is to run the query (runtime: ~3m) and look at all the
profiles. The table's primary index is about 7GB in size, so that
gives us a rough end-to-end bandwidth of ~40MB/s.

Here is a [bare_cpuprofile.pdf](https://github.com/cockroachdb/cockroach/files/1409052/bare_cpuprofile.pdf).



The CPU profile shows that a healthy portion (~65%) of the work
happens under `MVCCIterate`, but 12% under `tablereader.go`, and 10%
under `aggregator.go`. It's hard to disentangle exactly what this
means from this graph, so I switched to a [bare_calltree.pdf](https://github.com/cockroachdb/cockroach/files/1409054/bare_calltree.pdf) profile. Here's a
better picture: `aggregator.go` takes 10% of CPU, of which 4% is
memory allocation, and 4% is go channel traffic. Second,
`tablereader.go` is 14% of CPU, of which 4% is go channel traffic,
and 3% is in finalizing `EncDatumRow`s in `RowFetcher`.

Looking at heap profiles (`inuse_space`, `inuse_objects`,
`alloc_space`, `alloc_objects`) doesn't reveal anything significant at
the distsql transport layer. So it's not quite that DistSQL is
allocating huge amounts of memory, but that it's getting stuck
transporting it around.

I made `kvFetcher` throw rows away and performance improved by 20%,
which shows that improvements on the DistSQL side can only give us at
most that much throughput increase. Currently working on getting some
more visibility at the kvFetcher layer and below.

To investigate what I said earlier about "go channel traffic" : to
execute a `COUNT(*)`, tablereader transports the rows (one by one) via
a `RowChannel` to its destination processor (in this case,
[...]

PR #19471, opened 2017-10-23
WIP: distsql: benchmark serialization costs in distsql

rjnn:
This commit adds a benchmark to measure how fast we can pass rows through a `RowChannel` in DistSQL: the results are not great:

```
pkg: github.com/cockroachdb/cockroach/pkg/sql/distsqlrun
BenchmarkSerializingRows/BenchmarkSerializingRows100-4         	  100000	     18671 ns/op	  42.85 MB/s
BenchmarkSerializingRows/BenchmarkSerializingRows1000-4        	   10000	    156456 ns/op	  51.13 MB/s
BenchmarkSerializingRows/BenchmarkSerializingRows10000-4       	    1000	   1616575 ns/op	  49.49 MB/s
BenchmarkSerializingRows/BenchmarkSerializingRows100000-4      	     100	  16287727 ns/op	  49.12 MB/s
BenchmarkSerializingRows/BenchmarkSerializingRows1000000-4     	       5	 210394039 ns/op	  38.02 MB/s
PASS
```

This is a rather low amount of bandwidth to be pushing through the `RowChannel`, and gates performance for queries, as seen in #19288. Criticisms on how to improve this benchmark welcome.

### DIFF
--- a/pkg/sql/distsqlrun/serialization_test.go
+++ b/pkg/sql/distsqlrun/serialization_test.go
@@ -1,0 +1,263 @@
+// Copyright 2016 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package distsqlrun
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"fmt"
+
+	"strconv"
+
+	"sync"
+
+	"github.com/cockroachdb/cockroach/pkg/sql/parser"
+	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+)
+
+// Benchmark how long it takes to pass rows from one processor to another.
+
+type DevRandRowSource struct {
+	r             *rand.Rand
+	rowsToGen     int
+	cursor        int
+	backingDatums []sqlbase.EncDatumRow
+	types         []sqlbase.ColumnType
+}
+
+func (rs *DevRandRowSource) Init(types []sqlbase.ColumnType, rowsToGen int) {
+	rs.r = rand.New(rand.NewSource(int64(time.Now().UnixNano())))
+	rs.types = types
+	rs.rowsToGen = rowsToGen
+	rs.backingDatums = make([]sqlbase.EncDatumRow, rowsToGen)
+
+	for i := range rs.backingDatums {
+		rs.backingDatums[i] = genRandomEncDatumRow(types, rs.r)
+	}
+}
+
+func genRandomEncDatum(t sqlbase.ColumnType, r *rand.Rand) sqlbase.EncDatum {
+	var ed sqlbase.EncDatum
+	var d parser.Datum
+	switch t.SemanticType {
+	case sqlbase.ColumnType_BOOL:
+		b := r.Int31n(1)&0x01 == 0
+		d = parser.MakeDBool(parser.DBool(b))
+	case sqlbase.ColumnType_INT:
+		i := r.Int31()
+		d = parser.NewDInt(parser.DInt(i))
+	case sqlbase.ColumnType_FLOAT:
+		f := r.Float32()
+		d = parser.NewDFloat(parser.DFloat(f))
+	case sqlbase.ColumnType_STRING:
+		i := r.Int63()
+		s := strconv.Itoa(int(i))
+		d = parser.NewDString(s)
+	case sqlbase.ColumnType_BYTES:
+		i := r.Int63()
+		s := strconv.Itoa(int(i))
+		d = parser.NewDBytes(parser.DBytes(s))
+	default:
+		panic(fmt.Sprintf("unsuported type: %v"))
+	}
+
+	ed = sqlbase.DatumToEncDatum(t, d)
+	return ed
+}
+
+func genRandomEncDatumRow(ts []sqlbase.ColumnType, r *rand.Rand) sqlbase.EncDatumRow {
+	row := make(sqlbase.EncDatumRow, len(ts))
+	for i, t := range ts {
+		row[i] = genRandomEncDatum(t, r)
+	}
+	return row
+}
+
+func (rs *DevRandRowSource) Types() []sqlbase.ColumnType {
+	return rs.types
+}
+
+// DevRandRowSource.Next() returns the next row from the backing datums until
+// there are no more rows. It never returns a non-nil
+func (rs *DevRandRowSource) Next() (sqlbase.EncDatumRow, ProducerMetadata) {
+	var row sqlbase.EncDatumRow
+
+	if (rs.cursor + 1) >= rs.rowsToGen {
+		return nil, ProducerMetadata{}
+	}
+	row = rs.backingDatums[rs.cursor]
+	rs.cursor++
+	return row, ProducerMetadata{}
+}
+
+func (rs *DevRandRowSource) ConsumerDone()   {}
+func (rs *DevRandRowSource) ConsumerClosed() {}
+
+var _ RowSource = &DevRandRowSource{}
+
+type DevNullRowReceiver struct{}
+
+func (rr *DevNullRowReceiver) Push(row sqlbase.EncDatumRow, meta ProducerMetadata) ConsumerStatus {
+	if !meta.Empty() {
+		panic(meta)
+	}
+
+	if row == nil {
+		return ConsumerClosed
+	}
+	return NeedMoreRows
+}
+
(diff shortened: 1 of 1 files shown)


arjunravinarayan:
(This PR is just documenting the benchmarks for the experiment in #19288).

My main concern is that the bandwidth measurements are based off of a single integer being sent (i.e. what would occur in a `SUM` query). It undercounts larger rows, which are typical of JOIN queries (which always send at least one join key and one payload). This bandwidth might be considered more respectable in such scenarios. Nevertheless, a target bandwidth to aim for on a local machine should be closer to 1000 MB/s, with little to no CPU movement.

Next things to add would be to benchmark the `EncDatum` decoding allocations, and measure bandwidth, allocation, and CPU usage across machine barriers, using GRPC.

We ultimately want to use these benchmarks to then set a baseline to test the throughput of our batched implementation.

knz:
18671 ns/op for a batch of 100? That's 18ns per item. At 1000000 you have 21ns per item. It doesn't sound bad to me. If you add batching on top of that, you could have multiple items per 18ns perhaps?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-:-KxCJW8h0oZJTCN4ev6I:b-33zzf8)*

spencerkimball:
One idea is not to serializable at all. Use Cap’n proto. These are just static objects in memory and we never expect to change them as they’re passed through the system. Why bother encoding and decoding? Capnproto.org

RaduBerinde:
We're not really serializing anything (in the sense of encoding, sending, then decoding). We're effectively just moving pointers and slices around. I would call it `BenchmarkRowChannel`.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-:-KxDWApn2rNyWPEWa34z:bh8oa71)*

arjunravinarayan:
@spencerkimball: Yes, that's the ideal end-goal, but @knz had a thought that we could get most of the benefit (and the type safety) even if we just stuck to go structs that we passed around with pointers (basically free within a single node, but paying the serialization costs between nodes), as we could still make it fast enough to not be the bottleneck. But capnproto-like non-serialization of columnar batches is what [timely dataflow does](https://github.com/frankmcsherry/abomonation), and it works very well.

@knz its 20ns for what should basically be free. We're not doing any serialization. We're just passing stuff around.

knz:
"Passing stuff around" through a Go channel. 20ns is the price you pay for synchronization. It doesn't seem unreasonable. It also can be amortized by transferring more data each time. I suspect that a benchmark that transfers more rows, normalized to nanosecond per bytes, will see a lower (normalized) latency.

arjunravinarayan:
@knz, yes, that's my takeaway as well - we need to pass around larger batches of things at a time, because of the synchronization overhead involved in each `Push` call.

RaduBerinde:
Absolutely, we need to pass batches of rows, with a more efficient memory layout. I just clarified that we're not encoding/decoding on local streams.

knz:
FYI the call overhead of the Next() method  with scanNodes is about the same.

RaduBerinde:
Yes, the per-row method calls (on both the sending and receiving side) are also a significant part of the problem, which batching would also address.

arjunravinarayan:
@RaduBerinde thanks for the clarifications, agreed on all points.

petermattis:
For context, on my laptop a function call (not via an interface) is 1.86ns.

knz:
With 20ns to start with, the maximum speedup to be gained by shaving call overheads if we keep 1 synchronization per row is between 20 and 40 (assuming 0.5-1ns per sync) due to architectural limits. With row batching, assuming we can batch hundreds of rows or more, the speedup is potentially an order of magnitude larger.

danhhz:

---

*[pkg/sql/distsqlrun/serialization_test.go, line 236 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxDihN6QPKxdqBG4D0x:-KxDihN6QPKxdqBG4D0y:b-wlp83c) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L236)):*
> ```Go
> func BenchmarkSerializingRows(b *testing.B) {
> 	for _, rows := range []int{100, 1000, 10000, 100000, 1000000} {
> 		b.Run(fmt.Sprintf("BenchmarkSerializingRows%d", rows), func(b *testing.B) {
> ```

this is about as nit-y as it gets: you don't need the `BenchmarkSerializingRows` here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471)*

asubiotto:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/serialization_test.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxEXT8VRQbg1upDPZI0:-KxEXT8VRQbg1upDPZI1:b-tdg98t) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L34)):*
> ```Go
> // Benchmark how long it takes to pass rows from one processor to another.
> 
> type DevRandRowSource struct {
> ```

I would put any new `RowReceiver`s and `RowSource`s in `pkg/sql/distsqlrun/testutils.go` with the other ones we use for benchmarks.

---

*[pkg/sql/distsqlrun/serialization_test.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxEXFFI7wIpcPWjDMWO:-KxEXFFI7wIpcPWjDMWP:b-59ze51) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L53)):*
> ```Go
> }
> 
> func genRandomEncDatum(t sqlbase.ColumnType, r *rand.Rand) sqlbase.EncDatum {
> ```

There are some random row generator functions in `pkg/sql/sqlbase/testutils.go`. You could probably use those.

---

*[pkg/sql/distsqlrun/serialization_test.go, line 112 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxEWFa1F9ILwp1ltxyS:-KxEWFa1F9ILwp1ltxyT:b2ofzcc) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L112)):*
> ```Go
> var _ RowSource = &DevRandRowSource{}
> 
> type DevNullRowReceiver struct{}
> ```

You might be able to reuse `RowDisposer` in `pkg/sql/distsqlrun/testutils.go`

---

*[pkg/sql/distsqlrun/serialization_test.go, line 236 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxDihN6QPKxdqBG4D0x:-KxEXbco2twt6ouK9iJs:bzghsd1) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L236)):*

this is about as nit-y as it gets: you don't need the `BenchmarkSerializingRows` here
</blockquote></details>

The style we decided to use is along the lines of `b.Run(fmt.Sprintf("NumRows=%d", rows))`

---

*[pkg/sql/distsqlrun/serialization_test.go, line 238 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxEWxsg08bBPkBEabtF:-KxEWxshiwwMXZFobDSQ:b-51dtdw) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L238)):*
> ```Go
> 		b.Run(fmt.Sprintf("BenchmarkSerializingRows%d", rows), func(b *testing.B) {
> 			var source DevRandRowSource
> 			anInt := []sqlbase.ColumnType{{SemanticType: sqlbase.ColumnType_INT}}
> ```

There's a global `var intType` in `pkg/sql/distsqlrun/util_test.go` you can reuse.

---

*[pkg/sql/distsqlrun/serialization_test.go, line 239 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471#-KxEWgDv9vMRaxQBHB4M:-KxEWgDv9vMRaxQBHB4N:bdnk5ck) ([raw file](https://github.com/cockroachdb/cockroach/blob/3b26b3c1d97b72a737a121b4c6247062dc67861c/pkg/sql/distsqlrun/serialization_test.go#L239)):*
> ```Go
> 			var source DevRandRowSource
> 			anInt := []sqlbase.ColumnType{{SemanticType: sqlbase.ColumnType_INT}}
> 			source.Init(anInt, rows)
> ```

This is going to measure the time taken to initialize the rows, you should probably do this before `b.Run`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19471)*

arjunravinarayan:
Closing as we've proved our point and are moving away from using `RowChannel` in local DistSQL data transfers, e.g. #21220.

PR #19473, opened 2017-10-24
cherrypick-1.1: storageccl: use the official AWS SDK

### LINKED ISSUE #19435: sqlccl: import problem from different s3 region
Copied from https://stackoverflow.com/q/46852338/864236:

---

I try to bulk import into a cockroach db from a S3 bucket in the eu-west-1 region
    
    IMPORT TABLE osm.nodes (
            id INT PRIMARY KEY,
            version INT NOT NULL,
            lat DECIMAL NOT NULL,
            lon DECIMAL NOT NULL,
            changeset_id INT NOT NULL,
            visible BOOLEAN NOT NULL
        )
        CSV DATA ('s3://cockroach-import/nodes.csv?AWS_ACCESS_KEY_ID=<snip>&AWS_SECRET_ACCESS_KEY=<snip>')
        WITH
            temp = 's3://cockroach-import/?AWS_ACCESS_KEY_ID=<snip>&AWS_SECRET_ACCESS_KEY=<snip>',
            delimiter = ','
        ;

I get the error message:

    failed to create s3 reader: 400: "The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'eu-west-1'"

How can I specify the S3 region in the import statement?

madelynnblue:
This change is similar to #18902 except that it:
- omits the Size method and tests
- ensures that there are no changes to vendor, only Gopkg.lock

There are various other changes in Gopkg.lock which should have
already been in 1.1, but are included here for correctness.

See #18902
Fixes #19435


Gopkg.toml
```diff
@@ -52,3 +52,7 @@ ignored = [
 [[constraint]]
   name = "github.com/coreos/etcd"
   branch = "master"
+
+[[constraint]]
+  name = "github.com/rlmcpherson/s3gof3r"
+  revision = "864ae0bf7cf2e20c0002b7ea17f4d84fec1abc14"
```
dt: nit: newline

dt (approved):
just to confirm (since we can't have CIs to it): did you run the s3 storageccl tests by hand against us and non-us buckets?

mjibson:
Yes, I made a eu-west-1 bucket and ran the S3 tests on it.

bdarnell (approved):

PR #19480, opened 2017-10-24
storage: create RangeMatchingPred function, use for tscache accesses

nvb:
This change specializes the `keys.Range` functions that `Replica` uses
to determine the influence spans when adding and expanding
`tscache.Requests`. It restricts the `Range` function to
`roachpb.Requests` that update the `TimestampCache` when adding a
request and restricts the `Range` function to `roachpb.Requests` that
consult the `TimestampCache` for expanding requests. This means that
fewer unnecessary `tscache.Requests` will be expanded because unrelated
`roachpb.Requests` in a `BatchRequest` will no longer expand the
influence spans passed to the `tscache` methods.

This wasn't inspired by any particular workload, but it seemed like the
right thing to do. I also think the change makes purpose of the
`RSpan` passed to `AddRequest` and `ExpandRequests` more self-explanatory.

### DIFF
--- a/pkg/keys/keys.go
+++ b/pkg/keys/keys.go
@@ -672,13 +672,24 @@ func EnsureSafeSplitKey(key roachpb.Key) (roachpb.Key, error) {
 	return key[:idx], nil
 }
 
-// Range returns a key range encompassing all the keys in the Batch.
+// Range returns a key range encompassing the key ranges of all requests in the
+// Batch.
 func Range(ba roachpb.BatchRequest) (roachpb.RSpan, error) {
+	return RangeMatchingPred(ba, nil)
+}
+
+// RangeMatchingPred returns a key range encompassing the key ranges of all
+// requests in the Batch that match the provided predicate. If no predicate
+// is provided, no filtering will be performed on the requests in the Batch.
+func RangeMatchingPred(
+	ba roachpb.BatchRequest, pred func(roachpb.Request) bool,
+) (roachpb.RSpan, error) {
 	from := roachpb.RKeyMax
 	to := roachpb.RKeyMin
 	for _, arg := range ba.Requests {
 		req := arg.GetInner()
-		if _, ok := req.(*roachpb.NoopRequest); ok {
+		_, noop := req.(*roachpb.NoopRequest)
+		if noop || (pred != nil && !pred(req)) {
 			continue
 		}
 		h := req.Header()
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -1907,13 +1907,7 @@ func (ec *endCmds) done(br *roachpb.BatchResponse, pErr *roachpb.Error, retry pr
 	// marked as affecting the cache are processed. Inconsistent reads
 	// are excluded.
 	if pErr == nil && retry == proposalNoRetry && ec.ba.ReadConsistency != roachpb.INCONSISTENT {
-		span, err := keys.Range(ec.ba)
-		if err != nil {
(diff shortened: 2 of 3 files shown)


petermattis:
:lgtm:

---

---

*[pkg/storage/replica.go, line 1935 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480#-KxDf2XAfAMM7ZKUwLGZ:-KxDf2XAfAMM7ZKUwLG_:b-19ydrx) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e1791754f395ecb1b16952a5187f65428974b6b/pkg/storage/replica.go#L1935)):*
> ```Go
> 
> 	var err error
> 	if cr.Span, err = keys.RangeMatchingPred(*ba, roachpb.UpdatesTimestampCache); err != nil {
> ```

You'll want to buf allocate `cr.Span` when your other PR merges.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480#-:-KxDfCRjhase5WSQEjd3:bnfp4nl)*

bdarnell:
:lgtm:, although I'm not sure how much of a difference it will make. We don't mix reads and writes so I think most requests in the same batch will (currently) tend to have the same timestamp cache flags. 

---

---

*[pkg/keys/keys.go, line 685 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480#-KxEJqvXm7eSz_Px99Lc:-KxEJqvXm7eSz_Px99Ld:bud896o) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e1791754f395ecb1b16952a5187f65428974b6b/pkg/keys/keys.go#L685)):*
> ```Go
> 
> // RangeMatchingPred returns a key range encompassing the key ranges of all
> // requests in the Batch that match the provided predicate.
> ```

If `pred` ever returns true for a `NoopRequest`, the result will be nonsense (and we have a history of letting this slip through). I think `RangeMatchingPred` should filter out `NoopRequests` on its own before calling the predicate. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480#-:-KxEKQ50UN-3v5AaENnt:by8mbf8)*

nvanbenschoten:

---

*[pkg/keys/keys.go, line 685 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480#-KxEJqvXm7eSz_Px99Lc:-KxJDW4ewBI-lgsEfrze:b-qu6n94) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e1791754f395ecb1b16952a5187f65428974b6b/pkg/keys/keys.go#L685)):*

If `pred` ever returns true for a `NoopRequest`, the result will be nonsense (and we have a history of letting this slip through). I think `RangeMatchingPred` should filter out `NoopRequests` on its own before calling the predicate. 
</blockquote></details>

Good point. Done.

---

*[pkg/storage/replica.go, line 1935 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480#-KxDf2XAfAMM7ZKUwLGZ:-KxJDXG8hGHKZ7M-aB4w:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e1791754f395ecb1b16952a5187f65428974b6b/pkg/storage/replica.go#L1935)):*

You'll want to buf allocate `cr.Span` when your other PR merges.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19480)*

PR #19482, opened 2017-10-24
*: remove generated protobufs from version control

benesch:
The `go get` ship sailed long ago, when we moved to using our non-Go
dependencies' native build systems. PR #18349 pushed us even further in
that direction by automatically rebuilding the admin UI as necessary. We
might as well remove the generated protobuf code too; the Make rules to
generate them on-demand are already in place.


tschottdorf:
LGTM in spirit, though it makes me slightly nervous especially now, where we've discovered that our protoc-gen-gogoroach doesn't really get versioned correctly. Either way, this is probably the way to go. I won't miss having to push large diffs on bad connections.

dt:
Hm, I like the idea but I'll be a little sad if our packages (at least those that don't depend on c-deps) stop being importable elsewhere. What will this look like for a consumer (who we'll assumes vendors) of our sql/sqlbase or server/admin packages?

benesch:
Oh, that's an excellent point, David. Can't believe I forgot about that.
Grrrr. I suppose we should hold off on this.

tschottdorf:
Why is it important that our pkgs are importable elsewhere?

benesch:
Registration server, mostly, which vendors the packages containing the
protobufs it needs.

dt:
@tschottdorf currently we depend on sqlbase to read reported SQL data in the reg server. I think it's also not crazy that someone might want to depend on the admin ui API protos.

Maybe we should kick the protos into their own tree, and then allow easily depending on them (by generating into your own source tree however you want) without depending on our go code, which would depend on make-driving JIT generation. 

It's probably also be nice if we kept proto deps out of libs in util since those tend to be pretty reusable (we often reuse them elsewhere in experiments, tools, etc).

RaduBerinde:
We may also want to import packages like `sqlbase` or even `sql` into optimizer prototypes, like Peter's [opttoy](https://github.com/petermattis/opttoy/).

danhhz:
I may be the only one that does this, but I regularly delete the *.pb.go line from .gitattributes because I like seeing the diffs in generated code.

dt:
@danhhz fwiw, you can pass `--text` to `git diff` or `git show`

tschottdorf:
Or `-a`, saving another four key strokes!
-- 

-- Tobias

andreimatei:
@benesch I've read this thread and I'm left unsure about your motivation. What's wrong with checking in this generated code? I understand that `go get` can already not be used to download crdb for reasons, but how does that speak to whether this PR is good or not? I'm trying to figure out the pros and cons.

I've also got a side question - if this PR were to go in, would it have an effect on whether pb.go files need to be regenerated every time I switch branches across .proto changes? (currently they don't need to be regenerated).

tschottdorf:
@andreimatei:

pros:

- no need to re-run CI because the protos changed
- no merge conflicts because of proto changes (unless the conflict is in .proto)
- never pushing these rather large diffs, which is nice on bad connections

cons:
- Makefile really needs to work well so that switching branches auto-regenerates ;-)
- can't see the generated code in diffs

Probably missing something.

RaduBerinde:
Another con is that building after switching branches will be more expensive.

I'm mildly against doing this.

benesch:
As a result of the embedded.go removal, switching branches that span a protobuf change requires regenerating the JavaScript protobuf definitions. Regenerating the Go protobuf definitions is negligible in comparison. So we've already accepted that con, in a sense.

knz:
Hum we could store the javascript protobuf definitions in the repo too right?

bdarnell:
I'm also mildly against this, mainly because it will probably take more time than we expect to get the Makefile rules correct and ensure that everything is regenerated when it needs to be (and keep it that way over time). I also think that it's worth keeping proto-using packages relatively easy to use from other places. 

If we do make this change, we should consider doing it for all the other `go generate` rules too - the same arguments (pro and con) apply to yacc, stringer, etc.

bdarnell:
> Maybe we should kick the .protos into their own sub-tree, and then allow easily depending on them, by generating into your own source tree however you want, be the canonical way to consume them externally, without depending on any of our go code.

@dt for historical interest, I was working on this back when we were planning to build our own client drivers: https://github.com/cockroachdb/cockroach-proto

benesch:
@knz we definitely don't want to store the compiled JavaScript protobufs in the repository. They're generated into two big hunks (`pkg/ui/src/js/protos{.js,.d.ts}`), so checking them in means any commit that touches a protobuf anywhere conflicts with any other commit that touches a protobuf.

Go protobufs and C++ protobufs, by contrast, generate a XXX.pb.go or XXX.pb.cc for every XXX.proto, so conflicts only occur when two commits touch the same XXX.proto file. We could probably teach the JavaScript code to generate XXX.pb.js for each proto, instead of one big hunk, but not without some serious effort.

benesch:
Ok, closing for now, but filed #19552 to track.

PR #19483, opened 2017-10-24
config,keys,sqlbase: move MakeZoneKey[Prefix] from sqlbase to config

benesch:
Logically, encoding zone keys belongs in the config package.
Practically, partitioning requires that the config package contain the
logic to encode zone keys to avoid a circular dependency. (Finding split
keys now requires looking at the zone config to determine whether a
table partition exists, rather than just looking at the structure of the
key to determine if it spans a table boundary.)

----

Heads up: you'll probably want to filter out the first commit, which is just code movement.


danhhz:
:lgtm:

---

Review status: 0 of 23 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-:-KxIs1Anr0SOI0U9KL5S:bnfp4nl)*

BramGruneir:
:lgtm: 

---

Reviewed 16 of 17 files at r1, 9 of 9 files at r2.

---

*[pkg/config/system.proto, line 15 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-KxIy30z0s_GZdXkWCLW:-KxIy30z0s_GZdXkWCLX:b-xr4jeu) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1fee718266a5764ba92633983dd496504843fd4/pkg/config/system.proto#L15)):*
> ```protobuf
> // permissions and limitations under the License.
> 
> syntax = "proto2";
> ```

Worth changing this to a proto3?  Maybe in a follow up PR? Might be a good time since you're mucking around here.

---

*[pkg/sql/database.go, line 91 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-KxIyuhwx8Gc7p6-Yi02:-KxIyuhwx8Gc7p6-Yi03:bp387dy) ([raw file](https://github.com/cockroachdb/cockroach/blob/36422ee65f58ac0637b6d8acdf54e3d685b7ba6c/pkg/sql/database.go#L91)):*
> ```Go
> 	dbDesc *sqlbase.DatabaseDescriptor,
> ) (zoneKey roachpb.Key, nameKey roachpb.Key, descKey roachpb.Key) {
> 	zoneKey = config.MakeZoneKey(uint32(dbDesc.ID))
> ```

Isn't dbDesc.ID already an uint32?  It is in the proto.

Oh, it's type ID.  Perhaps that should move to config as well?  The repeated casts seems ugly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-:-KxIzuirgRKBUI_5b8dc:bg7igkv)*

danhhz:

---

*[pkg/sql/database.go, line 91 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-KxIyuhwx8Gc7p6-Yi02:-KxJ51djaa8c8nQ_SASg:b-k7ah38) ([raw file](https://github.com/cockroachdb/cockroach/blob/36422ee65f58ac0637b6d8acdf54e3d685b7ba6c/pkg/sql/database.go#L91)):*

Isn't dbDesc.ID already an uint32?  It is in the proto.

Oh, it's type ID.  Perhaps that should move to config as well?  The repeated casts seems ugly.
</blockquote></details>

ID is a SQL concept and should stay in sqlbase. Not sure I agree that these casts are ugly

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483)*

benesch:

---

*[pkg/config/system.proto, line 15 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-KxIy30z0s_GZdXkWCLW:-KxJ5RaskZUeoOEyUKwD:bchjspr) ([raw file](https://github.com/cockroachdb/cockroach/blob/b1fee718266a5764ba92633983dd496504843fd4/pkg/config/system.proto#L15)):*

Worth changing this to a proto3?  Maybe in a follow up PR? Might be a good time since you're mucking around here.
</blockquote></details>

Definitely don't want to mess with that now, but filed a follow-up issue! #19524

---

*[pkg/sql/database.go, line 91 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-KxIyuhwx8Gc7p6-Yi02:-KxJ5XGUpn3iFPnPsVAT:bis3qig) ([raw file](https://github.com/cockroachdb/cockroach/blob/36422ee65f58ac0637b6d8acdf54e3d685b7ba6c/pkg/sql/database.go#L91)):*

ID is a SQL concept and should stay in sqlbase. Not sure I agree that these casts are ugly
</blockquote></details>

Personally I wish we just used uint32 everywhere. I'm going to hold off on this for now because it'd be a massive undertaking to move `sqlbase.ID` elsewhere, but this definitely makes me itchy.

Filed https://github.com/cockroachdb/cockroach/issues/19525 to track the meta-issue of "config has gross inter-dependencies with the SQL layer".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483)*

benesch:
@BramGruneir, since you LGTM'd, I'm going to merge so I can start sending along some of my dependent followup PRs. Let me know if you're not satisfied with either of the two issues I filed, though—and TFTR!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-:-KxJ6P13VFkP__P1_X67:b-dnk566)*

BramGruneir:
SGTM

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-:-KxJ6f8tCgRTtHtSECaW:bb73djm)*

BramGruneir:
*[pkg/sql/database.go, line 91 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483#-KxIyuhwx8Gc7p6-Yi02:-KxJ6W_hq1zHpHb9EPOM:b3c7cbs) ([raw file](https://github.com/cockroachdb/cockroach/blob/36422ee65f58ac0637b6d8acdf54e3d685b7ba6c/pkg/sql/database.go#L91)):*

Personally I wish we just used uint32 everywhere. I'm going to hold off on this for now because it'd be a massive undertaking to move `sqlbase.ID` elsewhere, but this definitely makes me itchy.

Filed https://github.com/cockroachdb/cockroach/issues/19525 to track the meta-issue of "config has gross inter-dependencies with the SQL layer".
</blockquote></details>

Yeah, ID should be in sql.  Just a little messy is all.

SGTM

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19483)*

Issue #17603, opened 2017-08-11
storage: Make sure that DiversityScores make sense with unusual locality configurations


a-robinson:
As @BramGruneir wisely pointed out in #17570, our DiversityScore logic may not work properly with all possible locality configurations (such as wanting more replicas-per-range than the number of localities, or weird locality hierarchies like `a-1, a-2, b, c-i, c-ii, c-iii`), yet we don't restrict locality configurations at all.

We need to think through how things may break down with weird configurations and consider new approaches that are more generalizable.

a-robinson:
As with #17479, this would be too big of a change at this point in the release cycle. Moving to 1.2.

a6802739:
@a-robinson, could I have a try?

a-robinson:
Sure! Since this is pretty open ended, what we first need is for someone to think the problem through and write down a proposal with examples of how the proposal handles different kinds of locality configurations.

PR #19486, opened 2017-10-24
storage: make average diversity score for diversityRemovalScore

a6802739:
Fixes #17479.

@a-robinson, I didn't use `diversityRemovalAverageScore` to replace `diversityRemovalScore` in `rebalanceCandidates`.

If I replace it, it will fail in the test `TestReplicateQueueRebalance`.

PR #19489, opened 2017-10-24
storage: Trigger a rebalance if diversity can be improved

### LINKED ISSUE #17599: storage: Bad replica diversity should trigger a rebalance
Follow-up issue from #17570 that we don't want to include in 1.1 but want to get in early in the 1.2 cycle.

Right now, we don't consider bad diversity a reason to rebalance a range even though a human would probably consider something like all of a range's replicas being in the same locality a pretty great reason to rebalance. If the allocator could improve replica diversity by triggering a rebalance, it should consider doing so.

a-robinson:
Fixes #17599, which we wanted to do early in the 1.2 release cycle.

I don't think this should be merged quite yet, due to the TODO on one of the new test cases that brings out bad diversity scoring behavior, but am sending out to maybe inspire some thought related to #17603.

I'd also like to add a few more test cases to this once the diversity score changes have settled down.

### DIFF
--- a/pkg/storage/allocator_scorer.go
+++ b/pkg/storage/allocator_scorer.go
@@ -480,7 +480,7 @@ func rebalanceCandidates(
 	if !rebalanceConstraintsCheck {
 		for _, store := range sl.stores {
 			if _, ok := existingStoreIDs[store.StoreID]; ok {
-				if shouldRebalance(ctx, store, constraintsOkStoreList, rangeInfo, options) {
+				if shouldRebalance(ctx, store, constraintsOkStoreList, rangeInfo, existingNodeLocalities, options) {
 					shouldRebalanceCheck = true
 					break
 				}
@@ -585,12 +585,28 @@ func shouldRebalance(
 	store roachpb.StoreDescriptor,
 	sl StoreList,
 	rangeInfo RangeInfo,
+	existingNodeLocalities map[roachpb.NodeID]roachpb.Locality,
 	options scorerOptions,
 ) bool {
+	// Rebalance if this store is too full.
 	if store.Capacity.FractionUsed() >= maxFractionUsedThreshold {
 		log.VEventf(ctx, 2, "s%d: should-rebalance(disk-full): fraction-used=%.2f, capacity=(%v)",
 			store.StoreID, store.Capacity.FractionUsed(), store.Capacity)
 		return true
+	}
+
+	diversityScore := rangeDiversityScore(existingNodeLocalities)
+	for _, desc := range sl.stores {
+		if !preexistingReplicaCheck(desc.Node.NodeID, rangeInfo.Desc.Replicas) {
+			continue
+		}
+		otherScore := diversityRebalanceFromScore(desc, store.Node.NodeID, existingNodeLocalities)
+		if otherScore > diversityScore {
+			log.VEventf(ctx, 2,
+				"s%d: should-rebalance(better-diversity=s%d): diversityScore=%.5f, otherScore=%.5f",
+				store.StoreID, desc.StoreID, diversityScore, otherScore)
+			return true
+		}
 	}
 
 	if !options.statsBasedRebalancingEnabled {
@@ -732,6 +748,30 @@ func constraintCheck(store roachpb.StoreDescriptor, constraints config.Constrain
 	return true, positive
 }
 
+// rangeDiversityScore returns a value between 0 and 1 based on how diverse the
+// given range is. A higher score means the range is more diverse.
+// All below diversity-scoring methods should in theory be implemented by
+// calling into this one, but they aren't to avoid allocations.
+func rangeDiversityScore(existingNodeLocalities map[roachpb.NodeID]roachpb.Locality) float64 {
+	var sumScore float64
+	var numSamples int
+	for n1, l1 := range existingNodeLocalities {
+		for n2, l2 := range existingNodeLocalities {
+			// Only compare pairs of replicas where s2 > s1 to avoid computing the
+			// diversity score between each pair of localities twice.
+			if n2 <= n1 {
+				continue
+			}
+			sumScore += l1.DiversityScore(l2)
+			numSamples++
+		}
+	}
+	if numSamples == 0 {
+		return roachpb.MaxDiversityScore
+	}
+	return sumScore / float64(numSamples)
+}
+
 // diversityAllocateScore returns a value between 0 and 1 based on how
 // desirable it would be to add a replica to store. A higher score means the
 // store is a better fit.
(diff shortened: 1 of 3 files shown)


BramGruneir:
This change looks good, and we should be triggering rebalances based on diversity improvements.  But I'll leave it up to you for now to improve testing of it before merging.  I think we might want to create a few new tests similar to Example_rebalancing, that deal with a few different common scenarios with localities included.  This way, when we do change diversity score calculations, we can get an immediate sense of how that change affects common cluster setups.  They should probably move to a new test file too, since allocator_test is already huge.

pkg/storage/allocator_scorer.go
```diff
@@ -552,13 +552,30 @@ func shouldRebalance(
...
 	}
 
+	// Rebalance if we could achieve better locality diversity by doing so.
+	diversityScore := diversityRemovalScore(store.Node.NodeID, existingNodeLocalities)
+	for _, desc := range sl.stores {
+		if !preexistingReplicaCheck(desc.Node.NodeID, rangeInfo.Desc.Replicas) {
+			continue
+		}
+		otherScore := rebalanceToDiversityScore(desc, existingNodeLocalities)
+		if otherScore > diversityScore {
```
a6802739: @a-robinson , This is great!

I have some confusion here, how could we gurantee we could achieve better locality for this range if `otherScore > diversityScore`?

a6802739: @a-robinson , This is great!

I have some confusion here, how could we gurantee we could achieve better locality for this range if otherScore > diversityScore?

pkg/storage/allocator_scorer_test.go
```diff
@@ -489,6 +490,147 @@ func TestConstraintCheck(t *testing.T) {
...
+			s:               newStore(1, locUS),
+			sl:              sl3by3,
+			existingNodeIDs: []roachpb.NodeID{1, 4, 7},
+			expected:        false,
+		},
+		{
+			s:               newStore(1, locUS),
+			sl:              sl3by3,
+			existingNodeIDs: []roachpb.NodeID{1, 4, 5},
+			expected:        false,
```
a6802739: Why here `expected` should be false? From my understanding, we could rebalance to `locAS` to achieve better locality diversity.

pkg/storage/allocator_scorer_test.go
```diff
@@ -489,6 +490,147 @@ func TestConstraintCheck(t *testing.T) {
...
+			existingNodeIDs: []roachpb.NodeID{1, 4, 7, 10},
+			expected:        false,
+		},
+		{
+			// TODO(a-robinson/#17603): This is clearly not good. Rebalancing
+			// shouldn't be desired in this case, it'll just lead to thrashing.
+			s:               newStore(1, locUS),
+			sl:              sl4by3,
+			existingNodeIDs: []roachpb.NodeID{1, 2, 4, 7, 10},
+			expected:        true,
```
a6802739: How about we check if we rebalance to a target store, it will improve the overall diversity for this range?

In this methods, we should first get to know which store we are going to remove, and then calculates the overall diversityScore for each target rebalance store, and compare the diversityScore with the original overall diversity of the range. just like #19486.

a6802739 (commented):
@a-robinson  This is Great!

pkg/storage/allocator_scorer.go
a-robinson: I've added a test that ensures the method used to compute the current diversity and the method used to compute what the diversity would be if we added desc and removed store are directly comparable. I believe that resolves this, but do you still see anything wrong with it?

a-robinson (commented):
Thanks for the reviews! This should be good to go now that #19486 is in.

pkg/storage/allocator_scorer.go
```diff
@@ -697,6 +713,30 @@ func constraintCheck(store roachpb.StoreDescriptor, constraints config.Constrain
...
+// All below diversity-scoring methods should in theory be implemented by
+// calling into this one, but they aren't to avoid allocations.
+func rangeDiversityScore(existingNodeLocalities map[roachpb.NodeID]roachpb.Locality) float64 {
+	var sumScore float64
+	var numSamples int
+	for n1, l1 := range existingNodeLocalities {
+		for n2, l2 := range existingNodeLocalities {
+			// Only compare pairs of replicas where s2 > s1 to avoid computing the
+			// diversity score between each pair of localities twice.
+			if n2 <= n1 {
```
a6802739: Great idea!

pkg/storage/allocator_scorer_test.go
a6802739: Yeah, I see now. Thank you for your explanation.

pkg/storage/allocator_scorer_test.go
```diff
@@ -757,6 +903,64 @@ func TestRemovalDiversityScore(t *testing.T) {
...
+		{[]roachpb.StoreID{testStoreUSa15, testStoreUSa15Dupe, testStoreUSb, testStoreEurope}, 2.0 / 3.0},
+		{[]roachpb.StoreID{testStoreUSa15, testStoreUSa1, testStoreUSb, testStoreEurope}, 17.0 / 24.0},
+		{[]roachpb.StoreID{testStoreUSa15, testStoreUSa15Dupe, testStoreUSa1, testStoreUSb, testStoreEurope}, 3.0 / 5.0},
+	}
+
+	// Ensure that rangeDiversityScore and diversityRebalanceFromScore return
+	// the same results for the same configurations, enabling their results
+	// to be directly compared with each other. The same is not true for
+	// diversityAllocateScore and diversityRemovalScore as of their initial
+	// creation or else we would test them here as well.
```
a6802739: I could understand if the remove target and the rebalance target are same, `rangeDiversityScore` will be same with `diversityRebalanceFromScore`, but I think maybe we should have a test that we could achieve better diversity score for`diversityRebalanceFromScore` if we remove one replica and rebalance to one target comparing with `rangeDiversityScore`.

pkg/storage/allocator_scorer.go
```diff
@@ -550,14 +550,30 @@ func shouldRebalance(
...
 			store.StoreID, store.Capacity.FractionUsed(), store.Capacity)
 		return true
 	}
 
+	diversityScore := rangeDiversityScore(existingNodeLocalities)
+	for _, desc := range sl.stores {
+		if !preexistingReplicaCheck(desc.Node.NodeID, rangeInfo.Desc.Replicas) {
+			continue
+		}
+		otherScore := diversityRebalanceFromScore(desc, store.Node.NodeID, existingNodeLocalities)
```
a6802739: I think It's quite obvious now.

a6802739 (commented):
LGTM!

pkg/storage/allocator_scorer_test.go
a-robinson: I don't think I understand exactly what you're suggesting. It sounds to me like you're suggesting something along the lines of what I've already added in `TestShouldRebalanceDiversity`, which is testing that new code in `shouldRebalance` is able to use `rangeDiversityScore` and `diversityRebalanceFromScore` to properly determine when a rebalance would be beneficial for diversity. Could you explain in more detail what you're suggesting and what code it would be testing?

a6802739: Hmm, Sorry for my bad expression. What I mean is that maybe we could add a test. The test will display the result of both functions, and literally the result of `diversityRebalanceFromScore` is obviously bigger than that of `rangeDiversityScore`.
`TestShouldRebalanceDiversity` proves that the result of which function is bigger but doesn't display the result of both functions.
Maybe it's unnecessary, I just think it's a better way for other guys to understand.

a-robinson: Ok, I see now. That would help with understanding, but I don't think it provides any more code coverage. I'm going to skip it for now, but it would be a good test to have.

Issue #19427, opened 2017-10-22
Feature request: make CockroachDB survive datacentre reboots


jeremyvisser:
Currently, when you stop all CockroachDB nodes in a cluster, it is totally impossible to recover that cluster. This is because when starting a node, it attempts to find other nodes it has previously talked to. However, when it finds none, it simply hangs forever and never starts up, leading to a Catch-22 situation with all nodes deadlocked.

This is a total blocker for anything but the most basic of use cases, meaning something like a datacentre power outage requires reinitialising the cluster and restoring data from backup.

Please consider implementing a feature such that CockroachDB nodes can start up in a state where they are able to respond to requests from other nodes before they have found their own first node in order to bootstrap a quorum state.

Also please consider documenting this, as this situation is not documented anywhere and is suspiciously absent from any troubleshooting guides.

tschottdorf:
Thanks for filing this! It's correct that reasoning about cluster connectivity isn't as straightforward as we'd like to have it yet. We're discussing this over at https://github.com/cockroachdb/cockroach/issues/18027 and https://github.com/cockroachdb/cockroach/issues/18056.

Note though that for static deployments of smaller clusters specifying all join flags should work. In dynamically provisioned environments, there's usually a load balancer that can be pointed to that should also work (there are some remaining issues, but they are rare to hit in practice).

If you have a minute, could you briefly describe your deployment (cluster size, how nodes are provisioned, join flags for each node)?

a-robinson:
> Currently, when you stop all CockroachDB nodes in a cluster, it is totally impossible to recover that cluster. This is because when starting a node, it attempts to find other nodes it has previously talked to. However, when it finds none, it simply hangs forever and never starts up, leading to a Catch-22 situation with all nodes deadlocked.

That isn't true in most configurations. As @tschottdorf mentioned, there are situations where IP addresses / hostnames changing between restarts can cause problems if you don't change the join flags correspondingly, but it shouldn't affect standard situations like stopping and restarting all nodes/processes.

More info on your configuration that's having this problem would be helpful.

bdarnell:
This sounds like it might be the `--background` confusion - a node started with `--background` stays in the foreground until it has established contact with the cluster, so for a cold restart you need to start everything at once instead of one node at a time. Depending on how you're managing your deployment, some tools have difficulty with this pattern. For multi-node clusters, you may want to use shell backgrounding (add `&` at the end of the command) instead of `--background`.

[...]

PR #5441, opened 2016-03-21
storage: move node initialization to before serving

vivekmenezes:
This is not guaranteed to fix the bug but I'm marking it fixed for now. Let's see if we see it after this fix.

closes #5024 

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach/5441)

PR #19493, opened 2017-10-24
server: rework Gossip bootstrap sequence

### LINKED ISSUE #13027: gossip: Restarting a cluster with new hostnames/IPs doesn't work in many cases
This was found while testing out the restore functionality from a 10-node cluster backup onto a new set of 10 nodes. When starting up the cluster with three different configurations of `--join` parameters on the 10 nodes, something different happened each time:

1. With `--join=<node-1>` for nodes 2-10 and `--join=""` for node 1: none of the nodes initialized properly. They all got stuck waiting to be connected to the cluster via gossip.
1. With `--join=<node-(i-1)>` for nodes i=2 to i=10, and `--join=""` for node 1: nodes 4-10 initialized properly. Nodes 1-3 got stuck waiting to be connected to the cluster via gossip.
1. With `--join=<comma-separated-list-of-all-nodes>` for all the nodes in the cluster, including node 1: the cluster started up properly.

Because all of the nodes were new, it's worth noting that none of the addresses saved from the gossip network before the backup were usable on the new nodes. Thus, while all the nodes had a store full of data, they didn't know where the other nodes were other than what they got in the `--join` parameter.

While the details vary a little (see below), the core issue in both of the first two cases is that nodes don't serving incoming connections until they have initialized themselves, where "initialized themselves" effectively means that the node is connected enough to know the cluster ID. This means that only nodes with a copy of the first range can properly bootstrap themselves if they're unable to connect to another node, because they're the only ones that will add the cluster ID to their local gossip infostore.

In the first case above, because node 1 didn't have a copy of range 1, it is unable to initialize itself. Because it doesn't get initialized, it doesn't start accepting incoming connections, and thus it doesn't get the cluster ID even though it has incoming connections from 3 nodes that do have the cluster ID (the 3 that have a copy of range 1). Thus the cluster never gets going -- 3 nodes are up and running in some sense, but don't know about each other so they can't do anything.

In the second case above, nodes 4-10 work because node 4 is the lowest node ID that had a copy of range 1. Thus, it initializes itself and accepts the incoming connection from node 5, which allowed node 5 to initialize, which allowed node 6 to initialize, etc. Nodes 1-3 are stuck forever because they have no outgoing gossip connection to any of nodes 4-10.

In the third case above, all nodes work because they all attempt to connect to at least one node that has a copy of range 1.

-------------

It would be conceivable to say that the current state of things is ok, and that you should specify all the nodes in all the join flags (or at the very least ensure that it's a fully-connected directed graph). That seems a bit prescriptive and error-prone, though. If a node was able to learn from incoming gossip connections during startup, pretty much any configuration of `--join` flags would work.

@mjibson @spencerkimball

### LINKED ISSUE #18027: acceptance: local cluster is blocked on problems with empty join flag (#13027)
See https://github.com/cockroachdb/cockroach/pull/18007.

`TestMixedVersion` runs with three nodes as otherwise it's flaky: the node first restarted (i.e. the one without join flags) never manages to connect to gossip. See the comment within that test.

One theory is that the other three nodes manage to connect among each other sufficiently so that the don't feel that establishing new connections is helpful. In turn, the first node remains alone and unable to do anything about it. The theory is easy to test by running the test with a version of cockroach that always connects to everything it can.

It repros easily using `time go test -v -tags acceptance ./pkg/acceptance/ -run Mixed  -show-logs -count 100 -timeout 1h` within only a few iters (after making the test 4nodes, of course).

tbg:
This is a WIP because I want to add the script from #19427, which reproduces
this problem with 100% accuracy, as an acceptance test.

----

A long time ago, in #5441, we changed the server startup sequence so that we
would only open the main listener at the very end of the startup sequence. The
impetus was that by opening the listener early, various services could receive
requests when they weren't actually ready, which resulted in unexpected errors
during tests.

What we missed at the time was that this isn't actually a good idea because it
also robs the node of the ability to accept incoming Gossip connections. As a
result, if a node didn't manage to talk to Gossip peers (perhaps it was started
without a join flag and had nothing reasonable persisted), it would never get
its Gossip connected and would never leave `(*Node).start` which has to wait for
Gossip.

These deadlocks worsened the impact of some of our other insufficiently
streamlined aspects of the startup sequence, namely the choice of `--join`
flags, and were generally hard to debug especially since the HTTP endpoints, and
thus many debugging facilities, weren't available.

This commit reworks the sequence so that

- the HTTP endpoint opens early (so that the debug endpoints are accessible)
- opens the main cmux listener early
    - but delays opening the Postgres sublistener
    - but fails all non-Gossip and non-Ping gRPC traffic retryably

Fixes #18027.
Fixes #13027.

### DIFF
--- a/pkg/rpc/context.go
+++ b/pkg/rpc/context.go
@@ -15,7 +15,6 @@
 package rpc
 
 import (
-	"errors"
 	"fmt"
 	"math"
 	"net"
@@ -25,6 +24,7 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
 	"github.com/rubyist/circuitbreaker"
 	"golang.org/x/net/context"
 	"golang.org/x/sync/syncmap"
@@ -92,6 +92,15 @@ func spanInclusionFunc(
 // NewServer is a thin wrapper around grpc.NewServer that registers a heartbeat
 // service.
 func NewServer(ctx *Context) *grpc.Server {
+	return NewServerWithInterceptor(ctx, nil)
+}
+
+// NewServerWithInterceptor is like NewServer, but accepts an additional
+// interceptor which is called before streaming and unary RPCs and may inject an
+// error.
+func NewServerWithInterceptor(
+	ctx *Context, interceptor func(fullMethod string) error,
+) *grpc.Server {
 	opts := []grpc.ServerOption{
 		// The limiting factor for lowering the max message size is the fact
 		// that a single large kv can be sent over the network in one message.
@@ -131,15 +140,64 @@ func NewServer(ctx *Context) *grpc.Server {
 		}
 		opts = append(opts, grpc.Creds(credentials.NewTLS(tlsConfig)))
 	}
+
+	var unaryInterceptor grpc.UnaryServerInterceptor
+	var streamInterceptor grpc.StreamServerInterceptor
+
 	if tracer := ctx.AmbientCtx.Tracer; tracer != nil {
 		// We use a SpanInclusionFunc to save a bit of unnecessary work when
 		// tracing is disabled.
-		interceptor := otgrpc.OpenTracingServerInterceptor(
+		unaryInterceptor = otgrpc.OpenTracingServerInterceptor(
 			tracer,
 			otgrpc.IncludingSpans(otgrpc.SpanInclusionFunc(spanInclusionFunc)),
 		)
-		opts = append(opts, grpc.UnaryInterceptor(interceptor))
+		// TODO(tschottdorf): should set up tracing for stream-based RPCs as
+		// well. The otgrpc package has no such facility, but there's also this:
+		//
+		// https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/tracing/opentracing
 	}
+
+	// TODO(tschottdorf): when setting up the interceptors below, could make the
+	// functions a wee bit more performant by hoisting some of the nil checks
+	// out. Doubt measurements can tell the difference though.
+
+	if interceptor != nil {
+		prevUnaryInterceptor := unaryInterceptor
+		unaryInterceptor = func(
+			ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler,
+		) (interface{}, error) {
+			if err := interceptor(info.FullMethod); err != nil {
+				return nil, err
+			}
+			if prevUnaryInterceptor != nil {
+				return prevUnaryInterceptor(ctx, req, info, handler)
+			}
+			return handler(ctx, req)
+		}
+	}
+
+	if interceptor != nil {
+		prevStreamInterceptor := streamInterceptor
+		streamInterceptor = func(
+			srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler,
+		) error {
+			if err := interceptor(info.FullMethod); err != nil {
+				return err
+			}
+			if prevStreamInterceptor != nil {
+				return prevStreamInterceptor(srv, stream, info, handler)
+			}
+			return handler(srv, stream)
+		}
+	}
+
+	if unaryInterceptor != nil {
+		opts = append(opts, grpc.UnaryInterceptor(unaryInterceptor))
+	}
+	if streamInterceptor != nil {
+		opts = append(opts, grpc.StreamInterceptor(streamInterceptor))
+	}
+
 	s := grpc.NewServer(opts...)
 	RegisterHeartbeatServer(s, &HeartbeatService{
 		clock:              ctx.LocalClock,
--- a/pkg/acceptance/allocator_test.go
+++ b/pkg/acceptance/allocator_test.go
@@ -93,7 +93,7 @@ func (at *allocatorTest) Run(ctx context.Context, t *testing.T) {
 	if err := at.f.Resize(at.StartNodes); err != nil {
 		t.Fatal(err)
 	}
-	if err := CheckGossip(ctx, at.f, longWaitTime, HasPeers(at.StartNodes)); err != nil {
+	if err := CheckGossip(ctx, at.f, waitTime, HasPeers(at.StartNodes)); err != nil {
 		t.Fatal(err)
 	}
 	at.f.Assert(ctx, t)
@@ -152,7 +152,7 @@ func (at *allocatorTest) Run(ctx context.Context, t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if err := CheckGossip(ctx, at.f, longWaitTime, HasPeers(at.EndNodes)); err != nil {
+	if err := CheckGossip(ctx, at.f, waitTime, HasPeers(at.EndNodes)); err != nil {
 		t.Fatal(err)
 	}
 	at.f.Assert(ctx, t)
--- a/pkg/acceptance/continuous_load_test.go
+++ b/pkg/acceptance/continuous_load_test.go
@@ -56,6 +56,8 @@ type continuousLoadTest struct {
 	Process string
 }
 
+const longWaitTime = 2 * time.Minute
+
 // queryCount returns the total SQL queries executed by the cluster.
 func (cl continuousLoadTest) queryCount(f *terrafarm.Farmer) (float64, error) {
 	var client http.Client
--- a/pkg/acceptance/gossip_peerings_test.go
+++ b/pkg/acceptance/gossip_peerings_test.go
@@ -15,7 +15,9 @@
 package acceptance
 
 import (
+	"io/ioutil"
 	"math/rand"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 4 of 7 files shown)


bdarnell:
LGTM (I haven't read line-by-line but I assume it's basically just code movement)

Some more important context is that in 1.0, the server didn't open its port until it was ready to serve, so TCP-level health checks were sufficient. With the introduction of the `init` command in 1.1, a not-yet-initialized server will open its port early (which means that higher-level health checks are required for determining readiness). Now that we do that, we might as well also have initialized-but-not-connected servers do the same. 

---

---

*[pkg/server/server.go, line 200 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxEUJk9mjpkBT0KkHXu:-KxEUJk9mjpkBT0KkHXv:bdm1u12) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e533905830421c082ce6b839a9f96a374a686ef/pkg/server/server.go#L200)):*
> ```Go
> 
> 	{
> 		var earlyRPCWhiteList = map[string]struct{}{
> ```

Can we unify this with the init server? It would be nice to only have one version of this interceptor. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-:-KxETA_7wLO5oS9GWDxs:b-fyu2xs)*

a-robinson:
:lgtm:, but ditto on having glossed over some of the code movement. Have you verified that this fixes situations like your gist in #19427?

---

Reviewed 2 of 2 files at r1.

---

*[pkg/server/server.go, line 1103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxEqE8-02H9oX3qlvJe:-KxEqE8-02H9oX3qlvJf:b-r0v0u8) ([raw file](https://github.com/cockroachdb/cockroach/blob/26030c8a608ab8c8b5969840aef896e24d4aec80/pkg/server/server.go#L1103)):*
> 19427

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-:-KxEqSkoCpR_SXeJo147:b-chszww)*

tschottdorf:
> Have you verified that this fixes situations like your gist in #19427?

Yep, by now I finished translating that into an acceptance test, too.

> Some more important context is that in 1.0, the server didn't open its port until it was ready to serve, so TCP-level health checks were sufficient. With the introduction of the init command in 1.1, a not-yet-initialized server will open its port early (which means that higher-level health checks are required for determining readiness). Now that we do that, we might as well also have initialized-but-not-connected servers do the same.

Anything actionable in this PR? Should the health endpoint be opened later?

PTAL at the new test.

---

Review status: 1 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/server/server.go, line 200 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxEUJk9mjpkBT0KkHXu:-KxJ-8cyMxp1RrqtRUQ-:bey20oo) ([raw file](https://github.com/cockroachdb/cockroach/blob/6e533905830421c082ce6b839a9f96a374a686ef/pkg/server/server.go#L200)):*

Can we unify this with the init server? It would be nice to only have one version of this interceptor. 
</blockquote></details>

Good idea, though I'll file this for now: https://github.com/cockroachdb/cockroach/issues/19523

---

*[pkg/server/server.go, line 1103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxEqE8-02H9oX3qlvJe:-KxIqmqu6B1i-yboZuSf:b-4wrawl) ([raw file](https://github.com/cockroachdb/cockroach/blob/26030c8a608ab8c8b5969840aef896e24d4aec80/pkg/server/server.go#L1103)):*

> 19427
</blockquote></details>

Happy little accident?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-:-KxIpDa8a_HCDoVtU2m8:b-pwalof)*

bdarnell:
No, I wasn't suggesting any changes to this PR, but just thought that comment might be worth preserving somewhere (maybe in the commit message?)

---

Reviewed 1 of 1 files at r2, 4 of 5 files at r3, 1 of 1 files at r4.

---

*[pkg/acceptance/gossip_peerings_test.go, line 189 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxJ56yfX6Vi3vLVhOLM:-KxJ56yfX6Vi3vLVhOLN:b-15caap) ([raw file](https://github.com/cockroachdb/cockroach/blob/d70cd316867853589727289ded96650177bc2781/pkg/acceptance/gossip_peerings_test.go#L189)):*
> ```Go
> 
> // This test verifies that when the first node is restarted and has no valid
> // join flags or persisted Gossip bootstrap records, it will still be able to
> ```

Why doesn't the first node have persisted gossip bootstrap records?

---

*[pkg/acceptance/gossip_peerings_test.go, line 217 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxJ4m76QVbCHBEH-uUi:-KxJ4m76QVbCHBEH-uUj:b-qcuu2k) ([raw file](https://github.com/cockroachdb/cockroach/blob/d70cd316867853589727289ded96650177bc2781/pkg/acceptance/gossip_peerings_test.go#L217)):*
> ```Go
> 
> 	lc := c.(*localcluster.LocalCluster)
> 	// Upgrade the first node's binary to match the other nodes (i.e. the testing binary).
> ```

Copy-pasta comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-:-KxJ3mcWKiRbAxvui__F:b-bb99e1)*

tschottdorf:
Ack, done. TFTRs!

---

---

*[pkg/acceptance/gossip_peerings_test.go, line 189 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxJ56yfX6Vi3vLVhOLM:-KxJ7QtFOO_YcDoJSw-F:blm6a8h) ([raw file](https://github.com/cockroachdb/cockroach/blob/d70cd316867853589727289ded96650177bc2781/pkg/acceptance/gossip_peerings_test.go#L189)):*

Why doesn't the first node have persisted gossip bootstrap records?
</blockquote></details>

The test sets up things so that the first node's bootstrap info is all outdated. Added comment.

---

*[pkg/acceptance/gossip_peerings_test.go, line 217 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxJ4m76QVbCHBEH-uUi:-KxJ7a3DsAxatAdE9mYS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d70cd316867853589727289ded96650177bc2781/pkg/acceptance/gossip_peerings_test.go#L217)):*

Copy-pasta comment?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-:-KxJ8Tcje2IbPmIGwOza:b-swe23a)*

a-robinson:
Review status: 2 of 7 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/server/server.go, line 1103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493#-KxEqE8-02H9oX3qlvJe:-KxJ7T_92KiXHBIkLtyZ:b-rgzsmz) ([raw file](https://github.com/cockroachdb/cockroach/blob/26030c8a608ab8c8b5969840aef896e24d4aec80/pkg/server/server.go#L1103)):*

Happy little accident?
</blockquote></details>

"Happy little accident" could be a good replacement for the shame cube

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19493)*

pkg/rpc/context.go
```diff
@@ -131,15 +140,64 @@ func NewServer(ctx *Context) *grpc.Server {
...
+		// well. The otgrpc package has no such facility, but there's also this:
+		//
+		// https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/tracing/opentracing
+	}
+
+	// TODO(tschottdorf): when setting up the interceptors below, could make the
+	// functions a wee bit more performant by hoisting some of the nil checks
+	// out. Doubt measurements can tell the difference though.
+
+	if interceptor != nil {
```
tamird: Some of this code might benefit from https://godoc.org/github.com/grpc-ecosystem/go-grpc-middleware ?

tschottdorf:
@bdarnell what are your thoughts on cherry-picking this to 1.1.2? I think it can wait, though the issue that is being fixed is sure to cause a lot of problems when it occurs.

bdarnell:
I'm reluctant to make this change in a patch release; this is a potentially disruptive change (see #19559. It could also require changes to LB configurations). However, this depends on the availability of other solutions, since the problem being solved is pretty severe. If you follow best practices in setting join flags, how likely are you to run into this problem? Does it only occur if there are IP changes, or could it be more common?

a-robinson:
To me, this seems too big/risky of a change to be patched into 1.1.

As far as I'm aware, the only situations we've constructed in which it becomes a problem are when IPs or ports change, which shouldn't be common in real deployments.

tschottdorf:
I'm ok letting this bake until 2.0. I don't think you run into this much outside of the intentionally exotic things our acceptance tests do.

PR #19497, opened 2017-10-24
ui: require Yarn 1.0+

benesch:
99b35b3 accidentally introduced a dependency on Yarn 1.0+ with its usage
of `yarn --cwd`. Since Yarn 1.0 has been available since early
September, it seems reasonable to mandate its usage.


couchand (approved):
Seems like a reasonable period has passed, and this qualifies for Ben's criteria of needing an actual reason.

Since we're requiring this, can you remove the `--` on the script invocations that's throwing warnings?

a-robinson (approved):
Thanks, I just ran into this.

benesch:
@couchand 99b35b3 removed usage of `yarn run` entirely. You're not still seeing that `--` deprecation warning, are you?

couchand:
So you're right.  Nevermind, ignore me.

Issue #19237, opened 2017-10-13
sql: integer alias array columns fall back to `INT`


justinj:
`CREATE TABLE x (a INT2[])` is treated the same as `CREATE TABLE x (a INT[])`:

```
root@:26257/d> CREATE TABLE x (a INT2[]);
CREATE TABLE

Time: 5.542286ms

root@:26257/d> SHOW CREATE TABLE x;
+-------+----------------------------------+
| Table |           CreateTable            |
+-------+----------------------------------+
| x     | CREATE TABLE x (␤                |
|       |     a INT[] NULL,␤               |
|       |     FAMILY "primary" (a, rowid)␤ |
|       | )                                |
+-------+----------------------------------+
(1 row)

Time: 9.964743ms
```

PR #19498, opened 2017-10-24
sql,cli: support arrays in dump

### LINKED ISSUE #19487: dump error on ARRAY
```
cockroach dump test --certs-dir=/home/cockroach/certs/ --host 192.168.1.54 > test.sql
```
there is dump command.

```
create table test(
   values string array not null default array[]
);
insert into test(values) values(array['a','b','c']);
```
it works.

```
create table test(
   values int array not null default array[]
);
no data inserted.
```
it works.

```
create table test(
   values int array not null default array[]
);
insert into test(values) values(array[1,2,3,4,5]);
```
it didn't work.

error:

```
E171024 13:54:35.289938 1 util/log/crash_reporting.go:117  a panic has occurred!
panic: unknown []byte type: {}, values: INT[] [recovered]
        panic: unknown []byte type: {}, values: INT[]
goroutine 1 [running]:
github.com/cockroachdb/cockroach/pkg/util/log.RecoverAndReportPanic(0x7f3a7b3d48d8, 0xc420014218, 0xc420432000)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/crash_reporting.go:99 +0x9f
panic(0x1b44e00, 0xc4202f04e0)
        /usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/cli.dumpTableData(0x2ab4620, 0xc42000e018, 0xc420469a80, 0xc420271880, 0x1e, 0x174, 0xc4206cc580, 0x1, 0xc42026f330, 0x2, ...)
        /go/src/github.com/cockroachdb/cockroach/pkg/cli/dump.go:478 +0x126d
github.com/cockroachdb/cockroach/pkg/cli.runDump(0x2a9d9e0, 0xc420428f80, 0x1, 0x4, 0x0, 0x0)
        /go/src/github.com/cockroachdb/cockroach/pkg/cli/dump.go:113 +0xa3a
github.com/cockroachdb/cockroach/pkg/cli.MaybeDecorateGRPCError.func1(0x2a9d9e0, 0xc420428f80, 0x1, 0x4, 0x0, 0x0)
        /go/src/github.com/cockroachdb/cockroach/pkg/cli/error.go:35 +0x5a
github.com/cockroachdb/cockroach/vendor/github.com/spf13/cobra.(*Command).execute(0x2a9d9e0, 0xc420428e80, 0x4, 0x4, 0x2a9d9e0, 0xc420428e80)
        /go/src/github.com/cockroachdb/cockroach/vendor/github.com/spf13/cobra/command.go:644 +0x3f8
github.com/cockroachdb/cockroach/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x2a9b820, 0x5d876e, 0xc42027f390, 0xc42072ff10)
        /go/src/github.com/cockroachdb/cockroach/vendor/github.com/spf13/cobra/command.go:734 +0x339
github.com/cockroachdb/cockroach/vendor/github.com/spf13/cobra.(*Command).Execute(0x2a9b820, 0xc42072ff00, 0xc42027f360)
        /go/src/github.com/cockroachdb/cockroach/vendor/github.com/spf13/cobra/command.go:693 +0x2b
github.com/cockroachdb/cockroach/pkg/cli.Run(0xc4200100d0, 0x5, 0x5, 0xc420014218, 0xc420432000)
        /go/src/github.com/cockroachdb/cockroach/pkg/cli/cli.go:141 +0x6d
github.com/cockroachdb/cockroach/pkg/cli.Main()
        /go/src/github.com/cockroachdb/cockroach/pkg/cli/cli.go:50 +0x18d
main.main()
        /go/src/github.com/cockroachdb/cockroach/main.go:26 +0x20
```

justinj:
Fixes #19487.

I neglected to add the array handling case to dump. This commit adds it.
Added an `EachColType` function that will hopefully be able to kept up
to date which can be used to make tests which will continue to test
new column types as more get added.

### DIFF
--- a/pkg/cli/dump.go
+++ b/pkg/cli/dump.go
@@ -433,7 +433,18 @@ func dumpTableData(w io.Writer, conn *sqlConn, clusterTS string, md tableMetadat
 					default:
 						// STRING and DECIMAL types can have optional length
 						// suffixes, so only examine the prefix of the type.
-						if strings.HasPrefix(md.columnTypes[cols[si]], "STRING") {
+						// In addition, we can only observe ARRAY types by their [] suffix.
+						if strings.HasSuffix(md.columnTypes[cols[si]], "[]") {
+							typ := strings.TrimRight(md.columnTypes[cols[si]], "[]")
+							elemType, err := parser.StringToColType(typ)
+							if err != nil {
+								return err
+							}
+							d, err = parser.ParseDArrayFromString(parser.NewTestingEvalContext(), string(t), elemType)
+							if err != nil {
+								return err
+							}
+						} else if strings.HasPrefix(md.columnTypes[cols[si]], "STRING") {
 							d = parser.NewDString(string(t))
 						} else if strings.HasPrefix(md.columnTypes[cols[si]], "DECIMAL") {
 							d, err = parser.ParseDDecimal(string(t))
--- a/pkg/cli/dump_test.go
+++ b/pkg/cli/dump_test.go
@@ -59,11 +59,12 @@ func TestDumpRow(t *testing.T) {
 		e decimal,
 		u uuid,
 		ip inet,
+		ary string[],
 		tz timestamptz,
 		e1 decimal(2),
 		e2 decimal(2, 1),
 		s1 string(1),
-		FAMILY "primary" (i, f, d, t, n, o, u, ip, tz, e1, e2, s1, rowid),
+		FAMILY "primary" (i, f, d, t, n, o, u, ip, ary, tz, e1, e2, s1, rowid),
 		FAMILY fam_1_s (s),
 		FAMILY fam_2_b (b),
 		FAMILY fam_3_e (e)
@@ -78,6 +79,7 @@ func TestDumpRow(t *testing.T) {
 		1.2345,
 		'e9716c74-2638-443d-90ed-ffde7bea7d1d',
 		'192.168.0.1',
+		ARRAY['hello','world'],
 		'2016-01-25 10:10:10',
 		3.4,
 		4.5,
@@ -113,22 +115,23 @@ CREATE TABLE t (
 	e DECIMAL NULL,
 	u UUID NULL,
 	ip INET NULL,
+	ary STRING[] NULL,
 	tz TIMESTAMP WITH TIME ZONE NULL,
 	e1 DECIMAL(2) NULL,
 	e2 DECIMAL(2,1) NULL,
 	s1 STRING(1) NULL,
-	FAMILY "primary" (i, f, d, t, n, o, u, ip, tz, e1, e2, s1, rowid),
+	FAMILY "primary" (i, f, d, t, n, o, u, ip, ary, tz, e1, e2, s1, rowid),
 	FAMILY fam_1_s (s),
 	FAMILY fam_2_b (b),
 	FAMILY fam_3_e (e)
 );
 
-INSERT INTO t (i, f, s, b, d, t, n, o, e, u, ip, tz, e1, e2, s1) VALUES
-	(1, 2.3, 'striiing', '\x613162326333', '2016-03-26', '2016-01-25 10:10:10+00:00', '2h30m30s', true, 1.2345, 'e9716c74-2638-443d-90ed-ffde7bea7d1d', '192.168.0.1', '2016-01-25 10:10:10+00:00', 3, 4.5, 's'),
-	(NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL),
-	(NULL, '+Inf', NULL, NULL, NULL, NULL, NULL, NULL, 'Infinity', NULL, NULL, NULL, NULL, NULL, NULL),
-	(NULL, '-Inf', NULL, NULL, NULL, NULL, NULL, NULL, '-Infinity', NULL, NULL, NULL, NULL, NULL, NULL),
-	(NULL, 'NaN', NULL, NULL, NULL, NULL, NULL, NULL, 'NaN', NULL, NULL, NULL, NULL, NULL, NULL);
+INSERT INTO t (i, f, s, b, d, t, n, o, e, u, ip, ary, tz, e1, e2, s1) VALUES
+	(1, 2.3, 'striiing', '\x613162326333', '2016-03-26', '2016-01-25 10:10:10+00:00', '2h30m30s', true, 1.2345, 'e9716c74-2638-443d-90ed-ffde7bea7d1d', '192.168.0.1', ARRAY['hello':::STRING,'world':::STRING], '2016-01-25 10:10:10+00:00', 3, 4.5, 's'),
+	(NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL),
+	(NULL, '+Inf', NULL, NULL, NULL, NULL, NULL, NULL, 'Infinity', NULL, NULL, NULL, NULL, NULL, NULL, NULL),
+	(NULL, '-Inf', NULL, NULL, NULL, NULL, NULL, NULL, '-Infinity', NULL, NULL, NULL, NULL, NULL, NULL, NULL),
+	(NULL, 'NaN', NULL, NULL, NULL, NULL, NULL, NULL, 'NaN', NULL, NULL, NULL, NULL, NULL, NULL, NULL);
 `
 
 	if out != expect {
--- a/pkg/sql/parser/parse_array.go
+++ b/pkg/sql/parser/parse_array.go
@@ -144,6 +144,41 @@ func (p *parseState) parseElement() error {
 	return p.result.Append(d)
 }
 
+// StringToColType returns a column type given a string representation of the
+// type. Used by dump.
+func StringToColType(s string) (ColumnType, error) {
+	switch s {
+	case "BOOL":
+		return boolColTypeBool, nil
+	case "INT":
+		return intColTypeInt, nil
+	case "FLOAT":
+		return floatColTypeFloat, nil
+	case "DECIMAL":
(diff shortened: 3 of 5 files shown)


mjibson:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/cli/dump.go, line 436 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEZoaReoELI0JwepaL:-KxEZoaReoELI0JwepaM:bu5nouz) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/cli/dump.go#L436)):*
> ```Go
> 						// STRING and DECIMAL types can have optional length
> 						// suffixes, so only examine the prefix of the type.
> 						// In addition, we can only observe ARRAY types by their [] suffix.
> ```

Although I think the following code is correct, we could consider switching this whole thing to `crdb_internal.table_columns`, which has, I think, slightly better support for this. Also not opposed to expanding crdb_internal.table_columns to explicitly add columns for things like semantic_type and array_contents (as listed in the column_type field, but names could change) to make this easier. Summary is I'd like us to stop doing string parsing like this. However this can all happen in a later PR or never if no one cares.

---

*[pkg/sql/parser/col_types.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbjAHm_JUcRTc7h_G:-KxEbjAHm_JUcRTc7h_H:b-e8k1em) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/parser/col_types.go#L609)):*
> ```Go
> // EachColType iterates over all column types.
> // This is not exhaustive yet, due to #19237.
> func EachColType(f func(ColumnType)) {
> ```

Can we use parser.TypesAnyNonArray here instead? Otherwise this feels like just another list no one will maintain.

---

*[pkg/sql/sqlbase/table.go, line 1783 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbMYrLxEuO217QEfu:-KxEbMYrLxEuO217QEfv:bl5t57n) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/sqlbase/table.go#L1783)):*
> ```Go
> 	case parser.TypeDecimal:
> 		return encoding.Decimal, nil
> 	case parser.TypeBytes, parser.TypeString, parser.TypeName:
> ```

What problem did not including parser.TypeName cause?

---

*[pkg/sql/sqlbase/table.go, line 1785 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEazKOtlsJ_O7bu2OM:-KxEazKOtlsJ_O7bu2ON:b-f920wd) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/sqlbase/table.go#L1785)):*
> ```Go
> 	case parser.TypeBytes, parser.TypeString, parser.TypeName:
> 		return encoding.Bytes, nil
> 	case parser.TypeBytes:
> ```

I think this case is a duplicate and can go away.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498)*

justinj:

---

*[pkg/cli/dump.go, line 436 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEZoaReoELI0JwepaL:-KxEeKkRqesKfjZmBE66:bo6anzh) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/cli/dump.go#L436)):*

Although I think the following code is correct, we could consider switching this whole thing to `crdb_internal.table_columns`, which has, I think, slightly better support for this. Also not opposed to expanding crdb_internal.table_columns to explicitly add columns for things like semantic_type and array_contents (as listed in the column_type field, but names could change) to make this easier. Summary is I'd like us to stop doing string parsing like this. However this can all happen in a later PR or never if no one cares.
</blockquote></details>

Very strong agree. I'll open an issue for this.

---

*[pkg/sql/parser/col_types.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbjAHm_JUcRTc7h_G:-KxEeOWWgG4wmiqAeRC7:b-4nao74) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/parser/col_types.go#L609)):*

Can we use parser.TypesAnyNonArray here instead? Otherwise this feels like just another list no one will maintain.
</blockquote></details>

`TypesAnyNonArray` is a list of `parser.Type`s, which is different from column types (since we eventually want to add things like ints with different widths, etc, to this list). I agree that I'm nervous that this won't be maintained adequately, but I don't know a better solution.

---

*[pkg/sql/sqlbase/table.go, line 1783 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbMYrLxEuO217QEfu:-KxEeexpYpM-SWYPdyF6:b-jwyl3y) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/sqlbase/table.go#L1783)):*

What problem did not including parser.TypeName cause?
</blockquote></details>

This was semi-unrelated, but was uncovered by adding this test, this had an issue with encoding `NAME[]`s.

---

*[pkg/sql/sqlbase/table.go, line 1785 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEazKOtlsJ_O7bu2OM:-KxEel9s9xjUQJ8u5efO:b-pjfe1e) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/sqlbase/table.go#L1785)):*

I think this case is a duplicate and can go away.
</blockquote></details>

Good call, missed it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498)*

mjibson:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/parser/col_types.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbjAHm_JUcRTc7h_G:-KxEqYU7W9a-qMbAZK38:b1ju09x) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/parser/col_types.go#L609)):*

`TypesAnyNonArray` is a list of `parser.Type`s, which is different from column types (since we eventually want to add things like ints with different widths, etc, to this list). I agree that I'm nervous that this won't be maintained adequately, but I don't know a better solution.
</blockquote></details>

I guess my argument is that until we have canonical source for all column types, we shouldn't add this code or test, because it's not going to solve the problem. If someone adds a new type but forgets to add it to this list, then dump still won't work for that type. Since that problem isn't solved by this code, I don't see the point in adding it. It it solving another problem?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498)*

justinj:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/parser/col_types.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbjAHm_JUcRTc7h_G:-KxErS699xnn00xNNsvL:b-oguww6) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/parser/col_types.go#L609)):*

I guess my argument is that until we have canonical source for all column types, we shouldn't add this code or test, because it's not going to solve the problem. If someone adds a new type but forgets to add it to this list, then dump still won't work for that type. Since that problem isn't solved by this code, I don't see the point in adding it. It it solving another problem?
</blockquote></details>

Ok, that's fair. I'd like to push for indeed *having* a canonical source for all column types in the future, ideally at least linted to be comprehensive (for some meaning of comprehensive), but I'm not sure what that would look like.

I do think it's solving another problem in that it will make writing certain other kinds of tests easier. I'd like to delete all the repetition I added in the `array` logic test, and currently array testing is `O(# features * # column types)` which can make tests hard to write and to understand. But I do agree that it's a larger question given the possibility that it will fall out of sync, and out of scope for this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498)*

mjibson:
:lgtm:

---

---

*[pkg/sql/parser/col_types.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-KxEbjAHm_JUcRTc7h_G:-KxEwXlb8cb-SVpPZVK2:bmc7s5p) ([raw file](https://github.com/cockroachdb/cockroach/blob/07781afbf075bae38e3821d7212766a3f7614f54/pkg/sql/parser/col_types.go#L609)):*

Ok, that's fair. I'd like to push for indeed *having* a canonical source for all column types in the future, ideally at least linted to be comprehensive (for some meaning of comprehensive), but I'm not sure what that would look like.

I do think it's solving another problem in that it will make writing certain other kinds of tests easier. I'd like to delete all the repetition I added in the `array` logic test, and currently array testing is `O(# features * # column types)` which can make tests hard to write and to understand. But I do agree that it's a larger question given the possibility that it will fall out of sync, and out of scope for this PR.
</blockquote></details>

Yes, I fully agree we need this test, because we keep hitting this problem, so long as it's future-proof.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19498#-:-KxEwg4oCt1DuV1Xn_9Y:bnfp4nl)*

PR #19501, opened 2017-10-24
build: introduce maintainer-clean and unsafe-clean targets

benesch:
Teach our build system to better handle cleaning with the introduction
of three new cleaning targets. The distinctions between `clean`,
`maintainer-clean`, and `unsafe-clean` are as follows:

  * clean: remove all easily-regenerated files, where an
    "easily-regenerated" file only requires the build dependencies
    listed on the user-facing "Build from Source" install instructions.
    At present, that's a C++ compiler, a Go compiler, Bash, CMake, and
    autoconf. The cockroach binary, log files, C/C++ dependencies, Go
    compilation artifacts, and generated protobuf code are all fair
    game.

  * maintainer-clean: like clean, but also remove hard-to-regenerate
    files. At present, this means the UI build artifacts, which require
    Node and Yarn to regenerate.

  * unsafe-clean: like maintainer-clean, but also scrub the repository
    and its submodules of any untracked or ignored files.


knz:
:lgtm: but I'm not knowledgeable with the UI artifacts, so you'll want someone else to confirm.

---

Reviewed 5 of 6 files at r1.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19501#-:-KxEwUdcNI-k-C-fV7yj:b-rd5o88)*

benesch:
cc @couchand @mrtracy

couchand:
This makes me wonder if we should update the build from source instructions to require node & yarn?  What do we mean by "build from source"?

couchand:
but otherwise LGTM

benesch:
"Build from source" means users who click on that tab on https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html. We intentionally keep those dependencies as slim as possible; users building from source are usually doing so because we don't provide pre-built binaries for their platform/architecture, not because they want to. (I've been meaning to remove autoconf from that list, actually. )

Contributors, by contrast, are required to install Node and Yarn. (If they read CONTRIBUTING.md, at least.)

benesch:
TFTRs!

PR #19504, opened 2017-10-24
util/contextutil: silence overly verbose log message

petermattis:
Logging "canceling context" every time a context is cancelled is overly
verbose and a measurable performance hit. Only wrap the cancellation
closure if log.V(1) or higher is specified.


petermattis:
@mjibson You'll have to specify `--vmodule=context.go=1` or `--vmodule=context.go=2`. Hopefully that isn't too burdensome.

madelynnblue (approved):
Yes, I sadly agree this is required.

PR #19506, opened 2017-10-25
build: fix verify-archive

benesch:
verify-archive has been broken since August, but is quite useful
ensuring we don't accidentally introduce new build-time dependencies.

Fix it up by removing the Git dependency from compiling protobufs.


benesch:
TeamCity's verify archive build: https://teamcity.cockroachdb.com/viewType.html?buildTypeId=Cockroach_UnitTests_VerifyArchive&branch_Cockroach_UnitTests=19506&tab=buildTypeStatusDiv

build/protobuf.mk
```diff
@@ -30,7 +30,7 @@ PROTOC_PLUGIN   := $(LOCAL_BIN)/protoc-gen-$(PLUGIN_SUFFIX)
 GOGOPROTO_PROTO := $(GOGO_PROTOBUF_PATH)/gogoproto/gogo.proto
 
 COREOS_PATH := $(GITHUB_ROOT)/coreos
-COREOS_RAFT_PROTOS := $(addprefix $(COREOS_PATH)/etcd/raft/, $(sort $(shell git -C $(COREOS_PATH)/etcd/raft ls-files --exclude-standard --cached --others -- '*.proto')))
+COREOS_RAFT_PROTOS := $(sort $(shell find $(COREOS_PATH)/etcd/raft -type f -name '*.proto'))
```
lego: OOC, any reason for the change to find? Performance, consistency, or clarity?

build/verify-archive.sh
```diff
@@ -2,8 +2,9 @@
 
 # This script sanity-checks a source tarball, assuming a Debian-based Linux
 # environment with a Go version capable of building CockroachDB. Source tarballs
-# are expected to build and install a functional cockroach binary into the PATH,
-# even when the tarball is extracted outside of GOPATH.
+# are expected to build, build after cleaning, and install a functional
```
lego: Is the purpose of the clean-then-build to ensure we don't clean up the new precompiled dependencies  (`embedded.go`, etc.)?

Is there any reason for doing the first build?

lgo (approved):
LGTM, but it looks like verify-archive is still failing.

https://teamcity.cockroachdb.com/viewLog.html?buildId=388452&buildTypeId=Cockroach_UnitTests_VerifyArchive&tab=buildLog&_focus=1442
```
./cockroach/build/node-run.sh: line 42: yarn: command not found
```

build/protobuf.mk
benesch: The source tarballs strip the Git checkout to reduce the size of the tarball. (The `.git` directory on my machine is 1.6GB, which is _crazy_.) So `git ls-files` doesn't work, but `find` does. This `git ls-files`wasn't an issue before because `make build` didn't call `make protobuf`, but the removal of embedded.go means every `make build` invocation checks to make sure the protobufs are up-to-date.

build/verify-archive.sh
benesch: Heh, good point. Changed to clean then build.

PR #19508, opened 2017-10-25
tscache: introduce skiplist-backed TimestampCache

nvb:
This is still a major work in progress, but I wanted to publish it in-case others wanted
to play around with it as well. I'm going to continue experimenting with this over the
next few days.

The change creates a new implementation of `tscache.Cache` that's backed by a concurrent
skiplist. It uses a fork of https://github.com/andy-kimball/tscache, which contains a similar
timestamp cache built on top of an arena-based skiplist (https://github.com/andy-kimball/arenaskl).
A couple of important modifications were performed to make the structure usable in CockroachDB, which are located over in https://github.com/nvanbenschoten/tscache. The actual changes here
are minimal, and amount to creating a wrapper type around the library and ripping out now-unnecessary locking.

Nothing here is polished, but that wasn't the intention for now. The important part is that all tests pass with the new `tscache.Cache` implementation plugged in!

Preliminary benchmarking shows impressive results. Here, the effect of the arena allocator is obvious. The properties of the skiplist also allow us to avoid a `request` interval map like we have in the current code, which I predict is also contributing to speedup. The following lists some (hastily gathered) performance diffs:

#### kv, 100% reads
`kv --read-percent 100 --concurrency 16 --splits 100 --write-seq 100000`
- GCTrace time dropped from **7%** to **0%**
- Throughput increased by **14%**
- 99th percentile latency dropped by **40-60%**

#### kv, 80% reads
`kv --read-percent 80 --concurrency 16 --splits 100 --write-seq 100000`
- Throughput increased by **11%**
- 99th percentile latency dropped by **20%**

#### tpcc
`default settings`
- Throughput increased by **30%**
- 99th percentile latency dropped by **50-60%**

cc. @petermattis @spencerkimball @andy-kimball

### DIFF
--- a/pkg/storage/client_replica_test.go
+++ b/pkg/storage/client_replica_test.go
@@ -640,13 +640,11 @@ func TestRangeTransferLease(t *testing.T) {
 
 		// We'd like to verify the timestamp cache's low water mark, but this is
 		// impossible to determine precisely in all cases because it may have
-		// been subsumed by future tscache accesses. However, we know that there
-		// were no tscache accesses since we executed the transfer lease
-		// request, which means that the low water mark will still be equal to
-		// the high water mark. So instead of checking the low water mark, we
-		// make sure that the high water mark is set to the new lease start
-		// time, which is less than the previous lease's expiration time.
-		if highWater := replica1.GetTSCacheHighWater(); highWater != replica1Lease.Start {
+		// been subsumed by future tscache accesses. So instead of checking the
+		// low water mark, we make sure that the high water mark is equal to or
+		// greater than the new lease start time, which is less than the
+		// previous lease's expiration time.
+		if highWater := replica1.GetTSCacheHighWater(); highWater.Less(replica1Lease.Start) {
 			t.Fatalf("expected timestamp cache high water %s, but found %s",
 				replica1Lease.Start, highWater)
 		}
--- a/pkg/storage/helpers_test.go
+++ b/pkg/storage/helpers_test.go
@@ -24,6 +24,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/cockroach/pkg/config"
@@ -37,7 +38,6 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
-	"github.com/pkg/errors"
 )
 
 // AddReplica adds the replica to the store's replica map and to the sorted
@@ -312,8 +312,10 @@ func (r *Replica) CommandSizesLen() int {
 // GetTSCacheHighWater returns the high water mark of the replica's timestamp
 // cache.
 func (r *Replica) GetTSCacheHighWater() hlc.Timestamp {
-	r.store.tsCacheMu.Lock()
-	defer r.store.tsCacheMu.Unlock()
+	if !r.store.tsCacheMu.cache.ThreadSafe() {
+		r.store.tsCacheMu.Lock()
+		defer r.store.tsCacheMu.Unlock()
+	}
 
 	start := roachpb.Key(r.Desc().StartKey)
 	end := roachpb.Key(r.Desc().EndKey)
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -1956,9 +1956,13 @@ func (ec *endCmds) done(br *roachpb.BatchResponse, pErr *roachpb.Error, retry pr
 			// Clockless mode: all reads count as writes.
 			creq.Writes, creq.Reads = append(creq.Writes, creq.Reads...), nil
 		}
-		ec.repl.store.tsCacheMu.Lock()
+		if !ec.repl.store.tsCacheMu.cache.ThreadSafe() {
+			ec.repl.store.tsCacheMu.Lock()
+		}
 		ec.repl.store.tsCacheMu.cache.AddRequest(creq)
-		ec.repl.store.tsCacheMu.Unlock()
+		if !ec.repl.store.tsCacheMu.cache.ThreadSafe() {
+			ec.repl.store.tsCacheMu.Unlock()
+		}
 	}
 
 	if fn := ec.repl.store.cfg.TestingKnobs.OnCommandQueueAction; fn != nil {
@@ -2338,8 +2342,10 @@ func (r *Replica) applyTimestampCache(
 
 	// TODO(peter): We only need to hold a write lock during the ExpandRequests
 	// calls. Investigate whether using a RWMutex here reduces lock contention.
-	r.store.tsCacheMu.Lock()
-	defer r.store.tsCacheMu.Unlock()
+	if !r.store.tsCacheMu.cache.ThreadSafe() {
+		r.store.tsCacheMu.Lock()
+		defer r.store.tsCacheMu.Unlock()
+	}
 
 	if ba.Txn != nil {
 		r.store.tsCacheMu.cache.ExpandRequests(span, ba.Txn.Timestamp)
--- a/pkg/storage/replica_proposal.go
+++ b/pkg/storage/replica_proposal.go
@@ -209,11 +209,15 @@ func (r *Replica) leasePostApply(ctx context.Context, newLease roachpb.Lease) {
 		// lease's expiration but instead use the new lease's start to initialize
 		// the timestamp cache low water.
 		desc := r.Desc()
-		r.store.tsCacheMu.Lock()
+		if !r.store.tsCacheMu.cache.ThreadSafe() {
+			r.store.tsCacheMu.Lock()
+		}
 		for _, keyRange := range makeReplicatedKeyRanges(desc) {
 			r.store.tsCacheMu.cache.SetLowWater(keyRange.start.Key, keyRange.end.Key, newLease.Start)
 		}
-		r.store.tsCacheMu.Unlock()
+		if !r.store.tsCacheMu.cache.ThreadSafe() {
+			r.store.tsCacheMu.Unlock()
+		}
 
 		// Reset the request counts used to make lease placement decisions whenever
 		// starting a new lease.
--- a/pkg/storage/replica_test.go
+++ b/pkg/storage/replica_test.go
@@ -1963,8 +1963,10 @@ func TestReplicaUpdateTSCache(t *testing.T) {
 		t.Error(pErr)
 	}
 	// Verify the timestamp cache has rTS=1s and wTS=0s for "a".
-	tc.repl.store.tsCacheMu.Lock()
-	defer tc.repl.store.tsCacheMu.Unlock()
+	if !tc.repl.store.tsCacheMu.cache.ThreadSafe() {
+		tc.repl.store.tsCacheMu.Lock()
+		defer tc.repl.store.tsCacheMu.Unlock()
+	}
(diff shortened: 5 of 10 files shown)


tschottdorf:
Very exciting work, @nvanbenschoten! Does the `defaultCacheSize` parameter (64MB, which btw sounds tiny to me) allow for approximately the same number of entries in the cache or are we evicting much earlier now? I assume you ran all benchmarks long enough so that they would actually rotate the caches (since the cache is supposedly getting slower as it fills up due to how the skiplist is implemented?) Would be interesting to run the benchmarks with a large cache size such as 512MB (or even few gigs, just to see how far we can push it). It would also be interesting to see how the low watermark rises with this cache compared to the other one.

PS I think you may have accidentally amended one of Andy's commits (`Enhance benchmarks`) in `tscache`: 


---

Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 6 of 6 files at r3.

---

*[pkg/storage/helpers_test.go, line 335 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-KxIfZY24pMGIpicdVpC:-KxIfZY3-W3a7NZ7jG6T:b-svatt1) ([raw file](https://github.com/cockroachdb/cockroach/blob/cc4d267936c32bec6d19450119c131c031e499e3/pkg/storage/helpers_test.go#L335)):*
> ```Go
> // GetTimestampCacheLowWater returns the timestamp cache low water mark.
> func (r *Replica) GetTimestampCacheLowWater() hlc.Timestamp {
> 	// r.store.tsCacheMu.Lock()
> ```

Such a satisfying diff.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-:-KxIg8x-N7S7pqQk1TgG:bhihi34)*

petermattis:
Very cool. Did you come up with a solution for keeping entries in the cache for `MinTSCacheWindow` seconds? I was thinking about this morning. (Haven't looked at the code yet).

nvanbenschoten:
> Does the `defaultCacheSize` parameter (64MB, which btw sounds tiny to me) allow for approximately the same number of entries in the cache or are we evicting much earlier now?

I haven't verified this, but they should allow for roughly the same number of entries because the currently used [`cacheEntrySize`](https://github.com/cockroachdb/cockroach/blob/4e5d7ff865b22ffca832eb3fefc703e3a5a9b06d/pkg/storage/tscache/cache_impl.go#L70) accounts for roughly the same amount as what the skiplist's arena will account for (keys, hlc.Timestamp, txnID, entry overhead). 

I also agree that `64MB` sounds tiny. I think we'll want to size this as some fraction of available memory because that will also scale with the expected amount of load. Anyone benchmarking this may want to play around with increasing this by an order of magnitude or so.

> Did you come up with a solution for keeping entries in the cache for MinTSCacheWindow seconds?

@spencerkimball and I were discussing this last night. I was thinking that there could be some form of a dynamic `fixedCache` count instead of us maintaining only two. We could then add some policy to `rotateCaches` to only release older `fixedCaches` when their `maxWallTime` is more than `now-MinTSCacheWindow`. We'd probably also want to have some minimum count that the number of `fixedCaches` can shrink down to (maybe just keep it at 2?).

petermattis:
> @spencerkimball and I were discussing this last night. I was thinking that there could be some form of a dynamic fixedCache count instead of us maintaining only two. We could then add some policy to rotateCaches to only release older fixedCaches when their maxWallTime is more than now-MinTSCacheWindow. We'd probably also want to have some minimum count that the number of fixedCaches can shrink down to (maybe just keep it at 2?).

Heh, that's exactly what I was thinking. We'd have to maintain the max timestamp per fixedCache, but that seems relatively straightforward.

tschottdorf:
> @spencerkimball and I were discussing this last night. I was thinking that there could be some form of a dynamic fixedCache count instead of us maintaining only two. We could then add some policy to rotateCaches to only release older fixedCaches when their maxWallTime is more than now-MinTSCacheWindow. We'd probably also want to have some minimum count that the number of fixedCaches can shrink down to (maybe just keep it at 2?).

How would the lookup work in that scenario? You'd check the first cache, and if you fall through to the low watermark but you'd rather have something smaller, you check the next level, until you bottom out? That doesn't seem unreasonable, given that requests near the write front wouldn't usually have to check anything but the first level.

nvanbenschoten:
> We'd have to maintain the max timestamp per fixedCache

> You'd check the first cache, and if you fall through to the low watermark but you'd rather have something smaller, you check the next level, until you bottom out? That doesn't seem unreasonable, given that requests near the write front wouldn't usually have to check anything but the first level.

Actually, that's exactly what https://github.com/andy-kimball/tscache already does!

petermattis:
Finally got around to looking at the implementation. Very neat!

It is a bit wasteful that every read/write span for request gets a copy of the timestamp and the transaction ID, especially given that the data is immutable once it is stored in the arena. For a given request, it would be a significant space savings if we could store the timestamp/txnID once in the arena and then have each node store the offset. I think this could be done, though there are definite challenges. Certainly not worthwhile exploring in the near term.

nvanbenschoten:
> It is a bit wasteful that every read/write span for request gets a copy of the timestamp and the transaction ID, especially given that the data is immutable once it is stored in the arena. For a given request, it would be a significant space savings if we could store the timestamp/txnID once in the arena and then have each node store the offset.

@petermattis that's a good idea which could certainly be done, and it may even be something we'd want to look into sooner rather than later. In the case where all nodes have a key and a gap value, it would result in a ~22% space saving `(4*4+28)/(2*28)`. It becomes larger if the offset is used to point to timestamp and txnID together: ~35% savings `(2*4+28)/(2*28)`, but this comes with a few extra complications. Furthermore, I think these savings would actually start to compound as spans subsumed and ratcheted other's key and gap values to their same timestamps and txnIDs. The problem is, to take advantage of this compounding savings, we'd need to remove unused timestamps and txnIDs from the arena. Because the arena is immutable, this wouldn't be trivial. For now, I think we should leave this as is, but I'll add a TODO if we get far enough that this becomes something we should start to think about.

nvanbenschoten:
I've rebased this on top of #19974, so it's now using the `intervalSkl` from #19672. The two commits local to this PR are still works in progress (and not ready to be carefully reviewed), but they expose a `COCKROACH_USE_SKL_TSCACHE` environment variable to switch between `tscache.Cache` implementations.

@petermattis I did some testing with this using `kv`. I'm not seeing the txn retry errors anymore, so I'm thinking that we were right about them being due to one of the two race conditions addressed in #19672.

petermattis:
@nvanbenschoten Are you waiting on review of this or is there still more work to do here?

---

Review status: 0 of 11 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-:-KyvuWLRZhPQrOGUmd5L:bc974qh)*

nvanbenschoten:
No, I'm not waiting on a review at the moment. Before pushing this forward I want to improve the randomized testing we perform across `Cache` implementations to assure that they always give the same results, especially under concurrent workloads.

nvanbenschoten:
This has been rebased off of https://github.com/cockroachdb/cockroach/pull/20087 and is ready to be reviewed once that gets resolved.

nvanbenschoten:
@bdarnell would it be possible to stress this (using the `COCKROACH_USE_SKL_TSCACHE` env var) under Jepsen for a day or two?

bdarnell:
We're not set up to easily run multiple configurations of the jepsen tests, but you could temporarily add a commit to this PR to default that variable to true and then run a one-off build on this branch.

cuongdo:
@bdarnell aren't there current failures in Jepsen that might be confusing to sort through while testing this PR?

bdarnell:
Yes, but I'm assuming we don't want to wait for all of those to be resolved before testing this change. You'll just have to look at the logs (I can help with this) to see whether it's an inconsistency or some other automation/glue problem (All the current problems are the latter).

nvanbenschoten:
This has been rebased off master and is now ready for review. 

It got most of the way through Jepsen before an SSH failure. I'm running the two tests that didn't get to finish again.

tschottdorf:
:lgtm: but remove `[DNM]` from commits. Exciting!

---

Reviewed 9 of 10 files at r5, 3 of 3 files at r6.

---

*[build/jepsen-common.sh, line 54 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-Kz9_YYHIPzsJguY3Oq0:-Kz9_YYHIPzsJguY3Oq1:b3g84il) ([raw file](https://github.com/cockroachdb/cockroach/blob/cf53406d50afe37c540925d09492a28b0eab51d5/build/jepsen-common.sh#L54)):*
> ```Shell
> 
> tests=(
>     #"bank"
> ```

Just leaving a comment to double-check that this is reverted before merge.

---

*[pkg/storage/tscache/cache_test.go, line 515 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-Kz9aUIpkfrbjgfU1Fvd:-Kz9aUIpkfrbjgfU1Fve:bl6fwgv) ([raw file](https://github.com/cockroachdb/cockroach/blob/cf53406d50afe37c540925d09492a28b0eab51d5/pkg/storage/tscache/cache_test.go#L515)):*
> ```Go
> 
> 	// Run one subtest using a real clock to generate timestampts and
> ```

Timestampts :trollface: 
Do you need to update the commit here? I think you fixed this in the other PR.
Not reviewing the rest of the test, since I've seen it already. Please let me know if anything changed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-:-Kz9bMQ3ymmVV2oRl-ii:b-8sqdmp)*

nvanbenschoten:
TFTR! I think you mistook a new test as being the same as an existing one though.

---

Review status: 8 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/storage/tscache/cache_test.go, line 515 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-Kz9aUIpkfrbjgfU1Fvd:-KzAU8oUjBoOJlz4jy7Z:bxl0tco) ([raw file](https://github.com/cockroachdb/cockroach/blob/cf53406d50afe37c540925d09492a28b0eab51d5/pkg/storage/tscache/cache_test.go#L515)):*

Timestampts :trollface: 
Do you need to update the commit here? I think you fixed this in the other PR.
Not reviewing the rest of the test, since I've seen it already. Please let me know if anything changed.
</blockquote></details>

This comment was copied from `TestIntervalSklConcurrentVsSequential` and `TestIntervalSklConcurrency`, so that's why it had the same spelling error (fixed now!). This is a new test though. It's testing that `treeImpl` and `sklImpl` return the same results when accessed concurrently, which is a little different than the test that makes sure the `intervalSkl` returns the same results when accessed sequentially and when accessed concurrently.

---

*[build/jepsen-common.sh, line 54 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-Kz9_YYHIPzsJguY3Oq0:-KzAVGtrHTxVAXXFxk07:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/cf53406d50afe37c540925d09492a28b0eab51d5/build/jepsen-common.sh#L54)):*

Just leaving a comment to double-check that this is reverted before merge.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-:-KzAV1wvdCoJFKfW25i6:b-9q55js)*

tschottdorf:
:lgtm:

---

Reviewed 2 of 9 files at r4, 4 of 4 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19508#-:-KzVsJRtr6Wsf93vwZTU:bnfp4nl)*

PR #19527, opened 2017-10-25
build: various make improvements

benesch:


build/protobuf.mk
```diff
@@ -62,12 +62,6 @@ CPP_PROTOS := $(filter %/roachpb/metadata.proto %/roachpb/data.proto %/roachpb/i
 CPP_HEADERS := $(subst ./,$(NATIVE_ROOT)/,$(CPP_PROTOS:%.proto=%.pb.h))
 CPP_SOURCES := $(subst ./,$(NATIVE_ROOT)/,$(CPP_PROTOS:%.proto=%.pb.cc))
 
-$(PROTOC_PLUGIN):
-	$(GO_INSTALL) $(REPO_ROOT)/pkg/cmd/protoc-gen-gogoroach
-
-$(GRPC_GATEWAY_PLUGIN):
```
lego: It appears this plugin is still used. Is there another build rule taking care of this now? (e.g. this got replaced by a `pkg/cmd/...` command)

lgo (approved):
LGTM! I was just going to comment on a problem with bin files not cleaning (well, previously) but this is even better!

build/protobuf.mk
benesch: Yeah, it's included in the $(BOOTSTRAP_TARGET) recipe, so including it here was redundant. You should definitely check my eyes on this, though.

lego: 👍 looks good.

benesch:
Thanks, @lego!

PR #19534, opened 2017-10-25
server,sql: Refactor distSQLPlanner to only be initialized once

lgo:
This changes DistSQLPlanner to be exported. This will let us initialize a single DistSQLPlanner in pkg/server.

Note that a few methods on DistSQLPlanner were unexported as they don't escape pkg/sql. These exported methods are used outside of the file they're defined, but not out of the packaged.

This also refactors distSQLPlanner to only be initialized once. This moves distSQLPlanner initialization out from the SchemaChangeManager and Executor to instead use a shared instance initialized in Server.Start.

---
The first commit just renames things. @RaduBerinde thoughts on this? I'm not about our export/non-exported conventions as this does unexport a few functions that are called outside the definition file.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1074,6 +1074,24 @@ If problems persist, please see ` + base.DocsURL("cluster-setup-troubleshooting.
 		testingKnobs = new(sql.SchemaChangerTestingKnobs)
 	}
 
+	var distSQLPlannerKnobs sql.DistSQLPlannerTestingKnobs
+	if sqlExecutorTestingKnobs := s.cfg.TestingKnobs.SQLExecutor; sqlExecutorTestingKnobs != nil {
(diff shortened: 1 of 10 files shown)


tschottdorf:
I got alarmed by the word "singleton" in the message and scrolled through the diff, but it doesn't actually seem that you're making it a singleton (under which I understand a package-global variable). Unless that's actually what you're trying to do (I hope not), consider re-wording this appropriately.

lego:
Welp didn't mean to set off alarms. Renamed things to "initializing a single instance"

RaduBerinde:
:lgtm_strong:    Thanks!

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19534#-:-KxJbBNMYYW9xKSiYUtr:by1cciz)*

vivekmenezes:
Thanks!

PR #19537, opened 2017-10-25
sql: fix anonymous alias set up for natural join

### LINKED ISSUE #19536: sql: column alias mess-up when natural-joining anonymous table
I was looking at this code and it seemed wrong:
https://github.com/cockroachdb/cockroach/blob/8ac54961ab8edc641e41880f8e410032dc1843eb/pkg/sql/join_predicate.go#L312

I believe it should be `left.sourceColumns[leftEqualityIndices[i]].Name`. Indeed, I was able to craft a case which shows the bug:

```sql
root@:26257/test> SHOW CREATE TABLE t1;
+-------+-------------------------------------+
| Table |             CreateTable             |
+-------+-------------------------------------+
| t1    | CREATE TABLE t1 (␤                  |
|       |     a INT NULL,␤                    |
|       |     b INT NULL,␤                    |
|       |     FAMILY "primary" (a, b, rowid)␤ |
|       | )                                   |
+-------+-------------------------------------+
(1 row)

Time: 3.34088ms

root@:26257/test> SHOW CREATE TABLE t2;
+-------+----------------------------------------+
| Table |              CreateTable               |
+-------+----------------------------------------+
| t2    | CREATE TABLE t2 (␤                     |
|       |     c INT NULL,␤                       |
|       |     a INT NULL,␤                       |
|       |     b INT NULL,␤                       |
|       |     FAMILY "primary" (c, a, b, rowid)␤ |
|       | )                                      |
+-------+----------------------------------------+
(1 row)

Time: 2.922496ms

root@:26257/test> SELECT b FROM t1 NATURAL JOIN (SELECT * FROM T2);
pq: column reference "b" is ambiguous
```

This query should work - `b` is the merged column. We are incorrectly putting `c,a` instead of `a,b` in the hidden column list, which lets `b` have an extra anonymous alias.

RaduBerinde:
Fixes #19536.

Without the fix, the new test would fail with an ambiguous column error. I plan to cherry-pick this into 1.1.x

### DIFF
--- a/pkg/sql/join_predicate.go
+++ b/pkg/sql/join_predicate.go
@@ -309,8 +309,8 @@ func makeEqualityPredicate(
 	// anonymous data source.
 	for i := 0; i < numMergedEqualityColumns; i++ {
 		anonymousAlias.columnRange = append(anonymousAlias.columnRange, i)
-		hiddenLeftNames = append(hiddenLeftNames, left.sourceColumns[i].Name)
-		hiddenRightNames = append(hiddenRightNames, right.sourceColumns[i].Name)
+		hiddenLeftNames = append(hiddenLeftNames, left.sourceColumns[leftEqualityIndices[i]].Name)
+		hiddenRightNames = append(hiddenRightNames, right.sourceColumns[rightEqualityIndices[i]].Name)
 	}
 
 	// Now collect the other table-less columns into the anonymous data
--- a/pkg/sql/logictest/testdata/logic_test/join
+++ b/pkg/sql/logictest/testdata/logic_test/join
@@ -1218,6 +1218,32 @@ NULL       3      3
  201    NULL    201
  401    NULL    401
 
+# Test for #19536.
+query I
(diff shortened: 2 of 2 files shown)


jordanlewis:
:lgtm: nice catch - while you're changing that logic test you might as well add a query with results - it's easier to read that than the explain output. I think both are valuable.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19537#-:-KxJmi6yuGEsnw3pCF_H:buu0enz)*

RaduBerinde:
Thanks, done.

---

Review status: 1 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19537#-:-KxK-sYblo49uqq-DRlr:bo3vl6y)*

jordanlewis:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19537)*

PR #19538, opened 2017-10-25
storage: do not GC large transactions

tbg:
Not a real PR - please don't review. It will turn into a real PR at some point.

--

This code needs chunking or it will permanently wedge the GC queue, as observed in the wild
in #19216.

This is not a solution but merely a workaround.

### DIFF
--- a/pkg/storage/gc_queue.go
+++ b/pkg/storage/gc_queue.go
@@ -375,6 +375,8 @@ func processLocalKeyRange(
 
 	var gcKeys []roachpb.GCRequest_GCKey
 
+	const maxIntentHack = 800
+
 	handleOneTransaction := func(kv roachpb.KeyValue) error {
 		var txn roachpb.Transaction
 		if err := kv.Value.GetProto(&txn); err != nil {
@@ -407,6 +409,11 @@ func processLocalKeyRange(
 			// persisted, though they still might exist in the system.
 			infoMu.TransactionSpanGCAborted++
 			if err := func() error {
+				// Simply don't resolve these intents. We don't have to and don't want to get bogged down
+				// trying to do so.
+				if len(txn.Intents) > maxIntentHack {
+					return nil
+				}
 				infoMu.Unlock() // intentional
 				defer infoMu.Lock()
 				return resolveIntentsFn(roachpb.AsIntents(txn.Intents, &txn),
@@ -425,6 +432,11 @@ func processLocalKeyRange(
 			if err := func() error {
 				infoMu.Unlock() // intentional
 				defer infoMu.Lock()
+
+				if len(txn.Intents) > maxIntentHack {
+					// Return an error so that we leave this transaction alone. Don't try to resolve its intents.
+					return errors.Errorf("cannot GC transaction with %d intents", len(txn.Intents))
+				}
 				return resolveIntentsFn(roachpb.AsIntents(txn.Intents, &txn), ResolveOptions{Wait: true, Poison: false})
 			}(); err != nil {
 				// Returning the error here would abort the whole GC run, and


tschottdorf:
@bdarnell I'm thinking we should merge this as-is as I'm not sure we'll have anything better in 1.1.3 and this is pretty disastrous if it occurs. WDYT?

andreimatei:
If you want my review, please add more explanation to the commit msg. What is currently happening and what are the consequences of this change?

---

---

*[pkg/storage/gc_queue.go, line 378 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyI4qt0viHdvCGtxqC5:-KyI4qt0viHdvCGtxqC6:bace9yi) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L378)):*
> ```Go
> 	var gcKeys []roachpb.GCRequest_GCKey
> 
> 	const maxIntentHack = 800
> ```

lol. comment pls.
Also should it be named `maxIntentSpansHack`?

---

*[pkg/storage/gc_queue.go, line 438 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyI5ETeKY7_Ms5Fo2KL:-KyI5ETeKY7_Ms5Fo2KM:b-ezidbs) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L438)):*
> ```Go
> 				if len(txn.Intents) > maxIntentHack {
> 					// Return an error so that we leave this transaction alone. Don't try to resolve its intents.
> 					return errors.Errorf("cannot GC transaction with %d intents", len(txn.Intents))
> ```

nit: this msg is not clear. Did we fail for some other reason and are just printing #intents for debugging? I suggest `txn has too many intent spans to GC: %d`.

---

*[pkg/storage/gc_queue.go, line 438 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyI5YCFDcVejlSJLQYr:-KyI5YCFDcVejlSJLQYs:b-swlypu) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L438)):*
> ```Go
> 				if len(txn.Intents) > maxIntentHack {
> 					// Return an error so that we leave this transaction alone. Don't try to resolve its intents.
> 					return errors.Errorf("cannot GC transaction with %d intents", len(txn.Intents))
> ```

s/intents/intent spans

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-:-KyI5cg_cxJuo1yFY3wg:bnwzxxq)*

bdarnell:

---

*[pkg/storage/gc_queue.go, line 378 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyJNfouuwkka52JA_vU:-KyJNekk4GqPwi_4HfKy:br57bcn) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L378)):*
> ```Go
> 	var gcKeys []roachpb.GCRequest_GCKey
> 
> 	const maxIntentHack = 800
> ```

Is the limit really that low? How did you arrive at this number? If we can't GC a transaction of this size, we should probably disallow them (i.e. apply the same limit for `kv.maxIntents`). 

For a quick fix, I'd be more comfortable increasing the queue's process timeout to let these txns finish GC instead of leaving them to never finish. 

---

*[pkg/storage/gc_queue.go, line 414 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyJP5n9e5UnNyw30QU6:-KyJP5n9e5UnNyw30QU7:bji4r2k) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L414)):*
> ```Go
> 				// Simply don't resolve these intents. We don't have to and don't want to get bogged down
> 				// trying to do so.
> 				if len(txn.Intents) > maxIntentHack {
> ```

In the GC queue, aren't we more concerned about the cumulative intents being resolved by this replica than the size of any one transaction? I think we could break out of the `handleOne` loop early instead. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538)*

tschottdorf:

---

*[pkg/storage/gc_queue.go, line 378 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyJNfouuwkka52JA_vU:-KyMAGk2zxnNkEHomBDG:bkq4wkm) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L378)):*

Is the limit really that low? How did you arrive at this number? If we can't GC a transaction of this size, we should probably disallow them (i.e. apply the same limit for `kv.maxIntents`). 

For a quick fix, I'd be more comfortable increasing the queue's process timeout to let these txns finish GC instead of leaving them to never finish. 
</blockquote></details>

The limit is arbitrary. The case I saw in the wild had 80k intents over 4 transactions. Increasing the queue's timeout may not help for these cases. And even 1000 intents can be problematic: a transaction that is really a DeleteRange over 1000 ranges has 1000 intents. How much time do you need to send 1000 `DeleteRange`s that each scan all the range? I wouldn't be comfortable saying but certainly a lot. I'm not sure how to be more concise here, but I know that there is at least one user out there that runs this PR and will be in trouble once they don't run this patch any more.

In the long run we may want to split out the txn record processing into its own queue.

---

*[pkg/storage/gc_queue.go, line 414 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyJP5n9e5UnNyw30QU6:-KyMAcY03ucMh8SC_x6C:b-1cbk5q) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L414)):*

In the GC queue, aren't we more concerned about the cumulative intents being resolved by this replica than the size of any one transaction? I think we could break out of the `handleOne` loop early instead. 
</blockquote></details>

No, a transaction record is only removed after all intents are gone (at least for committed). So if you don't get all of them, you start at zero again the next time (though some intents may have actually been removed). The eventual plan is to be able to update the transaction entry as we go, and what you're saying is true.

BTW, since you're commenting on this code path -- I think it's generally a good idea not to touch intents belonging to a transaction that has been aborted. We don't have to, and the gc queue already picks up stray intents and resolves them. At worst, this may lead to the transaction entry being recreated a few times (due to the pushes) but that's ok, better than "always" duplicating the work. By the time we forcibly GC the txn record, the intents are usually gone.

We don't have that luxury for the committed case, where we have to prove that there aren't any of its intents out there any more, and so we're going to have to go through the intents and periodically update the txn record, removing those we've already dealt with. That's the real fix here, but since people run into these problems I was hoping to short-circuit it in the meantime. It may or may not be a good idea. (other options: hiding this behind an env var).

PS: I'm also wondering, generally, whether it's better to just try to *read* the data as opposed to resolving it blindly. We could send reads (perhaps even the kind that discards the result to save on bandwidth and memory) and let the intent resolution machinery at the serving replica deal with the intents that are *actually* there. That's much less heavyweight under the assumption that there aren't in fact any remaining intents. Worth exploring, I think.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538)*

bdarnell:

---

*[pkg/storage/gc_queue.go, line 378 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyJNfouuwkka52JA_vU:-KyMTEFGKWqpbqxxP4jW:b-kd3a2g) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L378)):*

The limit is arbitrary. The case I saw in the wild had 80k intents over 4 transactions. Increasing the queue's timeout may not help for these cases. And even 1000 intents can be problematic: a transaction that is really a DeleteRange over 1000 ranges has 1000 intents. How much time do you need to send 1000 `DeleteRange`s that each scan all the range? I wouldn't be comfortable saying but certainly a lot. I'm not sure how to be more concise here, but I know that there is at least one user out there that runs this PR and will be in trouble once they don't run this patch any more.

In the long run we may want to split out the txn record processing into its own queue.
</blockquote></details>

If 80k is a problem, we need to reduce the default for `kv.transaction.max_intents` (which allows 100k in a single transaction). 

---

*[pkg/storage/gc_queue.go, line 414 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-KyJP5n9e5UnNyw30QU6:-KyMTMoSbaCdOySwu8EW:b-p4g9eg) ([raw file](https://github.com/cockroachdb/cockroach/blob/92976f363369789325807c3b9e87a9594be2633b/pkg/storage/gc_queue.go#L414)):*

No, a transaction record is only removed after all intents are gone (at least for committed). So if you don't get all of them, you start at zero again the next time (though some intents may have actually been removed). The eventual plan is to be able to update the transaction entry as we go, and what you're saying is true.

BTW, since you're commenting on this code path -- I think it's generally a good idea not to touch intents belonging to a transaction that has been aborted. We don't have to, and the gc queue already picks up stray intents and resolves them. At worst, this may lead to the transaction entry being recreated a few times (due to the pushes) but that's ok, better than "always" duplicating the work. By the time we forcibly GC the txn record, the intents are usually gone.

We don't have that luxury for the committed case, where we have to prove that there aren't any of its intents out there any more, and so we're going to have to go through the intents and periodically update the txn record, removing those we've already dealt with. That's the real fix here, but since people run into these problems I was hoping to short-circuit it in the meantime. It may or may not be a good idea. (other options: hiding this behind an env var).

PS: I'm also wondering, generally, whether it's better to just try to *read* the data as opposed to resolving it blindly. We could send reads (perhaps even the kind that discards the result to save on bandwidth and memory) and let the intent resolution machinery at the serving replica deal with the intents that are *actually* there. That's much less heavyweight under the assumption that there aren't in fact any remaining intents. Worth exploring, I think.
</blockquote></details>

The reason I suggested breaking out of the `handleOne` loop is that we currently don't GC any txn records until we've resolved all the intents from all txns. If we stopped processing new txns once we had resolved 10k intents (for example), maybe we could clear these four txns one at a time. 

I agree that we can ignore aborted transactions more easily than committed ones. There's still some value in the cleanup here but we can use a low threshold here for deciding whether to resolve aborted intents on this path.

I think that changing to do reads of the affected rows instead of resolving the intents directly is going to be the long-term fix for these huge transactions, since it allows us to make gradual progress instead of processing 20k raft writes in one iteration of the GC queue. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538)*

bdarnell:
:lgtm: as a quick fix for 1.1.3 since we've seen multiple occurrences of this problem in the wild. 

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19538#-:-Kz0busv1WhzyxbZYKxR:b1odlij)*

PR #19539, opened 2017-10-25
sql: refactor Executor.execStmtInParallel

andreimatei:
... to not take in a StatementResult.

Before this patch, Executor.execStmtInParallel() was taking in a
StatementResult and writing mocked results to it. I've lifted the
mocking to the parent, since this matches an important concept -
parallel statements don't return results, so it was unnatural for
execStmtInParallel() to look like it's returning them.

### DIFF
--- a/pkg/sql/executor.go
+++ b/pkg/sql/executor.go
@@ -1812,7 +1812,20 @@ func (e *Executor) execStmtInOpenTxn(
 		// statements outside of a transaction are run synchronously with mocked
 		// results, which has the same effect as running asynchronously but
 		// immediately blocking.
-		err = e.execStmtInParallel(stmt, p, res)
+		err = func() error {
+			cols, err := e.execStmtInParallel(stmt, p)
+			if err != nil {
+				return err
+			}
+			// Produce mocked out results for the query - the "zero value" of the
+			// statement's result type:
+			// - tree.Rows -> an empty set of rows
+			// - tree.RowsAffected -> zero rows affected
+			if err := initStatementResult(res, stmt, cols); err != nil {
+				return err
+			}
+			return res.CloseResult()
+		}()
 	} else {
 		p.autoCommit = txnState.implicitTxn && !e.cfg.TestingKnobs.DisableAutoCommit
 		err = e.execStmt(stmt, p, automaticRetryCount, res)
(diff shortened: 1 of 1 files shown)


nvanbenschoten:

---

*[pkg/sql/executor.go, line 2200 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19539#-KxKYlFPVPdAHuyLkIbw:-KxKYlFPVPdAHuyLkIbx:b-eievnz) ([raw file](https://github.com/cockroachdb/cockroach/blob/a94ae5cac88cb02e6946580ca012f0b48b064551/pkg/sql/executor.go#L2200)):*
> ```Go
> // the query's plan which can be used by the caller to return mocked results
> // with the right schema to the client. Note that planning needs to be done
> // synchronously because it's needed by the query dependency analysis.
> ```

Maybe make a mention to the `planNode` we're returning.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19539)*

nvanbenschoten:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19539#-:-KxKYvS-rStQwE30ceU5:bnfp4nl)*

andreimatei:

---

*[pkg/sql/executor.go, line 2200 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19539#-KxKYlFPVPdAHuyLkIbw:-L0BgLegIhcA-yQfJnH1:b-7vl112) ([raw file](https://github.com/cockroachdb/cockroach/blob/a94ae5cac88cb02e6946580ca012f0b48b064551/pkg/sql/executor.go#L2200)):*

Maybe make a mention to the `planNode` we're returning.
</blockquote></details>

replaced the returned plan node with the list of cols, as discussed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19539)*

nvanbenschoten:
:lgtm:er

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19539#-:-L0Bh7dEMGBcYH4VRiDj:b593xvi)*

Issue #18657, opened 2017-09-21
perf: analyze INSERT INTO .... VALUES(...) ON CONFLICT DO UPDATE SET n = n+1


arctica:
Running v1.1-beta.20170907 on debian stretch with ext4 filesystem.
Nodes use --cache=20% --max-sql-memory=10GB

3 nodes in the same datacenter, all with the following hardware:
CPU: Intel(R) Xeon(R) CPU E3-1275 v5 @ 3.60GHz (8 threads)
RAM: 64GiB
Disks: 2x 512GiB NVME in software RAID1

Test benchmark table:
`CREATE TABLE bench (s string primary key, n int);`

A simple [load generating script](https://gist.github.com/anonymous/c2d43573f40c60f7f5be0d594bdf4b18) written in Go is creating 100 concurrent connections and runs the following query on one of the nodes where $1 is a random string of length 3:

`INSERT INTO bench (s, n) VALUES($1, 1) ON CONFLICT (s) DO UPDATE SET n = bench.n + 1`

Base performance without any further modifications is around **300 qps** which is extremely slow. This is likely due to the small datasize resulting in just **1 range** which seems like a unit of concurreny in the system. The default max range size is 64 MiB.

Then I dropped and recreated the table and modified the size of ranges to a min/max of 64/128 KiB to force it to split ranges very quickly. Inserts start slowly again at under 300 qps but at something like 8 ranges already, qps rise to around 700.

Once the system reaches **64 ranges** after a few minutes, performance stabalizes around **2500 qps**.

This is still quite low as some other databases can do nearly an order of magnitude more qps on the same hardware.

At the stable 2.5k qps, each node is using less than 50% of available CPU power, have plenty of RAM free and network throuput is about 1MiB up and down each on a gbit network.

The **disk io** though is quite worrying at around **26MiB writes/s** and 8% CPU spent in iowait as indicated by dstat.
The data being updated is very small (one integer). Granted, CockroachDB keeps all past values so let's assume each update is like an insert. The string has 3 bytes plus 4 byte integer plus overhead for metadata and encoding. Let's assume a generous 64 bytes per entry. At 2500 qps, that would be around 256KiB/s. LSM storage engines have write amplification. Not sure how many levels were generated in this test but I'd assume not too many. So let's assume each row is actually written 4 times as time goes by. That's 1MiB/s. Still off by a factor of 26. Not sure where all this disk io comes from but it seems excessive.

[...]

PR #19540, opened 2017-10-25
storage: Drop all but the last MsgAppResp per raft.Ready

### LINKED ISSUE #19423: storage: Dropped raft messages
Certain workloads result in a steady stream of "dropped" raft messages (measured [at the server](https://github.com/cockroachdb/cockroach/blob/c82e2c816f45a8984b6f7f0ee68f4cf3a368a4ff/pkg/storage/store.go#L2931) after receiving the messages from the network). #18657 is one such workload; I also have a binary load generator from a customer that exhibits this symptom. 

It is unclear whether these dropped messages are having any significant impact - increasing buffer sizes causes the dropped messages to go away without changing the performance of the workload in #18657. Still, there shouldn't be any dropped messages in normal usage so we should figure out what's going on here.

bdarnell:
Each new MsgAppResp makes all prior ones unnecessary, so when we get
more than one in the same batch we can discard them. Some workloads
have been seen to generate a significant number of these redundant
messages, leading to queue overflows, dropped messages, and reduced
performance. In one such test, this improved write throughput and
p99 latency by 30%.

Release note: Eliminated some redundant raft messages, improving write
performance for some workloads by up to 30%.

Fixes #19423

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -3496,9 +3496,11 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 	}
 	r.mu.Unlock()
 
+	var lastAppResp raftpb.Message
 	for _, message := range rd.Messages {
 		drop := false
-		if message.Type == raftpb.MsgApp {
+		switch message.Type {
+		case raftpb.MsgApp:
 			// Iterate over the entries to inline sideloaded commands.
 			for j := range message.Entries {
 				cow := false
@@ -3530,10 +3532,36 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 					message.Entries[j] = *newEnt
 				}
 			}
+
+		case raftpb.MsgAppResp:
+			// A successful (non-reject) MsgAppResp contains one piece of
+			// information: the highest log index. Raft currently queues up
+			// one MsgAppResp per incoming MsgApp, and we may process
+			// multiple messages in one handleRaftReady call (because
+			// multiple messages may arrive while we're blocked syncing to
+			// disk). If we get redundant MsgAppResps, drop all but the
+			// last (we've seen that too many MsgAppResps can overflow
+			// message queues on the receiving side).
+			//
+			// Note that this reorders the chosen MsgAppResp relative to
+			// other messages (including any MsgAppResps with the Reject flag),
+			// but raft is fine with this reordering.
+			//
+			// TODO(bdarnell): Consider pushing this optimization into etcd/raft.
+			// Similar optimizations may be possible for other message types,
+			// although MsgAppResp is the only one that has been seen as a
+			// problem in practice.
+			if !message.Reject && message.Index > lastAppResp.Index {
+				lastAppResp = message
+				drop = true
+			}
 		}
 		if !drop {
 			r.sendRaftMessage(ctx, message)
 		}
+	}
+	if lastAppResp.Index > 0 {
+		r.sendRaftMessage(ctx, lastAppResp)
 	}
 
 	for _, e := range rd.CommittedEntries {
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -3496,73 +3496,7 @@ func (r *Replica) handleRaftReadyRaftMuLocked(
 	}
 	r.mu.Unlock()
 
-	var lastAppResp raftpb.Message
-	for _, message := range rd.Messages {
-		drop := false
-		switch message.Type {
-		case raftpb.MsgApp:
-			// Iterate over the entries to inline sideloaded commands.
-			for j := range message.Entries {
-				cow := false
-				newEnt, err := maybeInlineSideloadedRaftCommand(
-					ctx,
-					r.RangeID,
-					message.Entries[j],
-					r.raftMu.sideloaded,
-					r.store.raftEntryCache,
-				)
-				if err != nil {
-					// We can simply drop the message since it could always get lost
-					// in transit anyway.
-					log.Errorf(ctx, "while inlining sideloaded commands: %s", err)
-					drop = true
-					continue
-				}
-				if newEnt != nil {
-					if !cow {
-						cow = true
-						// Copy the whole slice to avoid data races. Entries are
-						// usually shared between multiple outgoing messages,
-						// and while it would be possible to only modify them
-						// only the first time around, that isn't easy to
-						// implement (since you have to do nontrivial work to
-						// decide whether inlining needs to happen).
-						message.Entries = append([]raftpb.Entry(nil), message.Entries...)
-					}
-					message.Entries[j] = *newEnt
-				}
-			}
-
-		case raftpb.MsgAppResp:
-			// A successful (non-reject) MsgAppResp contains one piece of
-			// information: the highest log index. Raft currently queues up
-			// one MsgAppResp per incoming MsgApp, and we may process
-			// multiple messages in one handleRaftReady call (because
-			// multiple messages may arrive while we're blocked syncing to
-			// disk). If we get redundant MsgAppResps, drop all but the
-			// last (we've seen that too many MsgAppResps can overflow
-			// message queues on the receiving side).
-			//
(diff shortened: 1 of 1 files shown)


a-robinson:
:lgtm:, nice! On the #18657 workload the improvement isn't as pronounced -- it's giving a 5-10% throughput increase if I'm generous interpreting the graph -- but it does completely get rid of the dropped raft messages.

---

---

*[pkg/storage/replica.go, line 3550 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKG_rVYPbvPOgolZvI:-KxKG_rVYPbvPOgolZvJ:b-7g1iiw) ([raw file](https://github.com/cockroachdb/cockroach/blob/93152cad47829e5dbcce1771e212797431fc8468/pkg/storage/replica.go#L3550)):*
> ```Go
> 			// although MsgAppResp is the only one that has been seen as a
> 			// problem in practice.
> 			if !message.Reject && message.Index > lastAppResp.Index {
> ```

I'm pretty sure raft doesn't depend on the transport being ordered, but just to double check: the fact that this could potentially reorder messages if one of them has the `Reject` flag set is ok, right? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-:-KxKGlBX839CPJDxFL5T:b-eqdsjm)*

tschottdorf:
:lgtm: :thumbsup: 

---

Reviewed 1 of 1 files at r1.

---

*[pkg/storage/replica.go, line 3550 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKG_rVYPbvPOgolZvI:-KxKHZUu6Ns-jIMzpnim:bhh1jap) ([raw file](https://github.com/cockroachdb/cockroach/blob/93152cad47829e5dbcce1771e212797431fc8468/pkg/storage/replica.go#L3550)):*

I'm pretty sure raft doesn't depend on the transport being ordered, but just to double check: the fact that this could potentially reorder messages if one of them has the `Reject` flag set is ok, right? 
</blockquote></details>

That should be OK, but I think it's worth a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-:-KxKHxhicGRfKoRh1Ujn:b-wpm7nb)*

bdarnell:

---

*[pkg/storage/replica.go, line 3550 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKG_rVYPbvPOgolZvI:-KxKHbcAuKO86u69Y47Q:b11evir) ([raw file](https://github.com/cockroachdb/cockroach/blob/93152cad47829e5dbcce1771e212797431fc8468/pkg/storage/replica.go#L3550)):*

That should be OK, but I think it's worth a comment.
</blockquote></details>

Raft tolerates reordered messages. In some cases reordering can reduce performance, but it will never produce incorrect results. The main issue is that the order of MsgApps should be preserved (with respect to other MsgApps). Reordering MsgAppResps shouldn't be a problem (and rejections are rare in practice). I'll add a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540)*

petermattis:
:lgtm:

---

---

*[pkg/storage/replica.go, line 3500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKKPyy_9RWYBiboQoy:-KxKKPyy_9RWYBiboQoz:b-l890px) ([raw file](https://github.com/cockroachdb/cockroach/blob/56886292f92f1f2f0cd3d945d4bd65d9add8a10d/pkg/storage/replica.go#L3500)):*
> ```Go
> 
> 	var lastAppResp raftpb.Message
> 	for _, message := range rd.Messages {
> ```

Rather than adjusting the logic in this loop, another way to implement this would have been to add a new function which takes and returns a `raft.Ready` and trims off any unnecessary `MsgAppResp` from `raft.Ready.Messages`. That approach would alleviate the continued bloat of this method and allow for easier testing and an obvious path for upstreaming this change into `etcd/raft`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-:-KxKKMnlvto94fewo5To:bnfp4nl)*

bdarnell:

---

*[pkg/storage/replica.go, line 3500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKKPyy_9RWYBiboQoy:-KxKLTKwHrfa81Hv8Qln:b-l4uj8n) ([raw file](https://github.com/cockroachdb/cockroach/blob/56886292f92f1f2f0cd3d945d4bd65d9add8a10d/pkg/storage/replica.go#L3500)):*

Rather than adjusting the logic in this loop, another way to implement this would have been to add a new function which takes and returns a `raft.Ready` and trims off any unnecessary `MsgAppResp` from `raft.Ready.Messages`. That approach would alleviate the continued bloat of this method and allow for easier testing and an obvious path for upstreaming this change into `etcd/raft`. 
</blockquote></details>

That seems less efficient since you'd need to copy everything into a smaller array (or would you just leave gaps?). I'll reduce bloat by factoring this loop out into a separate method. 

My envisioned upstream solution would be to avoid adding MsgAppResp to the outgoing message queue at all (and synthesize it in `newReady()` instead). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540)*

petermattis:

---

*[pkg/storage/replica.go, line 3500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKKPyy_9RWYBiboQoy:-KxKQSr39Qqlzwf4RaF8:b-mftw5a) ([raw file](https://github.com/cockroachdb/cockroach/blob/56886292f92f1f2f0cd3d945d4bd65d9add8a10d/pkg/storage/replica.go#L3500)):*

That seems less efficient since you'd need to copy everything into a smaller array (or would you just leave gaps?). I'll reduce bloat by factoring this loop out into a separate method. 

My envisioned upstream solution would be to avoid adding MsgAppResp to the outgoing message queue at all (and synthesize it in `newReady()` instead). 
</blockquote></details>

Go supports a very concise way to do this trimming in an efficient manner:

```
func trimRaftReady(rd raft.Ready) raft.Ready {
  messages := rd.Messages
  rd.Messages = rd.Messages[:0]
  for _, msg := rd.Messages {
    if !dropMessage(msg) {
      rd.Messages = append(rd.Messages, msg)
    }
  }
  return rd
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540)*

petermattis:

---

*[pkg/storage/replica.go, line 3500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-KxKKPyy_9RWYBiboQoy:-KxKR9ajnV18ydo8kz6i:b-e654pg) ([raw file](https://github.com/cockroachdb/cockroach/blob/56886292f92f1f2f0cd3d945d4bd65d9add8a10d/pkg/storage/replica.go#L3500)):*

Go supports a very concise way to do this trimming in an efficient manner:

```
func trimRaftReady(rd raft.Ready) raft.Ready {
  messages := rd.Messages
  rd.Messages = rd.Messages[:0]
  for _, msg := rd.Messages {
    if !dropMessage(msg) {
      rd.Messages = append(rd.Messages, msg)
    }
  }
  return rd
}
```
</blockquote></details>

Err, that should be `for _, msg := range messages`, but I'm sure you would have caught that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540)*

nvanbenschoten:
:lgtm: nice!

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19540#-:-KxKVEevrLOPXxjCYxPl:b-7yytuz)*

PR #19545, opened 2017-10-25
sql: use FastIntSet for sourceAlias column set

RaduBerinde:
I was working with the join code and I found the columnRange terminology (as
well as the slice backing) confusing - it is actually a set.

Switching to FastIntSet, which is more efficient and makes it more clear what's
going on.

### DIFF
--- a/pkg/sql/data_source.go
+++ b/pkg/sql/data_source.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
 	"github.com/cockroachdb/cockroach/pkg/sql/privilege"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/cockroachdb/cockroach/pkg/util"
 )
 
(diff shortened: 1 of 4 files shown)


knz:
:lgtm:
Probably a nice small improvement on the number of allocs per plan with joins, too.

---

Reviewed 4 of 4 files at r1.

---

*[pkg/sql/data_source.go, line 893 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19545#-KxNVogY-5bWbOCZJnFZ:-KxNVog_HTbLX5mGEi86:b-8jfzkr) ([raw file](https://github.com/cockroachdb/cockroach/blob/c732eba65d37ac8b8aca5f3faff94d3f01890e0b/pkg/sql/data_source.go#L893)):*
> ```Go
> 		if !ok {
> 			// The data source "src" has no column for table tableName.
> 			// Try again with the net one.
> ```

nit: net -> next

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19545#-:-KxNVyLG5_YCo103HuVZ:btvqfzi)*

PR #19547, opened 2017-10-25
ui: exclude decommissioned nodes from staggered version warning

### LINKED ISSUE #15534: ui: "staggered version" banner considers dead nodes
In the current incarnation of blue.crdb.io, node 7 was recreated and was reborn as node 11, meaning that node 7 is permanently dead. However, since the cluster version was updated since then, the admin UI now displays the staggered version banner, presumably because node 7's node status still has the old version (and will never be updated).

It's also pretty annoying that this banner doesn't provide more detailed information (which nodes? which versions?).

tbg:
Fixes #15534.

### DIFF
--- a/pkg/ui/src/redux/alerts.spec.ts
+++ b/pkg/ui/src/redux/alerts.spec.ts
@@ -9,13 +9,14 @@ import { AdminUIState, createAdminUIStore } from "./state";
 import {
   AlertLevel,
   alertDataSync,
+  versionsSelector,
   staggeredVersionWarningSelector, staggeredVersionDismissedSetting,
   newVersionNotificationSelector, newVersionDismissedLocalSetting,
   disconnectedAlertSelector, disconnectedDismissedLocalSetting,
 } from "./alerts";
 import { VERSION_DISMISSED_KEY, setUIDataKey, isInFlight } from "./uiData";
 import {
-  versionReducerObj, nodesReducerObj, clusterReducerObj, healthReducerObj,
+  livenessReducerObj, versionReducerObj, nodesReducerObj, clusterReducerObj, healthReducerObj,
 } from "./apiReducers";
 
 describe("alerts", function() {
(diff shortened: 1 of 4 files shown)


tschottdorf:
I verified this manually, but that's not worth much. Happy to write the test given instructions or a related example.

vilterp:
You'd add a test around here: https://github.com/cockroachdb/cockroach/blob/ad6df63e975888a598e3a34f5ce45127e097bc1d/pkg/ui/src/redux/alerts.spec.ts#L38

```js
it("filters out decommissioned nodes with a different version", () => {
   ...
});
```

Also make sure to import the selector in the import statement up on line 9.

dianasaur323:
haha, i love how you just went ahead and did this. Thanks @tschottdorf! Much appreciated! We've been having some flux on the UI side, so haven't had time to get to bugs.

tschottdorf:
Okay, noob alert. This works and the test passes, but my editor (and `make lint`) complain about this:

```
	Error at src/redux/alerts.spec.ts:74:11: Argument of type '{ livenesses: { node_id: number; decommissioning: true; }[]; }' is not assignable to parameter of type 'LivenessResponse'.
[14:21:01]	Property 'toObject' is missing in type '{ livenesses: { node_id: number; decommissioning: true; }[]; }'.
```

I'm not sure what I'm missing.

tschottdorf:
@couchand kindly helped me out (was missing the proto constructor)

couchand:
:+1: for looking into this.  Thanks!  :lgtm:, but I left a comment on your map/filter if you feel like looking into it.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/ui/src/redux/alerts.ts, line 56 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19547#-KxOAR0Qz9y5kHjmznu7:-KxOAR0Qz9y5kHjmznu8:b14wynu) ([raw file](https://github.com/cockroachdb/cockroach/blob/05cda408abe1d1fa8823ae89b34ca11347d6a901/pkg/ui/src/redux/alerts.ts#L56)):*
> ```XML
>   livenessByNodeIDSelector,
>   (nodeStatuses, livenessStatusByNodeID) =>
>     nodeStatuses && _.filter(_.uniq(_.map(nodeStatuses, (status) =>
> ```

This is pretty cumbersome.  I think a more readable pipeline would filter first and map second.  @mrtracy would probably suggest using `_.chain` for this, something like:

```
_.chain(nodeStatuses)
  .filter((status) => !status.desc || ...)
  .map((status) => status.build_info.tag)
  .uniq()
  .values()
```

FYI @BramGruneir ran into some trouble with uniq on chain, can't remember offhand what it was...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19547#-:-KxOCGZpD50IwbWqjLUF:bydneze)*

tschottdorf:
TFTRs!

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/ui/src/redux/alerts.ts, line 56 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19547#-KxOAR0Qz9y5kHjmznu7:-KxcoIidMIh4FyPMJti5:bekoy3i) ([raw file](https://github.com/cockroachdb/cockroach/blob/05cda408abe1d1fa8823ae89b34ca11347d6a901/pkg/ui/src/redux/alerts.ts#L56)):*

This is pretty cumbersome.  I think a more readable pipeline would filter first and map second.  @mrtracy would probably suggest using `_.chain` for this, something like:

```
_.chain(nodeStatuses)
  .filter((status) => !status.desc || ...)
  .map((status) => status.build_info.tag)
  .uniq()
  .values()
```

FYI @BramGruneir ran into some trouble with uniq on chain, can't remember offhand what it was...
</blockquote></details>

Only one way to find out! I made the change. Matt didn't seem to have an issue suggesting chain and uniq here: https://github.com/cockroachdb/cockroach/pull/18603#issuecomment-330657226

@BramGruneir, if you remember.. share your wisdom please :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19547#-:-Kxconl3r0KGh0Yi_jjN:bz5mair)*

PR #19548, opened 2017-10-25
storage: use RangeList instead of RangeTree in CommandQueue

### LINKED ISSUE #19345: storage: reconsider command queue RangeGroup implementation

nvb:
Closes #19345.
Extracted from #19361.

This needs more experimentation, but I've done some testing on `kv` and
`tpcc`. I wasn't able to see the `RangeGroup` on profiles before or
after on `tpcc`.

On `kv` with `cycle-length=5` the change seemed to improve average latency
by about 4%. Again though, I wasn't able to reproduce the `RangeCache`
blowup on profiles for the before or after testing.

I'd like to see the effect it has on the Premier benchmarks.

cc. @asubiotto

### DIFF
--- a/pkg/storage/command_queue.go
+++ b/pkg/storage/command_queue.go
@@ -240,8 +240,8 @@ func NewCommandQueue(coveringOptimization bool) *CommandQueue {
 		readsBuffer:          make(map[*cmd]struct{}),
 		reads:                interval.NewTree(interval.ExclusiveOverlapper),
 		writes:               interval.NewTree(interval.ExclusiveOverlapper),
-		wRg:                  interval.NewRangeTree(),
-		rwRg:                 interval.NewRangeTree(),
+		wRg:                  interval.NewRangeList(),
+		rwRg:                 interval.NewRangeList(),
 		coveringOptimization: coveringOptimization,
 	}
 	return cq


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19548#-:-KxNvP4EEMESIsA5z73G:bnfp4nl)*

nvanbenschoten:
I just ran some benchmarks on this change and it looks like it's a clear improvement:

```
name                                          old time/op    new time/op    delta
CommandQueueReadWriteMix/readsPerWrite=1-4      8.89ms ± 1%    6.35ms ± 1%  -28.59%  (p=0.000 n=8+9)
CommandQueueReadWriteMix/readsPerWrite=4-4      4.99ms ± 2%    3.95ms ± 1%  -20.88%  (p=0.000 n=9+9)
CommandQueueReadWriteMix/readsPerWrite=16-4     2.85ms ± 2%    2.65ms ± 2%   -6.90%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=64-4     1.57ms ± 3%    1.57ms ± 5%     ~     (p=0.481 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=128-4    1.32ms ± 1%    1.33ms ± 4%     ~     (p=0.165 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=256-4    1.20ms ± 1%    1.24ms ± 4%   +3.66%  (p=0.000 n=9+9)

name                                          old alloc/op   new alloc/op   delta
CommandQueueReadWriteMix/readsPerWrite=1-4      1.30MB ± 0%    0.74MB ± 0%  -43.31%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=4-4       992kB ± 0%     691kB ± 0%  -30.35%  (p=0.000 n=9+9)
CommandQueueReadWriteMix/readsPerWrite=16-4      790kB ± 0%     667kB ± 0%  -15.57%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=64-4      533kB ± 0%     501kB ± 0%   -5.94%  (p=0.000 n=10+9)
CommandQueueReadWriteMix/readsPerWrite=128-4     487kB ± 0%     472kB ± 0%   -3.12%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=256-4     464kB ± 0%     457kB ± 0%   -1.50%  (p=0.000 n=10+10)

name                                          old allocs/op  new allocs/op  delta
CommandQueueReadWriteMix/readsPerWrite=1-4       29.0k ± 0%      9.1k ± 0%  -68.69%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=4-4       16.2k ± 0%      8.2k ± 0%  -49.32%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=16-4      9.08k ± 0%     7.07k ± 0%  -22.12%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=64-4      5.40k ± 0%     4.88k ± 0%   -9.65%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=128-4     4.74k ± 0%     4.48k ± 0%   -5.35%  (p=0.000 n=10+10)
CommandQueueReadWriteMix/readsPerWrite=256-4     4.41k ± 0%     4.29k ± 0%   -2.73%  (p=0.000 n=10+8)
```

tschottdorf:
These benchmarks look delicious, but I'm worried about scenarios with linear complexity that aren't covered by our coverage (I think we probably under-test high-contention scenarios in which the list size becomes significant).  Could you expand on what you expect for those? Have they been tested?

PR #19551, opened 2017-10-26
config: split subzones

benesch:
As described in the partitioning RFC, teach the config package how to
split subzones stored in a ZoneConfig. In the interest of minimizing
review burden, this commit does not introduce any codepaths outside of
tests that create ZoneConfigs with subzones.

Also do some cleanup work while I'm in the area.

### DIFF
--- a/pkg/config/system.go
+++ b/pkg/config/system.go
@@ -295,10 +295,6 @@ func (s SystemConfig) ComputeSplitKey(startKey, endKey roachpb.RKey) roachpb.RKe
 		// In either case, start looking for splits at the first ID usable
 		// by the user data span.
 		startID = keys.MaxSystemConfigDescID + 1
-	} else {
-		// The start key is either already a split key, or after the split
-		// key for its ID. We can skip straight to the next one.
-		startID++
 	}
 
 	// Build key prefixes for sequential table IDs until we reach endKey. Note
@@ -311,16 +307,36 @@ func (s SystemConfig) ComputeSplitKey(startKey, endKey roachpb.RKey) roachpb.RKe
 	findSplitKey := func(startID, endID uint32) roachpb.RKey {
 		// endID could be smaller than startID if we don't have user tables.
 		for id := startID; id <= endID; id++ {
-			key := roachpb.RKey(keys.MakeTablePrefix(id))
-			// Skip if this ID matches the provided startKey.
-			if !startKey.Less(key) {
+			tableKey := roachpb.RKey(keys.MakeTablePrefix(id))
+			// This logic is analogous to the well-commented static split logic above.
+			if startKey.Less(tableKey) {
+				if tableKey.Less(endKey) {
+					return tableKey
+				}
+				return nil
+			}
+
+			zoneVal := s.GetValue(MakeZoneKey(id))
+			if zoneVal == nil {
 				continue
 			}
-			// Handle the case where EndKey is already a table prefix.
-			if !key.Less(endKey) {
-				break
+			var zone ZoneConfig
+			if err := zoneVal.GetProto(&zone); err != nil {
+				// An error while decoding the zone proto is unfortunate, but logging a
+				// message here would be excessively spammy. Just move on, which
+				// effectively assumes there are no subzones for this table.
+				continue
 			}
-			return key
+			// This logic is analogous to the well-commented static split logic above.
+			for _, s := range zone.subzoneSplits() {
+				subzoneKey := append(tableKey, s...)
+				if startKey.Less(subzoneKey) {
+					if subzoneKey.Less(endKey) {
+						return subzoneKey
+					}
+					return nil
+				}
+			}
 		}
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


a-robinson:
:lgtm:, and very nice use of separate commits :clap:

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.

---

*[pkg/config/system.go, line 325 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-KxOAstYHrZVPXdZH-E1:-KxOAstYHrZVPXdZH-E2:b5d1pbs) ([raw file](https://github.com/cockroachdb/cockroach/blob/2ec64b8b6587a92730764e33402f1469d7a8cd87/pkg/config/system.go#L325)):*
> ```Go
> 			var zone ZoneConfig
> 			if err := zoneVal.GetProto(&zone); err != nil {
> 				// An error while decoding the zone proto is unfortunate, but logging a
> ```

Logging every N seconds would be nice, but I understand that this case is incredibly unlikely and will probably pop up somewhere else in the code anyway, so skipping it is ok.

---

*[pkg/config/system_test.go, line 182 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-KxO1TCnudYDTyawPX5o:-KxO1TCnudYDTyawPX5p:b-7k2518) ([raw file](https://github.com/cockroachdb/cockroach/blob/2acd336713a3f7121b6ae84c0d4fcd1f49357dcb/pkg/config/system_test.go#L182)):*
> ```Go
> 	for i := 1; i < len(splits); i++ {
> 		if !splits[i-1].Less(splits[i]) {
> 			t.Errorf("previous SplitKey %q should be less than next SplitPoint %q",
> ```

The message here should be updated to reflect that `SplitKey` and `SplitPoint` are now the same thing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-:-KxOBiTv75Y8Ar6Qh9K7:b-np9c80)*

benesch:
Heh, thanks. Trying to make up for my last zone config PR. TFTR!

---

---

*[pkg/config/system.go, line 325 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-KxOAstYHrZVPXdZH-E1:-KxkeyqOmehe2ha-L3ud:b-e9x2ed) ([raw file](https://github.com/cockroachdb/cockroach/blob/2ec64b8b6587a92730764e33402f1469d7a8cd87/pkg/config/system.go#L325)):*

Logging every N seconds would be nice, but I understand that this case is incredibly unlikely and will probably pop up somewhere else in the code anyway, so skipping it is ok.
</blockquote></details>

Ack. Filed #19671.

---

*[pkg/config/system_test.go, line 182 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-KxO1TCnudYDTyawPX5o:-Kxke2rEiFPOzNPnz-Dt:b-nvfudn) ([raw file](https://github.com/cockroachdb/cockroach/blob/2acd336713a3f7121b6ae84c0d4fcd1f49357dcb/pkg/config/system_test.go#L182)):*

The message here should be updated to reflect that `SplitKey` and `SplitPoint` are now the same thing.
</blockquote></details>

Ooh, good eye. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-:-Kxkfqx7DK8b3BVeNbz4:b-fwxx72)*

danhhz:
:lgtm:

---

Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/config/zone.go, line 310 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-Kxmj8k0ab3yrbFy9Kx1:-Kxmj8k1c-wNjISSNZQ_:b2mv606) ([raw file](https://github.com/cockroachdb/cockroach/blob/2871a9b8d21ac7364e18b887730369e1928d519e/pkg/config/zone.go#L310)):*
> ```Go
> 			endKey = span.Key.PrefixEnd()
> 		}
> 		if i >= len(z.SubzoneSpans)-1 || !z.SubzoneSpans[i+1].Key.Equal(endKey) {
> ```

optional: you could also keep a var outside the for loop which contains the last seen end key and skip adding start if they match. could be more readable than reasoning about indexes and off by ones

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-:-KxmklIcxyrF5v-Sj53t:bnfp4nl)*

benesch:
Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/config/zone.go, line 310 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551#-Kxmj8k0ab3yrbFy9Kx1:-Kxpx7i6oM_okyXz5Wlu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2871a9b8d21ac7364e18b887730369e1928d519e/pkg/config/zone.go#L310)):*

optional: you could also keep a var outside the for loop which contains the last seen end key and skip adding start if they match. could be more readable than reasoning about indexes and off by ones
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19551)*

PR #19554, opened 2017-10-26
settings: avoid multiple *settings.Values at runtime

### LINKED ISSUE #19533: log, settings: no crash reports from 1.1.x
We're not seeing any crash reports from 1.1 binaries, even when manually forced to crash with the `force_panic` helper.

Digging in a bit, it seems the crash reporter believes diagnostic reporting is disabled even when it isn't -- looks like in some of the settings refactors to de-singltonize and plumb them, we ended up with more than one values registry at runtime even in non-test binaries, and the one supplied to crash reporting isn't the one being updated by Server.

dt:
This eliminates all but one non-testing call to MakeClusterSettings -- which is critical to correctness.
Any *settings.Values other than the one maintained by the refresh loop will be stale/lies, so only in
testing do we want call sites creating settings value containers.

Fixes #19533.


dt:
I initially wanted to go add much more plumbing, e.g. to all the `newInMem` calls and into `Stopper`, but realized that that change was getting a bit big and conflict-y for a cherry-pick, so I figured I'd start small.

I also explored adding an assertion when a setting is read that the values pointer passed matches one in a package singleton, to try to prevent this from happening, but it's a little messy -- I've decided to checkpoint here for now, though I may look at that again tomorrow.

tschottdorf:
:lgtm: and we'll file a follow-up issue. Thanks for fixing this!

---

---

*[pkg/server/admin_test.go, line 112 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554#-KxO2Js_6S6x_MQGGZYb:-KxO2Js_6S6x_MQGGZYc:b-a1kh1l) ([raw file](https://github.com/cockroachdb/cockroach/blob/1686bd5864144c545c141ae175baef4c52c2fd39/pkg/server/admin_test.go#L112)):*
> ```Go
> 	defer s.Stopper().Stop(context.TODO())
> 
> 	st := cluster.MakeTestingClusterSettings()
> ```

I think the comment was helpful. Now you wonder why not to remove this.

---

*[pkg/settings/setting.go, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554#-KxO2S0gB22koCn4YCYZ:-KxO2S0gB22koCn4YCY_:b-drm8h7) ([raw file](https://github.com/cockroachdb/cockroach/blob/1686bd5864144c545c141ae175baef4c52c2fd39/pkg/settings/setting.go#L50)):*
> ```Go
> }
> 
> // SetCanonicalValuesContainer sets the Values constainer that will be refreshed
> ```

`constainer`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554#-:-KxO4yZSYDKu0r2K5Xj4:bzi92ow)*

dt:
Review status: 10 of 12 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/server/admin_test.go, line 112 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554#-KxO2Js_6S6x_MQGGZYb:-KxO7IT3tf-oHqoJY66D:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1686bd5864144c545c141ae175baef4c52c2fd39/pkg/server/admin_test.go#L112)):*

I think the comment was helpful. Now you wonder why not to remove this.
</blockquote></details>

Done.

---

*[pkg/settings/setting.go, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554#-KxO2S0gB22koCn4YCYZ:-KxO6Yd-0zOkyKNXk_uh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1686bd5864144c545c141ae175baef4c52c2fd39/pkg/settings/setting.go#L50)):*

`constainer`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554)*

dt:
cc @cockroachdb/release after this lands I'll want to cherrypick for 1.1.2

bdarnell:
:lgtm:

---

Reviewed 9 of 12 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19554#-:-KxOGmHUitFfCGuc2lTA:bnfp4nl)*

PR #19556, opened 2017-10-26
ccl/storageccl: retry s3 region fetch

### LINKED ISSUE #19502: backup/restore: transient network error
4 node cluster, SFO, NYC, NYC2, AMS. 

Running  show backup sometimes throws this error message: 

pq: importing 395 ranges: could not find s3 bucket's region: ServiceUnavailable: Service Unavailable
	status code: 503, request id: 207652C3D187982A, host id: age9rSo8QLjHrzOOkpdTm6vzuWFB6A9MTT+QsGHGGTsASX7L+5dXNk6Lcu7+8Rlm6UgrJauHJWg=

madelynnblue:
Retry an S3 operation that has been causing transient errors. If a
503 is returned, we may need to slow down our request rate. Although
the retry package does wait a bit, it is only some 10s of ms. We
instead wait 5s on these errors to attempt to slightly decrease the
request rate. This 5s is purely a guess, but it is also a low enough
time that it won't matter too much if we are waiting longer than
minimally needed.

For now put this only in the place we have seen the failure. We can
optionally retry the other locations if we see problems there. This
is the same approach we used for Azure and GCE with good results.

Fixes #19502

### DIFF
--- a/pkg/ccl/storageccl/export_storage.go
+++ b/pkg/ccl/storageccl/export_storage.go
@@ -20,6 +20,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"google.golang.org/api/option"
 
@@ -367,6 +368,25 @@ type s3Storage struct {
 	s3     *s3.S3
 }
 
+func s3Retry(ctx context.Context, fn func() error) error {
+	const maxAttempts = 3
+	return retry.WithMaxAttempts(ctx, base.DefaultRetryOptions(), maxAttempts, func() error {
+		err := fn()
+		if s3err, ok := err.(s3.RequestFailure); ok {
+			// A 503 error could mean we need to reduce our request rate. Impose an
+			// arbitrary slowdown in that case.
+			// See http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
+			if s3err.StatusCode() == 503 {
+				select {
+				case <-time.After(time.Second * 5):
+				case <-ctx.Done():
+				}
+			}
+		}
+		return err
+	})
+}
+
 var _ ExportStorage = &s3Storage{}
 
 func makeS3Storage(ctx context.Context, conf *roachpb.ExportStorage_S3) (ExportStorage, error) {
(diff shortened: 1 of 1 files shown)


dt:

---

*[pkg/ccl/storageccl/export_storage.go, line 373 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19556#-KxOzUNYxtlINgZJStzy:-KxOzUNYxtlINgZJStzz:b-u1hety) ([raw file](https://github.com/cockroachdb/cockroach/blob/9394a49a0a0082f715e8a2dff59168ab72d2d125/pkg/ccl/storageccl/export_storage.go#L373)):*
> ```Go
> func s3Retry(ctx context.Context, fn func() error) error {
> 	const maxAttempts = 3
> 	return retry.WithMaxAttempts(ctx, base.DefaultRetryOptions(), maxAttempts, func() error {
> ```

any motivation for rolling our own here v.s just passing e.g. `retry.Options{InitialBackoff: 1 * time.Second}` ? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19556)*

mjibson:

---

*[pkg/ccl/storageccl/export_storage.go, line 373 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19556#-KxOzUNYxtlINgZJStzy:-KxP-zwna7F3P1SoRHq6:b-d7as16) ([raw file](https://github.com/cockroachdb/cockroach/blob/9394a49a0a0082f715e8a2dff59168ab72d2d125/pkg/ccl/storageccl/export_storage.go#L373)):*

any motivation for rolling our own here v.s just passing e.g. `retry.Options{InitialBackoff: 1 * time.Second}` ? 
</blockquote></details>

My thoughts are that there are different kinds of errors. If it's a network error (but not S3 auth related) then we can probably retry right away. But if it's specifically THIS error, then we should wait a bit to appease the rate limiter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19556)*

dt:
:lgtm:

---

---

*[pkg/ccl/storageccl/export_storage.go, line 373 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19556#-KxOzUNYxtlINgZJStzy:-KxPFrbYLe_EG7MBKhPC:b-jnts55) ([raw file](https://github.com/cockroachdb/cockroach/blob/9394a49a0a0082f715e8a2dff59168ab72d2d125/pkg/ccl/storageccl/export_storage.go#L373)):*

My thoughts are that there are different kinds of errors. If it's a network error (but not S3 auth related) then we can probably retry right away. But if it's specifically THIS error, then we should wait a bit to appease the rate limiter.
</blockquote></details>

fair enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19556#-:-KxPFw8UJMEH2v5usNL3:bnfp4nl)*

PR #19563, opened 2017-10-26
cherrypick-1.1: settings: avoid multiple *settings.Values at runtime

### LINKED ISSUE #19533: log, settings: no crash reports from 1.1.x

dt:
This eliminates all but one non-testing call to MakeClusterSettings -- which is critical to correctness.
Any *settings.Values other than the one maintained by the refresh loop will be stale/lies, so only in
testing do we want call sites creating settings value containers.

Fixes #19533.


dt:
cc @cockroachdb/release

bdarnell (approved):

Issue #19555, opened 2017-10-26
sql: parser is too permissive for ANY/ALL/SOME expressions


jordanlewis:
We accept expressions with the `ANY`, `ALL` and `SOME` with or without parentheses in their antecedents. For example, we permit the query:

```
SELECT 1 = ANY ARRAY[1];
```

This isn't valid SQL in PostgreSQL or MySQL, both of which require the antecedent to be parenthesized. We should probably require the same thing, because otherwise our precedence rules create confusing situations like the following:

```
root@:26257/> SELECT 1 = any array[1]:::int[];
pq: incompatible type annotation for 1 = ANY ARRAY[1] as int[], found type: bool
```

Relatedly, our expression formatter for this variety of expression omits the enclosing parentheses unless the antecedent contains another comparison expression. This is an actual bug, causing expressions to fail to roundtrip. It's solved in #19424.

nvanbenschoten:
This was discussed during the original ANY/ALL/SOME code review (https://github.com/cockroachdb/cockroach/pull/12102#issuecomment-265269381). At the time, we weren't aware that it would cause any issues. Can this be addressed by changing our precedence rules?

nvanbenschoten:
I haven't tested this yet, but it may be that `d_expr` needs a rule for type casts and type annotations. Something like:
```
d_expr:
| d_expr TYPECAST cast_target
  {
    $$.val = ...
  }
| d_expr TYPEANNOTATE typename
  {
    $$.val = ...
  }
```

jordanlewis:
Yeah we might be able to just fix it this way. I'll take a look. In the meantime, #19569 to fix the broken.

PR #19569, opened 2017-10-26
sql: disambiguate ANY/ALL/SOME exprs in formatting

jordanlewis:
Previously, ANY/ALL/SOME expressions would be formatted without
parentheses around their antecedents. This could lead to invalid input.
For example, the following query doesn't parse properly since the type
annotation binds to the comparison expression, not the array:

```
SELECT 1 = ANY ARRAY[1]:::INT[]
```

This patch adds the disambiguating enclosing parenthesis.

Fixes the roundtrip problems mentioned in #19555.


jordanlewis:
Split from #19424

nvanbenschoten:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19569#-:-KxQ4hcEYxY3Haup81ct:bnfp4nl)*

jordanlewis:
TFTR

PR #19572, opened 2017-10-26
util: Move one-off methods into their corresponding sub-packages

a-robinson:
Just minor package hygiene -- there doesn't appear to be any reason keep these separated.


benesch (approved):
🙌

PR #19575, opened 2017-10-26
sql: fresh planners for queries run by virt tbls

jordanlewis:
forEachUser, crdb_internal.zones and crdb_internal.jobs previously ran
their `SELECT` statements using the planner that invoked it. This is
dangerous, since running plans can modify their associated planners.

Switch to creating a fresh internal planner for the queries, which is the
pattern that we use almost everywhere else for internal queries.


jordanlewis:
Split out from #19424.

jordanlewis:
Can someone give this a quick look?

justinj:
LGTM, not strictly part of this PR but I think `finishInternalPlanner` also close the memory account associated with the planner's EvalContext? (via `EvalContext.Stop()`).

jordanlewis:
TFTR!

PR #19578, opened 2017-10-26
sqlccl: cleanup failed or canceled RESTORE data

### LINKED ISSUE #17123: sqlccl: tests for BACKUP/RESTORE cancellations
In addition to tests that verify a BACKUP/RESTORE keeps going if the coordinating node goes down, we should also verify that if a user sends a BACKUP/RESTORE from `./cockroach sql` and ctrl-c's (or via golang's `sql.DB` and cancels the context) that it does _not_ keep going. This test should also verify that it tears down all the outstanding Export and Import requests.

### LINKED ISSUE #19398: sqlccl: restore failed, can't delete leftover data
Performed a multi gig restore tonight. It failed, now the restored data range is taking up space and can't be deleted. This is a known problem, but we thought it was theoretically impossible, so we weren't going to prioritize it too much. However apparently it can happen. Total table sizes is < 500MB, but cluster reports using 16GB.

See #19399 for failure reason.

madelynnblue:
Previously, if a RESTORE failed or was manually canceled, any data
it had committed would have been orphaned, and forever taken up
space. Since there was no table descriptor, nothing would have been
able to see or delete it.

Add optional callbacks in jobs code to allow for specific jobs
to implement an onFail method. Add the job's client.Txn to the
update method and this onFail callback to allow for a job to be
transactionally failed and cleaned up.

Release note: Cleanup partially restored data when a RESTORE fails
or is canceled.

Fixes #19398
Fixes #17123

@vivekmenezes could you review the table drop stuff in restore.go?

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -783,6 +783,16 @@ func TestBackupRestoreControlJob(t *testing.T) {
 	}(jobs.DefaultAdoptInterval)
 	jobs.DefaultAdoptInterval = 100 * time.Millisecond
 
+	serverArgs := base.TestServerArgs{}
+	// Disable external processing of mutations so that the final check of
+	// crdb_internal.tables is guaranteed to not be cleaned up. Although this
+	// was never observed by a stress test, it is here for safety.
+	serverArgs.Knobs.SQLSchemaChanger = &sql.SchemaChangerTestingKnobs{
+		AsyncExecNotification: func() error {
+			return errors.New("async schema changer disabled")
+		},
+	}
+
 	// PAUSE JOB and CANCEL JOB are racy in that it's hard to guarantee that the
 	// job is still running when executing a PAUSE or CANCEL--or that the job has
 	// even started running. To synchronize, we install a store response filter
@@ -794,7 +804,7 @@ func TestBackupRestoreControlJob(t *testing.T) {
 	// responses until we close the channel, and our backup or restore is large
 	// enough that it will generate more than one export or import response.
 	var allowResponse chan struct{}
-	params := base.TestClusterArgs{}
+	params := base.TestClusterArgs{ServerArgs: serverArgs}
 	params.ServerArgs.Knobs.Store = &storage.StoreTestingKnobs{
 		TestingResponseFilter: func(ba roachpb.BatchRequest, br *roachpb.BatchResponse) *roachpb.Error {
 			for _, res := range br.Responses {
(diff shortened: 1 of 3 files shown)


pkg/ccl/sqlccl/backup_test.go
```diff
@@ -864,9 +864,53 @@ func TestBackupRestoreControlJob(t *testing.T) {
...
+	}); err != nil {
+		t.Fatal(err)
+	}
+	_, err = sqlDB.DB.Exec(`RESTORE data.* FROM $1 WITH OPTIONS ('into_db'='restore')`, nodelocalDir)
+	if !testutils.IsError(err, "sst: no such file") {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	// Verify the failed RESTORE added some DROP tables.
+	sqlDB.CheckQueryResults(
+		`SELECT name FROM crdb_internal.tables WHERE database_name = 'restore' AND state = 'DROP'`,
```
benesch: I worry this might be flaky. What's to stop the schema changer from immediately cleaning up the failed backup, before this query runs? Maybe there's something we can do with `SchemaChangerTestingKnobs.AsyncExecNotification`.

I suppose if stress/stressrace seem happy I'm not particularly fussed.

pkg/sql/jobs/jobs.go
```diff
@@ -276,6 +277,11 @@ func (j *Job) Canceled(ctx context.Context) error {
 			return false, fmt.Errorf("job with status %s cannot be canceled", *status)
 		}
 		*status = StatusCanceled
+		if onfail, ok := payload.Details.(onFailer); ok {
+			if err := onfail.onFail(ctx, txn, j); err != nil {
+				return false, err
+			}
+		}
```
benesch: 👍

pkg/ccl/sqlccl/restore.go
```diff
@@ -1153,6 +1153,52 @@ func doRestorePlan(
...
+	}
+	return backupDescs, sqlDescs, nil
+}
+
+// restoreFailHook removes KV data that has been committed from a restore that
+// has failed or been canceled. It does this by adding the table descriptors
+// in DROP state, which causes the schema change stuff to delete the keys
+// in the background.
+func restoreFailHook(ctx context.Context, txn *client.Txn, settings *cluster.Settings, details *jobs.RestoreDetails) error {
+	// Needed to trigger the schema change manager.
```
benesch: More precisely it's needed to trigger a re-gossiping of the system config. Any transaction that touches the system config span is supposed to set this, which I'm realizing we don't do elsewhere in restore. I've filed #19617 to track, because `SetSystemConfigTrigger` confuses me every time I see it.

benesch (commented):
Your approach looks entirely perfect—just see my one little note about potential flakiness in the test.

pkg/sql/jobs/jobs.go
```diff
@@ -308,11 +319,26 @@ func (j *Job) Failed(ctx context.Context, err error) {
...
+}
+
+var _ onFailer = &Payload_Restore{}
+
+// RestoreFailHook is the func that is run when a RESTORE job has failed.
+var RestoreFailHook func(context.Context, *client.Txn, *cluster.Settings, *RestoreDetails) error
+
+func (r *Payload_Restore) onFail(ctx context.Context, txn *client.Txn, job *Job) error {
+	// This is set in the CCL code if RESTOREs are enabled, so no need to check
+	// for nil.
```
benesch: Heh, looks like this isn't true when running the tests for the jobs package. 🤦‍♂️

mjibson:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/backup_test.go, line 909 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578#-KxPMcHrLk9MFgxPJAnE-r1-909:-Kxg70sMsdzWi3OS_eri:bn6111w) ([raw file](https://github.com/cockroachdb/cockroach/blob/89b5673d8d5f7a8019184e990e1a164c095234e8/pkg/ccl/sqlccl/backup_test.go#L909)):*

I worry this might be flaky. What's to stop the schema changer from immediately cleaning up the failed backup, before this query runs? Maybe there's something we can do with `SchemaChangerTestingKnobs.AsyncExecNotification`.

I suppose if stress/stressrace seem happy I'm not particularly fussed.
</blockquote></details>

A stress didn't show this, but I've added it anyway just in case. Good eye.

---

*[pkg/ccl/sqlccl/restore.go, line 1177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578#-KxPMcHrLk9MFgxPJAnF-r1-1177:-Kxg7DhMuEBQFFIOHj8T:b-2rsho) ([raw file](https://github.com/cockroachdb/cockroach/blob/89b5673d8d5f7a8019184e990e1a164c095234e8/pkg/ccl/sqlccl/restore.go#L1177)):*

More precisely it's needed to trigger a re-gossiping of the system config. Any transaction that touches the system config span is supposed to set this, which I'm realizing we don't do elsewhere in restore. I've filed #19617 to track, because `SetSystemConfigTrigger` confuses me every time I see it.
</blockquote></details>

Ah, TIL. I had no idea about this until Vivek told me it was causing the lack of processing.

---

*[pkg/sql/jobs/jobs.go, line 333 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578#-KxPMcHrLk9MFgxPJAnG-r1-333:-Kxg83TIH_qWU9YOonIS:b-z0othp) ([raw file](https://github.com/cockroachdb/cockroach/blob/89b5673d8d5f7a8019184e990e1a164c095234e8/pkg/sql/jobs/jobs.go#L333)):*

Heh, looks like this isn't true when running the tests for the jobs package. 🤦‍♂️ 
</blockquote></details>

Oh indeed. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578)*

benesch:
:lgtm:

---

Reviewed 3 of 3 files at r2.

---

*[pkg/ccl/sqlccl/backup_test.go, line 797 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578#-Kxhy7ho_dWDJu0y7Zfa:-Kxhy7ho_dWDJu0y7Zfb:b-a3ito) ([raw file](https://github.com/cockroachdb/cockroach/blob/71c1a1e51d0f9b99be35615f36771cbb5588cd97/pkg/ccl/sqlccl/backup_test.go#L797)):*
> ```Go
> 	// enough that it will generate more than one export or import response.
> 	var allowResponse chan struct{}
> 	params := base.TestClusterArgs{}
> ```

I think you want the schema change knob here, too?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578#-:-KxhyBYDnfnV8vPs1prz:bnfp4nl)*

mjibson:

---

*[pkg/ccl/sqlccl/backup_test.go, line 797 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578#-Kxhy7ho_dWDJu0y7Zfa:-KxirrKy_FibTgAiRxVb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/71c1a1e51d0f9b99be35615f36771cbb5588cd97/pkg/ccl/sqlccl/backup_test.go#L797)):*

I think you want the schema change knob here, too?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19578)*

PR #19580, opened 2017-10-26
ui: use ReactRouter.Link's for active status

couchand:
This is a refactor to use ReactRouter.Link's built-in activeClassName
prop rather than rolling our own calculation of the active class.


mrtracy:
I suppose this is fine; there was some reason that I needed the class to be applied to the `li` element itself, but I can't remember it anymore. 

If we don't need to apply the active class to the `li` itself, then the entire ListLink class should be removed and inlined, applying that class was the entire purpose.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19580#-:-KxiT2b9aNkT3apOdSZu:b-htcr7u)*

couchand:
So I just noticed that there's this comment at the top of `listLink.tsx`:

>  * TODO(mrtracy): During construction of 'next', we are explicitly avoiding any
> * CSS changes from the existing ui. After publishing 'next', we should
> * restructure the css of these links slightly so that the 'active' class is
> * present directly on the <a> element, instead of the containing <li> element;
> * that will allow us to use the react-router's <Link> class to provide the
> * activeClass, rather than the logic below.

I don't think I have the context for that, but I'm guessing we've passed the point in time you were waiting for?

Made the changes.  It turned out to be a bit more than I expected, mostly because of the weird rules for the Raft debug page that could probably use some TLC.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19580#-:-KxilYtUXXCvFFFAVxxJ:bdfjrlc)*

mrtracy:
Hah, that is indeed quite old.  That was when we first rewrote it in react.  Yep, this is the right thing to do, and LGTM.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19580#-:-Kxixjwmv09tZpW19YQe:b-muo4xs)*

PR #19582, opened 2017-10-26
ui build: rebuild when styles are updated

couchand:
Rebuild the UI when only style updates are made.

### DIFF
--- a/pkg/ui/Makefile
+++ b/pkg/ui/Makefile
@@ -62,7 +62,7 @@ test-watch: $(DLLS)
 dist/%.dll.js %-manifest.json: webpack.%.js | protos
 	$(WEBPACK) -p --config $<
 
-$(GOBINDATA_TARGET): webpack.app.js $(DLLS) $(shell find src)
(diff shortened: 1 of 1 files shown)


pkg/ui/Makefile
```diff
@@ -62,7 +62,7 @@ test-watch: $(DLLS)
 dist/%.dll.js %-manifest.json: webpack.%.js | protos
 	$(WEBPACK) -p --config $<
 
-$(GOBINDATA_TARGET): webpack.app.js $(DLLS) $(shell find src)
+$(GOBINDATA_TARGET): webpack.app.js $(DLLS) $(shell find src) $(shell find styl)
```
benesch: nit: `$(shell find src styl)`

(Every shell command that appears in a prerequisite has to be executed on every invocation of `make`, so it's good to limit the number as much as possible.)

benesch (approved):
Thanks for catching this!

pkg/ui/Makefile
couchand: :+1: done

PR #19583, opened 2017-10-26
Add SCRUB TABLE ... INDEX and SCRUB TABLE ... INDEX(idxname) syntax

lgo:
`SCRUB TABLE ... INDEX` is placeholder syntax for the SCRUB feature of
checking secondary indexes. With no indexes specified, this checks all
indexes.

This also adds `SCRUB TABLE ... WITH INDEX(idx_name)`. This provides 
the ability to specify the indexes to check, as opposed to checking them all.

PR #19586, opened 2017-10-26
 sql: Implement SCRUB TABLE ... WITH INDEX check

lgo:
`SCRUB TABLE ... WITH OPTIONS INDEX ...` will run a check on secondary
indexes. In particular, this check will find all missing index
references or dangling index references.

Release note: Added the commands `SCRUB TABLE <table> WITH OPTIONS INDEX
ALL` and `SCRUB TABLE <table> WITH OPTIONS INDEX (<index_name>...)` to
check the integrity of secondary indexes.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/scrub
+++ b/pkg/sql/logictest/testdata/logic_test/scrub
@@ -24,4 +24,3 @@ EXPERIMENTAL SCRUB TABLE t WITH OPTIONS INDEX (name_idx)
 
 statement error specified indexes to check that do not exist on table "t": not_an_index, also_not
 EXPERIMENTAL SCRUB TABLE t WITH OPTIONS INDEX (not_an_index, also_not, name_idx)
------
--- a/pkg/sql/scrub.go
+++ b/pkg/sql/scrub.go
@@ -15,20 +15,59 @@
 package sql
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
-
-	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
-	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
-	"golang.org/x/net/context"
+	"time"
 
 	"github.com/cockroachdb/cockroach/pkg/sql/parser"
+	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
+	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/cockroachdb/cockroach/pkg/util/hlc"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+)
+
+const (
+	// ScrubErrorMissingIndexEntry occurs when a primary k/v is missing a
+	// corresponding secondary index k/v.
+	ScrubErrorMissingIndexEntry = "missing_index_entry"
+	// ScrubErrorDanglingIndexReference occurs when a secondary index k/v
+	// points to a non-existing primary k/v.
+	ScrubErrorDanglingIndexReference = "dangling_index_reference"
 )
 
 type scrubNode struct {
 	optColumnsSlot
 
-	n *parser.Scrub
+	n         *parser.Scrub
+	indexes   []sqlbase.IndexDescriptor
+	tableDesc *sqlbase.TableDescriptor
+	tableName *parser.TableName
+
+	currentIndex *checkedIndex
+
+	row parser.Datums
+}
+
+// checkedIndex holds the intermediate results for one index that is
+// checked.
+type checkedIndex struct {
+	// Intermediate values.
+	rows     *sqlbase.RowContainer
+	rowIndex int
+
+	// Context of the index check.
+	databaseName string
+	tableName    string
+	indexName    string
+	// columns is a list of the columns returned by one side of the
+	// queries join. The actual resulting rows from the RowContainer is
+	// twice this.
+	columns []*sqlbase.ColumnDescriptor
+	// primaryIndexColumnIndexes maps PrimaryIndex.Columns to the row
+	// indexes in the query result parser.Datums.
+	primaryIndexColumnIndexes []int
 }
 
 // Scrub checks the database.
@@ -48,37 +87,41 @@ var scrubColumns = sqlbase.ResultColumns{
 	{Name: "primary_key", Typ: parser.TypeString},
 	{Name: "timestamp", Typ: parser.TypeTimestamp},
 	{Name: "repaired", Typ: parser.TypeBool},
-	{Name: "details", Typ: parser.TypeString},
+	{Name: "details", Typ: parser.TypeJSON},
 }
 
 func (n *scrubNode) Start(params runParams) error {
-	tn, err := n.n.Table.NormalizeWithDatabaseName(params.p.session.Database)
+	var err error
+	n.tableName, err = n.n.Table.NormalizeWithDatabaseName(
+		params.p.session.Database)
 	if err != nil {
 		return err
 	}
 
-	tableDesc, err := params.p.getTableDesc(params.ctx, tn)
+	n.tableDesc, err = params.p.getTableDesc(params.ctx, n.tableName)
 	if err != nil {
 		return err
 	}
 
-	if tableDesc.IsView() {
-		return pgerror.NewErrorf(pgerror.CodeWrongObjectTypeError, "cannot run SCRUB on views")
+	if n.tableDesc.IsView() {
+		return pgerror.NewErrorf(pgerror.CodeWrongObjectTypeError,
+			"cannot run SCRUB on views")
 	}
 
-	// Process SCRUB options
-	var indexes []*sqlbase.IndexDescriptor
+	// Process SCRUB options.
+	var indexesSet bool
 	for _, option := range n.n.Options {
 		switch v := option.(type) {
 		case *parser.ScrubOptionIndex:
-			if indexes != nil {
+			if indexesSet {
 				return pgerror.NewErrorf(pgerror.CodeSyntaxError,
 					"cannot specify INDEX option more than once")
 			}
-			indexes, err = indexesToCheck(v.IndexNames, tableDesc)
+			n.indexes, err = indexesToCheck(v.IndexNames, n.tableDesc)
 			if err != nil {
 				return err
 			}
+			indexesSet = true
 		default:
 			panic(fmt.Sprintf("Unhandled SCRUB option received: %+v", v))
 		}
@@ -86,30 +129,282 @@ func (n *scrubNode) Start(params runParams) error {
 
 	// No options were provided. By default exhaustive checks are run.
 	if len(n.n.Options) == 0 {
-		indexes, err = indexesToCheck(nil /* indexNames */, tableDesc)
+		n.indexes, err = indexesToCheck(nil /* indexNames */, n.tableDesc)
 		if err != nil {
 			return err
 		}
 	}
 
-	// TODO(joey): Squelch the "indexes not used" lint error for now.
-	_ = indexes
-
 	return nil
+}
+
+// startIndexCheck will plan and run an index check using the distSQL
+// execution engine. The returned value contains the row iterator and
+// the information about the index that was checked.
+func (n *scrubNode) startIndexCheck(
+	ctx context.Context,
+	p *planner,
+	tableDesc *sqlbase.TableDescriptor,
+	tableName *parser.TableName,
+	indexDesc *sqlbase.IndexDescriptor,
+) (*checkedIndex, error) {
+	colToIdx := make(map[sqlbase.ColumnID]int)
+	for i, col := range tableDesc.Columns {
+		colToIdx[col.ID] = i
+	}
+
+	// Collect all of the columns we are fetching from the secondary
+	// index. This includes the columns involved in the index: columns,
+	// extra columns, and store columns.
+	var columns []*sqlbase.ColumnDescriptor
+	for _, colID := range indexDesc.ColumnIDs {
+		columns = append(columns, &tableDesc.Columns[colToIdx[colID]])
+	}
+	for _, colID := range indexDesc.ExtraColumnIDs {
+		columns = append(columns, &tableDesc.Columns[colToIdx[colID]])
+	}
+	for _, colID := range indexDesc.StoreColumnIDs {
+		columns = append(columns, &tableDesc.Columns[colToIdx[colID]])
+	}
+
+	// Collect the column names and types.
+	var columnNames []string
+	var columnTypes []sqlbase.ColumnType
+	for _, col := range columns {
+		columnNames = append(columnNames, col.Name)
+		columnTypes = append(columnTypes, col.Type)
+	}
+
+	// Because the row results include both primary key data and secondary
+	// key data, the row results will contain two copies of the column
+	// data.
+	columnTypes = append(columnTypes, columnTypes...)
+
+	// Find the row indexes for all of the primary index columns.
+	var primaryColumnRowIndexes []int
+	for i, colID := range tableDesc.PrimaryIndex.ColumnIDs {
+		rowIdx := -1
+		for idx, col := range columns {
+			if col.ID == colID {
+				rowIdx = idx
+				break
+			}
+		}
+		if rowIdx == -1 {
+			return nil, errors.Errorf(
+				"could not find primary index column in projection: columnID=%d columnName=%s",
+				colID,
+				tableDesc.PrimaryIndex.ColumnNames[i])
+		}
+		primaryColumnRowIndexes = append(primaryColumnRowIndexes, rowIdx)
+	}
+
+	checkQuery := createIndexCheckQuery(columnNames,
+		tableDesc.PrimaryIndex.ColumnNames, tableName, indexDesc.ID)
+	plan, err := p.delegateQuery(ctx, "SCRUB TABLE ... WITH OPTIONS INDEX", checkQuery, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// All columns projected in the plan generated from the query are
+	// needed. The columns are the index columns and extra columns in the
+	// index, twice -- for the primary and then secondary index.
+	needed := make([]bool, len(planColumns(plan)))
+	for i := range needed {
+		needed[i] = true
+	}
+
+	// Optimize the plan. This is required in order to populate scanNode
+	// spans.
+	plan, err = p.optimizePlan(ctx, plan, needed)
+	if err != nil {
+		plan.Close(ctx)
+		return nil, err
+	}
+	defer plan.Close(ctx)
+
+	rows, err := scrubIndexRunDistSQL(ctx, p, plan, columnTypes)
+	if err != nil {
+		rows.Close(ctx)
+		return nil, err
+	} else if rows == nil {
+		return nil, nil
+	}
+
+	return &checkedIndex{
+		rows:                      rows,
+		tableName:                 tableName.Table(),
+		databaseName:              tableName.Database(),
+		indexName:                 indexDesc.Name,
+		primaryIndexColumnIndexes: primaryColumnRowIndexes,
+		columns:                   columns,
+	}, nil
+}
+
+// scrubIndexRunDistSQL will prepare and run the plan in distSQL. If a
+// RowContainer is returned the caller must close it.
+func scrubIndexRunDistSQL(
+	ctx context.Context, p *planner, plan planNode, columnTypes []sqlbase.ColumnType,
+) (*sqlbase.RowContainer, error) {
+	ci := sqlbase.ColTypeInfoFromColTypes(columnTypes)
+	rows := sqlbase.NewRowContainer(*p.evalCtx.ActiveMemAcc, ci, 0)
+	rowResultWriter := NewRowResultWriter(parser.Rows, rows)
+	recv, err := makeDistSQLReceiver(
+		ctx,
+		rowResultWriter,
+		p.ExecCfg().RangeDescriptorCache,
+		p.ExecCfg().LeaseHolderCache,
+		p.txn,
+		func(ts hlc.Timestamp) {
+			_ = p.ExecCfg().Clock.Update(ts)
+		},
+	)
+	if err != nil {
+		return rows, err
+	}
+
+	err = p.session.distSQLPlanner.PlanAndRun(ctx, p.txn, plan, &recv, p.evalCtx)
+	if err != nil {
+		return rows, err
+	} else if recv.err != nil {
+		return rows, recv.err
+	}
+
+	if rows.Len() == 0 {
+		rows.Close(ctx)
+		return nil, nil
+	}
+
+	return rows, nil
+}
+
+// tableColumnsIsNullPredicate creates a predicate that checks if all of
+// the specified columns for a table are NULL. For example, given table
+// is t1 and the columns id, name, data, then the returned string is:
+//
+//   t1.id IS NULL AND t1.name IS NULL AND t1.data IS NULL
+func tableColumnsIsNullPredicate(tableName string, columns []string) string {
+	var buf bytes.Buffer
+	for i, col := range columns {
+		if i > 0 {
+			buf.WriteString(" AND ")
+		}
+		fmt.Fprintf(&buf, "%[1]s.%[2]s IS NULL", tableName, col)
+	}
+	return buf.String()
+}
+
+// tableColumnsEQ creates a predicate that checks if all of the
+// specified columns for two tables are equal. For example, given tables
+// t1, t2 and the columns id, name, data, then the returned string is:
+//
+//   t1.id = t2.id AND t1.name = t2.name AND t1.data = t2.data
+func tableColumnsEQ(tableName string, otherTableName string, columns []string) string {
+	var buf bytes.Buffer
+	for i, col := range columns {
+		if i > 0 {
+			buf.WriteString(" AND ")
+		}
+		fmt.Fprintf(&buf, "%[1]s.%[3]s = %[2]s.%[3]s", tableName, otherTableName, col)
+	}
+	return buf.String()
+}
+
+// tableColumnsProjection creates the select projection statement (a
+// comma delimetered column list), for the specified table and
+// columns. For example, if the table is t1 and the columns are id,
+// name, data, then the returned string is:
+//
+//   t1.id, t1.name, t1.data
+func tableColumnsProjection(tableName string, columns []string) string {
+	var buf bytes.Buffer
+	for i, col := range columns {
+		if i > 0 {
+			buf.WriteString(", ")
+		}
+		fmt.Fprintf(&buf, "%[1]s.%[2]s", tableName, col)
+	}
+	return buf.String()
+}
+
+// createIndexCheckQuery will make the index check query for a given
+// table and secondary index.
+// For example, given the following table schema:
+//
+//   CREATE TABLE test (
+//     k INT, s INT, v INT, misc INT,
+//     PRIMARY KEY (k, s),
+//     INDEX v_idx (v),
+//   )
+//
+// The generated query to check the `v_idx` will be:
+//
+//   SELECT left.k, left.s, left.v, right.k, right.s. right.v
+//   FROM
+//     test@{NO_INDEX_JOIN} as left
+//   FULL OUTER JOIN
+//     test@{FORCE_INDEX=v_idx,NO_INDEX_JOIN} as right
+//     ON left.k = right.k AND
+//        left.s = right.s AND
+//        left.v = right.v
+//   WHERE (left.k  IS NULL AND left.s  IS NULL) OR
+//         (right.k IS NULL AND right.s IS NULL)
+//
+// In short, this query is:
+// 1) Scanning the primary index and the secondary index.
+// 2) Joining them on all of the secondary index columns and
+//    extra columns that are equal.
+// 3) Filtering to achieve an anti-join. The first line of the predicate
+//    takes rows on the right for the anti-join. The second line of the
+//    predicate takes rows on the left for the anti-join.
+//
+// Because this is an anti-join, the results are as follows:
+// - If any primary index column on the left is NULL, that means the
+//   right columns are present. This is because of the invariant that
+//   primary index columns are never null.
+// - Otherwise, the left columns is present.
+//
+// TODO(joey): Once we support ANTI JOIN in distSQL this can be
+// simplified, as the the WHERE clause can be completely dropped.
+func createIndexCheckQuery(
+	columnNames []string,
+	primaryKeyColumnNames []string,
+	tableName *parser.TableName,
+	indexID sqlbase.IndexID,
+) string {
+	const checkIndexQuery = `
+				SELECT %[1]s, %[2]s
+				FROM
+					%[3]s@{NO_INDEX_JOIN} as leftside
+				FULL OUTER JOIN
+					%[3]s@{FORCE_INDEX=[%[4]d],NO_INDEX_JOIN} as rightside
+					ON %[5]s
+        WHERE (%[6]s) OR
+              (%[7]s)`
+
+	return fmt.Sprintf(checkIndexQuery,
+		tableColumnsProjection("leftside", columnNames),
+		tableColumnsProjection("rightside", columnNames),
+		tableName,
+		indexID,
+		tableColumnsEQ("leftside", "rightside", columnNames),
+		tableColumnsIsNullPredicate("leftside", primaryKeyColumnNames),
+		tableColumnsIsNullPredicate("rightside", primaryKeyColumnNames),
+	)
 }
 
 // indexesToCheck will return all of the indexes that are being checked.
 // If indexNames is nil, then all indexes are returned.
+// TODO(joey): This can be simplified with
+// TableDescriptor.FindIndexByName(), but this will only report the
+// first invalid index.
 func indexesToCheck(
 	indexNames parser.NameList, tableDesc *sqlbase.TableDescriptor,
-) (results []*sqlbase.IndexDescriptor, err error) {
+) (results []sqlbase.IndexDescriptor, err error) {
 	if indexNames == nil {
 		// Populate results with all secondary indexes of the
 		// table.
-		for _, idx := range tableDesc.Indexes {
-			results = append(results, &idx)
-		}
-		return results, nil
+		return tableDesc.Indexes, nil
 	}
 
 	// Find the indexes corresponding to the user input index names.
@@ -119,7 +414,7 @@ func indexesToCheck(
 	}
 	for _, idx := range tableDesc.Indexes {
 		if _, ok := names[idx.Name]; ok {
-			results = append(results, &idx)
+			results = append(results, idx)
 			delete(names, idx.Name)
 		}
 	}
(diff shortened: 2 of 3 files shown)


lego:
:)
```
root@:26257/test> experimental scrub table t2 with index;
+----------+--------------------------+----------+-------+-------------+----------------------------------+----------+------------------------------------------------------------+
| job_uuid |        error_type        | database | table | primary_key |            timestamp             | repaired |                          details                           |
+----------+--------------------------+----------+-------+-------------+----------------------------------+----------+------------------------------------------------------------+
| NULL     | missing_index_reference  | test     | t2    | (4)         | 2017-10-26 21:42:32.319766+00:00 | false    | {"index_name":"id2idx","row_data":{"id":"4","id2":"2"}}    |
| NULL     | missing_index_reference  | test     | t2    | (30)        | 2017-10-26 21:42:32.319766+00:00 | false    | {"index_name":"id2idx","row_data":{"id":"30","id2":"20"}}  |
| NULL     | dangling_index_reference | test     | t2    | (2)         | 2017-10-26 21:42:32.319766+00:00 | false    | {"index_name":"id2idx","row_data":{"id":"2","id2":"NULL"}} |
| NULL     | dangling_index_reference | test     | t2    | (4)         | 2017-10-26 21:42:32.319766+00:00 | false    | {"index_name":"id2idx","row_data":{"id":"4","id2":"3"}}    |
| NULL     | dangling_index_reference | test     | t2    | (10)        | 2017-10-26 21:42:32.319766+00:00 | false    | {"index_name":"id2idx","row_data":{"id":"10","id2":"10"}}  |
+----------+--------------------------+----------+-------+-------------+----------------------------------+----------+------------------------------------------------------------+
(5 rows)
```
(this is correct as the KVs were surgically changed after slicing out parts that update/create secondary KVs)

lego:
Review status: 0 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/scrub.go, line 38 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxPfHkDcCfGbKpZrqXP:-KxPfHkDcCfGbKpZrqXQ:bws6u8l) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b4ee98d921697fbf71a29e3a0db80de6b23aca4/pkg/sql/scrub.go#L38)):*
> ```Go
> 	n *parser.Scrub
> 
> 	checkedIndexes []*checkedIndex
> ```

This is a bit of a hack. During `n.Start` we plan and run all of the index checks, then toss the row containers in here. Then, later, we iterate through these `checkIndexes` for each of their row containers in the `n.Next` iterator.

---

*[pkg/sql/scrub.go, line 134 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxPflbtJ5WhCaZqD9__:-KxPflbtJ5WhCaZqD9_a:b-tohiyc) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b4ee98d921697fbf71a29e3a0db80de6b23aca4/pkg/sql/scrub.go#L134)):*
> ```Go
> // createIndexCheckQuery will make the index check query for a table and
> // secondary index.
> func createIndexCheckQuery(
> ```

This whooping function is just one giant string formatter. I don't think there was a better way to approach this because of all the commonness, so they are the chunks are broken up into closures. I'm definitely open to suggestions on this one.

---

*[pkg/sql/scrub.go, line 290 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxPgSCxCCUqxUzraRgq:-KxPgSCxCCUqxUzraRgr:b-ooxams) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b4ee98d921697fbf71a29e3a0db80de6b23aca4/pkg/sql/scrub.go#L290)):*
> ```Go
> 
> 	ci := sqlbase.ColTypeInfoFromColTypes(columnTypes)
> 	// FIXME(joey): Usage of rows.Close() may be incorrect. Currently it is:
> ```

I was a bit uncertain on the use of `rows.Close()`. Does it need only need to be closed if we receive rows? i.e. only in `Next`

---

*[pkg/sql/scrub.go, line 376 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxPgf-RWXIgsA_kExqz:-KxPgf-RWXIgsA_kExr-:b-42iacw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b4ee98d921697fbf71a29e3a0db80de6b23aca4/pkg/sql/scrub.go#L376)):*
> ```Go
> func (n *scrubNode) Next(params runParams) (bool, error) {
> 	if n.checkedIndexes != nil && len(n.checkedIndexes) > 0 {
> 		checkedIndex := n.checkedIndexes[0]
> ```

Related to the top comment, this is a bit hacky where we iterate through each of the checked index "buffer".

---

*[pkg/sql/scrub.go, line 424 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxPgptcOjQ_Vx-w2Gko:-KxPgptcOjQ_Vx-w2Gkp:brdplel) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b4ee98d921697fbf71a29e3a0db80de6b23aca4/pkg/sql/scrub.go#L424)):*
> ```Go
> 			for rowIdx, col := range checkedIndex.columns {
> 				// FIXME(joey): Should we maybe try to get the underlying type?
> 				rowDetails[col.Name] = row[rowIdx].String()
> ```

The `details` column is intended to be an unstructured field for the check details that are specific to only a few checks. At the moment this is formatted as a string, and so we end up building JSON to marshal it into this string.

I don't think there's a better approach to this. Also as this will be marked with `EXPERIEMENTAL` I'd assume that there is no problem changing the return column type to later use `JSON`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

lego:
I still haven't been able to get the tests working. After poking around all the functions on `"database/sql".Rows`, the columns and column types seem fine and doing `NextRowSet` does not seem to help. I'm beginning to think it might be how the plan execution works, so if you see anything off in there lmk. The query does work from the CLI though.

lego:
Woo nevermind, figured out the problem with the test -- it was an unchecked error from `rows.Scan(...)` :)

jordanlewis:
Review status: 0 of 10 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/scrub.go, line 138 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxntJTKjt2h0qP2oLRl:-KxntJTKjt2h0qP2oLRm:b2dat76) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L138)):*
> ```Go
> 	// Run the index check on each index.
> 	if indexes != nil {
> 		n.checkedIndexes = []*checkedIndex(nil)
> ```

I don't think you need this - `append(nil, something)` makes a valid slice! So in other words where you append to this down below will just work without explicit initialization. 

---

*[pkg/sql/scrub.go, line 140 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo6n1u_i_ArqBEsVSL:-Kxo6n1u_i_ArqBEsVSM:b1hcwc5) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L140)):*
> ```Go
> 		n.checkedIndexes = []*checkedIndex(nil)
> 		for _, index := range indexes {
> 			if err := n.startIndexCheck(params.ctx, params.p, tableDesc, tn, index); err != nil {
> ```

Hmm so is this going to create `n` concurrent `FULL OUTER JOIN` queries then? That seems quite less than ideal - it'd be better to limit the resource usage of this as much as possible, and I don't see a great reason why these should be all kicked off at once. Better to kick them off on demand as `Next()` runs out of rows in a particular index check.

Speaking of resource usage, this could be a good use case for the feature of running a query with limited memory usage that @knz has been talking about a little bit. It would be unfortunate if a `CHECK` command accidentally used up a ton of cluster memory and caused outages - of course it will be pausable when it gains jobs support, but still.

---

*[pkg/sql/scrub.go, line 185 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxnwcVKtKR5_Z3Lw7lw:-KxnwcVKtKR5_Z3Lw7lx:bhq04pw) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L185)):*
> ```Go
> 
> 	// Find the row indexes for all of the primary index columns.
> 	primaryColumnRowIndexes := []int(nil)
> ```

It's more Go-stylish to do `var primaryColumnRowIndexes []int`.

---

*[pkg/sql/scrub.go, line 278 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo220MIftYalAaXmQc:-Kxo220MIftYalAaXmQd:bysdqm2) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L278)):*
> ```Go
> 	columns []*sqlbase.ColumnDescriptor, tableName *parser.TableName, indexID sqlbase.IndexID,
> ) string {
> 	const checkIndexQuery = `
> ```

holy string interpolation batman! I guess this can't be helped with placeholders since most of the inputs are names.

Having jumped back to this thing multiple times though while reading the code that reads the results from this query I'd really appreciate if you include a fully interpolated example query in a comment above this.

---

*[pkg/sql/scrub.go, line 290 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxnymaNsodzvGcNV1pp:-KxnymaNsodzvGcNV1pq:boimsw9) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L290)):*
> ```Go
> 	// the specified columns for a table are NULL or NOT NULL, as
> 	// indicated by isNull.
> 	tableColumnsNullPredicate := func(
> ```

I don't think there's a strong reason that these functions need to be closures - I think it's more clear to make them ordinary functions and put their definitions above this one.

---

*[pkg/sql/scrub.go, line 397 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo73LL9QvC-AKuzn8C:-Kxo73LL9QvC-AKuzn8D:bqvb1jq) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L397)):*
> ```Go
> 		// in row is twice that, as it includes row data for both the
> 		// primary index and the secondary index.
> 		rowDataSize := len(row) / 2
> ```

Seems like you should probably store this when you create it rather than do a scary divide by 2 every time - it'll be clearer I think! Otherwise the reader of this code has to reason about why the row length is always even.

---

*[pkg/sql/scrub.go, line 399 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7IX2PGSKA4B95Axl:-Kxo7IX2PGSKA4B95Axm:b-3tillt) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L399)):*
> ```Go
> 		rowDataSize := len(row) / 2
> 
> 		// If any of the primary key values is present in the primary index
> ```

s/is/are/

---

*[pkg/sql/scrub.go, line 401 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7VasvJJldWZPHnVq:-Kxo7VasvJJldWZPHnVr:bad12d) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L401)):*
> ```Go
> 		// If any of the primary key values is present in the primary index
> 		// row data, then this row has no corresponding secondary index
> 		// entry.
> ```

I don't follow this logic. Why is that true?

---

*[pkg/sql/scrub.go, line 403 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7cyA25q2vCCI5Uuc:-Kxo7NlgoAhPcuh7MX3O:b-bkcnyg) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L403)):*
> ```Go
> 		// entry.
> 		var isMissingIndexReferenceError bool
> 		if row[checkedIndex.primaryIndexColumnIndexes[0]] != parser.DNull {
> ```

This explicit [0] seems somewhat suspicious - is it? It deserves a comment anyway. Is this the first column of the primary key?

---

*[pkg/sql/scrub.go, line 407 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo9p-Lkp4HRPz0oRrK:-Kxo9p-Lkp4HRPz0oRrL:b-dl686z) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L407)):*
> ```Go
> 		}
> 
> 		primaryKeyDatums := parser.Datums{}
> ```

more go stylish to say `var primaryKeyDatums parser.Datums`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

lego:
Review status: 0 of 10 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/scrub.go, line 140 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo6n1u_i_ArqBEsVSL:-KxoLOYltQldE6nBdb17:b5zarta) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L140)):*

Hmm so is this going to create `n` concurrent `FULL OUTER JOIN` queries then? That seems quite less than ideal - it'd be better to limit the resource usage of this as much as possible, and I don't see a great reason why these should be all kicked off at once. Better to kick them off on demand as `Next()` runs out of rows in a particular index check.

Speaking of resource usage, this could be a good use case for the feature of running a query with limited memory usage that @knz has been talking about a little bit. It would be unfortunate if a `CHECK` command accidentally used up a ton of cluster memory and caused outages - of course it will be pausable when it gains jobs support, but still.
</blockquote></details>

Does it do it concurrently? (i.e. the distSQL processors wait for us to iterate through streamed results? cool!)

In that case I'm all for moving this to `Next`.

---

*[pkg/sql/scrub.go, line 278 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo220MIftYalAaXmQc:-KxoLylJTVd5bTc_r4Mr:b-eyu0ld) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L278)):*

holy string interpolation batman! I guess this can't be helped with placeholders since most of the inputs are names.

Having jumped back to this thing multiple times though while reading the code that reads the results from this query I'd really appreciate if you include a fully interpolated example query in a comment above this.
</blockquote></details>

Definitely!

---

*[pkg/sql/scrub.go, line 401 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7VasvJJldWZPHnVq:-KxoKR_F1hP-HBgE7HcM:b-f0513m) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L401)):*

I don't follow this logic. Why is that true?
</blockquote></details>

The query is doing an `ANTIJOIN` with the primary index and secondary index, with the join predicate being all columns and extra columns in the index (I omitted store columns for now, although I will probably need to add that.) So either the primary index side will have values, or the secondary index side will.

Admittedly, this should be explicitly stated somewhere for how this check works.

---

*[pkg/sql/scrub.go, line 403 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7cyA25q2vCCI5Uuc:-KxoMiqz3MTAd59EB5kB:bh17fgj) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L403)):*

This explicit [0] seems somewhat suspicious - is it? It deserves a comment anyway. Is this the first column of the primary key?
</blockquote></details>

The specific column doesn't matter, but it comes from the primary index will only be null if the antijoin has only put through the secondary index side in this case.

I agree a comment is warranted.
(For completeness, it's the first column defined in the secondary index iirc)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

jordanlewis:
Review status: 0 of 10 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/scrub.go, line 140 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo6n1u_i_ArqBEsVSL:-KxobIEm7ceQsf1_1ILK:buju6dy) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L140)):*

Does it do it concurrently? (i.e. the distSQL processors wait for us to iterate through streamed results? cool!)

In that case I'm all for moving this to `Next`.
</blockquote></details>

Well you're calling `PlanAndRun` on all of the queries before you call `Next` on any of them, right? I'm not completely sure what DistSQL will do here, but it seems clear that doing even any amount of work per query before the results are strictly needed is not the most efficient way we can do things.

I encourage you to try it out! You could modify the code to put a big sleep before you call Next() on anything, and then run an index check against multiple indexes and see whether queries are active in `SHOW QUERIES`. There's probably even a better way of checking that I don't know about.

---

*[pkg/sql/scrub.go, line 401 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7VasvJJldWZPHnVq:-KxoccsZbelyJiqVVNQB:bz9zn0o) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L401)):*

The query is doing an `ANTIJOIN` with the primary index and secondary index, with the join predicate being all columns and extra columns in the index (I omitted store columns for now, although I will probably need to add that.) So either the primary index side will have values, or the secondary index side will.

Admittedly, this should be explicitly stated somewhere for how this check works.
</blockquote></details>

It would be great to include the sentiment of this review comment in the code somewhere. Perhaps either near the query interpolation itself or right here where we're interpreting the results.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

lego:
Review status: 0 of 10 files reviewed at latest revision, 16 unresolved discussions.

---

*[pkg/sql/scrub.go, line 138 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxntJTKjt2h0qP2oLRl:-KxpZoD9weriXkGOz6nK:b6lhd5a) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L138)):*

I don't think you need this - `append(nil, something)` makes a valid slice! So in other words where you append to this down below will just work without explicit initialization. 
</blockquote></details>

whaaaaaaat that's so much better :). I somehow assumed `append(nil, ...)` wasn't valid ;-; but it makes sense since this is a typed nil. (This slice doesn't exist anymore)

---

*[pkg/sql/scrub.go, line 140 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo6n1u_i_ArqBEsVSL:-KxpXVkxXSsnBDDYRU5j:b-73o079) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L140)):*

Well you're calling `PlanAndRun` on all of the queries before you call `Next` on any of them, right? I'm not completely sure what DistSQL will do here, but it seems clear that doing even any amount of work per query before the results are strictly needed is not the most efficient way we can do things.

I encourage you to try it out! You could modify the code to put a big sleep before you call Next() on anything, and then run an index check against multiple indexes and see whether queries are active in `SHOW QUERIES`. There's probably even a better way of checking that I don't know about.
</blockquote></details>

So `PlanAndRun` does run until completion (it doesn't appear in show queries, but it seems that this is the case after spelunking). Either way, it is cleaner calling this in `Next` so I shuffled things around for that.

---

*[pkg/sql/scrub.go, line 185 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxnwcVKtKR5_Z3Lw7lw:-Kxp_pfNXIZOy1MspT-_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L185)):*

It's more Go-stylish to do `var primaryColumnRowIndexes []int`.
</blockquote></details>

Done.

---

*[pkg/sql/scrub.go, line 278 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo220MIftYalAaXmQc:-Kxro_7sPeNcsXzBwvKO:bmfxgaw) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L278)):*

Definitely!
</blockquote></details>

Done :)

---

*[pkg/sql/scrub.go, line 290 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxnymaNsodzvGcNV1pp:-KxpdObX1dJgGSU9NZ91:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L290)):*

I don't think there's a strong reason that these functions need to be closures - I think it's more clear to make them ordinary functions and put their definitions above this one.
</blockquote></details>

Done.

---

*[pkg/sql/scrub.go, line 397 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo73LL9QvC-AKuzn8C:-KxpeAXMStC3VAfeT6Xd:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L397)):*

Seems like you should probably store this when you create it rather than do a scary divide by 2 every time - it'll be clearer I think! Otherwise the reader of this code has to reason about why the row length is always even.
</blockquote></details>

Done.

---

*[pkg/sql/scrub.go, line 399 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7IX2PGSKA4B95Axl:-KxpeODf9WUqGm__U7RS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L399)):*

s/is/are/
</blockquote></details>

Done.

---

*[pkg/sql/scrub.go, line 401 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7VasvJJldWZPHnVq:-KxpfcChoLAK1jwmssb6:b-vw6wgk) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L401)):*

It would be great to include the sentiment of this review comment in the code somewhere. Perhaps either near the query interpolation itself or right here where we're interpreting the results.
</blockquote></details>

Yep. Added it above the query interpolation function.

---

*[pkg/sql/scrub.go, line 403 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo7cyA25q2vCCI5Uuc:-KxrpRIQRTGDxoLPBMxb:bljq70o) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L403)):*

The specific column doesn't matter, but it comes from the primary index will only be null if the antijoin has only put through the secondary index side in this case.

I agree a comment is warranted.
(For completeness, it's the first column defined in the secondary index iirc)
</blockquote></details>

Done. (A comment that references the specifics was added.)

---

*[pkg/sql/scrub.go, line 407 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxo9p-Lkp4HRPz0oRrK:-KxpwRnifzF1TWzIF4hR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1f003a381d74d705269dd067c2c73474fbd60943/pkg/sql/scrub.go#L407)):*

more go stylish to say `var primaryKeyDatums parser.Datums`
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/scrub, line 22 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxpw_715Vrhg01ySt35:-Kxpw_715Vrhg01ySt36:b-e0ccwk) ([raw file](https://github.com/cockroachdb/cockroach/blob/56c38ba5f5fe2bb00338649c991cab8ae27f8d8c/pkg/sql/logictest/testdata/logic_test/scrub#L22)):*
> ```
> 
> query TTTTTTTT
> EXPERIMENTAL SCRUB TABLE t WITH OPTIONS INDEX(name_idx)
> ```

Hmm so something in the distSQL execution engine is causing a panic when there is no row data. I have a gut feeling it is because of the plan is unconventionally created and initialized or possibly there are no spans being inserted into the plan so it should not execute.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

lego:
Review status: 0 of 10 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[pkg/sql/logictest/testdata/logic_test/scrub, line 22 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxpw_715Vrhg01ySt35:-Kxs11gVX8-UJ8thDlY_:b-6aa74e) ([raw file](https://github.com/cockroachdb/cockroach/blob/56c38ba5f5fe2bb00338649c991cab8ae27f8d8c/pkg/sql/logictest/testdata/logic_test/scrub#L22)):*

Hmm so something in the distSQL execution engine is causing a panic when there is no row data. I have a gut feeling it is because of the plan is unconventionally created and initialized or possibly there are no spans being inserted into the plan so it should not execute.
</blockquote></details>

So, turns out when we call `makeDistSQLReceiver` we need to pass in the `DistSender.RangeDescriptorCache` and `DistSender.LeaseHolderCache`, and also probably a proper callback for `updateClock`. The two caches will need to be plumbed from the executor to the planner. There was also a TODO to fix this for `SchemaChanger`, but that one probably just barely avoids the panic codepath. https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/backfill.go#L539-L551

We may also want to plumb through the clock to update it, although I'm not completely sure what this does. I'll prop up another PR to do this which will also deal with the `SchemaChanger` codepath.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

knz:
I'll hold off reviewing and approving this until #19583 is cleaned up and resolved.

---

Reviewed 2 of 9 files at r7, 1 of 2 files at r9.
Review status: 1 of 10 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-:-KxsBJzGMT9MiNaF4SIs:bn7duid)*

lego:
This PR has been substantially cleaned up. I also added a few more test cases. :) PTAL

(also oof, I realize reviewable has a ton of revisions because of the past back and forth, my bad.)

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions.

---

*[pkg/sql/scrub.go, line 400 at r13](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxtJ8JgBDvLXEKeKdnM:-KxtJ8JgBDvLXEKeKdnN:b3om5bd) ([raw file](https://github.com/cockroachdb/cockroach/blob/38f5f32052c9cc9f457d95bcc1fe6d3e2e79aeec/pkg/sql/scrub.go#L400)):*
> ```Go
> // TODO(joey): This can be simplified with
> // TableDescriptor.FindIndexByName(), but this will only report the
> // first invalid index.
> ```

As noted in the TODO, is this a fair trade-off? (probably, it will dramatically reduce the function complexity). The only downside is reporting the first index that does not exist, which is pretty standard for other statements with multiple indexes for input.

---

*[pkg/sql/scrub_test.go, line 73 at r13](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxtJSrn0tvFe0vGtxVn:-KxtJSrn0tvFe0vGtxVo:b6vyncm) ([raw file](https://github.com/cockroachdb/cockroach/blob/38f5f32052c9cc9f457d95bcc1fe6d3e2e79aeec/pkg/sql/scrub_test.go#L73)):*
> ```Go
> 	return results, nil
> }
> ```

Possibly a few extra test cases would include when some columns are NULL in an index and/or unique index. At the moment though, there is no difference in our key encoding for these cases.

---

*[pkg/sql/logictest/testdata/logic_test/scrub, line 22 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Kxpw_715Vrhg01ySt35:-Kxs6CHkDuwdWTs27_eK:bvza2tv) ([raw file](https://github.com/cockroachdb/cockroach/blob/56c38ba5f5fe2bb00338649c991cab8ae27f8d8c/pkg/sql/logictest/testdata/logic_test/scrub#L22)):*

So, turns out when we call `makeDistSQLReceiver` we need to pass in the `DistSender.RangeDescriptorCache` and `DistSender.LeaseHolderCache`, and also probably a proper callback for `updateClock`. The two caches will need to be plumbed from the executor to the planner. There was also a TODO to fix this for `SchemaChanger`, but that one probably just barely avoids the panic codepath. https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/backfill.go#L539-L551

We may also want to plumb through the clock to update it, although I'm not completely sure what this does. I'll prop up another PR to do this which will also deal with the `SchemaChanger` codepath.
</blockquote></details>

:pray: `p.ExecCfg()` is a thing. I fixed up the schema changer one in a separate PR already, and was able to fix this one.

---

*[pkg/sql/logictest/testdata/logic_test/scrub, line 15 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxtH4mxvVYOywVWQJPq:-KxtH4mxvVYOywVWQJPr:b-txggrd) ([raw file](https://github.com/cockroachdb/cockroach/blob/89132b3e2b984bd7954404528f460b800500e6a6/pkg/sql/logictest/testdata/logic_test/scrub#L15)):*
> ```
> 
> query TTTTTTTT
> EXPERIMENTAL SCRUB TABLE t WITH OPTIONS INDEX
> ```

This test case that is currently failing. Specifically, during the `SCRUB TABLE` call we cannot find the index with the name `"and"`. Because the token is a `name`, the string literal includes quotations while the actual index name doesn't (it only has quotations added during `Format` output)

```sql
# Test that table, index, and column names are properly escaped during SCRUB's query generation.

statement ok
CREATE TABLE "all" (id int PRIMARY KEY, "any" string, INDEX "and" ("any" ASC))

query TTTTTTTT
EXPERIMENTAL SCRUB TABLE "all" WITH OPTIONS INDEX("and")
-----
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-:-KxtIMQC25lnOF9kxXxV:b-3mgeuy)*

knz:
Wow the results is quite fine. I find your tests extremely readable and easy to follow. The check code is slightly more complex but the commenting is fully appropriate. A few nits and I'm good on this.

---

Reviewed 2 of 9 files at r13, 1 of 1 files at r14.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[pkg/sql/scrub.go, line 290 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-KxPgSCxCCUqxUzraRgq:-Ky1kwCZYwD2A7c6H3Pc:b-mzxy1y) ([raw file](https://github.com/cockroachdb/cockroach/blob/5b4ee98d921697fbf71a29e3a0db80de6b23aca4/pkg/sql/scrub.go#L290)):*

I was a bit uncertain on the use of `rows.Close()`. Does it need only need to be closed if we receive rows? i.e. only in `Next`
</blockquote></details>

`Close()` needs to be called any time err is not nil, regardless of whether and how much you read from it.

---

*[pkg/sql/scrub.go, line 113 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1l6k1Uvv-eJScKzRm:-Ky1l6k1Uvv-eJScKzRn:b-bswyp8) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L113)):*
> ```Go
> 	}
> 
> 	// Process SCRUB options
> ```

period missing in comment.

---

*[pkg/sql/scrub.go, line 197 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1lM2rwvEHWJHSslbW:-Ky1lM2sGDKBgk9FsIrn:beg08je) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L197)):*
> ```Go
> 		// FIXME(joey): This is a bit of defensive programming, in the event
> 		// the secondary index descriptor doesn't contain all primary index
> 		// columns. Is this really necessary?
> ```

You're programming a scrub function. Defensive programming is appropriate here. 

---

*[pkg/sql/scrub.go, line 237 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1mHRfEHqqCjXIGlsG:-Ky1mHRfEHqqCjXIGlsH:b-ov7mcw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L237)):*
> ```Go
> 	// - Closed if an error is received, and we don't expect rows, or use
> 	//   the container.
> 	rows := sqlbase.NewRowContainer(*p.evalCtx.ActiveMemAcc, ci, 0)
> ```

I think you can do `defer rows.Close(ctx)` here and simplify the code below.

---

*[pkg/sql/scrub.go, line 288 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1ln38t0RYuDFQoFhn:-Ky1ln38t0RYuDFQoFho:b-lqtgny) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L288)):*
> ```Go
> 			buffer.WriteString(" AND ")
> 		}
> 		buffer.WriteString(fmt.Sprintf("%[1]s.%[2]s IS NULL", tableName, col))
> ```

Wow, no. `fmt.Fprintf(&buffer, "...", ...)` here and below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-:-Ky1mPNtdSMNaBhDwIhK:bnf8q1b)*

lego:
Review status: 2 of 3 files reviewed at latest revision, 19 unresolved discussions, some commit checks pending.

---

*[pkg/sql/scrub.go, line 113 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1l6k1Uvv-eJScKzRm:-KyGkQrJh4w32PDFz_Jq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L113)):*

period missing in comment.
</blockquote></details>

Done.

---

*[pkg/sql/scrub.go, line 237 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1mHRfEHqqCjXIGlsG:-KyGjb68WQRCtGSyht8y:b-xvbmeo) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L237)):*

I think you can do `defer rows.Close(ctx)` here and simplify the code below.
</blockquote></details>

Kinda. The issue is the case when we don't have an error and have rows we don't want to close it. I moved this out into a function so that there could be a single `if err != nil { rows.Close() }`

---

*[pkg/sql/scrub.go, line 288 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1ln38t0RYuDFQoFhn:-KyGkGTATar__8iJpw5e:bx0x8kz) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L288)):*

Wow, no. `fmt.Fprintf(&buffer, "...", ...)` here and below.
</blockquote></details>

Thanks for pointing that out. Done!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

knz:
Reviewed 1 of 1 files at r15.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks pending.

---

*[pkg/sql/scrub.go, line 237 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1mHRfEHqqCjXIGlsG:-KyGvPl7ZPTgpHxs0EzP:b-v6rzhl) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L237)):*

Kinda. The issue is the case when we don't have an error and have rows we don't want to close it. I moved this out into a function so that there could be a single `if err != nil { rows.Close() }`
</blockquote></details>

Nice! Thanks!

---

*[pkg/sql/scrub.go, line 288 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1ln38t0RYuDFQoFhn:-KyGvRX0ImhMkxKg-pbO:b-5ayn2v) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L288)):*

Thanks for pointing that out. Done!
</blockquote></details>

Just to clarify, this was a "wow" of surprise, not disapproval.

Thanks :beers: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586)*

lego:
TFTR!

---

Review status: all files reviewed at latest revision, 17 unresolved discussions, all commit checks successful.

---

*[pkg/sql/scrub.go, line 288 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-Ky1ln38t0RYuDFQoFhn:-KyH2OwcnLoCCWHUZi_c:b32vhw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5c70876a1b97be75da65131bab6bfca46cdb6fca/pkg/sql/scrub.go#L288)):*

Just to clarify, this was a "wow" of surprise, not disapproval.

Thanks :beers: 
</blockquote></details>

:)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19586#-:-KyH2ZYB2yMulsd0C_IG:b-814g8k)*

PR #19589, opened 2017-10-26
sql/parser: don't wrap unrestricted names in quotes

benesch:
Previously, a statement using the "@" form of specifying an index, like

    ALTER INDEX db.tbl@primary RENAME TO like

would roundtrip with "primary" and "like" wrapped in quotes

    ALTER INDEX db.tbl@"primary" RENAME TO "like"

because they're reserved keywords. This is particularly unfortunate
because "primary" is a built-in index name that cannot be changed. Teach
the parser to preserve the distinction between names and unrestricted
names after parsing, so that unrestricted names can be printed without
quotes even when they match a reserved keyword.

This commit introduces a new type, UnrestrictedName. In general,
unrestricted_name nonterminals are parsed into UnrestrictedNames, while
name nonterminals are parsed into Names. Upon formatting, Names are
wrapped in quotes if they match a reserved keyword, while
UnrestrictedNames are not.

A few unrestricted_name nonterminals continue to be parsed as Names
instead of UnrestrictedNames where the work involved was prohibitive;
these cases are no worse off than they were before.

Changes outside the parser package were required wherever code referred
to index names or zone names. Previously, index names and zone names
were `parser.Name`s, but are now `parser.UnrestrictedName`s due to where
they appear in the grammar.

This is pre-emptive polish for upcoming work to support per-index zone
configs. The CLI uses the SQL parser's logic for escaping zone
specifiers (e.g., "db.tbl@primary"); this change ensures that
`./cockroach zone get db.tbl@primary` won't print the zone's name as
`db.tbl@"primary"`.


benesch:
This actually passes tests now, so RFAL :)

justinj:
Hope you don't mind me jumping in uninvited - :lgtm: with one comment

---

Review status: 0 of 19 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/sql/parser/format.go, line 46 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-Kxkl8J_j1B7GwX5bOKP:-Kxkl8J_j1B7GwX5bOKQ:b-1u0uyd) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/format.go#L46)):*
> ```Go
> 	// If true, keywords used as identifiers will be rendered without wrapping
> 	// quotes.
> 	bareKeywords bool
> ```

I don't think this is ever read from, unless I missed somewhere? We might be able to either remove the boolean parameter in other places or remove this (would be nice if we could remove the boolean parameter)!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-:-KxklYekmhNWPccRp383:bddc0fd)*

jordanlewis:
:lgtm: with a few preferences for improved factorization

---

Reviewed 19 of 19 files at r1.

---

*[pkg/sql/parser/name_part.go, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnP-HK8FzgAqEwStsM:-KxnP-HK8FzgAqEwStsN:bproo9p) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/name_part.go#L83)):*
> ```Go
> }
> 
> // An UnrestrictedName is like Name, but it does not need to be wrapped in
> ```

This doesn't pass the linter, does it? I thought godocs had to have the name of the exported type as the first word in the comment?

Also I think it would be useful to explain why it doesn't need to be wrapped in quotes - I think the comment is a missing a little bit of context.

---

*[pkg/sql/parser/name_part.go, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnPR9yXV1yFbxMUQ0e:-KxnPR9yXV1yFbxMUQ0f:b-xra9de) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/name_part.go#L88)):*
> ```Go
> 
> func (u UnrestrictedName) Format(buf *bytes.Buffer, f FmtFlags) {
> ```

Can you please factor out the commonalities from this method and Name.Format? It seems like it would be easy to change one and not the other, especially with the special anonymize logic.

---

*[pkg/sql/parser/scan.go, line 895 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnPm3lqeJo4B340Z2R:-KxnPm3lqeJo4B340Z2S:bkq1al4) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/scan.go#L895)):*
> ```Go
> // identifier. If allowBareKeywords is false, additionally verify that the input
> // string is not a SQL keyword.
> func isBareIdentifier(s string, allowBareKeywords bool) bool {
> ```

Rather than renaming this and adding a boolean parameter, I'd prefer to see it split into 2 functions that share some code. I think `isNonKeywordBareIdentifier` is more clear than `isBareIdentifier(false)`.

---

*[pkg/sql/parser/sql.y, line 1090 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnNKCesp8_IDA-6WLL:-KxnNKCesp8_IDA-6WLM:b-f45gcv) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/sql.y#L1090)):*
> ```Yacc
>     /* SKIP DOC */
>     $$.val = &SetZoneConfig{
>       ZoneSpecifier: ZoneSpecifier{NamedZone: UnrestrictedName($3)},
> ```

Could you please add an inline comment above `unrestricted_name` while you're in here to just clarify the meaning of the rule? I assume it just means "a name in a position in the syntax tree that doesn't conflict with any SQL keywords" but I'm not sure and it would be nice to see that clearly stated when someone is trying to understand how the unrestricted name logic works.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-:-KxnQ1Hn0BxGp5WGe5kp:b56cn82)*

knz:
Thanks for approaching this!

I think I'm ok with the change however, as usual, I would be more comfortable if the PR would also state in prose what it intends to do next to providing the actual code. This way, your reviewer can compare your intent with your realization, and assess whether they match. Also it would be good to capture the reasoning for use by future readers of this area of the code.

In particular:
The commit message does not sufficiently explain what is being done here.
It should explain more that there is a new thing called `UnrestrictedName` and what that thing stands for, and what the difference is with `Name`.
Also the commit message should outline why many changes were needed outside of the `parser` package.
Finally, it should also teach the reader how to decide whether to use one or the other in the future, especially for any use outside of the `parser` package.

---

Reviewed 9 of 19 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-:-Kxrf8cQfE3R1Oz1MYot:b-jamt7b)*

benesch:
Ok, added a lot of detail to the commit message and code comments, and made most of the factorization changes you requested @jordanlewis. See below for details.

---

Review status: 3 of 18 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/parser/name_part.go, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnP-HK8FzgAqEwStsM:-Ky2jbEkmp_1wMvU39Qa:bbv2teq) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/name_part.go#L83)):*

This doesn't pass the linter, does it? I thought godocs had to have the name of the exported type as the first word in the comment?

Also I think it would be useful to explain why it doesn't need to be wrapped in quotes - I think the comment is a missing a little bit of context.
</blockquote></details>

It's ok if the comment starts with "A" or "An".

Added some detail here.

---

*[pkg/sql/parser/name_part.go, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnPR9yXV1yFbxMUQ0e:-Ky2oxLAaQjNPQLhn5ku:bxwu2ur) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/name_part.go#L88)):*

Can you please factor out the commonalities from this method and Name.Format? It seems like it would be easy to change one and not the other, especially with the special anonymize logic.
</blockquote></details>

Factoring out the commonalities and avoiding boolean parameters are mutually exclusive, I think. What do you suggest? I decided to prioritize avoiding boolean parameters over factoring out the one shared conditional, but happy to go the other way.

---

*[pkg/sql/parser/scan.go, line 895 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnPm3lqeJo4B340Z2R:-Ky2o5gv-9vbCqY8sOG-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/scan.go#L895)):*

Rather than renaming this and adding a boolean parameter, I'd prefer to see it split into 2 functions that share some code. I think `isNonKeywordBareIdentifier` is more clear than `isBareIdentifier(false)`.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 1090 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnNKCesp8_IDA-6WLL:-Ky2r4NucZtIxii4efVi:b-402m97) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/sql.y#L1090)):*

Could you please add an inline comment above `unrestricted_name` while you're in here to just clarify the meaning of the rule? I assume it just means "a name in a position in the syntax tree that doesn't conflict with any SQL keywords" but I'm not sure and it would be nice to see that clearly stated when someone is trying to understand how the unrestricted name logic works.
</blockquote></details>

I updated the comment on `unrestricted_name`, but the discussion that takes place starting at `Name classification hierarchy` is pretty comprehensive.

---

*[pkg/sql/parser/format.go, line 46 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-Kxkl8J_j1B7GwX5bOKP:-Ky2jUHBGrORtxlQQNkL:b-1xvpxi) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/format.go#L46)):*

I don't think this is ever read from, unless I missed somewhere? We might be able to either remove the boolean parameter in other places or remove this (would be nice if we could remove the boolean parameter)!
</blockquote></details>

Whoops, my bad. This was my first attempt at making this change, and I forgot to remove it. Sorry for the sloppiness.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-:-Ky6psg5NYPKivCfrYeR:bhlpygi)*

knz:
:lgtm:

---

Reviewed 2 of 19 files at r1, 16 of 16 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-:-KyG3DOFqdMo3IxV1nkI:bnfp4nl)*

jordanlewis:
Reviewed 16 of 16 files at r2.

---

*[pkg/sql/parser/name_part.go, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589#-KxnPR9yXV1yFbxMUQ0e:-KyIcDlF5uE8jb4BRjnx:b-1ns6gm) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0fdf9b27141ac0fd5ced3aaadac9186b3d556af/pkg/sql/parser/name_part.go#L88)):*

Factoring out the commonalities and avoiding boolean parameters are mutually exclusive, I think. What do you suggest? I decided to prioritize avoiding boolean parameters over factoring out the one shared conditional, but happy to go the other way.
</blockquote></details>

Heh, good point. I like what you have here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19589)*

benesch:
Pleasure doing business with y'all. Thanks for the reviews!

PR #19592, opened 2017-10-27
sql: introduce MakeJSON

justinj:
It occurred to me today that people using JSON for various purposes
within Cockroach will need a way to construct JSON objects, and it will
be a pain to construct the structs literally because 1. the object keys
need to be sorted, and 2. we want to be able to change the
representation (at least for the time being).

This commit adjusts interpretDJSON to rename it to MakeDJSON, export it
from the parser package, have a more usable API, and include tests.

### DIFF
--- a/pkg/sql/parser/datum.go
+++ b/pkg/sql/parser/datum.go
@@ -1953,6 +1953,22 @@ func ParseDJSON(s string) (Datum, error) {
 	return &DJSON{j}, nil
 }
 
+// MakeDJSON returns a JSON value given a Go-style representation of JSON.
+// * JSON null is Go `nil`,
+// * JSON true is Go `true`,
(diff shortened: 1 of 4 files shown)


lgo (approved):
💯.

You mentioned that the interface may change. Is there anything we may want to be aware of to make future changes to the interface painless? (seems simple enough though, and worst case I imagine a separate public constructor could be made)

justinj:
I don't expect this one to change too much/at all, it seems like the most convenient possible interface, though I could see maybe having one that would allow for more efficient construction in certain cases where it might matter (so you don't need to build up the whole `map[string]interface{}` structure before constructing your `DJSON`) but I think this one should exist regardless.

lego:
Completely agree. This made my day 🙌

PR #19593, opened 2017-10-27
 storage: minor tweaks to temporary disabling of incoming preemptive snapshots

a-robinson:
Also do a minor rename to make the variable/log message clearer. Intended to help debug #19010. The failures I'm seeing on my gceworker are very different from what teamcity has been running into lately, so try to get a little more info out of the teamcity runs.

The second commit *might* be a fix for #19010, but it's hard to say without being able to reproduce it locally.

The third commit is only tangentially related in that I ran into it while debugging the same problem. It just makes certain log messages more understandable without needing to refer back to the source code.

### DIFF
--- a/pkg/storage/client_test.go
+++ b/pkg/storage/client_test.go
@@ -918,8 +918,8 @@ func (m *multiTestContext) restartStoreWithoutHeartbeat(i int) {
 
 	// Normally, the newly restarted store would not be able to accept rebalances
 	// until the election timeout passed. But we're using a manual clock which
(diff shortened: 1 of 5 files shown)


tschottdorf:
:lgtm:

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19593#-:-KxTZi5375OxjMK5_Ued:bnfp4nl)*

PR #19594, opened 2017-10-27
sql: add some JSONB operators

justinj:
This adds support for the following JSONB operators:

* `->`
* `->>`
* `#>`
* `#>>`
* `-`
* `?`
* `?|`
* `?&`

It's worth noting some of these will need to change once the encoding happens.


knz:
:LGTM: however:

- Please extend the commit message to briefly remind the reader what the operators mean (e.g. give their name).
- outline which are those sensitive to the encoding.

---

Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19594#-:-KxmDUgA0Ah6YIwur7bz:bzgdzd8)*

justinj:
TFTR! I rebased this over to use the JSON interface (the implementation is basically the same, though) instead and updated the commit message.

PR #19596, opened 2017-10-27
WIP: rocksdb-level encryption with support for multiple keys.

mberhault:
**DO NOT MERGE** or use, this is definitely not ready.

This adds support for encryption at the rocksdb level.
We make use of the newish `env_encryption` which is a wrapper around all
file operations within rocksdb.

We reuse the technique from `rocksdb::CTREncryptionProvider` which adds
a page-sized block at the beginning of every file. This file contained
the IV and counter, but we now include a key ID.

Multiple keys can be passed (right now, hardcoded, need some way of
passing them in), with the highest ID used for new files.
We hook this into AES-128 as the block cipher.

We can include a special meaning for keyID 0 meaning no encryption,
this will fit well with this code and allow for transition from
non-encrypted to encrypted and back. However, that will not allow for
transitions from rocksdb files without the page-sized prefix. We may
want to have a sentinel file somewhere to indicate that we have them.
The absence of the sentinel would tell us to error out on existing
(older format) databases.

Lots of caveats in the current implementation, these are just the ones
that will remain assuming this PR is cleaned up and properly tested:
* no public use of rocksdb `env_encryption` and pretty much no tests,
  should be considered highly experimental
* this break compatibility with existing rocksdb files
* we would need a way to tell when a key is no longer needed
* is cryptopp the right choice?
* how do we make sure we use the AES instruction set? Will this cause
  problems?


mberhault:
Just a reminder in case the title and comment are not clear: this PR just plays around with rocksdb encryption. You can tell by the fact that there are no tests and two hard-coded keys.

petermattis:
This generally looks like what I was expecting. The ability to support multiple keys is nice. We'll eventually want to be able to determine how many files (or how much data) is encrypted with a particular key. Have you thought about how that could be done? RocksDB provides routines for listing all of the sstables and WAL files, so perhaps this is iteration over all of those files and reading the first page. 

> * this break compatibility with existing rocksdb files

We store a `COCKROACHDB_VERSION` file in the store directory. I think we could support existing RocksDB instances by checking the version and not using the `env_encryption` wrapper. Enabling encryption would then involve decommissioning a node, wiping it and recreating it. Somewhat onerous, but I think appropriate for the scope we're targeting initially. You'd want to add some checks if someone tried to enable encryption on a store that didn't support it.

---

Review status: 0 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[c-deps/libroach/db.cc, line 2155 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxTck7PqFu46zN388uC:-KxTck7PqFu46zN388uD:b-jnih9l) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/db.cc#L2155)):*
> ```cpp
>   rocksdb::ReadOptions opts;
>   opts.total_order_seek = true;
>   /*opts.table_filter = [min, max](const rocksdb::TableProperties& props) {
> ```

What is this about? Another caveat?

---

*[c-deps/libroach/encryption.cc, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxTcUY94PFa5J4FLbyC:-KxTcUY94PFa5J4FLbyD:bounfd9) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.cc#L1)):*
> ```cpp
> // Copyright 2014 The Cockroach Authors.
> ```

`s/2014/2017/g`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-:-KxTb_Bs4wioj8MXVxj2:bl0vwiz)*

mberhault:
Review status: 0 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[c-deps/libroach/db.cc, line 2155 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxTck7PqFu46zN388uC:-KxTf58Me9Ehx_Sdx6Jt:b-j1jshs) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/db.cc#L2155)):*

What is this about? Another caveat?
</blockquote></details>

Nah, I just didn't bother including @benesch's patch. It made it into rocksdb 5.8 though. Even if we don't want to go that high, we can just apply the patch on top of 5.7.

---

*[c-deps/libroach/encryption.cc, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxTcUY94PFa5J4FLbyC:-KxTfFIt2dJzmBG-Vaho:b-xmdlg) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.cc#L1)):*

`s/2014/2017/g`
</blockquote></details>

yeah, let's not open this for review quite yet, there are obviously a lot of issues.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596)*

bdarnell:
> * is cryptopp the right choice?

I think it's a reasonable choice. Our current needs are so simple (just basic AES) that there's very little difference between the different libraries. If we start doing something more complex I would look at nacl/libsodium. 

> * how do we make sure we use the AES instruction set? Will this cause
>   problems?

This is one area where AES implementations can differ - we need one that supports both hardware-accelerated AES and software fallbacks, and to choose between them at runtime based on CPU feature detection (as opposed to a static `ifdef`). I think cryptopp does the right thing here but we'll need to test. 

---

Review status: 0 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[.gitmodules, line 19 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZWdLdyQqA2Sip5l5F:-KxZWdLdyQqA2Sip5l5G:b8xczwd) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/.gitmodules#L19)):*
> ```
> [submodule "c-deps/cryptopp"]
> 	path = c-deps/cryptopp
> 	url = https://github.com/weidai11/cryptopp.git
> ```

Before this is finalized, be sure to fork the library we use under the `cockroachdb` namespace. 

---

*[c-deps/libroach/encryption.cc, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZZCuhYZtI7BaDOp4L:-KxZZCuhYZtI7BaDOp4M:b5o6tlx) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.cc#L25)):*
> ```cpp
> #include <hex.h>
> 
> byte key[CryptoPP::AES::DEFAULT_KEYLENGTH];
> ```

What's this global?

---

*[c-deps/libroach/encryption.cc, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZ_rF-bC_dg4BjKI8X:-KxZ_rF-bC_dg4BjKI8Y:b8bvvst) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.cc#L32)):*
> ```cpp
> 
> rocksdb::Status AESCipher::Encrypt(char *data) {
> 	CryptoPP::ECB_Mode<CryptoPP::AES>::Encryption enc((byte *)key, CryptoPP::AES::DEFAULT_KEYLENGTH);
> ```

I believe constructing the Encryption and Decryption objects has a non-trivial cost so we should do it in the constructor and reuse instead of redoing it on every block. 

---

*[c-deps/libroach/encryption.h, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZOth2VYEQIdDlEhur:-KxZOth2VYEQIdDlEhus:bvk0a3b) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.h#L25)):*
> ```objc
> rocksdb::Env* GetEncryptedEnv(rocksdb::Env* base_env);
> 
> // Implements a BlockCipher using AES-128.
> ```

I think we want AES-256 (we should at least offer that as an option, but I'd use 256 as the default)

---

*[c-deps/libroach/encryption.h, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZZMDLsxxr5lEtZ7BL:-KxZZ1kEjs8s0A56F8Ca:b-lxmwk5) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.h#L31)):*
> ```objc
>       size_t blockSize_;
>     public:
>       AESCipher(char *key) : key_(key) {}
> ```

AES supports different key lengths, so this should be a string or slice instead of a raw char pointer. 

---

*[c-deps/libroach/encryption.h, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZOmFBY_Qw0zCzmadR:-KxZOmFBY_Qw0zCzmadS:b-t59ml6) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.h#L34)):*
> ```objc
>       virtual ~AESCipher() {}
> 
> 			char *Key() { return key_; }
> ```

This file has a mix of tabs and spaces; all non-go code in cockroach should use spaces instead of tabs. 

---

*[c-deps/libroach/encryption.h, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-KxZUBaLCe5IB2AbCnDQ:-KxZUBaLCe5IB2AbCnDR:bhpmaco) ([raw file](https://github.com/cockroachdb/cockroach/blob/4bd569aaa21778bdd995d8aa57f5d458fc567bd1/c-deps/libroach/encryption.h#L50)):*
> ```objc
> typedef std::map<uint64_t, rocksdb::BlockCipher*> CipherList;
> 
> class CTREncryptionProviderWithKey : public rocksdb::EncryptionProvider {
> ```

Key rotation is such a fundamental need that I'm surprised we need to do this instead of using `CTREncryptionProvider`. Either we're missing something about how to use this API or it's poorly designed. 

It looks like one way to make the library do what we want is to subclass CTREncryptionProvider, pass a no-op cipher to the constructor, store the key ID in PopulateSecretPrefixPart, then use it in CreateCipherStreamFromPrefix. But that sounds pretty far from the way this interface is presumably intended to be used (on the other hand, what else is the "secret prefix" for if not for per-file keys? I just don't understand the top-level cipher's role)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-:-KxZWoRe0ypAqiOvYKCe:b9be8j0)*

mberhault:
I'll avoid addressing all the review comments, this was more of a playing around thing. I do agree with all your comments though and they will be addressed in the real PR.

About the main API: it seems to have been added for Arangodb's encryption at rest which does not support key rotation (https://docs.arangodb.com/devel/Manual/Administration/Encryption/). It's only in their enterprise version though, so no sample uses but I suspect they just do a straight up `AES(passed-in-key)`.

As for using the existing implementation, but we would have too many things to change. I also want to add support for "encryption flag set on this drive, but no key specified" which will still write the prefix (eg: keyid=0) but bypass everything else and just write the blocks as passed in.
At that point we'll just be overriding everything so might as well skip it.

---

Review status: 0 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19596#-:-KxTfKfRYFpcYo0zfkQD:b-82odl6)*

mberhault:
tweaked a bit: spaces vs tabs, copyright date, and global key (oops).
This will now wait until discussion and a likely RFC.

Issue #19559, opened 2017-10-26
stability: crash on startup.


mberhault:
`cyan` (continuous deployment) was just upgraded from 15b18a4 to 0224ad4.
Two nodes crashed on startup: `cockroach@cockroach-cyan-0002.crdb.io` and `cockroach@cockroach-cyan-0003.crdb.io`.
There are 6 merges between those two PRs, but all 6 failed to build so were never pushed.

```
I171026 14:02:38.790478 21 storage/engine/rocksdb.go:411  opening rocksdb instance at "/mnt/data1/cockroach-temp158446832"
W171026 14:02:38.849368 21 gossip/gossip.go:1277  [n?] no incoming or outgoing connections
I171026 14:02:38.849467 21 storage/engine/rocksdb.go:411  opening rocksdb instance at "/mnt/data1"
I171026 14:02:38.904256 51 gossip/client.go:129  [n?] started gossip client to 192.168.1.4:26257
I171026 14:02:40.951483 21 server/config.go:512  [n?] 1 storage engine initialized
I171026 14:02:40.951536 21 server/config.go:515  [n?] RocksDB cache size: 3.4 GiB
I171026 14:02:40.951555 21 server/config.go:515  [n?] store 0: RocksDB, max size 0 B, max open file limit 60536
F171026 14:02:41.398505 600 util/log/log.go:39  unexpected panic: runtime error: invalid memory address or nil pointer dereference
goroutine 600 [running]:
github.com/cockroachdb/cockroach/pkg/util/log.getStacks(0x344bb01, 0x14f123a994c97fc2, 0x2c40c78, 0xf)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/clog.go:833 +0xcf
github.com/cockroachdb/cockroach/pkg/util/log.(*loggingT).outputLogEntry(0x344bae0, 0xc400000004, 0x2c40c78, 0xf, 0x27, 0xc4208c1860, 0x52)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/clog.go:738 +0x5fc
github.com/cockroachdb/cockroach/pkg/util/log.addStructured(0x2f54d40, 0xc42001c018, 0x4, 0x2, 0x1f6c2a2, 0x14, 0xc420da58d8, 0x1, 0x1)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/structured.go:153 +0x2fc
[...]

PR #19591, opened 2017-10-26
acceptance: test rapid restarts and early queries

tbg:
See #19559, #19579, #19571 for issues touched by this test (all of which
it reproduced before the preceding commits fixed them).

The changes here (outside of the test) ensure that a single-node instance starts
as quickly as possible, without waiting for ports to be ready or for replication
to have occurred. This was done to remove any additional delay in
`TestRapidRestarts` that could otherwise obscur bugs (for example, consider that
going through `./cockroach quit` was not able to tickle any of the bugs).

I was also tempted to "bump" the version upgrade tests to use 1.1 instead of 1.0
since that would have removed a bunch of cruft (since 1.0 doesn't write its
listener files), but I decided against that. As a result, some of the new
accessors look a little unclean, but they should straighten themselves out once
we only rely on listener files.

PR #19598, opened 2017-10-27
server: open most HTTP endpoints later

tbg:
Extracted from https://github.com/cockroachdb/cockroach/pull/19591
which also adds a regression test. The tests in that PR pass, but
to make sure I verified manually using

```
while true; do curl -O /dev/null http://localhost:8080/_status/vars; done
```

and `./cockroach start --insecure` in another window, ten times.

Fixes https://github.com/cockroachdb/cockroach/issues/19559.


mberhault (approved):

a-robinson (approved):
LGTM, as in #19591 .

Although you referred to "fixing" #19493 again even though it's a merged PR. Is there an issue you meant to refer to instead?

tschottdorf:
Thanks! I meant #19559. Edited the commit and PR messages.

Issue #19600, opened 2017-10-27
sqlccl: `BACKUP... WITH revision_histroy` missing dropped tables


dt:
Currently the spans that must be included in a backup are generated based on the tables that match as of the end time, but that means that a table dropped since the previous backup will not be included and thus its spans will not have their history, up to the point the table was dropped, captured.

dt:
I want to gracefully handle changing spans soon, but in the meantime, I'm thinking of requiring the spans on revision_history backups exactly match the previous backup to prevent creating a broken backup. This breaks some legit use cases, e.g.
`BACKUP <everything> TO foo` followed by `BACKUP <small> TO bar INCREMENTAL FROM foo WITH revision_history`, but it makes sure you can't miss the fact that that your incremental with revision history is missing the changes up to the point the table was dropped.

dt:
After discussing some fringe / edge case user stories with @dianasaur323, our plan here, for now, is to include a descriptor if it:
* matches the target spec, as of the BACKUP time (stmt time / explicit `AS OF SYSTEM TIME`), i.e. our current matching behavior.
* at any point in the time window, has a parent ID matching a DB that is expanded by the target spec (`DATABASE foo` or `foo.*`).

Concretely, this means that a backup made by running `BACKUP DATABASE foo WITH revision_history` should include the history for up until dropping a table that was in `foo`, or the history for a table moved into or out of `foo`, the history both before and after a `truncate`, etc.

However, some places where this could potentially behave in an unexpected way come up when a table had a different name at some point in that window but is named explicitly in the BACKUP command -- e.g. `BACKUP foo.bar`. For example, if someone runs renames a table (ID 51)  `bar` to `baz` and then creates a new table (ID 52) `bar` before running `BACKUP foo.bar WITH revision_history`, while obviously we should include ID 52, we struggled a bit with how to handle ID 51. 

We went back and forth a bit here, but concluded the approach which seemed the simplest to understand/explain is the rules above, which implies that 51 is _not_ included. Adding an additional rule such as "has the same name at some point as a descriptor that is matched" would include 51, but seems to complicate the story -- e.g.  you could already include it if you wanted to, by adding `baz` to the target spec, but, if we added that additional rule, you'd be out of luck you _didn't_ want to include `baz`?

[...]

PR #19601, opened 2017-10-27
sqlccl: fix RESTORE of INCREMENTAL after dropping a table

dt:
This adds tests for the DROP and TRUNCATE cases.

Also, extend the handling of dropped tables during restore:

If a table was dropped between t1 and t2, it isn’t actually required that a backup of the state at time t2 include the span for that table.

However our checking of the span coverage was requiring that all spans in the backup match the desired time bounds, even those that were not required by the requested restore. This changes that enforcement to be limited to required spans.

This error would only occur when attempting to RESTORE that backup or in later attempts to create additional incremental backups on top of it : the initial backup, that stops including the span, would succeed, as it validates its _previous_ backups.

NB: for ’revision_history’ backups, this is still broken — unlike a latest backup, which is free to not include the span of the dropped table, a ‘revision_history’ backup *does* need to include that span,
to allow time travel to before the DROP. Tracked in #19600.

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -1792,9 +1792,9 @@ func TestTimestampMismatch(t *testing.T) {
 		}
 
 		// Missing data for one table in the most recent backup.
-		_, err = sqlDB.DB.Exec(`RESTORE data.* FROM $1, $2`,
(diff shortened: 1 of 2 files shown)


mjibson:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19601#-:-Kxg4eW36HIezTxSxoGK:bnfp4nl)*

PR #19602, opened 2017-10-27
[DO NOT MERGE] storage: benchmark difference between purego and cgo iteration

asubiotto:
NOTE: All numbers collected on a one-node cluster running on my laptop.

While investigating `COUNT(*)` performance (https://github.com/cockroachdb/cockroach/issues/19288), the throughput of various layers was measured (fb05b62) while performing `SELECT COUNT(*) FROM tpch.lineitem`. The throughput found for the `rocksdb` layer (which is simple iteration over the keys/values in `tpch.lineitem`) reported `~95MB/s`. This was a long way off from the `~500MB/s` measured through [`BenchmarkMVCCComputeStats_RocksDB`](https://github.com/cockroachdb/cockroach/blob/03a829c5f069a75a351edef36fe255055e73a6b1/pkg/storage/engine/bench_rocksdb_test.go#L75). The difference between these two iterations is that for the first one, we cross the cgo boundary once for every key/value pair while we only cross the cgo boundary once while computing stats, as the iteration is done in C++.

This warranted further investigation, so we set about measuring the cgo overhead by ingesting a `tpch.lineitem` sst and reading it with a `leveldb` pure go iteration and our normal `rocksdb` iteration (included in this PR), and got a pretty noticeable difference (the size of a key/value pair is `~200` bytes):
```
$ benchstat cgo purego
name             old time/op    new time/op     delta
MVCCIteration-8     511ms ± 4%      139ms ± 2%   -72.84%  (p=0.000 n=19+18)

name             old speed      new speed       delta
MVCCIteration-8  47.8MB/s ± 4%  176.2MB/s ± 2%  +268.82%  (p=0.000 n=20+18)
```
Although the purego iteration doesn't hit the `500MB/s` number, the throughput really depends on the size of the keys so is not comparable (for example, here is a run with an sst file where each key/value was `~700 bytes`):
```
BenchmarkMVCCIteration/Cgo-4         30      38456904 ns/op    1093.95 MB/s
BenchmarkMVCCIteration/Purego-4      50      27323478 ns/op    1539.69 MB/s
```
By moving our iteration at the MVCCScan level into C++, we could reduce our cgo overhead for scans and increase the throughput by a significant amount.
@arjunravinarayan @petermattis @jordanlewis @bdarnell @nvanbenschoten

### DIFF
--- a/pkg/ccl/storageccl/engineccl/mvcc_test.go
+++ b/pkg/ccl/storageccl/engineccl/mvcc_test.go
@@ -12,6 +12,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


petermattis:
> By moving our iteration at the MVCCScan level into C++, we could reduce our cgo overhead for scans and increase the throughput by a significant amount.

I agree, though there are significant challenges to doing this due to the cgo calling constraints and the restriction that only  C code can be called from Go, not C++. 

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19602#-:-KxUaCoiJh16-nNHMgAk:b2l6csv)*

petermattis:
> I agree, though there are significant challenges to doing this due to the cgo calling constraints and the restriction that only  C code can be called from Go, not C++.

To elaborate: the challenge is in passing the results of the scan from C++ back through Cgo to Go. On possibility is that we return the results in RocksDB WriteBatch format. This is independent of changing ScanResponse to use WriteBatch format. The benefit of using WriteBatch is that it is a single chunk of memory and not a complex data structure. 

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19602#-:-KxUePfdU7o1VmUoUlj9:b9va3au)*

bdarnell:
We could also reduce cgo boundary crossing while keeping the current structure by prefetching multiple k/v pairs whenever we call into rocksdb. It may be tricky to take advantage of these prefetched keys since we do a lot of `Seek()` calls and not just `Next()`.

PR #19581, opened 2017-10-26
scripts: fix gceworker

rjnn:
bootstrap-debian.sh does not build cockroach out of the box without a
git submodule update --init. Fix the git clone to recursively clone
submodules so that make works after running bootstrap-debian.sh.

PR #19606, opened 2017-10-27
build: check out submodules before `go install`

benesch:
I accidentally broke this ordering in #19527: submodules must be checked
out before we run `go install`. Fixing this eliminates the need for a
recursive clone in bootstrap-debian, which was added by #19581 as a
workaround. (Avoiding the recursive clone in bootstrap-debian means
someone will realize if the Makefile breaks again.)


lgo (approved):
LGTM

benesch:
Thank you! 🙇

PR #19608, opened 2017-10-27
sql: add random JSON generator

justinj:
This can be the basis for an RSG test in the future, right now it's just
used for a round-tripping test.

Considered making this a separate package but opted against it, no real
basis to that decision though, open to hearing otherwise.

### DIFF
--- a/pkg/internal/rsg/rsg.go
+++ b/pkg/internal/rsg/rsg.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/sql/sem/types"
 	"github.com/cockroachdb/cockroach/pkg/util/duration"
 	"github.com/cockroachdb/cockroach/pkg/util/ipaddr"
+	"github.com/cockroachdb/cockroach/pkg/util/json"
 	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
 	"github.com/cockroachdb/cockroach/pkg/util/timeofday"
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
@@ -269,6 +270,14 @@ func (r *RSG) GenerateRandomArg(typ types.T) string {
 		types.AnyArray,
 		types.Any:
 		v = "NULL"
+	case types.JSON:
+		r.lock.Lock()
+		j, err := json.Random(20, r.src)
+		r.lock.Unlock()
+		if err != nil {
+			panic(err)
+		}
+		v = fmt.Sprintf(`'%s'`, tree.DJSON{JSON: j})
 	default:
 		// Check types that can't be compared using equality
 		switch types.UnwrapType(typ).(type) {
--- a/pkg/util/json/json_test.go
+++ b/pkg/util/json/json_test.go
@@ -17,10 +17,12 @@ package json
 import (
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"strings"
 	"testing"
 
 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
+	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 )
 
 func TestJSONOrdering(t *testing.T) {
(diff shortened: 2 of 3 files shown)


mjibson:
All of our other rand datum code is in the sqlbase package. Does it make sense to put this there? We should add it to sqlbase.RandDatum at least I think.

---

---

*[pkg/sql/parser/json_random.go, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-KxkxdP_4GG8DakeGX43:-KxkxdP_4GG8DakeGX44:blf963k) ([raw file](https://github.com/cockroachdb/cockroach/blob/0cc321137c83854606c1f4700ac3c8eff848ae49/pkg/sql/parser/json_random.go#L23)):*
> ```Go
> )
> 
> func randomDJSON(complexity int) DJSON {
> ```

I'd remove this function and call it directly in the test. It feels unsafe to have a function that ignores errors in a non-test file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-:-KxkwDI6v7dCVLipMEdh:bzh32nm)*

justinj:
A recent PR added a `util/json` package so I've put it there now. It can't be added to `RandDatum` at the moment because it doesn't have a corresponding column type (but I'll make a note in my to-do list that it should go there with that PR). So this doesn't actually generate a Datum at the moment, but it's still useful for internal JSON testing.

---

---

*[pkg/sql/parser/json_random.go, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-KxkxdP_4GG8DakeGX43:-Kxtkw87xA7YinR_g8wh:b-etdmyo) ([raw file](https://github.com/cockroachdb/cockroach/blob/0cc321137c83854606c1f4700ac3c8eff848ae49/pkg/sql/parser/json_random.go#L23)):*

I'd remove this function and call it directly in the test. It feels unsafe to have a function that ignores errors in a non-test file.
</blockquote></details>

I just made it pass through the error so the test has to handle it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-:-Kxtl-1NUddMs2_gTG_3:bapjw4)*

justinj:
@mjibson anything else in the immediate term for this PR? I'd like to have this for testing round-tripping of encoding and decoding. Interested in particular if you have any thoughts on if the quality of the random objects could be improved in any way.

jordanlewis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-:-KyRMHaVdQ79WM4ZaFiU:bnfp4nl)*

mjibson:
Add something to https://github.com/cockroachdb/cockroach/blob/8fbbcee51ea89f72f8c6aee6ccf1c29a17b2a4cc/pkg/internal/rsg/rsg.go#L215 so the RSG tests work again.

I think the quality of the random objects is good.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-:-KyRucxKMlWRKiX0AxLB:b2pid37)*

justinj:
Ok, done.

I ran
```
make test PKG="./pkg/sql" TESTS=TestRandomSyntaxGeneration
```
and nothing exploded, is there anything else I need to update to get them running again?

mjibson:
Run `make test PKG=./pkg/sql 'TESTS=TestRandomSyntaxFunctions' 'TESTFLAGS=-rsg 1m'`

mjibson:

---

*[pkg/internal/rsg/rsg.go, line 267 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-KyYymowJntfWQw7a-Xx:-KyYymowJntfWQw7a-Xy:b7ci1r9) ([raw file](https://github.com/cockroachdb/cockroach/blob/ef884bb2c79468ed3ea77b022e91795f2349eb6d/pkg/internal/rsg/rsg.go#L267)):*
> ```Go
> 			panic(err)
> 		}
> 		v = parser.DJSON{JSON: j}
> ```

See a few lines below where `v` is used with %v in fmt.Sprintf. Will a parser.DJSON convert itself to a string in that case? If not, then v should be set to the correct string here, just like the other types.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608)*

justinj:
OK, updated and rebased!

mjibson:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19608#-:-Kyvr0oTNqsRCS3fVF2C:bnfp4nl)*

justinj:
Thanks for the reviews!

PR #19616, opened 2017-10-28
distsql: omit unused renders during planning

RaduBerinde:
Omitting unused renders when doing distsql planning for renderNodes.

In practice, the unused renders usually get projected out; but having them
there initially causes the planning process to use renders instead of a
projection. For example, without this change the plan for the new test case
shows `Render: @1` instead of `Out: @1`.


rjnn (approved):
LGTM

PR #19620, opened 2017-10-29
storage: interface *Replica to ReplicaEvalContext

tbg:
Introduce an interface implemented by `*Replica` and used by
`ReplicaEvalContext` to prepare the way for moving `ReplicaEvalContext` into a
subpackage of `storage`.

See #18779.

### DIFF
--- a/pkg/ccl/storageccl/export.go
+++ b/pkg/ccl/storageccl/export.go
@@ -112,7 +112,7 @@ func evalExport(
 	// threshold. If it's not, the mvcc tombstones could have been deleted and
 	// the resulting RocksDB tombstones compacted, which means we'd miss
 	// deletions in the incremental backup.
-	gcThreshold, err := cArgs.EvalCtx.GCThreshold()
+	gcThreshold, err := cArgs.EvalCtx.GetGCThreshold()
 	if err != nil {
 		return storage.EvalResult{}, err
 	}
--- a/pkg/storage/helpers_test.go
+++ b/pkg/storage/helpers_test.go
@@ -246,22 +246,6 @@ func (r *Replica) DescLocked() *roachpb.RangeDescriptor {
 	return r.mu.state.Desc
 }
 
-// GetGCThreshold returns the range's GCThreshold, acquiring a replica lock in
-// the process.
-func (r *Replica) GetGCThreshold() hlc.Timestamp {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
-	return *r.mu.state.GCThreshold
-}
-
-// GetTxnSpanGCThreshold returns the range's TxnSpanGCThreshold, acquiring a replica lock in
-// the process.
-func (r *Replica) GetTxnSpanGCThreshold() hlc.Timestamp {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
-	return *r.mu.state.TxnSpanGCThreshold
-}
-
 func (r *Replica) AssertState(ctx context.Context, reader engine.Reader) {
 	r.raftMu.Lock()
 	defer r.raftMu.Unlock()
@@ -284,13 +268,6 @@ func (r *Replica) GetLastIndex() (uint64, error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	return r.raftLastIndexLocked()
-}
-
-// GetLease exposes replica.getLease for tests.
-// If you just need information about the lease holder, consider issuing a
-// LeaseInfoRequest instead of using this internal method.
-func (r *Replica) GetLease() (roachpb.Lease, *roachpb.Lease) {
-	return r.getLease()
 }
 
 // SetQuotaPool allows the caller to set a replica's quota pool initialized to
--- a/pkg/storage/push_txn_queue.go
+++ b/pkg/storage/push_txn_queue.go
@@ -122,21 +122,23 @@ func (pt *pendingTxn) getDependentsSet() map[uuid.UUID]struct{} {
 	return set
 }
 
-// A pushTxnQueue enqueues PushTxn requests which are waiting on
-// extant txns with conflicting intents to abort or commit.
+// A PushTxnQueue enqueues PushTxn requests which are waiting on extant txns
+// with conflicting intents to abort or commit.
 //
-// Internally, it maintains a map from extant txn IDs to queues of
-// pending PushTxn requests.
+// Internally, it maintains a map from extant txn IDs to queues of pending
+// PushTxn requests.
 //
-// When a write intent is encountered, the command which encountered
-// it (called the "pusher" here) initiates a PushTxn request to
-// determine the disposition of the intent's transaction (called the
-// "pushee" here). This queue is where a PushTxn request will wait
-// if it discovers that the pushee's transaction is still pending,
-// and cannot be otherwise aborted or pushed forward.
+// When a write intent is encountered, the command which encountered it (called
+// the "pusher" here) initiates a PushTxn request to determine the disposition
+// of the intent's transaction (called the "pushee" here). This queue is where a
+// PushTxn request will wait if it discovers that the pushee's transaction is
+// still pending, and cannot be otherwise aborted or pushed forward.
 //
-// pushTxnQueue is thread safe.
-type pushTxnQueue struct {
+// PushTxnQueue is thread safe.
+//
+// TODO(tschottdorf): seems straightforward to factor this out and to depend on
+// *Store only through a narrow interface (Clock, DB, Stopper).
+type PushTxnQueue struct {
 	store *Store
 	mu    struct {
 		syncutil.Mutex
@@ -145,8 +147,8 @@ type pushTxnQueue struct {
 	}
 }
 
-func newPushTxnQueue(store *Store) *pushTxnQueue {
-	return &pushTxnQueue{
+func newPushTxnQueue(store *Store) *PushTxnQueue {
+	return &PushTxnQueue{
 		store: store,
 	}
 }
@@ -154,7 +156,7 @@ func newPushTxnQueue(store *Store) *pushTxnQueue {
 // Enable allows transactions to be enqueued and waiting pushers
 // added. This method must be idempotent as it can be invoked multiple
 // times as range leases are updated for the same replica.
-func (ptq *pushTxnQueue) Enable() {
+func (ptq *PushTxnQueue) Enable() {
 	ptq.mu.Lock()
 	defer ptq.mu.Unlock()
 	if ptq.mu.txns == nil {
@@ -170,7 +172,7 @@ func (ptq *pushTxnQueue) Enable() {
 // `disable` is true, future transactions may not be enqueued or
 // waiting pushers added. Call Enable() once the lease is again
 // acquired by the replica.
-func (ptq *pushTxnQueue) Clear(disable bool) {
+func (ptq *PushTxnQueue) Clear(disable bool) {
 	ptq.mu.Lock()
 	var pushWaiters []chan *roachpb.Transaction
 	for _, pt := range ptq.mu.txns {
@@ -214,7 +216,7 @@ func (ptq *pushTxnQueue) Clear(disable bool) {
 	}
 }
 
-func (ptq *pushTxnQueue) isEnabled() bool {
+func (ptq *PushTxnQueue) isEnabled() bool {
 	ptq.mu.Lock()
 	defer ptq.mu.Unlock()
 	return ptq.mu.txns != nil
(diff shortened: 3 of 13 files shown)


nvanbenschoten:

---

*[pkg/storage/replica_accessors.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhTkJJ2Aub8U9kyWBT:-KxhTkJJ2Aub8U9kyWBU:b-punvng) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_accessors.go#L32)):*
> ```Go
> )
> 
> // ReplicaI is the glue between a ReplicaEvalContext and a *Replica
> ```

This will be moving to the same package as `ReplicaEvalContext`, right?

---

*[pkg/storage/replica_accessors.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhSoyM_9c7CfFCEFG-:-KxhSoyM_9c7CfFCEFG0:b5dxya9) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_accessors.go#L34)):*
> ```Go
> // ReplicaI is the glue between a ReplicaEvalContext and a *Replica
> // to avoid a direct dependency.
> type ReplicaI interface {
> ```

This unordered list looks like it will get unwieldy as we continue to develop on it. Is there a good way to logically group some of these methods together to add some structure here (no need to pull out separate interfaces, just chunk up the methods)? If not, perhaps it would be better in alphabetical order. 

---

*[pkg/storage/replica_accessors.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhUygMlstiJrx5X1vC:-KxhUygMlstiJrx5X1vD:b-hn7e0l) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_accessors.go#L34)):*
> ```Go
> // ReplicaI is the glue between a ReplicaEvalContext and a *Replica
> // to avoid a direct dependency.
> type ReplicaI interface {
> ```

Think a `var _ ReplicaI = &Replica{}` is worth it somewhere? This should be statically asserting this already somewhere, but that makes it a bit more explicit and localizes any errors.

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-KxhURQVO9ILepYxp0QA:b-arswus) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*
> ```Go
> // the SpanSet if one is given.
> type ReplicaEvalContext struct {
> 	ReplicaI
> ```

I have mixed feelings about embedding this interface. I think the separation not embedding it provides by not leaking all `ReplicaI` methods to `ReplicaEvalContext` is valuable. What do you think?x.....,,

---

*[pkg/storage/replica_state.go, line 266 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhUK2IIb7nC6uZi0I7:-KxhUK2IIb7nC6uZi0I8:b4czd3o) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L266)):*
> ```Go
> }
> 
> // GetTxnSpanGCThreshold returns the time of the replica's last transactino span
> ```

transactino -> transaction

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620)*

bdarnell:

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-Kxht6CS3wix2h4exjps:b-hsjxx8) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*

I have mixed feelings about embedding this interface. I think the separation not embedding it provides by not leaking all `ReplicaI` methods to `ReplicaEvalContext` is valuable. What do you think?x.....,,
</blockquote></details>

I think I'd make some renamings: the new interface should be `ReplicaEvalContext`, and the old `ReplicaEvalContext` is now `SpanSetReplicaEvalContext` (which wraps a `ReplicaEvalContext`, adds the `SpanSet` checks, and also implements `ReplicaEvalContext`). 

Embedding the wrapped object means there's less code since we don't need a bunch of one-line forwarding methods, but it means that we could easily miss one that needs a `SpanSet` check. I'd rather not embed and have all the forwarding methods. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620)*

tschottdorf:
Review status: 5 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/storage/replica_accessors.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhTkJJ2Aub8U9kyWBT:-Kxi-8zsZsqwaMqL0FM1:b-hd0rce) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_accessors.go#L32)):*

This will be moving to the same package as `ReplicaEvalContext`, right?
</blockquote></details>

Yep, planning to move both into the (newly created) `batcheval`.

---

*[pkg/storage/replica_accessors.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhSoyM_9c7CfFCEFG-:-Kxi-AcnDSE9Q5BsrAm0:b-45jf3v) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_accessors.go#L34)):*

This unordered list looks like it will get unwieldy as we continue to develop on it. Is there a good way to logically group some of these methods together to add some structure here (no need to pull out separate interfaces, just chunk up the methods)? If not, perhaps it would be better in alphabetical order. 
</blockquote></details>

Cleaned up the list a bit. The naming is still inconsistent but unfortunately that's due to the methods on Replica. I think it's better to clean those up separately since the diff will be noisy.

---

*[pkg/storage/replica_accessors.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhUygMlstiJrx5X1vC:-Kxi_NGFTAGk5n-68Ria:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_accessors.go#L34)):*

Think a `var _ ReplicaI = &Replica{}` is worth it somewhere? This should be statically asserting this already somewhere, but that makes it a bit more explicit and localizes any errors.
</blockquote></details>

Done.

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-Kxi-H5aVbM3abtuu5LE:bt46j5s) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*

I think I'd make some renamings: the new interface should be `ReplicaEvalContext`, and the old `ReplicaEvalContext` is now `SpanSetReplicaEvalContext` (which wraps a `ReplicaEvalContext`, adds the `SpanSet` checks, and also implements `ReplicaEvalContext`). 

Embedding the wrapped object means there's less code since we don't need a bunch of one-line forwarding methods, but it means that we could easily miss one that needs a `SpanSet` check. I'd rather not embed and have all the forwarding methods. 
</blockquote></details>

I originally made `ReplicaEvalContext` implement an interface, but that forces heap allocations which didn't seem like a good idea. The current code has the benefit of only hiding `*Replica` behind an interface, so that no allocations are added. Let me know if you disagree, but I think it's better as is.

I changed the embedding though and added the appropriate passthroughs.

---

*[pkg/storage/replica_state.go, line 266 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhUK2IIb7nC6uZi0I7:-Kxid5dg535ei40ju_KP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L266)):*

transactino -> transaction
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620)*

bdarnell:
Review status: 5 of 13 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-KxiejTI-9tfxkamYeg8:bnjyiaj) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*

I originally made `ReplicaEvalContext` implement an interface, but that forces heap allocations which didn't seem like a good idea. The current code has the benefit of only hiding `*Replica` behind an interface, so that no allocations are added. Let me know if you disagree, but I think it's better as is.

I changed the embedding though and added the appropriate passthroughs.
</blockquote></details>

Since both `*Replica` (which is already heap-allocated) and `ReplicaEvalContext` both implement the same interface, and `SpanSets` are disabled in production, we could pass the `*Replica` directly in most cases and only allocate a `ReplicaEvalContext` when we need `SpanSets`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620)*

tschottdorf:
PTAL

---

Review status: 5 of 13 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-KxinnuJ0CFoWJJT_olf:bmwgc5n) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*

Since both `*Replica` (which is already heap-allocated) and `ReplicaEvalContext` both implement the same interface, and `SpanSets` are disabled in production, we could pass the `*Replica` directly in most cases and only allocate a `ReplicaEvalContext` when we need `SpanSets`.
</blockquote></details>

That's a good idea and I started pulling on this thread, but it started incurring a large diff that I wasn't comfortable pulling into this PR. I'd like to move things first and then unravel them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-:-Kxio-IeJko0bldmT0WW:bb71pro)*

tschottdorf:
Review status: 5 of 13 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-Kxiq3-eeTqDbNKysmun:b-f1xu80) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*

That's a good idea and I started pulling on this thread, but it started incurring a large diff that I wasn't comfortable pulling into this PR. I'd like to move things first and then unravel them.
</blockquote></details>

(btw, they don't impl the same interface -- the asserted version returns more errors. That can be fixed, but is it worth introducing the awkwardness? I'm not sure what we win by changing the way it works except some probably negligible `ss != nil` checks).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620)*

tschottdorf:
Review status: 5 of 13 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica_state.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-KxhURQVO9ILepYxp0Q9:-KxiurLdu3FeH7jbza3U:b-s2tj8g) ([raw file](https://github.com/cockroachdb/cockroach/blob/eb0045e9493b471bc978bdf24e006d73b443358c/pkg/storage/replica_state.go#L123)):*

(btw, they don't impl the same interface -- the asserted version returns more errors. That can be fixed, but is it worth introducing the awkwardness? I'm not sure what we win by changing the way it works except some probably negligible `ss != nil` checks).
</blockquote></details>

Hmm, it would help by hiding the SpanSet from the subpackage. I'll give it another try.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620)*

bdarnell:
:lgtm:

---

Reviewed 8 of 8 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19620#-:-Kxj0QfX6ZgvkXhJ5Ntk:bnfp4nl)*

tschottdorf:
TFTRs! I piled two more commits on top of this but I'll just open them in a new PR to avoid conflicts below.

PR #19626, opened 2017-10-29
cherrypick-1.1: ui: exclude decommissioned nodes from staggered version warning

### LINKED ISSUE #15534: ui: "staggered version" banner considers dead nodes

tbg:
Fixes #15534.

cc @cockroachdb/release

### DIFF
--- a/pkg/ui/src/redux/alerts.spec.ts
+++ b/pkg/ui/src/redux/alerts.spec.ts
@@ -9,13 +9,14 @@ import { AdminUIState, createAdminUIStore } from "./state";
 import {
   AlertLevel,
   alertDataSync,
+  versionsSelector,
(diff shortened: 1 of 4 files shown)


couchand (approved):

bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19626#-:-KxhwPaic3hKQ3CHsmXT:bnfp4nl)*

PR #19627, opened 2017-10-29
sql: add <@, @> JSON operators

justinj:
The behaviour of this operator is described at
https://www.postgresql.org/docs/9.5/static/datatype-json.html#JSON-CONTAINMENT

This will probably have to change a bit once we have an encoding, but I
wanted to implement it now for the simpler case.

I debated first going with a simpler but always O(n^2) implementation of jsonArrayContains, it looked like this:

```go
func jsonArrayContains(ctx *EvalContext, ary DJSON, other DJSON) bool {
	for _, v := range other.arrayVal {
		found := false
		for _, c := range ary.arrayVal {
			if c.typ == v.typ && c.contains(ctx, v) {
				found = true
				break
			}
		}
		if !found {
			return false
		}
	}
	return true
}
```

The current implementation is quite a bit more complex, but significantly faster.
On this benchmark:

```go
func BenchmarkJSONArrayContains(b *testing.B) {
	ctx := NewTestingEvalContext()
	for arySize := 0; arySize <= 20; arySize += 5 {
		b.Run(fmt.Sprintf("JSON array of size %d", arySize), func(b *testing.B) {
			var buf bytes.Buffer
			buf.WriteByte('[')
			for i := 0; i < arySize; i++ {
				if i > 0 {
					buf.WriteByte(',')
				}
				buf.WriteString(fmt.Sprintf("%d", i))
			}
			buf.WriteByte(']')
			j, _ := ParseDJSON(buf.String())
			ary := *j.(*DJSON)

			// ary is now an array of the form [0, 1, 2, ...].
			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				jsonArrayContains(ctx, ary, ary)
			}
			b.StopTimer()
		})
	}
}
```

The original implementation:
```
BenchmarkJSONArrayContains/JSON_array_of_size_0-8          200000000          9.30 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_5-8            500000       3274 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_10-8           100000      11784 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_15-8            50000      25897 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_20-8            30000      45286 ns/op
```

The current implementation:
```
BenchmarkJSONArrayContains/JSON_array_of_size_0-8           5000000        285 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_5-8           1000000       2088 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_10-8           300000       3756 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_15-8           200000       6238 ns/op
BenchmarkJSONArrayContains/JSON_array_of_size_20-8           200000       8365 ns/op
```

Re: the second TODO on `preprocessJSONArrayForContains`,
We could maybe transform
```
a @> b
```
into
```
PREPROCESS(a) @>' PREPROCESS(b)
```
where @>' assumes its arguments are preprocessed, and then the constant folding code
would take care of this. cc @knz for thoughts on this?

### DIFF
--- a/docs/generated/sql/operators.md
+++ b/docs/generated/sql/operators.md
@@ -192,6 +192,11 @@
 <tr><td><a href="uuid.html">uuid</a> <code><=</code> <a href="uuid.html">uuid</a></td><td><a href="bool.html">bool</a></td></tr>
 </tbody></table>
 <table><thead>
+<tr><td><code><@</code></td><td>Return</td></tr>
+</thead><tbody>
+<tr><td>jsonb <code><@</code> jsonb</td><td><a href="bool.html">bool</a></td></tr>
+</tbody></table>
+<table><thead>
 <tr><td><code>=</code></td><td>Return</td></tr>
 </thead><tbody>
 <tr><td><a href="bool.html">bool</a> <code>=</code> <a href="bool.html">bool</a></td><td><a href="bool.html">bool</a></td></tr>
@@ -258,6 +263,11 @@
 <tr><td><code>?|</code></td><td>Return</td></tr>
 </thead><tbody>
 <tr><td>jsonb <code>?|</code> <a href="string.html">string[]</a></td><td><a href="bool.html">bool</a></td></tr>
+</tbody></table>
+<table><thead>
+<tr><td><code>@></code></td><td>Return</td></tr>
+</thead><tbody>
+<tr><td>jsonb <code>@></code> jsonb</td><td><a href="bool.html">bool</a></td></tr>
 </tbody></table>
 <table><thead>
 <tr><td><code>ILIKE</code></td><td>Return</td></tr>
--- a/pkg/sql/logictest/testdata/logic_test/json
+++ b/pkg/sql/logictest/testdata/logic_test/json
@@ -336,3 +336,23 @@ SELECT '3'::JSONB - 'b'
 
 statement error pgcode 22023 cannot delete from object using integer index
 SELECT '{}'::JSONB - 1
+
+query B
+SELECT '[1, 2, 3]'::JSONB <@ '[1, 2]'::JSONB
+----
+false
+
+query B
+SELECT '[1, 2]'::JSONB <@ '[1, 2, 3]'::JSONB
+----
+true
+
+query B
+SELECT '[1, 2]'::JSONB @> '[1, 2, 3]'::JSONB
+----
+false
+
+query B
+SELECT '[1, 2, 3]'::JSONB @> '[1, 2]'::JSONB
+----
+true
--- a/pkg/sql/sem/tree/eval.go
+++ b/pkg/sql/sem/tree/eval.go
@@ -1895,6 +1895,26 @@ var CmpOps = map[ComparisonOperator]cmpOpOverload{
 			},
 		},
 	},
+
+	Contains: {
+		CmpOp{
+			LeftType:  types.JSON,
+			RightType: types.JSON,
+			fn: func(ctx *EvalContext, left Datum, right Datum) (Datum, error) {
+				return MakeDBool(DBool(json.Contains(left.(*DJSON).JSON, right.(*DJSON).JSON))), nil
+			},
+		},
+	},
+
+	ContainedBy: {
+		CmpOp{
+			LeftType:  types.JSON,
+			RightType: types.JSON,
+			fn: func(ctx *EvalContext, left Datum, right Datum) (Datum, error) {
+				return MakeDBool(DBool(json.Contains(right.(*DJSON).JSON, left.(*DJSON).JSON))), nil
+			},
+		},
+	},
 }
 
 func boolFromCmp(cmp int, op ComparisonOperator) *DBool {
--- a/pkg/util/json/contains.go
+++ b/pkg/util/json/contains.go
@@ -1,0 +1,188 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package json
+
+import "sort"
+
+// Contains returns true if a contains b. This implements the @>, <@ operators.
+// See the Postgres docs for the expected semantics of Contains.
+// https://www.postgresql.org/docs/10/static/datatype-json.html#JSON-CONTAINMENT
+// The naive approach to doing array containment would be to do an O(n^2)
+// nested loop through the arrays to check if one is contained in the
+// other.  We're out of luck when the arrays contain other arrays or
+// objects (there might actually be something fancy we can do, but there's nothing
+// obvious).
+// When the arrays contain scalars however, we can optimize this by
+// pre-sorting both arrays and iterating through them in lockstep.
+// To this end, we preprocess the JSON document to sort all of its arrays so
+// that when we perform contains we can extract the scalars sorted, and then
+// also the arrays and objects in separate arrays, so that we can do the fast
+// thing for the subset of the arrays which are scalars.
+func Contains(a, b JSON) bool {
+	preA := a.preprocessForContains()
+	preB := b.preprocessForContains()
+	return preA.contains(preB)
+}
+
+// containsable is an interface used internally for the implementation of @>.
+type containsable interface {
+	contains(other containsable) bool
+}
+
+// containsableScalar is a preprocessed JSON scalar. The JSON it holds will
+// never be a JSON object or a JSON array.
+type containsableScalar struct{ JSON }
+
+// containsableArray is a preprocessed JSON array.
+// * scalars will always be scalars and will always be sorted,
+// * arrays will only contain containsableArrays,
+// * objects will only contain containsableObjects
+// (the last two are stored interfaces for reuse, though)
+type containsableArray struct {
+	scalars []containsableScalar
+	arrays  []containsable
+	objects []containsable
+}
+
+type containsableKeyValuePair struct {
+	k jsonString
+	v containsable
+}
+
+// containsableObject is a preprocessed JSON object.
+// Same as a jsonObject, it is stored as a sorted-by-key list of key-value
+// pairs.
+type containsableObject []containsableKeyValuePair
+
+func (j jsonNull) preprocessForContains() containsable   { return containsableScalar{j} }
+func (j jsonFalse) preprocessForContains() containsable  { return containsableScalar{j} }
+func (j jsonTrue) preprocessForContains() containsable   { return containsableScalar{j} }
+func (j jsonNumber) preprocessForContains() containsable { return containsableScalar{j} }
+func (j jsonString) preprocessForContains() containsable { return containsableScalar{j} }
+
+func (j jsonArray) preprocessForContains() containsable {
+	result := containsableArray{}
+	for _, e := range j {
+		switch v := e.(type) {
+		case jsonArray:
+			result.arrays = append(result.arrays, v.preprocessForContains())
+		case jsonObject:
+			result.objects = append(result.objects, v.preprocessForContains())
+		default:
+			result.scalars = append(result.scalars, e.preprocessForContains().(containsableScalar))
+		}
+	}
+
+	sort.Slice(result.scalars, func(i, j int) bool {
+		return result.scalars[i].JSON.Compare(result.scalars[j].JSON) == -1
+	})
+
+	return result
+}
+
+func (j jsonObject) preprocessForContains() containsable {
+	preprocessed := make(containsableObject, len(j))
+
+	for i := range preprocessed {
+		preprocessed[i].k = j[i].k
+		preprocessed[i].v = j[i].v.preprocessForContains()
+	}
+
+	return preprocessed
+}
+
+func (j containsableScalar) contains(other containsable) bool {
+	if o, ok := other.(containsableScalar); ok {
+		return j.JSON.Compare(o.JSON) == 0
+	}
+	return false
+}
+
+func (j containsableArray) contains(other containsable) bool {
+	// This is a unique case of contains (and is described as such in the
+	// Postgres docs) - an array contains a scalar which is an element of it.
+	// This contradicts the general rule of contains that the contained object
+	// must have the same "shape" as the containing object.
+	if o, ok := other.(containsableScalar); ok {
+		found := sort.Search(len(j.scalars), func(i int) bool {
+			return j.scalars[i].JSON.Compare(o.JSON) >= 0
+		})
+		return found < len(j.scalars) && j.scalars[found].JSON.Compare(o.JSON) == 0
+	}
+
+	if contained, ok := other.(containsableArray); ok {
+		// Since both slices of scalars are sorted via the preprocessing, we can
+		// step through them together via binary search.
+		remainingScalars := j.scalars[:]
+		for _, val := range contained.scalars {
+			found := sort.Search(len(remainingScalars), func(i int) bool {
+				return remainingScalars[i].JSON.Compare(val.JSON) >= 0
+			})
+			if found == len(remainingScalars) || remainingScalars[found].JSON.Compare(val.JSON) != 0 {
+				return false
+			}
+			remainingScalars = remainingScalars[found:]
+		}
+
+		// TODO(justin): there's possibly(?) something fancier we can do with the
+		// objects and arrays, but for now just do the quadratic check.
+		if !quadraticJSONArrayContains(j.objects, contained.objects) ||
+			!quadraticJSONArrayContains(j.arrays, contained.arrays) {
+			return false
+		}
+		return true
+	}
+	return false
+}
+
+// quadraticJSONArrayContains does an O(n^2) check to see if every value in
+// `other` is contained within a value in `container`. `container` and `other`
+// should not contain scalars.
+func quadraticJSONArrayContains(container, other []containsable) bool {
+	for _, otherVal := range other {
+		found := false
+		for _, containerVal := range container {
+			if containerVal.contains(otherVal) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	return true
+}
+
+func (j containsableObject) contains(other containsable) bool {
+	if contained, ok := other.(containsableObject); ok {
+		// We can iterate through the keys of `other` and scan through to find the
+		// corresponding keys in `j` since they're both sorted.
+		objIdx := 0
+		for _, rightEntry := range contained {
+			for objIdx < len(j) && j[objIdx].k < rightEntry.k {
+				objIdx++
+			}
+			if objIdx >= len(j) ||
+				j[objIdx].k != rightEntry.k ||
+				!j[objIdx].v.contains(rightEntry.v) {
+				return false
+			}
+			objIdx++
+		}
+		return true
+	}
+	return false
+}
--- a/pkg/util/json/contains_testers.go
+++ b/pkg/util/json/contains_testers.go
@@ -1,0 +1,122 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+// This file contains (ha!) helpers to test @>.
+
+package json
+
+import (
+	"math/rand"
+)
+
+type containsTester interface {
+	// slowContains is a slower but simpler implementation of contains to check
+	// against the substantially more complex actual implementation.
+	slowContains(other JSON) bool
+
+	// subdocument returns a JSON document which is contained by this one.
+	subdocument(isRoot bool, rng *rand.Rand) JSON
+}
+
+func (j jsonNull) slowContains(other JSON) bool   { return j.Compare(other) == 0 }
+func (j jsonTrue) slowContains(other JSON) bool   { return j.Compare(other) == 0 }
+func (j jsonFalse) slowContains(other JSON) bool  { return j.Compare(other) == 0 }
+func (j jsonNumber) slowContains(other JSON) bool { return j.Compare(other) == 0 }
+func (j jsonString) slowContains(other JSON) bool { return j.Compare(other) == 0 }
+
+func (j jsonArray) slowContains(other JSON) bool {
+	if other.isScalar() {
+		for i := 0; i < len(j); i++ {
+			if j[i].Compare(other) == 0 {
+				return true
+			}
+		}
+	}
+
+	if ary, ok := other.(jsonArray); ok {
+		for i := 0; i < len(ary); i++ {
+			found := false
+			for k := 0; k < len(j); k++ {
+				if j[k].jsonType() == ary[i].jsonType() && j[k].(containsTester).slowContains(ary[i]) {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return false
+			}
+		}
+		return true
+	}
+	return false
+}
+
+func (j jsonObject) slowContains(other JSON) bool {
+	if obj, ok := other.(jsonObject); ok {
+		for i := 0; i < len(obj); i++ {
+			leftVal := j.FetchValKey(string(obj[i].k))
+			if leftVal == nil || !leftVal.(containsTester).slowContains(obj[i].v) {
+				return false
+			}
+		}
+		return true
+	}
+	return false
+}
+
+func (j jsonNull) subdocument(_ bool, _ *rand.Rand) JSON   { return j }
+func (j jsonTrue) subdocument(_ bool, _ *rand.Rand) JSON   { return j }
+func (j jsonFalse) subdocument(_ bool, _ *rand.Rand) JSON  { return j }
+func (j jsonNumber) subdocument(_ bool, _ *rand.Rand) JSON { return j }
+func (j jsonString) subdocument(_ bool, _ *rand.Rand) JSON { return j }
+
+func (j jsonArray) subdocument(isRoot bool, rng *rand.Rand) JSON {
+	// Root arrays contain their scalar elements.
+	if isRoot && rng.Intn(5) == 0 {
+		idx := rng.Intn(len(j))
+		if j[idx].isScalar() {
+			return j[idx]
+		}
+	}
+	result := make(jsonArray, 0)
+	i := 0
+	for i < len(j) {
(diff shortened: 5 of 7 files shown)


knz:
Please extend the commit message to explain how the two algorithms (the naive one and the better one) differ.

I would welcome exposing the normalization ("preprocess") function as a builtin, for use in tests, and then you could also experiment doing the desugaring you just proposed and measure whether it's beneficial for performance.

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-:-KxsJ_M565QwjZBoecVD:b-aks0k8)*

justinj:
Good call @knz, I updated with a more detailed comment of how the check is performed and augmented the commit message.

knz:
LGTM!

out of curiosity, what do you intend to do wrt the idea to preprocess before `@>`?

justinj:
I figured I would leave that out of scope for this PR, it seems pretty low-priority to me right now given the other JSON stuff that remains to be done, but I can leave a TODO for it.

jordanlewis:
:lgtm:

---

Reviewed 7 of 7 files at r2.

---

*[pkg/util/json/json.go, line 599 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky2wnjQxMJrm3RgL9xy:-Ky2wnjQxMJrm3RgL9xz:b-xjqa49) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L599)):*
> ```Go
> 		i := 0
> 		for _, val := range otherScalars {
> 			for i < len(containerScalars) && containerScalars[i].Compare(val) == -1 {
> ```

You could use binary search here I think.

```
remainingScalars := containerScalars[:]
for _, val := range otherScalars {
  found := sort.Search(len(remainingScalars), func(i int) bool { return remainingScalars[i].Compare(val) >= 0 } )
  if found == len(remainingScalars) || remainingScalars[found].Compare(val) != 0 {
    return false
  }
  remainingScalars = remainingScalars[found+1:]
}
```

Disclaimer: I didn't test that

---

*[pkg/util/json/json.go, line 647 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky2w8BVz6lcr6jg-OgU:-Ky2w8BVz6lcr6jg-OgV:b-alck85) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L647)):*
> ```Go
> // permute the input array and then pass back up a way to permute it back to
> // the way it was once the caller was done with it.
> // TODO(justin): if a JSON document is given as a literal for the @> operator,
> ```

Seems like you could do this in `constant.go` where `StrVal` gets converted to JSON. Now that I look, though, it seems like there isn't a case there for JSON. Is that deliberate?

---

*[pkg/util/json/json.go, line 656 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky3-jsQdmb8dcRQb74B:-Ky3-jsQdmb8dcRQb74C:b8rc7zp) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L656)):*
> ```Go
> 	// objects.
> 	sort.Slice(sorted, func(i, j int) bool {
> 		return sorted[i].Compare(sorted[j]) == -1
> ```

Wait won't this be really slow, since it has to actually sort the arrays and objects too? Is it worth writing a CompareShallow or similar for this purpose?

---

*[pkg/util/json/json_test.go, line 551 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky2swBSpul_emLvlAPd:-Ky2swBSpul_emLvlAPe:b-wx5cxc) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json_test.go#L551)):*
> ```Go
> 			{`[[1, 2]]`, true},
> 			{`[3, [1, 2]]`, true},
> 			{`[[1, 2], 3]`, true},
> ```

What about `[[1],[2],3]`? That seems sufficiently different that it's worth testing.  I'm not even sure if it's supposed to be true or false!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-:-Ky303hGlOrXcr7oakn4:bnfp4nl)*

justinj:
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/util/json/json.go, line 599 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky2wnjQxMJrm3RgL9xy:-Ky37_WoebV1NDX4Jnoi:b8xob0a) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L599)):*

You could use binary search here I think.

```
remainingScalars := containerScalars[:]
for _, val := range otherScalars {
  found := sort.Search(len(remainingScalars), func(i int) bool { return remainingScalars[i].Compare(val) >= 0 } )
  if found == len(remainingScalars) || remainingScalars[found].Compare(val) != 0 {
    return false
  }
  remainingScalars = remainingScalars[found+1:]
}
```

Disclaimer: I didn't test that
</blockquote></details>

:fire: cool

---

*[pkg/util/json/json.go, line 647 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky2w8BVz6lcr6jg-OgU:-Ky39m-nnpZn687gWBNA:b-ymd0ab) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L647)):*

Seems like you could do this in `constant.go` where `StrVal` gets converted to JSON. Now that I look, though, it seems like there isn't a case there for JSON. Is that deliberate?
</blockquote></details>

No it isn't, I'll make a note to add one in a future PR. And as we discussed it might be easiest to do this preprocessing during normalization.

---

*[pkg/util/json/json.go, line 656 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky3-jsQdmb8dcRQb74B:-Ky37hSb_UysGANcyoyn:b-oxe0md) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L656)):*

Wait won't this be really slow, since it has to actually sort the arrays and objects too? Is it worth writing a CompareShallow or similar for this purpose?
</blockquote></details>

If I understand you correctly, no it won't have to sort the arrays and objects for `Compare`. It will have to sort them when it gets to the point where it checks containment of them in which case it will have to sort them when `Contain` gets called on them.

But your point does make me realize that `[[...]] @> [<a bunch of arrays>]` will re-sort `[...]` a whole bunch of times to check containment because the preprocessing won't get passed up to the parent `Contains` caller :( I think that's a big issue but it's not immediately clear to me the best way to resolve it, so I'm going to think about it a bit (maybe this is what you were getting at?).

---

*[pkg/util/json/json_test.go, line 551 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky2swBSpul_emLvlAPd:-Ky39Azxw-RKqB9PtGTN:bh9nva6) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json_test.go#L551)):*

What about `[[1],[2],3]`? That seems sufficiently different that it's worth testing.  I'm not even sure if it's supposed to be true or false!
</blockquote></details>

Done (it's true!).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

jordanlewis:
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/util/json/json.go, line 656 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky3-jsQdmb8dcRQb74B:-Ky3D0iUfLlFYO1_5vAk:b-fq8vyq) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L656)):*

If I understand you correctly, no it won't have to sort the arrays and objects for `Compare`. It will have to sort them when it gets to the point where it checks containment of them in which case it will have to sort them when `Contain` gets called on them.

But your point does make me realize that `[[...]] @> [<a bunch of arrays>]` will re-sort `[...]` a whole bunch of times to check containment because the preprocessing won't get passed up to the parent `Contains` caller :( I think that's a big issue but it's not immediately clear to me the best way to resolve it, so I'm going to think about it a bit (maybe this is what you were getting at?).
</blockquote></details>

Good point about the resorting, but that's not what I was getting at.

I meant *compare* not sort. Like, it has to deep-compare all the objects and arrays even though as you point out we don't care about the sortedness of them - just that they're partitioned from the scalars.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/util/json/json.go, line 656 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Ky3-jsQdmb8dcRQb74B:-Ky3PDRzuYl6-sq2NZga:b-jlv6th) ([raw file](https://github.com/cockroachdb/cockroach/blob/c668e8367ccb746f3323fd66f0a38eb14a8d8086/pkg/util/json/json.go#L656)):*

Good point about the resorting, but that's not what I was getting at.

I meant *compare* not sort. Like, it has to deep-compare all the objects and arrays even though as you point out we don't care about the sortedness of them - just that they're partitioned from the scalars.

</blockquote></details>

Ahh good call, done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Ok @jordanlewis I fixed the re-sorting problem, PTAL. I also pulled all the contains logic into its own file.

jordanlewis:
Reviewed 4 of 4 files at r3.

---

*[pkg/util/json/contains.go, line 21 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRZuO2ru0xVxr3kZva:-KyRZuO2ru0xVxr3kZvb:b-yxlqrt) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L21)):*
> ```Go
> )
> 
> // Contains is the public interface to the @>, <@ operators.
> ```

IMO this docstring is at the wrong level of detail. These implementation details/commentaries are great but they don't belong on the public api godoc. You should move them into the body of `Contains`.
.
The docstring should read something like `Contains returns true if a contains b. This is the public interface to @> and <@.` and maybe include a note on the general guiding principle like you described offline - "contains really means "can delete stuff from the bigger one" except for in the case of array containership of scalars" or whatever.

---

*[pkg/util/json/contains.go, line 46 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRUeYvDwEO0n2FAQni:-KyRUeYvDwEO0n2FAQnj:b9dvg2r) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L46)):*
> ```Go
> 
> // shallowLess is a less than implementation for JSON values which considers
> // all objects equal, and all arrays equal (all objects are still less than all
> ```

This comment seems wrong as we discussed

---

*[pkg/util/json/contains.go, line 66 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRVGzc5064BkKI3oqq:-KyRVGzdJoTFIhk-RF9E:bfksoo3) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L66)):*
> ```Go
> // * the scalars in the given slice, sorted,
> // * the arrays in the input slice,
> // * and the objects in the input slice.
> ```

This function does not seem to return 3 slices at all!

---

*[pkg/util/json/contains.go, line 95 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRZIVI8BaEkRWsSkb8:-KyRZIVI8BaEkRWsSkb9:b-b529nr) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L95)):*
> ```Go
> }
> 
> func (j jsonNull) contains(other JSON) bool   { return j.Compare(other) == 0 }
> ```

Couldn't these all be `j` == `jsonFalseVal` etc? Not sure if it matters too much but seems like a quick win.

---

*[pkg/util/json/json.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRZgrTputDNa26mhZz:-KyRZgrTputDNa26mh_-:by9w0no) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/json.go#L93)):*
> ```Go
> 	// contains implements the <@, @> operators. It should only be called on a
> 	// JSON document which has already been preprocessed by
> 	// preprocessForContains.
> ```

Might be time for a sanity check bit in `JSON`? `wasPreprocessed bool`. See also `DTuple.Sorted`.

---

*[pkg/util/json/json_test.go, line 525 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyR_YYObUFpWGCDB1Uv:-KyR_YYObUFpWGCDB1Uw:bc4smz1) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/json_test.go#L525)):*
> ```Go
> 		other    string
> 		expected bool
> 	}{
> ```

Looks like you're missing tests for containership in empty arrays and objects. 

I wonder if you should add a test using the random json generator. You can recheck your algorithm using the simple recursive containership check maybe. I'm nervous that the tests here, while quite comprehensive, are missing edge cases. The algorithm has gotten fairly complicated, especially the array containership one. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Review status: 5 of 6 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/contains.go, line 21 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRZuO2ru0xVxr3kZva:-KyRgttFkYN6O-yfsdLP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L21)):*

IMO this docstring is at the wrong level of detail. These implementation details/commentaries are great but they don't belong on the public api godoc. You should move them into the body of `Contains`.
.
The docstring should read something like `Contains returns true if a contains b. This is the public interface to @> and <@.` and maybe include a note on the general guiding principle like you described offline - "contains really means "can delete stuff from the bigger one" except for in the case of array containership of scalars" or whatever.
</blockquote></details>

Done.

---

*[pkg/util/json/contains.go, line 46 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRUeYvDwEO0n2FAQni:-KyRguMTabKOvA-Dg1WM:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L46)):*

This comment seems wrong as we discussed
</blockquote></details>

Done.

---

*[pkg/util/json/contains.go, line 66 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRVGzc5064BkKI3oqq:-KyRhT1L0s6zQ7fIi_mj:bt2vr0y) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L66)):*

This function does not seem to return 3 slices at all!
</blockquote></details>

Fixed!

---

*[pkg/util/json/contains.go, line 95 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRZIVI8BaEkRWsSkb8:-KyRhUociHEx7Q1Xm__0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/contains.go#L95)):*

Couldn't these all be `j` == `jsonFalseVal` etc? Not sure if it matters too much but seems like a quick win.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyRZgrTputDNa26mhZz:-KyRhVGdSsRZR8KelQZG:b-en78l3) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/json.go#L93)):*

Might be time for a sanity check bit in `JSON`? `wasPreprocessed bool`. See also `DTuple.Sorted`.
</blockquote></details>

I'd rather not add that additional cognitive overhead to `jsonArray` and `jsonObject` if possible for just one operator, maybe I should just bite the bullet and make a separate type for preprocessed documents :S I'll look into that I guess.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

knz:
:lgtm: but I have a big concern remaining on memory usage via the preprocess methods. Either here or in a follow-up PR, please consider plumbing the appropriate memory monitor / account.

---

Reviewed 2 of 7 files at r2, 3 of 4 files at r3, 1 of 1 files at r4.

---

*[pkg/util/json/contains.go, line 89 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KySVcnOJkM082gQ8B96:-KySVcnOJkM082gQ8B97:b-3aib9y) ([raw file](https://github.com/cockroachdb/cockroach/blob/9c057e3b770fe2ea96747cf8573ac54b7255f81b/pkg/util/json/contains.go#L89)):*
> ```Go
> 
> func (j jsonObject) preprocessForContains() JSON {
> 	preprocessed := make(jsonObject, len(j))
> ```

Given the allocation behavior of this method, and that JSON objects can be arbitrarily large, I'd like to see a memory account plumbed through this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-:-KySVnIV2hsJNAiqOGYg:b-2mst9r)*

justinj:
Ok, PTAL, unfortunate the amount of churn this PR has had :(

@knz can you explain your reasoning for the memory monitor a bit more? The allocated data isn't kept around so it would have to be freed at the end of the call, and also all the keys, numbers, strings, and actual data will be shared with the original object (which should already be accounted for).

knz:
Oh it hadn't occurred to me that the original leaf objects were already accounted for. Carry on.

jordanlewis:
Reviewed 3 of 3 files at r5.

---

*[pkg/sql/pgwire/pgerror/codes.go, line 1 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KycPAPrcTEjSFK9c__x:-KycPAPrcTEjSFK9c__y:b-9b1ndl) ([raw file](https://github.com/cockroachdb/cockroach/blob/805058650fbde57982b3ba6f9e6c3152b59d1e49/pkg/sql/pgwire/pgerror/codes.go#L1)):*
> ```Go
> // * All the arrays come after all the objects
> ```

not sure this belongs here =)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Review status: 5 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/util/json/json_test.go, line 525 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyR_YYObUFpWGCDB1Uv:-Kyv1jJ4WjPfw-0DW0IS:bf4r5df) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/json_test.go#L525)):*

Looks like you're missing tests for containership in empty arrays and objects. 

I wonder if you should add a test using the random json generator. You can recheck your algorithm using the simple recursive containership check maybe. I'm nervous that the tests here, while quite comprehensive, are missing edge cases. The algorithm has gotten fairly complicated, especially the array containership one. 
</blockquote></details>

Are your thoughts that we would generate two random JSON documents and check if one was contained in the other? It seems like it would be unlikely for one to be contained in the other by chance. Maybe we could come up with a (complete) set of operations that retain containership? So you generate a document `a` and then apply some set of the operations to it to get a document which `a` contains? I think this would do it?

* Array element deletion
* Array element duplication
* Array shuffling
* Object key deletion
* Perform these operations on a child
* [Only if we're at the root] promote a scalar array element to the root

However, that doesn't handle testing the *negative* case, so maybe both this and the approach you outlined together would be a good combination?

---

*[pkg/sql/pgwire/pgerror/codes.go, line 1 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KycPAPrcTEjSFK9c__x:-Kyv0Z0AWlbvKtAUQvgT:bbvudf6) ([raw file](https://github.com/cockroachdb/cockroach/blob/805058650fbde57982b3ba6f9e6c3152b59d1e49/pkg/sql/pgwire/pgerror/codes.go#L1)):*

not sure this belongs here =)
</blockquote></details>

erg, thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Review status: 1 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/util/json/json_test.go, line 525 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyR_YYObUFpWGCDB1Uv:-KywTamTWH4p6jDhIZ1e:bymo1u5) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/json_test.go#L525)):*

Are your thoughts that we would generate two random JSON documents and check if one was contained in the other? It seems like it would be unlikely for one to be contained in the other by chance. Maybe we could come up with a (complete) set of operations that retain containership? So you generate a document `a` and then apply some set of the operations to it to get a document which `a` contains? I think this would do it?

* Array element deletion
* Array element duplication
* Array shuffling
* Object key deletion
* Perform these operations on a child
* [Only if we're at the root] promote a scalar array element to the root

However, that doesn't handle testing the *negative* case, so maybe both this and the approach you outlined together would be a good combination?
</blockquote></details>

Ok, I gave this a try and I think it's pretty neat!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

jordanlewis:
Review status: 1 of 7 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/json.go, line 49 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Kz-C6ef6-K9sjNCdgeL:-Kz-C6egTTv5VJ0pEWHg:blmhjo) ([raw file](https://github.com/cockroachdb/cockroach/blob/896661a3ce274df0b2c958b1574c111601440b3f/pkg/util/json/json.go#L49)):*
> ```Go
> type JSON interface {
> 	fmt.Stringer
> 	containsTester
> ```

Is there a way to avoid embedding the tester into the production interface?

---

*[pkg/util/json/json_test.go, line 525 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-KyR_YYObUFpWGCDB1Uv:-Kyv5O2rwLOQGTdvwueY:b-cwfhca) ([raw file](https://github.com/cockroachdb/cockroach/blob/4180583f5c47300f00da549231348aae4ef31a61/pkg/util/json/json_test.go#L525)):*

Ok, I gave this a try and I think it's pretty neat!
</blockquote></details>

Sweet! I love it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Review status: 1 of 7 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/json.go, line 49 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Kz-C6ef6-K9sjNCdgeL:-Kz-FymJ9xXuSklKaaSx:b-8c9prh) ([raw file](https://github.com/cockroachdb/cockroach/blob/896661a3ce274df0b2c958b1574c111601440b3f/pkg/util/json/json.go#L49)):*

Is there a way to avoid embedding the tester into the production interface?
</blockquote></details>

I thought about that, and I considered just doing an assertion in the tests, but I think it's possible that we'll add additional implementors of the `JSON` interface (say, to represent numbers that we don't need to encode an entire `DECIMAL` for, and I have a branch that adds an `encoded` implementor) and it would be nice if those gave a type error to remind the implementor to include them in the tests. Is there a downside to having this interface embedded here given that it's not exported?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

jordanlewis:
Review status: 1 of 7 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/json.go, line 49 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Kz-C6ef6-K9sjNCdgeL:-Kz0CiAA-jwZF25kRQte:bv1fgx7) ([raw file](https://github.com/cockroachdb/cockroach/blob/896661a3ce274df0b2c958b1574c111601440b3f/pkg/util/json/json.go#L49)):*

I thought about that, and I considered just doing an assertion in the tests, but I think it's possible that we'll add additional implementors of the `JSON` interface (say, to represent numbers that we don't need to encode an entire `DECIMAL` for, and I have a branch that adds an `encoded` implementor) and it would be nice if those gave a type error to remind the implementor to include them in the tests. Is there a downside to having this interface embedded here given that it's not exported?
</blockquote></details>

I'm not sure if there's a downside besides the general distastefulness (IMO) of having a test-only declaration in a production interface and file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

justinj:
Review status: 1 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/util/json/json.go, line 49 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-Kz-C6ef6-K9sjNCdgeL:-Kz0FarC9feXSiLgxDtJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/896661a3ce274df0b2c958b1574c111601440b3f/pkg/util/json/json.go#L49)):*

I'm not sure if there's a downside besides the general distastefulness (IMO) of having a test-only declaration in a production interface and file.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627)*

jordanlewis:
:lgtm:

---

Reviewed 3 of 8 files at r6, 3 of 3 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19627#-:-Kz0Ft71DUTdfopV6EOe:bnfp4nl)*

justinj:
TFTRs!

PR #19629, opened 2017-10-29
storage: Don't ignore quiesceLocked's return value

a-robinson:
This doesn't look like it was a real correctness issue, but it
certainly seems wasteful to send quiesce messages to all followers
if the leader didn't actually quiesce itself.

### DIFF
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -3924,7 +3924,9 @@ func (r *Replica) quiesceAndNotifyLocked(ctx context.Context, status *raft.Statu
 		return false
 	}
 
-	r.quiesceLocked()
+	if !r.quiesceLocked() {
+		return false
+	}
 	for id := range status.Progress {
 		if roachpb.ReplicaID(id) == r.mu.replicaID {
 			continue


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629#-:-KxhdMwt08_UtDDnvHKN:bnfp4nl)*

petermattis:

---

*[pkg/storage/replica.go, line 3927 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629#-KxhdhlEq22JfDn5JXYQ:-KxhdhlEq22JfDn5JXYR:bs8f73k) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3866648749f06f7239e773efe6ccd236dbd1b1d/pkg/storage/replica.go#L3927)):*
> ```Go
> 	}
> 
> 	if !r.quiesceLocked() {
> ```

I don't think `quiesceLocked()` would ever return false from here because we previously called `shouldQuiesceLocked()`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629)*

a-robinson:

---

*[pkg/storage/replica.go, line 3927 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629#-KxhdhlEq22JfDn5JXYQ:-Kxhed4izF8J2W8K3R8D:b-a3yjpy) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3866648749f06f7239e773efe6ccd236dbd1b1d/pkg/storage/replica.go#L3927)):*

I don't think `quiesceLocked()` would ever return false from here because we previously called `shouldQuiesceLocked()`. 
</blockquote></details>

Yeah, looks like you're right. I'm still going to merge this to protect against future skew between `quiesceLocked` and `shouldQuiesceLocked`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629)*

petermattis:

---

*[pkg/storage/replica.go, line 3927 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629#-KxhdhlEq22JfDn5JXYQ:-KxhfS8LkaYCbr5VGj3g:bn1wz0t) ([raw file](https://github.com/cockroachdb/cockroach/blob/a3866648749f06f7239e773efe6ccd236dbd1b1d/pkg/storage/replica.go#L3927)):*

Yeah, looks like you're right. I'm still going to merge this to protect against future skew between `quiesceLocked` and `shouldQuiesceLocked`.
</blockquote></details>

:thumbsup:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19629)*

Issue #19523, opened 2017-10-25
server: fold the init service into the startup sequence interceptors


tschottdorf:
for the `cockroach init` command, we currently start a grpc server that only supports the `Init` service and, once initialized, open the actual server.

#19493 reworked the way servers initialize their services to allow gossip to be served earlier than other services. It seems that we should be able to avoid the special casing for the init service by registering it with the same server used by anything else, and by introducing a zero-th startup phase: if the server needs an init, allow only the init service, and then switch over to the "only-gossip" phase.

This is fairly self-contained, but the startup code has a lot going on, so I'd call this an good first ambitious issue.

tschottdorf:
Done by @dsaveliev in #19634!

PR #19634, opened 2017-10-30
server: implement the startup sequence interceptors

dsaveliev:
Related to #19523

This is a rough sketch for the startup sequence interceptors.

Main changes:
* **Init** service moved to the common gRPC server.
* `RegisterInitServer` moved to `NewServer` function so `initServer` added to the `Server` struct.
* Server got the states for bootstrap process.
* Implemented a generic interceptor with rules for each server state.
* `noopInitServer` logic moved to the `Bootstrap` method.

Some concerns:
* I didn't use any third-party state machine library for the sake of simplicity.
* I have big doubts about naming.
* I could have missed some subtle logic, obviously.

I would appreciate any thoughts and advice.

### DIFF
--- a/pkg/server/init.go
+++ b/pkg/server/init.go
@@ -15,71 +15,44 @@
 package server
 
 import (
-	"net"
-
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-
-	"github.com/cockroachdb/cockroach/pkg/rpc"
 	"github.com/cockroachdb/cockroach/pkg/server/serverpb"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
-	"github.com/cockroachdb/cockroach/pkg/util/netutil"
-	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
-	"github.com/cockroachdb/cockroach/pkg/util/uuid"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 )
 
-// noopInitServer implements the Init server interface for a cluster that has
-// already been initialized. Its purpose is simply to provide better error
-// messages if someone tries to initialize a cluster that's already been
-// initialized.
-type noopInitServer struct {
-	clusterID func() uuid.UUID
+type semaphore chan struct{}
+
+func newSemaphore() semaphore {
+	ch := make(semaphore, 1)
+	ch <- struct{}{}
+	return ch
 }
 
-func (s *noopInitServer) Bootstrap(
-	_ context.Context, _ *serverpb.BootstrapRequest,
-) (*serverpb.BootstrapResponse, error) {
-	return nil, grpc.Errorf(
-		codes.AlreadyExists, "cluster has already been initialized with ID %s", s.clusterID())
+func (s semaphore) acquire() {
+	<-s
 }
 
-// initListener wraps a net.Listener and turns its Close() method into
-// a no-op. This is used so that the initServer's grpc.Server can be
-// stopped without closing the listener (which it shares with the main
-// grpc.Server).
-type initListener struct {
-	net.Listener
-}
-
-func (initListener) Close() error {
-	return nil
+func (s semaphore) release() {
+	s <- struct{}{}
 }
 
 // initServer manages the temporary init server used during
 // bootstrapping.
 type initServer struct {
 	server       *Server
-	grpc         *grpc.Server
 	bootstrapped chan struct{}
-	mu           struct {
-		syncutil.Mutex
-		awaitDone bool
-	}
+	semaphore
 }
 
 func newInitServer(s *Server) *initServer {
-	return &initServer{server: s, bootstrapped: make(chan struct{})}
-}
-
-func (s *initServer) serve(ctx context.Context, ln net.Listener) {
-	s.grpc = rpc.NewServer(s.server.rpcContext)
-	serverpb.RegisterInitServer(s.grpc, s)
-
-	s.server.stopper.RunWorker(ctx, func(context.Context) {
-		netutil.FatalIfUnexpected(s.grpc.Serve(initListener{ln}))
-	})
+	return &initServer{
+		server:       s,
+		semaphore:    newSemaphore(),
+		bootstrapped: make(chan struct{}),
+	}
 }
 
 func (s *initServer) awaitBootstrap() error {
@@ -89,10 +62,6 @@ func (s *initServer) awaitBootstrap() error {
 	case <-s.server.stopper.ShouldStop():
 		return errors.New("stop called while waiting to bootstrap")
 	}
-	s.mu.Lock()
-	s.mu.awaitDone = true
-	s.mu.Unlock()
-	s.grpc.GracefulStop()
 
 	return nil
 }
@@ -100,13 +69,13 @@ func (s *initServer) awaitBootstrap() error {
 func (s *initServer) Bootstrap(
 	ctx context.Context, request *serverpb.BootstrapRequest,
 ) (response *serverpb.BootstrapResponse, err error) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	if s.mu.awaitDone {
-		return nil, errors.New("bootstrap called after cluster already initialized")
-	}
+	s.semaphore.acquire()
+	defer s.semaphore.release()
 
 	if err := s.server.node.bootstrap(ctx, s.server.engines, s.server.cfg.Settings.Version.BootstrapVersion()); err != nil {
+		if _, ok := err.(*duplicateBootstrapError); ok {
+			return nil, grpc.Errorf(codes.AlreadyExists, err.Error())
+		}
 		log.Error(ctx, "node bootstrap failed: ", err)
 		return nil, err
 	}
--- a/pkg/server/node.go
+++ b/pkg/server/node.go
@@ -245,6 +245,15 @@ func bootstrapCluster(
 	return clusterID, nil
 }
 
+// duplicateBootstrapError is returned by Node.bootstrap when the node is already initialized.
+type duplicateBootstrapError struct {
+	ClusterID uuid.UUID
+}
+
+func (e *duplicateBootstrapError) Error() string {
+	return fmt.Sprintf("cluster has already been initialized with ID %s", e.ClusterID)
+}
+
 // NewNode returns a new instance of Node.
 func NewNode(
 	cfg storage.StoreConfig,
@@ -338,6 +347,9 @@ func (n *Node) initNodeID(ctx context.Context, id roachpb.NodeID) {
 func (n *Node) bootstrap(
 	ctx context.Context, engines []engine.Engine, bootstrapVersion cluster.ClusterVersion,
 ) error {
+	if n.initialBoot || n.ClusterID != (uuid.UUID{}) {
+		return &duplicateBootstrapError{ClusterID: n.ClusterID}
+	}
 	n.initialBoot = true
 	clusterID, err := bootstrapCluster(ctx, n.storeCfg, engines, bootstrapVersion, n.txnMetrics)
 	if err != nil {
--- a/pkg/server/servemode.go
+++ b/pkg/server/servemode.go
@@ -1,0 +1,63 @@
+// Copyright 2014 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package server
+
+import (
+	"sync/atomic"
+
+	"google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+)
+
+// A list of the server states for bootstrap process.
+const (
+	// modeInitializing is intended for server initialization process.
+	// It allows only bootstrap, heartbeat and gossip methods
+	// to prevent calls to potentially uninitialized services.
+	modeInitializing serveMode = iota
+	// modeOperational is intended for completely initialized server
+	// and thus allows all RPC methods.
+	modeOperational
+)
+
+type serveMode int32
+
+// Intercept implements filtering rules for each server state.
+func (s *Server) Intercept() func(string) error {
+	interceptors := map[string]struct{}{
+		"/cockroach.rpc.Heartbeat/Ping":             {},
+		"/cockroach.gossip.Gossip/Gossip":           {},
+		"/cockroach.server.serverpb.Init/Bootstrap": {},
+	}
+	return func(fullName string) error {
+		if s.serveMode.operational() {
+			return nil
+		}
+		if _, allowed := interceptors[fullName]; !allowed {
+			return grpcstatus.Errorf(
+				codes.Unavailable, "node waiting for init; %s not available", fullName,
+			)
+		}
+		return nil
+	}
+}
+
+func (s *serveMode) set(mode serveMode) {
+	atomic.StoreInt32((*int32)(s), int32(mode))
+}
+
+func (s *serveMode) operational() bool {
+	return atomic.LoadInt32((*int32)(s)) == int32(modeOperational)
+}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -40,8 +40,6 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	grpcstatus "google.golang.org/grpc/status"
 
 	"github.com/cockroachdb/cmux"
 	"github.com/cockroachdb/cockroach/pkg/base"
@@ -123,6 +121,7 @@ type Server struct {
 	admin              *adminServer
 	status             *statusServer
 	authentication     *authenticationServer
+	initServer         *initServer
 	tsDB               *ts.DB
 	tsServer           ts.Server
 	raftTransport      *storage.RaftTransport
@@ -134,8 +133,7 @@ type Server struct {
 	engines            Engines
 	internalMemMetrics sql.MemoryMetrics
 	adminMemMetrics    sql.MemoryMetrics
-
-	serveNonGossip int32 // atomically updated
+	serveMode
 }
 
 // NewServer creates a Server from a server.Context.
@@ -158,6 +156,7 @@ func NewServer(cfg Config, stopper *stop.Stopper) (*Server, error) {
 		cfg:      cfg,
 		registry: metric.NewRegistry(),
 	}
+	s.serveMode.set(modeInitializing)
 
 	// If the tracer has a Close function, call it after the server stops.
 	if tr, ok := cfg.AmbientCtx.Tracer.(stop.Closer); ok {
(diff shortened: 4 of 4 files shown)


tschottdorf:
Thanks @dsaveliev, much appreciated! The logic looks good. I had a few comments about naming as you anticipated, though nothing substantial.

---

Reviewed 3 of 3 files at r1.

---

*[pkg/server/interceptors.go, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2r0RhkmO6NNJrTwjl:-Ky2r0RigYImQaRbJAQC:b-7fehrx) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/interceptors.go#L26)):*
> ```Go
> // A list of the server states for bootstrap process.
> const (
> 	GossipMode int32 = iota
> ```

Add a comment on each mode here.

---

*[pkg/server/interceptors.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2raUHxjy8KKqWCZSz:-Ky2raUHxjy8KKqWCZT-:bktwmjp) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/interceptors.go#L32)):*
> ```Go
> 
> // newInterceptor implements filtering rules for each server state.
> func newInterceptor(s *Server) func(string) error {
> ```

I would make the interceptor a struct:

```go
type grpcMode struct {
  mode int32 // accessed atomically
}

func (i *interceptor) Intercept(s string) error {
  // replaces newInterceptor
}

and then make `(Server).grpcMode` of type `grpcMode`.
```


---

*[pkg/server/interceptors.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2rB3eJsX4hCfzufus:-Ky2rB3eJsX4hCfzufut:b-85pwcb) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/interceptors.go#L53)):*
> ```Go
> 		if _, allowed := interceptors[mode][fullName]; !allowed {
> 			return grpcstatus.Errorf(
> 				codes.Unavailable, "node waiting for %s network; %s not available", modeNames[mode], fullName,
> ```

This would print "node waiting for init network", just move the "network" into `GossipMode: "gossip"` above.

---

*[pkg/server/interceptors.go, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2s1vZyMZTJzlISCmQ:-Ky2s1vZyMZTJzlISCmR:btmihey) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/interceptors.go#L60)):*
> ```Go
> }
> 
> func turnOnGossipMode(s *Server) {
> ```

And these would sit on `(*grpcMode)` as well. The file would be better named `grpcmode.go` at that point :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-:-Ky2tMigFFPkGBaj-hNS:bma9kdw)*

tschottdorf:
Two more things:

1. you may want to rebase, we fixed some test flakiness since you opened this PR
2. if you're interested in a somewhat-related follow-up -- there is a small optimization to the testing infrastructure that has to do with the init command [here](https://github.com/cockroachdb/cockroach/issues/19791). No pressure though! 😄

tschottdorf:

---

*[pkg/server/init.go, line 64 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky3zrcWgp7vM_II-ItJ:-Ky3zrcWgp7vM_II-ItK:b-cw13v4) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/init.go#L64)):*
> ```Go
> 	}
> 
> 	s.mu.Lock()
> ```

Actually there's another problem here that predates your changes, which is that it's possible for multiple clients to race into `Bootstrap` with requests. You need to introduce a semaphore here, and while you're at it, it would be nice if the "cluster has already been initialized" message were the only one that clients could receive (i.e. the "bootstrap called after ..." should go away).

Roughly, you put a `chan struct{}` which is populated as `make(chan struct{}, 1)` and filled with a `struct{}` onto `initServer` and then add this to the beginning of `Bootstrap`:

```go
<-s.semaphore
defer func() {
  s.semaphore <- struct{}{}
}()
```

Next, I think you can remove `s.awaitDone`. I think `server.node.bootstrap` already returns a sentinel error when the node has already been bootstrapped (or you can make it so). Then just convert that sentinel error into the same error that is returned above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634)*

dsaveliev:
Thanks for the thorough comments.
I'm sorry for waiting, diving into the details took some time.

Changes:
* I moved the interceptors logic into `grpcMode` struct and added comments.
* Semaphore is added to `initServer`
* Rebase is applied to PR.

There are two things that took me most of the time:

The first one is the sentinel error from `node.bootstrap`.
For the time being there is no such type of error and the way to implement it a bit unclear for me.
The point is that an obvious way to track successful bootstrap is to look at the NodeID/ClusterID in the node. But the node gets these IDs in the `node.start` method and we can not rely on them during the `node.bootstrap` call. 
We can copy [this logic](https://github.com/dsaveliev/cockroach/blob/master/pkg/server/server.go#L555) and check each engine in the node, or we can introduce some **bool** flag to marking node. Both options do not look very elegant so I decided to check `server.CluserID` and `server.currentMode` inside `initServer`.

Another one is an ugly matcher for [postgres listener](https://github.com/cockroachdb/cockroach/pull/19634/files#diff-09856fe9becddf0199651f451409356aL701)
When I started to work on https://github.com/cockroachdb/cockroach/issues/19791 I got a problem with `TestInitModeNone`. This method totally hung test execution at [this point](https://github.com/cockroachdb/cockroach/blob/master/pkg/acceptance/init_test.go#L100).
It seems that the test should be able to establish pg connection, but it should receive an error during the bootstrap phase, and the `initL` listener implements exactly this behaviour.
So, without `initL` listener I was forced to complicate matcher for `pgL` listener. I am still looking for the better solution though.

For https://github.com/cockroachdb/cockroach/issues/19791 I implemented separate grpc server to return the current `grpcMode`, but it felt like an overkill and  I left your recent changes for `DockerCluster.RunInitCommand`.

dsaveliev:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/server/grpcmode.go, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2r0RhkmO6NNJrTwjl:-KypvLzpc-SNZxpz4xDt:bmrywwh) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L26)):*

Add a comment on each mode here.
</blockquote></details>

Comments have been added.

---

*[pkg/server/grpcmode.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2raUHxjy8KKqWCZSz:-KypvuDewCaP5LOr7fKk:bb6un1j) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L32)):*

I would make the interceptor a struct:

```go
type grpcMode struct {
  mode int32 // accessed atomically
}

func (i *interceptor) Intercept(s string) error {
  // replaces newInterceptor
}

and then make `(Server).grpcMode` of type `grpcMode`.
</blockquote></details>

Done

---

*[pkg/server/grpcmode.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2rB3eJsX4hCfzufus:-Kypul9oCdIrZljPMkhk:b-886gsx) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L53)):*

This would print "node waiting for init network", just move the "network" into `GossipMode: "gossip"` above.
</blockquote></details>

Fixed

---

*[pkg/server/grpcmode.go, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2s1vZyMZTJzlISCmQ:-KypuyiL0ML3X6ACezGe:b194unx) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L60)):*

And these would sit on `(*grpcMode)` as well. The file would be better named `grpcmode.go` at that point :-)
</blockquote></details>

Done, `grpcMode` struct was made.

---

*[pkg/server/init.go, line 64 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky3zrcWgp7vM_II-ItJ:-KypwHDEvTK4hmT8EYKV:bb6un1j) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/init.go#L64)):*

Actually there's another problem here that predates your changes, which is that it's possible for multiple clients to race into `Bootstrap` with requests. You need to introduce a semaphore here, and while you're at it, it would be nice if the "cluster has already been initialized" message were the only one that clients could receive (i.e. the "bootstrap called after ..." should go away).

Roughly, you put a `chan struct{}` which is populated as `make(chan struct{}, 1)` and filled with a `struct{}` onto `initServer` and then add this to the beginning of `Bootstrap`:

```go
<-s.semaphore
defer func() {
  s.semaphore <- struct{}{}
}()
```

Next, I think you can remove `s.awaitDone`. I think `server.node.bootstrap` already returns a sentinel error when the node has already been bootstrapped (or you can make it so). Then just convert that sentinel error into the same error that is returned above.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634)*

tschottdorf:
Thanks for your patience! I have a few more comments on how to simplify this now that the picture gets clearer (thanks to your work). Please let me know if that makes sense. Really appreciate your work here!

---

Reviewed 5 of 5 files at r2.

---

*[pkg/server/grpcmode.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2raUHxjy8KKqWCZSz:-Kz0NUhkar2vLC1U8iBn:b-sub928) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L32)):*
<details><summary><i>Previously, dsaveliev (Dmitry Saveliev) wrote…</i></summary><blockquote>

Done
</blockquote></details>

Sorry, my bad:

```go
type grpcMode int32
```

should do it. No need to create a one-element struct. :smile: 

---

*[pkg/server/node.go, line 249 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0Oj6VZzftosg7wms8:-Kz0OfUoi7Zx393y0YVD:b-bpuxnr) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/node.go#L249)):*
> ```Go
> 
> // DuplicateBootstrapError is returned by Node.bootstrap when the node is already initialized.
> type DuplicateBootstrapError struct {
> ```

This can be unexported (rename to `duplicateBootstrapError`).

---

*[pkg/server/server.go, line 688 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0PjKfa00ahCYD55MZ:-Kz0PjKfa00ahCYD55M_:b-eu7df3) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/server.go#L688)):*
> ```Go
> 	// Inject an initialization listener that will intercept all
> 	// connections while the cluster is initializing.
> 	initL := m.Match(func(_ io.Reader) bool {
> ```

I'm probably missing something, but why do you still need `initL`? I thought it would go away (because now the init service is just another service, like gossip, and the grpcmode-interceptor handles access to it)?

---

*[pkg/server/server.go, line 898 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0QE6X0hE7_md5zsYm:-Kz0QE6X0hE7_md5zsYn:b9ek3zx) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/server.go#L898)):*
> ```Go
> 		log.Info(ctx, "no stores bootstrapped and --join flag specified, awaiting init command.")
> 
> 		s.turnOnInitMode()
> ```

This code has a minor problem (that we're sure to hit in tests, however): we allow first gossip and not init, and only then allow init also. This means that there's a period of time during which `init` is disallowed (even though it'll ultimately be allowed).

I think you can fix that using `initServer.semaphore`: it starts out empty, and at this point in the code you "feed" the semaphore. After `awaitBootstrap`, you drain it again. Then, after the `else` block terminates, you feed the semaphore once more (this is just to insure that anyone who snuck in in the meantime gets an error and doesn't get stuck forever).

Additionally, I think you can actually merge `InitMode` and `GossipMode`. The distinction only exists to prevent users from calling `Init` when they're not supposed to. But you've already changed things so that that's harmless (you get the AlreadyExists error). That'll be a nice simplification. Summary: you'll have a GossipInitMode and an EverythingMode (please chose better names than I just did :smile:).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-:-Kz0TUUK5BYJpcBqP17U:bhojk3j)*

dsaveliev:
Thank for the comments and for your patience too!
I made the advised changes and this time added detailed comments to them (I apologize for my literary skill 😄).
I look forward to your feedback.

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/server/grpcmode.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2raUHxjy8KKqWCZSz:-KzQGS-eFCNiQkKkqaB6:b-w8i3i7) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L32)):*

Sorry, my bad:

```go
type grpcMode int32
```

should do it. No need to create a one-element struct. :smile: 
</blockquote></details>

Yes, I had this idea, but in this case it's a bit hard to work with `atomic` functions.  At the moment, I have removed thе `atomic` functions from `grpcmode`, because there is only concurrent reading and I think this is not a big deal (but I have some doubts).

---

*[pkg/server/node.go, line 249 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0Oj6VZzftosg7wms8:-KzQGsMLB3rcVdZ9Q-27:b2ct678) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/node.go#L249)):*

This can be unexported (rename to `duplicateBootstrapError`).
</blockquote></details>

My bad, fixed.

---

*[pkg/server/server.go, line 688 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0PjKfa00ahCYD55MZ:-Kz0WuZTtLtw0Zbb0H28:b-38up1f) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/server.go#L688)):*

I'm probably missing something, but why do you still need `initL`? I thought it would go away (because now the init service is just another service, like gossip, and the grpcmode-interceptor handles access to it)?
</blockquote></details>

There is a large gap between the initialization of `pgL` and the serving of `pgL`, and all 
**pg** requests to the uninitialized cluster will wait for the bootstrap finish. `initL` was introduced to intercept such requests in the `InitMode` and return an error. For example, without `initL` [this line](https://github.com/cockroachdb/cockroach/blob/master/pkg/acceptance/init_test.go#L100) hangs while executing `TestInitModeNone`. Therefore, after removing `initL` I was forced to add the `grpcMode` check into the `pgL` matcher.

---

*[pkg/server/server.go, line 898 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0QE6X0hE7_md5zsYm:-KzQQAipiONeR5O3Wxb3:b3vy56l) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/server.go#L898)):*

This code has a minor problem (that we're sure to hit in tests, however): we allow first gossip and not init, and only then allow init also. This means that there's a period of time during which `init` is disallowed (even though it'll ultimately be allowed).

I think you can fix that using `initServer.semaphore`: it starts out empty, and at this point in the code you "feed" the semaphore. After `awaitBootstrap`, you drain it again. Then, after the `else` block terminates, you feed the semaphore once more (this is just to insure that anyone who snuck in in the meantime gets an error and doesn't get stuck forever).

Additionally, I think you can actually merge `InitMode` and `GossipMode`. The distinction only exists to prevent users from calling `Init` when they're not supposed to. But you've already changed things so that that's harmless (you get the AlreadyExists error). That'll be a nice simplification. Summary: you'll have a GossipInitMode and an EverythingMode (please chose better names than I just did :smile:).
</blockquote></details>

* `InitMode` and `GossipMode` merged together. As is known, naming is the hardest problem so I left `InitMode` name for the result.
* I made advised changes for the semaphore. Furthermore, I introduced auxiliary methods for the semaphore in the hope of making the code more readable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-:-KzQSYXShv0xPWBSq-1o:bti73hi)*

tschottdorf:
@dsaveliev, I made some changes to this branch (look at r3-r4) to avoid the data races, and added a bit of additional commentary. (I took this liberty because, well, you've already done all of the work and going back-and-forth for weeks about cleanup is boring). I will merge this when it turns green, but please do look at the changes I piled on top to make sure I wasn't doing anything silly. Thanks so much for your contribution! If you're looking for something else to work on -- I'll be happy to find something!

---

Reviewed 1 of 6 files at r2, 1 of 5 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/server/init.go, line 72 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-KzyuA983pQUAmUYnrHk:-KzyuA983pQUAmUYnrHl:boz6ky9) ([raw file](https://github.com/cockroachdb/cockroach/blob/61c6785e6a18b2f0341b9c3e3e38b572c6322013/pkg/server/init.go#L72)):*
> ```Go
> 	s.semaphore.acquire()
> 	defer func() {
> 		s.semaphore.release()
> ```

Removed the unnecessary closure.

---

*[pkg/server/server.go, line 688 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Kz0PjKfa00ahCYD55MZ:-KzysqyjqUeha7NwvG03:bo9b6lh) ([raw file](https://github.com/cockroachdb/cockroach/blob/ee34bdf09814f02a8f8208dfce64bd09a5484c82/pkg/server/server.go#L688)):*
<details><summary><i>Previously, dsaveliev (Dmitry Saveliev) wrote…</i></summary><blockquote>

There is a large gap between the initialization of `pgL` and the serving of `pgL`, and all 
**pg** requests to the uninitialized cluster will wait for the bootstrap finish. `initL` was introduced to intercept such requests in the `InitMode` and return an error. For example, without `initL` [this line](https://github.com/cockroachdb/cockroach/blob/master/pkg/acceptance/init_test.go#L100) hangs while executing `TestInitModeNone`. Therefore, after removing `initL` I was forced to add the `grpcMode` check into the `pgL` matcher.
</blockquote></details>

Good point. I solved this by adding a check after `Accept()`. This will make it easier (in some future change) to return a "nice" error to sql clients that connect while the server is in this state.

---

*[pkg/server/grpcmode.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-Ky2raUHxjy8KKqWCZSz:-KzygpaOrak7YRrOrOb6:bhjopp) ([raw file](https://github.com/cockroachdb/cockroach/blob/4b79b5d9eda1f27ad175e55791010cb58fa4fe1a/pkg/server/grpcmode.go#L32)):*
<details><summary><i>Previously, dsaveliev (Dmitry Saveliev) wrote…</i></summary><blockquote>

Yes, I had this idea, but in this case it's a bit hard to work with `atomic` functions.  At the moment, I have removed thе `atomic` functions from `grpcmode`, because there is only concurrent reading and I think this is not a big deal (but I have some doubts).
</blockquote></details>

You'll need atomic access there. I reintroduced it!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-:-Kzzblegdsqn-YSZUoqj:bppynw9)*

dsaveliev:
Thank you for your improvements and patience!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19634#-:-L-18f9E3rZWuycUVTjj:b-xd4pqa)*

PR #19642, opened 2017-10-30
release-1.1: ui: regenerate embedded.go

tbg:


a-robinson (approved):
This is needed due to https://github.com/cockroachdb/cockroach/pull/19626, right?

tschottdorf:
Yeah, I'd have to double check, but I think we maybe changed our CI during the PR to not care whether `embedded.go` has been touched, but then the release build still checks? As far as I see it, my PR should've never turned green in the first place.

PR #19649, opened 2017-10-30
sql: add some logging for distributed import

madelynnblue:


pkg/sql/distsql_physical_planner.go
```diff
@@ -1036,6 +1036,9 @@ func (l *DistLoader) LoadCSV(
 	if err != nil {
 		return err
 	}
+	if log.V(1) {
+		log.Infof(ctx, "import-distsql: begin sampling phase of job %s", job.Record.Description)
```
asubiotto: I would tag the context `log.WithLogTag` to not have to write `import-distsql` every time.

pkg/sql/distsql_physical_planner.go
```diff
@@ -1193,6 +1199,9 @@ func (l *DistLoader) LoadCSV(
 	if recv.err != nil {
 		return recv.err
 	}
+	if log.V(1) {
+		log.Infof(ctx, "import-distsql: finished job %s", job.Record.Description)
```
asubiotto: Consider deferring this at the top to print even on error.

andreimatei:

---

*[pkg/sql/distsql_physical_planner.go, line 1039 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19649#-Kxig-9GTkxor05vTMrp:-Kxig-9GTkxor05vTMrq:by0q4bk) ([raw file](https://github.com/cockroachdb/cockroach/blob/45755bd1c890acb0dd5fbe07b42741e13f25b3f2/pkg/sql/distsql_physical_planner.go#L1039)):*
> ```Go
> 		return err
> 	}
> 	if log.V(1) {
> ```

There's `VEventf(ctx, 1, ...)`.
If you want to gate the Eventf call itself, use `if log.ExpensiveLogEnabled(ctx, 2) {}`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19649)*

a-robinson (commented):
Is this logging expensive enough to even need gating? Three log messages for an entire import job seems pretty insignificant, since imports don't seem like the sort of thing anyone would be doing a lot of at once.

mjibson:

---

*[pkg/sql/distsql_physical_planner.go, line 1039 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19649#-Kxig-9GTkxor05vTMrp:-KxigzckWQ1v3t2VGU7q:b-rmw1b3) ([raw file](https://github.com/cockroachdb/cockroach/blob/45755bd1c890acb0dd5fbe07b42741e13f25b3f2/pkg/sql/distsql_physical_planner.go#L1039)):*

There's `VEventf(ctx, 1, ...)`.
If you want to gate the Eventf call itself, use `if log.ExpensiveLogEnabled(ctx, 2) {}`
</blockquote></details>

TIL. Thanks.

---

*[pkg/sql/distsql_physical_planner.go, line 1040 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19649#-Kxidh2dAcexA7apjvnY-r1-1040:-Kxih-_IFfgmYsNh0oYq:b-rmw1b3) ([raw file](https://github.com/cockroachdb/cockroach/blob/45755bd1c890acb0dd5fbe07b42741e13f25b3f2/pkg/sql/distsql_physical_planner.go#L1040)):*

I would tag the context `log.WithLogTag` to not have to write `import-distsql` every time.
</blockquote></details>

TIL. Thanks.

---

*[pkg/sql/distsql_physical_planner.go, line 1203 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19649#-Kxidh2dAcexA7apjvnY-r1-1203:-Kxih0Voa2YBs9zlwPXJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/45755bd1c890acb0dd5fbe07b42741e13f25b3f2/pkg/sql/distsql_physical_planner.go#L1203)):*

Consider deferring this at the top to print even on error.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19649)*

mjibson:
I have no opinion on the gating. @a-robinson does gating it behind V1 make it too cumbersome to be useful?

a-robinson (approved):
I'm fine running with `vmodule=distsql_physical_planner=1` for my tests, but it's really more up to you and whether you think you'll be able to debug issues on users' clusters without this info. The code LGTM.

mjibson:
I'm ok leaving it as is, and adjusting later if necessary.

PR #19577, opened 2017-10-26
rfc: SELECT FOR UPDATE

rytaft:
RFC regarding support of `SELECT ... FOR UPDATE` SQL syntax.  See [Issue #6583](https://github.com/cockroachdb/cockroach/issues/6583).

Issue #6583, opened 2016-05-09
sql: explicit lock syntax (SELECT FOR {SHARE,UPDATE} {skip locked,nowait})


tbg:
Postgres (and MySQL as well) provide a `FOR UPDATE` clause which instructs a SELECT operation to acquire row-level locks for the data it is reading.

We currently don't support this syntax, and supporting it with the same semantics is more or less impossible anyway with our underlying concurrency model (we have no locks, and while we could put "intents" in place to simulate an intent to write, it wouldn't quite be the same).

We should discuss whether we want to add least accept the syntax so that frameworks which make use of it can at least proceed (even if they don't get the intended semantics, though it's not clear what the implications of that are).

Apparently the [Quartz scheduler](http://www.quartz-scheduler.org) makes use of `SELECT FOR UPDATE`. It is in this context that this came up.

See also http://www.postgresql.org/docs/9.1/static/explicit-locking.html for the explicit locking modes exposed by Postgres.

karthik-gorthi:
I think it is a better idea to only **accept** the syntax without supporting the functionality. 

While this would need a lot of highlighting in the docs, many current RDBMS support this explicit locking mechanism (except MSSQL), so accepting it would relieve a lot of pain in modifying the apps. 

In an open source library like quartz, accepting this without actually supporting means, possible errors at the commit time (due to OCC) which it doesn't handle. 
But, in a custom application which someone looks to migrate to cockroach, handling errors upon commits would / should have been handled at a higher (framework?) level, so it should be fine IMO.

bdarnell:
Note that `FOR UPDATE` is not standard SQL, so a portable application shouldn't rely on it without having an alternative, although it's reasonable to assume that `FOR UPDATE` exists in postgres mode so it would still be good to support it if we can.

We shouldn't support the syntax without providing the semantics that it implies. Fortunately, the locking modifiers are mostly semantically relevant at lower isolation levels; at `SERIALIZABLE` we'd be safe to accept the syntax as no-ops. I'd have to think more about how `SNAPSHOT` and `FOR UPDATE` would interact. 

Even though it's not required for `SERIALIZABLE` semantics, we *might* want to write a dummy intent in `SELECT FOR UPDATE`. This would allow a transaction to assert its timestamp and priority during an initial read-only phase so it can abort other conflicting transactions. I'm not sure whether that's helpful, though, or if it is consistent with expectations from this command in other databases.


[...]

PR #19650, opened 2017-10-30
sql: Support SELECT FOR UPDATE SQL syntax

rytaft:
Update the parser to recognize the `SELECT ... FOR UPDATE` SQL
syntax.  Update the SQL layer so scan nodes in the query plan
are flagged as lockForUpdate if the `SELECT ... FOR UPDATE` syntax
is used.  If a scan node with lockForUpdate is encountered during
plan execution, return an error so the user knows that 
`SELECT ... FOR UPDATE` is not yet fully supported.

This is the first of several commits/PRs needed to support
`SELECT ... FOR UPDATE`.  Future commits will implement the
changes required at the KV layer, connect the two layers with
changes to the KV api, and eliminate the temporary error message
introduced in this commit.

See PR #19577 for the RFC, and Issue #6583.


RaduBerinde:
*[pkg/sql/distsqlrun/processors.proto, line 173 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-Kxipjcccgzk1AAb2anc:-Kxipjcccgzk1AAb2and:bhmuuw9) ([raw file](https://github.com/cockroachdb/cockroach/blob/ccb69040da2ac8f7304c1548c660f5ebe0aa95ee/pkg/sql/distsqlrun/processors.proto#L173)):*
> ```protobuf
>   // (that value will be used for sizing batches instead).
>   optional int64 limit_hint = 5 [(gogoproto.nullable) = false];
>   optional bool lockForUpdate = 6 [(gogoproto.nullable) = false];
> ```

Unfortunately, distsql doesn't currently support write operations - this is because the KV TxnCoordSender assumes all writes happen on the same node. I am not quite sure if this will apply to the new KV operations, but I suspect it will (CC @andreimatei - see RFC #19577).

I would not add any support in distsql at this time and disallow such queries from using distsql (see `DistSQLPlanner.checkSupportForNode`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650)*

rytaft:
Review status: 0 of 23 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/distsqlrun/processors.proto, line 173 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-Kxipjcccgzk1AAb2anc:-Kxj8QU2eQ5MJhdd8BX4:btufykf) ([raw file](https://github.com/cockroachdb/cockroach/blob/ccb69040da2ac8f7304c1548c660f5ebe0aa95ee/pkg/sql/distsqlrun/processors.proto#L173)):*

Unfortunately, distsql doesn't currently support write operations - this is because the KV TxnCoordSender assumes all writes happen on the same node. I am not quite sure if this will apply to the new KV operations, but I suspect it will (CC @andreimatei - see RFC #19577).

I would not add any support in distsql at this time and disallow such queries from using distsql (see `DistSQLPlanner.checkSupportForNode`).
</blockquote></details>

Got it.  I've removed support from distsql and added a check in `DistSQLPlanner.checkSupportForNode`. Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650)*

RaduBerinde:
:lgtm:

---

Reviewed 21 of 26 files at r1, 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-:-KxjAYQ1lPD0Nx6VGBVG:bnfp4nl)*

jordanlewis:
:lgtm: - but have you considered adding the `lockForUpdate` bit to some of the structures that already get passed around in most of the method signatures you're modifying though? I think ideally we could avoid having to plumb this extra bool through everything, especially since undoubtedly as time goes on we will want to add more and more feature flags like this one.

Concretely, I think adding the bool to the `SelectClause` and `planDataSource` structs would probably eliminate most of the method signature changes.

Forgive me if you've already thought this through and decided that it's infeasible!

---

Reviewed 21 of 26 files at r1, 5 of 5 files at r2.

---

*[pkg/sql/check.go, line 156 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-KxjGYEcGLw0SpdC6Wgl:-KxjGYEcGLw0SpdC6Wgm:bkg4j28) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fbeb9e76f9488073966f7ef4d00167aa20997b/pkg/sql/check.go#L156)):*
> ```Go
> 	// would be marginal compared to the work of the actual query, so the added
> 	// complexity seems unjustified.
> 	rows, err := p.SelectClause(ctx, sel, nil /* orderBy */, lim, false, /* lockForUpdate */
> ```

Did you consider adding the `lockForUpdate` bit to the `SelectClause` AST node?

---

*[pkg/sql/data_source.go, line 265 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-KxjGxa0Byflbgy6_K_j:-KxjGxa0Byflbgy6_K_k:b-pnil6x) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fbeb9e76f9488073966f7ef4d00167aa20997b/pkg/sql/data_source.go#L265)):*
> ```Go
> 			return planDataSource{}, err
> 		}
> 		right, err := p.getSources(ctx, sources[1:], scanVisibility, lockForUpdate)
> ```

Would it make sense to add the `lockForUpdate` bit to the `planDataSource` itself? That would eliminate this argument plumbing.

---

*[pkg/sql/plan.go, line 398 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-KxjHEqsOlbfoBtrWwGx:-KxjHEqsOlbfoBtrWwGy:bkydqx6) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fbeb9e76f9488073966f7ef4d00167aa20997b/pkg/sql/plan.go#L398)):*
> ```Go
> 		return p.Select(ctx, n, desiredTypes)
> 	case *parser.SelectClause:
> 		return p.SelectClause(ctx, n, nil /* orderBy */, nil /* limit */, false, /* lockForUpdate */
> ```

Adding the bit to the SelectClause expression would eliminate this threading

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-:-KxjHQEBsAkyOvvH7XIS:b-skoqo8)*

jordanlewis:
Whoops looks like I was a bit late on the uptake here!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-:-KxjIfj07p0RjiQw0GOX:blpgoe7)*

rytaft:
Thanks for the comments!  Unfortunately I don't think it's easy to eliminate the extra argument.  As described below in my responses to the inline comments, the `FOR UPDATE` clause is logically grouped with the `ORDER BY` and `LIMIT` clauses in the Postgres parser, and changing that would likely change the grammar.  I don't think adding `lockForUpdate` to `planDataSource` works either, since it is needed by the `scanNode` during plan execution.  Let me know if you think I'm missing something... thanks!

---

---

*[pkg/sql/check.go, line 156 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-KxjGYEcGLw0SpdC6Wgl:-KxjKJGuHdM_bohpaglN:b-14unas) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fbeb9e76f9488073966f7ef4d00167aa20997b/pkg/sql/check.go#L156)):*

Did you consider adding the `lockForUpdate` bit to the `SelectClause` AST node?
</blockquote></details>

I added it to the `Select` rather than `SelectClause` in order to match the [Postgres yacc file](https://github.com/postgres/postgres/blob/86182b18957b8f9e8045d55b137aeef7c9af9916/src/backend/parser/gram.y#L10834) (`ORDER BY`, `LIMIT`, and `FOR UPDATE` are all handled similarly).  I think it would be difficult to add `lockForUpdate` to the `SelectClause` without changing the grammar...

---

*[pkg/sql/data_source.go, line 265 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-KxjGxa0Byflbgy6_K_j:-KxjpIpSBPIsGECevBbk:bqcv71d) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fbeb9e76f9488073966f7ef4d00167aa20997b/pkg/sql/data_source.go#L265)):*

Would it make sense to add the `lockForUpdate` bit to the `planDataSource` itself? That would eliminate this argument plumbing.
</blockquote></details>

I don't think this would work because the `lockForUpdate` value is needed by the `scanNode` at plan execution.  The `scanNode` won't be able to access the `planDataSource`, since that is higher up in the plan tree.

---

*[pkg/sql/plan.go, line 398 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-KxjHEqsOlbfoBtrWwGx:-KxjrIhtom-SG4D--Rh4:b-i640og) ([raw file](https://github.com/cockroachdb/cockroach/blob/04fbeb9e76f9488073966f7ef4d00167aa20997b/pkg/sql/plan.go#L398)):*

Adding the bit to the SelectClause expression would eliminate this threading
</blockquote></details>

Same comment as above - the `SelectClause` struct is populated in a different section of `sql.y`.  Changing this would depart from the grammar used by Postgres.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-:-KxjrmAbsGuTqUp9Co6P:b-yn8899)*

jordanlewis:
I was suggesting augmenting the `SelectClause` struct outside of the parser just as a way of passing the `lockForUpdate` bit without changing method signatures - but now that I actually survey the code it seems like we're not really doing that anywhere else so it'd be a bit of a break in style.

Thanks for the explanation!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19650#-:-KxkECgFi9Gmx-AXIEcX:btrvpn6)*

PR #19653, opened 2017-10-30
cherrypick-1.1: sqlccl: cleanup failed or canceled RESTORE data

### LINKED ISSUE #17123: sqlccl: tests for BACKUP/RESTORE cancellations

### LINKED ISSUE #19398: sqlccl: restore failed, can't delete leftover data

madelynnblue:
Previously, if a RESTORE failed or was manually canceled, any data
it had committed would have been orphaned, and forever taken up
space. Since there was no table descriptor, nothing would have been
able to see or delete it.

Add optional callbacks in jobs code to allow for specific jobs
to implement an onFail method. Add the job's client.Txn to the
update method and this onFail callback to allow for a job to be
transactionally failed and cleaned up.

Release note: Cleanup partially restored data when a RESTORE fails
or is canceled.

This PR is from #19578 but with all of the settings passing removed,
since 1.1 doesn't have any cluster settings for backup/restore.

Fixes #19398
Fixes #17123

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -812,6 +812,16 @@ func TestBackupRestoreControlJob(t *testing.T) {
 	}(jobs.DefaultAdoptInterval)
 	jobs.DefaultAdoptInterval = 100 * time.Millisecond
 
+	serverArgs := base.TestServerArgs{}
+	// Disable external processing of mutations so that the final check of
+	// crdb_internal.tables is guaranteed to not be cleaned up. Although this
+	// was never observed by a stress test, it is here for safety.
+	serverArgs.Knobs.SQLSchemaChanger = &sql.SchemaChangerTestingKnobs{
+		AsyncExecNotification: func() error {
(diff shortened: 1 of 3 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19653#-:-Kxj3VEQ1dA2ClBblHMv:bnfp4nl)*

PR #19654, opened 2017-10-30
sqlccl: add a function from tabledesc -> spans for zone configs

danhhz:
GenerateSubzoneSpans constructs from a TableDescriptor the entries mapping
zone config spans to subzones for use in the SubzoneSpans field of
config.ZoneConfig. SubzoneSpans controls which splits are created, so only
the spans corresponding to entries in subzones are returned.

Zone configs target indexes and partitions via `subzones`, which are attached
to a table-scoped row in `system.zones`. Each subzone represents one index
(primary or secondary) or one partition (or subpartition) and contains the
usual zone config constraints. They are saved to `system.zones` sparsely
(only when set by a user) and are the most specific entry in the normal
cluster-default/database/table/subzone config hierarchy.

Each non-interleaved index and partition can be mapped to spans in the
keyspace. Indexes and range partitions each map to one span, while each list
partition maps to one or more spans. Each partition span is contained by some
index span and each subpartition span is contained by one of its parent
partition's spans. The spans for a given level of a range partitioning
(corresponding to one `PARTITION BY` in sql or one `PartitionDescriptor`) are
disjoint, but the spans for a given level of a list partitioning may overlap
if DEFAULT is used. A list partitioning which includes both (1, DEFAULT) and
(1, 2) will overlap with the latter getting precedence in the zone config
hierarchy. NB: In a valid PartitionDescriptor, no partitions with the same
number of DEFAULTs will overlap (this property is used by
`indexCoveringsForPartitioning`).

These subzone spans are kept denormalized to the relevant `system.zone` row
for performance. Given a TableDescriptor, the spans for every
index/partition/subpartition are created, filtered out if they don't have a
config set for them, and precedence applied (via `OverlapCoveringMerge`) to
produce a set of non-overlapping spans, which each map to a subzone. There
may be "holes" (uncovered spans) in this set.

The returned spans are returned in exactly the format required by
`system.zones`. They must be sorted and non-overlapping. Each contains an
IndexID, which maps to one of the input `subzones` by indexing into the
slice. As space optimizations, all `Key`s and `EndKey`s of `SubzoneSpan` omit
the common prefix (the encoded table ID) and if `EndKey` is equal to
`Key.PrefixEnd()` it is omitted.

Release Note: none

### DIFF
--- a/pkg/ccl/sqlccl/zone.go
+++ b/pkg/ccl/sqlccl/zone.go
@@ -1,0 +1,258 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed as a CockroachDB Enterprise file under the Cockroach Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+//     https://github.com/cockroachdb/cockroach/blob/master/LICENSE
+
+package sqlccl
+
+import (
+	"bytes"
+
+	"github.com/cockroachdb/cockroach/pkg/keys"
+
+	"github.com/cockroachdb/cockroach/pkg/ccl/utilccl/intervalccl"
+	"github.com/cockroachdb/cockroach/pkg/config"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
+	"github.com/cockroachdb/cockroach/pkg/sql/parser"
+	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/pkg/errors"
+)
+
+// GenerateSubzoneSpans constructs from a TableDescriptor the entries mapping
+// zone config spans to subzones for use in the SubzoneSpans field of
+// config.ZoneConfig. SubzoneSpans controls which splits are created, so only
+// the spans corresponding to entries in subzones are returned.
+//
+// Zone configs target indexes and partitions via `subzones`, which are attached
+// to a table-scoped row in `system.zones`. Each subzone represents one index
+// (primary or secondary) or one partition (or subpartition) and contains the
+// usual zone config constraints. They are saved to `system.zones` sparsely
+// (only when set by a user) and are the most specific entry in the normal
+// cluster-default/database/table/subzone config hierarchy.
+//
+// Each non-interleaved index and partition can be mapped to spans in the
+// keyspace. Indexes and range partitions each map to one span, while each list
+// partition maps to one or more spans. Each partition span is contained by some
+// index span and each subpartition span is contained by one of its parent
+// partition's spans. The spans for a given level of a range partitioning
+// (corresponding to one `PARTITION BY` in sql or one `PartitionDescriptor`) are
+// disjoint, but the spans for a given level of a list partitioning may overlap
+// if DEFAULT is used. A list partitioning which includes both (1, DEFAULT) and
+// (1, 2) will overlap with the latter getting precedence in the zone config
+// hierarchy. NB: In a valid PartitionDescriptor, no partitions with the same
+// number of DEFAULTs will overlap (this property is used by
+// `indexCoveringsForPartitioning`).
+//
+// These subzone spans are kept denormalized to the relevant `system.zone` row
+// for performance. Given a TableDescriptor, the spans for every
+// index/partition/subpartition are created, filtered out if they don't have a
+// config set for them, and precedence applied (via `OverlapCoveringMerge`) to
+// produce a set of non-overlapping spans, which each map to a subzone. There
+// may be "holes" (uncovered spans) in this set.
+//
+// The returned spans are returned in exactly the format required by
+// `system.zones`. They must be sorted and non-overlapping. Each contains an
+// IndexID, which maps to one of the input `subzones` by indexing into the
+// slice. As space optimizations, all `Key`s and `EndKey`s of `SubzoneSpan` omit
+// the common prefix (the encoded table ID) and if `EndKey` is equal to
+// `Key.PrefixEnd()` it is omitted.
+func GenerateSubzoneSpans(
+	tableDesc *sqlbase.TableDescriptor, subzones []config.Subzone,
+) ([]config.SubzoneSpan, error) {
+	a := &sqlbase.DatumAlloc{}
+
+	subzoneIndexByIndexID := make(map[sqlbase.IndexID]int32)
+	subzoneIndexByPartition := make(map[string]int32)
+	for i, subzone := range subzones {
+		if len(subzone.PartitionName) > 0 {
+			subzoneIndexByPartition[subzone.PartitionName] = int32(i)
+		} else {
+			subzoneIndexByIndexID[sqlbase.IndexID(subzone.IndexID)] = int32(i)
+		}
+	}
+
+	var indexCovering intervalccl.Covering
+	var partitionCoverings []intervalccl.Covering
+	if err := tableDesc.ForeachNonDropIndex(func(idxDesc *sqlbase.IndexDescriptor) error {
+		_, indexSubzoneExists := subzoneIndexByIndexID[idxDesc.ID]
+		if indexSubzoneExists {
+			idxSpan := tableDesc.IndexSpan(idxDesc.ID)
+			// Each index starts with a unique prefix, so (from a precedence
+			// perspective) it's safe to append them all together.
+			indexCovering = append(indexCovering, intervalccl.Range{
+				Start: idxSpan.Key, End: idxSpan.EndKey,
+				Payload: config.Subzone{IndexID: uint32(idxDesc.ID)},
+			})
+		}
+
+		var emptyPrefix []parser.Datum
+		indexPartitionCoverings, err := indexCoveringsForPartitioning(
+			a, tableDesc, idxDesc, &idxDesc.Partitioning, subzoneIndexByPartition, emptyPrefix)
+		if err != nil {
+			return err
+		}
+		// The returned indexPartitionCoverings are sorted with highest
+		// precedence first. They all start with the index prefix, so cannot
+		// overlap with the partition coverings for any other index, so (from a
+		// precedence perspective) it's safe to append them all together.
+		partitionCoverings = append(partitionCoverings, indexPartitionCoverings...)
+
+		// TODO(dan): A bunch of this logic is easier if we disallow setting
+		// zone configs on indexes that are interleaved into another index.
+		// InterleavedBy is fine, so using the root of the interleave hierarchy
+		// will work. It is expected that this is sufficient for real-world use
+		// cases. Revisit this restriction if that expectation is wrong.
+		indexUsed := indexSubzoneExists || len(indexPartitionCoverings) > 0
+		if indexUsed && len(idxDesc.Interleave.Ancestors) > 0 {
+			return errors.Errorf("cannot set a zone config for interleaved index %s; "+
+				"set it on the root of the interleaved hierarchy instead", idxDesc.Name)
+		}
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	// OverlapCoveringMerge returns the payloads for any coverings that overlap
+	// in the same order they were input. So, we require that they be ordered
+	// with highest precedence first, so the first payload of each range is the
+	// one we need.
+	ranges := intervalccl.OverlapCoveringMerge(append(partitionCoverings, indexCovering))
+
+	// NB: This assumes that none of the indexes are interleaved, which is
+	// checked above.
+	sharedPrefix := keys.MakeTablePrefix(uint32(tableDesc.ID))
+
+	var subzoneSpans []config.SubzoneSpan
+	for _, r := range ranges {
+		payloads := r.Payload.([]interface{})
+		if len(payloads) == 0 {
+			continue
+		}
+		subzoneSpan := config.SubzoneSpan{
+			Key:    bytes.TrimPrefix(r.Start, sharedPrefix),
+			EndKey: bytes.TrimPrefix(r.End, sharedPrefix),
+		}
+		var ok bool
+		if subzone := payloads[0].(config.Subzone); len(subzone.PartitionName) > 0 {
+			subzoneSpan.SubzoneIndex, ok = subzoneIndexByPartition[subzone.PartitionName]
+		} else {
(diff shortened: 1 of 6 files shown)


benesch:
:raised_hands: 

Going to make time for this tomorrow. Barely scratched the surface tonight.

---

Reviewed 4 of 4 files at r1.
Review status: 4 of 6 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-:-KxpvbeskiqrTTbgKLdA:b-gjzz9b)*

benesch:
Reviewed 4 of 4 files at r1, 2 of 2 files at r2.

---

*[pkg/ccl/sqlccl/zone.go, line 176 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu6cpc96wF1eCwKPu6:-Kxu6cpc96wF1eCwKPu7:b-34q2zz) ([raw file](https://github.com/cockroachdb/cockroach/blob/4df7e00fa7cc2f490655190b0e4121e84839a72a/pkg/ccl/sqlccl/zone.go#L176)):*
> ```Go
> 
> 	if len(partDesc.List) > 0 {
> 		// The returned spans are required to be ordered with hightest
> ```

"hightest"

---

*[pkg/ccl/sqlccl/zone.go, line 225 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu6PNa9uZCpayAbBQd:-Kxu6PNa9uZCpayAbBQe:bp4vxws) ([raw file](https://github.com/cockroachdb/cockroach/blob/4df7e00fa7cc2f490655190b0e4121e84839a72a/pkg/ccl/sqlccl/zone.go#L225)):*
> ```Go
> 					return nil, err
> 				}
> 				colTypes[i] = column.Type.ToDatumType()
> ```

`colTypes` seems unused here too?

---

*[pkg/sql/create.go, line 1143 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu15_rKSJS5XtcpSWo:-Kxu15_siRuk-8onMYHt:b-nv2ai4) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/create.go#L1143)):*
> ```Go
> ) ([]byte, error) {
> 	if len(tuple.Exprs) != len(cols) {
> 		return nil, errors.Errorf("partition has %d columns but %d values were supplied",
> ```

This makes my ears bleed a little bit, but I assume you're doing it for consistency with error messages elsewhere?

---

*[pkg/sql/sqlbase/structured.go, line 1323 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu3FqLaygWe3uMkQel:-Kxu3FqM8sgXleeLXsyd:b-94xuil) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/sqlbase/structured.go#L1323)):*
> ```Go
> 			return nil, nil, err
> 		}
> 		colTypes[i] = column.Type.ToDatumType()
> ```

`colTypes` seems unused?

Also seems like constructing this col map deserves its own helper on `IndexDescriptor`.

---

*[pkg/sql/sqlbase/structured.go, line 1428 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu5k5A2wYuKSs288gg:-Kxu5k5A2wYuKSs288gh:b-fyul69) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/sqlbase/structured.go#L1428)):*
> ```Go
> 				); err != nil {
> 					return err
> 				}
> ```

Huh, it's not at all clear to me why you had to move this into the loop.

---

*[pkg/sql/sqlbase/structured.go, line 1827 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu2D21EOkMENV251ek:-Kxu2D21EOkMENV251el:bmhd1bx) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/sqlbase/structured.go#L1827)):*
> ```Go
> 	if desc.IsPhysicalTable() && desc.PrimaryIndex.Name == name {
> 		return desc.PrimaryIndex, false, nil
> 	}
> ```

:flushed: I'm shocked this worked before.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654)*

danhhz:
Review status: 3 of 6 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/zone.go, line 176 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu6cpc96wF1eCwKPu6:-KxxIMTTIxVuHOLcSohH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4df7e00fa7cc2f490655190b0e4121e84839a72a/pkg/ccl/sqlccl/zone.go#L176)):*

"hightest"
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/zone.go, line 225 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu6PNa9uZCpayAbBQd:-KxxI_KEY_1avAtw2EuP:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4df7e00fa7cc2f490655190b0e4121e84839a72a/pkg/ccl/sqlccl/zone.go#L225)):*

`colTypes` seems unused here too?
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1143 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu15_rKSJS5XtcpSWo:-KxxIdvOv8_hQrbrf8NP:b5p7ho3) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/create.go#L1143)):*

This makes my ears bleed a little bit, but I assume you're doing it for consistency with error messages elsewhere?
</blockquote></details>

yup. i looked into this after this merged and noticed that we don't do "foo(s)" anywhere else

---

*[pkg/sql/sqlbase/structured.go, line 1323 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu3FqLaygWe3uMkQel:-KxxIyIUZuVNER9ruvBD:bjol7yb) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/sqlbase/structured.go#L1323)):*
> colTypes seems unused?

Done

> Also seems like constructing this col map deserves its own helper on IndexDescriptor.

We're only doing a prefix, which is a probably a weird use. I think instead I'd rather define a nil map passed to EncodePartialIndexKey to mean the datums are in the same order as the columns, so we can save the construction (and allocation) of the map. Don't particularly want to add that to this PR though

---

*[pkg/sql/sqlbase/structured.go, line 1428 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu5k5A2wYuKSs288gg:-KxxJbDRCLFqg_PYnVbe:b-uq5spv) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/sqlbase/structured.go#L1428)):*

Huh, it's not at all clear to me why you had to move this into the loop.
</blockquote></details>

lol at BEFORE MERGE leaking here. done

---

*[pkg/sql/sqlbase/structured.go, line 1827 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxu2D21EOkMENV251ek:-KxxNUKrJnXtJZjB44pt:b-8mbbe0) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa498184e01bd22d5637ff217f9c675fc5410986/pkg/sql/sqlbase/structured.go#L1827)):*

:flushed: I'm shocked this worked before.
</blockquote></details>

I briefly glanced at the existing calls and they're all editing indexes, which we don't allow for primary keys

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654)*

benesch:
This mostly LGTM, but need to wrap my head around the overlap covering
merges one last time.

benesch:
:lgtm: modulo one question below

---

Reviewed 3 of 3 files at r3, 2 of 2 files at r4.

---

*[pkg/ccl/sqlccl/zone.go, line 80 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxy0sLRVFesSQZ_nncB:-Kxy0sLRVFesSQZ_nncC:b-y6n139) ([raw file](https://github.com/cockroachdb/cockroach/blob/cea0312b682f2cd40fb085f1386e98b314ce44c1/pkg/ccl/sqlccl/zone.go#L80)):*
> ```Go
> 	var partitionCoverings []intervalccl.Covering
> 	if err := tableDesc.ForeachNonDropIndex(func(idxDesc *sqlbase.IndexDescriptor) error {
> 		_, subzoneIncludesIndex := subzoneIndexByIndexID[idxDesc.ID]
> ```

optional nit: I think this would be better called `indexHasSubzone` or `indexSubzoneExists`. In my mind `subzoneIncludesIndex` is semantically equivalent to `indexUsed` (i.e., whether any subzone, including partition subzones, refers to the index), when it is in fact a subset.

---

*[pkg/ccl/sqlccl/zone.go, line 111 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxy1sat7m5x7XtApwUM:-Kxy1sat7m5x7XtApwUN:b-y2l7tk) ([raw file](https://github.com/cockroachdb/cockroach/blob/cea0312b682f2cd40fb085f1386e98b314ce44c1/pkg/ccl/sqlccl/zone.go#L111)):*
> ```Go
> 		if indexUsed && len(idxDesc.Interleave.Ancestors) > 0 {
> 			return errors.Errorf("cannot set a zone config for interleaved index %s. "+
> 				"set it on the root of the interleaved hierarchy instead", idxDesc.Name)
> ```

optional nit: I _think_ we normally use semicolons in cases like these. (You should check me on that.) There's a zone config error that's something like "cannot set zone config on system config table; try setting it on the "system" database instead", at least.

---

*[pkg/ccl/sqlccl/zone.go, line 247 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxy3JY0znqNzoz_9MIv:-Kxy3JY0znqNzoz_9MIw:b-76mdum) ([raw file](https://github.com/cockroachdb/cockroach/blob/cea0312b682f2cd40fb085f1386e98b314ce44c1/pkg/ccl/sqlccl/zone.go#L247)):*
> ```Go
> 			newPrefixDatums := append(prefixDatums, datums...)
> 			subpartitionCoverings, err := indexCoveringsForPartitioning(
> 				a, tableDesc, idxDesc, &p.Subpartitioning, relevantPartitions, newPrefixDatums)
> ```

Since ranges can't have subpartitions is this codepath ever taken? In fact, can we just remove the `Subpartitioning` field from `message PartitioningDescriptor_Range`? Fine with me to punt that to another commit  but would prefer to drop the recursion (or `panic("unreachable")`) if it's indeed unused.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-:-Kxy5DXpKtrV5AVHIzrR:bljfdn0)*

benesch:
Discussed via Slack. Happy to leave the range subpartioning stuff to a
follow up PR.

danhhz:
Thanks for the reviews! This one was complicated

---

Review status: 5 of 6 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/zone.go, line 80 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxy0sLRVFesSQZ_nncB:-Kxy7HSnE5Q46xXxehCS:bifpk69) ([raw file](https://github.com/cockroachdb/cockroach/blob/cea0312b682f2cd40fb085f1386e98b314ce44c1/pkg/ccl/sqlccl/zone.go#L80)):*

optional nit: I think this would be better called `indexHasSubzone` or `indexSubzoneExists`. In my mind `subzoneIncludesIndex` is semantically equivalent to `indexUsed` (i.e., whether any subzone, including partition subzones, refers to the index), when it is in fact a subset.
</blockquote></details>

went with indexSubzoneExists

---

*[pkg/ccl/sqlccl/zone.go, line 111 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxy1sat7m5x7XtApwUM:-Kxy6h88HCpVzFKazSJj:bq2in8x) ([raw file](https://github.com/cockroachdb/cockroach/blob/cea0312b682f2cd40fb085f1386e98b314ce44c1/pkg/ccl/sqlccl/zone.go#L111)):*

optional nit: I _think_ we normally use semicolons in cases like these. (You should check me on that.) There's a zone config error that's something like "cannot set zone config on system config table; try setting it on the "system" database instead", at least.
</blockquote></details>

good enough for me. the period was bugging me too

---

*[pkg/ccl/sqlccl/zone.go, line 247 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-Kxy3JY0znqNzoz_9MIv:-Kxy6oZgX4brJpoZ9pB5:b-2cn1x) ([raw file](https://github.com/cockroachdb/cockroach/blob/cea0312b682f2cd40fb085f1386e98b314ce44c1/pkg/ccl/sqlccl/zone.go#L247)):*

Since ranges can't have subpartitions is this codepath ever taken? In fact, can we just remove the `Subpartitioning` field from `message PartitioningDescriptor_Range`? Fine with me to punt that to another commit  but would prefer to drop the recursion (or `panic("unreachable")`) if it's indeed unused.
</blockquote></details>

you are correct. i'm working on removing "support" for subpartitioning of ranges right now in a local branch, so i'd rather leave this for that followup

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19654#-:-Kxy7qmy_jkXg7UwLdD2:bcrxhaw)*

Issue #19643, opened 2017-10-30
ui clusterviz: naming things and URLs


couchand:
Currently, the time series pages are titled "Cluster Overview", and are located (along with nodes and events lists) under the path `/cluster`.  Given the upcoming introduction of the cluster viz page, these names are likely to be confusing.  The paths might be fine (if the cluster viz just becomes the root of `/cluster`, say), but there might be value in thinking about them as well.

What names should be applied for:

- the cluster viz
  - page title
  - sidebar link text
  - sidebar link icon
  - path
- the time series pages
  - page titles
  - sidebar link text
  - sidebar link icon
  - paths

dianasaur323:
Hmm... the annoying thing about all this is that we actually do usage tracking based on URLs. Let me check with our metrics platform to see if they have suggestions. Regardless, even if this isn't backwards compatible, we should still fix it to make it make more sense. How about:

- Cluster Viz (remains /cluster -> I would also be okay with /overview)
   - Page Title: Cluster Overview (seems fine to me)
   - Side Bar: Overview (?)
   - Sidebar link icon @kuanluo 
   - Path: /cluster/overview -> I think we can get rid of the all?
      - /cluster/overview/locality1/locality2/nodes
      - /cluster/overview/map/locality1/locality2/nodes
- Time Series (becomes /metrics/all/overview) -> and just keep the paths as we have them
   - Page Title: Cluster Metrics (?)
   - Side Bar: Metrics
   - Side bar link icon @kuanluo 
   - Paths - lets just keep everything as is?

Also, it looks like /cluster/nodes will still be around, and can be navigated to form our new cluster overview page? We might need to iterate on this...

kuanluo:
- Cluster viz -> **Overview** for sure
- Time Series -> Monitor?

couchand:
cc @mrtracy @Amruta-Ranade @vilterp  in case any of you have thoughts on these questions.

Amruta-Ranade:
+1 for **Metrics**

couchand:
For context, here is our current route structure:

```
/ - redirects to /cluster/all/overview
  /cluster - redirects to /cluster/all/overview
    /all/:dashboardName
    /node/:nodeId/:dashboardName
    /nodes
      /:nodeId
        /logs
  /databases - redirects to /databases/tables
    /tables
    /grants
    /database/:databaseName/table/:tableName
  /jobs
  /raft - redirects to /raft/ranges
    /ranges
    /messages/all
[...]

PR #19657, opened 2017-10-30
ui: cluster overview

### LINKED ISSUE #19512: clusterviz: cluster overview
image

- [X] Implement the above view
- ~[ ] Pull top entry of event log as the last event + timestamp (when it started / occurred)~ (moved to #20540)
- ~[ ] Link to full event log~ (moved to #20540)
- [X] Node liveness numbers should reflect the same coloring as our current one - suspect is yellow, dead is red. We can keep live as the standard blue. 
- ~[ ] Node liveness needs to point to the node list - we may need to add a link here @kuanluo~
- [X] Replication status ordering should be Total Ranges (Blue), Under-Replicated Ranges (Yellow), then Unavailable ranges (Red).
- [x] Let's make sure we show something appropriate and consistent for the loading state here
- ~[ ] Let's break up the different shades of grey by 20%. So lightest grey is 20% full, 2nd lightest grey is 40% full, and so on.~
- ~[ ] Support collapsing the overview to get more space for the node list/node canvas.~ (moved to #20540)
- [x] Work on top three browsers (might be hard to capture all issues for now, but let's get the ones that are reported by CockroachDB internal users)
- [x] Make the title consistent in formatting with the other tabs
- ~[ ] Top bar is responsive, where as node list is fixed width - find a more elegant solution for this~ (moved to #20540)

cc @kuanluo @vilterp @couchand @mrtracy @cuongdo 

- [X] For the open source version, create a page with the cluster overview + node list.

couchand:
Part of the initial steps of making a homepage for the admin UI with the cluster viz as the main attraction, here is some work towards a structure for the page and the overview on the top.  Clearly lots of work to do here, but I thought I'd get it up so we could get some eyes on it early.

screen shot 2017-10-30 at 5 28 14 pm

Contributes to #19512.
See #19643 regarding naming things.

### DIFF
--- a/pkg/ui/src/index.tsx
+++ b/pkg/ui/src/index.tsx
@@ -34,6 +34,7 @@ import TableDetails from "src/views/databases/containers/tableDetails";
 
 import JobsPage from "src/views/jobs";
 
+import ClusterOverview from "src/views/cluster/containers/clusterOverview";
 import NodesOverview from "src/views/cluster/containers/nodesOverview";
 import NodeOverview from "src/views/cluster/containers/nodeOverview";
 import NodeGraphs from "src/views/cluster/containers/nodeGraphs";
@@ -67,7 +68,10 @@ ReactDOM.render(
   <Provider store={store}>
     <Router history={history}>
       <Route path="/" component={Layout}>
-        <IndexRedirect to="cluster" />
+        <IndexRedirect to="overview" />
+        <Route path="overview">
+          <IndexRoute component={ClusterOverview} />
+        </Route>
         <Route path="cluster">
           <IndexRedirect to="all/overview" />
           <Route path={`all/:${dashboardNameAttr}`} component={NodeGraphs} />
--- a/pkg/ui/src/redux/nodes.ts
+++ b/pkg/ui/src/redux/nodes.ts
@@ -140,6 +140,8 @@ const nodeSumsSelector = createSelector(
       capacityTotal: 0,
       usedBytes: 0,
       usedMem: 0,
+      totalRanges: 0,
+      underReplicatedRanges: 0,
       unavailableRanges: 0,
       replicas: 0,
     };
@@ -170,6 +172,8 @@ const nodeSumsSelector = createSelector(
           result.capacityTotal += n.metrics[MetricConstants.capacity];
           result.usedBytes += BytesUsed(n);
           result.usedMem += n.metrics[MetricConstants.rss];
+          result.totalRanges += n.metrics[MetricConstants.ranges];
+          result.underReplicatedRanges += n.metrics[MetricConstants.underReplicatedRanges];
           result.unavailableRanges += n.metrics[MetricConstants.unavailableRanges];
           result.replicas += n.metrics[MetricConstants.replicas];
         }
--- a/pkg/ui/src/util/format.ts
+++ b/pkg/ui/src/util/format.ts
@@ -24,11 +24,25 @@ export function ComputePrefixExponent(value: number, prefixMultiple: number, pre
   return prefixScale;
 }
 
-export function BytesToUnitValue(bytes: number): UnitValue {
+/**
+ * ComputeByteScale calculates the appropriate scale factor and unit to use to
+ * display a given byte value, without actually converting the value.
+ *
+ * This is used to prescale byte values before passing them to a d3-axis.
+ */
+export function ComputeByteScale(bytes: number): UnitValue {
   const scale = ComputePrefixExponent(bytes, kibi, byteUnits);
   return {
-    value: bytes / Math.pow(kibi, scale),
+    value: Math.pow(kibi, scale),
     units: byteUnits[scale],
+  };
+}
+
+export function BytesToUnitValue(bytes: number): UnitValue {
+  const scale = ComputeByteScale(bytes);
+  return {
+    value: bytes / scale.value,
+    units: scale.units,
   };
 }
 
--- a/pkg/ui/src/views/app/components/layoutSidebar/index.tsx
+++ b/pkg/ui/src/views/app/components/layoutSidebar/index.tsx
@@ -44,7 +44,8 @@ export default class extends React.Component<{}, {}> {
   render() {
     return <nav className="navigation-bar">
       <ul className="navigation-bar__list">
-        <IconLink to="/cluster" icon={Icons.clusterIcon} title="Cluster" />
+        <IconLink to="/overview" icon={Icons.clusterIcon} title="Overview" />
+        <IconLink to="/cluster" icon={Icons.nodesIcon} title="Cluster" />
         <IconLink to="/databases" icon={Icons.databaseIcon} title="Databases"/>
         <IconLink to="/jobs" icon={Icons.jobsIcon} title="Jobs"/>
       </ul>
--- a/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx
+++ b/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx
@@ -1,0 +1,107 @@
+import d3 from "d3";
+
+import { ComputeByteScale } from "src/util/format";
+
+interface CapacityChartProps {
+  used: number;
+  usable: number;
+}
+
+/**
+ * capacityChart is the small bar chart showing capacity usage displayed on the
+ * cluster summary panel of the cluster overview page.
+ */
+function capacityChart() {
+  const size = {
+    width: 250,
+    height: 20,
+  };
+
+  const margin = {
+    top: 12,
+    right: 35,
+    bottom: 25,
+    left: 20,
+  };
+
+  const TICK_SIZE = 6;
+  const AXIS_MARGIN = 4;
+
+  const scale = d3.scale.linear()
+    .range([0, size.width]);
+
+  const axis = d3.svg.axis()
+    .scale(scale)
+    .tickSize(TICK_SIZE)
+    .ticks(5);
+
+  function recomputeScale(capacity: CapacityChartProps) {
+    // Compute the appropriate scale factor for a value slightly smaller than the
+    // usable capacity, so that if the usable capacity is exactly 1 {MiB,GiB,etc}
+    // we show the scale in the next-smaller unit.
+    const byteScale = ComputeByteScale(capacity.usable - 1);
+
+    const scaled = {
+      used: capacity.used / byteScale.value,
+      usable: capacity.usable / byteScale.value,
+    };
+
+    axis.tickFormat(function (d) {
+      return d + " " + byteScale.units;
+    });
+    scale.domain([0, scaled.usable]);
+
+    return scaled;
+  }
+
+  return function chart(svg: d3.Selection<CapacityChartProps>) {
+    svg
+      .attr("width", size.width + margin.left + margin.right)
+      .attr("height", size.height + margin.top + margin.bottom);
+
+    const el = svg.selectAll(".main")
+      .data((d: CapacityChartProps) => [recomputeScale(d)]);
+
+    el.enter()
+      .append("g")
+      .attr("class", "main")
+      .attr("transform", `translate(${margin.left},${margin.top})`);
+
+    const axisGroup = el.selectAll("g.axis")
+      .data(() => [0]);
+
+    axisGroup.enter()
+      .append("g")
+      .attr("class", "axis")
+      .attr("transform", `translate(0,${size.height + AXIS_MARGIN})`);
+
+    axisGroup.call(axis);
+    axisGroup.selectAll("text").attr("y", AXIS_MARGIN + TICK_SIZE);
+
+    const bg = el.selectAll(".bg")
+      .data((d: CapacityChartProps) => [d]);
+
+    bg.enter()
+      .append("rect")
+      .attr("class", "bg");
+
+    bg
+      .attr("width", size.width)
+      .attr("height", size.height);
+
+    const bar = el.selectAll(".bar")
+      .data((d: CapacityChartProps) => [d]);
+
+    bar.enter()
+      .append("rect")
+      .attr("class", "bar")
+      .attr("height", 10)
+      .attr("y", 5);
+
+    bar
+      .attr("width", (d: CapacityChartProps) => scale(d.used));
+
+  };
+}
+
+export default capacityChart;
--- a/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl
+++ b/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl
@@ -1,0 +1,150 @@
+@require "~styl/base/palette.styl"
+@require "~styl/layout/layout.styl"
+
+.cluster-overview
+  /* TODO(couchand): Update the layout styles to get rid of this hack. */
+  margin-top -1 * $page-padding-top
+  border-bottom 1px solid $table-border-color
+
+  .cluster-ticker
+    padding-top 0
+    padding-bottom 0
+
+  .cluster-summary
+    background-color white
+    padding-top $page-padding-top
+    padding-bottom $page-padding-top
+    padding-left 20px
+    color gray
+    display grid
+    align-items end
+    grid-template-columns 0.8fr 3fr 1fr 1fr 1fr 1fr 0.5fr 1fr 1.2fr 1fr 0.5fr
+    grid-template-rows repeat(3, auto)
+    grid-template-areas "cap-t cap-t . live-t live-t live-t . rep-t rep-t rep-t ." "cap-m cap-c . live-a live-b live-c . rep-a rep-b rep-c ." "cap-a cap-a . live-1 live-2 live-3 . rep-1 rep-2 rep-3 ."
+
+    &__section
+      display contents
+
+      &.capacity-usage
+        .cluster-summary__title
+            grid-area cap-t
+
+        .cluster-summary__chart
+            grid-area cap-c
+
+        .cluster-summary__metric
+            grid-area cap-m
+
+        .cluster-summary__aside
+            grid-area cap-a
+
+      &.node-liveness
+        .cluster-summary__title
+            grid-area live-t
+
+        .cluster-summary__metric.live-nodes
+            grid-area live-a
+
+        .cluster-summary__metric.suspect-nodes
+            grid-area live-b
+
+        .cluster-summary__metric.dead-nodes
+            grid-area live-c
+
+        .cluster-summary__label.live-nodes
+            grid-area live-1
+
+        .cluster-summary__label.suspect-nodes
+            grid-area live-2
+
+        .cluster-summary__label.dead-nodes
+            grid-area live-3
+
+      &.replication-status
+        .cluster-summary__title
+            grid-area rep-t
+
+        .cluster-summary__metric.total-ranges
+            grid-area rep-a
+
+        .cluster-summary__metric.under-replicated-ranges
+            grid-area rep-b
+
+        .cluster-summary__metric.unavailable-ranges
+            grid-area rep-c
+
+        .cluster-summary__label.total-ranges
+            grid-area rep-1
+
+        .cluster-summary__label.under-replicated-ranges
+            grid-area rep-2
+
+        .cluster-summary__label.unavailable-ranges
+            grid-area rep-3
+
+    &__title
+      font-weight 400
+      margin-bottom 20px
+
+    &__metric
+      align-self start
+      font-size 30px
+      line-height 48px
+      color $link-color
+
+
+      &.warning
+        color $warning-color
+
+      &.alert
+        color $alert-color
+
+    &__label
+      font-size 11px
+      line-height 14px
+      padding-top 7px
+      text-transform uppercase
+
+    &__aside
+      .label
+        text-transform uppercase
+        margin-right 14px
+
+      .value
+        color $link-color
+
+      .label, .value
+        font-size 11px
+
+  .capacity-usage .cluster-summary__chart
+    align-self start
+    position relative
+
+    svg
+      position absolute
+      shape-rendering crispEdges
+
+    .axis path
+      fill none
+      stroke none
+
+    .axis line
+      fill none
+      stroke gray
+      shape-rendering crispEdges
+
+    .axis text
+      font-size 12px
+      fill gray
+
+    .axis .tick:nth-child(2n) text
+      display none
+
+    rect
+      shape-rendering crispEdges
+
+      &.bg
+        fill lightgray
+
+      &.bar
+        fill $link-color
--- a/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx
+++ b/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx
@@ -1,0 +1,174 @@
+import d3 from "d3";
+import React from "react";
+import { connect } from "react-redux";
+import { createSelector } from "reselect";
+
+import { nodesSummarySelector, NodesSummary } from "src/redux/nodes";
+import { Bytes as formatBytes } from "src/util/format";
+import { NodesOverview } from "src/views/cluster/containers/nodesOverview";
+import createChartComponent from "src/views/shared/util/d3-react";
+import capacityChart from "./capacity";
+
+import "./cluster.styl";
+
+// tslint:disable-next-line:variable-name
+const CapacityChart = createChartComponent(capacityChart);
+
+class ClusterTicker extends React.Component<{}, {}> {
+  render() {
+    return (
+      <section className="section cluster-ticker">
+        <h2>Cluster Overview</h2>
+      </section>
+    );
+  }
+}
+
+interface CapacityUsageProps {
+  usedCapacity: number;
+  usableCapacity: number;
+}
+
+const formatPercentage = d3.format("0.1%");
+
+class CapacityUsage extends React.Component<CapacityUsageProps, {}> {
+  render() {
+    const { usedCapacity, usableCapacity } = this.props;
+    const usedPercentage = usableCapacity !== 0 ? usedCapacity / usableCapacity : 0;
+    return (
+      <div className="cluster-summary__section capacity-usage">
+        <h3 className="cluster-summary__title">Capacity Usage</h3>
+        <div className="cluster-summary__metric">{ formatPercentage(usedPercentage) }</div>
+        <div className="cluster-summary__chart">
+          <CapacityChart used={usedCapacity} usable={usableCapacity} />
+        </div>
+        <div className="cluster-summary__aside">
+          <span className="label">Current Usage</span>
+          <span className="value">{ formatBytes(usedCapacity) }</span>
+        </div>
+      </div>
+    );
+  }
+}
+
+const mapStateToCapacityUsageProps = createSelector(
+  nodesSummarySelector,
+  function (nodesSummary: NodesSummary) {
+    const { capacityAvailable, capacityUsed } = nodesSummary.nodeSums;
+    const usableCapacity = capacityAvailable + capacityUsed;
+    return {
+      usedCapacity: capacityUsed,
+      usableCapacity: usableCapacity,
+    };
+  },
+);
+
+// tslint:disable-next-line:variable-name
+const CapacityUsageConnected = connect(mapStateToCapacityUsageProps)(CapacityUsage);
+
+interface NodeLivenessProps {
+  liveNodes: number;
+  suspectNodes: number;
+  deadNodes: number;
+}
+
+class NodeLiveness extends React.Component<NodeLivenessProps, {}> {
+  render() {
+    const { liveNodes, suspectNodes, deadNodes } = this.props;
+    return (
+      <div className="cluster-summary__section node-liveness">
+        <h3 className="cluster-summary__title">Node Liveness</h3>
+        <div className="cluster-summary__metric live-nodes">{ liveNodes }</div>
+        <div className="cluster-summary__label live-nodes">Live<br />Nodes</div>
+        <div className={ "cluster-summary__metric suspect-nodes" + (suspectNodes ? " warning" : "") }>{ suspectNodes }</div>
+        <div className="cluster-summary__label suspect-nodes">Suspect<br />Nodes</div>
+        <div className={ "cluster-summary__metric dead-nodes" + (deadNodes ? " alert" : "") }>{ deadNodes }</div>
+        <div className="cluster-summary__label dead-nodes">Dead<br />Nodes</div>
+      </div>
+    );
+  }
+}
+
+const mapStateToNodeLivenessProps = createSelector(
+  nodesSummarySelector,
+  function (nodesSummary: NodesSummary) {
+    const { nodeCounts } = nodesSummary.nodeSums;
+    return {
+      liveNodes: nodeCounts.healthy,
+      suspectNodes: nodeCounts.suspect,
+      deadNodes: nodeCounts.dead,
+    };
+  },
+);
+
+// tslint:disable-next-line:variable-name
+const NodeLivenessConnected = connect(mapStateToNodeLivenessProps)(NodeLiveness);
+
+interface ReplicationStatusProps {
+  totalRanges: number;
+  underReplicatedRanges: number;
+  unavailableRanges: number;
+}
+
+class ReplicationStatus extends React.Component<ReplicationStatusProps, {}> {
+  render() {
+    const { totalRanges, underReplicatedRanges, unavailableRanges } = this.props;
+    return (
+      <div className="cluster-summary__section replication-status">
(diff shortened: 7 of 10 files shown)


couchand:
Just wrapped up the first pass of the bar chart for capacity:

screen shot 2017-10-31 at 2 28 45 pm

Now I'm gonna turn to getting the real data in there.

dianasaur323:
@couchand what kind of feedback do you think would be most helpful? I'm assuming more so from the engineering team than from design and product?

mrtracy:
I had a couple of organization suggestions, mostly looks fine though.

---

Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo3cC5BuuqSQC26UKj:-Kxo3cC5BuuqSQC26UKk:b-3p2qrt) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L23)):*
> ```XML
> 
> const formatPercentage = d3.format("0.1%");
> // TODO(couchand): Actual byte formatter.
> ```

We already have a byte formatting function available under src/util/format

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo2CBbO85K98aMIxfw:-Kxo2CBbO85K98aMIxfx:b-e6avd0) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L107)):*
> ```XML
>     return (
>       <div className="cluster-summary__section">
>         <h3 className="cluster-summary__section--title">Replication Status</h3>
> ```

A few BEM notes:

+ Use of the double-dash (`--`) indicates a modifier class; i.e. a class that is applied to indicate a different mode.  This appears to be trying to make a sub-element.

+ Sub-elements are discouraged.  That is, instead of `cluster-summary__section__title`, use `cluster-summary__title`.

---

*[pkg/ui/styl/pages/cluster.styl, line 9 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo1VwX0StoUgWiHKNW:-Kxo1VwX0StoUgWiHKNX:b-cwpjib) ([raw file](https://github.com/cockroachdb/cockroach/blob/d63168f395ded868b3da26135557bb4a51254e15/pkg/ui/styl/pages/cluster.styl#L9)):*
> ```Stylus
>     width $chart-width
>     &:first-child
>       margin-top 10px
> ```

New styles should go right alongside the code... i.e. this should be in `/views/cluster/containers/clusterOverview/cluster.styl`.  I would like to get rid of these top-level stylus files as much as possible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-:-Kxo4cc2Ee6JhjxKVJLQ:b-rlgax2)*

dianasaur323:
@couchand okay great! agreed that there are still some deviations from the mock ups, but i'll leave that to you for now. i'll take a look at that issue, although I don't have time today or tomorrow.

couchand:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo3cC5BuuqSQC26UKj:-Kxo7Nl_xav4gdhZUwMo:b-azwgla) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L23)):*

We already have a byte formatting function available under src/util/format
</blockquote></details>

Yep!  I've just been pushing up the intermediate commits, look at the latest pls.  Or not yet, since I'm about to send up some more.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo2CBbO85K98aMIxfw:-Kxo7_FtmXbitRe0_4y_:b56r8u0) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L107)):*

A few BEM notes:

+ Use of the double-dash (`--`) indicates a modifier class; i.e. a class that is applied to indicate a different mode.  This appears to be trying to make a sub-element.

+ Sub-elements are discouraged.  That is, instead of `cluster-summary__section__title`, use `cluster-summary__title`.
</blockquote></details>

Ugh at least BEM provides some structure, but it's still crazy trying to apply it to real components.  As far as I can tell it just assumes everything is a two-level structure which, of course, oversimplifies things.  As you can see below, at some point I just stopped bothering, figuring I'd go back and do a pass renaming everything at the end.

---

*[pkg/ui/styl/pages/cluster.styl, line 9 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo1VwX0StoUgWiHKNW:-KxoAghAzBSskCitltbq:b2zhidu) ([raw file](https://github.com/cockroachdb/cockroach/blob/d63168f395ded868b3da26135557bb4a51254e15/pkg/ui/styl/pages/cluster.styl#L9)):*

New styles should go right alongside the code... i.e. this should be in `/views/cluster/containers/clusterOverview/cluster.styl`.  I would like to get rid of these top-level stylus files as much as possible.
</blockquote></details>

Aside: what are your thoughts on using one of the React inline style libraries?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

couchand:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/ui/src/redux/nodes.ts, line 175 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-KxoF69su0JVJlm8jlSY:-KxoF69su0JVJlm8jlSZ:b-fbokcu) ([raw file](https://github.com/cockroachdb/cockroach/blob/a224630df7fa283d8d6750ba282afd6fabfeb23a/pkg/ui/src/redux/nodes.ts#L175)):*
> ```XML
>           result.usedBytes += BytesUsed(n);
>           result.usedMem += n.metrics[MetricConstants.rss];
>           result.totalRanges += n.metrics[MetricConstants.ranges];
> ```

@mrtracy can you confirm if this change is sound?  it really feels wrong, but I just cargo-culted the rest...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

mrtracy:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/ui/src/redux/nodes.ts, line 175 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-KxoF69su0JVJlm8jlSY:-KxoFg0EtHx2GlZMh72y:b-tcdjgc) ([raw file](https://github.com/cockroachdb/cockroach/blob/a224630df7fa283d8d6750ba282afd6fabfeb23a/pkg/ui/src/redux/nodes.ts#L175)):*

@mrtracy can you confirm if this change is sound?  it really feels wrong, but I just cargo-culted the rest...
</blockquote></details>

That looks correct.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo2CBbO85K98aMIxfw:-KxoFl5LQSuS5lhJYAWu:bv4e4c7) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L107)):*

Ugh at least BEM provides some structure, but it's still crazy trying to apply it to real components.  As far as I can tell it just assumes everything is a two-level structure which, of course, oversimplifies things.  As you can see below, at some point I just stopped bothering, figuring I'd go back and do a pass renaming everything at the end.
</blockquote></details>

BEM wants a certain level of complexity in terms of a block; even if elements of a block are nested, each element is treated as a child of the block (and not a child of another element in the block).

If you encounter a situation where the two-level naming is not adequate because of the complicated structure of the block, BEM would prefer breaking that block into multiple blocks (rather than using element-of-element).  I don't think that applies in this situation, though.

---

*[pkg/ui/styl/pages/cluster.styl, line 9 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo1VwX0StoUgWiHKNW:-KxoGMk2U2djZ-L4OaU5:b-6tkxns) ([raw file](https://github.com/cockroachdb/cockroach/blob/d63168f395ded868b3da26135557bb4a51254e15/pkg/ui/styl/pages/cluster.styl#L9)):*

Aside: what are your thoughts on using one of the React inline style libraries?
</blockquote></details>

It's not high on my priority list.  We've gone through two CSS refactorings already, and I don't think we've given this one a chance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

couchand:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/ui/src/redux/nodes.ts, line 175 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-KxoF69su0JVJlm8jlSY:-KxoQ_F58sx0WW9e8e0K:bcq3u31) ([raw file](https://github.com/cockroachdb/cockroach/blob/a224630df7fa283d8d6750ba282afd6fabfeb23a/pkg/ui/src/redux/nodes.ts#L175)):*

That looks correct.
</blockquote></details>

To clarify my concern: if multiple nodes have replicas of the same range, will this not double-count them?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

mrtracy:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/ui/src/redux/nodes.ts, line 175 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-KxoF69su0JVJlm8jlSY:-KxoR_QCPAyU-lNi3WME:b-h8sgux) ([raw file](https://github.com/cockroachdb/cockroach/blob/a224630df7fa283d8d6750ba282afd6fabfeb23a/pkg/ui/src/redux/nodes.ts#L175)):*

To clarify my concern: if multiple nodes have replicas of the same range, will this not double-count them?
</blockquote></details>

For this metric, each node counts only the ranges for which it either 
1.) The leader or 2.) There is no leader, and it is the first node in the rangedescriptor list

The intent is for this to be the total count of ranges when aggregated.  It is possible to double-count due to active membership changes, but this number is accurate enough for monitoring purposes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

couchand:
Thanks for the clarification @mrtracy, that makes sense.

I think this is ready to go (module squashing commits), PTAL, thanks!

screen shot 2017-11-02 at 11 30 34 am

mrtracy:
Review status: 0 of 9 files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 48 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8HFa0H25kuJrsyTxE:-Ky8HFa0H25kuJrsyTxF:b-xdbthe) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L48)):*
> ```XML
>   el.datum(scaled);
> 
>   axis.tickFormat(function (d) {
> ```

You're modifying a package global here.  These functions and variables should be moved into the CapacityChart class.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 70 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8GJQfpLUsD8LqDkKT:-Ky8GJQfpLUsD8LqDkKU:bxwetja) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L70)):*
> ```XML
>   axisGroup.selectAll("text").attr("y", AXIS_MARGIN + TICK_SIZE);
> 
>   const bg = el.selectAll(".bg")
> ```

I think it would be much clearer just to forgo all this enter/exit logic and just add these elements to the JSX; there's always exactly one of each of these elements, d3 can easily select them if they are already present.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 95 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8FJrdNRgxoygywZe5:-Ky8FJrdNRgxoygywZe6:bkqiubt) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L95)):*
> ```XML
> }
> 
> export class CapacityChart extends React.Component<CapacityChartProps, {}> {
> ```

Should have descriptive comments on exported members.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 104 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8FWKpDEGXrD0Hxh0w:-Ky8FWKpDEGXrD0Hxh0x:b8sogo3) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L104)):*
> ```XML
>   }
> 
>   shouldComponentUpdate(props: CapacityChartProps) {
> ```

Why is this being done in shouldComponentUpdate?  I would expect it to be in componentDidUpdate.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl, line 19 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8L_OewhuvBUzSHUG6:-Ky8L_OewhuvBUzSHUG7:bqjyohr) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl#L19)):*
> ```Stylus
>     padding-left 20px
>     color gray
>     display grid
> ```

This is really kind of... odd.  I'm not sure how I feel about it, but it seems to make the CSS *very* closely bound to *exactly* whats currently being displayed.   Would this not have been possible with a flexbox, or without using grid-template-areas?

Maybe it's fine, it just looks different than anything i've seen in CSS before and my immediate instinct is that it's too tightly coupled.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl, line 25 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8IQoq8jtRpBUCecFG:-Ky8IQoq8jtRpBUCecFH:bqw58fk) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl#L25)):*
> ```Stylus
>     grid-template-areas "cap-t cap-t . live-t live-t live-t . rep-t rep-t rep-t ." "cap-m cap-c . live-a live-b live-c . rep-a rep-b rep-c ." "cap-a cap-a . live-1 live-2 live-3 . rep-1 rep-2 rep-3 ."
> 
>     .cluster-summary__section
> ```

Use `&__section` here.   The `&` parent reference operator doesn't just insert the parent's name, it also promotes the child rule to the parent's level.

That is, this currently compiles to:
```css
.cluster-summary .cluster-summary__section{
}
```

But `&__section` would compile to:
```css
.cluster-summary__section
```

Avoiding the nesting of CSS rules is one of the design goals of BEM, to keep the specificity flat.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 82 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8Mo6xrvDRODOaxMWo:-Ky8Mo6xrvDRODOaxMWp:b2v0srw) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L82)):*
> ```XML
>           <span className="value">{ liveNodes }</span>
>         </div>
>         <div className="cluster-summary__label live-nodes">
> ```

The metric/label groups are repeated a lot, seems like a great opportunity just to replace it with a function.

```typescript
function clusterSummaryMetric(label: JSX.Element, value: JSX.Element)
```

Then these groups could look more like:

```typescript
{clusterSummaryMetric(
  <span>Live<br />Nodes</span>,
  suspectNodes, 
  suspectNodes ? "warning" : ""
)}
{clusterSummaryMetric(
  <span>Dead<br />Nodes</span>,
  deadNodes, 
  deadNodes ? "alert" : ""
)}

```

You could even do them as stateless components, although that's up to you:

```typescript
<ClusterSummaryMetric
  label={<span>Dead<br />Nodes</span>}
  value={deadNodes}, 
  valueClassName={deadNodes ? "alert" : ""}
/>
```

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 89 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8P7YHfuzk1UwQYqQo:-Ky8P7YHfuzk1UwQYqQp:b-ey1xw3) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L89)):*
> ```XML
>         </div>
>         <div className="cluster-summary__label suspect-nodes">
>           <span className="label">Suspect<br />Nodes</span>
> ```

I don't think the internal span is necessary, you're in the label div and the span isn't adding any additional information.

Also, is there no way to express this in CS without using the <br /> labels?

---

*[pkg/ui/src/views/cluster/containers/nodesOverview/index.tsx, line 373 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-KxygpKfEvecuw-Q9gJB:-KxygpKfEvecuw-Q9gJC:b-u9e9wb) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/nodesOverview/index.tsx#L373)):*
> ```XML
>  */
> function NodesPage() {
>   return <div>
> ```

Parenthesis around Multi-line JSX 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

couchand:
Review status: 0 of 9 files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 48 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8HFa0H25kuJrsyTxE:-KyGylahk5B8f4ONVlYl:b-7ywrc9) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L48)):*

You're modifying a package global here.  These functions and variables should be moved into the CapacityChart class.
</blockquote></details>

Welp.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 70 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8GJQfpLUsD8LqDkKT:-KyGz1B9gWRApM8DpXDE:b4eoc4y) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L70)):*

I think it would be much clearer just to forgo all this enter/exit logic and just add these elements to the JSX; there's always exactly one of each of these elements, d3 can easily select them if they are already present.
</blockquote></details>

I'll respectfully disagree.  I think the whole chart can be encapsulated in a D3 component without coupling together a React component and a D3 component.  The React integration can be generalized further, per the thread below on `shouldComponentUpdate`, then all the chart's details can be self-contained.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 95 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8FJrdNRgxoygywZe5:-KyHV8C6pXCdh744_icB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L95)):*

Should have descriptive comments on exported members.
</blockquote></details>

Done.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 104 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8FWKpDEGXrD0Hxh0w:-KyGtlt1r9bdZCKFsWqI:bnzonl2) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L104)):*

Why is this being done in shouldComponentUpdate?  I would expect it to be in componentDidUpdate.
</blockquote></details>

Well we'd rather that React not muck about the internals of this component after the initial render, so we want `shouldComponentUpdate` to always return false.  It's returning `undefined` implicitly, which I think is as good, except it does spit out a warning, so I'll update that.  I think the pattern is not as clear when it's all inlined with the implementation of the chart, so I'll pull that out as an HOC to clarify.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl, line 19 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8L_OewhuvBUzSHUG6:-KyGxT-ob6f1VyIv2EO3:b-h8a1jc) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl#L19)):*

This is really kind of... odd.  I'm not sure how I feel about it, but it seems to make the CSS *very* closely bound to *exactly* whats currently being displayed.   Would this not have been possible with a flexbox, or without using grid-template-areas?

Maybe it's fine, it just looks different than anything i've seen in CSS before and my immediate instinct is that it's too tightly coupled.
</blockquote></details>

I did it with flexbox first, but it's really tough to get things to line up precisely in both dimensions.  CSS grid is specifically meant for lining things up in two dimensions, whereas flexbox is intended for layout in one dimension.

I agree that the CSS gets coupled tightly to the contents, that's why I asked about inline styling.  There's not really any getting around it here, since getting the contents to look right demands that each be treated individually.  Putting the styles inline would make that slightly less gross.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl, line 25 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8IQoq8jtRpBUCecFG:-KyGxRTjn0UiQxLjEIKs:b-ap1neb) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl#L25)):*

Use `&__section` here.   The `&` parent reference operator doesn't just insert the parent's name, it also promotes the child rule to the parent's level.

That is, this currently compiles to:
```css
.cluster-summary .cluster-summary__section{
}
```

But `&__section` would compile to:
```css
.cluster-summary__section
```

Avoiding the nesting of CSS rules is one of the design goals of BEM, to keep the specificity flat.
</blockquote></details>

Oh that's slick. Done.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo2CBbO85K98aMIxfw:-KyGy8RvtXy7ZVwBoLT-:bupx35t) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L107)):*

BEM wants a certain level of complexity in terms of a block; even if elements of a block are nested, each element is treated as a child of the block (and not a child of another element in the block).

If you encounter a situation where the two-level naming is not adequate because of the complicated structure of the block, BEM would prefer breaking that block into multiple blocks (rather than using element-of-element).  I don't think that applies in this situation, though.
</blockquote></details>

I don't know, just using your first point as an example, `cluster-summary__title` counds like it's the title for the whole cluster summary (which exists elsewhere), not the title for a particular section.  Maybe it should be `cluster-summary__section-title` as a workaround?  I'm not really worried about it, just trying to illustrate the point.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 82 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8Mo6xrvDRODOaxMWo:-KyGyTrARtFX66blkT68:bsu0sg3) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L82)):*

The metric/label groups are repeated a lot, seems like a great opportunity just to replace it with a function.

```typescript
function clusterSummaryMetric(label: JSX.Element, value: JSX.Element)
```

Then these groups could look more like:

```typescript
{clusterSummaryMetric(
  <span>Live<br />Nodes</span>,
  suspectNodes, 
  suspectNodes ? "warning" : ""
)}
{clusterSummaryMetric(
  <span>Dead<br />Nodes</span>,
  deadNodes, 
  deadNodes ? "alert" : ""
)}

```

You could even do them as stateless components, although that's up to you:

```typescript
<ClusterSummaryMetric
  label={<span>Dead<br />Nodes</span>}
  value={deadNodes}, 
  valueClassName={deadNodes ? "alert" : ""}
/>
```
</blockquote></details>

Good call.  Can't do it as a component yet, that requires React 16.

...

So I tried this, but after removing the internal span, the explicit version is clearer than trying to use a function.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 89 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8P7YHfuzk1UwQYqQo:-KyGyWTmO-NwFRUfmTO_:b-68gxfa) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L89)):*

I don't think the internal span is necessary, you're in the label div and the span isn't adding any additional information.

Also, is there no way to express this in CS without using the <br /> labels?
</blockquote></details>

Yeah, the internal span isn't needed, it's an artifact of some restructuring.  I left it in since that's where the warning/alert class was set, but that can easily be moved up a level.  Done.

---

*[pkg/ui/src/views/cluster/containers/nodesOverview/index.tsx, line 373 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-KxygpKfEvecuw-Q9gJB:-KyHAWY7zl3Gb9SrxyEV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/nodesOverview/index.tsx#L373)):*

Parenthesis around Multi-line JSX 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657)*

mrtracy:
LGTM

---

Review status: 0 of 10 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx, line 104 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8FWKpDEGXrD0Hxh0w:-KyN9qIq3DsTai44itIe:b-m93lsx) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/capacity.tsx#L104)):*

Well we'd rather that React not muck about the internals of this component after the initial render, so we want `shouldComponentUpdate` to always return false.  It's returning `undefined` implicitly, which I think is as good, except it does spit out a warning, so I'll update that.  I think the pattern is not as clear when it's all inlined with the implementation of the chart, so I'll pull that out as an HOC to clarify.
</blockquote></details>

Okay, I understand the reasoning behind this. We can disagree on the use of D3 for now, but *if* we're escaping to D3, this works.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl, line 19 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8L_OewhuvBUzSHUG6:-KyNA-GUHdafVp98_w9-:b-f93gg6) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl#L19)):*

I did it with flexbox first, but it's really tough to get things to line up precisely in both dimensions.  CSS grid is specifically meant for lining things up in two dimensions, whereas flexbox is intended for layout in one dimension.

I agree that the CSS gets coupled tightly to the contents, that's why I asked about inline styling.  There's not really any getting around it here, since getting the contents to look right demands that each be treated individually.  Putting the styles inline would make that slightly less gross.
</blockquote></details>

I see the benefits of grid, but I'm less sold on the use of `grid-template-areas` in this way (where the top-level container is linked to its components rigidly).  It seems like something that might work if there were fewer components, or a more exotic layout, but for me it's breaking down when you have 16 named elements which have their own visual relationships to each other (i.e. the live nodes title appearing on top of the life counts.

This is fine for now, no other components directly depend on it.  However, I would like to address this before the strategy gets copy/pasted to different components.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 107 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Kxo2CBbO85K98aMIxfw:-KyNBL06Z1ZBBgJH4Cbz:bhid0l7) ([raw file](https://github.com/cockroachdb/cockroach/blob/92ffcd7001112c3dcb40ce395120fb1617522e83/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L107)):*

I don't know, just using your first point as an example, `cluster-summary__title` counds like it's the title for the whole cluster summary (which exists elsewhere), not the title for a particular section.  Maybe it should be `cluster-summary__section-title` as a workaround?  I'm not really worried about it, just trying to illustrate the point.
</blockquote></details>

Then the problem is that "cluster-summary" isn't the proper name for this block, because its being confused with the entire "cluster summary" section, even though it doesn't describe the entire section.

However, even in the case where this *did* contain the entire section, the use of "title" for the section title (i.e. "Cluster Overview") and the use of "section-title" for the section titles would be appropriate.  The difference between "section-title" and "section__title" is subtle, but it's an important rule that gives predictable structure. 

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 82 at r25](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-Ky8Mo6xrvDRODOaxMWo:-KyNB6kOafdennRRa1xt:baf7wo7) ([raw file](https://github.com/cockroachdb/cockroach/blob/0ab109da769d6cc23eb50216f90c84b2575c0481/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L82)):*

Good call.  Can't do it as a component yet, that requires React 16.

...

So I tried this, but after removing the internal span, the explicit version is clearer than trying to use a function.
</blockquote></details>

I'll take your word for it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19657#-:-KyNQEjziIrqahOcyQj5:bb6ywmx)*

Issue #19640, opened 2017-10-30
sql: "opening" join filters incorrect in some cases


RaduBerinde:
We have code that converts a post-join filter like `x > 10` into `a.x > 10 AND b.x > 10`. This is not quite correct for left or right outer joins, where one of `a.x, b.x` can be NULL.
```sql
root@:26257/test> show create table a;
+-------+----------------------------------------+
| Table |              CreateTable               |
+-------+----------------------------------------+
| a     | CREATE TABLE a (␤                      |
|       |     x INT NULL,␤                       |
|       |     y INT NULL,␤                       |
|       |     u INT NULL,␤                       |
|       |     FAMILY "primary" (x, y, u, rowid)␤ |
|       | )                                      |
+-------+----------------------------------------+
(1 row)

root@:26257/test> show create table b;
+-------+----------------------------------------+
| Table |              CreateTable               |
+-------+----------------------------------------+
| b     | CREATE TABLE b (␤                      |
|       |     x INT NULL,␤                       |
|       |     y INT NULL,␤                       |
|       |     v INT NULL,␤                       |
|       |     FAMILY "primary" (x, y, v, rowid)␤ |
|       | )                                      |
+-------+----------------------------------------+
(1 row)

root@:26257/test> select *,a.x,b.x from a left outer join b USING (x,y);
+---+---+---+------+---+------+
| x | y | u |  v   | x |  x   |
+---+---+---+------+---+------+
| 1 | 1 | 1 | NULL | 1 | NULL |
| 2 | 2 | 2 |    2 | 2 |    2 |
+---+---+---+------+---+------+
(2 rows)

root@:26257/test> select *,a.x,b.x from a left outer join b USING (x,y) where x is not null;
+---+---+---+---+---+---+
| x | y | u | v | x | x |
+---+---+---+---+---+---+
| 2 | 2 | 2 | 2 | 2 | 2 |
+---+---+---+---+---+---+
(1 row)

root@:26257/test> select *,a.x,b.x from a left outer join b USING (x,y) where x > 0;
+---+---+---+---+---+---+
| x | y | u | v | x | x |
+---+---+---+---+---+---+
| 2 | 2 | 2 | 2 | 2 | 2 |
+---+---+---+---+---+---+
(1 row)
```

RaduBerinde:
CC @knz

knz:
yeah, I messed up when I did the outer join stuff. My bad. Thanks for spotting it.

knz:
[...]

PR #19660, opened 2017-10-30
 sql: remove merged column functionality from join code

RaduBerinde:
Removing the merged columns from the join code. The functionality is instead
provided as a `renderNode` on top of the `joinNode`.

As fallout from this change, we no longer propagate filters from merged columns
to both sides. For inner and left outer joins, they now get propagated only to
the left side. This will be addressed in a subsequent change which transfers
constraints between the two sides. The disabled `openJoinFilter` code is kept
temporarily until that change.

This change fixes the correctness problems described in #19640.

CC @richardwu


jordanlewis:
Haven't reviewed carefully yet, but just calling out that the double renders in some of the explain outputs seem suspicious. Is that correct? Could you explain why in a comment somewhere if so?

---

Reviewed 2 of 2 files at r1, 13 of 13 files at r2.

---

*[pkg/sql/logictest/testdata/logic_test/select_index, line 149 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19660#-KxjK5yU03f7gqN-Alfk:-KxjK5yU03f7gqN-Alfl:b-r9gooo) ([raw file](https://github.com/cockroachdb/cockroach/blob/568c983b93dcf2d3997fa19b85ee5e2081a9b411/pkg/sql/logictest/testdata/logic_test/select_index#L149)):*
> ```
> ----
> 0  render      ·         ·
> 1  render      ·         ·
> ```

This seems mildly suspicious. Is it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19660#-:-KxjKdahIkj7D7muHo5o:b-7m01mj)*

RaduBerinde:
I left a TODO somewhere to coalesce render nodes. It is expected with this change because we add a new renderNode.  It was already possible for certain queries. Note that for distsql, the planning process automatically coalesces projections/renders.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19660#-:-KxjTB1DYjLgOjj6A7vH:b-thc5yx)*

knz:
:lgtm_strong: 
Thank you for picking this up!

---

Reviewed 2 of 2 files at r1, 13 of 13 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19660#-:-Kxm4ROMqWxWaGDj5g4n:bvhfyzm)*

PR #19196, opened 2017-10-11
rfc: SQL Sequences

vilterp:

PR #19664, opened 2017-10-30
sql: basic sequences functionality

vilterp:
Implement `CREATE SEQUENCE` and the `nextval` function, as the first steps of adding sequences support.

The `CREATE SEQUENCE` statement:
- creates and saves a table descriptor with its `sequence_settings` field set. The presence of the `sequence_settings` field on a descriptor signifies that it represents a sequence, not a table or a view.
- initializes the sequence value at the key `/Table/<id>/SequenceValue` as a bare int64.

This PR leaves off the `AS` and `OWNED BY` options of the `CREATE TABLE` statement, to be revisited in later PRs which handle bounds checks and the column <=> sequence dependency relationship, respectively.

The `nextval` function looks up the descriptor by name, calls the `Inc` KV operation on the key, and returns the result.

RFC: #19196 
Tracking issue: #19723 (shows future work)


pkg/sql/parser/eval.go
```diff
@@ -1846,6 +1846,8 @@ type EvalPlanner interface {
 	// QualifyWithDatabase resolves a possibly unqualified table name into a
 	// normalized table name that is qualified by database.
 	QualifyWithDatabase(ctx context.Context, t *NormalizableTableName) (*TableName, error)
+
+	IncrementSequence(context context.Context, seqName string) (int64, error)
```
vilterp: This is a very specific method to have in this interface. Since this is the first function which makes KV writes, not sure what the best way is to plumb things through.

vilterp:
cc @awoods187 @robert-s-lee

justinj:
Is anything going to need to change in `pg_catalog` or `information_schema` for this change (adding Java acceptance tests might help you figure out the answer to this)?

---

Review status: 0 of 27 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/create.go, line 1333 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnZCEN7tM5G64HxXN-:-KxnZCEN7tM5G64HxXN0:b-rmdzhb) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1333)):*
> ```Go
> 
> 	settings := &sqlbase.TableDescriptor_SequenceSettings{}
> 	settings.Cache = 1
> ```

nit: these can be inlined in the struct literal I think?

---

*[pkg/sql/create.go, line 1334 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnZHLT6UKZ8qEjUuMX:-KxnZHLT6UKZ8qEjUuMY:bxl5ho9) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1334)):*
> ```Go
> 	settings := &sqlbase.TableDescriptor_SequenceSettings{}
> 	settings.Cache = 1
> 	settings.Cycle = false
> ```

just to be clear, this will be a no-op (since `false` is the zero value of bool) but I also like the explicitness, I don't particularly care if you set it or not.

---

*[pkg/sql/create.go, line 1343 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnZkLwlRh9qwPemrfV:-KxnZkLwlRh9qwPemrfW:bgcvt28) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1343)):*
> ```Go
> 		case parser.IncrementOption:
> 			settings.Increment = *opt.Increment
> 		default:
> ```

nit: don't need `default:`. this might be clearer as an `if` type check than as a switch but I don't really care too much

---

*[pkg/sql/create.go, line 1831 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnaHv8G7Y4EIxKFeVy:-KxnaHv8G7Y4EIxKFeVz:b-3hrksc) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1831)):*
> ```Go
> 	}
> 
> 	// initialize the sequence value
> ```

nit: capitalize and period

---

*[pkg/sql/create.go, line 1838 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnaREsGkXIbRwzcJLh:-KxnaREsGkXIbRwzcJLi:b-6fxxx7) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1838)):*
> ```Go
> 
> 	b := &client.Batch{}
> 	// TODO(vilterp): setting with Put then incrementing with Inc is not supposed to work, but does.
> ```

someone else will have to help you with this one :D

---

*[pkg/sql/event_log.go, line 59 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnY8YCGQH_-r6hIFVl:-KxnY8YCGQH_-r6hIFVm:b-spv0dv) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/event_log.go#L59)):*
> ```Go
> 	EventLogCreateSequence EventLogType = "create_sequence"
> 	// EventLogDropSequence is recorded when a sequence is dropped.
> 	// EventLogDropSequence EventLogType = "drop_sequence"
> ```

Is this meant to be commented out?

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 3 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnXYb7ODCJ69JB7Z7b:-KxnXYb7ODCJ69JB7Z7c:b-o3cmg8) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/logictest/testdata/logic_test/sequences#L3)):*
> ```
> # LogicTest: default
> 
> # SEQUENCE CREATION
> ```

These tests are great - could you also add some integration tests? For Java at the minimum, since JDBC tends to do some wackier stuff than other drivers. I can help you with this if you want, check out `pkg/acceptance/testdata/java` (also node or even others if you're feeling ambitious, but I think those are all less important than JDBC)

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 8 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnULdSiOOJa-8rRFsE:-KxnULdSiOOJa-8rRFsF:b-y0e6cd) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/logictest/testdata/logic_test/sequences#L8)):*
> ```
> CREATE SEQUENCE foo
> 
> # can't create again
> ```

nit: comments are generally capitalized and have a period (this might be less strictly enforced in logic tests but I think it's a good thing to do)

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 18 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnUCbckS6zs7MxM-pR:-KxnUCbdV9Gtrxx8mr-4:b-rbpsnh) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/logictest/testdata/logic_test/sequences#L18)):*
> ```
> # can't create with 0 increment
> statement error INCREMENT must not be zero
> CREATE SEQUENCE zero_test INCREMENT 0
> ```

We don't do so in many existing logic tests, but I think using `pgerror _____` to assert the pg code in error cases is a good thing to do!

---

*[pkg/sql/parser/create.go, line 857 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnbV49RQ2saYqsDGCF:-KxnbV49RQ2saYqsDGCG:bqiexgd) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/create.go#L857)):*
> ```Go
> 	FormatNode(buf, f, &node.Name)
> 	for _, option := range node.Options {
> 		buf.WriteRune(' ')
> ```

nit: we use `WriteByte` elsewhere in this file, so might as well use it here for consistency

---

*[pkg/sql/parser/create.go, line 863 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxncCeKWk332Ig9llba:-KxncCeKWk332Ig9llbb:b-k6v7x9) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/create.go#L863)):*
> ```Go
> 
> // SequenceOption represents an option on a CREATE SEQUENCE statement.
> type SequenceOption interface {
> ```

I'm not sure these all need to be their own structs, seems like we could maybe get by with just a `SequenceOption` struct that had all these as fields (you can indicate presence/nonpresence by using a pointer)

---

*[pkg/sql/parser/sql.y, line 215 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnUpE2TRnvMGEuK4tx:-KxnUpE3yvY5EGKwQyVV:bdunrmw) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L215)):*
> ```Yacc
> }
> func (u *sqlSymUnion) intVal() *int64 {
> 	return u.val.(*int64)
> ```

Reviewable is showing me some kind of red arrow here that I'm not sure about the implication of, this might have tabs instead of spaces (sql.y is spaces for some reason)?

err, mostly, there are two lines in it with tabs instead of spaces D: we should fix those

---

*[pkg/sql/parser/sql.y, line 709 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWMzmrPdO0qlEw_mL:-KxnWMzmrPdO0qlEw_mM:bf3f5ax) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L709)):*
> ```Yacc
> %type <[]SequenceOption> sequence_option_list opt_sequence_option_list
> %type <SequenceOption> sequence_option_elem
> %type <*int64> just_an_int
> ```

Postgres calls this `NumericOnly` and also accepts floats, will we have any problems if we don't accept floats (and then assert that they're whole numbers)?

If not, is there a reason we need a new nonterminal rather than reusing `signed_iconst`?

---

*[pkg/sql/parser/sql.y, line 3125 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWoYvb7VP6nmiVNC2:-KxnWoYvb7VP6nmiVNC3:bity3te) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L3125)):*
> ```Yacc
> | CACHE just_an_int       { $$.val = CacheOption{Cache: $2.intVal()} }
> | CYCLE                   { $$.val = CycleOption{Cycle: true} }
> | NO CYCLE								{ $$.val = CycleOption{Cycle: false} }
> ```

nit: tabs instead of spaces, we should really lint sql.y

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-:-KxnYghyPXGwXeZDXEEu:bk5bnxv)*

vilterp:
I'm going to add stuff to `pg_catalog` and `information_schema` in a separate PR (just in the spirit of small PRs). Added a Java test that makes a sequence and uses it; it worked without changing those internal tables.

---

Reviewed 5 of 24 files at r1, 2 of 17 files at r4, 20 of 20 files at r8, 1 of 11 files at r11, 2 of 5 files at r14, 1 of 1 files at r15.
Review status: 14 of 28 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

*[pkg/sql/create.go, line 1333 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnZCEN7tM5G64HxXN-:-KxnrpNnElM4-vsDiVFc:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1333)):*

nit: these can be inlined in the struct literal I think?
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1334 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnZHLT6UKZ8qEjUuMX:-KxntyMxGnKS5u3k6Qv0:b-qhbs0n) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1334)):*

just to be clear, this will be a no-op (since `false` is the zero value of bool) but I also like the explicitness, I don't particularly care if you set it or not.
</blockquote></details>

Done. Leaving it in to be explicit (but in a struct literal)

---

*[pkg/sql/create.go, line 1343 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnZkLwlRh9qwPemrfV:-KxnzhW58Btp9op7mNT9:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1343)):*

nit: don't need `default:`. this might be clearer as an `if` type check than as a switch but I don't really care too much
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1831 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnaHv8G7Y4EIxKFeVy:-Kxns2Og8bEAyJKgX0db:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1831)):*

nit: capitalize and period
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1838 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnaREsGkXIbRwzcJLh:-KxnrXOkQt7F4l8wXbt_:bv2rce3) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/create.go#L1838)):*

someone else will have to help you with this one :D
</blockquote></details>

Done. (Realized I could just use `Inc`)

---

*[pkg/sql/event_log.go, line 59 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnY8YCGQH_-r6hIFVl:-KxnzXZktWLkrXbD_aAM:b-snf2a2) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/event_log.go#L59)):*

Is this meant to be commented out?
</blockquote></details>

Done. Will add that in a later PR when I do the DROP command

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 3 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnXYb7ODCJ69JB7Z7b:-Kxnr6w9wwZ27KJVvBV0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/logictest/testdata/logic_test/sequences#L3)):*

These tests are great - could you also add some integration tests? For Java at the minimum, since JDBC tends to do some wackier stuff than other drivers. I can help you with this if you want, check out `pkg/acceptance/testdata/java` (also node or even others if you're feeling ambitious, but I think those are all less important than JDBC)
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 8 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnULdSiOOJa-8rRFsE:-KxnrRdSl7pgpB4iBtbr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/logictest/testdata/logic_test/sequences#L8)):*

nit: comments are generally capitalized and have a period (this might be less strictly enforced in logic tests but I think it's a good thing to do)
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 18 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnUCbckS6zs7MxM-pR:-KxntuP3Ny2EkKBHpxlR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/logictest/testdata/logic_test/sequences#L18)):*

We don't do so in many existing logic tests, but I think using `pgerror _____` to assert the pg code in error cases is a good thing to do!
</blockquote></details>

Done.

---

*[pkg/sql/parser/create.go, line 857 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnbV49RQ2saYqsDGCF:-Kxnr-c_g0e3wW6kXd-o:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/create.go#L857)):*

nit: we use `WriteByte` elsewhere in this file, so might as well use it here for consistency
</blockquote></details>

Done.

---

*[pkg/sql/parser/create.go, line 863 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxncCeKWk332Ig9llba:-Kxnze_cEhjn0LSbEf1t:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/create.go#L863)):*

I'm not sure these all need to be their own structs, seems like we could maybe get by with just a `SequenceOption` struct that had all these as fields (you can indicate presence/nonpresence by using a pointer)
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 215 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnUpE2TRnvMGEuK4tx:-KxnqhUBzl-FlL0u3wZY:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L215)):*

Reviewable is showing me some kind of red arrow here that I'm not sure about the implication of, this might have tabs instead of spaces (sql.y is spaces for some reason)?

err, mostly, there are two lines in it with tabs instead of spaces D: we should fix those
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 3125 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWoYvb7VP6nmiVNC2:-KxnqekkEuyYOVcqriax:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L3125)):*

nit: tabs instead of spaces, we should really lint sql.y
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-:-KxtRR5Y74_gVhutNl5E:bpv25pg)*

justinj:
:lgtm: but @knz should probably take a look as well.

---

Review status: 14 of 28 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/planner.go, line 269 at r15](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxtVwwrrZW0JHWiLu6Q:-KxtVwwrrZW0JHWiLu6R:b-auemqz) ([raw file](https://github.com/cockroachdb/cockroach/blob/c5dc443beed338e31c597b0347f25311b39e6701/pkg/sql/planner.go#L269)):*
> ```Go
> 	}
> 	seqValueKey := keys.MakeSequenceKey(uint32(descriptor.ID))
> 	res, err := client.IncrementValRetryable(
> ```

nit i missed before: you could just return this directly

---

*[pkg/sql/parser/create.go, line 863 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxncCeKWk332Ig9llba:-KxtSuEE_AHX2b4B6mE9:b-ei0txl) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/create.go#L863)):*

Done.
</blockquote></details>

Sorry if I was unclear, I meant that we could have a single struct that gets threaded through like
```
type SequenceOptions struct {
    Cycle bool
    Increment *int
    ...
}
```
etc. but on second thought that might be more difficult than I'm imagining with the grammar, maybe check what another statement does here, it might just be what you had before? Sorry if that's the case!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-:-KxtVT7SRNpsZwaK7BZv:bfv89ra)*

justinj:
Actually, would you mind including the Java test in this PR as well (and then the README change isn't so out-of-place :D)?

---

Review status: 14 of 28 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-:-KxtYkJ9SMDRx92PYo-L:b-8p7cur)*

vilterp:
Reviewed 2 of 12 files at r9, 4 of 11 files at r11, 1 of 4 files at r13, 7 of 8 files at r16, 3 of 3 files at r17.

---

*[pkg/sql/planner.go, line 269 at r15](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxtVwwrrZW0JHWiLu6Q:-KxtbhllxfwkZKIv3aC7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c5dc443beed338e31c597b0347f25311b39e6701/pkg/sql/planner.go#L269)):*

nit i missed before: you could just return this directly
</blockquote></details>

Done.

---

*[pkg/sql/parser/create.go, line 863 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxncCeKWk332Ig9llba:-Kxtbk-082dRWhWru1vJ:b-iwq7um) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/create.go#L863)):*

Sorry if I was unclear, I meant that we could have a single struct that gets threaded through like
```
type SequenceOptions struct {
    Cycle bool
    Increment *int
    ...
}
```
etc. but on second thought that might be more difficult than I'm imagining with the grammar, maybe check what another statement does here, it might just be what you had before? Sorry if that's the case!
</blockquote></details>

The current implementation (`SequenceOption` with `IntVal` and `BoolVal`) isn't too awkward — mimics the PG grammar, and some other rules that thread through a slice. Maybe I could thread that struct through in a similar way; will give it a try.

---

*[pkg/sql/parser/sql.y, line 709 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWMzmrPdO0qlEw_mL:-KxtbzuES3pLRs_GDvUj:b-yqbmb3) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L709)):*

Postgres calls this `NumericOnly` and also accepts floats, will we have any problems if we don't accept floats (and then assert that they're whole numbers)?

If not, is there a reason we need a new nonterminal rather than reusing `signed_iconst`?
</blockquote></details>

Renamed it to `iconst_no_err` and added an explanatory comment. @knz, is there a better way to factor this?

---

*[pkg/sql/sqlbase/structured.go, line 325 at r17](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxtdqKI-rBPXWgMjPw4:-KxtdqKI-rBPXWgMjPw5:bkcziul) ([raw file](https://github.com/cockroachdb/cockroach/blob/7762037ef6ad3ad3c29a12a9aaa28bd04380cb09/pkg/sql/sqlbase/structured.go#L325)):*
> ```Go
> 
> // KindName returns what kind of database object this descriptor describes.
> func (desc *TableDescriptor) KindName() string {
> ```

Wasn't sure what to call this, so I said `Kind`. (`TypeName` was already taken, and always returns "relation"). Does this name seem reasonable? Do we already have a term for "sequence"/"table"/"view"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664)*

jordanlewis:
Reviewed 5 of 24 files at r1, 2 of 17 files at r4, 3 of 20 files at r8, 20 of 20 files at r18, 12 of 12 files at r19.
Review status: 11 of 33 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[pkg/keys/keys.go, line 619 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyxG_1hGkC2_FabIEX:-KxyxG_1hGkC2_FabIEY:b-cj2dqe) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/keys/keys.go#L619)):*
> ```Go
> 
> // MakeSequenceKey returns the key used to store the value of a sequence.
> func MakeSequenceKey(tableID uint32) []byte {
> ```

Kinda silly, but this probably should be tested somehow if possible. Changing this function later should cause tests to fail, as it would represent a backward incompatible change to sequences (old sequences would become inaccessible).

---

*[pkg/sql/planner.go, line 266 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxysNQresqUS9nRjYJY:-KxysNQresqUS9nRjYJZ:b1ctt8a) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/planner.go#L266)):*
> ```Go
> 	}
> 	if descriptor.SequenceSettings == nil {
> 		return 0, pgerror.NewErrorf(pgerror.CodeWrongObjectTypeError, `"%s" is not a sequence`, seqName)
> ```

Please add a logic test for this if you can - wouldn't want it to regress later.

---

*[pkg/sql/update.go, line 56 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Kxywy4RKdDM_8VUbSE-:-Kxywy4SfQ_jYZTN-fqn:b-ek6v0g) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/update.go#L56)):*
> ```Go
> 	}
> 	// We don't support update on views or sequences, only real tables.
> 	tableDesc.TypeName()
> ```

this seems like a stray line :)

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 29 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxysFHjsK1Il2U2FZL0:-KxysFHjsK1Il2U2FZL1:bp6hbn6) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L29)):*
> ```
> # DML ERRORS
> 
> statement error pgcode XX000 cannot run INSERT on sequence "foo" - sequences are not updateable
> ```

You should find a more appropriate pgerror code for this one.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 44 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxysajJHQcpQKbwae_T:-KxysajJHQcpQKbwae_U:b-vuoe8i) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L44)):*
> ```
> 
> # `select * from foo` currently returns 0 columns & 0 rows;
> # testing against that is a Test-Script syntax error.
> ```

What do you mean by Test-Script syntax error? I think you can put `statement ok` on top of it and it'll work, maybe?

---

*[pkg/sql/parser/eval.go, line 1850 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxjO2K_E2ZXeViTjfhO-r3-1850:-Kxyw7COZf35Dz_JlKW7:b29hzcn) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3058c0e0163c77572b9eb185187cfa5b9bad235/pkg/sql/parser/eval.go#L1850)):*

This is a very specific method to have in this interface. Since this is the first builtin function which makes KV writes, not sure what the best way is to plumb things through.
</blockquote></details>

I'm surprised the linter isn't failing this - but exported methods should have comments.

---

*[pkg/sql/parser/sql.y, line 3174 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyuY63yz_R8CUJOuhG:-KxyuY63yz_R8CUJOuhH:b-lyaspk) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/sql.y#L3174)):*
> ```Yacc
>     val, err := $1.numVal().AsInt64()
>     if err != nil {
>       sqllex.Error(err.Error()); return 1
> ```

This seems like a useful rule! Could you add some tests for it in `parse_test`? There is a section for examples that are supposed to return errors, I believe.

---

*[pkg/sql/sqlbase/structured.go, line 325 at r17](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxtdqKI-rBPXWgMjPw4:-KxyvvgC4IdKsuk6Nllr:b-shfzzw) ([raw file](https://github.com/cockroachdb/cockroach/blob/7762037ef6ad3ad3c29a12a9aaa28bd04380cb09/pkg/sql/sqlbase/structured.go#L325)):*

Wasn't sure what to call this, so I said `Kind`. (`TypeName` was already taken, and always returns "relation"). Does this name seem reasonable? Do we already have a term for "sequence"/"table"/"view"?
</blockquote></details>

Kind seems reasonable enough to me, but maybe just call it `Kind` instead of `KindName`? We don't have a type named `kind` so it's not ambiguous.

---

*[pkg/sql/sqlbase/structured.go, line 336 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyvGkS8U7S-VBnvlII:-KxyvGkS8U7S-VBnvlIJ:bpl859y) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.go#L336)):*
> ```Go
> 
> // KindNamePlural returns the plural form of the kind of database object this descriptor describes.
> func (desc *TableDescriptor) KindNamePlural() string {
> ```

This method seems kind of unnecessary to me, since the plural is always formed by adding `s` to the end, which you can do in the corresponding format string. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664)*

jordanlewis:
This is generally looking really good! Nice work.

petermattis:
Review status: 11 of 33 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[pkg/keys/keys.go, line 620 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1RglX3p2vGlxz3QFM:-Ky1RglX3p2vGlxz3QFN:borcjt7) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/keys/keys.go#L620)):*
> ```Go
> // MakeSequenceKey returns the key used to store the value of a sequence.
> func MakeSequenceKey(tableID uint32) []byte {
> 	return makeKey(MakeTablePrefix(tableID), roachpb.RKey("SequenceValue"))
> ```

You need to consider how these keys will be formatted by the pretty-printer. Because you're using the table prefix, I'm suspicious this will be garbled because the second component is not an index ID. Worthwhile to add a test for how sequence keys pretty-print. 

Also, sort of a small item, but we usually use shorter constants and we use lowercase. Perhaps `s/SequenceValue/seqval/g`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664)*

knz:
I'm mostly good with this too.
Just one error remaining in the name resolution, and some missing tests.

---

Reviewed 2 of 24 files at r1, 13 of 20 files at r18, 5 of 11 files at r21, 1 of 4 files at r23, 2 of 5 files at r24, 4 of 7 files at r25, 3 of 3 files at r26, 5 of 5 files at r27, 2 of 2 files at r28.

---

*[pkg/sql/planner.go, line 256 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1bXFodDKpVqZWFkbC:-Ky1bXFpRjmFg5jfIJwa:bxjdr1i) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/sql/planner.go#L256)):*
> ```Go
> 
> func (p *planner) IncrementSequence(
> 	ctx context.Context, seqName string,
> ```

`seqName` must be a `*parser.TableName`, not `string`.

---

*[pkg/sql/planner.go, line 258 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1bbD8FJ6crTR7HzHt:-Ky1bbD8FJ6crTR7HzHu:b-a61van) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/sql/planner.go#L258)):*
> ```Go
> 	ctx context.Context, seqName string,
> ) (int64, error) {
> 	tableName, err := parser.ParseTableName(seqName)
> ```

call `ParseTableName` in the builtin definition for `nextval`, not here.

---

*[pkg/sql/planner.go, line 262 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1bfEQnD-0vOFF6znm:-Ky1bfEQnD-0vOFF6znn:bq4b3q) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/sql/planner.go#L262)):*
> ```Go
> 		return 0, err
> 	}
> 	tableName.DatabaseName = parser.Name(p.session.Database)
> ```

That doesn't seem right. As per the pg doc (https://www.postgresql.org/docs/10/static/functions-sequence.html), the sequence name is determined using the search path.
Use `searchAndQualifyDatabase` instead.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 84 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1dSqz0r_spbeRojL6:-Ky1dSqz0r_spbeRojL7:bnn670q) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L84)):*
> ```
> SELECT nextval('baz')
> ----
> 7
> ```

Add tests that check that sequences can be found in other databases via the search path.
Also add at least one test that checks sequence names can use special characters.

---

*[pkg/sql/parser/builtins.go, line 1148 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1ctXhUZztEtRWnYBO:-Ky1ctXhUZztEtRWnYBP:b-mhumyx) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/builtins.go#L1148)):*
> ```Go
> 			fn: func(evalCtx *EvalContext, args Datums) (Datum, error) {
> 				seqName := MustBeDString(args[0])
> 				res, err := evalCtx.Planner.IncrementSequence(evalCtx.Ctx(), string(seqName))
> ```

as I explain in a previous comment, the translation from string to TableName must occur here, not in IncrementSequence.

---

*[pkg/sql/parser/parse_test.go, line 224 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1d5kndeGd6EePkCbc:-Ky1d5kou5UYJMVtonmP:bk0a21) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/parse_test.go#L224)):*
> ```Go
> 		{`CREATE VIEW a AS TABLE b`},
> 
> 		{`CREATE SEQUENCE a`},
> ```

Some `IF NOT EXISTS` test is missing here.

---

*[pkg/sql/parser/sql.y, line 709 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWMzmrPdO0qlEw_mL:-Ky1cYEfK5fFUKhnPel5:bcgjugz) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L709)):*

Renamed it to `iconst_no_err` and added an explanatory comment. @knz, is there a better way to factor this?
</blockquote></details>

I don't understand the question; what is the problem to be solved here?

Also I'd use `%type <int64> iconst_no_err` - no need for a pointer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-:-Ky1dvJRkVlkkEs7mm_y:b-h0ifea)*

nvanbenschoten:
Reviewed 4 of 24 files at r1, 2 of 17 files at r4, 3 of 20 files at r8, 1 of 20 files at r18, 2 of 12 files at r19, 4 of 11 files at r21, 1 of 4 files at r23, 1 of 5 files at r24, 4 of 7 files at r25, 3 of 3 files at r26, 5 of 5 files at r27, 2 of 2 files at r28.
Review status: all files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[pkg/keys/keys.go, line 620 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1RglX3p2vGlxz3QFM:-KyHnt0QVIyyM2firkid:b-phy28b) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/keys/keys.go#L620)):*

You need to consider how these keys will be formatted by the pretty-printer. Because you're using the table prefix, I'm suspicious this will be garbled because the second component is not an index ID. Worthwhile to add a test for how sequence keys pretty-print. 

Also, sort of a small item, but we usually use shorter constants and we use lowercase. Perhaps `s/SequenceValue/seqval/g`. 
</blockquote></details>

`"seqval"` should also be pulled into `constants.go`

---

*[pkg/sql/create.go, line 1389 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHtZNI19qzl4XTH8RM:-KyHtZNI19qzl4XTH8RN:b-reubm9) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/create.go#L1389)):*
> ```Go
> 			settings.Cache = *option.IntVal
> 		case parser.SeqOptCycle:
> 			settings.Cycle = option.BoolVal
> ```

Should we be throwing errors for the options that are not yet supported?

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 1 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHoUVECDKEy8KCWGe3:-KyHoUVECDKEy8KCWGe4:b-bmum0a) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L1)):*
> ```
> # LogicTest: default
> ```

Why only the `default` config?

---

*[pkg/sql/parser/sql.y, line 3149 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHqJAKk278f8plnriR:-KyHqJAKk278f8plnriS:blr8axn) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/sql.y#L3149)):*
> ```Yacc
> opt_sequence_option_list:
>   sequence_option_list
> | /* EMPTY */          { $$.val = []SequenceOption{} }
> ```

`s/[]SequenceOption{}/[]SequenceOption(nil)/`

---

*[pkg/sql/sqlbase/structured.go, line 326 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHpKg4iknm4Znv6O8j:-KyHpKg4iknm4Znv6O8k:bijc29v) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.go#L326)):*
> ```Go
> // KindName returns what kind of database object this descriptor describes.
> func (desc *TableDescriptor) KindName() string {
> 	if desc.IsTable() {
> ```

nit: I'd do something like:

```
switch {
case desc.IsTable():
    return ...
case desc.IsView():
    ...
case desc.IsSequence():
    ...
default:
    panic("what is going on")
}

to catch any logic errors.
```


---

*[pkg/sql/sqlbase/structured.go, line 336 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyvGkS8U7S-VBnvlII:-KyHpIW4dQd2L8yRdbVs:b32v57) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.go#L336)):*

This method seems kind of unnecessary to me, since the plural is always formed by adding `s` to the end, which you can do in the corresponding format string. 
</blockquote></details>

+1

---

*[pkg/sql/sqlbase/structured.proto, line 611 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHsOiUFnDVky7mhnB3:-KyHsOiUFnDVky7mhnB4:b95kvfg) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.proto#L611)):*
> ```protobuf
>   repeated MutationJob mutationJobs = 27 [(gogoproto.nullable) = false];
> 
>   message SequenceSettings {
> ```

These fields should be commented.

I'd also push to rename this to `SequenceOpts` or `SequenceDescriptor`, as the word "settings" has a specific meaning in Cockroach (session settings, cluster settings, etc.) and has not traditionally been used in this way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664)*

vilterp:
Reviewed 20 of 20 files at r18, 1 of 12 files at r19, 1 of 5 files at r27, 30 of 30 files at r29.
Review status: 9 of 37 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[pkg/keys/keys.go, line 620 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1RglX3p2vGlxz3QFM:-KyHz4m7w2h1lgvVUcDy:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/keys/keys.go#L620)):*

`"seqval"` should also be pulled into `constants.go`
</blockquote></details>

Done.

---

*[pkg/keys/keys.go, line 619 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyxG_1hGkC2_FabIEX:-KyIE5vf6g2Q1-lfG0Yu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/keys/keys.go#L619)):*

Kinda silly, but this probably should be tested somehow if possible. Changing this function later should cause tests to fail, as it would represent a backward incompatible change to sequences (old sequences would become inaccessible).
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1389 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHtZNI19qzl4XTH8RM:-KySnhWCznSYWzT7h7BH:b-zel4sm) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/create.go#L1389)):*

Should we be throwing errors for the options that are not yet supported?
</blockquote></details>

Done. The parser now throws errors for the options we don't support

---

*[pkg/sql/planner.go, line 256 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1bXFodDKpVqZWFkbC:-KySnnEsxY6pnCbtMRq7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/sql/planner.go#L256)):*

`seqName` must be a `*parser.TableName`, not `string`.
</blockquote></details>

Done.

---

*[pkg/sql/planner.go, line 258 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1bbD8FJ6crTR7HzHt:-KyIphVkKX7_D3-s0Qj2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/sql/planner.go#L258)):*

call `ParseTableName` in the builtin definition for `nextval`, not here.
</blockquote></details>

Done.

---

*[pkg/sql/planner.go, line 262 at r18](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1bfEQnD-0vOFF6znm:-KyRXhskNF6t32PLhRqQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/adfd9e7116196eb2e001025ab22ad1a179fa0075/pkg/sql/planner.go#L262)):*

That doesn't seem right. As per the pg doc (https://www.postgresql.org/docs/10/static/functions-sequence.html), the sequence name is determined using the search path.
Use `searchAndQualifyDatabase` instead.
</blockquote></details>

Done.

---

*[pkg/sql/planner.go, line 266 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxysNQresqUS9nRjYJY:-KyIMeB9SLnCbLDj2lgO:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/planner.go#L266)):*

Please add a logic test for this if you can - wouldn't want it to regress later.
</blockquote></details>

Done.

---

*[pkg/sql/update.go, line 56 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Kxywy4RKdDM_8VUbSE-:-KyRXkW4BI8XhVaManWe:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/update.go#L56)):*

this seems like a stray line :)
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 1 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHoUVECDKEy8KCWGe3:-KyRXv986ok9D7-b1H9D:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L1)):*

Why only the `default` config?
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 29 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxysFHjsK1Il2U2FZL0:-KyIVt-q-w0a6zImapye:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L29)):*

You should find a more appropriate pgerror code for this one.
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 44 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxysajJHQcpQKbwae_T:-KyRXn2iNTKuwbQopTCH:bmjwrdm) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L44)):*

What do you mean by Test-Script syntax error? I think you can put `statement ok` on top of it and it'll work, maybe?
</blockquote></details>

Done. Changed this to return an error until we actually support it.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 84 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1dSqz0r_spbeRojL6:-KyIblKxVpzrC_HoUgvo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/logictest/testdata/logic_test/sequences#L84)):*

Add tests that check that sequences can be found in other databases via the search path.
Also add at least one test that checks sequence names can use special characters.
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 1148 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1ctXhUZztEtRWnYBO:-KyRXxI9tCC4y2JrCRpS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/builtins.go#L1148)):*

as I explain in a previous comment, the translation from string to TableName must occur here, not in IncrementSequence.
</blockquote></details>

Done.

---

*[pkg/sql/parser/eval.go, line 1850 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxjO2K_E2ZXeViTjfhO-r3-1850:-KyId1xYkKlfCNxRq6et:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d3058c0e0163c77572b9eb185187cfa5b9bad235/pkg/sql/parser/eval.go#L1850)):*

I'm surprised the linter isn't failing this - but exported methods should have comments.
</blockquote></details>

Done.

---

*[pkg/sql/parser/parse_test.go, line 224 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1d5kndeGd6EePkCbc:-KySzNSMRiaVh9DW7b1J:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/parse_test.go#L224)):*

Some `IF NOT EXISTS` test is missing here.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 709 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWMzmrPdO0qlEw_mL:-KySzyWsZOowx4ikKoiZ:bfhczl) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L709)):*

I don't understand the question; what is the problem to be solved here?

Also I'd use `%type <int64> iconst_no_err` - no need for a pointer.
</blockquote></details>

The problem is that it seems weird to have the suffix `_no_err` in a rule name. Rules should just be named for what we are — if there's an error, the caller should move on to the next option. So I guess I was wondering if `signed_iconst` should just behave like this, and then `iconst_no_err` would not have to exist. Would have to refactor `numeric_only`; maybe I'll take a run at that, but didn't feel like messing with it.

I can't seem to change it to plain `int64` without strange build errors of the form `sqlDollar[2].int64 undefined (type sqlSymType has no field or method int64)`. Not sure what's going on there.

---

*[pkg/sql/parser/sql.y, line 3149 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHqJAKk278f8plnriR:-KyRhbjBKqkfGA7hfaUZ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/sql.y#L3149)):*

`s/[]SequenceOption{}/[]SequenceOption(nil)/`
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 3174 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyuY63yz_R8CUJOuhG:-KySzSr3LuneACpYvqs7:b-lsr8jt) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/sql.y#L3174)):*

This seems like a useful rule! Could you add some tests for it in `parse_test`? There is a section for examples that are supposed to return errors, I believe.
</blockquote></details>

Not quite sure how to add tests for an individual rule… I see some tests for syntax errors, but they all seem to be parsing statements from the top level.

---

*[pkg/sql/sqlbase/structured.go, line 325 at r17](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxtdqKI-rBPXWgMjPw4:-KySoscSDjtTq9-NPXGz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7762037ef6ad3ad3c29a12a9aaa28bd04380cb09/pkg/sql/sqlbase/structured.go#L325)):*

Kind seems reasonable enough to me, but maybe just call it `Kind` instead of `KindName`? We don't have a type named `kind` so it's not ambiguous.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.go, line 326 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHpKg4iknm4Znv6O8j:-KyRjV2c4kqt75zo2F9g:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.go#L326)):*

nit: I'd do something like:

```
switch {
case desc.IsTable():
    return ...
case desc.IsView():
    ...
case desc.IsSequence():
    ...
default:
    panic("what is going on")
}

to catch any logic errors.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.go, line 336 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyvGkS8U7S-VBnvlII:-KyRiazWRJcAfcgyy5LG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.go#L336)):*

+1
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/structured.proto, line 611 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyHsOiUFnDVky7mhnB3:-KySoulLqLzse1HdhdCv:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/sqlbase/structured.proto#L611)):*

These fields should be commented.

I'd also push to rename this to `SequenceOpts` or `SequenceDescriptor`, as the word "settings" has a specific meaning in Cockroach (session settings, cluster settings, etc.) and has not traditionally been used in this way.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664)*
<!-- Sent from Reviewable.io -->

**Comment by vilterp** [comment]
Reviewed 24 of 24 files at r30, 37 of 37 files at r35.
Review status: 0 of 38 files reviewed at latest revision, 23 unresolved discussions.

---

*[pkg/sql/planner.go, line 269 at r33](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyVKgF7ZSA5Jv7RR1V2:-KyWzyYujthV_Q-zm1gl:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/632a56ff988b0d1dfee2c0f2743b584b49b3e689/pkg/sql/planner.go#L269)):*

Given how you've modified the code here, the comment must be extended with: `The caller must ensure that seqName is fully qualified already.`
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 124 at r34](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyVJgAJMDZlDUhh5BNo:-KyWy11YFGuLq-GExMV3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2dc4cb2c520883b2bdf73f22e8dcab82118aa8a3/pkg/sql/logictest/testdata/logic_test/sequences#L124)):*

We don't typically use USE in logic tests -- SET DATABASE instead. Consistency is desirable here as the editor of a test will often grep for SET DATABASE to determine what to expect.
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/sequences, line 133 at r34](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KyVJt1lXLctz7MNjyGw:-KyWy4JwIgr2dXMfAO7P:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2dc4cb2c520883b2bdf73f22e8dcab82118aa8a3/pkg/sql/logictest/testdata/logic_test/sequences#L133)):*

This is OK but insufficient.

This must work:

```
CREATE DATABASE foo
CREATE DATABASE bar
CREATE SEQUENCE foo.x
SET DATABASE = bar
SET SEARCH_PATH = foo
SELECT nextval('x')
```
</blockquote></details>

Done.

---

*[pkg/sql/parser/builtins.go, line 1148 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1ctXhUZztEtRWnYBO:-KyWz1N7czX50NGZr3VN:b4ts8al) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/builtins.go#L1148)):*

I don't think so -- `QualifyWithDatabase` does not use the search path. See my comment on the test for details of what needs to work.
</blockquote></details>

`QualifyWithDatabase` [calls `searchAndQualifyWithDatabase`](https://github.com/cockroachdb/cockroach/blob/64ad6c6491e71fc67310f5d369bcce8d88371f91/pkg/sql/data_source.go#L465), which uses the search path. I used it because it was already on the `EvalPlanner` interface, and if we were going to pass a TableName instead of a string to `IncrementSequence` (as you suggested), it seemed desirable to ensure that that TableName was completely ready to go use.

---

*[pkg/sql/parser/sql.y, line 709 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxnWMzmrPdO0qlEw_mL:-KyX9OokiYu6AO0z9u0f:boyexb3) ([raw file](https://github.com/cockroachdb/cockroach/blob/32cc9c20606c32f96d70f5cc3dd51f4e04a28b91/pkg/sql/parser/sql.y#L709)):*

Oh I know what happened. I had forgotten something :)
Here's the trick: using `%type <int64>` here only works if you *also* add the following function definition alongside the others:

```
func (u *sqlSymUnion) int64() int64 {
   return u.val.(int64)
}
```


</blockquote></details>

Done. Thanks!

---

*[pkg/sql/parser/sql.y, line 3174 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-KxyuY63yz_R8CUJOuhG:-KyX9ReA3XC9p9be0OBu:b-bbg66q) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/sql.y#L3174)):*

I had to look again at the entire patch to grok what's going on here. :)
FYI the reason why `signed_iconst` cannot be constrainted to 64 bits, is that we also use it for DECIMAL literal values which can be larger.

Now since this is what it does really, what you currently call `iconst_no_err` would be better called `signed_iconst64`!

This is a good rule to have indeed! And documented as such:

`// signed_iconst64 is a variant of signed_iconst which only accepts (signed) integer literals that fit in an int64.` 

Then:

1) have the non-terminal produce a `int64` as I explain in my other comment above
2) review/audit the rest of the grammar for calls to AsInt64(), I suspect that those other uses correspond to a use of ICONST / signed_iconst that can be replaced by ~iconst_no_err~ signed_iconst64. For example `index_hints_params`, `opt_index_hints` etc. (You might need to add another rule `iconst64` (not signed) while you're at it)
</blockquote></details>

Done. Thanks for your direction; I suspected a refactor like this was a good idea! Put the introduction of `iconst64` and `signed_iconst64` in a separate commit at the beginning of the PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664)*

**Comment by knz** [comment]
:lgtm: with nit

---

Reviewed 10 of 10 files at r36, 14 of 15 files at r37, 4 of 5 files at r38, 7 of 7 files at r39, 3 of 3 files at r40.
Review status: all files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/builtins.go, line 1148 at r28](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky1ctXhUZztEtRWnYBO:-Ky_UEhrV4PQ5Ra6_IRz:b-rwwlih) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f452130d6b3a5b175ee4fe7a9b114bd0743141d/pkg/sql/parser/builtins.go#L1148)):*

`QualifyWithDatabase` [calls `searchAndQualifyWithDatabase`](https://github.com/cockroachdb/cockroach/blob/64ad6c6491e71fc67310f5d369bcce8d88371f91/pkg/sql/data_source.go#L465), which uses the search path. I used it because it was already on the `EvalPlanner` interface, and if we were going to pass a TableName instead of a string to `IncrementSequence` (as you suggested), it seemed desirable to ensure that that TableName was completely ready to go use.
</blockquote></details>

Oh interesting. TIL. Thanks!

---

*[pkg/sql/parser/eval.go, line 2004 at r39](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-Ky_UL1UfRT955hdlZ-p:-Ky_UL1UfRT955hdlZ-q:bipfn7q) ([raw file](https://github.com/cockroachdb/cockroach/blob/ca52203d817f2a576c1deb9d2ead5da455bbdf4c/pkg/sql/parser/eval.go#L2004)):*
> ```Go
> 
> 	// QualifyWithDatabase resolves a possibly unqualified table name into a
> 	// normalized table name that is qualified by database.
> ```

Extend the comment to explain that QualifyWithDatabase also uses the search path. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19664#-:-Ky_Uss7c2DPBN7VH9hL:b-xv64lu)*

File: pkg/sql/logictest/testdata/logic_test/sequences [hunk 2]
```diff
@@ -0,0 +1,180 @@
...
+
+statement ok
+INSERT INTO blog_posts (title) values ('bar')
+
+# TODO(vilterp) change this to [1, 2] once initial-evaluation bug has been fixed
+query I
+SELECT id FROM blog_posts ORDER BY id
+----
+2
+3
```
vilterp: Curious for some advice from you, @knz:

In this example, we create a new sequence and then use it in a DEFAULT expression. However, when we insert the first row into the table, the primary key comes out as 2, not 1 (where the sequence is supposed to start). This is because default expressions are run once at table creation time to see if they work (https://github.com/cockroachdb/cockroach/blob/d73de92d4e4f1e4b3c85e4c13b14ff6474391d6a/pkg/sql/sqlbase/table.go#L198-L205).

Would it be reasonable to flag functions which have side effects (e.g. writing to the database, as in this case), and exclude them from being run here? (Note also that we can't run it in a transaction and roll it back, because the `nextval` function operates in its own transaction)

If the solution is introduce a new flag on `Builtin` and use it to mark `nextval` as side effectful, maybe using the same terminology as postgres's volatility categories (https://www.postgresql.org/docs/8.2/static/xfunc-volatility.html) for these flags would be appropriate.

Also FWIW, I think this could be fixed in another PR if all else is good to go.

knz:
We shouldn't be evaluating expressions "just to see if they work". That's just wrong thinking. I'm OK with removing that evaluation. If for any reason a user discovers a bug that makes a default expression invalid, they can use ALTER to change it.
-- 

vilterp:
@knz I agree;  happy to remove it. Asked you because you [added the check](https://github.com/cockroachdb/cockroach/commit/581207aa5bc2037f81874afb4c2a1cc706615e4c). Is there some case we're forgetting? The logic tests still pass with it gone. Anyway, I'm going to make that a separate PR; we can discuss there.

knz:
I didn't add it -- I merely moved it from where it was (once per backfill processor) to another place (once for the entire backfill). It shouldn't have been there in the first place.

pkg/sql/parser/sql.y
```diff
@@ -6563,6 +6527,28 @@ signed_iconst:
     $$.val = &tree.NumVal{Value: constant.UnaryOp(token.SUB, $2.numVal().Value, 0)}
   }
 
+// signed_iconst64 is a variant of signed_iconst which only accepts (signed) integer literals that fit in an int64.
+signed_iconst64:
```
jordanlewis: i didn't totally follow the discussion around this one but is it intentional that this rule isn't used?

pkg/sql/sqlbase/structured.proto
```diff
@@ -606,6 +606,24 @@ message TableDescriptor {
...
+    // How much to increment the sequence by when nextval() is called.
+    optional int64 increment = 1 [(gogoproto.nullable) = false];
+    // Minimum value of the sequence.
+    optional int64 min_value = 2 [(gogoproto.nullable) = false];
+    // Maximum value of the sequence.
+    optional int64 max_value = 3 [(gogoproto.nullable) = false];
+    // Start value of the sequence.
+    optional int64 start = 4 [(gogoproto.nullable) = false];
+    // How many values to pre-allocate and cache in memory (not currently supported)
+    optional int64 cache = 5 [(gogoproto.nullable) = false];
```
jordanlewis: If we're not planning to support the cache option anytime soon we probably don't need to add this to the proto, especially not as a non-nullable field. An extra int64 per descriptor obviously won't kill us, though. And since descriptors are cached this might just not be worth worrying about.

pkg/sql/logictest/testdata/logic_test/sequences
jordanlewis: So we do already have an `impure` flag on builtins and it does already propagate to those volatility categories, FWIW.

jordanlewis (approved):
LGTM!

pkg/sql/sqlbase/structured.proto
vilterp: Agreed. Removed it.

pkg/sql/parser/sql.y
vilterp: Added comment with more explanation. It's used in a later commit than the one that introduces it.

PR #19665, opened 2017-10-30
 distsql: introduce InterleavedReaderJoiner DistSQL processor for handling interleaved joins

richardwu:
As per the [interleaved table joins RFC](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20171025_interleaved_table_joins.md),
this PR introduces an `InterleavedReaderJoiner` processor that logically
combines two table readers and a merge joiner for INNER JOINs for a
__full interleave prefix__ join between a parent and child table.

Note that the current state of `InterleavedReaderJoiner` only supports
INNER JOINs on a full interleave prefix join since this is the simplest
case (implementation-wise) and does not require buffering any rows at
all. We can still formulate an upper bound on the performance gains we
get from using an `InterleavedReaderJoiner` as opposed to two `TableReader`.
Obviously, the planner must be aware that `InterleavedReaderJoiner` only
supports this specific join scenario.


RaduBerinde:
*[pkg/sql/distsqlrun/processors.proto, line 524 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxme6pxtYXjwnMwtBU0:-Kxme6pxtYXjwnMwtBU1:b-8d5fk5) ([raw file](https://github.com/cockroachdb/cockroach/blob/761f188a044e18d9a94930ced8d7bb065c649235/pkg/sql/distsqlrun/processors.proto#L524)):*
> ```protobuf
> 
>   optional bool reverse = 2 [(gogoproto.nullable) = false];
>   repeated TableReaderSpan spans = 3 [(gogoproto.nullable) = false];
> ```

It's not obvious why a single set of spans for both tables is sufficient. Filters can become spans, and we could have different filters on the two sides. It should be explained how this should be set. Perhaps we need to set this to the "intersection" of the spans?

This is a convoluted example but say we have a child table with key `x,y` interleaved in a parent table with key `x` and
`SELECT * FROM (SELECT * FROM parent WHERE x > 10) JOIN (SELECT * FROM child WHERE x > 15) USING x`

Of course, this query would be equivalent if we had `x > 15` in both places, but we don't currently infer constraints from subtrees like that.

An even more annoying case is the same query with left outer join:
`SELECT * FROM (SELECT * FROM parent WHERE x > 10) LEFT OUTER JOIN (SELECT * FROM child WHERE x > 15) USING x`
Here we actually need to scan and output rows between 10 and 15. I don't think we can capture this one with a single set of spans (unless we keep filter expressions around in addition to the spans). It's fine if we don't support such a case but 1) the limitation must be documented clearly and 2) we need to make sure we correctly detect these cases during planning (I guess we can just require the spans to be identical).

An alternative would be to have both tables have their own spans, and have the interleave reader "merge" them while it's running, outputting any "outer" rows as necessary but this is more complicated (of course, we could start by requiring that the spans are identical initially).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

vivekmenezes:
@arjunravinarayan assigning you as a reviewer.

richardwu:
Review status: 0 of 21 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/distsqlrun/processors.proto, line 524 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxme6pxtYXjwnMwtBU0:-KxmrRamPRzN5ar5WetF:b7tb8ys) ([raw file](https://github.com/cockroachdb/cockroach/blob/761f188a044e18d9a94930ced8d7bb065c649235/pkg/sql/distsqlrun/processors.proto#L524)):*

It's not obvious why a single set of spans for both tables is sufficient. Filters can become spans, and we could have different filters on the two sides. It should be explained how this should be set. Perhaps we need to set this to the "intersection" of the spans?

This is a convoluted example but say we have a child table with key `x,y` interleaved in a parent table with key `x` and
`SELECT * FROM (SELECT * FROM parent WHERE x > 10) JOIN (SELECT * FROM child WHERE x > 15) USING x`

Of course, this query would be equivalent if we had `x > 15` in both places, but we don't currently infer constraints from subtrees like that.

An even more annoying case is the same query with left outer join:
`SELECT * FROM (SELECT * FROM parent WHERE x > 10) LEFT OUTER JOIN (SELECT * FROM child WHERE x > 15) USING x`
Here we actually need to scan and output rows between 10 and 15. I don't think we can capture this one with a single set of spans (unless we keep filter expressions around in addition to the spans). It's fine if we don't support such a case but 1) the limitation must be documented clearly and 2) we need to make sure we correctly detect these cases during planning (I guess we can just require the spans to be identical).

An alternative would be to have both tables have their own spans, and have the interleave reader "merge" them while it's running, outputting any "outer" rows as necessary but this is more complicated (of course, we could start by requiring that the spans are identical initially).
</blockquote></details>

I had originally planned for `InterleaveReaderJoiner` to be agnostic to how the spans for its tables are structured. That is: the planner should handle the heavy-lifting of taking the intersection (for equality inner joins) or the union of spans (`MergeSpans`) for all other types of joins (with potential heuristics where we know an intersection is sufficient). The processor simply delegates the scan to `MultiRowFetcher` and spits out whatever rows it gets back (joined together on the join condition).

For example, there are inner joins that require a union of spans
```
SELECT * FROM (SELECT * FROM parent WHERE x > 10) as parent JOIN (SELECT * FROM child WHERE x > 15) as child ON parent.x <= child.x
```

I definitely agree that I should add a comment on how it should be set.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

RaduBerinde:
*[pkg/sql/distsqlrun/processors.proto, line 524 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxme6pxtYXjwnMwtBU0:-KxmvAPe79zyWvoPELvb:bsu82ib) ([raw file](https://github.com/cockroachdb/cockroach/blob/761f188a044e18d9a94930ced8d7bb065c649235/pkg/sql/distsqlrun/processors.proto#L524)):*

I had originally planned for `InterleaveReaderJoiner` to be agnostic to how the spans for its tables are structured. That is: the planner should handle the heavy-lifting of taking the intersection (for equality inner joins) or the union of spans (`MergeSpans`) for all other types of joins (with potential heuristics where we know an intersection is sufficient). The processor simply delegates the scan to `MultiRowFetcher` and spits out whatever rows it gets back (joined together on the join condition).

For example, there are inner joins that require a union of spans
```
SELECT * FROM (SELECT * FROM parent WHERE x > 10) as parent JOIN (SELECT * FROM child WHERE x > 15) as child ON parent.x <= child.x
```

I definitely agree that I should add a comment on how it should be set.
</blockquote></details>

In my last example, there is no way to calculate a single set of spans that works (unless maybe we add back in some filtering expressions)..

In your example, I don't see how doing the union of spans is correct. The spans are `1/10-` and `1/15-`. If we scan everything in the `1/10-` span, we will include results form the child where x is between 10 and 15. We would have to keep `child.x > 15` as a filtering expression which doesn't currently happen (we remove filtering expressions that we were able to exactly convert to spans).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

richardwu:
Review status: 0 of 21 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/sql/distsqlrun/processors.proto, line 524 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxme6pxtYXjwnMwtBU0:-KxmwULey9c1i_l1hKeP:bvhfbs0) ([raw file](https://github.com/cockroachdb/cockroach/blob/761f188a044e18d9a94930ced8d7bb065c649235/pkg/sql/distsqlrun/processors.proto#L524)):*

In my last example, there is no way to calculate a single set of spans that works (unless maybe we add back in some filtering expressions)..

In your example, I don't see how doing the union of spans is correct. The spans are `1/10-` and `1/15-`. If we scan everything in the `1/10-` span, we will include results form the child where x is between 10 and 15. We would have to keep `child.x > 15` as a filtering expression which doesn't currently happen (we remove filtering expressions that we were able to exactly convert to spans).
</blockquote></details>

Ahh gotcha.

What do you think about the planner peeking into `scanNode.origFilter` and applyiing that to the `PostProcessSpec` under the above edge cases?

What we could do for correctness at first is to always the the union of spans and to always apply `origFilter` to the `PostProcessSpec` of each table. I reckon this would still be more efficient than the current solution. We can then introduce heuristics where we know that we can take the intersection of spans instead at the planner-level.

I prefer doing this during planning since we'd still have access to all the fields of our `planNode`s. It also keeps the proto interface simple. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

RaduBerinde:
*[pkg/sql/distsqlrun/processors.proto, line 524 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxme6pxtYXjwnMwtBU0:-Kxn1H0-XFXNkGfBqPub:b-sfhsf) ([raw file](https://github.com/cockroachdb/cockroach/blob/761f188a044e18d9a94930ced8d7bb065c649235/pkg/sql/distsqlrun/processors.proto#L524)):*

Ahh gotcha.

What do you think about the planner peeking into `scanNode.origFilter` and applyiing that to the `PostProcessSpec` under the above edge cases?

What we could do for correctness at first is to always the the union of spans and to always apply `origFilter` to the `PostProcessSpec` of each table. I reckon this would still be more efficient than the current solution. We can then introduce heuristics where we know that we can take the intersection of spans instead at the planner-level.

I prefer doing this during planning since we'd still have access to all the fields of our `planNode`s. It also keeps the proto interface simple. 
</blockquote></details>

That works. We could also first only allow cases that require intersecting the spans, (e.g. inner joins with equality constraints) and avoid the extra filtering in these common cases. We can worry about using `origFilter` to cover more cases later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

richardwu:
Added the comment with `spans` and rebased!

---

Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-:-KxnaBacIsnASweOboal:bvb98vs)*

RaduBerinde:
Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/processors.proto, line 536 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KxnhGeguBFV86vZid-7:-KxnhGeguBFV86vZid-8:buius0f) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/processors.proto#L536)):*
> ```protobuf
>   // Typically, one would want to pass in the union of all spans from the
>   // tables. This ensures all original rows are scanned. To ensure correctness,
>   // we need to propagate the original filters:
> ```

left-over sentence? (it's explained more clearly below)

---

*[pkg/sql/distsqlrun/processors.proto, line 545 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KxnhShbLR9WJSds3n2D:-KxnhShbLR9WJSds3n2E:b-lai237) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/processors.proto#L545)):*
> ```protobuf
>   //
>   // For example's sake, take the following query
>   //    SELECT * FROM (SELECT * FROM parent WHERE x > 5) LEFT JOIN (SELECT * FROM child WHERE x > 10) USING x
> ```

[nit] long line (can wrap around LEFT JOIN)

---

*[pkg/sql/distsqlrun/processors.proto, line 558 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KxnhoDVdOvJtEnSX0Fz:-KxnhoDVdOvJtEnSX0G-:bw2vx5j) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/processors.proto#L558)):*
> ```protobuf
>   //
>   // TODO(richardwu): Update this once other types of joins are implemented.
>   // The current iteration only supports equality inner joins: one can in fact
> ```

I would just say "There are common cases (like equality inner joins) where we can take the intersection of the spans to avoid having to re-introduce filters." It should be a consideration of the higher-level planning code how this is used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

RaduBerinde:
Please bump the `DistSQLVersion` in `distsqlrun/server.go` (while you're there, `s/version_history.txt/this comment`).

Great stuff, you're making very quick progress!! Skimmed over some parts, will need to do a more detailed review.

---

Review status: 0 of 14 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsqlrun/base.go, line 46 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KxnlG_yPWJowZvY6InO:-KxnlG_yPWJowZvY6InP:baf5c1w) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/base.go#L46)):*
> ```Go
> // We ignore any limits that are higher than this value to avoid any
> // overflows.
> const readerOverflowProtection = 1000000000
> ```

This should be much higher. A billion rows is not that astronomical.

---

*[pkg/sql/distsqlrun/interleave_reader_joiner.go, line 141 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KxnmQ_TMenM3_grfT0z:-KxnmQ_TMenM3_grfT1-:bqevy72) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/interleave_reader_joiner.go#L141)):*
> ```Go
> 		irj.limitHint = int64(post.Limit)
> 	} else if spec.LimitHint != 0 && spec.LimitHint <= readerOverflowProtection {
> 		// If it turns out that limiHint rows are sufficient for our consumer, we
> ```

This stuff is copy/pasted from tablereader, no? Maybe we can extract them in a common function?

---

*[pkg/sql/distsqlrun/interleave_reader_joiner.go, line 198 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxnn4xn1SpNaSqu-3da:-Kxnn4xn1SpNaSqu-3db:b-eg72xg) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/interleave_reader_joiner.go#L198)):*
> ```Go
> 		}
> 
> 		colIdxMap := make(map[sqlbase.ColumnID]int, len(desc.Columns))
> ```

Sorry for not mentioning this in the other review - we should try to use `util.FastIntSet` instead of `[]bool`. Not sure if it's worth refactoring this code, just wanted to make sure you're aware of it.

---

*[pkg/sql/distsqlrun/interleave_reader_joiner.go, line 219 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxnn_4xW8sbajQJYbNo:-Kxnn_4xW8sbajQJYbNp:b-vqcjgr) ([raw file](https://github.com/cockroachdb/cockroach/blob/92218cc6e2c80412bf64b7cc1c0078ce6986de22/pkg/sql/distsqlrun/interleave_reader_joiner.go#L219)):*
> ```Go
> // This should be called after the fetcher was used to read everything this
> // interleaveReaderJoiner was supposed to read.
> func (irj *interleaveReaderJoiner) sendMisplannedRangesMetadata(ctx context.Context) {
> ```

Same comment on duplicating code, consider making a helper function.

---

*[pkg/sql/distsqlrun/processors.proto, line 524 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-Kxme6pxtYXjwnMwtBU0:-KxnmjspAUlS-7ljBGM0:b-k9ukpm) ([raw file](https://github.com/cockroachdb/cockroach/blob/761f188a044e18d9a94930ced8d7bb065c649235/pkg/sql/distsqlrun/processors.proto#L524)):*

That works. We could also first only allow cases that require intersecting the spans, (e.g. inner joins with equality constraints) and avoid the extra filtering in these common cases. We can worry about using `origFilter` to cover more cases later.
</blockquote></details>

(That the processor itself only supports INNER is a somewhat separate limitation, it should be mentioned in the comment for InterleaveReaderJoinerSpec

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-:-KxnlpOJlnbStD9zW8KP:bycjgpn)*

richardwu:
In the latest revision, I separated out the one `repeated TableReaderSpan spans` field under `InterleaveReaderJoinerSpec` into individual `repeated TableReaderSpan spans` per `InterleaveReaderJoinerTable`. This means that the consumer (i.e. `distSQLPlanner`) will pass in the relevant spans per table and delegate the responsibility of the "merging" part to `InterleaveReaderJoiner`.

This was discussed as the cleaner and less error-prone approach in the planning PR https://reviewable.io/reviews/cockroachdb/cockroach/19853#-KyRIBAoAZP6LvsXIJ1Q.

In short, `InterleaveReaderJoiner` merges all spans of the tables `MergeSpans` and for every kv retrieved corresponding to a table, checks if the key is contained in the specified `spans` (via `ContainsKey`).

Note I had to move a bunch of test helper methods to a `sq/testutils` source file in order to use `makeSpans` for testing  `InterleaveReaderJoiner` with intermediary spans that have a constraint/filter applied.

---

Review status: 0 of 24 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-:-KyWB_lMKYLqprgcUbsM:brow6d7)*

RaduBerinde:
I like the updated approach!  The implementation is simpler than I thought, thanks to your MergeSpans+ContainsKey idea.

---

Review status: 0 of 26 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 501 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KyX8g8rHLjeAXbLDx30:-KyX8g8s3N6b4hgwkN0V:bmbmjvy) ([raw file](https://github.com/cockroachdb/cockroach/blob/0c91695f079b5c5125f3842859f0e4af74463f12/pkg/sql/sqlbase/multirowfetcher.go#L501)):*
> ```Go
> 	// The index key is not within our specified span of keys for the
> 	// particular table.
> 	if !mrf.tables[tableIdx].spans.ContainsKey(initialKey) {
> ```

It's not great that we look through all spans for each key. It's fine for the first implementation (and I like the simplicity) but please add a TODO. We could just look at the first span and throw it away after we go beyond it (though reverse scans might make this annoying). A simpler solution would be to also return a "hint" (span index) from `ContainsKey` that we remember and pass to the next call (which looks at that span first).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-:-KyX9oCLsWjvqoX96hst:bgi5vp2)*

richardwu:
Review status: 0 of 26 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 501 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-KyX8g8rHLjeAXbLDx30:-Kyq_X82M7lwK_gyt86N:by55k68) ([raw file](https://github.com/cockroachdb/cockroach/blob/0c91695f079b5c5125f3842859f0e4af74463f12/pkg/sql/sqlbase/multirowfetcher.go#L501)):*

It's not great that we look through all spans for each key. It's fine for the first implementation (and I like the simplicity) but please add a TODO. We could just look at the first span and throw it away after we go beyond it (though reverse scans might make this annoying). A simpler solution would be to also return a "hint" (span index) from `ContainsKey` that we remember and pass to the next call (which looks at that span first).
</blockquote></details>

Done. For reverse, we could also pass in the hint to iterate in the reverse direction into `ContainsKey`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

RaduBerinde:
:lgtm_strong: 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-:-L-3-1tosqABitr-x3vb:b-uwkowd)*

RaduBerinde:
Reviewed 2 of 29 files at r4, 5 of 25 files at r5, 21 of 21 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665)*

richardwu:
Thanks for the review!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19665#-:-L-3azuEdXCPSmUJo99w:b-9qdoz3)*

PR #19682, opened 2017-10-31
sql: prefer hard limit

### LINKED ISSUE #19677: sql: prefer hard limit
We are seeing a case where we have a soft limit coming from above a limit node, and a sort node underneath. We are using the soft limit because it's smaller, but this makes the sort node store all the rows in memory. The hard limit would have avoided that memory usage.

Without more information, it's hard in general to choose one way or the other, as this comment explains: https://github.com/cockroachdb/cockroach/blob/1aa51c7c90c3a2f8b41f58535d8ebbb3304b4bdd/pkg/sql/limit_opt.go#L62

In the future, this decision should be explored by the optimizer. In the meantime, I think we should use the hard limit because it can make a dramatic impact on memory usage.

RaduBerinde:
We are seeing a case where we have a soft limit coming from above a limit node,
and a sort node underneath. We are using the soft limit because it's smaller,
but this makes the sort node store all the rows in memory. The hard limit avoids
this.

Switch to preferring the hard limit.

Fixes #19677.

### DIFF
--- a/pkg/sql/limit_opt.go
+++ b/pkg/sql/limit_opt.go
@@ -57,20 +57,20 @@ func applyLimit(plan planNode, numRows int64, soft bool) {
 
 	case *limitNode:
 		// A higher-level limitNode or EXPLAIN is pushing a limit down onto
-		// this node. Accept it unless the local limit is definitely
-		// smaller, in which case we propagate that as a hard limit instead.
+		// this node. Prefer the local "hard" limit, unless the limit pushed
+		// down is "hard" and smaller than the local limit.
+		//
 		// TODO(radu): we may get a smaller "soft" limit from the upper node
 		// and we may have a larger "hard" limit locally. In general, it's
 		// not clear which of those results in less work.
 		if !n.evaluated {
 			n.estimateLimit()
 		}
-		hintCount := numRows
-		if hintCount > n.count {
-			hintCount = n.count
-			soft = false
+		count := n.count
+		if !soft && numRows < count {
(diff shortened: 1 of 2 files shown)


RaduBerinde:
I plan to cherry-pick this into 1.1.x.

knz:
:lgtm: but please include an example query in the commit message and show why there was a problem. 

Also prepare a one sentence release note summary, as discussed by Ben over the contributor guide two weeks ago.

---

Reviewed 2 of 2 files at r1.

---

*[pkg/sql/logictest/testdata/logic_test/limit, line 75 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19682#-Kxmy9sGtxqbu-xgEImc:-Kxmy9sGtxqbu-xgEImd:bb7phvt) ([raw file](https://github.com/cockroachdb/cockroach/blob/dda9486379d5eb1a89c68b49025ceae26f252b14/pkg/sql/logictest/testdata/logic_test/limit#L75)):*
> ```
> 
> # This kind of query can be used to work around memory usage limits. We need to
> # choose the "hard" llmit of 100 over the "soft" limit of 25 (with the hard
> ```

typo

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19682#-:-Kxmyfyjc22UPX7jScS-:b-xfd5ta)*

asubiotto (approved):

RaduBerinde:
TFTR, done!

PR #19683, opened 2017-10-31
sql: rework join filter propagation

RaduBerinde:
With the previous change of removing the merged columns, we can no longer
map constraints on merged columns to constraints on the left and right side.

Instead, we implement a more general mechanism that "transfers" constraints from
one side to the other. This handles more cases well: e.g.
`SELECT FROM a JOIN b USING (x) WHERE a.x > 1` will push down `x > 1` on both
`a` and `b`.

For inner and left outer joins, merged columns are aliases for the corresponding
left side column, so any constraints on them end up being propagated through the
renderNode and then to both sides of the join. For right outer joins, merged
columns are aliases for the corresponding right side column (except for collated
strings and the like).

Updates #19640.

### DIFF
--- a/pkg/sql/filter_opt.go
+++ b/pkg/sql/filter_opt.go
@@ -196,18 +196,7 @@ func (p *planner) propagateFilters(
 		return p.addRenderFilter(ctx, n, extraFilter)
 
 	case *joinNode:
-		switch n.joinType {
-		case joinTypeInner, joinTypeLeftOuter, joinTypeRightOuter:
-			return p.addJoinFilter(ctx, n, extraFilter)
-		default:
-			// There's nothing we can do for full outer joins; simply
-			// trigger filter optimization in the sub-nodes.
-			var err error
-			if n.left.plan, err = p.triggerFilterPropagation(ctx, n.left.plan); err == nil {
-				n.right.plan, err = p.triggerFilterPropagation(ctx, n.right.plan)
-			}
-			return n, extraFilter, err
-		}
+		return p.addJoinFilter(ctx, n, extraFilter)
 
 	case *indexJoinNode:
 		panic("filter optimization must occur before index selection")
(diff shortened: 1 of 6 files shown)


knz:
Code LGTM but I'm not happy about the large decrease in comments.

---

Reviewed 6 of 6 files at r1.

---

*[pkg/sql/filter_opt.go, line 535 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-Kxmzy4aMticYoz6rvIu:-Kxmzy4aMticYoz6rvIv:b-nal4uo) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0afcfce7cb2792302a026011ff57e3ccd7e1aef/pkg/sql/filter_opt.go#L535)):*
> ```Go
> 	var result parser.TypedExpr = parser.DBoolTrue
> 
> 	andExprs := splitAndExpr(&n.planner.evalCtx, cond, nil)
> ```

I think you should keep a long-ish comment like the one that was there before, that explains in prose what the code below does.
The explanation is not that different btw.

---

*[pkg/sql/filter_opt.go, line 630 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-Kxn-DsIlnOa2fAuDDu6:-Kxn-DsIlnOa2fAuDDu7:b53e8hv) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0afcfce7cb2792302a026011ff57e3ccd7e1aef/pkg/sql/filter_opt.go#L630)):*
> ```Go
> 	ctx context.Context, n *joinNode, extraFilter parser.TypedExpr,
> ) (planNode, parser.TypedExpr, error) {
> 	numLeft := len(n.left.info.sourceColumns)
> ```

ditto, use a comment to explain the overall structure of the code below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-:-Kxn-kmyCNdDA3vSgyY7:bfw527)*

RaduBerinde:

---

*[pkg/sql/filter_opt.go, line 535 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-Kxmzy4aMticYoz6rvIu:-KxnAEVkAHRFeMy6700J:b-mk1hrc) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0afcfce7cb2792302a026011ff57e3ccd7e1aef/pkg/sql/filter_opt.go#L535)):*

I think you should keep a long-ish comment like the one that was there before, that explains in prose what the code below does.
The explanation is not that different btw.
</blockquote></details>

Done, PTAL. While writing this I realized we can be a bit smarter about keeping the original expression and I made a small change.

---

*[pkg/sql/filter_opt.go, line 630 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-Kxn-DsIlnOa2fAuDDu6:-KxnBpwseWplKorKEmUz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b0afcfce7cb2792302a026011ff57e3ccd7e1aef/pkg/sql/filter_opt.go#L630)):*

ditto, use a comment to explain the overall structure of the code below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683)*

knz:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/filter_opt.go, line 562 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnE_AC3hr_WxWbnTHh:-KxnE_AC3hr_WxWbnTHi:b26g36m) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/filter_opt.go#L562)):*
> ```Go
> 	//      result:             left.x > 10 AND right.x > 10
> 	//
> 	//      conjunction:        left.x > right.v
> ```

Why do you get a result with the right splitFilter here, but none for `left.v > right.w` below?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683)*

RaduBerinde:
*[pkg/sql/filter_opt.go, line 562 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnE_AC3hr_WxWbnTHh:-KxnJ7w5Y9M6B8rQ66Nm:bjsd53b) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/filter_opt.go#L562)):*

Why do you get a result with the right splitFilter here, but none for `left.v > right.w` below?
</blockquote></details>

`v` and `w` are not equality columns, so we can't transfer anything

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683)*

richardwu:

---

*[pkg/sql/filter_opt.go, line 549 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnEA7gtsMjo7eam5wT:-KxnEA7hXGlUWz4JDQTS:btlyld8) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/filter_opt.go#L549)):*
> ```Go
> 	// not keeping redundant constraints.
> 	//
> 	// We split the condition into conjunctions and use splitFilter on each
> ```

Do you mean __disjunction__/literal clauses (of CNF)?

---

*[pkg/sql/logictest/testdata/logic_test/join, line 1540 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnLv860ArCKEL_3pOY:-KxnLv860ArCKEL_3pOZ:b1ddzlq) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/logictest/testdata/logic_test/join#L1540)):*
> ```
> 
> statement ok
> CREATE TABLE xyu (x INT, y INT, u INT, PRIMARY KEY(x,y,u)); INSERT INTO xyu VALUES (1, 1, 11), (1, 2, 12), (2, 2, 22)
> ```

Since you're using `x > 0` filters, can you add some values into `xyu` and `xyv` where `x <= 0`? 

---

*[pkg/sql/logictest/testdata/logic_test/join, line 1736 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnMx2N-DPCKzvQeD8m:-KxnMx2N-DPCKzvQeD8n:b-dscrcc) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/logictest/testdata/logic_test/join#L1736)):*
> ```
> 1  scan  ·               ·                                       (x, y, u)           x!=NULL; y!=NULL; u!=NULL; key(x,y,u); +x,+y
> 1  ·     table           xyu@primary                             ·                   ·
> 1  ·     spans           ALL                                     ·                   ·
> ```

Hrm, shouldn't this also be filtered? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683)*

RaduBerinde:
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/filter_opt.go, line 549 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnEA7gtsMjo7eam5wT:-Kxnej8SKEG7twaQMzWU:b-w0x0b8) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/filter_opt.go#L549)):*

Do you mean __disjunction__/literal clauses (of CNF)?
</blockquote></details>

Ah. Renamed to "conjunct" (operand of conjunction).

---

*[pkg/sql/logictest/testdata/logic_test/join, line 1540 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnLv860ArCKEL_3pOY:-Kxnez9KhLq7M5gNAVwN:b-80tlqk) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/logictest/testdata/logic_test/join#L1540)):*

Since you're using `x > 0` filters, can you add some values into `xyu` and `xyv` where `x <= 0`? 
</blockquote></details>

Sure.

---

*[pkg/sql/logictest/testdata/logic_test/join, line 1736 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-KxnMx2N-DPCKzvQeD8m:-KxnfXsQLb95YdAJLCiz:b-8byu45) ([raw file](https://github.com/cockroachdb/cockroach/blob/ff6e01491f506289e7b2f98fae05ee28dc9c92f1/pkg/sql/logictest/testdata/logic_test/join#L1736)):*

Hrm, shouldn't this also be filtered? 
</blockquote></details>

No.. it's a left outer with no filter, we actually need to output all rows from the left side.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683)*

knz:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19683#-:-KxoT_MawAclsIk96-k5:bnfp4nl)*

PR #19688, opened 2017-10-31
cherry-pick 1.1: sql: prefer hard limit

### LINKED ISSUE #19677: sql: prefer hard limit

RaduBerinde:
We are seeing a case where we have a soft limit coming from above a limit node,
and a sort node underneath, something along the lines of:
  SELECT DISTINCT (<some query> ORDER BY x LIMIT 100) ORDER BY x LIMIT 25

We are using the soft limit because it's smaller, but this makes the sort node
store all the rows in memory. The hard limit avoids this.

Switch to preferring the hard limit.

Fixes #19677.

Release note: "improved memory usage for certain queries that use limits at
multiple levels"
CC @cockroachdb/release

### DIFF
--- a/pkg/sql/limit_opt.go
+++ b/pkg/sql/limit_opt.go
@@ -57,20 +57,20 @@ func applyLimit(plan planNode, numRows int64, soft bool) {
 
 	case *limitNode:
 		// A higher-level limitNode or EXPLAIN is pushing a limit down onto
-		// this node. Accept it unless the local limit is definitely
-		// smaller, in which case we propagate that as a hard limit instead.
+		// this node. Prefer the local "hard" limit, unless the limit pushed
+		// down is "hard" and smaller than the local limit.
+		//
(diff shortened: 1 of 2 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19688#-:-KxnUK20uZMGoRS8umXs:bnfp4nl)*

PR #19691, opened 2017-10-31
Update honnef.co/go/tools, unpin golang.org/x/sys

tamird:
Includes a small drive-by cleanup in storage.

### DIFF
--- a/build/style_test.go
+++ b/build/style_test.go
@@ -954,7 +954,7 @@ func TestStyle(t *testing.T) {
 			},
 			staticcheck.NewChecker(): {
 				// sql/ir/irgen/parser/yaccpar:362:3: this value of irgenDollar is never used (SA4006)
-				&lint.GlobIgnore{Pattern: "github.com/cockroachdb/cockroach/pkg/sql/ir/irgen/parser/yaccpar", Checks: []string{"SA4006"}},
+				&lint.GlobIgnore{Pattern: "github.com/cockroachdb/cockroach/pkg/sql/ir/irgen/parser/irgen.go", Checks: []string{"SA4006"}},
 				// The generated parser is full of `case` arms such as:
 				//
 				// case 1:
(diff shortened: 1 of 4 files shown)


build/style_test.go
```diff
@@ -1015,15 +1015,15 @@ func TestStyle(t *testing.T) {
 				// which results in the unused warning:
 				//
 				// sql/parser/yaccpar:362:3: this value of sqlDollar is never used (SA4006)
-				&lint.GlobIgnore{Pattern: "github.com/cockroachdb/cockroach/pkg/sql/parser/yaccpar", Checks: []string{"SA4006"}},
+				&lint.GlobIgnore{Pattern: "github.com/cockroachdb/cockroach/pkg/sql/parser/sql.go", Checks: []string{"SA4006"}},
```
tschottdorf: Why this change? Do the comments need an update? Ditto below.

pkg/storage/allocator.go
```diff
@@ -534,11 +534,8 @@ func (a Allocator) RebalanceTarget(
 			ReplicaID: rangeInfo.Desc.NextReplicaID,
 		}
 
-		// Deep-copy the Replicas slice since we'll mutate it.
 		desc := *rangeInfo.Desc
-		desc.Replicas = append([]roachpb.ReplicaDescriptor(nil), desc.Replicas...)
-
-		desc.Replicas = append(desc.Replicas, newReplica)
+		desc.Replicas = append(desc.Replicas[:len(desc.Replicas):len(desc.Replicas)], newReplica)
```
tschottdorf: Please restore the comment. This code is nice and compressed but pretty sure most people will need to squint at it for a minute or two to figure out what's going on.

tbg (approved):
LGTM with some comments and also @bdarnell to approve the dep bumps.

build/style_test.go
tamird: This is just reverting a previous change that was made necessary by an upstairs bug (see the comment being removed from Gopkg.toml)

pkg/storage/allocator.go
tamird: Restored.

build/style_test.go
tamird: Er, I misspoke - there's no comment referring to the upstream bug, but here's where it got fixed: https://github.com/dominikh/go-tools/commit/80dfe5569e5c609daca40943b2385b4c6cc33b48

bdarnell:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19691#-:-KxuLTjgVa4rNEv8cJqd:bnfp4nl)*

PR #19693, opened 2017-10-31
specify go 1.9+ for math/bits in fast_int_set.go in go get ...cockroach

phelanm:


RaduBerinde:
LGTM, thanks!

RaduBerinde:
I think it's 1.10 by now..

RaduBerinde:
The text says "Note that the Makefile enforces the specific version required, as it is updated frequently" so I think the lack of a specific version was intentional.

PR #19700, opened 2017-10-31
sql: change JSON datums to use interfaces

### LINKED ISSUE #19668: sql: struct vs. interface for DJSON
It's known that interface dispatch in Go can be somewhat slow. DJSON is a natural union type (a DJSON can be one of `true`, `false`, `null`, a number, a string, an array, or an object). This leads to two obvious implementations: an interface type with a struct implementation of each of the seven, or a single struct containing a tag field and a separate field for each type requiring data. The second option is what has been implemented [at the moment](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/parser/datum.go#L1982-L1992) with an eye towards trying the other option. This issue tracks the other option and is a place to document the discussion regarding it.

I have a branch that implements all the features currently supported on master via an interface, and a very [rudimentary benchmark](https://gist.github.com/justinj/ac53d2ee36dcce403a2a6c5919d8a265) (`Compare` was the only operation on master when I wrote this and the document is just from the Wikipedia page).

The results of the benchmark are this:

Struct implementation (currently on master):
```
BenchmarkJSON-8          2000000               894 ns/op
```
Interface implementation (on my local branch):
```
BenchmarkJSON-8          1000000              2095 ns/op
```

However, the struct version is *significantly* heavier in terms of memory size of a datum - even just a `true` datum takes ~120 bytes to store given the bloat of all the other fields, compared to just a pointer for the interface implementation.

It's not immediately clear to me which direction to take this trade-off, although I'm leaning towards the interface implementation given that the memory difference is so significant. It's possible (though I haven't tried this yet) that some kind of hybrid approach is possible where we store some smaller/common types in a struct (`true`, `false`, `null`, ints) with a pointer to the larger ones?

I could also take both implementations further down the line to support storing them in columns to allow for a more detailed benchmark.

Interested in your thoughts @knz @jordanlewis @m-schneider @nvanbenschoten @petermattis

justinj:
Resolves #19668.

The `return nil, nil` at the end of interpretJSON is addressed by #19592 which I'll rebase after this merges.


jordanlewis:
Here is a radical idea: what if we move all of the JSON internals to their own package? I would personally love to see that. Doesn't have to be this PR obviously but just thought of it.

---

Reviewed 3 of 3 files at r1.

---

*[pkg/sql/parser/datum.go, line 1992 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs394K860ZeGpU079m:-Kxs394K860ZeGpU079n:b-5v1eum) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/datum.go#L1992)):*
> ```Go
> // IsMin implements the Datum interface.
> func (d *DJSON) IsMin(_ *EvalContext) bool {
> 	_, ok := d.jsonVal.(jsonNull)
> ```

You should be able to keep this I think - there should be a constant null value (`var nullJSONValue = jsonNull{}` or something) that is always used to construct and test for null values.

---

*[pkg/sql/parser/datum.go, line 2003 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs3LoXv_ZsepgBG6R3:-Kxs3LoXv_ZsepgBG6R4:b-z3p5ge) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/datum.go#L2003)):*
> ```Go
> // min implements the Datum interface.
> func (d *DJSON) min(_ *EvalContext) (Datum, bool) {
> 	return &DJSON{jsonNull{}}, true
> ```

ditto above

---

*[pkg/sql/parser/json.go, line 29 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs3VV0rsfYKTupXlwc:-Kxs3VV0rsfYKTupXlwd:b-nym2n3) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L29)):*
> ```Go
> type jsonType int
> 
> // This enum defines the ordering of types.
> ```

Add "It should not be reordered." or something to make the intention super clear

---

*[pkg/sql/parser/json.go, line 43 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs4DToGrRkGd4J4WWi:-Kxs4DToGrRkGd4J4WWj:b-nkvuad) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L43)):*
> ```Go
> type jsonVal interface {
> 	compare(*EvalContext, jsonVal) int
> 	typeOrdinal() jsonType
> ```

maybe just call this jsonType()? jsonType doesn't seem like an ordinal unless you look under the covers to the underlying int.

---

*[pkg/sql/parser/json.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs6cAu_AX0fO2hKwdS:-Kxs6cAu_AX0fO2hKwdT:b-o09uri) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L53)):*
> ```Go
> 
> type jsonNumber DDecimal
> type jsonString DString
> ```

Is there a reason why this is a DString and not just a native string? I'm not sure about what the tradeoffs here but wondering if you've thought about it.

---

*[pkg/sql/parser/json.go, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs6jX_cz9cDL-JiSgw:-Kxs6jX_cz9cDL-JiSgx:b-isav0a) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L58)):*
> ```Go
> 
> type jsonKeyValuePair struct {
> 	k DString
> ```

ditto string vs DString

---

*[pkg/sql/parser/json.go, line 84 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs5d0O_k4x0Yuilzde:-Kxs5d0O_k4x0Yuilzdf:bux4f0t) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L84)):*
> ```Go
> 
> func (j jsonFalse) compare(_ *EvalContext, other jsonVal) int {
> 	if other.typeOrdinal() > j.typeOrdinal() {
> ```

does the compiler know not to call both of these functions twice? I hate to ask... but could you benchmark it or look at the generated code? All of this compare stuff will be on the hot path for everything involving json including index searches so it'd be good to make it quick

Also I've always been curious when Go can detect when it doesn't have to re-call a method.

---

*[pkg/sql/parser/json.go, line 225 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs5QVtzhNQVL-fWwzn:-Kxs5QVtzhNQVL-fWwzo:b-pgekn8) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L225)):*
> ```Go
> 
> func (j jsonNumber) size() uintptr {
> 	dec := DDecimal(j)
> ```

would `return DDecimal(j).Size()` work as well?

---

*[pkg/sql/parser/json.go, line 231 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs6tb4_KDmqTodhk44:-Kxs6tb4_KDmqTodhk45:b-z3p5ge) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L231)):*
> ```Go
> func (j jsonString) size() uintptr {
> 	str := DString(j)
> 	return str.Size()
> ```

ditto above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-:-Kxs799ruqNyuMO9lU7Q:blo5q32)*

justinj:
it is done and it is a big improvement

---

Review status: 0 of 6 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/sql/parser/datum.go, line 1992 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs394K860ZeGpU079m:-KxsGrNoW01-aBJ3MGRg:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/datum.go#L1992)):*

You should be able to keep this I think - there should be a constant null value (`var nullJSONValue = jsonNull{}` or something) that is always used to construct and test for null values.
</blockquote></details>

Done.

---

*[pkg/sql/parser/datum.go, line 2003 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs3LoXv_ZsepgBG6R3:-KxsOyslu4C6WCY091R-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/datum.go#L2003)):*

ditto above
</blockquote></details>

Done.

---

*[pkg/sql/parser/json.go, line 29 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs3VV0rsfYKTupXlwc:-KxsP2-00qPDIk1cC-7g:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L29)):*

Add "It should not be reordered." or something to make the intention super clear
</blockquote></details>

Done.

---

*[pkg/sql/parser/json.go, line 43 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs4DToGrRkGd4J4WWi:-KxsQOPn2g2w_vsKE2OZ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L43)):*

maybe just call this jsonType()? jsonType doesn't seem like an ordinal unless you look under the covers to the underlying int.
</blockquote></details>

Done.

---

*[pkg/sql/parser/json.go, line 53 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs6cAu_AX0fO2hKwdS:-KxsQPk4ztRwhKxR73pJ:blais6z) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L53)):*

Is there a reason why this is a DString and not just a native string? I'm not sure about what the tradeoffs here but wondering if you've thought about it.
</blockquote></details>

It just simplifies calling `Compare` and `Format`, no real other reason.

---

*[pkg/sql/parser/json.go, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs6jX_cz9cDL-JiSgw:-KxsmnZr962CBtMUfoLf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L58)):*

ditto string vs DString
</blockquote></details>

Done.

---

*[pkg/sql/parser/json.go, line 84 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs5d0O_k4x0Yuilzde:-KxsmtseRERAeaRE4KlI:b-5ho3it) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L84)):*

does the compiler know not to call both of these functions twice? I hate to ask... but could you benchmark it or look at the generated code? All of this compare stuff will be on the hot path for everything involving json including index searches so it'd be good to make it quick

Also I've always been curious when Go can detect when it doesn't have to re-call a method.
</blockquote></details>

/shrug I ended up making it clearer I think by removing the extra call

---

*[pkg/sql/parser/json.go, line 225 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs5QVtzhNQVL-fWwzn:-Kxsn5iRbWDum-zaenKx:bvbt5qr) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L225)):*

would `return DDecimal(j).Size()` work as well?
</blockquote></details>

not a concern anymore, but no, since `Size` was a pointer method and you can't call them in this position for some reason.

---

*[pkg/sql/parser/json.go, line 231 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxs6tb4_KDmqTodhk44:-KxsnAppyqUeAWfZ09DJ:bb7od0r) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/json.go#L231)):*

ditto above
</blockquote></details>

same

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-:-KxsGUSqgcy3voksGFqr:b-4618jv)*

jordanlewis:
:lgtm:

---

Reviewed 7 of 7 files at r2.

---

*[pkg/sql/parser/datum.go, line 612 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsudCu7DeFIB-I6jE5:-KxsudCu7DeFIB-I6jE6:byb1ff9) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/datum.go#L612)):*
> ```Go
> // ParseDDecimalVal parses and returns the DDecimal value represented by the
> // provided string, or an error if parsing is unsuccessful.
> func ParseDDecimalVal(s string) (DDecimal, error) {
> ```

I guess this was unused?

---

*[pkg/sql/parser/encode.go, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxssKRr2Uys9vz_eJUJ:-KxssKRr2Uys9vz_eJUK:bgqpjcm) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/sql/parser/encode.go#L75)):*
> ```Go
> 				bareStrings = false
> 			}
> 			if stringencoding.EncodeMap[ch] == stringencoding.DontEscape && ch != '\'' {
> ```

This is kind of weird that you have to use 2 package variables in one line like this (and again below) - i think it should be a function `stringencoding.NeedEscape(ch)`.

---

*[pkg/sql/parser/encode.go, line 111 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxssVCLZDO5WKBEv4DW:-KxssVCLZDO5WKBEv4DX:byr4ffv) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/sql/parser/encode.go#L111)):*
> ```Go
> 	for i, r := range in {
> 		ch := byte(r)
> 		if r >= 0x20 && r < 0x7F && stringencoding.EncodeMap[ch] == stringencoding.DontEscape && ch != '"' {
> ```

see above

---

*[pkg/sql/parser/encode.go, line 159 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxssm1eywhm3PVA8saK:-Kxssm1eywhm3PVA8saL:b-5u1hbh) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/sql/parser/encode.go#L159)):*
> ```Go
> 	for i, n := 0, len(in); i < n; i++ {
> 		ch := in[i]
> 		if encodedChar := stringencoding.EncodeMap[ch]; encodedChar != stringencoding.DontEscape {
> ```

oo another one!

---

*[pkg/util/json/json.go, line 92 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsvAGjU0Kr3LRdcuPC:-KxsvAGjU0Kr3LRdcuPD:b-igx6i1) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L92)):*
> ```Go
> func (jsonObject) jsonType() jsonType { return objectJSONType }
> 
> func cmpJSONTypes(a JSON, b JSON) int {
> ```

Having this thing accept `jsonType` and the callers be the ones to call `jsonType()` is cleaner I think - since the function is called `cmpJSONTypes`. Also, it's more efficient to do it that way, since the runtime type of the left argument is already known at the callsites.

---

*[pkg/util/json/json.go, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsutnmhFiFlm62TYNL:-KxsutnmhFiFlm62TYNM:b7v2afu) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L213)):*
> ```Go
> // Very similar to encodeSQLStringInsideArray. Primary difference is that it is
> // legal to directly print out unicode characters.
> func encodeJSONString(buf *bytes.Buffer, in string) {
> ```

Is the unicode logic in this guy sufficiently tested?

---

*[pkg/util/json/json.go, line 218 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxstN3cf-b2x90J4JD8:-KxstN3cf-b2x90J4JD9:b1fcw3u) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L218)):*
> ```Go
> 	for i, r := range in {
> 		ch := byte(r)
> 		if unicode.IsPrint(r) && stringencoding.EncodeMap[ch] == stringencoding.DontEscape && ch != '"' {
> ```

here's another one

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-:-KxsvRBNfozLNVRcb14G:bnfp4nl)*

justinj:
Review status: 2 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[pkg/sql/parser/datum.go, line 612 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsudCu7DeFIB-I6jE5:-Kxt6dFz7ZB38Q1Cvy3o:b-dop14r) ([raw file](https://github.com/cockroachdb/cockroach/blob/fecfe9d5593be0568a22159bfd7d130ed6804b6c/pkg/sql/parser/datum.go#L612)):*

I guess this was unused?
</blockquote></details>

I added it for use in `interpretDJSON` but it's unused now that that uses `apd.Decimal` directly instead of `DDecimal`.

---

*[pkg/sql/parser/encode.go, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxssKRr2Uys9vz_eJUJ:-Kxt78vzazklExaqF5Dr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/sql/parser/encode.go#L75)):*

This is kind of weird that you have to use 2 package variables in one line like this (and again below) - i think it should be a function `stringencoding.NeedEscape(ch)`.
</blockquote></details>

Done.

---

*[pkg/sql/parser/encode.go, line 111 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxssVCLZDO5WKBEv4DW:-Kxt7Ec5LkGJwk-OlzK5:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/sql/parser/encode.go#L111)):*

see above
</blockquote></details>

Done.

---

*[pkg/sql/parser/encode.go, line 159 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-Kxssm1eywhm3PVA8saK:-Kxt7F62XDta3zy3jiej:bj6sqf8) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/sql/parser/encode.go#L159)):*

oo another one!
</blockquote></details>

I think this one is easier to just leave this way, since otherwise we have to do a second lookup (without restructuring everything).

---

*[pkg/util/json/json.go, line 92 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsvAGjU0Kr3LRdcuPC:-KxtANlWMUsf65FTbaTN:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L92)):*

Having this thing accept `jsonType` and the callers be the ones to call `jsonType()` is cleaner I think - since the function is called `cmpJSONTypes`. Also, it's more efficient to do it that way, since the runtime type of the left argument is already known at the callsites.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsutnmhFiFlm62TYNL:-KxtAsiTiAtcMuyWibss:b-oz616m) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L213)):*

Is the unicode logic in this guy sufficiently tested?
</blockquote></details>

There's a unicode test in the node acceptance tests, it could maybe do with some more thorough testing though... I added some tests involving unicode to `TestJSONRoundTrip`, I'm not sure how to be more extensive.

---

*[pkg/util/json/json.go, line 218 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxstN3cf-b2x90J4JD8:-KxtBfoiLro6AVcISmUx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L218)):*

here's another one
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700)*

jordanlewis:
Reviewed 1 of 7 files at r2, 4 of 4 files at r3.

---

*[pkg/util/json/json.go, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700#-KxsutnmhFiFlm62TYNL:-KxtXi_XyzVcxNCKYplE:ba5ydg0) ([raw file](https://github.com/cockroachdb/cockroach/blob/b79b09b239816497ee61da6f2bc0fd4c6a9fc230/pkg/util/json/json.go#L213)):*

There's a unicode test in the node acceptance tests, it could maybe do with some more thorough testing though... I added some tests involving unicode to `TestJSONRoundTrip`, I'm not sure how to be more extensive.
</blockquote></details>

Just wanted to make sure the logic for escaping vs non escaping of unicode is covered. Sounds like it is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19700)*

PR #19701, opened 2017-11-01
sql: removed RowFetcher and replaced with MultiRowFetcher

richardwu:
In https://github.com/cockroachdb/cockroach/pull/19228, `MultiRowFetcher` was "forked" from `RowFetcher`. `MultiRowFetcher` permitted specifying multiple tables (indexes to be precise) from which to retrieve rows. It returns a `RowResponse` which includes the `row EncDatumRow`, the `TableDescriptor` and `IndexDescriptor` that corresponds to the index from which the row was scanned.

In the simple 1 table case, `MultiRowFetcher` is logically and for the most part equivalent to `RowFetcher`. 

A simple micro-benchmark with `TableReader` shows that there is trivial overhead
```
name           old time/op    new time/op    delta
TableReader-8    11.6ms ± 3%    11.6ms ± 2%   ~     (p=0.720 n=10+9)

name           old alloc/op   new alloc/op   delta
TableReader-8    7.88MB ± 0%    7.88MB ± 0%   ~     (p=0.481 n=10+10)

name           old allocs/op  new allocs/op  delta
TableReader-8      418 ±126%      422 ±127%   ~     (p=0.288 n=10+10)
````
`InterleavedSelect` SQL benchmark shows similar results
```
name                                          old time/op    new time/op    delta
SQL/Cockroach/InterleavedSelect/count=1000-8    2.98ms ± 9%    2.94ms ± 4%    ~     (p=0.481 n=10+10)

name                                          old alloc/op   new alloc/op   delta
SQL/Cockroach/InterleavedSelect/count=1000-8    1.50MB ± 0%    1.50MB ± 0%  +0.07%  (p=0.000 n=10+10)

name                                          old allocs/op  new allocs/op  delta
SQL/Cockroach/InterleavedSelect/count=1000-8     6.33k ± 1%     6.34k ± 0%    ~     (p=0.343 n=10+10)
```

cc: @rytaft might want to rebase if you're touching more of `RowFetcher`.

### CODE REVIEW

**Comment by jordanlewis** [comment]
Very excited to see this! Can you include the before/after results with `-benchmem` as well? Add that flag to `TESTFLAGS` in  `scripts/bench`.

---

Review status: 0 of 15 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-:-Kxs25wh0q6GxTMH2rc-:boc7xrd)*

**Comment by richardwu** [comment]
Et voila
```
name           old time/op    new time/op    delta
TableReader-8    11.6ms ± 3%    11.6ms ± 2%   ~     (p=0.720 n=10+9)

name           old alloc/op   new alloc/op   delta
TableReader-8    7.88MB ± 0%    7.88MB ± 0%   ~     (p=0.481 n=10+10)

name           old allocs/op  new allocs/op  delta
TableReader-8      418 ±126%      422 ±127%   ~     (p=0.288 n=10+10)
```

---

Review status: 0 of 15 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-:-Kxs3JcvNXjCdIV7VwNU:b-z1wauc)*

File: scripts/bench [hunk 1]
```diff
@@ -15,7 +15,7 @@ echo "Writing to ${dest}"
 
 for branch in "${2}" "${1}"; do
   git checkout "${branch}"
-  make bench PKG="${PKG}" BENCHES="${BENCHES}" TESTFLAGS="-count 10" > "${dest}/bench.${branch}" 2> "${dest}/log.txt"
+  make bench PKG="${PKG}" BENCHES="${BENCHES}" TESTFLAGS="-count 10 -benchmem" > "${dest}/bench.${branch}" 2> "${dest}/log.txt"
```
benesch: Would be great if you could split this into its own commit with some rationale. Only if you need go through CI again for something else, though.

RaduBerinde:
*[pkg/sql/scan.go, line 189 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KxtK-3MSzQq0lozahUp:-KxtK-3MSzQq0lozahUq:bkz8sej) ([raw file](https://github.com/cockroachdb/cockroach/blob/7d93e681b1041817aaa8e77c9cc411afae802d1b/pkg/sql/scan.go#L189)):*
> ```Go
> 	// We fetch one row at a time until we find one that passes the filter.
> 	for n.hardLimit == 0 || n.rowIndex < n.hardLimit {
> 		resp, err := n.fetcher.NextRowDecoded(params.ctx)
> ```

[nit] If NextRow/NextRowDecoded would return the results as separate arguments instead of in a struct, it would be easier to ignore the unused stuff.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

richardwu:
Review status: 0 of 24 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/scan.go, line 189 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KxtK-3MSzQq0lozahUp:-KxtfMXZgFd3a63QzCbo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7d93e681b1041817aaa8e77c9cc411afae802d1b/pkg/sql/scan.go#L189)):*

[nit] If NextRow/NextRowDecoded would return the results as separate arguments instead of in a struct, it would be easier to ignore the unused stuff.
</blockquote></details>

Done.

---

*[scripts/bench, line 18 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-Kxs3q6wq90ovGNyM6Tf-r1-18:-KxtfJWssZz2uakY3Jrg:b-dzg07c) ([raw file](https://github.com/cockroachdb/cockroach/blob/c5bfd92ff761fa957373b0f970b490063a30b2e8/scripts/bench#L18)):*

Would be great if you could split this into its own commit with some rationale. Only if you need go through CI again for something else, though.
</blockquote></details>

Good point :+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

arjunravinarayan:
:lgtm_strong:

---

Reviewed 2 of 16 files at r1, 15 of 19 files at r5, 16 of 16 files at r7.

---

*[pkg/sql/distsqlrun/tablereader_test.go, line 248 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KyqVvaS1Jxwr-DVsFb-:-KyqVvaS1Jxwr-DVsFb0:byzub06) ([raw file](https://github.com/cockroachdb/cockroach/blob/153ecbada7907a35e5ff77b7e4f06bd7e2fd61f8/pkg/sql/distsqlrun/tablereader_test.go#L248)):*
> ```Go
> }
> 
> func BenchmarkTableReader(b *testing.B) {
> ```

Not saying you should do this now, but next time you could probably adapt the existing schemas set up in `TestTableReader` to benchmark them all without having to rewrite a bunch of code, and you'd get more benchmarks as a result. This isn't a criticism, just a tip, as this is great as-is. 

---

*[pkg/util/fast_int_set_test.go, line 210 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-Kyq_LdGylynZIlAcwse:-Kyq_LdGylynZIlAcwsf:bt0hcqm) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8d9de672c6237c8d752ffc9005985bc18176472/pkg/util/fast_int_set_test.go#L210)):*
> ```Go
> 	checkBuffer := 100
> 
> 	for _, n := range []int{0, 1, 20, 32, 42, 62, 63, 64, 100} {
> ```

I'd test the entirety of the `smallCutoff`, and then the random set that you already have above and beyond. It's too tiny to not test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-:-KyqaCYCVdjdS7kHzn8z:b-jbsf8z)*

richardwu:
Review status: 22 of 23 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlrun/tablereader_test.go, line 248 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KyqVvaS1Jxwr-DVsFb-:-KyqivOne_saxoLDkplt:b-bc2t3p) ([raw file](https://github.com/cockroachdb/cockroach/blob/153ecbada7907a35e5ff77b7e4f06bd7e2fd61f8/pkg/sql/distsqlrun/tablereader_test.go#L248)):*

Not saying you should do this now, but next time you could probably adapt the existing schemas set up in `TestTableReader` to benchmark them all without having to rewrite a bunch of code, and you'd get more benchmarks as a result. This isn't a criticism, just a tip, as this is great as-is. 
</blockquote></details>

Good point, duly noted!

---

*[pkg/util/fast_int_set_test.go, line 210 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-Kyq_LdGylynZIlAcwse:-KyqlyDe8o0Av86X0rXr:b-ytryfo) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8d9de672c6237c8d752ffc9005985bc18176472/pkg/util/fast_int_set_test.go#L210)):*

I'd test the entirety of the `smallCutoff`, and then the random set that you already have above and beyond. It's too tiny to not test.
</blockquote></details>

I've modified this to test everything up to `smallCutoff + 100` which should be sufficiently fast. Also changed the way the test checks for the values (instead of `Contains`, it iterates through every value with `ForEach`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

RaduBerinde:
*[pkg/util/fast_int_set.go, line 86 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-Kyr9x8wOp9QhOgwW_Kx:-Kyr9x8wOp9QhOgwW_Ky:b-eux91b) ([raw file](https://github.com/cockroachdb/cockroach/blob/c353d0bc7902dd0cef4a0f82aaed73078900ac62/pkg/util/fast_int_set.go#L86)):*
> ```Go
> 
> // AddUpTo adds values 0 up to n (inclusively) to the set.
> func (s *FastIntSet) AddUpTo(n int) {
> ```

[nitty] I don't like the hardcoding of 0. Existing APIs are not prescriptive on what kind of indexing the caller should use. I think we should make this more general, eg `AddRange(from, to int)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

richardwu:
Review status: 16 of 23 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/util/fast_int_set.go, line 86 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-Kyr9x8wOp9QhOgwW_Kx:-KyrQvw627dsoW0AqN5x:bv9tpl2) ([raw file](https://github.com/cockroachdb/cockroach/blob/c353d0bc7902dd0cef4a0f82aaed73078900ac62/pkg/util/fast_int_set.go#L86)):*

[nitty] I don't like the hardcoding of 0. Existing APIs are not prescriptive on what kind of indexing the caller should use. I think we should make this more general, eg `AddRange(from, to int)`
</blockquote></details>

I like it, fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

RaduBerinde:
*[pkg/util/fast_int_set.go, line 89 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KyrRHbD9rC1xP46pA3W:-KyrRHbEDuqTYFYpPupm:b-ou1onl) ([raw file](https://github.com/cockroachdb/cockroach/blob/3f15dff9ac1bd350f1e337784a9e84f8f115636e/pkg/util/fast_int_set.go#L89)):*
> ```Go
> // 'from' must be >= 0 and 'to' must be >= 'from'.
> func (s *FastIntSet) AddRange(from, to int) {
> 	if from < 0 {
> ```

We can.. We don't need to restrict this (just make sure we're in the "large" case)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

richardwu:
Review status: 16 of 23 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[pkg/util/fast_int_set.go, line 89 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KyrRHbD9rC1xP46pA3W:-KyrWxoPpXE5B-_zEZU0:bfv3b4q) ([raw file](https://github.com/cockroachdb/cockroach/blob/3f15dff9ac1bd350f1e337784a9e84f8f115636e/pkg/util/fast_int_set.go#L89)):*

We can.. We don't need to restrict this (just make sure we're in the "large" case)
</blockquote></details>

Oops, good catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

RaduBerinde:
*[pkg/util/fast_int_set.go, line 89 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701#-KyrRHbD9rC1xP46pA3W:-KysQmS1F3yQFnSeQhmw:b2c0zcb) ([raw file](https://github.com/cockroachdb/cockroach/blob/3f15dff9ac1bd350f1e337784a9e84f8f115636e/pkg/util/fast_int_set.go#L89)):*

Oops, good catch.
</blockquote></details>

LGTM, thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19701)*

PR #19713, opened 2017-11-01
sqlccl: use checkpoint to lock out concurrent ops

dt:
Accidentially starting concurrent backups pointed at the same location
would result in them clobbering eachother. While we already bail if a
finished backup has written a BACKUP file, we still have a relatively
large window while it is in progress.

This switches to check for both the finished BACKUP file and the
BACKUP-CHECKPOINT file that may exist while the backup is still running,
and ensures it *does* exist by writting an empty one immediately after
checking for it.

This cuts the window during which a concurrent op could start to just
the check-then-set round-trip delay.

Additionally, it verifies that we can _write_ to the designated location
*before* sending potentially expensive work out that would ultimately
be wasted if the target isn't writable.

### DIFF
--- a/pkg/ccl/sqlccl/backup.go
+++ b/pkg/ccl/sqlccl/backup.go
@@ -412,7 +412,6 @@ func backup(
 		files          []BackupDescriptor_File
 		exported       roachpb.BulkOpSummary
 		lastCheckpoint time.Time
-		checkpointed   bool
 	}{}
 
 	var checkpointMu syncutil.Mutex
@@ -434,11 +433,16 @@ func backup(
 		// TODO(benesch): verify these files, rather than accepting them as truth
 		// blindly.
 		// No concurrency yet, so these assignments are safe.
-		mu.checkpointed = true
 		mu.files = checkpointDesc.Files
 		mu.exported = checkpointDesc.EntryCounts
 		for _, file := range checkpointDesc.Files {
 			completedSpans = append(completedSpans, file.Span)
+		}
+	}
+
+	cleanupCheckpoint := func() {
+		if err := exportStore.Delete(ctx, BackupDescriptorCheckpointName); err != nil {
+			log.Warningf(ctx, "unable to delete checkpointed backup descriptor: %+v", err)
 		}
 	}
 
@@ -454,7 +458,12 @@ func backup(
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
-	if err := job.Created(ctx, cancel); err != nil {
+	cancelFn := func() {
+		cancel()
+		// this fn closes over the non-cancelled ctx above for its call to Delete.
+		cleanupCheckpoint()
+	}
+	if err := job.Created(ctx, cancelFn); err != nil {
 		return err
 	}
 	if err := job.Started(ctx); err != nil {
@@ -540,9 +549,6 @@ func backup(
 				checkpointMu.Unlock()
 				if err != nil {
 					log.Errorf(ctx, "unable to checkpoint backup descriptor: %+v", err)
-					mu.Lock()
-					mu.checkpointed = true
-					mu.Unlock()
 				}
 			}
 			return nil
@@ -550,6 +556,9 @@ func backup(
 	}
 
 	if err := g.Wait(); err != nil {
+		if jobs.IsCancellation(err) {
+			cleanupCheckpoint()
+		}
 		return errors.Wrapf(err, "exporting %d ranges", len(spans))
 	}
 
@@ -561,13 +570,37 @@ func backup(
 	if err := writeBackupDescriptor(ctx, exportStore, BackupDescriptorName, backupDesc); err != nil {
 		return err
 	}
+	cleanupCheckpoint()
 
-	if mu.checkpointed {
-		if err := exportStore.Delete(ctx, BackupDescriptorCheckpointName); err != nil {
-			log.Warningf(ctx, "unable to delete checkpointed backup descriptor: %+v", err)
-		}
+	return nil
+}
+
+// verifyUsableExportTarget ensures that the target location does not already
+// contain a BACKUP or checkpoint and writes an empty checkpoint, both verifying
+// that the location is writable and locking out accidental concurrent
+// operations on that location if subsequently try this check. Callers must
+// clean up the written checkpoint file (BackupDescriptorCheckpointName) only
+// after writing to the backup file location (BackupDescriptorName).
+func verifyUsableExportTarget(
+	ctx context.Context, exportStore storageccl.ExportStorage, readable string,
+) error {
+	if r, err := exportStore.ReadFile(ctx, BackupDescriptorName); err == nil {
+		// TODO(dt): If we audit exactly what not-exists error each ExportStorage
+		// returns (and then wrap/tag them), we could narrow this check.
+		r.Close()
+		return errors.Errorf("%s already contains a %s file",
+			readable, BackupDescriptorName)
 	}
-
+	if r, err := exportStore.ReadFile(ctx, BackupDescriptorCheckpointName); err == nil {
+		r.Close()
+		return errors.Errorf("%s already contains a %s file (is another operation already in progress?)",
+			readable, BackupDescriptorCheckpointName)
+	}
+	if err := writeBackupDescriptor(
+		ctx, exportStore, BackupDescriptorCheckpointName, &BackupDescriptor{},
+	); err != nil {
+		return errors.Wrapf(err, "cannot write to %s", readable)
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


dt:
Added cleanup of CHECKPOINT on cancel.

mjibson:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/ccl/sqlccl/backup.go, line 463 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-Kxsovrzwl6WNw4-yuyv:-Kxsovs-ZJC650Mj-ouR:b4s0ils) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L463)):*
> ```Go
> 	cancelFn := func() {
> 		cancel()
> 		cleanupCheckpoint()
> ```

It is slightly confusing that the ctx in the cleanupCheckpoint function is different than the ctx three lines above. This is important because the Delete call will fail (I assume) if the ctx is already canceled. The `cancel()` call one line above would seem to do that, but doesn't because ctx is shadowed. I guess this is fine, but it confused me a bit.

---

*[pkg/ccl/sqlccl/backup.go, line 558 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxspVVtCJz7iLNhkRPv:-KxspVVtCJz7iLNhkRPw:b3x5ume) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L558)):*
> ```Go
> 
> 	if err := g.Wait(); err != nil {
> 		if jobs.IsCancellation(err) {
> ```

Why do we care about deleting the checkpoint file on cancellation?

---

*[pkg/ccl/sqlccl/backup.go, line 595 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxsqZmV137YwL5_m_--:-KxsqZmV137YwL5_m_-0:beg1in8) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L595)):*
> ```Go
> 	if r, err := exportStore.ReadFile(ctx, BackupDescriptorCheckpointName); err == nil {
> 		r.Close()
> 		return errors.Errorf("Is another operation already in progress? %s contains a %s file",
> ```

should we make these error messages the same? also this one begins with a capital letter and is a question. we don't have a lot of errors like that. also ok just leaving it because it's already there.

---

*[pkg/ccl/sqlccl/backup.go, line 846 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxsrkdOR3jDlObZcABf:-KxsrkdOR3jDlObZcABg:blceg5y) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L846)):*
> ```Go
> 			// If the checkpoint is from a different cluster, it's meaningless to us.
> 			// More likely though are dummy/lock-out checkpoints with no ClusterID.
> 			if desc.ClusterID.Equal(job.ClusterID()) {
> ```

I don't understand the purpose of this. Why is it ok to not set checkpointDesc if the cluster IDs aren't equal?

---

*[pkg/ccl/sqlccl/backup_test.go, line 821 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxssQqR1oHr59J2b-o0:-KxssQqS7axA33v49aMP:b-8bhd6x) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup_test.go#L821)):*
> ```Go
> 
> 	const numAccounts = 1000
> 	_, dir, _, outerDB, cleanup := backupRestoreTestSetupWithParams(t, multiNode, numAccounts, initNone, params)
> ```

What is this change for? So one subtest can fail without failing all the others?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713)*

dt:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/ccl/sqlccl/backup.go, line 463 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-Kxsovrzwl6WNw4-yuyv:-KxsvUCAx0VSTqaNBtpJ:b-cj6kmv) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L463)):*

It is slightly confusing that the ctx in the cleanupCheckpoint function is different than the ctx three lines above. This is important because the Delete call will fail (I assume) if the ctx is already canceled. The `cancel()` call one line above would seem to do that, but doesn't because ctx is shadowed. I guess this is fine, but it confused me a bit.
</blockquote></details>

We don't actually want the canceled context used for the delete -- if we're trying to delete after a cancellation, we don't want to skip it (though in reality I don't actually think any of the impls check the ctx during delete :/). 

---

*[pkg/ccl/sqlccl/backup.go, line 558 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxspVVtCJz7iLNhkRPv:-KxsupAgxWBlrjx1MbLQ:b-hn9rk7) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L558)):*

Why do we care about deleting the checkpoint file on cancellation?
</blockquote></details>

The existence of the checkpoint is used to signal a running (or paused) job has exclusive use of the export location. If you leave it after a cancel, the location remains unusable and requires manual cleanup -- not the end of the world, but still somewhat annoying. Particularly once we have saved keys, the user might not even have access to do the manual cleanup.

---

*[pkg/ccl/sqlccl/backup.go, line 846 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxsrkdOR3jDlObZcABf:-KxstCmSop6EpXofSSkI:b-hfxjpg) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L846)):*

I don't understand the purpose of this. Why is it ok to not set checkpointDesc if the cluster IDs aren't equal?
</blockquote></details>

it's okay to not set it -- passing nil just means we don't have any prior work to resume from. Indeed, if we can't read it (doesn't exist yet or whatever), we already pass nil.

I added the check to ignore the empty checkpoint written above, though, based on the way we currently use the checkpoint desc, it wouldn't actually hurt anything to pass an empty one: while the non-nil check would no longer filter it, the empty files fields wouldn't hurt anything. However, if we ever changed to use the other fields, this could be confusing, so it seemed safer to pass the explicitly handled nil instead.

and if the cluster IDs aren't equal on a non-empty checkpoint, then any prior work recorded in the checkpoint is meaningless to this cluster and _should_ be discarded, so the check seems like a good idea either way.

---

*[pkg/ccl/sqlccl/backup_test.go, line 821 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-KxssQqR1oHr59J2b-o0:-Kxssms_8DgOUe_A5T_9:b-koeymz) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup_test.go#L821)):*

What is this change for? So one subtest can fail without failing all the others?
</blockquote></details>

Unfortunately it is really easy to accidentally call the outer `testing.T` from inside a subtest when you close over it in a sqlrunner like this, if you don't remember to make a new `sqlrunner` in each subtest with the right `testing.t`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713)*

mjibson:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/ccl/sqlccl/backup.go, line 463 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-Kxsovrzwl6WNw4-yuyv:-Kxsw_8MIadeny1IOQQG:bde32bq) ([raw file](https://github.com/cockroachdb/cockroach/blob/fa46055da53dfd788f70f8310a6229094eead03e/pkg/ccl/sqlccl/backup.go#L463)):*

We don't actually want the canceled context used for the delete -- if we're trying to delete after a cancellation, we don't want to skip it (though in reality I don't actually think any of the impls check the ctx during delete :/). 
</blockquote></details>

Yes, that was my point. It took me a while to understand that the context sent to Delete would be the non-cancelled one, even though it is invoked after the call to `cancel()` because it is shadowed. I just think the code is confusing because of reuse of the name ctx. We have this all over the codebase, and we sometimes use `gCtx` for errgroups, for example, to make it explicit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-:-KxsxLplqQQ840pu9ZTF:bnfp4nl)*

dt:
Added comment about the closed over ctx, and reworded the error messages.

mjibson:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19713#-:-KxuYVLIMDACu_3sZld5:bnfp4nl)*

PR #19714, opened 2017-11-01
sql: Plumb RangeDescriptorCache and LeaseHolderCache to distSQLBackfill

lgo:
This plumbs through caches to the distSQLReceiver initialized during
schema change backfilling. These caches are updated during distSQL
execution.

This will prevent a panic if the spans during distSQL execution
do not match the planned spans. It is not currently known if this was a
problem, but panics are happening on another code path where a
distSQLReceiver is initialized during query execution (on #19586).


andreimatei:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19714#-:-KxseKz05zfRTJ6X3mOt:bnfp4nl)*

PR #19718, opened 2017-11-01
Change New Version alert link to upgrade doc

sploiselle:
Instead of linking to the install doc when we detect a cluster's running an old version of `cockroach`, we should link to the doc that shows them how to upgrade. 

This should ultimately link to a version that you know you can upgrade to, but because we're avoiding dealing with the Gordian knot of versions, I suggest we just improving the current behavior by leaving this as a link to `stable`.


couchand:
Thanks for taking care of this, that's a much better place to link to.  Content LGTM.  Please take a look at the commit message guidelines in [our contributing guide](https://github.com/cockroachdb/cockroach/blob/master/CONTRIBUTING.md) (search for "Good Commit Message"... there's not a direct link (there should be)).

couchand (approved):

PR #19720, opened 2017-11-01
sql: fix recursive use of prepare and execute

### LINKED ISSUE #19716: sql: prepare/execute and query datasources don't work together correctly
Observe the following sketchy interaction:

```
root@:26257/test> prepare u as SELECT $1:::int i, 'foo' t;
PREPARE

Time: 424.189µs

root@:26257/test> prepare v as select * from [execute u(3)] where t=$1;
pq: unsupported comparison operator: <string> = <int>
```

The problem here is evidently that the placeholder type information for the inner query overwrites that of the outer query. Inner queries of this form need to properly preserve the state of placeholders in the outer queries and restore them after they're finished.

cc @knz 

Hoping to roll a fix in for this in with the leaf placeholders PR sequence related to #19424.

jordanlewis:
Previously, calling `PREPARE` on a statement that included an `EXECUTE`
in a statement datasource (the `SELECT * FROM [...]` syntax) was buggy.
It did not properly segregate the placeholders from the inner statement
from those of the outer statement.

Fixes #19716.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/prepare
+++ b/pkg/sql/logictest/testdata/logic_test/prepare
@@ -334,7 +334,7 @@ query error pq: job with ID 123 does not exist
 EXECUTE p6(123)
 
 # Ensure that SET / SET CLUSTER SETTING know about placeholders
-statement
+statement ok
 PREPARE setp(string) AS SET application_name = $1
 
 query T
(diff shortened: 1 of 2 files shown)


knz:
it's interestingly simple (I like that) but I think the tests are saying it's not sufficient :)

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19720#-:-KxwxLVK0OX9JgOJQfS0:bkslemn)*

jordanlewis:
Yeah... the problem is that internal planner is insufficient for most tasks. This is a thread to a larger point of confusion in the SQL code base - there are lots of ways to make "internal queries"!

It's not obvious from any particular point of the code whether one should use planner's QueryRow methods directly, make an internal planner, or use an internal executor.

What I think we need is a concept of a subplanner or planner frames for placeholder (and indexed var? but I won't go there for now) bindings. Going down a frame happens when you run an internal query or statement with placeholders, and it permits executing a statement with fresh bindings, restoring the old bindings when the plan completes.

But I'm also nervous about introducing yet another way of executing queries internally. I'd like to brainstorm about this issue with you at some point soon.

jordanlewis:
Another way to fix this is to modify all the planner's QueryRow methods to do this directly, and get rid of internal planner altogether. I'm not exactly sure of the purpose of the internal planner - but it seems to have some strange behavior. Specifically I'd like to hilight that internal planners have separate (and unlimited) memory accounts.

I don't really understand that - when and why would it be okay to give internal queries issued by user queries their own separate memory accounts? Isn't it more accurate to have those internal queries *share* the memory account of the outer query? Everything downstream of a query should probably be sharing a memory account so we have proper attribution of memory use in the system.

jordanlewis:
PTAL, just did the simple thing for now

knz:
Simple thing doesn't seem to bad actually. Clever. Thanks!
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19720#-:-Ky5Qi8mEFt2dogBSE30:b-kdzrex)*

jordanlewis:
TFTR!

PR #19724, opened 2017-11-01
build: File issues when jepsen tests fail

### LINKED ISSUE #19694: jepsen: File issues on failures
Currently, jepsen failures are not tracked anywhere and rely on someone manually watching the logs. This should file an issue instead, similar to the nightly stress runner and other CI jobs.

bdarnell:
Release note: None

Fixes #19694 

What's the best way to test this without the risk of spamming everyone in an overnight run?


tbg (approved):
I've only skimmed this. You can test this by sourcing the secret env locally and running a slightly modified version that posts against your fork and not the main repo.

We should also remind ourselves to check after this merges that the nightly stress tests and failed master merges still post. The nightlies are guaranteed to fail at least 2 tests every night, so I doubt we'll miss it.

benesch (approved):
LGTM too

bdarnell:
This was replaced by #19766.

PR #19725, opened 2017-11-01
storageccl,cli: add external IO dir cli flag

dt:
okay, still need to test the CLI side of this one (need to figure out how to do that in ccl tree), but wanted to get some early feedback and see what the CI thinks on all this plumbing.

### DIFF
--- a/pkg/base/test_server_args.go
+++ b/pkg/base/test_server_args.go
@@ -69,6 +69,9 @@ type TestServerArgs struct {
 	// If not initialized, will default to DefaultTestTempStorageConfig.
 	TempStorageConfig TempStorageConfig
 
+	// ExternalIODir is used to initialize field in cluster.Settings.
+	ExternalIODir string
+
 	// Fields copied to the server.Config.
 	Insecure                 bool
 	RetryOptions             retry.Options
--- a/pkg/ccl/sqlccl/backup_cloud_test.go
+++ b/pkg/ccl/sqlccl/backup_cloud_test.go
@@ -55,7 +55,7 @@ func TestCloudBackupRestoreS3(t *testing.T) {
 	defer sql.TestDisableTableLeases()()
 	const numAccounts = 1000
 
-	ctx, _, _, sqlDB, cleanupFn := backupRestoreTestSetup(t, 1, numAccounts, initNone)
+	ctx, _, tc, _, cleanupFn := backupRestoreTestSetup(t, 1, numAccounts, initNone)
 	defer cleanupFn()
 	prefix := fmt.Sprintf("TestBackupRestoreS3-%d", timeutil.Now().UnixNano())
 	uri := url.URL{Scheme: "s3", Host: bucket, Path: prefix}
@@ -64,7 +64,7 @@ func TestCloudBackupRestoreS3(t *testing.T) {
 	values.Add(storageccl.S3SecretParam, creds.SecretAccessKey)
 	uri.RawQuery = values.Encode()
 
-	backupAndRestore(ctx, t, sqlDB, uri.String(), numAccounts)
+	backupAndRestore(ctx, t, tc, uri.String(), numAccounts)
 }
 
 // TestBackupRestoreGoogleCloudStorage hits the real GCS and so could
@@ -88,11 +88,11 @@ func TestCloudBackupRestoreGoogleCloudStorage(t *testing.T) {
 	}(http.DefaultTransport.(*http.Transport).DisableKeepAlives)
 	http.DefaultTransport.(*http.Transport).DisableKeepAlives = true
 
-	ctx, _, _, sqlDB, cleanupFn := backupRestoreTestSetup(t, 1, numAccounts, initNone)
+	ctx, _, tc, _, cleanupFn := backupRestoreTestSetup(t, 1, numAccounts, initNone)
 	defer cleanupFn()
 	prefix := fmt.Sprintf("TestBackupRestoreGoogleCloudStorage-%d", timeutil.Now().UnixNano())
 	uri := url.URL{Scheme: "gs", Host: bucket, Path: prefix}
-	backupAndRestore(ctx, t, sqlDB, uri.String(), numAccounts)
+	backupAndRestore(ctx, t, tc, uri.String(), numAccounts)
 }
 
 // TestBackupRestoreAzure hits the real Azure Blob Storage and so could
@@ -120,7 +120,7 @@ func TestCloudBackupRestoreAzure(t *testing.T) {
 	}(http.DefaultTransport.(*http.Transport).DisableKeepAlives)
 	http.DefaultTransport.(*http.Transport).DisableKeepAlives = true
 
-	ctx, _, _, sqlDB, cleanupFn := backupRestoreTestSetup(t, 1, numAccounts, initNone)
+	ctx, _, tc, _, cleanupFn := backupRestoreTestSetup(t, 1, numAccounts, initNone)
 	defer cleanupFn()
 	prefix := fmt.Sprintf("TestBackupRestoreAzure-%d", timeutil.Now().UnixNano())
 	uri := url.URL{Scheme: "azure", Host: bucket, Path: prefix}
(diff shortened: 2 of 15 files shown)


mjibson:
Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/storageccl/export_storage.go, line 227 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-KxuaOhL-xpCqeexjOhv:-KxuaOhL-xpCqeexjOhw:b-cdfswd) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/ccl/storageccl/export_storage.go#L227)):*
> ```Go
> 			return nil, errors.Errorf("local file access is disabled")
> 		}
> 		if !strings.HasPrefix(base, settings.ExternalIODir) {
> ```

I think it is incorrect to have a per-node setting that a user is then required to know, and be the same across the entire cluster, for them to use this feature. I think we should instead require relative paths so that clusters can be configured however they need to be, possibly with differing settings, and the user's query won't have to change.

---

*[pkg/cli/start.go, line 410 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-Kxua6c1Q9tmNaD21t8Q:-Kxua6c1Q9tmNaD21t8R:bqekom1) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/cli/start.go#L410)):*
> ```Go
> 		externalIODir = filepath.Join(firstStore.Path, "extern")
> 	}
> 	if externalIODir == "" || externalIODir == "disabled" {
> ```

"disabled" looks like leftover from testing

---

*[pkg/cli/start.go, line 411 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-Kxua43Uf3PH_x0RuBr4:-Kxua43Uf3PH_x0RuBr5:b-1ngo2e) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/cli/start.go#L411)):*
> ```Go
> 	}
> 	if externalIODir == "" || externalIODir == "disabled" {
> 		panic(externalIODir)
> ```

debug leftover? maybe you already know about it...

---

*[pkg/cli/cliflags/flags.go, line 483 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-Kxu_W3UpMMiRo9dRrq-:-Kxu_W3UpMMiRo9dRrq0:b-rycpe5) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/cli/cliflags/flags.go#L483)):*
> ```Go
> 		Name: "external-io-dir",
> 		Description: `
> The directory path under which operations like BACKUP/RESTORE/IMPORT are allowed to read and write files.
> ```

May need to specify that this only applies to nodelocal and nfs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725)*

dt:
okay, I've added tests that this does what it says on the tin.

After discussion in person with mjibson though, I'm planning to move forward with not just restricting access to just the local dir, but rather prefixing all accessed paths with it -- as a cli-flag it is node-specific, it's just generally not something we want specified in a query that is executed across many nodes.

The biggest downside to automatically prefixing is that it could be confusing in the case where you say "backup db to file:///home/david/desktop/bar" and then don't find "bar" on your desktop. 

We discussed a few ways we might try to prevent this confusion:
- Disallow arbitrary paths (that could be mistaken to be real paths), and only allow the base name to be specified -- e.g. only a host field in our URIs and not a Path. This unfortunately disallows e.g. organizing backups into folders, which could be annoying, or even a roadblock for filesystems that don't do well with many files in a folder.
- Require some prefix, like `extern`, to make it obvious that this isn't a local path.
- Return an error explaining it, then dismiss the error with a setting.
In the end, we don't like any of these. We decided to just auto-prefix. If you try to _read_ that doesn't exist after prefixing, we'll make sure the error makes it clear that the file was searched for _in_ the external IO dir.

dt:
The switch to prefixing all local file paths with the external IO dir changes just about every single backup/restore/import test... @bdarnell @mjibson would you rather review that as a separate CL or add do this one?

mjibson:
Separate is my preference.

dt:
okay, it's here https://github.com/dt/cockroach/compare/extern...dt:extern-prefix, but I'll wait for this one to land then before opening a PR for it.

dt:
Review status: 0 of 15 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/ccl/storageccl/export_storage.go, line 227 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-KxuaOhL-xpCqeexjOhv:-KyIEwTwX2OAFtrUbRzN:b5gs27j) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/ccl/storageccl/export_storage.go#L227)):*

I think it is incorrect to have a per-node setting that a user is then required to know, and be the same across the entire cluster, for them to use this feature. I think we should instead require relative paths so that clusters can be configured however they need to be, possibly with differing settings, and the user's query won't have to change.
</blockquote></details>

As discussed, the big change to switch to relative IO paths will be a follow-up CL.

---

*[pkg/cli/start.go, line 410 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-Kxua6c1Q9tmNaD21t8Q:-KxwcgYXdA7d9Ha56qQC:bqcq4xi) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/cli/start.go#L410)):*

"disabled" looks like leftover from testing
</blockquote></details>

We'll want a way to completely disable local disk IO, so if "" defaults to first store dir, we'll need a placeholder value like "disabled". There's no risk of collisions with an intended path since actual paths must be absolute. Added a cli test that "disabled" prints the right message.

---

*[pkg/cli/start.go, line 411 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-Kxua43Uf3PH_x0RuBr4:-KxwcfFcEWeFMtdDeJ6S:bczcwj9) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/cli/start.go#L411)):*

debug leftover? maybe you already know about it...
</blockquote></details>

whoops, yes, thanks,

---

*[pkg/cli/cliflags/flags.go, line 483 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-Kxu_W3UpMMiRo9dRrq-:-KyIF5YDLdlpIl-9Dtuh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eaf3f9731828c19b352c9689657704d50309286/pkg/cli/cliflags/flags.go#L483)):*

May need to specify that this only applies to nodelocal and nfs.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725)*

mjibson:
:lgtm:

---

Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/ccl/storageccl/export_storage_test.go, line 224 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-KyKg4sn32KlN542AO5h:-KyKg4sn32KlN542AO5i:b-twajy7) ([raw file](https://github.com/cockroachdb/cockroach/blob/e2c34084f3637caaf9acef6a4d79686da77bfebb/pkg/ccl/storageccl/export_storage_test.go#L224)):*
> ```Go
> 
> 	testExportStore(t, dest, false)
> }
> ```

newline here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-:-KyKgyZHCCZAXibHUE3w:bnfp4nl)*

dt:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/storageccl/export_storage_test.go, line 224 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725#-KyKg4sn32KlN542AO5h:-KyLwOzApKZ53AkFgtIW:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e2c34084f3637caaf9acef6a4d79686da77bfebb/pkg/ccl/storageccl/export_storage_test.go#L224)):*

newline here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19725)*

PR #19728, opened 2017-11-01
sql: pre-evaluate arguments to set

### LINKED ISSUE #19727: sql: panic in SET caused by use of invalid evalContext
The arguments to SET get evaluated with an incomplete `evalContext`. This can panic the server. Here's an admittedly contrived example:

```
root@:26257/> SET application_name = current_timestamp()::string;
driver: bad connection
connection lost; opening new connection: all session settings will be lost
```

jordanlewis:
Run Eval() on all arguments passed to Set before beginning the Set plan.
This is necessary because the implementations of all of the different
variable setters don't get passed the proper EvalContext, and therefore
won't be able to properly resolve Placeholders once they become leaf
values.

This wouldn't be necessary if the Set implementations were able to get
the correct EvalContext, but doing that properly would require a larger
refactor.

Release note: fix a rare panic in `SET`.
Fixes #19727.

I'd like to cherrypick this into 1.1.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/set
+++ b/pkg/sql/logictest/testdata/logic_test/set
@@ -203,3 +203,7 @@ query T
 SHOW "time zone"
 ----
 UTC
+
+# Regression test for #19727 - invalid EvalContext used to evaluate arguments to set.
+statement ok
+SET APPLICATION_NAME = current_timestamp()::string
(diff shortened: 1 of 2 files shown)


justinj:
LGTM

knz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19728#-:-KxwYP1oAaOwZ2dGIt3f:bnfp4nl)*

PR #19729, opened 2017-11-01
Add 'uuid_generate_v4' alias for 'uuid_v4' builtin.

novabyte:
This is a very small changeset which adds `uuid_generate_v4()` as an alias for the `uuid_v4()` builtin. This improves compatibility with PostgreSQL's equivalent function in the `uuid-ossp` module: https://www.postgresql.org/docs/9.4/static/uuid-ossp.html

### DIFF
--- a/pkg/sql/parser/builtins.go
+++ b/pkg/sql/parser/builtins.go
@@ -1138,6 +1138,7 @@ CockroachDB supports the following flags:
 	},
 
 	"experimental_uuid_v4": {uuidV4Impl},
+	"uuid_generate_v4":     {uuidV4Impl},
 	"uuid_v4":              {uuidV4Impl},
 
 	"greatest": {


novabyte:
Also let me know if you want me to add tests. I just wasn't sure it was needed given the `uuidV4Impl` struct is already tested.

justinj:
Hey Chris, thanks for the PR!

Unfortunately our `uuid_v4` was added to Cockroach before we had dedicated `UUID` types, so it returns `BYTES` instead of `UUID`, which is not ideal (but we don't change it for backwards-compat). #19379 added a variant of this function which correctly returns a `UUID`. Would you mind making this an alias to that one instead? You should be able to pull out the definition in the same way it's been done with `uuidV4Impl`.

You'll also have to re-generate the docs with `make generate PKG=./docs/...` since you're adding a builtin (or CI will complain).

And tests would be welcome (but indeed maybe not necessary for this change), if you'd like to, the easiest way is probably to do something similar to the `uuid_v4` tests in `pkg/sql/logictest/testdata/logic_test/builtin_function`. You can use `pg_typeof` to verify that the returned value is indeed a `UUID`.

novabyte:
@justinj ok great. I'll give it another look tomorrow. Thanks for the pointers. 😄

justinj:
Hey @novabyte, gonna close this for the time being, but feel free to re-open it!

hazcod:
@justinj / @novabyte  : is this in cockroachdb? The function does not seem to exist: `select uuid_generate_v4();`

knz:
@hazcod for nwo the alias is called `gen_random_uuid()`. We haven't added `uuid_generate_v4()` yet.

hazcod:
Ohh, so it's non-compatible with postgres. :/ 
That function does not exist in postgres.

knz:
@hazcod also see issue #56014

knz:
> Ohh, so it's non-compatible with postgres

How so? It generated a v4 UUID all the same?

hazcod:
It means cockroach is not postgres compliant, I cannot re-use my queries on both.

novabyte:
@hazcod This is not quite true anymore. In PG 13 they added `gen_random_uuid()` so Postgres became compatible with CockroachDB with the built-in function before CRDB did with Postgres. 🤷‍♂️ 

https://www.postgresql.org/docs/13/functions-uuid.html

hazcod:
@novabyte my bad, I am using v12. 👍

PR #19732, opened 2017-11-01
testutils: create RunTrueAndFalse helper function

nvb:
This creates a `RunTrueAndFalse` function that runs a provided closure that
takes a boolean in a pair of subtests, first with the bool set to true and then
with the bool set to false. This is a very common pattern during testing so
using this function removes some repetitive code. More importantly, though, it
removes a level of indentation, which add up in tests with nested subtests.


knz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19732#-:-KxwZKJdyGKvqcqXdTM0:bnfp4nl)*

PR #19738, opened 2017-11-02
sqlccl: lock temp storage against concurrent IMPORTs

### LINKED ISSUE #19637: bulkio: verify target is writable before starting expensive jobs
Before starting a large write operation, we should ensure we can actually write to the destination before doing lots of expensive work, namely, the target storage for a BACKUP or the `temp` storage for an IMPORT. Currently, especially in IMPORT, hours of work may occur before the first attempt to write to storage.

We potentially could also use this canary write to drop to a placeholder/lock, which would prevent another instance of the same operation from interfering with ours if mistakenly started pointed at the same location (easy to do when copy/pasting e.g. the temp path for an import job, or hardcoding a backup path in a recurring job).

dt:
Import can take hours of processing CSV data before its first write to
its temp store. Ensuring the temp store is writable before starting the
operation avoids potentially wasting significant time and cluster
resources only to discover a typo or permissions issue.

This should also prevent accidental reuse of a temp store by multiple
IMPORT jobs, which would overwrite eachother's hard-won results.

Fixes #19637.

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -975,6 +975,15 @@ func importPlanHook(
 			return errors.Errorf("must provide a temporary storage location")
 		}
 
+		tempStorage, err := exportStorageFromURI(ctx, temp, p.ExecCfg().Settings)
+		if err != nil {
+			return err
+		}
+		defer tempStorage.Close()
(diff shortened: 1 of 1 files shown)


mjibson:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19738#-:-KxxqptyaxJOnH6zlBfw:bnfp4nl)*

PR #19741, opened 2017-11-02
sql: some cleanup to planner-internal queries

jordanlewis:
- remove usages of `planner.query` from outside of the planner
- rename `planner.query` to `makeInternalPlan`
- restore placeholders after executing internal queries

### DIFF
--- a/pkg/sql/information_schema.go
+++ b/pkg/sql/information_schema.go
@@ -870,13 +870,9 @@ func forEachUser(ctx context.Context, origPlanner *planner, fn func(username str
 	query := `SELECT username FROM system.users`
 	p := makeInternalPlanner("for-each-user", origPlanner.txn, security.RootUser, origPlanner.session.memMetrics)
 	defer finishInternalPlanner(p)
-	plan, err := p.query(ctx, query)
+	rows, err := p.queryRows(ctx, query)
 	if err != nil {
 		return nil
-	}
-	defer plan.Close(ctx)
-	if err := p.startPlan(ctx, plan); err != nil {
-		return err
 	}
 
 	// TODO(cuongdo/asubiotto): Get rid of root user special-casing if/when a row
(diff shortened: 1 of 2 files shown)


knz:
:lgtm: with potential perf nit

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[pkg/sql/planner.go, line 262 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19741#-KxxmfxQo-wZyhWA-fcG:-KxxmfxQo-wZyhWA-fcH:b-y42gvy) ([raw file](https://github.com/cockroachdb/cockroach/blob/e973f50ad68f76e2bb9824b67a984135c6d001eb/pkg/sql/planner.go#L262)):*
> ```Go
> ) ([]parser.Datums, error) {
> 	oldPlaceholders := p.semaCtx.Placeholders
> 	defer func() { p.semaCtx.Placeholders = oldPlaceholders }()
> ```

Like I've explained before, you should double check the generated code in particular what happens wrt allocations when you capture a variable from the surrounding scope.
I wouldn't be surprised if you need the following to get good code:
```
defer func(prev PlaceholderInfo) { p.semaCtx.Placeholders = prev }(p.semaCtx.Placeholders)
```

---

*[pkg/sql/planner.go, line 293 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19741#-Kxxmw-95h0_tdU-gZjc:-Kxxmw-95h0_tdU-gZjd:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/e973f50ad68f76e2bb9824b67a984135c6d001eb/pkg/sql/planner.go#L293)):*
> ```Go
> func (p *planner) exec(ctx context.Context, sql string, args ...interface{}) (int, error) {
> 	oldPlaceholders := p.semaCtx.Placeholders
> 	defer func() { p.semaCtx.Placeholders = oldPlaceholders }()
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19741#-:-Kxxn022xttA8zjGhhLH:bs0q4xb)*

jordanlewis:
TFTR - I checked with @nvanbenschoten's `goescape` script that no allocations are performed there.

andreimatei:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19741#-:-KxxsfVoxnRpk0qlDmri:bnfp4nl)*

PR #19744, opened 2017-11-02
sql/parser: move DEFAULT from ctext_expr to d_expr

danhhz:
This appears to have been done originally because DEFAULT can only
appear in a VALUES list within INSERT or on the right side of a SET.
Otherwise, these ctext_exprs would have been a_exprs.

Because of the way the final PARTITION BY syntax works, we need DEFAULT
to be in a_expr, so this commit moves it there and rips out the ctext
stuff.

ctext_expr, ctext_expr_list, and ctext_row are no longer used and so are
removed.

Release note: none. no user facing change


knz:
cc @jordanlewis

jordanlewis:
@danhhz could you explain why `DEFAULT` needs to live in `a_expr` for the final syntax to work?

This seems somewhat dangerous. Moving `DEFAULT` into `a_expr` (well, `d_expr` actually) causes `DEFAULT` to become available in many new places.

For example, `SELECT DEFAULT` now parses. This is okay because `DefaultVal`'s `TypeCheck` method throws a useful error that re-validates that the `DEFAULT` expression is living in an okay spot. Are there any situations in which we don't call `TypeCheck` on every expression in an expression tree? If so, a `DEFAULT` in that expression tree would be a bug.

---

Reviewed 4 of 4 files at r1.

---

*[pkg/sql/create.go, line 1149 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-Kxy1ei0trW1K62affcb:-Kxy1ei181Z596KWn0YI:bxi7dj5) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8bec8d166666ce6c51850af431bcb05792e996/pkg/sql/create.go#L1149)):*
> ```Go
> 	// Because of some parsing oddness, DEFAULT in the context of PARTITION BY
> 	// is parsed as parser.DefaultVal instead of parser.PartitionDefault. Fix it
> 	// up.
> ```

I was going to complain that this method seems like an awfully generic place to put this unconditional transformation, until I realized that `valueEncodeTuple` is only called in the context of partition by tuples. Is that true? Incorrect commit to be talking about this but as a code reader I'd find it helpful if that were explained either in the comment or preferably in the method name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-Kxy1H-5fv0Mo0Sgv6uV:bh5lels)*

knz:
@danhhz I think you can clarify by describing a bit more details the ambiguity that arises otherwise.

danhhz:
Yeah, let me dig up the email thread

danhhz:
> @danhhz could you explain why DEFAULT needs to live in a_expr for the final syntax to work?

Good call. Added the background to the commit message for any future explorers.

> This seems somewhat dangerous. Moving DEFAULT into a_expr (well, d_expr actually) causes DEFAULT to become available in many new places.
> For example, SELECT DEFAULT now parses. This is okay because DefaultVal's TypeCheck method throws a useful error that re-validates that the DEFAULT expression is living in an okay spot. Are there any situations in which we don't call TypeCheck on every expression in an expression tree? If so, a DEFAULT in that expression tree would be a bug.

I have the same concerns. I'm all ears for ways to mitigate this risk.

---

Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/create.go, line 1149 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-Kxy1ei0trW1K62affcb:-KxyCT0L4U275L022lDl:bok36iz) ([raw file](https://github.com/cockroachdb/cockroach/blob/cab84eb380ebee356bc8f809c72a99e064c8738a/pkg/sql/create.go#L1149)):*

I was going to complain that this method seems like an awfully generic place to put this unconditional transformation, until I realized that `valueEncodeTuple` is only called in the context of partition by tuples. Is that true? Incorrect commit to be talking about this but as a code reader I'd find it helpful if that were explained either in the comment or preferably in the method name.
</blockquote></details>

whoops. this was originally scoped inside one of the partitioning methods but moved out during code review and I should have renamed it then. done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-KxyBEtUfA0avoSOtxnH:btf7ej5)*

jordanlewis:
Thanks for the explanation - that helps. How about restricting the contents of the expression trees inside of `VALUES IN`? From the perspective of someone who doesn't know much about `PARTITION` yet, it doesn't seem particularly important to permit arbitrary expressions inside the partition list. But I'm sure you guys have some good reasons I haven't thought of...

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-KxyKmzrjYU0EY0L3wZf:b-p01l3d)*

jordanlewis:
We discussed offline. Hoping to see `DEFAULT` moved to `a_expr` from `d_expr` to help reduce the scope a little bit. Otherwise :lgtm_strong: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-KxyT2wV6JyEXEIpMFCG:b-ptsfgf)*

danhhz:
Pushed the move to `a_expr`. Good idea!

---

Review status: 1 of 4 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-KxygW19BylgyfxtfHeA:boj6099)*

benesch:
I'm confused about why we still need `PartitionDefault{}`. This is more than :lgtm: enough to merge at the moment though!

Also assume you're saving the VALUES -> VALUES IN work for another PR? Or waiting on the RFC to merge?

---

Reviewed 1 of 4 files at r1, 3 of 3 files at r3.

---

*[pkg/sql/create.go, line 1154 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-KxyuPNePEmBHzX2f_k6:-KxyuPNePEmBHzX2f_k7:b-66o6ov) ([raw file](https://github.com/cockroachdb/cockroach/blob/3272dcff08536416204333a84a4fb28f20d4471f/pkg/sql/create.go#L1154)):*
> ```Go
> 			tuple.Exprs[i] = parser.PartitionDefault{}
> 		}
> 	}
> ```

Wait, can't we kill `PartitionDefault{}` now? Seems like we could just check for `parser.DefaultVal` below.

---

*[pkg/sql/parser/sql.y, line 2763 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-Kxyuka8mC0xlqmjv_-N:-Kxyuka8mC0xlqmjv_-O:bbugls) ([raw file](https://github.com/cockroachdb/cockroach/blob/3272dcff08536416204333a84a4fb28f20d4471f/pkg/sql/parser/sql.y#L2763)):*
> ```Yacc
> 
> list_partition_values:
>   '(' partition_exprs ')'
> ```

If you get rid of `PartitionDefault{}` you could make this `DefaultVal{}`, no?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-KxyupjI06VYkIkN7ApH:b-174ooc)*

danhhz:
Thanks for the reviews, y'all!

> Also assume you're saving the VALUES -> VALUES IN work for another PR? Or waiting on the RFC to merge?

Yup separate PR. Ditto for PartitionDefault{}

---

---

*[pkg/sql/create.go, line 1154 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-KxyuPNePEmBHzX2f_k6:-KxyxI4puVe0ByH9FyPA:b9qejcf) ([raw file](https://github.com/cockroachdb/cockroach/blob/3272dcff08536416204333a84a4fb28f20d4471f/pkg/sql/create.go#L1154)):*

Wait, can't we kill `PartitionDefault{}` now? Seems like we could just check for `parser.DefaultVal` below.
</blockquote></details>

Yup and I'll do that in the PR that switches to the new syntax. Just want to keep this PR as tight as possible because of the scariness voiced above

---

*[pkg/sql/parser/sql.y, line 2763 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-Kxyuka8mC0xlqmjv_-N:-KxyxfhbSM_QUwn1ZZt5:bqg0br7) ([raw file](https://github.com/cockroachdb/cockroach/blob/3272dcff08536416204333a84a4fb28f20d4471f/pkg/sql/parser/sql.y#L2763)):*

If you get rid of `PartitionDefault{}` you could make this `DefaultVal{}`, no?
</blockquote></details>

Not quite yet. MAXVALUE has to move too, which will come in the followup that implements `VALUES IN` and `VALUES <`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19744#-:-KxyxYFY9lC82QNou83W:b-c0hoa2)*

PR #19750, opened 2017-11-02
cherrypick-1.1: sql: pre-evaluate arguments to set

jordanlewis:
Run Eval() on all arguments passed to Set before beginning the Set plan.
This is necessary because the implementations of all of the different
variable setters don't get passed the proper EvalContext, and therefore
won't be able to properly resolve Placeholders once they become leaf
values.

This wouldn't be necessary if the Set implementations were able to get
the correct EvalContext, but doing that properly would require a larger
refactor.

Release notes: fix a rare panic in `SET`.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/set
+++ b/pkg/sql/logictest/testdata/logic_test/set
@@ -196,3 +196,7 @@ query T
 SHOW "time zone"
 ----
 UTC
+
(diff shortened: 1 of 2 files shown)


jordanlewis:
cc @cockroachdb/release

knz:
LGTM

bdarnell (approved):

PR #19752, opened 2017-11-02
sql: fix panic when UPDATE ran after ALTER ADD COLUMN on table with index on all columns

richardwu:
I discovered this nasty bug during https://github.com/cockroachdb/cockroach/pull/19701.

## Replication steps

1. Create a table
2. Create an index on that table on all the columns
3. Run an `ADD COLUMN` schema change, then run a DML statement (i.e. `UPDATE`) concurrently with the schema change (one can accomplish this interactively by creating a table with a large .# of rows, then quickly switching to another console to run the DML query)

## Example

```sql
CREATE TABLE kv (k INT, v INT);

CREATE INDEX allidx ON kv (k, v);   # important

# Generate enough rows to cause schema change to take O(seconds) for running DML concurrently
INSERT INTO kv (kv) SELECT k, v FROM
  GENERATE_SERIES(1,1000) AS K(k),
  GENERATE_SERIES(1,100) AS V(v);

ALTER TABLE kv ADD COLUMN x DECIMAL DEFAULT (DECIMAL '1.4');
# The following should be run concurrently with the above schema change
UPDATE kv SET v = 1337 WHERE k = 123;
```

## Cause

1. [`isCoveringIndex` is invoked](https://github.com/cockroachdb/cockroach/blob/4c420daeefedfd5b14249ffdd4b4f12797d89d13/pkg/sql/index_selection.go#L802-L809) to figure out which index to use for `UPDATE` 
2. During schema changes, `valNeededForCol` is [initialized with all columns (including mutations)](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/scan.go#L370-L373). From my understanding of schema changes, certain delete/write/read capabilities are granted after each stage. I'm assuming this is correct behavior to initialize this with all columns in case those capabilities are granted during execution.

## Fix

We check if the index exceeds the current columns in the descriptor during index selection. I've also added a regression test that fails without the fix.

### DIFF
--- a/pkg/sql/index_selection.go
+++ b/pkg/sql/index_selection.go
@@ -800,6 +800,11 @@ func (v *indexInfo) isCoveringIndex(scan *scanNode) bool {
 	}
 
 	for i, needed := range scan.valNeededForCol {
+		// This is possible during a schema change when we have
+		// additional columns.
+		if i >= len(v.desc.Columns) {
+			return false
+		}
 		if needed {
 			colID := v.desc.Columns[i].ID
 			if !v.index.ContainsColumnID(colID) {
--- a/pkg/sql/schema_changer_test.go
+++ b/pkg/sql/schema_changer_test.go
@@ -415,7 +415,7 @@ CREATE INDEX foo ON t.test (v)
 }
 
 // checkTableKeyCount returns the number of KVs in the DB, the multiple should be the
-// number of columns.
+// number of indexes (including primary).
 func checkTableKeyCount(ctx context.Context, kvDB *client.DB, multiple int, maxValue int) error {
 	tableDesc := sqlbase.GetTableDescriptor(kvDB, "t", "test")
 	tablePrefix := roachpb.Key(keys.MakeTablePrefix(uint32(tableDesc.ID)))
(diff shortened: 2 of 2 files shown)


knz:
please make the 1 line commit title be a more descriptive explanation, i.e. it should mention "update after alter add column" or something.

richardwu:
This will need to be cherrypicked for 1.1.

pkg/sql/index_selection.go
```diff
@@ -800,6 +800,11 @@ func (v *indexInfo) isCoveringIndex(scan *scanNode) bool {
 	}
 
 	for i, needed := range scan.valNeededForCol {
+		// This is possible during a schema change when we have
+		// additional columns.
+		if i >= len(v.desc.Columns) {
+			return false
```
RaduBerinde: If the column not needed, should we still return false?

richardwu: Hrm, you're right. I'm not too sure why `valNeededForCol` would ever include the mutated columns up to this point (perhaps someone more familiar with schema changes will have more context on this). Is it valid for queries during the planning phase to refer to `TableDescriptor.mutations` columns, but have those columns be resolved during execution? Because if not, maybe we shouldn't initialize `resultColumns, colIdxMap, valNeededForCol` with `n.cols`.

RaduBerinde:
*[pkg/sql/index_selection.go, line 806 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19752#-Kxy8TQLoyoaWeRafn2L-r1-806:-KxyFZjo6CBNYf7O4bG0:b-g9qi53) ([raw file](https://github.com/cockroachdb/cockroach/blob/84068d1c8b738c229c2ef195be555e19915b6dbf/pkg/sql/index_selection.go#L806)):*

Hrm, you're right. I'm not too sure why `valNeededForCol` would ever include the mutation columns up to this point (perhaps someone more familiar with schema changes will have more context on this). EDIT: Actually, I suppose mutation columns could include `DROP`s which can still have read capabilities up to this point?

Is it valid for queries during the planning phase to refer to `TableDescriptor.mutations` columns, but have those columns be resolved during execution? Because if not, maybe we shouldn't initialize `resultColumns, colIdxMap, valNeededForCol` with `n.cols`.
</blockquote></details>

I don't think it's valid for a query to refer to mutation columns but I'm not sure. CC @vivekmenezes @jordanlewis 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19752)*

jordanlewis:
Reviewed 2 of 2 files at r1.

---

*[pkg/sql/index_selection.go, line 806 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19752#-Kxy8TQLoyoaWeRafn2L-r1-806:-Kz0IZ4U5CNGuNfE5Ukk:b-pthdu5) ([raw file](https://github.com/cockroachdb/cockroach/blob/84068d1c8b738c229c2ef195be555e19915b6dbf/pkg/sql/index_selection.go#L806)):*

I don't think it's valid for a query to refer to mutation columns but I'm not sure. CC @vivekmenezes @jordanlewis 
</blockquote></details>

It is not valid for a query to explicitly refer to mutation columns.

Queries may implicitly modify mutation columns, but may never read them.

For example, an INSERT into a table with a mutation column in the DELETE_AND_WRITE_ONLY state will write the column's default value if available.

A DELETE will always delete the value in the mutation column.

I'm also confused. Why does `valNeededForCol` contain a mutation column? It seems generally wrong to me that it would need that. By definition, mutation columns never need to be read. @vivekmenezes isn't that correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19752)*

richardwu:
Tests added in https://github.com/cockroachdb/cockroach/pull/20063 and fixed in https://github.com/cockroachdb/cockroach/pull/19701. Closing but feel free to continue discussion.

vivekmenezes:

---

*[pkg/sql/index_selection.go, line 806 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19752#-Kxy8TQLoyoaWeRafn2L-r1-806:-Kz0UCcdFl-BG2FqsPiv:b-fx6t2j) ([raw file](https://github.com/cockroachdb/cockroach/blob/84068d1c8b738c229c2ef195be555e19915b6dbf/pkg/sql/index_selection.go#L806)):*

It is not valid for a query to explicitly refer to mutation columns.

Queries may implicitly modify mutation columns, but may never read them.

For example, an INSERT into a table with a mutation column in the DELETE_AND_WRITE_ONLY state will write the column's default value if available.

A DELETE will always delete the value in the mutation column.

I'm also confused. Why does `valNeededForCol` contain a mutation column? It seems generally wrong to me that it would need that. By definition, mutation columns never need to be read. @vivekmenezes isn't that correct?
</blockquote></details>

We allow an UPDATE to read columns still being added. 

https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/update.go#L280

see how we pass in a special flag publicAndNonPublicColumns to the SELECT here to allow it to read mutation column. Don't remember why we need it, but if you pass in a different value you're probably going to break a test and uncover why. It wouldn't surprise me if it has to do with UNIQUE columns.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19752)*

PR #19758, opened 2017-11-02
ui build: run make watch through webpack-dashboard

couchand:
This change updates `make run` to pipe the webpack status through webpack-dashboard, a tool that shows the current build progress and presents a few useful stats about asset and bundle size, in addition to showing the debug output of webpack.

screen shot 2017-11-02 at 4 02 50 pm

### DIFF
--- a/pkg/ui/Makefile
+++ b/pkg/ui/Makefile
@@ -75,7 +75,7 @@ $(GOBINDATA_TARGET): webpack.app.js $(DLLS) $(shell find src styl)
 watch: export TARGET := http://localhost:8080
 watch: PORT := 3000
 watch: $(DLLS)
-	$(WEBPACK_DEV_SERVER) --config webpack.app.js --port $(PORT)
+	./node_modules/.bin/webpack-dashboard -- $(WEBPACK_DEV_SERVER) --config webpack.app.js --port $(PORT)
 
 .PHONY: maintainer-clean
 maintainer-clean:
--- a/pkg/ui/package.json
+++ b/pkg/ui/package.json
@@ -110,6 +110,7 @@
     "typescript": "^2.2.1",
     "url-loader": "^0.5.8",
     "webpack": "^3.5.6",
+    "webpack-dashboard": "^1.0.0",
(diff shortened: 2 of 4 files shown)


mrtracy:
I think this is fine, looks like an easy usability win; that dashboard makes much more sense to me than whatever the current spew is.  Off the top of my head, I don't think anyone is using `make watch` in a way that would conflict with this; if @benesch doesn't have an issue with it, it's LGTM.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19758#-:-Kxya1piTnLu1Y_WH-gZ:beij7w5)*

benesch (approved):
LGTM @couchand, modulo one safety harness.

pkg/ui/Makefile
```diff
@@ -75,7 +75,7 @@ $(GOBINDATA_TARGET): webpack.app.js $(DLLS) $(shell find src styl)
 watch: export TARGET := http://localhost:8080
 watch: PORT := 3000
 watch: $(DLLS)
-	$(WEBPACK_DEV_SERVER) --config webpack.app.js --port $(PORT)
+	./node_modules/.bin/webpack-dashboard -- $(WEBPACK_DEV_SERVER) --config webpack.app.js --port $(PORT)
```
benesch: Can you also wrap webpack-dashboard in $(NODE_RUN), and then stick that in a variable named `WEBPACK_DASHBOARD` at the top of the file? I know we don't currently call `make watch` from CI, but I never ever ever want to have to debug the issue detailed in `node-run` again.

benesch (approved):

pkg/ui/Makefile
couchand: Didn't do that intentionally.  Looks like for that to work properly we'd need to address your TODO regarding isatty.  I thought that would be more work that is worth it here.  Also, I think it's highly unlikely that the make watch task would be run within docker.

benesch: Ah, ok, fair enough. LGTM.

PR #19766, opened 2017-11-02
build: Move issue-posting script from TC config to repo

bdarnell:
This is cleaned up from the previous version and modified to work
for any project, not just merge-to-master.

I got a little carried away from cleaning up all the python one-liners in the previous script (but not so carried away as to rewrite the whole thing in go)

### DIFF
--- a/build/teamcity-post-failures.py
+++ b/build/teamcity-post-failures.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+"""Post failures from the current teamcity job as github issues.
+
+Requires the following environment variables:
+- TC_API_PASSWORD
+- TC_BUILD_BRANCH
+- TC_BUILD_ID
+- GITHUB_API_TOKEN
+"""
+
+import json
+import os
+import urllib.error
+import urllib.request
+import xml.etree.ElementTree as ET
+
+from urllib.parse import urljoin, urlencode
+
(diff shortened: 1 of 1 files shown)


jordanlewis:
:lgtm_strong: +2 shipit etc. Thank you! Not that I'm in any position to complain, since this is so much better than the hacked up bash script that it is replacing, but the script could use a brief top-level comment about what it does and what environment variables it requires.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19766#-:-Kxyjl0A-mUI6zgbT9NY:bnjocyo)*

benesch:
It's so readable! ❤️ 

(I'd be in favor of doing more of this stuff in Python. I don't find Go well-suited to this sort of task.)

bdarnell:
Added comments.

To complete the change, I should cherry-pick this into both release branches and then change teamcity to call it instead of the inline shell script (since this step isn't run for open PRs there won't be any other transitional steps), right?

jordanlewis:
Yes, that should do it.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19766#-:-Ky1G33ULJ_0QN9c9hY2:bio3xyw)*

PR #19767, opened 2017-11-02
sql: Allowed columns and indexes to be added and dropped in same transaction

### LINKED ISSUE #19210: sql: Cannot add and then drop a column within the same transaction
You cannot `add` a column and then `drop` it within the scope of the same `transaction` in cockroachdb sql.  When I try this I get an error on the drop statement that the column I am trying to `drop` is in the middle of being added.

Obviously this is ostensibly an odd thing to need to do but the use case is actually pretty common - we have a database migrations project which loops through sql scripts running those which haven't yet been applied to the database. If you run the migrator somewhere where the `add` statement hasn't been applied yet ( build server/fresh deployment ) then the two statements will come together in a single transaction and the migration will fail.

jrote1:
fixes #19210

### DIFF
--- a/pkg/sql/alter_table.go
+++ b/pkg/sql/alter_table.go
@@ -306,7 +306,12 @@ func (n *alterTableNode) Start(params runParams) error {
 				}
 			}
 			if !found {
-				return fmt.Errorf("column %q in the middle of being added, try again later", t.Column)
+				for _, mutation := range n.tableDesc.Mutations {
+					if mutation.GetColumn().ID == col.ID {
+						params.p.removeSchemaChangerForMutation(mutation.MutationID)
+						break
+					}
+				}
 			}
 
 		case *parser.AlterTableDropConstraint:
--- a/pkg/sql/drop.go
+++ b/pkg/sql/drop.go
@@ -432,16 +432,6 @@ func (p *planner) dropIndexByName(
 	return nil
 }
 
-func (p *planner) removeSchemaChangerForMutation(mutationID sqlbase.MutationID) {
-	var schemaChangers = p.session.TxnState.schemaChangers.schemaChangers
-	for si, schemaChanger := range schemaChangers {
-		if schemaChanger.sc.mutationID == mutationID {
-			p.session.TxnState.schemaChangers.schemaChangers = append(schemaChangers[:si], schemaChangers[si+1:]...)
-			break
-		}
-	}
-}
-
 func (*dropIndexNode) Next(runParams) (bool, error) { return false, nil }
 func (*dropIndexNode) Close(context.Context)        {}
 func (*dropIndexNode) Values() parser.Datums        { return parser.Datums{} }
--- a/pkg/sql/drop.go
+++ b/pkg/sql/drop.go
@@ -388,7 +388,12 @@ func (p *planner) dropIndexByName(
 		}
 	}
 	if !found {
-		return fmt.Errorf("index %q in the middle of being added, try again later", idxName)
+		for _, mutation := range tableDesc.Mutations {
+			if mutation.GetIndex().ID == idx.ID {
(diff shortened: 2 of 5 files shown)


vivekmenezes:
I can review this

jrote1:
Have seen the failing test fixing now

jrote1:
@vivekmenezes is all good now

pkg/sql/logictest/testdata/logic_test/alter_table
```diff
@@ -663,3 +663,23 @@ query I
 SELECT COUNT(*) FROM crdb_internal.jobs WHERE status = 'pending' OR status = 'started'
 ----
 0
+
+# Can add and drop column in same transaction
+statement ok
+BEGIN; CREATE TABLE add_drop_column (id INT); ALTER TABLE add_drop_column ADD COLUMN column_to_drop INT; ALTER TABLE add_drop_column DROP COLUMN column_to_drop; COMMIT;
```
vivekmenezes: I'd use table "t" for both these tests

pkg/sql/drop.go
```diff
@@ -427,6 +432,16 @@ func (p *planner) dropIndexByName(
 	return nil
 }
 
+func (p *planner) removeSchemaChangerForMutation(mutationID sqlbase.MutationID) {
```
vivekmenezes: You need to find whether the column or indexes is being added through one of the outstanding schema changers. This method would then remove that schema changer and return true

pkg/sql/alter_table.go
```diff
@@ -306,7 +306,12 @@ func (n *alterTableNode) Start(params runParams) error {
 				}
 			}
 			if !found {
-				return fmt.Errorf("column %q in the middle of being added, try again later", t.Column)
+				for _, mutation := range n.tableDesc.Mutations {
+					if mutation.GetColumn().ID == col.ID {
```
vivekmenezes: the column needs to be a mutation and the mutationID must be part of p.session.TxnState.schemaChangers.schemaChangers (see my comment below)

Once that's true you can safely remove it from the mutation list through

n.tableDesc.Mutations = append(n.tableDesc.Mutations[:i], n.tableDesc.Mutations[i+1:]...)

pkg/sql/drop.go
```diff
@@ -388,7 +388,12 @@ func (p *planner) dropIndexByName(
 		}
 	}
 	if !found {
-		return fmt.Errorf("index %q in the middle of being added, try again later", idxName)
+		for _, mutation := range tableDesc.Mutations {
+			if mutation.GetIndex().ID == idx.ID {
+				p.removeSchemaChangerForMutation(mutation.MutationID)
```
vivekmenezes: same here

pkg/sql/logictest/testdata/logic_test/alter_table
vivekmenezes: Also add a schema change that adds more than one column

ALTER TABLE add_drop_column ADD COLUMN column_that_stays INT, column_to_drop INT

vivekmenezes: You also want to test this with a UNIQUE column and ensure that the implicit index is dropped.

vivekmenezes (commented):
Thanks for making this change! We really appreciate it

jrote1:
I'll look into this this evening

jrote1:
Sorry have been very busy recently, will get this ready to go ASAP

jrote1:
Am having issues with some tests, but don't have time to investigate currently. Closing PR. May reattempt when have time but don't want to hold anything up

vivekmenezes:
Hey thanks for your efforts! I appreciate you giving it a try :-)

PR #19768, opened 2017-11-02
sql,sqlbase: add some additional PARTITION BY validations

danhhz:
It turns out that generating the spans for a range partition that is
subpartitioned (by either list or range) depends on the data in the
table, so we disallow it. Tables partitioned by both list and range will
have to be list, then range.

Additionally, this commit disallows partitioning any index or table that
is interleaved into another. The root of the interleave hierarchy can be
partitioned and all children will be partitioned the same way. It's
expected that this will be sufficient for our users, but if not, this
restriction can be lifted with additional work.

Release note: none. in-progress feature


benesch:
I like the way this turned out a lot. Allowing it at parse-time but removing it from the descriptor entirely seems perfect.

:lgtm:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19768#-:-KxysVB3DJxW2iaNR4oz:bmswhko)*

danhhz:
TFTR!

---

Review status: 5 of 6 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19768#-:-Kxz1o0JGXs865CtIC1F:b-814g8k)*

Issue #19777, opened 2017-11-03
sql/pgwire: TestBinaryIntArray failed under stress


jordanlewis:
SHA: https://github.com/cockroachdb/cockroach/commits/2ce4414c1ee5c31a498a9b3fb9c0b9445127e64a

Parameters:
```

```

Stress build found a failed test: https://teamcity.cockroachdb.com/viewLog.html?buildId=31337&tab=buildLog

```
	binary_test.go:132: Testing the auto stress failure assigner
```

PR #19779, opened 2017-11-03
cmd/github-post: assign stress issues by git blame

jordanlewis:
The automatic stress test failure issue assigner now uses git blame
information to determine the assignee.

The current heuristic is very simplistic. The assignee is set to the
author of the last commit that modified the test's function signature.


jordanlewis:
Example issue (my API key and a test I created): https://github.com/cockroachdb/cockroach/issues/19777

tschottdorf:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/cmd/github-post/main.go, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky1yayDeS9C2bi8HWZT:-Ky1yayDeS9C2bi8HWZU:bszfnq8) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L115)):*
> ```Go
> 	testName = subtests[0]
> 	packageName = strings.TrimPrefix(packageName, "github.com/cockroachdb/cockroach/")
> 	cmd := exec.Command(`/bin/bash`, `-c`, fmt.Sprintf(`git grep -n "func %s" $(git rev-parse --show-toplevel)/%s/*_test.go`, testName, packageName))
> ```

Can't you replace the whole rev-parse by `./`? (workdir is already equal to that rev-parses output), and I don't think you need bash then `(git grep` understands `'*_test.go').

---

*[pkg/cmd/github-post/main.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky2--UKWRktPm3HXPsZ:-Ky2--UKWRktPm3HXPs_:bkl39ki) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L132)):*
> ```Go
> 
> 	// Now run git blame.
> 	cmd = exec.Command(`/bin/bash`, `-c`, fmt.Sprintf(`git blame --porcelain -L%s,+1 %s | grep author-mail`, linenum, filename))
> ```

Doesn't seem that you need the grep here, after all you regex match below? Then you could ditch bash again.

Also, this seems to generate the whole blame history since the beginning of time -- I checked the man page and couldn't find a `-n 1` type of switch, so I *guess* that's what it's gonna be.

---

*[pkg/cmd/github-post/main.go, line 253 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky2-lG-dYjib8fEXvAg:-Ky2-lG-dYjib8fEXvAh:bmxfybj) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L253)):*
> ```Go
> 				if len(commits) > 0 {
> 					assignee = *commits[0].Author.Login
> 				}
> ```

if we *can't* assign anyone, sigh, feel free to hard-code me. I think otherwise it may take a while until we find potential problems here. Everything should have a blame. If it doesn't, there should still be someone assigned who can triage.

Ideally the issue posting message indicates that there wasn't a better candidate, but don't worry if it's awkward.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-:-Ky20C1IXBVSLXvCbUs1:bnfp4nl)*

jordanlewis:

---

*[pkg/cmd/github-post/main.go, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky1yayDeS9C2bi8HWZT:-Ky2SSUXkjvJ075Bo_sk:bgxruuw) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L115)):*

Can't you replace the whole rev-parse by `./`? (workdir is already equal to that rev-parses output), and I don't think you need bash then `(git grep` understands `'*_test.go').
</blockquote></details>

This breaks the tests, otherwise I would have done it.

---

*[pkg/cmd/github-post/main.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky2--UKWRktPm3HXPsZ:-Ky2Sr8vekcZv8XvwCD0:bajc172) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L132)):*

Doesn't seem that you need the grep here, after all you regex match below? Then you could ditch bash again.

Also, this seems to generate the whole blame history since the beginning of time -- I checked the man page and couldn't find a `-n 1` type of switch, so I *guess* that's what it's gonna be.
</blockquote></details>

Yeah... but bash makes it so you don't have to split your arguments. I left this alone primarily out of laziness.

---

*[pkg/cmd/github-post/main.go, line 253 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky2-lG-dYjib8fEXvAg:-Ky2i1Hlr12retzfyu4u:belbebn) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L253)):*

if we *can't* assign anyone, sigh, feel free to hard-code me. I think otherwise it may take a while until we find potential problems here. Everything should have a blame. If it doesn't, there should still be someone assigned who can triage.

Ideally the issue posting message indicates that there wasn't a better candidate, but don't worry if it's awkward.
</blockquote></details>

Added a message if we can't find an assignee, defaulted to you, and made tamird redirect to you as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779)*

tschottdorf:
LGTM!

---

Reviewed 3 of 5 files at r2, 2 of 2 files at r3.

---

*[pkg/cmd/github-post/main.go, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky1yayDeS9C2bi8HWZT:-Ky2oiU2LLouOMTesjJB:bizzdjj) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L115)):*

This breaks the tests, otherwise I would have done it.
</blockquote></details>

Curious what you mean. It's ok as-is, either way.

---

*[pkg/cmd/github-post/main.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky2--UKWRktPm3HXPsZ:-Ky2oyI_IV2l7im220iI:b-bk5ip0) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L132)):*

Yeah... but bash makes it so you don't have to split your arguments. I left this alone primarily out of laziness.
</blockquote></details>

Works for me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-:-Ky2ppkG8S32Lu70wvts:b-85i648)*

jordanlewis:
TFTR!

---

---

*[pkg/cmd/github-post/main.go, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-Ky1yayDeS9C2bi8HWZT:-Ky2q05SHfmS8rzUQX_o:b-5pagp9) ([raw file](https://github.com/cockroachdb/cockroach/blob/099f3bcd1dc09b72cdd6f894092e874cbfbf320f/pkg/cmd/github-post/main.go#L115)):*

Curious what you mean. It's ok as-is, either way.
</blockquote></details>

The go tests for github-post I mean. They actually call this whole thing (minus the github api part) with a real test name, and expect git blame to find things. I think the tests get run with a different CWD, don't they?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19779#-:-Ky2qBWmUOUsADx8vTj5:b-814g8k)*

PR #19780, opened 2017-11-03
log: better error reporting

tbg:
The previous code tended to lose the type when it mattered most, for example in
`errors.Wrap(unknownSentinelErr, "something")`, the output would have only an
empty string for the unknown sentinel error. This will now print its type
instead (which could still be useless, but it's better than nothing).

### DIFF
--- a/pkg/util/log/crash_reporting.go
+++ b/pkg/util/log/crash_reporting.go
@@ -181,21 +181,37 @@ func (e *safeError) Error() string {
 	return e.message
 }
 
-func redact(r interface{}) string {
+// redact returns the redacted string and a boolean indicating whether the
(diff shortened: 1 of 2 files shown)


tschottdorf:
Added a second commit that I think is very useful because it reports all the involved file:line tuples for any `pkg/errors`-wrapped error chains.

dt:
:lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19780#-:-KyHq7B9-UE9JkFnqf5b:bg7igkv)*

tschottdorf:
TFTR!

PR #19789, opened 2017-11-03
sql: add JSON encoding

justinj:
This commit introduces the ability to encode JSON values into bytes.
This will allow storing it in column types.

This commit does not actually enable storing JSON values in columns,
however, and it doesn't implement any of the intelligent selective
decoding of JSON values for particular operators.


justinj:
Going to have a follow up PR to extend the encoding tech-note for this as you suggested @knz!

jordanlewis:
Reviewed 4 of 4 files at r1.

---

*[pkg/util/encoding/encoding.go, line 125 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHBiz00J2PzO8TSKMg:-KyHBiz00J2PzO8TSKMh:b-ed5tvj) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/encoding/encoding.go#L125)):*
> ```Go
> 
> // PutUint32Ascending encodes the uint32 value using a big-endian 4 byte
> // representation at the specified index.
> ```

s/index./index, lengthening the input slice if necessary./

---

*[pkg/util/json/encode.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHBxQL7O2Sn00kPu3i:-KyHBxQL7O2Sn00kPu3j:bki4lfy) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L54)):*
> ```Go
> }
> 
> func (jsonNull) encode(appendTo []byte) (jEntry uint32, b []byte, err error) {
> ```

These functions confuse me. Why don't we modify `appendTo`?

On second look I guess it's because the JEntry's contain the type information, or something like that. It would be useful to write a brief doc explaining what's going on.

---

*[pkg/util/json/encode.go, line 143 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHD8mJruQ-28apFGFy:-KyHD8mJruQ-28apFGFz:b-n6pk9x) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L143)):*
> ```Go
> }
> 
> // EncodeJSON encodes a JSON value as a sequences of bytes.
> ```

nit: s/sequences/sequence/

---

*[pkg/util/json/encode.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHDJeSjMkSR_oxG0Q-:-KyHDABovK-NKWUTUfpy:bf9g4zr) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L144)):*
> ```Go
> 
> // EncodeJSON encodes a JSON value as a sequences of bytes.
> func EncodeJSON(appendTo []byte, j JSON) ([]byte, error) {
> ```

Not necessarily relevant to this PR, but I wonder if we'll want to make a function that estimates the size of an encoded json object to help the caller right-size their byte array.

---

*[pkg/util/json/json.go, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRsGjovvWNGd7iZQtC:-KyRsGjovvWNGd7iZQtD:b-k9j7p7) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json.go#L83)):*
> ```Go
> 	Exists(string) bool
> 
> 	// encode returns the encoded bytes for the JSON document along with its jEntry.
> ```

sometimes `appendTo` isn't modified - should explain why here

---

*[pkg/util/json/json_test.go, line 524 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRsYEYMj0Rf5SwdgKI:-KyRsYEYMj0Rf5SwdgKJ:b4i81ji) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json_test.go#L524)):*
> ```Go
> func TestJSONEncodeRoundTrip(t *testing.T) {
> 	cases := []string{
> 		`true`,
> ```

I feel like we need some large-object test cases here. Wasn't there some special logic for bigger objects? The stride stuff? Did we end up skipping that? I don't see references to that in this code - possibly I'm thinking of something that doesn't apply to this situation.

I bet there's like a canonical "i'm here to screw up your codecs" json object lying around on the internet for testing that you can find.

Also now would be a good time to add an encode/decode benchmark we can keep around for later.

---

*[pkg/util/json/json_test.go, line 554 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRt3vXxg5SSiarhEq-:-KyRt3vXxg5SSiarhEq0:br2s4l8) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json_test.go#L554)):*
> ```Go
> 		`{"a": [1, 2, 3]}`,
> 		`{"a": [{"b": 2}, {"b": 4}]}`,
> 		`[1, {"a": 3}, null, {"b": null}]`,
> ```

Is there testing we need to do for duplicate keys? Does that get filtered out at an earlier stage? Should the code here fail gracefully in that case or is it okay?

---

*[pkg/util/json/json_test.go, line 555 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRtAOwV4mhe4q4ukLw:-KyRtAOxKOVdXHxm_Us2:b-no0oog) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json_test.go#L555)):*
> ```Go
> 		`{"a": [{"b": 2}, {"b": 4}]}`,
> 		`[1, {"a": 3}, null, {"b": null}]`,
> 		`{"🤔": "foo"}`,
> ```

What about unicode values?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789)*

m-schneider:
Looks good, do you think we should share some of the type tags for indexing?

---

---

*[pkg/util/json/encode.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHBxQL7O2Sn00kPu3i:-KySI35hyB46Zb8w96zA:b-bxzt2z) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L54)):*

These functions confuse me. Why don't we modify `appendTo`?

On second look I guess it's because the JEntry's contain the type information, or something like that. It would be useful to write a brief doc explaining what's going on.
</blockquote></details>

Agreed, I'm a bit confused too.

---

*[pkg/util/json/encode.go, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KySK-kvbDDhRDKxxapv:-KySK-kvbDDhRDKxxapw:b45l25u) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L99)):*
> ```Go
> 			return 0, appendTo, nil
> 		}
> 		appendTo = encoding.PutUint32Ascending(appendTo, nextJEntry, jEntryIdx+i*4)
> ```

I'm a bit confused as to what you're doing here. Are you adding the encoded value of the array entry into a an int32?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-:-KySJ5KZCrR9qRvcB_Te:b-9kaotv)*

justinj:
I added extra tests cases in a separate commit.

Perhaps we could! What would you need for that?

---

Review status: 0 of 99 files reviewed at latest revision, 9 unresolved discussions.

---

*[pkg/util/encoding/encoding.go, line 125 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHBiz00J2PzO8TSKMg:-KyWM2maLQhqf6Oc5qo0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/encoding/encoding.go#L125)):*

s/index./index, lengthening the input slice if necessary./
</blockquote></details>

Done.

---

*[pkg/util/json/encode.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHBxQL7O2Sn00kPu3i:-KyWIlCLdPAExA_p7Mto:bg2j3bc) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L54)):*

Agreed, I'm a bit confused too.
</blockquote></details>

Yeah, it's because the encoding of `true`, `false`, and `null` is the 0-length byte sequence. I'll add a comment clarifying that this is the case. The [RFC](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20171005_jsonb_encoding.md) is the canonical doc for the format.

---

*[pkg/util/json/encode.go, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KySK-kvbDDhRDKxxapv:-KyWJj2TazdZrapQ3Jvw:b-vo94nj) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L99)):*

I'm a bit confused as to what you're doing here. Are you adding the encoded value of the array entry into a an int32?
</blockquote></details>

This is filling in the JEntry headers for each of the array elements, see [the relevant section](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20171005_jsonb_encoding.md#array) in the encoding RFC.

---

*[pkg/util/json/encode.go, line 143 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHD8mJruQ-28apFGFy:-KyWK0BVdpVzgOZDAHyT:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L143)):*

nit: s/sequences/sequence/
</blockquote></details>

Done.

---

*[pkg/util/json/encode.go, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyHDJeSjMkSR_oxG0Q-:-KyWK0_prBzBxiMzFVsZ:b-sbyptw) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/encode.go#L144)):*

Not necessarily relevant to this PR, but I wonder if we'll want to make a function that estimates the size of an encoded json object to help the caller right-size their byte array.
</blockquote></details>

That sounds doable - we could maybe even construct such an estimate during parsing?

---

*[pkg/util/json/json.go, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRsGjovvWNGd7iZQtC:-KyWKOV0xQM2pDfXszUz:b68d5ow) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json.go#L83)):*

sometimes `appendTo` isn't modified - should explain why here
</blockquote></details>

This doesn't really feel like an edge case to me - it just so happens that the encoding of `true`/`false`/`null` is the empty byte sequence, but I added a clarifying comment.

---

*[pkg/util/json/json_test.go, line 524 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRsYEYMj0Rf5SwdgKI:-KyWXB5ld7RILsXsoUHB:b-l13l2f) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json_test.go#L524)):*

I feel like we need some large-object test cases here. Wasn't there some special logic for bigger objects? The stride stuff? Did we end up skipping that? I don't see references to that in this code - possibly I'm thinking of something that doesn't apply to this situation.

I bet there's like a canonical "i'm here to screw up your codecs" json object lying around on the internet for testing that you can find.

Also now would be a good time to add an encode/decode benchmark we can keep around for later.
</blockquote></details>

We decided to leave the stride as infinite for the moment, so everything is len-encoded (but backwards-compatible).

---

*[pkg/util/json/json_test.go, line 554 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRt3vXxg5SSiarhEq-:-KyWz6G68xsYVt8cv_2U:b-e5ny3s) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json_test.go#L554)):*

Is there testing we need to do for duplicate keys? Does that get filtered out at an earlier stage? Should the code here fail gracefully in that case or is it okay?
</blockquote></details>

Yes that gets filtered out by the JSON parser (I believe the Go parser takes the value of the final key). Currently all ways of constructing a JSON object handle this.

---

*[pkg/util/json/json_test.go, line 555 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-KyRtAOwV4mhe4q4ukLw:-KyX-5_QjLzBQsYzUYdK:b-fd9mwj) ([raw file](https://github.com/cockroachdb/cockroach/blob/b916fbdcd7775fa0558ef2ebb20e650a95119f5e/pkg/util/json/json_test.go#L555)):*

What about unicode values?
</blockquote></details>

There are some explicit unicode values in the bunch of test cases I just added.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-:-KyX1CBBlZ0hI5lU3JPf:b-ht38te)*

justinj:
Ok, now that the random generator merged I updated this with a round-tripping test involving that. I also verified that breaking the encoding code causes that test to fail.

jordanlewis:
:lgtm:

---

Reviewed 99 of 99 files at r2, 4 of 4 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19789#-:-Kz0HPYjvLNXbJ7a1j_U:bnfp4nl)*

justinj:
TFTRs!

PR #19796, opened 2017-11-04
acceptance: avoid runaway loop RunInitCommand

tbg:
The retry loop @bdarnell was nervous about in review of the code that
previously touched this turned out to wreak havoc due to matching the
wrong context error (`Canceled` instead of `DeadlineExceeded`). Make
this code generally safer and also more fun to look at.

### DIFF
--- a/pkg/acceptance/cluster/dockercluster.go
+++ b/pkg/acceptance/cluster/dockercluster.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


tschottdorf:
Rewritten after I found out what was really going on.

tschottdorf:
Hmm, this also doesn't work because the error matching doesn't ever fire (the returned error is just "exit code 1" and the interesting part just ends up going to stderr directly).

Guess I'll have to rewrite this again.

tschottdorf:
Ok, rewritten again.

bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19796#-:-KyELHMsjB_JsPIwS6ql:bnfp4nl)*

PR #19802, opened 2017-11-04
sql: convert placeholders to leaf values

jordanlewis:
Had to recreate this - apparently, you can't re-open a PR (#19424) if you force pushed to it while it was closed. I resolved all the comments from the previous PR.

Previously, the Placeholder expression type was merely a marker that
was replaced during type checking with an actual Datum value by a tree
traversal.

This was inefficient - every query with placeholders would end up
getting path-copied during that replacement step. Additionally, it was
an obstacle to optimizations like plan caching that requires that
expression trees retain their meaning after type checking.

Now, a Placeholder is a fully-fledged Datum that can be evaluated.
Evaluating a Placeholder returns the Datum that it points to in the
PlaceholderInfo map.

This transformation required a lot of plumbing, which makes up the bulk
of this patch. EvalContext had to be augmented to include a
PlaceholderInfo map so that the placeholder values can be known at
Placeholder.Eval() time. EvalContexts and SemaContexts also needed
to be plumbed to every place that needs to check for constant
expressions, since checking for constant expressions now requires that
placeholder values be known (placeholders without values are not
constant expressions; placeholders with values are).

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -330,11 +330,12 @@ func makeCSVTableDescriptor(
 			return nil, errors.Errorf("unsupported table definition: %s", parser.AsString(def))
 		}
 	}
+	semaCtx := parser.SemaContext{}
+	evalCtx := parser.EvalContext{}
 	tableDesc, err := sql.MakeTableDesc(
 		ctx,
 		nil, /* txn */
 		sql.NilVirtualTabler,
-		parser.SearchPath{},
 		create,
 		parentID,
 		tableID,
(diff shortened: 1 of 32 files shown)


jordanlewis:
First commit is #19809.

knz:
Reviewed 3 of 3 files at r1, 11 of 11 files at r2, 4 of 4 files at r3.
Review status: 14 of 31 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19802)*

knz:
:lgtm: - well done!

---

Reviewed 17 of 17 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19802#-:-KyH9bB4r7W4lxZMSTlN:b-qf6lo1)*

jordanlewis:
TFTR!

PR #19809, opened 2017-11-05
sql: restrict when EXECUTE can be used

jordanlewis:
This commit prevents users from using `EXECUTE` inside of `PREPARE`
statements and alongside other `EXECUTE` statements, both of which are
made possible by the `... FROM [<stmt>]` syntax.

For example, `PREPARE x AS SELECT * FROM [EXECUTE y]` is no longer
permitted. Nor is `SELECT * FROM [EXECUTE x] CROSS JOIN [execute y]`.

It's difficult to properly support this, because special care must be
taken all over the place to maintain a proper stack of prepared
statement parameters. It's also currently buggy, despite the effort made
in #19720. The problem is `EXPLAIN` statements - since `EXECUTE`
statements don't get their own plan nodes, the `EXPLAIN` plan walker has
no ability to maintain a prepared statement argument stack, since
there's no marker inside of a plan tree for where an `EXECUTE` with
parameters begins. There's no technical reason why a new plan node for
`EXECUTE` couldn't be added, but it seems like extra complexity for no
reason at the moment.

Reverts #19720.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/prepare
+++ b/pkg/sql/logictest/testdata/logic_test/prepare
@@ -361,17 +361,11 @@ EXECUTE sets('hello')
 statement error could not determine data type of placeholder
 PREPARE x19597 AS SELECT $1 IN ($2, null);
 
-# Regression for #19716: ensure that placeholders for EXECUTE statements inside
-# of PREPARE statements are properly segregated from those of the outer
-# statement.
-
 statement ok
 PREPARE innerStmt AS SELECT $1:::int i, 'foo' t
 
-statement ok
+statement error can't prepare an EXECUTE statement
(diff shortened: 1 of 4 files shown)


knz:
I concur that this restriction is a good price to pay to restore sanity -- we can work to add what you suggest later.

:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19809#-:-KyFxq1JsG3yGrT10k_0:b-rgykf4)*

jordanlewis:
TFTR!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19809#-:-KyGriHa0bDAgQQ07wKG:b-814g8k)*

PR #19811, opened 2017-11-06
build: fail if TESTS specified without PKG

jordanlewis:
A common error is to try to run a named test or benchmark with TESTS= or
BENCHES= without remembering to specify the package that the test is in
with PKG=. Doing this is always(?) a waste of time and requires a ctrl-c
and re-run. Prevent this failure mode by failing in this case.

Users that really want to run a named test against all packages can
specify PKG=./... to get that behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,15 @@
 # `help` target. They look a bit awkward in the variable declarations below
 # since any whitespace added would become part of the variable's default value.
 
+ifeq ($(PKG),)
+ifneq ($(TESTS),)
+$(error TESTS must be specified with PKG (e.g. PKG=./pkg/sql/))
+endif
+ifneq ($(BENCHES),)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -239,6 +248,7 @@ bench: TESTTIMEOUT := $(BENCHTIMEOUT)
 
 .PHONY: check test testshort testrace testlogic bench
 check test testshort testrace bench: gotestdashi
+	$(warning You specified TESTS=$(TESTS) without specifying PKG - your tests will run against all packages)
```
benesch: (This looks it's been superseded by L25-32.)

jordanlewis: Oops, yeah. Sloppy!

benesch:
LGTM!

benesch (approved):

Issue #13651, opened 2017-02-17
sql: arrays can't be cast to their own type


jordanlewis:
```
root@:26257/> SELECT ARRAY[1, 2]::int[];
pq: invalid cast: int[] -> INT[]
```

knz:
Note for implementer: there's a commented test for this already in `pgwire_test.go`.

justinj:
Can this be closed now @jordanlewis ? Also note [this comment](https://github.com/justinj/cockroach/blob/ed2720d4e7b3613819fba550e1e60cd0bc3cae55/pkg/sql/pgwire/pgwire_test.go#L910-L913)

jordanlewis:
Will send a PR for that comment.

PR #19816, opened 2017-11-06
sql: allow arrays to be cast to their own type

jkominek:
@jordanlewis 

I ran into #13651 while trying to use epgsql against cockroach, which [runs some queries under the hood](https://github.com/epgsql/epgsql/blob/3652aa9c0a7cb58beb4736cdbdfa0dc4f7eafa20/src/epgsql.erl#L181) that contain array casts.

#13651 mentions that there's a test for this, here: https://github.com/cockroachdb/cockroach/blob/0c8b4d9370caedce43caacd339d99f5993ece257/pkg/sql/pgwire/pgwire_test.go#L879

Unfortunately the test doesn't pass; with this patch the error changes from the previous "invalid cast: int[] -> INT[]", to the following:

```
--- FAIL: TestPGPreparedQuery (0.34s)
    --- FAIL: TestPGPreparedQuery/exec (0.11s)
        --- FAIL: TestPGPreparedQuery/exec/SELECT_$1::INT[] (0.00s)
            --- FAIL: TestPGPreparedQuery/exec/SELECT_$1::INT[]/0 (0.00s)
                pgwire_test.go:945: sql: Scan error on column index 0: pq: destination <nil> is not a pointer to array or slice
                pgwire_test.go:951: SELECT $1::INT[]: [{[10]}]: expected [{[10]}], got [{<nil>}]
    --- FAIL: TestPGPreparedQuery/prepare (0.04s)
        --- FAIL: TestPGPreparedQuery/prepare/SELECT_$1::INT[] (0.00s)
            --- FAIL: TestPGPreparedQuery/prepare/SELECT_$1::INT[]/0 (0.00s)
                pgwire_test.go:945: sql: Scan error on column index 0: pq: destination <nil> is not a pointer to array or slice
                pgwire_test.go:951: SELECT $1::INT[]: [{[10]}]: expected [{[10]}], got [{<nil>}]
```

having stared at the referenced code for a bit, I think the test framework in there has troubles with arrays, but my go skills are insufficient at this point to sort it out. So I left that test commented out, and included another test of the fix in sql/logictest/testdata/logic_test/array; hopefully that's sufficient.

Do let me know if there's anything more I can do, here.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/array
+++ b/pkg/sql/logictest/testdata/logic_test/array
@@ -125,6 +125,23 @@ SELECT '{hello}'::VARCHAR(2)[]
 ----
 {"he"}
 
+# array casting
+
+query T
+SELECT ARRAY[1,2,3]::INT[]
+----
+{1,2,3}
+
+query error invalid cast: int[] -> UUID[]
+SELECT ARRAY[1,2,3]::UUID[]
+
+query error invalid cast: inet[] -> INT[]
+SELECT ARRAY['8.8.8.8'::INET, '8.8.4.4'::INET]::INT[]
+
+# this _should_ be valid, but isn't, yet
+query error invalid cast: int[] -> TEXT[]
+SELECT ARRAY[1,2,3]::TEXT[]
+
 # array subscript access
 
 query T
--- a/pkg/sql/sem/tree/eval.go
+++ b/pkg/sql/sem/tree/eval.go
@@ -2740,8 +2740,13 @@ func performCast(ctx *EvalContext, d Datum, t coltypes.CastTargetType) (Datum, e
 			return v, nil
 		}
 	case *coltypes.TArray:
-		if s, ok := d.(*DString); ok {
-			return ParseDArrayFromString(ctx, string(*s), typ.ParamType)
+		switch v := d.(type) {
+		case *DString:
+			return ParseDArrayFromString(ctx, string(*v), typ.ParamType)
+		case *DArray:
+			if (*v).ParamTyp == coltypes.CastTargetToDatumType((*typ).ParamType) {
+				return d, nil
+			}
 		}
 	case *coltypes.TOid:
 		switch v := d.(type) {
--- a/pkg/sql/sem/tree/expr.go
+++ b/pkg/sql/sem/tree/expr.go
@@ -1210,7 +1210,10 @@ func validCastTypes(t types.T) []types.T {
 		if t.FamilyEqual(types.FamCollatedString) {
 			return stringCastTypes
 		} else if t.FamilyEqual(types.FamArray) {
-			return arrayCastTypes
+			ret := make([]types.T, len(arrayCastTypes)+1)
+			copy(ret, arrayCastTypes)
+			ret[len(ret)-1] = t
+			return ret
 		}
 		return nil
 	}


justinj:
Hey @jkominek! Thanks for the PR, this is a great change to have! I left a few comments on the code.

---

---

*[pkg/sql/logictest/testdata/logic_test/array, line 962 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHzcAC9Lga9EXeCIpG:-KyHzcAC9Lga9EXeCIpH:bbi4jwm) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/logictest/testdata/logic_test/array#L962)):*
> ```
> 
> query T
> SELECT ARRAY[1,2,3]::INT[]
> ```

Can you also add a test that we can't convert between incompatible array types?

---

*[pkg/sql/parser/expr.go, line 1162 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHwO22psTuUcevW-_1:-KyHwO22psTuUcevW-_2:b-7agpte) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/parser/expr.go#L1162)):*
> ```Go
> 			return stringCastTypes
> 		} else if t.FamilyEqual(TypeArray) {
> 			return append(arrayCastTypes, t)
> ```

I think this is a little sketchy - while I don't suspect this particular case will cause any actual problems, it's not a great idea to call `append` on a slice you're intending to re-use. Further, if we look at the caller (there's only one, in `type_check.go`) we can see that it uses `FamilyEqual` to compare to the elements of this slice (and array types are `FamilyEqual` with each other), so I think (but haven't actually checked) that every array type will end up being considered castable to every other array type, which is probably not what we want. I think the most reasonable thing to do would be to remove all of the identity cases (so those which represent a cast from a type to itself) from the lists in `expr.go`, and then add an extra check in `type_check.go` where this function is called that if the two types are considered equal we always allow the cast. Or we could stop using FamilyEqual, I'm not sure what the original reason for using it was, it may or may not be applicable anymore, you might have to run the tests and find out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-:-KyI-md8_M22mMLPvkEB:b-chop1z)*

jkominek:

---

*[pkg/sql/logictest/testdata/logic_test/array, line 962 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHzcAC9Lga9EXeCIpG:-KyJ0P8Vw_omg1cLjGzz:bbvt2i2) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/logictest/testdata/logic_test/array#L962)):*

Can you also add a test that we can't convert between incompatible array types?
</blockquote></details>

i've added three additional tests (and moved them all up near the other casting tests).

---

*[pkg/sql/parser/eval.go, line 2744 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyJ2k4ZU_aq8WBgeWkx:-KyJ2k4ZU_aq8WBgeWky:b-w6ohkc) ([raw file](https://github.com/cockroachdb/cockroach/blob/7578f1e94665610faac88ac374655a139314c23c/pkg/sql/parser/eval.go#L2744)):*
> ```Go
> 				return d, nil
> 			}
> 		}
> ```

falling through here lands us in another invalid cast error condition, and is why these changes (and the string -> array casting) are (IMO :smile:) safe and reasonable.

---

*[pkg/sql/parser/expr.go, line 1162 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHwO22psTuUcevW-_1:-KyJ0iQuaXhhZypg5oiW:b-aepmtc) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/parser/expr.go#L1162)):*

I think this is a little sketchy - while I don't suspect this particular case will cause any actual problems, it's not a great idea to call `append` on a slice you're intending to re-use. Further, if we look at the caller (there's only one, in `type_check.go`) we can see that it uses `FamilyEqual` to compare to the elements of this slice (and array types are `FamilyEqual` with each other), so I think (but haven't actually checked) that every array type will end up being considered castable to every other array type, which is probably not what we want. I think the most reasonable thing to do would be to remove all of the identity cases (so those which represent a cast from a type to itself) from the lists in `expr.go`, and then add an extra check in `type_check.go` where this function is called that if the two types are considered equal we always allow the cast. Or we could stop using FamilyEqual, I'm not sure what the original reason for using it was, it may or may not be applicable anymore, you might have to run the tests and find out.
</blockquote></details>

in retrospect, yes, that was more than a little sketchy. also unnecessary because of the whole `FamilyEqual` thing.  adding `TypeArray` to `arrayCastTypes` is sufficient to get the expressions past this stage of type checking. unimplemented (and impossible) conversions then fail when `performCast` attempts to process them.

FWIW, this feels OK-ish to me because it's the same thing that happens when it attempts to perform a string to array cast on a bad string. `'{1,2,3'::INT[]` passes type check, but fails at `performCast`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816)*

justinj:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/array, line 962 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHzcAC9Lga9EXeCIpG:-KyJ7hxn53v_4PG5wKmw:b-ucfknh) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/logictest/testdata/logic_test/array#L962)):*
<details><summary><i>Previously, jkominek (Jay Kominek) wrote…</i></summary><blockquote>

i've added three additional tests (and moved them all up near the other casting tests).
</blockquote></details>

Thanks!

---

*[pkg/sql/logictest/testdata/logic_test/array, line 128 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyJ6IQH-1C0R0om4Hn-:-KyJ6IQH-1C0R0om4Hn0:b-lo0sxr) ([raw file](https://github.com/cockroachdb/cockroach/blob/7578f1e94665610faac88ac374655a139314c23c/pkg/sql/logictest/testdata/logic_test/array#L128)):*
> ```
> {1,2,3}
> 
> query error pq: invalid cast: int[] -> UUID[]
> ```

nit: we usually omit the `pq:` prefix (this string is actually a regex).

---

*[pkg/sql/logictest/testdata/logic_test/array, line 132 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyJ7kZZtxK_YEHMU14b:-KyJ7kZZtxK_YEHMU14c:bkz0u9q) ([raw file](https://github.com/cockroachdb/cockroach/blob/7578f1e94665610faac88ac374655a139314c23c/pkg/sql/logictest/testdata/logic_test/array#L132)):*
> ```
> 
> query error pq: invalid cast: inet[] -> INT[]
> SeLECT ARRAY['8.8.8.8'::INET, '8.8.4.4'::INET]::INT[]
> ```

nit: `s/SeLECT/SELECT/`

---

*[pkg/sql/parser/eval.go, line 2744 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyJ2k4ZU_aq8WBgeWkx:-KyJ6_CZxYYVrrT0uBKP:b-hk32ne) ([raw file](https://github.com/cockroachdb/cockroach/blob/7578f1e94665610faac88ac374655a139314c23c/pkg/sql/parser/eval.go#L2744)):*
<details><summary><i>Previously, jkominek (Jay Kominek) wrote…</i></summary><blockquote>

falling through here lands us in another invalid cast error condition, and is why these changes (and the string -> array casting) are (IMO :smile:) safe and reasonable.
</blockquote></details>

Hm, @knz do you have any thoughts on this? That sounds safe (for now) but it feels undesirable for invalid casts to make it past type-checking and into the evaluation stage - it seems to me like any cases where we currently do this are (perhaps non-observable) bugs (but my thoughts there might be unfounded!). If not, then it seems odd for us to have this tested during type-checking at all if it's just going to be validated again during evaluation.

---

*[pkg/sql/parser/expr.go, line 1162 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHwO22psTuUcevW-_1:-KyJ79Jv6-IqyepKXdXs:b-g58wl9) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/parser/expr.go#L1162)):*
<details><summary><i>Previously, jkominek (Jay Kominek) wrote…</i></summary><blockquote>

in retrospect, yes, that was more than a little sketchy. also unnecessary because of the whole `FamilyEqual` thing.  adding `TypeArray` to `arrayCastTypes` is sufficient to get the expressions past this stage of type checking. unimplemented (and impossible) conversions then fail when `performCast` attempts to process them.

FWIW, this feels OK-ish to me because it's the same thing that happens when it attempts to perform a string to array cast on a bad string. `'{1,2,3'::INT[]` passes type check, but fails at `performCast`.
</blockquote></details>

I see what you're saying, and it sounds somewhat reasonable to me, I'll let @knz weigh in when it comes to things like that re: the type system, though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816)*

knz:
Hi Jay! Thank you for helping us with this!

---

Reviewed 2 of 2 files at r1, 1 of 2 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/expr.go, line 1162 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-KyHwO22psTuUcevW-_1:-KyL0zxVYcpw-URyieVs:b-hxlwzp) ([raw file](https://github.com/cockroachdb/cockroach/blob/38fe57f3e6e8d6ebd831211960fcb71fc9362514/pkg/sql/parser/expr.go#L1162)):*

I see what you're saying, and it sounds somewhat reasonable to me, I'll let @knz weigh in when it comes to things like that re: the type system, though.
</blockquote></details>

Thanks for inviting me to the conversation. It is indeed undesirable to let type errors propagate to query execution -- if a query passes type checking, it is expected that all the evaluations will be working with valid types. Although your code change in `eval.go` currently "catches" this type error (late), we can't assume we'll be working with that code forever: soon enough we'll be looking at query compilation, and there's no way whatsoever that the compiled form of an expression will be able to check types.

So I would simply request that either you move the check for array element type compatibility from `eval.go` to here, or do what justin suggested initially.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-:-KyL1YYwpwlQtAZSgEf8:bqasi1j)*

jordanlewis:
Thanks for your contribution, @jkominek! I took the liberty of updating your patch to respond to the comments - hope that's okay!

jordanlewis:
@justinj PTAL

justinj:
:lgtm: modulo the append thing we talked about

---

Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19816#-:-L0kTtd5uOYk6dPmakLx:b-g3bp38)*

PR #19818, opened 2017-11-06
sql: extract the SQL types into their own package

knz:
Visible changes:

- `parser.TypeXXX` -> `types.XXX` (overall much shorter to type!)
- `parser.Type` -> `types.T`
- `parser.Type{Tuple,Table,Array,CollatedString,Placeholder}` -> `types.Fam{Tuple,Table,Array,CollagedString,Placeholder}` to highlight more clearly that they are special.


knz:
cc @cockroachdb/sql

nvanbenschoten:
Nice change! Let's try to get this in as soon as possible to avoid conflicts
with other changes.

The only commit I'm not sure about is:
`types: type families are families, not types. rename them.`

Type families are types, just parametric types, so I'm not sure the logic holds.
Also, even if the reasoning was technically true, making the distinction doesn't
make anything cleaner. I think it actually makes the few places where the new
"Fam" variables are used more confusing. In SQL an ARRAY is considered a type,
so why isn't it here? It also brings up other questions:
- should they even be in the `types` package?
- `Any` is parametric. Is it a type family?

Another comment is that if a commit consists of simply running a sed command
without any manual intervention, that command should be included in the commit
message. This makes the change more clear, reproducible, and helps reviewers
know what kinds of issues to look for.

---

Reviewed 5 of 5 files at r1, 16 of 16 files at r2, 8 of 8 files at r3, 5 of 5 files at r4, 5 of 5 files at r5, 66 of 66 files at r6, 84 of 84 files at r7, 12 of 12 files at r8, 7 of 7 files at r9, 84 of 84 files at r10, 19 of 19 files at r11.
Review status: all files reviewed at latest revision, 17 unresolved discussions, all commit checks successful.

---

*[pkg/sql/distsqlrun/expr_test.go, line 25 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH8XS3ZyNmjt0CjCkp:-KyH8XS3ZyNmjt0CjCkq:bo7rqpc) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/distsqlrun/expr_test.go#L25)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
> 
> 	"github.com/cockroachdb/cockroach/pkg/sql/sem/types"
> ```

nit: move up

---

*[pkg/sql/parser/col_types.go, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5N7RQWLxF_iPn8Sh:-KyH5N7RQWLxF_iPn8Si:b-nwf53d) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/col_types.go#L24)):*
> ```Go
> 	"github.com/cockroachdb/apd"
> 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
> 	types "github.com/cockroachdb/cockroach/pkg/sql/sem/types"
> ```

alias not needed

---

*[pkg/sql/parser/constant.go, line 28 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH8exfgaUzS8s26F7F:-KyH8exfgaUzS8s26F7G:bo7rqpc) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/parser/constant.go#L28)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
> 	"github.com/cockroachdb/cockroach/pkg/sql/sem/types"
> 	"github.com/pkg/errors"
> ```

nit: move up

---

*[pkg/sql/parser/create.go, line 28 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH8malYrJ7x4spkQ18:-KyH8malYrJ7x4spkQ19:bodkqx5) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/parser/create.go#L28)):*
> ```Go
> 
> 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
> 	types "github.com/cockroachdb/cockroach/pkg/sql/sem/types"
> ```

Before pushing next, let's make sure all of these are gone.

---

*[pkg/sql/parser/datum.go, line 2852 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH4eR3psa8oELk4-KG:-KyH4eR3psa8oELk4-KH:bl5h39s) ([raw file](https://github.com/cockroachdb/cockroach/blob/15d7b4decebafbffb4ee3624322f2c0b960e35b9/pkg/sql/parser/datum.go#L2852)):*
> ```Go
> }
> 
> // DatumTypeSize returns a lower bound on the total size of a Datum
> ```

Do you think we need a test for this? A few variants, like `TTuple`, actually have a good amount of logic.

---

*[pkg/sql/parser/datum.go, line 2853 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH3hrvIx_yRvlaF8s7:-KyH3hrwhK7I73IVShfu:b-y8x6ks) ([raw file](https://github.com/cockroachdb/cockroach/blob/15d7b4decebafbffb4ee3624322f2c0b960e35b9/pkg/sql/parser/datum.go#L2853)):*
> ```Go
> 
> // DatumTypeSize returns a lower bound on the total size of a Datum
> // whose type is the receiver in bytes, including memory that is
> ```

This isn't a method, so there is no "receiver".

---

*[pkg/sql/parser/datum.go, line 2864 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH4XVwYmkBaEje0QXt:-KyH4XVxmi8rdUP_v8qK:b-mg8bua) ([raw file](https://github.com/cockroachdb/cockroach/blob/15d7b4decebafbffb4ee3624322f2c0b960e35b9/pkg/sql/parser/datum.go#L2864)):*
> ```Go
> 	// The following are composite types.
> 	switch ty := t.(type) {
> 	case tOid:
> ```

Why can't this be in the `baseDatumTypeSizes` map?

---

*[pkg/sql/parser/datum.go, line 39 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5QEx7OYOQJJPAGp_:-KyH5QEx7OYOQJJPAGpa:bpp6stj) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/datum.go#L39)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/roachpb"
> 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
> 	types "github.com/cockroachdb/cockroach/pkg/sql/sem/types"
> ```

not needed

---

*[pkg/sql/parser/datum.go, line 2715 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH9-AIIgFYfChvo8p9:-KyH9-AIIgFYfChvo8pA:bgiqoty) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/parser/datum.go#L2715)):*
> parser.types.T

I'd also make sure none of these exist.

---

*[pkg/sql/parser/eval.go, line 35 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5WImYP0CNt4STJ9y:-KyH5WImYP0CNt4STJ9z:b-qtozme) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/eval.go#L35)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/roachpb"
> 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
> 	types "github.com/cockroachdb/cockroach/pkg/sql/sem/types"
> ```

None of these are needed. I'll stop saying it from here on though.

---

*[pkg/sql/parser/type.go, line 22 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5nX-99yw_xwGd1-T:-KyH5nX-99yw_xwGd1-U:buirxis) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/type.go#L22)):*
> ```Go
> 
> // Type is a transitionary alias.
> type Type = types.Type
> ```

Our first type alias, aliasing type Type to types.Type!

---

*[pkg/sql/pgwire/v3.go, line 665 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyHEwAr_sU3eEl4AN_K:-KyHEwAr_sU3eEl4AN_L:bb687mf) ([raw file](https://github.com/cockroachdb/cockroach/blob/4ada76c849aa3ad1d9219d25ab9fe633c3b8f680/pkg/sql/pgwire/v3.go#L665)):*
> ```Go
> 			continue
> 		}
> 		v, ok := types.OidToType[t]
> ```

Why is this changing in this commit?

---

*[pkg/sql/sem/types/oid.go, line 287 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH6a2KyGDJ3xsFKF-M:-KyH6a2KyGDJ3xsFKF-N:b9qjnj2) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8ead5088200e81ec00c2dc2a5dd646f75a40e9/pkg/sql/sem/types/oid.go#L287)):*
> ```Go
> 	}
> 	// Base types.
> 	return baseTypeOids[t]
> ```

Should we panic if this lookup fails?

---

*[pkg/sql/sem/types/oid.go, line 57 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH7OFBcTcErGpeyjTE:-KyH7OFBcTcErGpeyjTF:b68b6dv) ([raw file](https://github.com/cockroachdb/cockroach/blob/110ea4ab7998f765a951ee21f6316301db3ff47e/pkg/sql/sem/types/oid.go#L57)):*
> ```Go
> )
> 
> // OidToType maps Postgres object IDs to CockroachDB types.
> ```

Add to the comment that we export the map instead of a method so that other packages can iterate over the map directly.

---

*[pkg/sql/sem/types/sql_name.go, line 34 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH6q297xNfqec59Ij9:-KyH6q297xNfqec59IjA:buy3mya) ([raw file](https://github.com/cockroachdb/cockroach/blob/11b876d804dba00f8ebe93fd8efddaf15240d54c/pkg/sql/sem/types/sql_name.go#L34)):*
> ```Go
> 		return SQLName(ty.Typ) + "[]"
> 	}
> 	return sqlTypeNames[t]
> ```

Again, should this panic if we're missing the type?

---

*[pkg/sql/sem/types/types.go, line 75 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyHDcX74d60EJ2BdC5E:-KyHDcX74d60EJ2BdC5F:b-mfp8u4) ([raw file](https://github.com/cockroachdb/cockroach/blob/4ada76c849aa3ad1d9219d25ab9fe633c3b8f680/pkg/sql/sem/types/types.go#L75)):*
> ```Go
> 	Any T = tAny{}
> 
> 	// TypesAnyNonArray contains all non-array types.
> ```

Should this be `AnyNonArray` now?

---

*[pkg/sql/sqlbase/table_test.go, line 31 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH9snJkbcDJJ4pYbFW:-KyH9snJkbcDJJ4pYbFX:b2r2qqe) ([raw file](https://github.com/cockroachdb/cockroach/blob/4280440d29149156e2e22ff7ecc7fb9c5b76b07d/pkg/sql/sqlbase/table_test.go#L31)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
> 	"github.com/cockroachdb/cockroach/pkg/util/randutil"
> 	"github.com/pkg/errors"
> ```

This should not have moved.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-:-KyHFXmU7VPlh79FEcRc:bumgmdp)*

RaduBerinde:
[nit] I don't think TypeNull means "unknown", I think it's a type where NULL is the only possible value (eg SELECT NULL).
I hesitate having an "unknown" type unless we really need it because it begs the question of whether a given type is unknown wherever we are given a type.

RaduBerinde:
LGTM

Out of curiosity, what is the thinking behind moving some things (Oid) to a single function with a `switch` instead of each type implementing them ?

knz:
Radu: the thinking is DRY -- less code. A lot of that functionality was repeating the same thing over and over across types.

> if a commit consists of simply running a sed command without any manual intervention, that command should be included in the commit message

Agreed. However none of the patches here required no intervention.

> Type families are types, just parametric types, so I'm not sure the logic holds.

That's not exactly correct IMHO. 

The type families are of a higher kind than parameterized types.

It's like the difference in ML between `'a list` and `list`. `'a list` is parameterized (with a free variable) but is a valid type that can be used to define functions, data structures etc. For example `fun fst = list.nth 0` has `'a list` as type of its argument. (Sometimes it's valid to have a free type variable left on concrete things!) Meanwhile, `list` with no parameter is not a type. I'd say it's a higher kind, a function from types to types which given a type (e.g. `int` but also possibly `'a`) gives another type (`int list` or `'a list`). 

In CockroachDB we thus have a (good) distinction between FamArray (the higher kind type, e.g. not valid as input type for builtin functions) and TypeAnyArray (a valid type for e.g. builtin input arguments).
We also have a (good) distinction between FamCollatedString (not a type -- there's nothing like "collated string over any locale", it's never valid as input type to a builtin function) and the particular instances of TCollatedString (all valid as input type for a builtin function). 
I don't really want to lose these distinctions.

> I think it actually makes the few places where the new "Fam" variables are used more confusing.

I actually think that using the different name has revealed places where we're probably not doing the right thing. I added TODOs for this purpose.

> In SQL an ARRAY is considered a type so why isn't it here?

1) no, ARRAY is not a single type. 
2) "array" is a type constructor. TypeAnyArray is the array type over a free type variable, and is a type. TypeInt2Array is the array type over the int2 oid type, and is a type. TArray is not a type, and nor is FamArray.

> should they even be in the types package?

That's a discussion for another PR.

> `Any` is parametric. Is it a type family?

No - it's a free type variable, and is (arguably) a type.

Also the discussion here is misstepping away from the distinction that was highlighted by the comments in the code: the requirement not to use `==` for things that are "type families" (higher-kinded types). "Any" wasn't given that treatment, so there was no reason for me to change it here.

I have updated the commit message accordingly.

---

Reviewed 1 of 66 files at r6.

---

*[pkg/sql/distsqlrun/expr_test.go, line 25 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH8XS3ZyNmjt0CjCkp:-KyLfO4M8KSLChd6adlS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/distsqlrun/expr_test.go#L25)):*

nit: move up
</blockquote></details>

Done.

---

*[pkg/sql/parser/col_types.go, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5N7RQWLxF_iPn8Sh:-KyLcRVSeo_soC4LKow-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/col_types.go#L24)):*

alias not needed
</blockquote></details>

Done.

---

*[pkg/sql/parser/constant.go, line 28 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH8exfgaUzS8s26F7F:-KyLfP-9xdvlRUoGkO5d:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/parser/constant.go#L28)):*

nit: move up
</blockquote></details>

Done.

---

*[pkg/sql/parser/create.go, line 28 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH8malYrJ7x4spkQ18:-KyLcTbiaESZj2d1JRep:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/parser/create.go#L28)):*

Before pushing next, let's make sure all of these are gone.
</blockquote></details>

Done.

---

*[pkg/sql/parser/datum.go, line 2852 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH4eR3psa8oELk4-KG:-KyLcWtaJBQ6mFT9dCrg:b-1y4ky8) ([raw file](https://github.com/cockroachdb/cockroach/blob/15d7b4decebafbffb4ee3624322f2c0b960e35b9/pkg/sql/parser/datum.go#L2852)):*

Do you think we need a test for this? A few variants, like `TTuple`, actually have a good amount of logic.
</blockquote></details>

It's out of scope for this PR. IF you think that's important, please create an issue and cc me.

---

*[pkg/sql/parser/datum.go, line 2853 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH3hrvIx_yRvlaF8s7:-KyLeD6NF0XscH0cjQtH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/15d7b4decebafbffb4ee3624322f2c0b960e35b9/pkg/sql/parser/datum.go#L2853)):*

This isn't a method, so there is no "receiver".
</blockquote></details>

Done.

---

*[pkg/sql/parser/datum.go, line 2864 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH4XVwYmkBaEje0QXt:-KyLeeSfudNysVByM9eO:b-uyf7ca) ([raw file](https://github.com/cockroachdb/cockroach/blob/15d7b4decebafbffb4ee3624322f2c0b960e35b9/pkg/sql/parser/datum.go#L2864)):*

Why can't this be in the `baseDatumTypeSizes` map?
</blockquote></details>

Clarified in comment.

---

*[pkg/sql/parser/datum.go, line 39 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5QEx7OYOQJJPAGp_:-KyLcUTCZ6UdZgwF3RqL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/datum.go#L39)):*

not needed
</blockquote></details>

Done.

---

*[pkg/sql/parser/datum.go, line 2715 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH9-AIIgFYfChvo8p9:-KyLcUvEwQ1GhFEZ7Dap:bsturw4) ([raw file](https://github.com/cockroachdb/cockroach/blob/551cddb344ec989d2e0258be2d4e97f3ac7a9fa8/pkg/sql/parser/datum.go#L2715)):*

> parser.types.T

I'd also make sure none of these exist.
</blockquote></details>

Fixed throughout.

---

*[pkg/sql/parser/eval.go, line 35 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH5WImYP0CNt4STJ9y:-KyLcd2GQ60nJaQvexVZ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3e36229669ecbb62f2e0caec8402195cdfecdb2b/pkg/sql/parser/eval.go#L35)):*

None of these are needed. I'll stop saying it from here on though.
</blockquote></details>

Done.

---

*[pkg/sql/pgwire/v3.go, line 665 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyHEwAr_sU3eEl4AN_K:-KyLceR5P3xpN3Nq-4Jc:b-du42dp) ([raw file](https://github.com/cockroachdb/cockroach/blob/4ada76c849aa3ad1d9219d25ab9fe633c3b8f680/pkg/sql/pgwire/v3.go#L665)):*

Why is this changing in this commit?
</blockquote></details>

Mistake -- fixed.

---

*[pkg/sql/sem/types/oid.go, line 287 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH6a2KyGDJ3xsFKF-M:-KyLggmiIrtrCJmmsJKQ:bpypyor) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8ead5088200e81ec00c2dc2a5dd646f75a40e9/pkg/sql/sem/types/oid.go#L287)):*

Should we panic if this lookup fails?
</blockquote></details>

Hmm I always forget that Go doesn't panic on map accesses... added.

---

*[pkg/sql/sem/types/oid.go, line 57 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH7OFBcTcErGpeyjTE:-KyLgRYKJOBJU0-Q2Jl8:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/110ea4ab7998f765a951ee21f6316301db3ff47e/pkg/sql/sem/types/oid.go#L57)):*

Add to the comment that we export the map instead of a method so that other packages can iterate over the map directly.
</blockquote></details>

Done.

---

*[pkg/sql/sem/types/sql_name.go, line 34 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH6q297xNfqec59Ij9:-KyLhWMJORddpnB6feAo:bjcs5dc) ([raw file](https://github.com/cockroachdb/cockroach/blob/11b876d804dba00f8ebe93fd8efddaf15240d54c/pkg/sql/sem/types/sql_name.go#L34)):*

Again, should this panic if we're missing the type?
</blockquote></details>

Indeed. Done.

---

*[pkg/sql/sem/types/types.go, line 75 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyHDcX74d60EJ2BdC5E:-KyLhqboc4Xy5BHc89CF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4ada76c849aa3ad1d9219d25ab9fe633c3b8f680/pkg/sql/sem/types/types.go#L75)):*

Should this be `AnyNonArray` now?
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/table_test.go, line 31 at r7](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyH9snJkbcDJJ4pYbFW:-KyLVc4bGi9G_Rt-2kP5:b-g5y6k4) ([raw file](https://github.com/cockroachdb/cockroach/blob/4280440d29149156e2e22ff7ecc7fb9c5b76b07d/pkg/sql/sqlbase/table_test.go#L31)):*

This should not have moved.
</blockquote></details>

Fixed.

I'm sorry I'm not responsible for what crlfmt/gofmt/goimports chooses to do. This is much too much work to change manually and I don't know how to automate, so I propose if there are any other left to revisit in a followup PR. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-:-KyLcFh-AMP7HtwBApH4:b-1fcl5l)*

jordanlewis:
<all files reviewed @ r22 but i can't tell reviewable that i've done that for some reason> 
:lgtm:

-1 on r14 and r15, the ones that replace Type.SQLName and Type.OID with type switches. I don't see what advantage that change gives us  - AFAICT there are no package dependencies being untangled there.

---

Reviewed 103 of 103 files at r12, 16 of 16 files at r13, 8 of 8 files at r14, 5 of 5 files at r15.
Review status: 0 of 103 files reviewed at latest revision, 19 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/datum.go, line 2929 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyM2cVM-2txCCSIFhB3:-KyM2cVM-2txCCSIFhB4:boz4ruw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d6e0167c88fffb53561d0d7ea162b868ae216156/pkg/sql/parser/datum.go#L2929)):*
> ```Go
> //
> // It holds for every Datum d that d.Size() >= DatumSize(d.ResolvedType())
> func DatumTypeSize(t Type) (uintptr, bool) {
> ```

It makes sense that you're doing this to untangle package dependencies, but I'd like to register my displeasure at replacing type methods (simple to find when implementing new types) with type switches (scattered all around the source code). In my opinion, like I've stated before, scattering the implementation of a type all around is a recipe for bugs when adding new types - it's hard to remember to add new types to all of the relevant type switches (or even find them)!

Just food for thought. I'd be interested in exploring a further refactor (with a new shared package dependency or something?) after this lands to convert these back to type methods.

---

*[pkg/sql/sem/types/oid.go, line 270 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyM5aKod5MQo-isbF44:-KyM5aKod5MQo-isbF45:bi3un0d) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2b1c4297cbe66001dd31e4ee68f89cf0508ec1d/pkg/sql/sem/types/oid.go#L270)):*
> ```Go
> 
> // Oid returns the type's Postgres object ID.
> func Oid(t Type) oid.Oid {
> ```

-1 on this patch - I don't see a good reason to switch this away from being a method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-:-KyM75eaUBtyO1t5KPbK:b-mgm81w)*

knz:
Review status: 0 of 103 files reviewed at latest revision, 19 unresolved discussions, all commit checks successful.

---

*[pkg/sql/parser/datum.go, line 2929 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyM2cVM-2txCCSIFhB3:-KyMApdmMpAmCR51aZqW:b-hngm0a) ([raw file](https://github.com/cockroachdb/cockroach/blob/d6e0167c88fffb53561d0d7ea162b868ae216156/pkg/sql/parser/datum.go#L2929)):*

It makes sense that you're doing this to untangle package dependencies, but I'd like to register my displeasure at replacing type methods (simple to find when implementing new types) with type switches (scattered all around the source code). In my opinion, like I've stated before, scattering the implementation of a type all around is a recipe for bugs when adding new types - it's hard to remember to add new types to all of the relevant type switches (or even find them)!

Just food for thought. I'd be interested in exploring a further refactor (with a new shared package dependency or something?) after this lands to convert these back to type methods.
</blockquote></details>

The problem you're going to hit though is that Go doesn't like adding new methods to types in a different package, nor does it support generics, nor does it support type classes
In general here we have a major clash between a productivity need to split independent things into independent packages, and what you feel "looks good" from past experience but *made impossible by Go*. I believe we have reached a tipping point where we should not let Go come in the way of making our work on this code base more scalable.

---

*[pkg/sql/sem/types/oid.go, line 270 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyM5aKod5MQo-isbF44:-KyMBSDQd-Rs0hgMKnAQ:b42eqq9) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2b1c4297cbe66001dd31e4ee68f89cf0508ec1d/pkg/sql/sem/types/oid.go#L270)):*

-1 on this patch - I don't see a good reason to switch this away from being a method.
</blockquote></details>

The reason is DRY.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818)*

RaduBerinde:
*[pkg/sql/sem/types/oid.go, line 270 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyM5aKod5MQo-isbF44:-KyME29DxZQfQpJFL0Lz:b-pud3cg) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2b1c4297cbe66001dd31e4ee68f89cf0508ec1d/pkg/sql/sem/types/oid.go#L270)):*

The reason is DRY.
</blockquote></details>

I'm with Jordan on this (though not very strongly). I don't think DRY is enough of a motivation, there are still a bunch of cases here. And there is a readability cost to spreading out the type-specific code, especially when the objects implement a common interface (there's an implicit expectation that everything specific to that type is in those methods).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818)*

RaduBerinde:
*[pkg/sql/sem/types/oid.go, line 270 at r14](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-KyM5aKod5MQo-isbF44:-KyMF7U0VIRS9AkxgYnw:bmw7qrf) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2b1c4297cbe66001dd31e4ee68f89cf0508ec1d/pkg/sql/sem/types/oid.go#L270)):*

I'm with Jordan on this (though not very strongly). I don't think DRY is enough of a motivation, there are still a bunch of cases here. And there is a readability cost to spreading out the type-specific code, especially when the objects implement a common interface (there's an implicit expectation that everything specific to that type is in those methods).
</blockquote></details>

BTW we could have a "base" struct that all type structs embed which has an`Oid()` function that panics. Then only the types that actually support it need to implement their own method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818)*

knz:
So to summarize the discussion I just had with Jordan:

- I'll revert that part of the patch because it's not in-scope for this PR anyway
- we still need a discussion about things, and it's possible that the changes made here will re-appear later; some example considerations:
   - the benefit to have the definition of the behavior "close to the type" is to be appropriately compared/opposed to having the definition of similar behavior across multiple types in the same place, to *share common behavior* and *reveal inconsistencies*. In an ideal world the language should allows us to switch between one layout and another at no cost (e.g. during reviews). But that is not permitted by Go.
   - separately, for this code like other code we've already considered, we might switch to using a single Go type with a header integer, so as to simplify / improve the allocation overheads, and that will mandate switches anyway
   - separately, from a perf perspective, dynamic method dispatch is strictly worse than a switch or a map (the discussion about whether a switch or a map is better from a perf perspective is an orthogonal concern, and that's to be decided on a case by case basis)

knz:
Edited the controversial changes away from the PR. Will merge on green.

---

Review status: 0 of 102 files reviewed at latest revision, 19 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19818#-:-KyMQn_W5uqN-3jMz3h8:b-ric196)*

nvanbenschoten:
> from a perf perspective, dynamic method dispatch is strictly worse than a switch or a map (the discussion about whether a switch or a map is better from a perf perspective is an orthogonal concern, and that's to be decided on a case by case basis)

We shouldn't be saying anything about performance without backing it up with real numbers and sources, especially in such a definitive tone that makes it sound like the experiment has already been run.

```
package src

import "testing"

type a struct{ x bool }
type b struct{ x string }
type c struct{ x int }
type d struct{ x int }

type i interface {
	i() bool
}

func (v a) i() bool { return true }
func (v b) i() bool { return false }
func (v c) i() bool { return false }
func (v d) i() bool { return true }

var aVar i = a{x: true}
var bVar i = b{x: "str"}
var cVar i = c{x: 123}
var dVar i = d{x: 321}
var vars = []i{aVar, bVar, cVar, dVar, aVar, bVar, cVar, dVar, aVar, bVar, cVar, dVar}

var iMap = map[i]bool{
	aVar: true,
	bVar: false,
	cVar: false,
	dVar: true,
}

func BenchmarkDispatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var res bool
		for _, v := range vars {
			res = v.i()
		}
		_ = res
	}
}

func BenchmarkMap(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var res bool
		for _, v := range vars {
			res = iMap[v]
		}
		_ = res
	}
}

func BenchmarkSwitch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var res bool
		for _, v := range vars {
			switch v {
			case aVar:
				res = true
			case bVar:
				res = false
			case cVar:
				res = false
			case dVar:
				res = true
			}
		}
		_ = res
	}
}
```

```
name        time/op
Dispatch-4   192ns ± 7%
Map-4       1.76µs ± 9%
Switch-4     462ns ± 7%
```

PR #19819, opened 2017-11-06
storage: enhance the proposal quota logic

### LINKED ISSUE #18342: storage: proposal quota and bad replicas
#17524 revealed a bit of fragility in the use of proposal quota: a replica that is unable to make progress for some reason will freeze up the entire range even though 2 good replicas remain. The specific bug causing a replica to be unable to make progress is now fixed, but we should consider enhancing the proposal quota logic to lessen the impact of such bugs in the future. 

The current proposal quota logic "finds the minimum index that active followers have acknowledged" and uses that value to determine when quota should be returned to the pool. An "active follower" is one that is on a healthy node where health is determined by the per-connection RPC heartbeats. This is an insufficient definition of health if a particular replica on an otherwise good node is having problems. When that occurs proposal quota is exhausted and the range will stop accepting writes.

Enhancing the proposal quota logic should be relatively straightforward. Rather than simply consider connection health, we could also have some measure of whether a replica has made any progress at all within a certain time period. For example, we could consider a follower inactive if its `progress.Match` has not changed in the last 10 seconds. 

Cc @nvanbenschoten, @tschottdorf

a6802739:
Fixes #18342.  

@petermattis @bdarnell , I wonder If I should a test for this PR, but I found it's quite difficult to construct the test scenario.

### DIFF
--- a/pkg/storage/client_raft_test.go
+++ b/pkg/storage/client_raft_test.go
@@ -1894,6 +1894,84 @@ func TestQuotaPool(t *testing.T) {
 	}
 }
 
+// TestWedgedReplicaDetection verifies that a leader replica is able to correctly detect a wedged follower replica.
+func TestWedgedReplicaDetection(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+
+	const numReplicas = 3
+	const rangeID = 1
+
+	var manuals []*hlc.ManualClock
+	var clocks []*hlc.Clock
+	for i := 0; i < numReplicas; i++ {
+		manuals = append(manuals, hlc.NewManualClock(123))
+		clocks = append(clocks, hlc.NewClock(manuals[i].UnixNano, time.Nanosecond))
+	}
+
+	sc := storage.TestStoreConfig(nil)
+	// Suppress timeout-based elections to avoid leadership changes in ways
+	// this test doesn't expect.
+	sc.RaftElectionTimeoutTicks = 100000
+	mtc := &multiTestContext{storeConfig: &sc, clocks: clocks}
+	mtc.Start(t, numReplicas)
+	defer mtc.Stop()
+
+	mtc.replicateRange(rangeID, 1, 2)
+
+	leaderRepl := mtc.getRaftLeader(rangeID)
+	// We need four steps in this test:
+	// 1. Wedged the specified follower replica.
+	// 2. Increment leader's clock past MinQuotaReplicaLivenessDuration+1.
+	// 3. Send An request to the leader replica.
+	// 4. Check whether leader replica's last update time for the follower replica exceeds
+	// MinQuotaReplicaLivenessDuration.
+	followerRepl := func() *storage.Replica {
+		for _, store := range mtc.stores {
+			repl, err := store.GetReplica(rangeID)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if repl == leaderRepl {
+				continue
+			}
+			return repl
+		}
+		return nil
+	}()
+	if followerRepl == nil {
+		t.Fatal("could not get a handle on a follower replica")
+	}
+
+	// NB: See TestRaftBlockedReplica/#9914 for why we use a separate
+	// goroutine.
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		followerRepl.RaftLock()
+		wg.Done()
+	}()
+	wg.Wait()
+	leaderReplicaID := leaderRepl.ReplicaID()
+	manuals[leaderReplicaID].Increment(int64(storage.MinQuotaReplicaLivenessDuration) + int64(1))
+
+	key := roachpb.Key("k")
+	value := []byte("value")
+	_, pErr := client.SendWrapped(context.Background(), leaderRepl, putArgs(key, value))
+	if pErr != nil {
+		t.Fatal(pErr)
+	}
+
+	now := timeutil.Unix(0, clocks[leaderReplicaID].PhysicalNow())
+	followerIsNotWedged := func(lastUpdateTime time.Time) bool {
+		return !lastUpdateTime.IsZero() && now.Sub(lastUpdateTime) <= storage.MinQuotaReplicaLivenessDuration
+	}
+	lastUpdateTime := leaderRepl.GetLastUpdateTimeForReplica(followerRepl.ReplicaID())
+	if followerIsNotWedged(lastUpdateTime) {
+		t.Fatalf("expected time duration for replica %d exceed %d, but got %d.", followerRepl.ReplicaIDLocked(), storage.MinQuotaReplicaLivenessDuration, now.Sub(lastUpdateTime))
+	}
+	followerRepl.RaftUnlock()
+}
+
 // TestRaftHeartbeats verifies that coalesced heartbeats are correctly
 // suppressing elections in an idle cluster.
 func TestRaftHeartbeats(t *testing.T) {
--- a/pkg/storage/helpers_test.go
+++ b/pkg/storage/helpers_test.go
@@ -240,8 +240,20 @@ func NewTestStorePool(cfg StoreConfig) *StorePool {
 	)
 }
 
+func (r *Replica) ReplicaID() roachpb.ReplicaID {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.ReplicaIDLocked()
+}
+
 func (r *Replica) ReplicaIDLocked() roachpb.ReplicaID {
 	return r.mu.replicaID
+}
+
+func (r *Replica) GetLastUpdateTimeForReplica(replicaID roachpb.ReplicaID) time.Time {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.mu.lastUpdateTimes[replicaID]
 }
 
 func (r *Replica) DescLocked() *roachpb.RangeDescriptor {
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -88,6 +88,9 @@ const (
 	splitTxnName         = "split"
 	mergeTxnName         = "merge"
 
+	// MinQuotaReplicaLivenessDuration is the minimum duration that we could make sure
+	// the follower of this replica is active.
+	MinQuotaReplicaLivenessDuration   = 10 * time.Second
 	defaultReplicaRaftMuWarnThreshold = 500 * time.Millisecond
 )
 
@@ -409,6 +412,9 @@ type Replica struct {
 		// we know that the replica has caught up.
 		lastReplicaAdded     roachpb.ReplicaID
 		lastReplicaAddedTime time.Time
+
+		// The most recently updated time for each follower of this range.
+		lastUpdateTimes map[roachpb.ReplicaID]time.Time
 
 		// The last seen replica descriptors from incoming Raft messages. These are
 		// stored so that the replica still knows the replica descriptors for itself
@@ -1010,6 +1016,7 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 			r.mu.proposalQuota.close()
 		}
 		r.mu.proposalQuota = nil
+		r.mu.lastUpdateTimes = nil
 		r.mu.quotaReleaseQueue = nil
 		r.mu.commandSizes = nil
 		return
@@ -1036,6 +1043,7 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 			// offset this we reset the quota pool whenever leadership changes
 			// hands.
 			r.mu.proposalQuota = newQuotaPool(defaultProposalQuota)
+			r.mu.lastUpdateTimes = make(map[roachpb.ReplicaID]time.Time)
 			r.mu.commandSizes = make(map[storagebase.CmdIDKey]int)
 		} else if r.mu.proposalQuota != nil {
 			// We're becoming a follower.
@@ -1044,6 +1052,7 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 			// goroutines (if any).
 			r.mu.proposalQuota.close()
 			r.mu.proposalQuota = nil
+			r.mu.lastUpdateTimes = nil
 			r.mu.quotaReleaseQueue = nil
 			r.mu.commandSizes = nil
 		}
@@ -1070,16 +1079,20 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 	// Find the minimum index that active followers have acknowledged.
 	minIndex := status.Commit
 
+	now := timeutil.Unix(0, r.store.Clock().PhysicalNow())
 	for _, rep := range r.mu.state.Desc.Replicas {
-		// Only consider followers that have "healthy" RPC connections. We
-		// don't use node liveness here as doing so could lead to deadlock
-		// unless we avoided enforcing proposal quota for node liveness ranges.
+		// Only consider followers that that have "healthy" RPC connections and
+		// have made communication with leader in the last MinQuotaReplicaLivenessDuration.
 		if r.store.cfg.Transport.resolver != nil {
 			addr, err := r.store.cfg.Transport.resolver(rep.NodeID)
 			if err != nil {
 				continue
 			}
 			if err := r.store.cfg.Transport.rpcContext.ConnHealth(addr.String()); err != nil {
+				continue
+			}
+			if lastUpdateTime, ok := r.mu.lastUpdateTimes[rep.ReplicaID]; !ok ||
+				now.Sub(lastUpdateTime) > MinQuotaReplicaLivenessDuration {
 				continue
 			}
 		}
@@ -1105,8 +1118,8 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 
 	if r.mu.proposalQuotaBaseIndex < minIndex {
 		// We've persisted minIndex - r.mu.proposalQuotaBaseIndex entries to
-		// the raft log on all replicas with "healthy" RPC connections since
-		// last we checked, we 'should' be able to release the difference back to
+		// the raft log on all 'active' replicas since last we checked,
+		// we 'should' be able to release the difference back to
 		// the quota pool. But consider the scenario where we have a single
 		// replica that we're writing to, we only construct the
 		// quotaReleaseQueue when entries 'come out' of Raft via
@@ -1709,6 +1722,23 @@ func (r *Replica) setQueueLastProcessed(
 ) error {
 	key := keys.QueueLastProcessedKey(r.Desc().StartKey, queue)
 	return r.store.DB().PutInline(ctx, key, &timestamp)
+}
+
+func (r *Replica) setLastUpdateTimeForReplicaLocked(replicaID roachpb.ReplicaID) {
+	if r.mu.replicaID == r.mu.leaderID {
+		r.mu.lastUpdateTimes[replicaID] = r.store.Clock().PhysicalTime()
+	}
+}
+
+func (r *Replica) setLastUpdateTimesLocked() {
+	if r.mu.replicaID == r.mu.leaderID {
+		for _, rep := range r.mu.state.Desc.Replicas {
+			if rep.ReplicaID == r.mu.replicaID {
+				continue
+			}
+			r.mu.lastUpdateTimes[rep.ReplicaID] = r.store.Clock().PhysicalTime()
+		}
+	}
 }
 
 // RaftStatus returns the current raft status of the replica. It returns nil
@@ -3294,6 +3324,7 @@ func (r *Replica) unquiesceLocked() {
 			log.Infof(ctx, "unquiescing")
 		}
 		r.mu.quiescent = false
+		r.setLastUpdateTimesLocked()
 	}
 }
 
@@ -3320,6 +3351,7 @@ func (r *Replica) stepRaftGroup(req *RaftMessageRequest) error {
 		if req.Message.Type == raftpb.MsgApp {
 			r.setEstimatedCommitIndexLocked(req.Message.Commit)
 		}
+		r.setLastUpdateTimeForReplicaLocked(req.FromReplica.ReplicaID)
 		return false, /* unquiesceAndWakeLeader */
 			raftGroup.Step(req.Message)
 	})


petermattis:
This looks like a good start. Yes, this absolute needs a test. Writing a test is often the hardest part of these changes. I haven't looked at it recently, but I'd suggest taking a look at `TestQuotaPool`. Either extend it or use it as inspiration for a new test. The test will need to wedge one of the replicas while leaving the node up. This can be accomplished by grabbing `Replica.raftMu` on the replica you want to wedge.

---

---

*[pkg/storage/replica.go, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KyGpUSo73UcWLgxSs0E:-KyGpUSo73UcWLgxSs0F:brg8tyb) ([raw file](https://github.com/cockroachdb/cockroach/blob/89728dd71e9a029ddc040c1be71dd40dbe246943/pkg/storage/replica.go#L88)):*
> ```Go
> 	mergeTxnName         = "merge"
> 
> 	minUpdatesDuration                = 10 * time.Second
> ```

Let's name this constant so that it refers to quota. Something like: `minQuotaReplicaLivenessDuration`. 

---

*[pkg/storage/replica.go, line 666 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KyGp5zuWKWVrlQJuZwj:-KyGp5zuWKWVrlQJuZwk:b-obw4ze) ([raw file](https://github.com/cockroachdb/cockroach/blob/89728dd71e9a029ddc040c1be71dd40dbe246943/pkg/storage/replica.go#L666)):*
> ```Go
> 
> 	r.mu.proposals = map[storagebase.CmdIDKey]*ProposalData{}
> 	r.mu.lastUpdateTimes = make(map[uint64]time.Time)
> ```

I think this map only needs to be initialized when a replica becomes the leader. Also, it should be set to nil when a replica loses leadership. Take a look at where `Replica.mu.proposalQuota` is initialized and cleared.

---

*[pkg/storage/replica.go, line 1043 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KyGpKchQvHqPCXQiBcx:-KyGpKchQvHqPCXQiBcy:bmuaw2e) ([raw file](https://github.com/cockroachdb/cockroach/blob/89728dd71e9a029ddc040c1be71dd40dbe246943/pkg/storage/replica.go#L1043)):*
> ```Go
> 
> 	for _, rep := range r.mu.state.Desc.Replicas {
> 		now := timeutil.Unix(0, r.store.Clock().PhysicalNow())
> ```

This can be moved outside of the loop. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-:-KyGoiCURggKtvEhkXKV:b-bz4ve7)*

a6802739:
@petermattis , I have write a test for the PR. 

Still not sure if it's appropriate.

Thank your for your review.

nvanbenschoten:
Reviewed 1 of 4 files at r2.
Review status: 1 of 4 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica.go, line 381 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-Kyq4i98fnbnohll2FZl:-Kyq4i99tbtzXaUOAUy_:b-lbe8lx) ([raw file](https://github.com/cockroachdb/cockroach/blob/89728dd71e9a029ddc040c1be71dd40dbe246943/pkg/storage/replica.go#L381)):*
> ```Go
> 
> 		// The most recently updated time for each follower of this range.
> 		lastUpdateTimes map[uint64]time.Time
> ```

The `uint64` should be replaced with a `roachpb.ReplicaID`.

---

*[pkg/storage/replica.go, line 1091 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-Kyq6Ql6VhxluQ1t1CAv:-Kyq6Ql6VhxluQ1t1CAw:b-12jkhp) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5043ac51db73f9aab80ef4023c95409aae1b71e/pkg/storage/replica.go#L1091)):*
> ```Go
> 	now := timeutil.Unix(0, r.store.Clock().PhysicalNow())
> 	for _, rep := range r.mu.state.Desc.Replicas {
> 		// Only consider followers that have made communication with leader in the last 10s.
> ```

The 10s is parameterized in `minQuotaReplicaLivenessDuration`. Make this `Only consider followers that have made communication with leader in the last minQuotaReplicaLivenessDuration.`

Also, what was the logic behind getting rid of the healthy RPC connection check? I'd expect that we'd still want that as well. So the definition of health would become `healthyRPCConnection && updatedInLast10s`. Is that what you had in mind @petermattis?

Finally, I'm not sure this will interact well with quiescent replicas. How will this work if a leader quiesces, waits 11 seconds, and then unquiesces to find that it hasn't received any updates from any of its followers for more than `minQuotaReplicaLivenessDuration`? I guess the worst that can happen is that the leader won't consider any followers as live and will release all of its quota. Is this ok?

---

*[pkg/storage/replica.go, line 1117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KyqCJlHL3e9_N6uzqGB:-KyqCJlHL3e9_N6uzqGC:b-xcq4vh) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5043ac51db73f9aab80ef4023c95409aae1b71e/pkg/storage/replica.go#L1117)):*
> ```Go
> 
> 	if r.mu.proposalQuotaBaseIndex < minIndex {
> 		// We've persisted minIndex - r.mu.proposalQuotaBaseIndex entries to
> ```

This comment needs an update.

---

*[pkg/storage/replica.go, line 1728 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-Kyq5JhgdhbVZhhjVeVT:-Kyq5JhgdhbVZhhjVeVU:b-cl6x0d) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5043ac51db73f9aab80ef4023c95409aae1b71e/pkg/storage/replica.go#L1728)):*
> ```Go
> }
> 
> func (r *Replica) getLastUpdateTimeForReplicaLocked(replicaID roachpb.ReplicaID) (time.Time, bool) {
> ```

This method doesn't need to exist, especially once we change the map key to a `ReplicaID`.

---

*[pkg/storage/store.go, line 2988 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-Kyq8oBh9x8W9CA2SMbC:-Kyq8oBh9x8W9CA2SMbD:b1tdyd2) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5043ac51db73f9aab80ef4023c95409aae1b71e/pkg/storage/store.go#L2988)):*
> ```Go
> 	r.mu.Lock()
> 	if r.mu.replicaID == r.mu.leaderID {
> 		r.setLastUpdateTimeForReplicaLocked(req.FromReplica.ReplicaID)
> ```

I think a better place for this would be in `stepRaftGroup`. There, we have already locked the `Replica` and have access to the `*RaftMessageRequest`. It is also called by both the non-snapshot and the snapshot paths, so it feels like a more natural place for replica update tracking (even though the snapshot path won't ever need this at the moment since we only keep track on the leader, but this could always change).

Also, for the sake of encapsulation, I think the `r.mu.replicaID == r.mu.leaderID` check should go in `setLastUpdateTimeForReplicaLocked`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819)*

pkg/storage/replica.go
```diff
@@ -410,6 +411,9 @@ type Replica struct {
 		lastReplicaAdded     roachpb.ReplicaID
 		lastReplicaAddedTime time.Time
 
+		// The most recently updated time for each follower of this range.
+		lastUpdateTimes map[roachpb.ReplicaID]time.Time
```
a6802739: Done.

pkg/storage/replica.go
```diff
@@ -1079,18 +1086,12 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
...
-		// unless we avoided enforcing proposal quota for node liveness ranges.
-		if r.store.cfg.Transport.resolver != nil {
-			addr, err := r.store.cfg.Transport.resolver(rep.NodeID)
-			if err != nil {
-				continue
-			}
-			if err := r.store.cfg.Transport.rpcContext.ConnHealth(addr.String()); err != nil {
-				continue
-			}
+		// Only consider followers that have made communication with leader in the last minQuotaReplicaLivenessDuration.
```
a6802739: Done.

pkg/storage/replica.go
```diff
@@ -1720,6 +1721,12 @@ func (r *Replica) setQueueLastProcessed(
 	return r.store.DB().PutInline(ctx, key, &timestamp)
 }
 
+func (r *Replica) setLastUpdateTimeForReplicaLocked(replicaID roachpb.ReplicaID) {
+	if r.mu.replicaID == r.mu.leaderID {
```
a6802739: Done.

pkg/storage/replica.go
```diff
@@ -3310,6 +3317,7 @@ func (r *Replica) stepRaftGroup(req *RaftMessageRequest) error {
 		if req.Message.Type == raftpb.MsgApp {
 			r.setEstimatedCommitIndexLocked(req.Message.Commit)
 		}
+		r.setLastUpdateTimeForReplicaLocked(req.FromReplica.ReplicaID)
```
a6802739: Done.

a6802739 (commented):
@nvanbenschoten , @petermattis, How about when the leader's status changed from`quiescent` to `unquiescent`, we just update each follower's update time to the newest timestamp?

nvanbenschoten:
This is better @a6802739! My question about still considering the RPC connection health remains.

> How about when the leader's status changed fromquiescent to unquiescent, we just update each follower's update time to the newest timestamp?

That approach makes sense to me.

---

Reviewed 1 of 4 files at r3, 1 of 1 files at r4.
Review status: 1 of 3 files reviewed at latest revision, 17 unresolved discussions, all commit checks successful.

---

*[pkg/storage/client_raft_test.go, line 1898 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KzOXeal20X8fExlHmQX:-KzOXeal20X8fExlHmQY:b-xzfd83) ([raw file](https://github.com/cockroachdb/cockroach/blob/cfd896dbf0d0b40a55e1e2b15dce628050468be1/pkg/storage/client_raft_test.go#L1898)):*
> ```Go
> 
> // TestWedgedReplica could help us to correctly detect out the wedged replica.
> func TestWedgedReplica(t *testing.T) {
> ```

This name and comment could be improved. Maybe `TestWedgedReplicaDetection verifies that a leader replica is able to correctly detect a wedged follower replica.`

---

*[pkg/storage/client_raft_test.go, line 1952 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KzOXZCNNtF5xP4DKTk-:-KzOXZCOhZ4M-zupSQQB:bflf169) ([raw file](https://github.com/cockroachdb/cockroach/blob/cfd896dbf0d0b40a55e1e2b15dce628050468be1/pkg/storage/client_raft_test.go#L1952)):*
> ```Go
> 	wg.Wait()
> 	leaderReplicaID := leaderRepl.ReplicaID()
> 	manuals[leaderReplicaID].Increment(int64(10 * time.Second))
> ```

`s/10 * time.Second/minQuotaReplicaLivenessDuration`

---

*[pkg/storage/client_raft_test.go, line 1952 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KzOXum2NHItvjSj4p8q:-KzOXum2NHItvjSj4p8r:beqg4mp) ([raw file](https://github.com/cockroachdb/cockroach/blob/cfd896dbf0d0b40a55e1e2b15dce628050468be1/pkg/storage/client_raft_test.go#L1952)):*
> ```Go
> 	wg.Wait()
> 	leaderReplicaID := leaderRepl.ReplicaID()
> 	manuals[leaderReplicaID].Increment(int64(10 * time.Second))
> ```

Could you add some comments to describe what's going on here. They can be as simple as:
- `Increment leader's clock past minQuotaReplicaLivenessDuration`
- `Send request to leader replica to ...`
- `Check leader replica's last update time for the follower replica`

---

*[pkg/storage/client_raft_test.go, line 1963 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KzOYNc3c2IJTKfiWSYS:-KzOYNc3c2IJTKfiWSYT:b-42xb0) ([raw file](https://github.com/cockroachdb/cockroach/blob/cfd896dbf0d0b40a55e1e2b15dce628050468be1/pkg/storage/client_raft_test.go#L1963)):*
> ```Go
> 	now := timeutil.Unix(0, clocks[leaderReplicaID].PhysicalNow())
> 	lastUpdateTime := leaderRepl.GetLastUpdateTimeForReplica(followerRepl.ReplicaID())
> 	if expectedDuration := 10 * time.Second; !lastUpdateTime.IsZero() && now.Sub(lastUpdateTime) < expectedDuration {
> ```

So this test should pass if `lastUpdateTime.IsZero()`?

Also, we're again hard-coding `minQuotaReplicaLivenessDuration`.

---

*[pkg/storage/client_raft_test.go, line 1963 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KzOZ7Zg9jPLBB_f8MNm:-KzOZ7Zg9jPLBB_f8MNn:b-bdluu2) ([raw file](https://github.com/cockroachdb/cockroach/blob/cfd896dbf0d0b40a55e1e2b15dce628050468be1/pkg/storage/client_raft_test.go#L1963)):*
> now.Sub(lastUpdateTime) < expectedDuration

Isn't the definition for a wedged replica `now.Sub(lastUpdateTime) > minQuotaReplicaLivenessDuration`, not `now.Sub(lastUpdateTime) >= minQuotaReplicaLivenessDuration`? Maybe it's worth pulling this into a function somewhere.

---

*[pkg/storage/helpers_test.go, line 307 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-KzOW6BTxitgbrCQImlM:-KzOW6BTxitgbrCQImlN:bhhin8v) ([raw file](https://github.com/cockroachdb/cockroach/blob/cfd896dbf0d0b40a55e1e2b15dce628050468be1/pkg/storage/helpers_test.go#L307)):*
> ```Go
> 	r.mu.Lock()
> 	defer r.mu.Unlock()
> 	r.mu.lastUpdateTimes = make(map[roachpb.ReplicaID]time.Time)
> ```

Should this just be added into `InitQuotaPool`? That method sets all the same fields as we set in `updateProposalQuotaRaftMuLocked`, so it probably makes sense to keep all initialization together.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-:-KzOVOE4E2oMi0ccq5g5:boe5aog)*

petermattis:
Generally looks good and I agree with @nvanbenschoten's comments.

---

Review status: 1 of 3 files reviewed at latest revision, 17 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica.go, line 1091 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-Kyq6Ql6VhxluQ1t1CAv:-KzOtjOduaFoEaD239tW:brdswlx) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5043ac51db73f9aab80ef4023c95409aae1b71e/pkg/storage/replica.go#L1091)):*

The 10s is parameterized in `minQuotaReplicaLivenessDuration`. Make this `Only consider followers that have made communication with leader in the last minQuotaReplicaLivenessDuration.`

Also, what was the logic behind getting rid of the healthy RPC connection check? I'd expect that we'd still want that as well. So the definition of health would become `healthyRPCConnection && updatedInLast10s`. Is that what you had in mind @petermattis?

Finally, I'm not sure this will interact well with quiescent replicas. How will this work if a leader quiesces, waits 11 seconds, and then unquiesces to find that it hasn't received any updates from any of its followers for more than `minQuotaReplicaLivenessDuration`? I guess the worst that can happen is that the leader won't consider any followers as live and will release all of its quota. Is this ok?
</blockquote></details>

Quiescent replicas shouldn't have any outstanding quota (if they did, they wouldn't quiesce). Is there a concern here that I'm not seeing?

I also imagined that the new logic would be in addition to checking the RPC connection health. The RPC connection health is a faster failure detector if the remote node dies. The replica update check detects if a specific replica on an otherwise healthy node is wedged.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-:-KzOuTMOmODnNZ9l9t39:bpjlevr)*

nvanbenschoten:
Review status: 1 of 3 files reviewed at latest revision, 17 unresolved discussions, all commit checks successful.

---

*[pkg/storage/replica.go, line 1091 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-Kyq6Ql6VhxluQ1t1CAv:-KzPGN6kasO44-9_kpxB:b-yvqw9n) ([raw file](https://github.com/cockroachdb/cockroach/blob/a5043ac51db73f9aab80ef4023c95409aae1b71e/pkg/storage/replica.go#L1091)):*
> Quiescent replicas shouldn't have any outstanding quota (if they did, they wouldn't quiesce). Is there a concern here that I'm not seeing?

That's true, but once the leader unquiesces it will begin acquiring quota again. If we don't do anything extra here (and a quiescence phase takes longer than `minQuotaReplicaLivenessDuration`), then follower replicas won't begin counting towards the newly outstanding quota until after they have responded to the leader, even though they were all live when quiescing began. 

This might be ok in practice. In fact, it may even be desirable because unquiescing has a high chance of finding dead nodes (although this should be handled by the RPC health check). Either way, I think we need to be explicit about the behavior we want here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819)*

pkg/storage/client_raft_test.go
```diff
@@ -1894,6 +1894,84 @@ func TestQuotaPool(t *testing.T) {
 	}
 }
 
+// TestWedgedReplicaDetection verifies that a leader replica is able to correctly detect a wedged follower replica.
```
a6802739: Done

pkg/storage/client_raft_test.go
```diff
@@ -1894,6 +1894,84 @@ func TestQuotaPool(t *testing.T) {
...
+
+	key := roachpb.Key("k")
+	value := []byte("value")
+	_, pErr := client.SendWrapped(context.Background(), leaderRepl, putArgs(key, value))
+	if pErr != nil {
+		t.Fatal(pErr)
+	}
+
+	now := timeutil.Unix(0, clocks[leaderReplicaID].PhysicalNow())
+	followerIsNotWedged := func(lastUpdateTime time.Time) bool {
```
a6802739: I make a function here.

pkg/storage/client_raft_test.go
```diff
@@ -1894,6 +1894,84 @@ func TestQuotaPool(t *testing.T) {
...
+		return !lastUpdateTime.IsZero() && now.Sub(lastUpdateTime) <= storage.MinQuotaReplicaLivenessDuration
```
a6802739: Done

pkg/storage/client_raft_test.go
```diff
@@ -1894,6 +1894,84 @@ func TestQuotaPool(t *testing.T) {
...
+	// this test doesn't expect.
+	sc.RaftElectionTimeoutTicks = 100000
+	mtc := &multiTestContext{storeConfig: &sc, clocks: clocks}
+	mtc.Start(t, numReplicas)
+	defer mtc.Stop()
+
+	mtc.replicateRange(rangeID, 1, 2)
+
+	leaderRepl := mtc.getRaftLeader(rangeID)
+	// We need four steps in this test:
```
a6802739: Thanks for your comments!

Forgive my poor English expression. @nvanbenschoten

pkg/storage/helpers_test.go
```diff
@@ -287,6 +299,8 @@ func (r *Replica) InitQuotaPool(quota int64) {
 	r.mu.proposalQuota = newQuotaPool(quota)
 	r.mu.quotaReleaseQueue = nil
 	r.mu.commandSizes = make(map[storagebase.CmdIDKey]int)
+	r.mu.lastUpdateTimes = make(map[roachpb.ReplicaID]time.Time)
```
a6802739: Done.

pkg/storage/replica.go
```diff
@@ -1079,10 +1088,10 @@ func (r *Replica) updateProposalQuotaRaftMuLocked(
 	// Find the minimum index that active followers have acknowledged.
 	minIndex := status.Commit
 
+	now := timeutil.Unix(0, r.store.Clock().PhysicalNow())
 	for _, rep := range r.mu.state.Desc.Replicas {
-		// Only consider followers that have "healthy" RPC connections. We
-		// don't use node liveness here as doing so could lead to deadlock
-		// unless we avoided enforcing proposal quota for node liveness ranges.
+		// Only consider followers that that have "healthy" RPC connections and
+		// have made communication with leader in the last MinQuotaReplicaLivenessDuration.
```
a6802739: I take the logic of check "healthy" RPC back now.

pkg/storage/replica.go
```diff
@@ -3297,6 +3327,7 @@ func (r *Replica) unquiesceLocked() {
 			log.Infof(ctx, "unquiescing")
 		}
 		r.mu.quiescent = false
+		r.setLastUpdateTimesLocked()
```
a6802739: Now when the leader becomes from `quiescent` to `unquiescent`, it will update each follower's lastUpdateTime.

a6802739 (commented):
@nvanbenschoten  @petermattis  Thank you for your review!

nvanbenschoten:
:lgtm: Thanks for all the help here @a6802739!

---

Reviewed 1 of 3 files at r5, 1 of 2 files at r7.
Review status: 2 of 3 files reviewed at latest revision, 22 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19819#-:-KzyQfaqHc0GP38uoUpb:b-wni4jd)*

PR #19795, opened 2017-11-03
build: fix compilation on arm64

nvb:
Fixes #18319.

This change makes two adjustments to fix the build on arm64:
- Avoid passing the `-sse3` flag to RocksDB when it's not available
- Add preprocessor directives to avoid sse4.2 on ARM

I'm not sure we'll actually want to merge this if we plan to update RocksDB soon,
but it demonstrates the steps we'll need to take to get ARM builds up and running
again.

PR #19824, opened 2017-11-06
Update rocksdb to 5.9

mberhault:
There are still a few SSE patches to apply (see https://github.com/cockroachdb/rocksdb/commits/nvanbenschoten/fixArm)

### DIFF
--- a/c-deps/libroach/db.cc
+++ b/c-deps/libroach/db.cc
@@ -2573,7 +2573,7 @@ DBStatus DBSstFileWriterOpen(DBSstFileWriter* fw) {
 }
 
 DBStatus DBSstFileWriterAdd(DBSstFileWriter* fw, DBKey key, DBSlice val) {
-  rocksdb::Status status = fw->rep.Add(EncodeKey(key), ToSlice(val));
+  rocksdb::Status status = fw->rep.Put(EncodeKey(key), ToSlice(val));
   if (!status.ok()) {
     return ToDBStatus(status);
   }


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19824#-:-KyGsahFHgrNViVJbfBX:bnfp4nl)*

benesch:
@mberhault want me to cherry-pick the ARM fixes since I'm planning to upstream them anyway?

mberhault:
That would be nice. I took a quick peek, and it looks like they're going to conflict.

nvanbenschoten:
@benesch which ARM fixes are you planning on upstreaming? The upstream [`crc32c.cc`](https://github.com/facebook/rocksdb/blob/master/util/crc32c.cc) is pretty different than the old revision we were working on, and I think https://github.com/facebook/rocksdb/commit/5449c0990b32ac0fcc69faeed9bada5ff1fa4a8a already addressed the same issue that #19795 did. There might not be anything to do on our end.

benesch:
Yeah, they added fast-CRC32 on PPC support upstream. which is why the file looks so different. The underlying problem I was trying to solve is still there, though. If you compile across the board with `-msse4.2`, like RocksDB does when you say `USE_SSE42=1`, GCC will toss some SSE4.2 instructions in other files which aren't guarded by a runtime `cpuid` check. So it's a complete waste to guard the CRC32 instructions behind a `IsFastCrc32Supported` check, because the binary will crash in other SSE4.2-using functions on older CPUs. `target("sse4.2")` solves that, but it broke the ARM build until you fixed it.

benesch:
Ok, here's a patch for 5.9: https://github.com/benesch/rocksdb/tree/5.9-sse

@nvanbenschoten could you verify that works on ARM?

nvanbenschoten:
@benesch yeah, I'll check it out tonight. Thanks for putting that together!

nvanbenschoten:
Report back from ARM-land: compilations succeed with this change as-is and with @benesch's patch!

a-robinson (approved):

mberhault:
Should we merge this and separately patch `cockrochdb/rocksdb`, or would you prefer I wait until the patch is in?

benesch:
@mberhault I just pushed my patch to the crl-release-5.9.0 branch, so you should be able to update this PR to point at the new tip of that branch.

mberhault:
Thanks. Updated to tip. I'll wait for a green build.

mberhault:
Build's green. PTAL.

PR #19837, opened 2017-11-06
*: link ccl code using build tags, not entrypoints

benesch:
CCL code has expanded beyond its originally-envisioned home, pkg/ccl.
libroachccl lives in c-deps, and UI CCL code is likely to live in
pkg/ui/ccl. Our current plan to support conditional CCL compilation in
pkg/ui is to introduce a `ccl` build tag that's set iff we're building a
CCL binary.

This commit introduces that `ccl` build tag. In the interest of reducing
complexity, this commit removes the pkg/cmd/cockroach-oss/main.go
entrypoint, which was a copy of the normal entrypoint (main.go) with CCL
imports removed, and replaces it with a main_oss.go file that's
conditionally compiled when the `ccl` build tag is present.

Additionally move tests that check for forbidden imports into the
linter. They're otherwise awkwardly shoved into packages where they
don't belong; e.g., pkg/cli should not be asserting whether the entire
binary is free of forbidden imports.


tschottdorf:
This might be the right thing to do, but what immediately makes me cautious is that using build tags for this purpose often gets in the way of tooling such as gorename, etc, and usually means finding out much later that something has been broken. It also means that it'll be more critical to test both the ccl and non-ccl builds in separation (arguably not more than now).

That said, again, maybe this is the right thing -- but I wouldn't do it with conditional unless we really want to and then preferably only where we have to.

dt:
There was some discussion back when we first started the ccl tree about hooks vs tags that led to the current design. Plumbing tags to tools in our builds and in our tooling was once concern (we used to be pretty bad at it even in our Make driven builds).

Is there an existing discussion about the UI code layout plan? I'm surprised that we need a non-apache2 tree there.

benesch:
> This might be the right thing to do, but what immediately makes me cautious is that using build tags for this purpose often gets in the way of tooling such as gorename

In general, I agree, but in this case we're not hiding any code from gorename and friends because all the code still exists under `pkg` without the build tag.

>  It also means that it'll be more critical to test both the ccl and non-ccl builds in separation (arguably not more than now).

Yeah, I don't think we're any worse off. We already don't do much (any?) testing of the OSS build, except inasmuch as some `FOO_test` packages don't link CCL packages.

> That said, again, maybe this is the right thing -- but I wouldn't do it with conditional unless we really want to and then preferably only where we have to.

@couchand and I just talked over ways to CCL parts of the UI, and it's going to be a pain to use the equivalent of Go's `init()` hooks with Webpack. I.e., we'd much rather compile a CCL version of the UI and a non-CCL version of the UI, rather than having the CCL version inject itself into the non-CCL version at binary runtime. That requires conditional compilation any way you skin it.

benesch:
> There was some discussion back when we first started the ccl tree about hooks vs tags that led to the current design. Plumbing tags to tools in our builds and in our tooling was once concern (we used to be pretty bad at it even in our Make driven builds).

Yeah, I'm definitely not advocating for getting rid of the hooks and sticking every file in `pkg/ccl` behind `+build ccl`. I'm fairly positive that this doesn't change our tooling situation at all; it's just a question of whether we let Make or `go build` handle the "include CCL" decision.

dt:
| a CCL version of the UI and a non-CCL version of the UI

I'm curious about motivation for putting parts of the UI under non-APL2. Is there an enterprise feature that we anticipate is primarily implemented *in the UI*?  My understanding was that features identified as enterprise features, have the core of their implementation distributed under the CCL, and can thus require paid enterprise license for use, but we only really need to be restrictive on the core of the feature, not everything related. Indeed, the APL2 distribution still includes code (like the grammar, parser and help) that is aware of and can interact with enterprise features -- the implementation is just missing in a pure-FLOSS binary. I guess I assumed the web UI would be similar to the cli -- it can refer to and interact with enterprise features, but may degrade if they're not implemented or are disabled due to license terms.

couchand:
@dt we're starting to develop admin UI functionality that's going to be covered under the CCL, not simply a frontend piece of backend CCL features.  Happy to chat about this if you'd like more context.

Makefile
```diff
@@ -169,10 +168,10 @@ COCKROACH := ./cockroach$(SUFFIX)$(shell $(XGO) env GOEXE)
 .DEFAULT_GOAL := all
 all: $(COCKROACH)
 
-buildoss: BUILDTARGET = ./pkg/cmd/cockroach-oss
 buildoss: $(C_LIBS_OSS)
 
 $(COCKROACH) build go-install gotestdashi generate lint lintshort: $(C_LIBS_CCL)
+$(COCKROACH) build go-install gotestdashi generate lint lintshort: TAGS += ccl
```
couchand: Maybe a first pass can be to just add this line, and leave the rest as-is.  Would that be acceptable to folks, or is introducing this at all a concern?

I'm generally a fan of having just one way to do things, so if there's a build tag for CCL, that seems like the way all the CCL code should get built in.  Going the other way is harder: it's not immediately clear how to nicely swap out which `embedded.go` to use from the top-level entry point.

benesch: Yeah, if you want to sidecar this line into your PR, @couchand, I'll just file this as future cleanup work.

bdarnell:
:lgtm:

---

---

*[Makefile, line 171 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-KyJQFMnIiOdHbW5Bxwe:-KyJQFMnIiOdHbW5Bxwf:byscn85) ([raw file](https://github.com/cockroachdb/cockroach/blob/99de5688d113f07e913581c6978510bbb30bfaeb/Makefile#L171)):*
> ```Makefile
> all: $(COCKROACH)
> 
> buildoss: $(C_LIBS_OSS)
> ```

I think it's worth preserving the behavior that `buildoss` produces a different output file than `build`, so that you don't overwrite one licensed file with another. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-:-KyJRBgjicJAnCVXtILi:bnfp4nl)*

benesch:
@bdarnell do you have thoughts on the tags vs entrypoint debate?

---

---

*[Makefile, line 171 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-KyJQFMnIiOdHbW5Bxwe:-KyJUknXSlFT_A0Rl7c2:b-o8glm9) ([raw file](https://github.com/cockroachdb/cockroach/blob/99de5688d113f07e913581c6978510bbb30bfaeb/Makefile#L171)):*

I think it's worth preserving the behavior that `buildoss` produces a different output file than `build`, so that you don't overwrite one licensed file with another. 
</blockquote></details>

Wait, I don't think that's the case now. Both `build` and `buildoss` specify `go build ... -o $(COCKROACH)`, and `COCKROACH` isn't conditional on the build type.

Sounds like a good idea in any case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-:-KyJUzcrLKzNtTT96rpU:b-wvr41d)*

bdarnell:
What's the debate? Just whether this change is a good idea or not? Both this change and the status quo are fine as far as I'm concerned. I'm wary of increasing our use of build tags, but I'll wait to see what's in the works for the UI before passing judgment there. 

---

---

*[Makefile, line 171 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-KyJQFMnIiOdHbW5Bxwe:-KyJViPdUcle_nPU5B7i:b79tp4j) ([raw file](https://github.com/cockroachdb/cockroach/blob/99de5688d113f07e913581c6978510bbb30bfaeb/Makefile#L171)):*

Wait, I don't think that's the case now. Both `build` and `buildoss` specify `go build ... -o $(COCKROACH)`, and `COCKROACH` isn't conditional on the build type.

Sounds like a good idea in any case.
</blockquote></details>

Ah, the `-o` is not obvious. I thought it was producing `cockroach-oss`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-:-KyJW7eP40-PBlAdQX8d:bohko2z)*

benesch:
Yes, that's exactly the debate. @couchand has a sketch of how this would fit into the UI build system if you want to take a look: https://github.com/cockroachdb/cockroach/pull/19850

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19837#-:-KyJlXuZmGmou79V9Z6R:bkuf7w9)*

benesch:
Solved another way by #20051 without a single build tag. partyparrot Closing this now-defunct attempt!

PR #19844, opened 2017-11-06
storage: make ReplicaEvalContext the interface

tbg:
As suggested by @bdarnell, this makes ReplicaEvalContext an interface
(previously `ReplicaI`) and in turn has it implemented by `*Replica` and
`SpanSetReplicaEvalContext` (formerly `ReplicaEvalContext`).

As fallout, some methods that previously returned errors due to assertions will
now Fatal. This was deemed more ergonomic than returning `nil` errors from
various `*Replica` accessors.

Introduce a second implementation, `*SpanSetReplicaEvalContext`, which is the
one we use in race-enabled builds (where we want to assert span accesses).

As a side effect, this also removes an oversight: we were previously passing the
`SpanSet` to the `ReplicaEvalContext` even when that wasn't desired (i.e.
outside of race builds).

### DIFF
--- a/pkg/ccl/storageccl/export.go
+++ b/pkg/ccl/storageccl/export.go
@@ -112,10 +112,7 @@ func evalExport(
 	// threshold. If it's not, the mvcc tombstones could have been deleted and
 	// the resulting RocksDB tombstones compacted, which means we'd miss
 	// deletions in the incremental backup.
-	gcThreshold, err := cArgs.EvalCtx.GetGCThreshold()
-	if err != nil {
-		return storage.EvalResult{}, err
-	}
+	gcThreshold := cArgs.EvalCtx.GetGCThreshold()
 	if args.StartTime != (hlc.Timestamp{}) {
 		if !gcThreshold.Less(args.StartTime) {
 			return storage.EvalResult{}, errors.Errorf("start timestamp %v must be after replica GC threshold %v", args.StartTime, gcThreshold)
--- a/pkg/storage/replica.go
+++ b/pkg/storage/replica.go
@@ -460,7 +460,7 @@ type Replica struct {
 	}
 }
 
-var _ ReplicaI = &Replica{}
+var _ ReplicaEvalContext = &Replica{}
 
 // KeyRange is an interface type for the replicasByKey BTree, to compare
 // Replica and ReplicaPlaceholder.
@@ -2438,7 +2438,7 @@ func (r *Replica) executeAdminBatch(
 
 	case *roachpb.ImportRequest:
 		cArgs := CommandArgs{
-			EvalCtx: ReplicaEvalContext{r, nil},
+			EvalCtx: NewReplicaEvalContext(r, &SpanSet{}),
 			Header:  ba.Header,
 			Args:    args,
 		}
@@ -2519,7 +2519,7 @@ func (r *Replica) executeReadOnlyBatch(
 	// that holding readOnlyCmdMu throughout is important to avoid reads from the
 	// "wrong" key range being served after the range has been split.
 	var result EvalResult
-	rec := ReplicaEvalContext{r, spans}
+	rec := NewReplicaEvalContext(r, spans)
 	readOnly := r.store.Engine().NewReadOnly()
 	defer readOnly.Close()
 	br, result, pErr = evaluateBatch(ctx, storagebase.CmdIDKey(""), readOnly, rec, nil, ba)
@@ -4983,7 +4983,7 @@ func (r *Replica) evaluateTxnWriteBatch(
 		if util.RaceEnabled && spans != nil {
 			batch = makeSpanSetBatch(batch, spans)
 		}
-		rec := ReplicaEvalContext{r, spans}
+		rec := NewReplicaEvalContext(r, spans)
 		br, result, pErr := evaluateBatch(ctx, idKey, batch, rec, &ms, strippedBa)
 		if pErr == nil && ba.Timestamp == br.Timestamp {
 			clonedTxn := ba.Txn.Clone()
@@ -5035,7 +5035,7 @@ func (r *Replica) evaluateTxnWriteBatch(
 	if util.RaceEnabled && spans != nil {
 		batch = makeSpanSetBatch(batch, spans)
 	}
-	rec := ReplicaEvalContext{r, spans}
+	rec := NewReplicaEvalContext(r, spans)
 	br, result, pErr := evaluateBatch(ctx, idKey, batch, rec, &ms, ba)
 	return batch, ms, br, result, pErr
 }
@@ -5492,7 +5492,7 @@ func (r *Replica) maybeGossipNodeLiveness(ctx context.Context, span roachpb.Span
 	ba.Timestamp = r.store.Clock().Now()
 	ba.Add(&roachpb.ScanRequest{Span: span})
 	// Call evaluateBatch instead of Send to avoid command queue reentrance.
-	rec := ReplicaEvalContext{r, nil}
+	rec := NewReplicaEvalContext(r, todoSpanSet)
 	br, result, pErr :=
 		evaluateBatch(ctx, storagebase.CmdIDKey(""), r.store.Engine(), rec, nil, ba)
 	if pErr != nil {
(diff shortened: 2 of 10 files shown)


nvanbenschoten:
Reviewed 10 of 11 files at r1, 3 of 3 files at r2.

---

*[pkg/storage/replica_eval_context_span.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI2kVBMnGTG14See2A:-KyI2kVBMnGTG14See2B:b-lqmtr1) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_eval_context_span.go#L123)):*
> ```Go
> // on Replica.ContainsKey.
> func (rec SpanSetReplicaEvalContext) ContainsKey(key roachpb.Key) bool {
> 	desc := rec.i.Desc()
> ```

Can we just call `rec.Desc`?

---

*[pkg/storage/replica_eval_context_span.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI2yHI4J5duiTnz0oP:-KyI2yHI4J5duiTnz0oQ:b-5xubak) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_eval_context_span.go#L157)):*
> ```Go
> // transaction span GC.
> func (rec SpanSetReplicaEvalContext) GetTxnSpanGCThreshold() hlc.Timestamp {
> 	if rec.ss != nil {
> ```

It would be nice to avoid worrying about `nil` `*SpanSets` here. Perhaps we should create an `allSpans` `*SpanSet` and use that instead of ever using `nil`. Or we could change `SpanSet` to treat a nil receiver in `assertAllowed` as a no-op.

---

*[pkg/storage/replica_state.go, line 163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI3WNQhBT4pFe_DEcZ:-KyI3WNQhBT4pFe_DEc_:b-fib7t4) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_state.go#L163)):*
> ```Go
> }
> 
> // GetPushTxnQueue returns the Replica's pushTxnQueue.
> ```

This doesn't look right.

---

*[pkg/storage/replica_test.go, line 778 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI3irAnyFVGjA1KPH9:-KyI3irAnyFVGjA1KPHA:b-5mhgps) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_test.go#L778)):*
> ```Go
> 		if _, err := evalRequestLease(context.Background(), tc.store.Engine(),
> 			CommandArgs{
> 				EvalCtx: &SpanSetReplicaEvalContext{tc.repl, nil},
> ```

Why aren't we using the constructor here and below?

---

*[pkg/storage/span_set.go, line 103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI3sKUumK-sDFkUIEO:-KyI3sKUumK-sDFkUIEP:bu40eqy) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/span_set.go#L103)):*
> ```Go
> }
> 
> // TODO(tschottdorf): take a context, make sure it's well-annotated.
> ```

And add a comment on this function please.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844)*

tschottdorf:
Updated, PTAL!

---

Review status: 4 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/storage/replica_eval_context_span.go, line 123 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI2kVBMnGTG14See2A:-KyI5hapfrNhtOlFC9Qk:b-48vilg) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_eval_context_span.go#L123)):*

Can we just call `rec.Desc`?
</blockquote></details>

Good call, done.

---

*[pkg/storage/replica_eval_context_span.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI2yHI4J5duiTnz0oP:-KyI5izyO23iil07fCXl:bew5zf5) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_eval_context_span.go#L157)):*

It would be nice to avoid worrying about `nil` `*SpanSets` here. Perhaps we should create an `allSpans` `*SpanSet` and use that instead of ever using `nil`. Or we could change `SpanSet` to treat a nil receiver in `assertAllowed` as a no-op.
</blockquote></details>

Good idea, done. In doing so, I discovered some sloppiness (https://github.com/cockroachdb/cockroach/issues/19851) that I won't be able to fix in this PR.

---

*[pkg/storage/replica_state.go, line 163 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI3WNQhBT4pFe_DEcZ:-KyI4wYw0IqLR3VKdQ6G:b-f1vfyx) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_state.go#L163)):*

This doesn't look right.
</blockquote></details>

Already fixed.

---

*[pkg/storage/replica_test.go, line 778 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI3irAnyFVGjA1KPH9:-KyI55jRRL1S6lsKl0xZ:buwwhcb) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/replica_test.go#L778)):*

Why aren't we using the constructor here and below?
</blockquote></details>

No good reason, done.

---

*[pkg/storage/span_set.go, line 103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyI3sKUumK-sDFkUIEO:-KyI5_9IrJGqcDocZOZ7:b-amf69g) ([raw file](https://github.com/cockroachdb/cockroach/blob/42e71513f110e1fdc22b9a3c42f023093edda69e/pkg/storage/span_set.go#L103)):*

And add a comment on this function please.
</blockquote></details>

Done. I also removed the TODO because the key is good enough for finding out what failed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-:-KyIO_lyrH6tPTCF3bZQ:b-rr168o)*

nvanbenschoten:
:lgtm:

---

Reviewed 1 of 11 files at r1, 6 of 6 files at r3.
Review status: 9 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/storage/replica_eval_context.go, line 50 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyIZ5eLZcvVu0ykKY5p:-KyIZ5eLZcvVu0ykKY5q:b-8gmk4w) ([raw file](https://github.com/cockroachdb/cockroach/blob/36893cbc6684652c5a351af89ad55fbe8b15a46c/pkg/storage/replica_eval_context.go#L50)):*
> ```Go
> func NewReplicaEvalContext(r *Replica, ss *SpanSet) ReplicaEvalContext {
> 	if util.RaceEnabled {
> 		if ss == nil {
> ```

We should probably move this out of the `util.RaceEnabled` block so that any regressions are caught immediately.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-:-KyIZISTo_Cjkp2Fc1d4:bnfp4nl)*

tschottdorf:
TFTR!

---

Review status: 8 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/storage/replica_eval_context.go, line 50 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-KyIZ5eLZcvVu0ykKY5p:-KyI_8RUOhomMvZZwlGE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/36893cbc6684652c5a351af89ad55fbe8b15a46c/pkg/storage/replica_eval_context.go#L50)):*

We should probably move this out of the `util.RaceEnabled` block so that any regressions are caught immediately.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19844#-:-KyI_7Bxs2D7MTDtHMr0:b-814g8k)*

PR #19848, opened 2017-11-06
cherrypick-1.1: improve sentry error reporting

tbg:
Cherry-pick of #19780.

Release note (general change): report the file and line of error causes in anonymized crash reports.

cc @cockroachdb/release

### DIFF
--- a/pkg/util/log/crash_reporting.go
+++ b/pkg/util/log/crash_reporting.go
@@ -181,11 +181,9 @@ func (e *safeError) Error() string {
 	return e.message
 }
 
-// redact returns the redacted string and a boolean indicating whether the
-// supplied item is considered safe. That is, if the returned `bool` is true,
-// then the caller may choose to report the item verbatim (which might have been
-// scrubbed internally), discarding the returned string.
-func redact(r interface{}) (string, bool) {
+// redact returns a redacted version of the supplied item that is safe to use in
+// anonymized reporting.
+func redact(r interface{}) string {
 	handleSafeType := func(v *SafeType) string {
 		return fmt.Sprintf("%+v", v.V)
 	}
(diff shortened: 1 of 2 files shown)


pkg/util/log/crash_reporting.go
```diff
@@ -260,29 +290,22 @@ func reportablesToSafeError(depth int, format string, reportables []interface{})
...
-		msg := redact(reportables[i])
-		typ := fmt.Sprintf("<%T>", reportables[i])
-		if msg == "" {
-			redacted = append(redacted, typ)
-		} else if typ == "<log.SafeType>" || typ == "<*log.SafeType>" {
-			redacted = append(redacted, msg)
-		} else {
-			redacted = append(redacted, typ+": "+msg)
-		}
+		s, _ := redact(reportables[i])
```
bdarnell: This is a weird design, with two call sites each of which only uses one of the return values. It's getting really hard to follow this code. Do we really need to keep iterating on this in patch releases?

tschottdorf: If you don't care about getting better sentry error reporting, then no, we don't. But I think we should because a lot of the reports we're getting in sentry are not actionable, and not finding out until next year sets us back.
Re: the design, I'm open for suggestions.

bdarnell: I haven't been paying much attention to sentry so I'm not sure how much we need this. As for design, I can't really suggest a better one because I'm having a hard time telling what we're even trying to do here. Can we reduce some conditional logic by e.g. *always* including the type no matter what?

Basically, my confidence that this code is redacting everything that needs to be redacted is being reduced by every change we add to it. We need to prioritize simplicity here, especially for patch releases.

tschottdorf:
@bdarnell I cherry-picked #19873 on top of this. See if this is more palatable for 1.1.3.

bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19848#-:-KyMc-8TB1keebLqhswH:bnfp4nl)*

tschottdorf:
TFTR!

PR #19849, opened 2017-11-06
test: Improve leaktest linting

lgo:
Previously the leaktest regex was fairly relaxed, and it would allow

    defer leaktest.AfterTest(...)

This makes the regex stricter, preventing that incorrect use of
leaktest.

There is also a correction to one of the incorrect uses of the leaktests.


benesch (approved):
Ha! Nice.

nit that's definitely not worth going through CI for: I usually indent code in commit messages by four spaces instead of wrapping it in triple backticks. That looks much nicer in the `git log`, but still renders as a code block in the PR description.

lego:
TFT(quick)R.

Amazingly CI failed for other reasons, but good to know.

PR #19854, opened 2017-11-06
config,sql: support zone configs on indices and partitions

benesch:
As described in the partitioning RFC, allow specifying zone configs on
indices and partitions of indices.

To specify an index zone config:

    EXPERIMENTAL ALTER INDEX <index> CONFIGURE ZONE 'yaml'

To specify a table zone config:

    EXPERIMENTAL ALTER TABLE <t> PARTITION <p> CONFIGURE ZONE 'yaml'

Support for setting zone configs via the CLI will be implemented in
another commit.

@danhhz I know this is still an overwhelming number of lines, but I'm hopeful that it's much easier to follow.


danhhz:
:lgtm_strong: this is great stuff!

---

Review status: 0 of 17 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/config/zone.go, line 110 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS9PgrT6cCK06ZSScM:-KyS9PgrT6cCK06ZSScN:be9deq7) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/config/zone.go#L110)):*
> ```Go
> 		partition = un[2].(parser.Name)
> 		parsed.Table.TableNameReference = un[:2]
> 	}
> ```

maybe throw in a comment to mention that anything other cases (`len(un) == 0 || len(un) > 3`) get handled by parsed.Table.Normalize()

---

*[pkg/config/zone_test.go, line 259 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS92L9r2wpxLRNJIYR:-KyS92L9r2wpxLRNJIYS:b-o2yx67) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/config/zone_test.go#L259)):*
> ```Go
> 		{`db.tbl@idx`, 51, ""},
> 		{"db.tbl.too.many.dots", 0, "invalid table name"},
> 		{`db.tbl@primary`, 51, ""},
> ```

what about both an index and partition name? should error right?

---

*[pkg/sql/config.go, line 75 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyRwUQ8S7kQ_GuBaRy9:-KyRwUQ8S7kQ_GuBaRyA:b-8kx4mt) ([raw file](https://github.com/cockroachdb/cockroach/blob/c0cb6b4423c64009f5fd8ca347a01624be7a661f/pkg/sql/config.go#L75)):*
> ```Go
> 		if tableDesc := desc.GetTable(); tableDesc != nil {
> 			// This is a table descriptor. Look up its parent database zone config.
> 			// Don't forward keySuffix, because only tables can have subzones.
> ```

I think this keySuffix comment has rotted. Also, is this a bug? I guess we don't expect databases to have subzones, so I don't see how it would return a wrong answer. I'm not convinced that this is a great idea, but you could pass nil for getSubzone (or perhaps a function that always returns nil

---

*[pkg/sql/config.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyRxY3g5Cg5lbHpzkWe:-KyRxY3g5Cg5lbHpzkWf:b-lzj0w4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c0cb6b4423c64009f5fd8ca347a01624be7a661f/pkg/sql/config.go#L93)):*
> ```Go
> // cached system config. If keySuffix is within a subzone, the subzone's config
> // is returned instead.
> func ZoneConfigHook(
> ```

you wrote a benchmark for this, right? can you include a benchstat in the commit message, please? i want to make sure the closures aren't doing anything funky to performance

---

*[pkg/sql/parser/table_name.go, line 269 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS-3nejFXtsd4h2XON:-KyS-3nejFXtsd4h2XOO:bw257hx) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/table_name.go#L269)):*
> ```Go
> }
> 
> func (n *TableNameWithIndex) String() string { return AsString(n) }
> ```

this should probably be next to the others at the bottom of `pkg/sql/parser/expr.go`

---

*[pkg/sql/parser/table_name_test.go, line 47 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS-_gPqjZOm1PoMhXr:-KyS-_gPqjZOm1PoMhXs:b-m9rnt8) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/table_name_test.go#L47)):*
> ```Go
> 
> 	for _, tc := range testCases {
> 		tn, err := func() (*TableName, error) {
> ```

why the closure? i feel like i've mostly seen that for defer

---

*[pkg/sql/parser/zone.go, line 26 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS0Mxs5JbFrCHaMQIr:-KyS0Mxs5JbFrCHaMQIs:b2k47fu) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/zone.go#L26)):*
> ```Go
> 	NamedZone UnrestrictedName
> 	Database  Name
> 	Table     TableNameWithIndex
> ```

calling this field Table is a bit confusing since it could be either a table or an index (and the subtle way TableNameWithIndex works complicates it further). is TableOrIndex too wordy?

either way, this definitely deserves a comment calling out that TableNameWithIndex is subtle and pointing you to read its godoc

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-:-KyS9rCheXDRuQArZen1:bfxnpaf)*

danhhz:
not so fast dan, reviewable tricked me and i missed a bunch of this PR

---

Review status: 0 of 17 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-:-KySC4DGNUe_4WVgqab4:b7iajvk)*

benesch:
Responded to some comments, but holding off on pushing until you finish your review to avoid revision explosion.

---

Review status: 0 of 17 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[pkg/sql/parser/table_name.go, line 269 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS-3nejFXtsd4h2XON:-KySG2r9BuUsNrPCIIQ-:b-o28baz) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/table_name.go#L269)):*

this should probably be next to the others at the bottom of `pkg/sql/parser/expr.go`
</blockquote></details>

`TableName.String()` and `TableNames.String()` are defined in this file. I _think_ only `FOOExpr` and `FOODatum`s' string methods are supposed to go in expr.go, though that rule definitely isn't absolute.

---

*[pkg/sql/parser/table_name_test.go, line 47 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS-_gPqjZOm1PoMhXr:-KySFdrJnXNEBdXRScpp:bvfui9s) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/table_name_test.go#L47)):*

why the closure? i feel like i've mostly seen that for defer
</blockquote></details>

The error we're matching against might be from `ParseOne`, or it might be from `Normalize()`, or it might be from `QualifyWithDatabase`. Doing it this way avoids doubly-nested `if err == nil` blocks.

---

*[pkg/sql/parser/zone.go, line 26 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS0Mxs5JbFrCHaMQIr:-KySFVZNDzVfwpf7bYVv:b-y7qs73) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/zone.go#L26)):*

calling this field Table is a bit confusing since it could be either a table or an index (and the subtle way TableNameWithIndex works complicates it further). is TableOrIndex too wordy?

either way, this definitely deserves a comment calling out that TableNameWithIndex is subtle and pointing you to read its godoc
</blockquote></details>

Ooh, I like `TableOrIndex`. Will do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-:-KySGSOJUmPxlHO9av34:b23dmv)*

danhhz:
okay, i think i looked at everything now. thanks for waiting on pushing the updates : - )

---

Review status: 0 of 17 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[pkg/sql/zone.go, line 98 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KySKWZF_RUNSjcpOuSa:-KySKWZF_RUNSjcpOuSb:b-uen8u7) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/zone.go#L98)):*
> ```Go
> 		}
> 		if index != nil && pi.ID != index.ID {
> 			return nil, nil, "", fmt.Errorf("partition %q belongs to index %q, not %q",
> ```

test for this?

---

*[pkg/sql/parser/table_name.go, line 269 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS-3nejFXtsd4h2XON:-KySJVBnwHyqaPdutVfQ:b-q8ilnv) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/table_name.go#L269)):*

`TableName.String()` and `TableNames.String()` are defined in this file. I _think_ only `FOOExpr` and `FOODatum`s' string methods are supposed to go in expr.go, though that rule definitely isn't absolute.
</blockquote></details>

gotcha

---

*[pkg/sql/parser/table_name_test.go, line 47 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS-_gPqjZOm1PoMhXr:-KySLfSA379ik4gH5Zui:bhkkvyz) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/table_name_test.go#L47)):*

The error we're matching against might be from `ParseOne`, or it might be from `Normalize()`, or it might be from `QualifyWithDatabase`. Doing it this way avoids doubly-nested `if err == nil` blocks.
</blockquote></details>

aha! clever

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-:-KySLXy6QoZksV0rB2y7:b28unk4)*

benesch:
Whew, ok, everything is addressed!

---

Review status: 0 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/config/zone.go, line 110 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS9PgrT6cCK06ZSScM:-KySSr1iM4B8m-W1oxA1:bhz3jjm) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/config/zone.go#L110)):*

maybe throw in a comment to mention that anything other cases (`len(un) == 0 || len(un) > 3`) get handled by parsed.Table.Normalize()
</blockquote></details>

Good idea. This is subtle. Done.

---

*[pkg/config/zone_test.go, line 259 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS92L9r2wpxLRNJIYR:-KySYTgN3guk4ECcUt-Y:b-9rkjre) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/config/zone_test.go#L259)):*

what about both an index and partition name? should error right?
</blockquote></details>

Good point. Added the test case (and the code to make it pass).

---

*[pkg/sql/config.go, line 75 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyRwUQ8S7kQ_GuBaRy9:-KySaR9ERamGYhBYSABt:bwzraak) ([raw file](https://github.com/cockroachdb/cockroach/blob/c0cb6b4423c64009f5fd8ca347a01624be7a661f/pkg/sql/config.go#L75)):*

I think this keySuffix comment has rotted. Also, is this a bug? I guess we don't expect databases to have subzones, so I don't see how it would return a wrong answer. I'm not convinced that this is a great idea, but you could pass nil for getSubzone (or perhaps a function that always returns nil
</blockquote></details>

Ooh, good catch. Fixed. Not causing incorrectness since both implementations of `getSubzone` will simply return nil when the list of subzones is empty.

---

*[pkg/sql/config.go, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyRxY3g5Cg5lbHpzkWe:-KySf7jg8MAFbvIm3e3X:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c0cb6b4423c64009f5fd8ca347a01624be7a661f/pkg/sql/config.go#L93)):*

you wrote a benchmark for this, right? can you include a benchstat in the commit message, please? i want to make sure the closures aren't doing anything funky to performance
</blockquote></details>

Done.

---

*[pkg/sql/zone.go, line 98 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KySKWZF_RUNSjcpOuSa:-KySZxAZxpgLCeXrgG9V:b-grloba) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/zone.go#L98)):*

test for this?
</blockquote></details>

This is impossible now that zone specifiers can't specify an index and partition simultaneously. Updated the code to reflect that reality.

---

*[pkg/sql/parser/zone.go, line 26 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-KyS0Mxs5JbFrCHaMQIr:-KySTFrjmlmHCuyUs5QS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1d6be77a72267c14f6a100424a4e3466da61ac8c/pkg/sql/parser/zone.go#L26)):*

Ooh, I like `TableOrIndex`. Will do.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854#-:-KySflt1bCqv5dfzIklq:b4z96lf)*

danhhz:
Reviewed 1 of 1 files at r6, 2 of 3 files at r7, 14 of 14 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19854)*

benesch:
Woohoo, green build! TFTR!

Issue #19607, opened 2017-10-27
sqlccl: Estimation of time remaining in job is way off


a-robinson:
I realize estimating how long a job will take is super hard, but the estimate that I've been seeing for a a CSV IMPORT is so far off (by an order of magnitude, potentially) that it seems worth at least filing an aspirational issue for:

<img width="1279" alt="screen shot 2017-10-27 at 5 41 46 pm" src="https://user-images.githubusercontent.com/7085343/32126343-225e8aee-bb3e-11e7-8891-e5a310d8109c.png">

It's been estimating "01:21:40 remaining" for a while now, which means the 33% number is clearly not correct, but also this is a job that apparently takes at least 7 hours according to @dt's comment that his identical IMPORT job broke after more than 7 hours of running (https://github.com/cockroachdb/cockroach/pull/19436#issuecomment-339206066).

To be specific, this is a TPC-H SF=100 lineitem IMPORT on 3 GCE n1-highcpu-4 nodes using local SSD.

a-robinson:
Sorry for the incomplete initial post. Edited.

dt:
AFAIK, the time estimate is extrapolating from % complete and time elapsed... we're just wrong to say 33% after first read of the CSVs (to sample the key distribution) is done.... probably makes more sense to guess that that is... 5 or 10%?

dt:
@benesch how would you feel about an alternative signature for `Progressed` that takes a percentage to add rather than an absolute one, and then handles the retry loop required to read/add/write?

I'm thinking that could simplify distributed progress, letting each chunk write a "i finished 1/x" update as it wraps up, rather than needing to plumb those back to a coordinator that writes the sums? I guess the downside is a write storm in widely distributed jobs if everyone finishes at the same time

benesch:
Certainly not opposed, but fair warning: in the days leading up to the 1.1 release, progress tracking could occasionally brick a restore. We had a bug where, if we were unlucky, several dozen concurrent writers would try to update the progress at once. This would overwhelm the command queue, causing additional concurrent writers to pile up; eventually we'd end up with fifty or sixty goroutines all trying to update the progress at once, essentially serializing the restore. I'm not convinced that we were entirely at fault—our performance under contention leaves a lot to be desired—but tread carefully.

dt:
We might be able to get away with this in IMPORT where concurrency is limited by number of import files... I'll play with it, thanks!

mjibson:
Fixed.

PR #19855, opened 2017-11-06
sql: only claim 1/10 done after sampling csv

dt:
1/3 was pulled out of thin air and is wildly inaccurate.
1/10 is too, but hopefully closer, and at least not overpromising quite as badly.

### DIFF
--- a/pkg/sql/distsql_physical_planner.go
+++ b/pkg/sql/distsql_physical_planner.go
@@ -1090,7 +1090,7 @@ func (l *DistLoader) LoadCSV(
 	}
 
 	log.VEventf(ctx, 1, "generated %d splits; begin routing for job %s", len(spans), job.Record.Description)
-	if err := job.Progressed(ctx, 1.0/3.0, jobs.Noop); err != nil {
+	if err := job.Progressed(ctx, 1.0/10.0, jobs.Noop); err != nil {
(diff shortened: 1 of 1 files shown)


RaduBerinde:
LGTM

I'm surprised only one place uses this value, don't the rest of the progress updates have to interpolate from 1/10 to 1? Or is the only other progress update at the very end?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19855#-:-KyImyvH9pNCieOjNXIF:b9sjv4d)*

dt:
bingo (sadly).

dt:
@RaduBerinde more details about the current progress situation and some thoughts on how we might add more updates on https://github.com/cockroachdb/cockroach/issues/19607, but TLDR is we'll need to switch to updating the job progress remotely, from within the individual distsql processors, or add some flow of that progress info back to the coordinator. IIRC @mjibson is thinking about ways to do the former, but in the mean time, this is just to make our one existing progress update a little less wrong.

madelynnblue (approved):

mjibson:
I've already sketched out how to do it with distributed updating and started working on the code. Just a low priority.

a-robinson (commented):
👍

PR #19856, opened 2017-11-06
sql: introduce utils for Session communication

andreimatei:
First parts of a larger Session rewrite. This commit contains i/o utils
for a session to receive statements that it needs to run and to produce
results that another module needs to deliver to the client.

session_refactor.go contains a diagram hinting to how the various pieces
will fit together.

### DIFF
--- a/pkg/sql/conn_executor.go
+++ b/pkg/sql/conn_executor.go
@@ -1,0 +1,112 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package sql
+
+// A ConnExecutor is in charge of executing queries received on a given client
+// connection. The ConnExecutor implements a state machine (mostly defined by
+// the Postgres/pgwire session semantics). The state machine is supposed to run
+// asynchronously wrt the client connection: it receives input statements
+// through a stmtBuf and produces results through a clientComm interface. The
+// ConnExecutor maintains a cursor over the statementBuffer and executes
+// statements / produces results for one statement at a time. The cursor points
+// at all times to the statement that the ConnExecutor is currently executing.
+// Results for statements before the cursor have already been produced (but not
+// necessarily delivered to the client). Statements after the cursor are queued
+// for future execution. Keeping already executed statements in the buffer is
+// useful in case of automatic retries (in which case statements from the
+// retried transaction have to be executed again); the ConnExecutor is in charge
+// of removing old statements that are no longer needed for retries from the
+// (head of the) buffer. Separately, the implementer of the clientComm interface
+// (e.g. the pgwire module) is in charge of keeping track of what results have
+// been delivered to the client and what results haven't (yet).
+//
+// The ConnExecutor has two main responsibilities: to dispatch queries to the
+// execution engine(s) and relay their results to the clientComm, and to
+// implement the state machine maintaining the various aspects of a connection's
+// state. The state machine implementation is further divided into two aspects:
+// maintaining the transaction status of the connection (outside of a txn,
+// inside a txn, in an aborted txn, in a txn awaiting client restart, etc.) and
+// maintaining the cursor position (i.e. correctly jumping to whatever the
+// "next" statement to execute is in various situations).
+//
+// The cursor normally advances one statement at a time, but it can also skip
+// some statements (remaining statements in a query string are skipped once an
+// error is encountered) and it can sometimes be rewound when performing
+// automatic retries. Rewinding can only be done if results for the rewound
+// statements have not actually been delivered to the client; see below.
+//
+//                                                  +-----------------+
+//     +--------------------+                       |ConnExecutor     |
+//     |stmtBuf             |                       |                 |
+//     |                    | statements are read   |                 |
+//     | +-+-+ +-+-+ +-+-+  +-----------------------+  +------------+ |
+//     | | | | | | | | | |  |                       |  |txnState    | |
+// +---> +-+-+ +++-+ +-+-+  |                       |  +------------+ |
+// |   |        ^           |                       |  +------------+ |
+// |   |        |   +-------------------------------+  |session data| |
+// |   |        +   v       | cursor is advanced    |  +------------+ |
+// |   |       cursor       |           +           |                 |
+// |   +--------------------+           |           +------------+----+
+// |                                    |                        |
+// |                                    |                        |
+// +-------------+                      |                        |
+//               +--------+             |                        |
+//               ||parser |             |results are produced    |
+//               +--------+             |                        |
+//               |                      |                        |
+//               |                      |                        v
+//               |                      |            +-----------+----+
+//       +-------+------+               |            |execution engine|
+//       | pgwire conn  |               |            |(local/DistSQL) |
+//       |              |               |            +----------------+
+//       |   +----------+               |
+//       |   |clientComm<---------------+
+//       |   +----------+
+//       |              |
+//       +-------^------+
+//               |
+//               |
+//       +-------+------+
+//       | SQL client   |
+//       +--------------+
+//
+// The ConnExecutor is disconnected from client communication (i.e. generally
+// network communication - pgwire); the module doing client communication is
+// responsible for pushing statements into the buffer and for providing an
+// implementation of the clientConn interface (and thus consume results and send
+// them to the client). The ConnExecutor does not control when results are
+// delivered to the client, but still it does have some influence over that;
+// this is because of the fact that the possibility of doing automatic retries
+// is broken the moment results for the transaction in question are delivered to
+// the client. The communication module has full freedom in sending results
+// whenever it sees fit; however the ConnExecutor influences communication in
+// the following ways:
+//
+// a) The ConnExecutor calls clientComm.flush(), informing the implementer that
+// all the previous results can be delivered to the client at will.
+//
+// b) When deciding whether an automatic retry can be performed for a
+// transaction, the ConnExecutor needs to:
+//
+//   1) query the communication status to check that no results for the txn have
+//   been delivered to the client and, if this check passes:
+//   2) lock the communication so that no further results are delivered to the
+//   client, and, eventually:
+//   3) rewind the clientComm to a certain position corresponding to the start
+//   of the transaction, thereby discarding all the results that had been
+//   accumulated for the previous attempt to run the transaction in question.
+//
+// These steps are all orchestrated through clientComm.lockCommunication() and
+// rewindCapability{}.
--- a/pkg/sql/conn_io.go
+++ b/pkg/sql/conn_io.go
@@ -1,0 +1,380 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package sql
+
+import (
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+
+	"github.com/cockroachdb/cockroach/pkg/sql/sem/tree"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
+)
+
+// This file contains utils and interfaces used by a ConnExecutor to communicate
+// with a SQL client. There's stmtBuf used for input and clientComm used for
+// output.
+
+// queryStrPos represents the index of a query string relative to the start of a
+// ConnExecutor. The first query string received on a connection has
+// queryStrPos = 0.
+type queryStrPos int64
+
+// stmtIdx represents the index of a statement within the query string that it's
+// part of.
+type stmtIdx int
+
+// cursorPosition represents a position of a cursor iterating over statements in
+// a stmtBuf.
+type cursorPosition struct {
+	// queryStrPos indicates the current query string (a.k.a. batch of statements
+	// received from the client as a unit). The position is relative to the start
+	// of the connection, even though the stmtBuf may not contain query strings
+	// from the beginning of the connection (old query strings are cleared).
+	queryStrPos queryStrPos
+	// stmtIdx indicates the index of a query within a query string.
+	stmtIdx stmtIdx
+}
+
+func (c cursorPosition) String() string {
+	return fmt.Sprintf("%d:%d", c.queryStrPos, c.stmtIdx)
+}
+
+// compare return -1 if c < other, 0 if c == other, 1 if c > other.
+func (c cursorPosition) compare(other cursorPosition) int {
+	if c.queryStrPos < other.queryStrPos ||
+		(c.queryStrPos == other.queryStrPos && c.stmtIdx < other.stmtIdx) {
+		return -1
+	}
+	if c.queryStrPos == other.queryStrPos && c.stmtIdx == other.stmtIdx {
+		return 0
+	}
+	return 1
+}
+
+// stmtBuf represents a list of query strings (batches of SQL statements sent as
+// a unit by a client) that a SQL client has sent for execution. The buffer
+// is supposed to be used by one reader and one writer. The writer adds
+// statements to the buffer using push(). The reader reads one statement at a
+// time using curStmt(). The buffer internally maintains a cursor representing
+// the reader's position. The reader has to manually move the cursor using
+// advanceOne(), seekToNextQueryStr() and rewind().
+//
+// In practice, the writer is a module responsible for communicating with a SQL
+// client (i.e.  pgwire) and the reader is a ConnExecutor.
+//
+// push() can be called concurrently with curStmt().
+//
+// The ConnExecutor will use the buffer to maintain a window around the
+// statement it is currently executing. It will maintain enough history for
+// executing statements again in case of an automatic retry. The ConnExecutor is
+// in charge of trimming completed statements from the buffer when it's done
+// with them.
+type stmtBuf struct {
+	mu struct {
+		syncutil.Mutex
+
+		// closed, if set, means that the writer has closed the buffer. See Close().
+		closed bool
+
+		// cond is signaled when new statements are pushed.
+		cond *sync.Cond
+
+		queryStrings []tree.StatementList
+		// startPos indicates the index of queryStrings[0] relative to the start of
+		// the connection.
+		startPos queryStrPos
+		// curPos is the current position of the cursor going through the statements.
+		// At any time, curPos indicates the position of the statement to be returned
+		// by curStmt().
+		curPos cursorPosition
+	}
+}
+
+// newStmtBuf creates a stmtBuf.
+func newStmtBuf() *stmtBuf {
+	var buf stmtBuf
+	buf.mu.cond = sync.NewCond(&buf.mu.Mutex)
+	return &buf
+}
+
+// Close marks the buffer as closed. Once Close() is called, no further push()es
+// are allowed. If a reader is blocked on a curStmt() call, it is unblocked with
+// io.EOF. Any further curStmt() call also returns io.EOF (even if some
+// statements were already available in the buffer before the Close()).
+func (buf *stmtBuf) Close() {
+	buf.mu.Lock()
+	buf.mu.closed = true
+	buf.mu.cond.Signal()
+	buf.mu.Unlock()
+}
+
+// push adds a query string to the end of the buffer. If a curStmt() call was
+// blocked waiting for this query string to arrive, it will be woken up.
+func (buf *stmtBuf) push(ctx context.Context, stmts tree.StatementList) {
+	buf.mu.Lock()
+	defer buf.mu.Unlock()
+	if buf.mu.closed {
+		log.Fatal(ctx, "cannot push after Close()")
+	}
+	buf.mu.queryStrings = append(buf.mu.queryStrings, stmts)
+	buf.mu.cond.Signal()
+}
+
+// curStmt returns the statement currently indicated by the cursor. Besides
+// the statement itself, the statement's position is also returned; the position
+// can be used to later rewind() to this statement.
+// If the cursor is positioned at the beginning of the (next) query string and
+// that query string hasn't arrived yet, this blocks until the query string is
+// pushed to the buffer.
+//
+// If the buffer has previously been Close()d, or is closed while this is
+// blocked, io.EOF is returned.
+func (buf *stmtBuf) curStmt(ctx context.Context) (tree.Statement, cursorPosition, error) {
+	buf.mu.Lock()
+	defer buf.mu.Unlock()
+	for {
+		if buf.mu.closed {
+			return nil, cursorPosition{}, io.EOF
+		}
+		curPos := buf.mu.curPos
+		qsIdx, err := buf.translateQueryStrPosLocked(curPos.queryStrPos)
+		if err != nil {
+			return nil, cursorPosition{}, err
+		}
+		if qsIdx < len(buf.mu.queryStrings) {
+			qs := buf.mu.queryStrings[qsIdx]
+			if int(curPos.stmtIdx) >= len(qs) {
+				log.Fatalf(ctx, "corrupt cursor: %s", curPos)
+			}
+			return buf.mu.queryStrings[qsIdx][curPos.stmtIdx], curPos, nil
+		}
+		if (qsIdx != len(buf.mu.queryStrings)) || (curPos.stmtIdx != 0) {
+			return nil, cursorPosition{}, errors.Errorf(
+				"can only wait for next query string; corrupt cursor: %s", curPos)
+		}
+		// Wait for the next statement to arrive to the buffer.
+		buf.mu.cond.Wait()
+	}
+}
+
+// translateQueryStrPosLocked translates an absolute position of a query string
+// (counting from the connection start) to the index of the respective query
+// string among the query strings currently buffered in buf (so, it returns an
+// index relative to the start of the buffer).
+//
+// Attempting to translate a position that's below buf.startPos returns an
+// error.
+func (buf *stmtBuf) translateQueryStrPosLocked(pos queryStrPos) (int, error) {
+	if pos < buf.mu.startPos {
+		return 0, errors.Errorf(
+			"position %d no longer in buffer (buffer starting at %d)",
+			pos, buf.mu.startPos)
+	}
+	return int(pos - buf.mu.startPos), nil
+}
+
+// ltrim iterates over the buffer forward and removes all query strings up to
+// (not including) the query string indicated by pos.
+// Nothing is done for statements in pos' query string (even if pos is higher
+// than some statements).
+//
+// It's illegal to ltrim to a position higher than the current cursor.
+func (buf *stmtBuf) ltrim(ctx context.Context, pos cursorPosition) {
+	buf.mu.Lock()
+	defer buf.mu.Unlock()
+	if pos.queryStrPos < buf.mu.startPos {
+		log.Fatalf(ctx, "invalid ltrim query string position: %s. buf starting at: %d",
+			pos, buf.mu.startPos)
+	}
+	if buf.mu.curPos.compare(pos) < 0 {
+		log.Fatalf(ctx, "invalid ltrim position: %s when cursor is: %s",
+			pos, buf.mu.curPos)
+	}
+	// Remove query strings one by one.
+	for {
+		if (len(buf.mu.queryStrings) == 0) || (buf.mu.startPos == pos.queryStrPos) {
+			break
+		}
+		buf.mu.queryStrings = buf.mu.queryStrings[1:]
+		buf.mu.startPos++
+	}
+}
+
+// advanceOne advances the cursor one statement over. The statement over which
+// the cursor will be positioned when this returns may not be in the
+// buffer yet.
+func (buf *stmtBuf) advanceOne(ctx context.Context) {
+	buf.mu.Lock()
+	defer buf.mu.Unlock()
+	curBatchIdx, err := buf.translateQueryStrPosLocked(buf.mu.curPos.queryStrPos)
+	if err != nil {
+		log.Fatal(ctx, err)
+	}
+	if int(buf.mu.curPos.stmtIdx) == len(buf.mu.queryStrings[curBatchIdx])-1 {
+		// We were positioned on the last statement in a query string. Time to
+		// advance to the next query string.
+		buf.mu.curPos.queryStrPos++
+		buf.mu.curPos.stmtIdx = 0
+	} else {
+		buf.mu.curPos.stmtIdx++
+	}
+}
+
+// seekToNextQueryStr moves the cursor position to the start of the next query
+// string, skipping past any statements left in the current query string (if
+// any).
+// The statement over which the cursor will be positioned when this returns may
+// not be in the buffer yet.
+func (buf *stmtBuf) seekToNextQueryStr() {
+	buf.mu.Lock()
+	buf.mu.curPos.stmtIdx = 0
+	buf.mu.curPos.queryStrPos++
+	buf.mu.Unlock()
+}
+
+// rewind resets the buffer's position to pos.
+func (buf *stmtBuf) rewind(ctx context.Context, pos cursorPosition) {
+	buf.mu.Lock()
+	defer buf.mu.Unlock()
+	if pos.queryStrPos < buf.mu.startPos {
+		log.Fatalf(ctx, "attempting to rewind below buffer start")
+	}
+	buf.mu.curPos = pos
+}
+
+// clientComm is the interface used by the ConnExecutor for creating results to
+// be communicated to client and for exerting some control over this
+// communication.
+//
+// clientComm is implemented by the pgwire connection.
+type clientComm interface {
+	// createStatementResult creates a StatementResult for stmt. pos is the stmt's
+	// position within the connection and is used to enforce that results are
+	// created in order and also to discard results through clientLock.rtrim(pos).
+	createStatementResult(stmt tree.Statement, pos cursorPosition) StatementResult
+
+	// lockCommunication ensures that no further results are delivered to the
+	// client. The returned clientLock can be queried to see what results have
+	// been already delivered to the client and to discard results that haven't
+	// been delivered.
+	//
+	// clientLock.Close() needs to be called on the returned lock once
+	// communication can be unlocked (i.e. results can be delivered to the client
+	// again).
+	lockCommunication() clientLock
+
+	// flush tells the implementation that all the results produced so far can be
+	// delivered to the SQL client. In other words, this is promising that
+	// the corresponding stmtBuf will not be rewind()ed to positions <= the
+	// last pos passed to createStatementResult().
+	//
+	// An error is returned if the client connection is broken. In this case, no
+	// further calls can be made on this clientComm. The caller should interrupt
+	// any statements running on the respective connection.
+	flush() error
+}
+
+// WIP(andrei)
+var _ clientComm
+
+// clientLock is an interface returned by clientComm.lockCommunication(). It
+// represents a lock on the delivery of results to a SQL client. While such a
+// lock is used, no more results are delivered. The lock itself can be used to
+// query what results have already been delivered and to discard results that
+// haven't been delivered.
+type clientLock interface {
+	// Close unlocks the clientComm from whence this clientLock came from. After
+	// Close is called, buffered results may again be sent to the client,
+	// according to the result streaming policy.
+	//
+	// Once Close() is called, the clientLock cannot be used anymore.
+	Close()
+
+	// clientPos returns the position of the latest statement for which results
+	// have been sent to the client. The position is relative to the start of the
+	// connection.
+	clientPos() cursorPosition
+
+	// rtrim iterates backwards through the results and drops all results with
+	// position >= pos.
+	// It is illegal to call rtrim with a position <= clientPos(). In other words,
+	// results can
+	rtrim(pos cursorPosition)
+}
+
+// rewindCapability is used pass rewinding instructions in between different
+// layers of the ConnExecutor state machine. It ties together a position to
+// which we want to rewind within the stream of statements with:
+// a) a clientLock that guarantees that the rewind to the respective position is
+// (and remains) possible.
+// b) the stmtBuf that needs to be rewound at the same time as the results.
+//
+// rewindAndUnlock() needs to be called eventually in order to actually perform
+// the rewinding and unlock the respective clientComm.
+type rewindCapability struct {
+	cl  clientLock
+	buf *stmtBuf
+
+	rewindPos cursorPosition
+}
+
+// WIP(andrei)
+var _ = func() {
+	var rc rewindCapability
+	// use the rewindAndUnlock member function
+	var _ = rc.rewindAndUnlock
+}
+var _ = rewindCapability{}.cl
+var _ = rewindCapability{}.buf
+var _ = rewindCapability{}.rewindPos
+
+// rewindAndUnlock performs the rewinding described by the rewindCapability and
+// unlocks the respective clientComm.
+func (rc *rewindCapability) rewindAndUnlock(ctx context.Context) {
+	rc.cl.rtrim(rc.rewindPos)
+	rc.buf.rewind(ctx, rc.rewindPos)
+	rc.cl.Close()
+}
+
+// WIP(andrei):
+//
+// // getRewindTxnCapability locks the clientComm if it's possible to rewind to the
+// // point where the transaction started. If that is possible, a rewindCapability
+// // is returned (and thus the clientComm is locked) and the returned bool is
+// // true. In this case, rewindAndUnlock() needs to be called on the returned
+// // capability.
+// //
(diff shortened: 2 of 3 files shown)


knz:
I appreciate that you are avoiding the RFC process by what amounts to a RFC in a comment hidden in a source file. It's all right, but I'd still like to invite Nathan and also @bdarnell in this discussion.

There are a few large points here that I'd like to discuss before you're much further in the process:

1. what are the guarantees attached with the current "cursor" in the incoming batches? This is not sufficiently clear yet (to me) -- is this a cursor of the statements already completed but for which the results are still buffered (and not sent to the client)? Or is this a cursor of the statements currently in-flight? (for which we don't have results yet)

We may need two cursors here btw. This architecture strongly reminds me of parallel execution inside a processor. You're proposing what amounts to an issue window, and our parallel executor amounts to out-of-order issue. Of course we are constrained by SQL semantics to out-of-order completion, that's why we need data structures to organize the reordering of results. 

I would be grateful if your proposal/comment could be extended to explain what you want to happen with parallel execution.

2. have you considered the asymmetric nature of TCP: a client can stuff their outgoing TCP stream with SQL statements, close the sending stream (causing the server to get a short read on input), but keep their incoming TCP stream to retrieve the results of all the statements they have sent. Can you confirm this will be handled gracefully here? 

I expect the mechanism to mark the end-of-stream with a special entry in the incoming statement list, and still try to execute whatever is left there *until either completion of everything, or until the outgoing connection (client's incoming) is also closed*. (Note that I do not know, but expect that Go's `net.Conn` reports a conn is closed only when both input and output TCP streams are closed; as long as either is still open, progress can be made, and thus it's possible to have a short read on receive while still being able to send. Perhaps Ben can comment).

3. What of statement cancellation? It seems to me that statement cancellation should cause any buffered statements to be cancelled as well (semantically, from the client's perspective any statement that the client has sent asynchronously for pipelining was sent with the expectation that the behavior of previous statements is "controlled" -- a cancellation would violate that expectation, and thus violate whatever semantic guarantees the client was relying on to pipeline). Can you outline in your proposal/comment how cancellation would be handled with this?



Also I'll hold off reviewing the code until you've resolved my comment about your use of the word "session" as discussed below.

---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/session_refactor.go, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyL7lvYGB6U8EzSeHU1:-KyL7lvYGB6U8EzSeHU2:bcdyd0d) ([raw file](https://github.com/cockroachdb/cockroach/blob/9934c8d76c6763cd8b18a16ded9cf2bd524a8736/pkg/sql/session_refactor.go#L17)):*
> ```Go
> package sql
> 
> // A SQL Session is in charge of executing queries received on a given client
> ```

Terminology: the word "session" has meaning both in our current code base and in other SQL products, and it certainly never carries txn management in its scope. You can't just reuse it for txn management and hope for the best. Not only people will be confused; it will prevent us from doing more stuff / different stuff with sessions down the line.

Please use another word for this functionality. 

What's wrong with the word "executor" btw? "A SQL executor is in charge of executing queries..." sounds much better to me.

Otherwise "transaction manager" is the other common word that other products use.
Some more alternatives: any combination of "transaction"/"state"/"flow"/"SQL" followed by "translator"/"manager"/"handler"/"executor"/"organizer". Just no "session".

---

*[pkg/sql/session_refactor.go, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyL90R77nHd08WYLmfi:-KyL90R77nHd08WYLmfj:bo9m8ou) ([raw file](https://github.com/cockroachdb/cockroach/blob/9934c8d76c6763cd8b18a16ded9cf2bd524a8736/pkg/sql/session_refactor.go#L83)):*
> ```Go
> // full freedom in sending results whenever it sees fit; however the Session
> // influences communication in the following ways:
> // a) The Session calls clientComm.flushAll(), informing the implementer that
> ```

Please add more empty lines in these ordered lists, and put the 1-2-3 points below on their own lines.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KyL9BC_dibCPbY4Wuqz:b-ghmhgv)*

andreimatei:
> what are the guarantees attached with the current "cursor" in the incoming batches? This is not sufficiently clear yet (to me) -- is this a cursor of the statements already completed but for which the results are still buffered (and not sent to the client)? Or is this a cursor of the statements currently in-flight? (for which we don't have results yet)

Added:
```
The
// cursor points at all times to the statement that the Session is currently
// executing. Results for statements before the cursor have already been
// produced (but not necessarily delivered to the client). Statements above the
// cursor are queued for future execution. Keeping already executed statements
// in the buffer is useful in case of automatic retries (in which case
// statements from the retried transaction have to be executed again); the
// Session is in charge of trimming old statements that are no longer needed for
// retries.
```

You say something about needing "two cursors". I'm not sure what you're thinking of, but note that there is also implicitly a cursor hidden inside the `clientComm` interface. The implementer of that interface needs to keep track of what results have been delivered to the client and which weren't (yet).

> have you considered the asymmetric nature of TCP: a client can stuff their outgoing TCP stream with SQL statements, close the sending stream (causing the server to get a short read on input), but keep their incoming TCP stream to retrieve the results of all the statements they have sent. Can you confirm this will be handled gracefully here?

I think this issue is not pertinent to the code in this patch; management of the client connection is the responsibility of another module. The Session will keep executing statements until its Context is cancelled (this part is not featured here yet either). The documentation will have something to say about this once I actually write that code and see how it looks.
Having said this, I have not considered handling of the client's FIN. I'll probably do whatever go lets me do and not ask too many questions. When a read says `io.EOF`, I'll probably cancel the session context and hope for swift interruption of everything (and stop taking new statements out of the buffer). I don't know how this translates to this asymmetric close of the TCP conn, but I don't think any client would do what you describe - close one stream and still wait for results. If it were to do that, what would be the way to abruptly close a pg connection and signal to the server that the client bailed? Send a RST?
FWIW, the pgwire protocol has a Terminate msg that a client is supposed to send before closing the connection - so I guess this means that the client is not supposed to just send that FIN.

> What of statement cancellation? It seems to me that statement cancellation should cause any buffered statements to be cancelled as well

Ah, client expectation in the face of errors - my favorite pgwire topic. Stmt cancelation cause an error for a statement, which also toasts the stmt's transaction (if any). 
Since the protocol does not contain any kind of resets / handshakes after an error, I believe what you say doesn't add up. There's no way for the server to know which statement has been "pipelined" before the client learned of the error and which were sent after the client had a chance to react to the error.
So, what will happen is also what currently happens - an error toasts the current transaction; further statements in the error's query string are _skipped_. If we're in a txn, further statements in further query strings and rejected with a "transaction is aborted" error, until we see a `ROLLBACK` (and it follows that a `ROLLBACK` needs to be the first in a query string, otherwise it is skipped).

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/session_refactor.go, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyL7lvYGB6U8EzSeHU1:-KyMyi7iJNNtNFhINPsz:bvqidfm) ([raw file](https://github.com/cockroachdb/cockroach/blob/9934c8d76c6763cd8b18a16ded9cf2bd524a8736/pkg/sql/session_refactor.go#L17)):*
> Terminology: the word "session" has meaning both in our current code base and in other SQL products, and it certainly never carries txn management in its scope.

What does a session mean to you?
I think the way I use it is quite coherent with the current code. The Session owned the txn state, and all the statement execution happened in the context of a session. We have an `Executor` doing the execution but that's just a bag of methods, all of them taking a `Session` arg. The current structure is just OOP failure; all those methods might as well have been on the Session.

---

*[pkg/sql/session_refactor.go, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyL90R77nHd08WYLmfi:-KyN6qMyClZOH4BiOmXK:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9934c8d76c6763cd8b18a16ded9cf2bd524a8736/pkg/sql/session_refactor.go#L83)):*

Please add more empty lines in these ordered lists, and put the 1-2-3 points below on their own lines.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KyN1q3e7DvNlZwurTUj:bmhprcy)*

nvanbenschoten:
Reviewed 1 of 1 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[pkg/sql/session_io.go, line 33 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRTu65CYDRo1_Husw1:-KyRTu65CYDRo1_Husw2:b-rb406r) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L33)):*
> ```Go
> // output.
> 
> // queryStrPos represents the index of a query string relative to the start of a
> ```

It's a little hard to visualize the interaction between these three types on first read. Honestly, a diagram might be helpful here.

---

*[pkg/sql/session_io.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRfuHBopHZDX4f13ef:-KyRfuHBopHZDX4f13eg:b-v4bqbw) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L37)):*
> ```Go
> type queryStrPos int64
> 
> // queryStrIdx represents the index of a query string within a collection of
> ```

This type is only used in one place, and it's internal to this abstraction. Is it really needed? I think keeping it makes the other two more confusing.

---

*[pkg/sql/session_io.go, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRUiEf-a8n-qtK6Tws:-KyRUiEf-a8n-qtK6Twt:bz92ia5) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L75)):*
> The buffer allows one reader and one writer

What does "allows" mean? Is it not thread-safe? Why won't multiple writes work? What if I wanted to have multiple workers read off a single `stmtBuf`?

---

*[pkg/sql/session_io.go, line 88 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRVO4733FgVuLJRw5K:-KyRVO4733FgVuLJRw5L:b-4et7qw) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L88)):*
> ```Go
> // The Session will use the buffer to maintain a window around the statement it
> // is currently executing. It will maintain enough history for executing
> // statements again in case of an automatic retry.  The Session is in charge
> ```

Two spaces.

---

*[pkg/sql/session_io.go, line 91 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRVSY8i5G2k4sgOG4I:-KyRVSY8i5G2k4sgOG4J:b76kagh) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L91)):*
> ```Go
> // trimming completed statements from the buffer when it's done with them.
> type stmtBuf struct {
> 	mu struct {
> ```

I'm not sure it's idiomatic to use this pattern if all fields in a struct require the lock. That said, I don't hate it.

---

*[pkg/sql/session_io.go, line 129 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRVyR2xUnpljnpFxr7:-KyRVyR2xUnpljnpFxr8:b-n9garm) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L129)):*
> ```Go
> // that query string hasn't arrived yet, this blocks until the query string is
> // pushed to the buffer.
> func (buf *stmtBuf) getCurStmt(ctx context.Context) (parser.Statement, cursorPosition) {
> ```

nit: `s/getCurStmt/curStmt/`

---

*[pkg/sql/session_io.go, line 150 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRfgzk-gcQi4_t34OR:-KyRfgzk-gcQi4_t34OS:bf5cqla) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L150)):*
> ```Go
> 		// Wait for the next statement to arrive to the buffer.
> 		buf.mu.cond.Wait()
> ```

nit: extra line

---

*[pkg/sql/session_io.go, line 200 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRfIUmzqb2NcpNXjQd:-KyRfIUmzqb2NcpNXjQe:b-ysq8dt) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L200)):*
> ```Go
> // which the cursor will be positioned when this returns may not be in the
> // buffer yet.
> func (buf *stmtBuf) advanceOneStmt(ctx context.Context) {
> ```

nit: some of these methods have the word `Stmt` in them and some don't. For instance, `advanceOnStmt` vs. `push`. We should make this consistent:
- `advanceOnStmt` and `pushStmt`
- `advance` and `push`

---

*[pkg/sql/session_io.go, line 255 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRgO4oa1y0b1SPe50J:-KyRgO4oa1y0b1SPe50K:bbk7mt1) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L255)):*
> ```Go
> 	// been delivered.
> 	//
> 	// clientLock.Close() needs to be called on the returned lock once results
> ```

punctuation

---

*[pkg/sql/session_io.go, line 280 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRg_Nb1xy9EGNY6V3y:-KyRg_Nb1xy9EGNY6V3z:bpvb36f) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L280)):*
> ```Go
> 	// Close unlocks the clientComm from whence this clientLock came from. After
> 	// Close is called, buffered results may again be sent to the client,
> 	// according to the result streaming policy.
> ```

Make a note that the `clientLock` should not be used after Close is called.

---

*[pkg/sql/session_io.go, line 314 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRgqSqrFZvpAmwVRMD:-KyRgqSqrFZvpAmwVRME:b-6pgyqx) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L314)):*
> ```Go
> var _ rewindCapability
> 
> // rewindAndUnlock performs the rewinding described by the rewindCapability and
> ```

`s/rewindAndUnlock/rewindAndClose/`

---

*[pkg/sql/session_refactor.go, line 26 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRRnqV3km7IVNWmHjN:-KyRRnqW2wiciQrYiNGu:b-jbe5bw) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_refactor.go#L26)):*
> ```Go
> // cursor points at all times to the statement that the Session is currently
> // executing. Results for statements before the cursor have already been
> // produced (but not necessarily delivered to the client). Statements above the
> ```

`s/above/after/`

---

*[pkg/sql/session_refactor.go, line 30 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRS-AAjAp0Wb9axUlA:-KyRS-AAjAp0Wb9axUlB:b-jqvi57) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_refactor.go#L30)):*
> // the
> // Session is in charge of trimming old statements that are no longer needed for
> // retries.

What do you mean by "trimming" here?

---

*[pkg/sql/session_refactor.go, line 45 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRSL2TB4bepScEvS01:-KyRSL2TB4bepScEvS02:bo7ysu6) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_refactor.go#L45)):*
> ```Go
> // some statements and it can sometimes be rewound when performing automatic
> // retries. Rewinding can only be done if results for the rewound statements
> // have not actually been delivered to the client; see below.
> ```

When can it skip statements?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856)*

andreimatei:
TFYR!
I've added another commit with a Close() method for the buf.

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions.

---

*[pkg/sql/session_io.go, line 33 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRTu65CYDRo1_Husw1:-KyRjKpFLOXIW3g7NAnp:b-z5jrwy) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L33)):*

It's a little hard to visualize the interaction between these three types on first read. Honestly, a diagram might be helpful here.
</blockquote></details>

Removed queryStrIdx. Do you still want a diagram?

---

*[pkg/sql/session_io.go, line 37 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRfuHBopHZDX4f13ef:-KyRjFqxS-2mx881vn39:b-nwgmcj) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L37)):*

This type is only used in one place, and it's internal to this abstraction. Is it really needed? I think keeping it makes the other two more confusing.
</blockquote></details>

ok, removed. 

---

*[pkg/sql/session_io.go, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRUiEf-a8n-qtK6Tws:-KyRjY0OVfJcDFDNzvAG:bxxno22) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L75)):*

> The buffer allows one reader and one writer

What does "allows" mean? Is it not thread-safe? Why won't multiple writes work? What if I wanted to have multiple workers read off a single `stmtBuf`?
</blockquote></details>

well, this buffer is tied to a session/connection because it implicitly assigns positions relative to statements relative to the start of a transaction, and the reader needs to be able to reset to the beginning of a transaction.
For readers, you could in principle have multiple of them, but the results need to be produced in order (so, when a statement is read from the buffer, the reader is supposed to immediately call `clientComm.createStatementResult()`).

Technically, a question about allowing a more general would be on the condvar signal vs broadcast use; I'd rather not have to think about it.

So I dunno, I don't think of this as a general buffer. I've replaced "allows" with "is supposed to be used by a single reader and a single writer". Sounds good?

---

*[pkg/sql/session_io.go, line 88 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRVO4733FgVuLJRw5K:-KyRl5ZIJeOT1S-v66fb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L88)):*

Two spaces.
</blockquote></details>

Done.

---

*[pkg/sql/session_io.go, line 129 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRVyR2xUnpljnpFxr7:-KyRlMNXvDtGvHesAtbQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L129)):*

nit: `s/getCurStmt/curStmt/`
</blockquote></details>

Done.

---

*[pkg/sql/session_io.go, line 150 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRfgzk-gcQi4_t34OR:-KyRlObvDyK1JKHDPG_o:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L150)):*

nit: extra line
</blockquote></details>

Done.

---

*[pkg/sql/session_io.go, line 200 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRfIUmzqb2NcpNXjQd:-KyRld8aUWie2u0CS4BT:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L200)):*

nit: some of these methods have the word `Stmt` in them and some don't. For instance, `advanceOnStmt` vs. `push`. We should make this consistent:
- `advanceOnStmt` and `pushStmt`
- `advance` and `push`
</blockquote></details>

Done.

---

*[pkg/sql/session_io.go, line 255 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRgO4oa1y0b1SPe50J:-KyRlx_bncMJt8YD7gOj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L255)):*

punctuation
</blockquote></details>

Done.

---

*[pkg/sql/session_io.go, line 280 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRg_Nb1xy9EGNY6V3y:-KyRm6R2ApeFV6l97byx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L280)):*

Make a note that the `clientLock` should not be used after Close is called.
</blockquote></details>

Done.

---

*[pkg/sql/session_io.go, line 314 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRgqSqrFZvpAmwVRMD:-KyRm9dZOM4J-YitX171:bjblhuy) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_io.go#L314)):*

`s/rewindAndUnlock/rewindAndClose/`
</blockquote></details>

well, "unlock" is more descriptive since it refers to the `clientLock`. I thought about renaming `clientLock.Close` to `unlock`, but `Close` is the Closer interface...

---

*[pkg/sql/session_refactor.go, line 26 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRRnqV3km7IVNWmHjN:-KyRmTBYF1l1ysQDNeeU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_refactor.go#L26)):*

`s/above/after/`
</blockquote></details>

Done.

---

*[pkg/sql/session_refactor.go, line 30 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRS-AAjAp0Wb9axUlA:-KyRmanVSL-POQt6daUa:b-fa9sz8) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_refactor.go#L30)):*

> // the
> // Session is in charge of trimming old statements that are no longer needed for
> // retries.

What do you mean by "trimming" here?
</blockquote></details>

clarified

---

*[pkg/sql/session_refactor.go, line 45 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyRSL2TB4bepScEvS01:-KyRmjVNPpK3XdhVLTfL:b-97gcr3) ([raw file](https://github.com/cockroachdb/cockroach/blob/81e2dfc4a4ec969d05cbec96ffe611a82b1ec16a/pkg/sql/session_refactor.go#L45)):*

When can it skip statements?
</blockquote></details>

explained

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KyRwWRerrK9A0invtsw:b-f0znfy)*

knz:
There is something very concrete, shared by all SQL engines, an object that holds data on the server on behalf of the client, in particular those items that extend outside of the lifetime of transactions:

- prepared statements
- prepared portals
- cached plans
- session variables
- (later) temporary tables

That object is called *everywhere* a **session**, and it has nothing to do with the flow of states for SQL transactions.

Of course in theory you could envision merging the concerns between what you're doing here (which I'll call "transaction manager" for the time being) and the session, and merge them also in the code. But wouldn't you agree that *good software engineering practices* will strongly recommend we disentangle separate concerns into separate data structures?

Then, consider this: suppose that you did not agree, that you do not think that good software engineering invites us from separating concerns. What then? I'd then highlight that since your switching teams, you won't be the person who has to live with and maintain that code. In that quality, I'd then *kindly request* that you do this separation out of concern for the rest of the team.

andreimatei:
Raphael, please acknowledge that currently the `Session` object does mix all these things that you don't want mixed. It contains the 'TxnState' object, with txn-related state, and it also contains the prepared statements and all the `evalContext` variables.

Out of all the faults that our current `Session` has, I don't find this to particularly be one of them. If you want, I can group all the things you have listed in a `SessionData` struct or such, holding all this session-scoped items. If you really want, I can also give a different name to the object that receives statements, executes statements and moves forward the state machine... Although to me `Session` seems like a very natural name for that... Seems natural to say that statements are executed "on a session" and also that the transaction state is part of the current "state of the session".
If you don't like the name `Session` for this, then what's a better one? `TransactionManager` doesn't seem to fit well - we need something that receives statements and executes them, TM does't suggest that to me (I don't know what it suggests).

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KyS8jS7U5eXy4J8w57e:by2fya4)*

knz:
I think you just found the right name then - what about `TxnState`?

andreimatei:
Well, but that `TxnState` is really just a component of the `Session` (and I intend to keep it as a separate struct). Remember, the object we're trying to name receives statements and executes them (and also maintains this `TxnState` and also holds on to the `items you've mentioned before (my suggested `SessionData`).
So I don't really like something with `State` in the name - that suggests a bag of fields grouped together, not a complex objects with goroutines and logic and varied responsibilities. I also don't really like `Txn` in the name haha

Perhaps... `ConnectionExecutor`? Or just `Connection` (it'd clash with `pgwire.Conn` but it'd be in a different package). But... I like `Session`. I didn't win your heart `Session` vs `SessionData`?

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KySZfArPB4HPGTA9NZT:b-rknnas)*

knz:
On 08-11-17 22:45, Andrei Matei wrote:
>  I didn't win your heart |Session| vs |SessionData|?

would you be willing to compromise with `SessionData` vs `SessionExecutor`?


-- 
Raphael 'kena' Poss

andreimatei:
> would you be willing to compromise with `SessionData` vs `SessionExecutor`?

Maybe... Although between `SessionExecutor` and `ConnectionExecutor`, I might prefer the 2nd.
If we go with `ConnExecutor`, I might even give you `Session` for `SessionData` if you want, although I personally wouldn't encourage that :)

andreimatei:
deal!

knz:
> Added:
> 
> |The cursor points at all times to the statement that the Session is
> currently [...] |

Thanks, this clarifies.

> You say something about needing "two cursors". I'm not sure what you're
> thinking of, but note that there is also implicitly a cursor hidden
> inside the |clientComm| interface. The implementer of that interface
> needs to keep track of what results have been delivered to the client
> and which weren't (yet).

It would be nice if this was also stated as a reminder to the reader in
the longer explanatory comment.

>     have you considered the asymmetric nature of TCP [...]
> 
> Having said this, I have not considered handling of the client's FIN.

I'm OK with leaving "invalid protocol short reads" - FIN before a
Terminate message - out of scope here, however there are two constraints
on this:

> I'll probably cancel the session context and
> hope for swift interruption of everything (and stop taking new
> statements out of the buffer).

If you adopt a "hard" approach then be sure that the `net.Conn` is
closed, so that no more results are sent to the client. This would be in
the case of FIN before Terminate. Meanwhile...

> FWIW, the pgwire protocol has a Terminate msg that a client is supposed
> to send before closing the connection - so I guess this means that the
> client is not supposed to just send that FIN.

Yes, and in that case we must be handling this properly: if a FIN does
come, *after* a Terminate message, but *before* the other side of the
TCP connection is closed, then I would argue that the execution logic
must process all remaining statements and still send their results onto
the remaining TCP stream.

How do you propose to handle Terminate by the way? Will there be a
special "EOF" entry in the statement buffer?

>     What of statement cancellation? It seems to me that statement
>     cancellation should cause any buffered statements to be cancelled as
>     well
> 
> Ah, client expectation in the face of errors - my favorite pgwire topic.
> Stmt cancelation cause an error for a statement, which also toasts the
> stmt's transaction (if any).
> [...] 
> So, what will happen is also what currently happens - an error toasts
> the current transaction; further statements in the error's query string
> are /skipped/. 

Wow I'm not sure that's correct. That's because the client will be
waiting for a response! I think they should be processed but an error
message returned to the client for each of them. Can you confirm?

> If we're in a txn, further statements in further query
> strings and rejected with a "transaction is aborted" error, until we see
> a |ROLLBACK| (and it follows that a |ROLLBACK| needs to be the first in
> a query string, otherwise it is skipped).

Yes that I fully agree on.

The topics of "flushing the queue on requested cancellation" should be
also outlined in the explanatory comment.

-- 
Raphael 'kena' Poss

knz:
Also I would be glad if you could outline here (in the review thread, not necessarily in the explanatory comment) how much this work will facilitate throttling of incoming queries, to cap the concurrency load on a node. It seems to me that once we have this mechanism in place, we can extend it with a node-wide semaphore on the reading of the statement buffer. am I correct?

---

Reviewed 1 of 3 files at r3, 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[pkg/sql/session_io.go, line 35 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyVaoIg8p7Zgi5HZTXO:-KyVaoIg8p7Zgi5HZTXP:b-67eg2g) ([raw file](https://github.com/cockroachdb/cockroach/blob/a1f0e1cb03f7bf9fac18819f401e4133e0107ab6/pkg/sql/session_io.go#L35)):*
> ```Go
> 
> // queryStrPos represents the index of a query string relative to the start of a
> // Session. The first query string received on a session has queryStrPos = 0.
> ```

How unlikely that is, I'd like the code to assert we're never wrapping around.

Alternatively, you could also state that the value is reset to zero every time the buffer is empty and there's no result set remaining to be sent.

---

*[pkg/sql/session_io.go, line 86 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyVbJ3iuViT8qv3jTHN:-KyVbJ3iuViT8qv3jTHO:b-g3kuwy) ([raw file](https://github.com/cockroachdb/cockroach/blob/a1f0e1cb03f7bf9fac18819f401e4133e0107ab6/pkg/sql/session_io.go#L86)):*
> ```Go
> // is currently executing. It will maintain enough history for executing
> // statements again in case of an automatic retry. The Session is in charge
> // trimming completed statements from the buffer when it's done with them.
> ```

"is in charge *of* trimming"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KyVblYbifi0I4D35VER:bg94cyi)*

andreimatei:
>
> It would be nice if this was also stated as a reminder to the reader in
> the longer explanatory comment.
>

Done.


> > I'll probably cancel the session context and
> > hope for swift interruption of everything (and stop taking new
> > statements out of the buffer).
>
> If you adopt a "hard" approach then be sure that the `net.Conn` is
> closed, so that no more results are sent to the client. This would be in
> the case of FIN before Terminate. Meanwhile...
>
> > FWIW, the pgwire protocol has a Terminate msg that a client is supposed
> > to send before closing the connection - so I guess this means that the
> > client is not supposed to just send that FIN.
>
> Yes, and in that case we must be handling this properly: if a FIN does
> come, *after* a Terminate message, but *before* the other side of the
> TCP connection is closed, then I would argue that the execution logic
> must process all remaining statements and still send their results onto
> the remaining TCP stream.
>
> How do you propose to handle Terminate by the way? Will there be a
> special "EOF" entry in the statement buffer?



I think the pg protocol did not make your considerations. This is what the
docs
<https://www.postgresql.org/docs/current/static/protocol-flow.html#idm46428663883648>
say:
"""
The normal, graceful termination procedure is that the frontend sends a
Terminate message and immediately closes the connection. On receipt of this
message, the backend closes the connection and terminates.
"""

And later
""'
If either frontend or backend sees an unexpected closure of the connection,
it should clean up and terminate.
"""

I'd interpret this as saying that if you see a Terminate msg, bail out.
That's also what the current code does (so, for example, we currently don't
send errors encountered by the parallel execution queue when we see a
Terminate).

The buffer has a Close() method, which makes it not return any more
statements. The plan is for pgwire to call this at the same time that in
cancel's the connection's ctx. And I was going to do it either when
conn.read() returns io.EOF or when we see a Terminate message.
But if you don't agree, I think we can argue when we get to the pgwire part
of this change.




>
>
> > What of statement cancellation? It seems to me that statement
> > cancellation should cause any buffered statements to be cancelled as
> > well
> >
> > Ah, client expectation in the face of errors - my favorite pgwire topic.
> > Stmt cancelation cause an error for a statement, which also toasts the
> > stmt's transaction (if any).
> > [...]
> > So, what will happen is also what currently happens - an error toasts
> > the current transaction; further statements in the error's query string
> > are /skipped/.
>
> Wow I'm not sure that's correct. That's because the client will be
> waiting for a response! I think they should be processed but an error
> message returned to the client for each of them. Can you confirm?


No, I don't think that's correct. The protocol is clear that for statements
sent in a single query string, a single error is returned. That's also our
current behavior, btw.
I do agree that this inconsistency with the pipelined statements case is
unfortunate.



>
>
> > If we're in a txn, further statements in further query
> > strings and rejected with a "transaction is aborted" error, until we see
> > a |ROLLBACK| (and it follows that a |ROLLBACK| needs to be the first in
> > a query string, otherwise it is skipped).
>
> Yes that I fully agree on.
>

Good :). But note that here the "one error per query string" also applies.
If the transaction is aborted and the server gets a query string with
multiple statements, only one error is generate.


>
> The topics of "flushing the queue on requested cancellation" should be
> also outlined in the explanatory comment.


Which flushing exactly are we talking about? Btw, this comment is not
supposed to be particularly exhaustive at this point. More stuff can come
to it as I write/send you more code. You won't be shy about asking for it
and that's good :).

andreimatei:
>
> Also I would be glad if you could outline here (in the review thread, not
> necessarily in the explanatory comment) how much this work will facilitate
> throttling of incoming queries, to cap the concurrency load on a node. It
> seems to me that once we have this mechanism in place, we can extend it
> with a node-wide semaphore on the reading of the statement buffer. am I
> correct?
>

I think that untangling reading from network connection from execution can
definitely help in having some node-wide semaphore (certainly can't hurt
:P). But I haven't thought about it; I'm definitely trying to approach this
as a feature-neutral change for start.
Another somewhat-related thing that I've avoided so far is the subject of
memory accounting for this buffer.

// queryStrPos represents the index of a query string relative to the
start of a// Session. The first query string received on a session has
queryStrPos = 0.
>
> How unlikely that is, I'd like the code to assert we're never wrapping
> around.
>
It's an int64; there's no wrapping around. Also keep in mind that this is a
per-connection counter, not per node.



> Alternatively, you could also state that the value is reset to zero every
> time the buffer is empty and there's no result set remaining to be sent.
> ------------------------------
>
> I kinda like the idea of counting statements from the beginning of the
session. Makes everything simpler than having to think about resets of the
counter and such, and also this natural counter could be a nice thing to
show in traces and such.




> "is in charge *of* trimming"
>
Done.

andreimatei:
Renamed to ConnExecutor.

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-KyXoS2JCFPYT9s1v76A:b4vxe6h)*

andreimatei:
Rafa can I merge this sucker?

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-Kz_CCy2YlSrX9aPaIjY:b-ei19oa)*

knz:
Yes - LGTM with nits. Please fix the nits and then you're free to merge.

---

Reviewed 6 of 6 files at r5.
Review status: all files reviewed at latest revision, 18 unresolved discussions, some commit checks pending.

---

*[pkg/sql/conn_io.go, line 49 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-Kz_DTYP79SXQoxMQ6rb:-Kz_DTYP79SXQoxMQ6rc:byoc8pp) ([raw file](https://github.com/cockroachdb/cockroach/blob/eba91bf9c360dbbca597a53131e1a4399fc937c1/pkg/sql/conn_io.go#L49)):*
> ```Go
> 	// received from the client as a unit). The position is relative to the start
> 	// of the connection, even though the stmtBuf may not contain query strings
> 	// from the beginning of the connection(old query strings are cleared).
> ```

nit: space between `connection` and `(`

---

*[pkg/sql/conn_io.go, line 339 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-Kz_DwaPmKr8Y-uwoA9h:-Kz_DwaPmKr8Y-uwoA9i:bz367la) ([raw file](https://github.com/cockroachdb/cockroach/blob/eba91bf9c360dbbca597a53131e1a4399fc937c1/pkg/sql/conn_io.go#L339)):*
> ```Go
> 
> // WIP(andrei)
> var _rc rewindCapability
> ```

`var _ rewindCapability` perhaps?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-:-Kz_EM7mtYE_d7p72qNp:b-90wkkt)*

andreimatei:

---

*[pkg/sql/conn_io.go, line 339 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-Kz_DwaPmKr8Y-uwoA9h:-Kz_JhHlvJ0iOwXxgKWh:b-fi9tb7) ([raw file](https://github.com/cockroachdb/cockroach/blob/eba91bf9c360dbbca597a53131e1a4399fc937c1/pkg/sql/conn_io.go#L339)):*

`var _ rewindCapability` perhaps?
</blockquote></details>

well it's like this because I use the variable a few lines down to reference a method. I couldn't figure out another way to access a method on a pointer. Do you know?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856)*

knz:
Review status: all files reviewed at latest revision, 18 unresolved discussions, some commit checks pending.

---

*[pkg/sql/conn_io.go, line 339 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-Kz_DwaPmKr8Y-uwoA9h:-KztNGyAvwgvHpFa6fte:bpn6hu2) ([raw file](https://github.com/cockroachdb/cockroach/blob/eba91bf9c360dbbca597a53131e1a4399fc937c1/pkg/sql/conn_io.go#L339)):*

well it's like this because I use the variable a few lines down to reference a method. I couldn't figure out another way to access a method on a pointer. Do you know?
</blockquote></details>

`var _ = func() { var rc rewindCapability; _ := rc.rewindAndUnlock }`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856)*

andreimatei:
Review status: all files reviewed at latest revision, 18 unresolved discussions, some commit checks pending.

---

*[pkg/sql/conn_io.go, line 86 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-KyVbJ3iuViT8qv3jTHN:-KzyIAAV28e5Mx40cBi0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a1f0e1cb03f7bf9fac18819f401e4133e0107ab6/pkg/sql/conn_io.go#L86)):*

"is in charge *of* trimming"
</blockquote></details>

Done.

---

*[pkg/sql/conn_io.go, line 49 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-Kz_DTYP79SXQoxMQ6rb:-KzyI7B-hWI8LfujeQ9m:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eba91bf9c360dbbca597a53131e1a4399fc937c1/pkg/sql/conn_io.go#L49)):*

nit: space between `connection` and `(`
</blockquote></details>

Done.

---

*[pkg/sql/conn_io.go, line 339 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856#-Kz_DwaPmKr8Y-uwoA9h:-KzyIBCGuoOc2cCoa7AV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eba91bf9c360dbbca597a53131e1a4399fc937c1/pkg/sql/conn_io.go#L339)):*

`var _ = func() { var rc rewindCapability; _ := rc.rewindAndUnlock }`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19856)*

PR #19859, opened 2017-11-07
build/packer: install non-buggy Git on teamcity agents

benesch:
Git v2.7 has a bug where submodule metadata sometimes uses absolute
paths instead of relative paths, which means the affected submodules
cannot be shared with Docker containers. Install the latest Git from a
PPA on TeamCity agents for now.


mberhault:
LGTM.

PR #19860, opened 2017-11-07
storageccl: support S3 endpoints

### LINKED ISSUE #19832: storageccl: support s3-protocol compatible internal clouds in IMPORT/BACKUP/RESTORE
There are packages for running your own storage server that is compatible with the s3 protocol. Presumably, we should be able to support this in IMPORT/BACKUP/RESTORE by passing host overrides down to ExportStorage.

This is originally from a user request on the forum: https://forum.cockroachlabs.com/t/import-csv-does-not-work-with-minio-s3-store/1083/1

madelynnblue:
Add endpoint and region parameters to S3 URLs for use with third
party S3-like stores. A region is required because the aws package
verifies that the region isn't an empty string, even though it may
not be strictly required by the store.

Release note: Add AWS_ENDPOINT and AWS_REGION parameters in S3
URIs to specify the AWS endpoint or region. The endpoint can be any
S3-compatible service.

Fixes #19832


dt:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19860#-:-KyNc7Q6mNaz9825dITr:bnfp4nl)*

sahlex:
I'm running version 1.1.4 and still experience the region-error.
Which version is supposed to support it?
> pq: could not find s3 bucket's region: BadRequest: Bad Request

mjibson:
The host portion of the URL is not the endpoint, it is the bucket name. The path portion of the URL is the path in the bucket. So your URL above is using the S3 bucket `172.17.42.50:9001`, which is not a valid bucket.

CockroachDB 1.1 and 1.0 only support Amazon S3. A 2.0 alpha cluster, however, added new support for other endpoints. Use the docs below to correctly form your URL, but add `AWS_ENDPOINT` and `AWS_REGION` parameters to the URL. We are working on documenting this change now (https://github.com/cockroachdb/docs/issues/2255).

See: https://www.cockroachlabs.com/docs/stable/import.html#import-file-temp-directory-urls

sahlex:
Thanks for coming back so quickly.

According to the above mentioned document the URL format is: 
> `[scheme]://[host]/[path]?[parameters]`

When I use the format you suggested (`s3://import-temp/?AWS_ACCESS_KEY_ID=testtest&AWS_SECRET_ACCESS_KEY=testtest&AWS_REGION=us-east-1&AWS_ENDPOINT=172.17.42.50:9001`) it's giving me:
> `pq: cannot write to s3://import-temp/?AWS_ACCESS_KEY_ID=testtest&AWS_SECRET_ACCESS_KEY=testtest&AWS_REGION=us-east-1&AWS_ENDPOINT=172.17.42.50:9001: failed to put s3 object: RequestError: send request failed
caused by: Put https://import-temp.172.17.42.50:9001/BACKUP-CHECKPOINT: dial tcp: lookup import-temp.172.17.42.50 on 172.17.4.13:53: no such host`

mjibson:
I'm really unclear about something. Are you attempting to use Amazon's S3 service, or a S3-compatible service? If you are using Amazon's S3, then omit the AWS_ENDPOINT and AWS_REGION paramaters. They are only needed if connecting to a S3-compatible service.

sahlex:
It's a [minio](https://www.minio.io/) container offering s3 compatible interface. 
BTW, it's still not working. I think I'm making progress but still I've got problems:
Now it's tellinhg me:
>dosdev.de:9001: failed to put s3 object: RequestError: send request failed
caused by: Put https://import-temp.s3.dev.net:9001/BACKUP-CHECKPOINT: EOF

Where I set s3.dev.net and import-temp.s3.dev.net in /etc/hosts to point to my minio server.

mjibson:
Hmm. Unclear if that's a cockroach problem or not. We haven't tested specifically with minio, so it's possible there's a bug here. But I'm not sure just from the error message. Can you open a new issue with these details so we can be sure to track it?

sahlex:
OK, sure.

sahlex:
#21412

Issue #19843, opened 2017-11-06
*log.safeError: replica.go:1724: on-disk and in-memory state diverged: [Lease.Expiration: nil != &hlc.Timestamp{}...


danhhz:
https://sentry.io/cockroach-labs/cockroachdb/issues/400115795/

```
*log.safeError: replica.go:1724: on-disk and in-memory state diverged: [Lease.Expiration: nil != &hlc.Timestamp{} Lease.DeprecatedStartStasis: nil != &hlc.Timestamp{}]
  File "github.com/cockroachdb/cockroach/pkg/storage/replica_raftstorage.go", line 851, in applySnapshot
  File "github.com/cockroachdb/cockroach/pkg/storage/store.go", line 3227, in func1
  File "github.com/cockroachdb/cockroach/pkg/storage/store.go", line 2970, in withReplicaForRequest
  File "github.com/cockroachdb/cockroach/pkg/storage/store.go", line 3036, in processRaftSnapshotRequest
  File "github.com/cockroachdb/cockroach/pkg/storage/store.go", line 2839, in HandleSnapshot
...
(3 additional frame(s) were not displayed)

replica.go:1724: on-disk and in-memory state diverged: [Lease.Expiration: nil != &hlc.Timestamp{} Lease.DeprecatedStartStasis: nil != &hlc.Timestamp{}]
```

petermattis:
This was for one of our clusters, right? Can you point me to the cluster?

danhhz:
This happened while doing a rolling upgrade of adriatic node 1 to current master. I think it was previously on 1.1.2, diving the logs to make sure

danhhz:
If my grep-ing is right, the last version it started up with before this was `CockroachDB CCL v1.1.1-46-gc726df2 (linux amd64, built 2017/10/26 22:27:17, go1.8.3)`

petermattis:
`v1.1.1-46-gc726df2` is also known as `v1.1.2`. And the new version (which crashed) is `CockroachDB CCL 3b90c52 (linux amd64, built 2017/11/06 17:16:45, go1.9)`.

petermattis:
Looks like another problem with making those fields nullable. Consider this test (where `DeprecatedLease` is the old lease proto with `Expiration` and `DeprecatedStartStasis` marked non-nullable):

```
func TestZeroLeaseMarshalRoundTrip(t *testing.T) {
	b, err := protoutil.Marshal(&DeprecatedLease{})
[...]

PR #19867, opened 2017-11-07
roachpb: tweak Lease.Equal

petermattis:
Allow the Expiration and DeprecatedStartStasis fields to compare equal
if either is nil or the zero timestamp (&hlc.Timestamp{}). This is
necessary because this fields were previously non-nullable which meant
they were encoded on the wire. When they were changed to nullable we can
encounter situations where the same lease was encoded on the wire in two
different ways which causes their in-memory representation upon decoding
to be slightly different.

See #19843


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19867#-:-KyN3qmEi5wLzV3g4Ycg:bnfp4nl)*

tschottdorf:
:lgtm:

Can we ever take this out again? Doesn't seem like it, right?

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19867#-:-KyN6u4UdMhl3jaLjn7t:b-l35afc)*

petermattis:
> Can we ever take this out again? Doesn't seem like it, right?

No, I don't think we can. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19867#-:-KyN7Iaf3gJ6QV12qyxG:bivd2nw)*

Issue #4735, opened 2016-02-29
ui: GiB vs. GB (1024^3 vs. 1000^3)


spencerkimball:
tl;dr Let's allow the *iB designations, but make the *B designations also use 1024^i and also make them the defaults for documentation and user interface output.

Let's not confuse our users the way that hard drive manufacturers confused theirs in the 90s. I'm fine with allowing the [MGTPE]iB designations, but I strongly disagree with making [MBTPE]B mean 1000^i bytes instead of 1024^i bytes. Most people expect 1024^i, so that's what we should use. If you Google the difference, you'll find _many_ explanations. Most people are very confused when they see the "i". I didn't know what it meant, for example. I'm sure others here didn't know. My expectation is that byte sizes are specified in 1024^i increments. Why are we actually agreeing with the idiocy of hard drive manufacturers and adopting their nomenclature?

Other tools I've used which specify byte sizes in 1024^i increments using the standard [MGTPE]B designations:

```
pprof
colossus
leveldb/rocksdb
hdfs
cassandra
java
```

cc @BramGruneir

mrtracy:
Normal metric prefixes (without the 'i') should always denote 1000^x units, according to both IEC and SI.  The usage of normal metric prefixes to denote 1024^x units is the incorrect notation; that's actually the reason that the 'i' prefixes were created in the first place, since the new 'i' prefixes were unambiguous in their meaning.

Note that MacOSX and Ubuntu Linux report file sizes in regular metric prefixes, and they do mean the correct 1000^i. In fact, this is a relatively recent development; wikipedia says that MacOSX had the incorrect behavior (i.e. "Megabyte" = 1024^3) until OSX 1.6, when it was changed.

So, if you want to report units in normal metric prefixes such as "MB", that is fine - but they _must_ mean 1000^x units, or we are perpetuating the problem.

spencerkimball:
The issue is that MiB is not familiar to many users. You shouldn't have to know what a "mebibyte" or "gibibyte" is to view the admin UI. It certainly _is_ a goal for our users to see units in the dashboard  without being confused as to what they mean.

To that end, why don't we compromise and allow [MGTPE] as suffix units, meaning 1024^3, and also use those as the default units in our user interfaces. MB and GB can continue as 10^6 and 10^9 in the small set of use cases where operators might choose to input sizes using those units via the command line.

mrtracy:
Spencer and I discussed this a bit offline.

[...]

PR #19870, opened 2017-11-07
ui: label byte axes with the measure name and unit symbol

### LINKED ISSUE #18949: ui: inconsistent use of byte prefixes
All charts which show byte metrics have a labeling and computation inconsistency.  The axis label is written with SI unit names, but the tooltips use IEC prefixes, and the computation is done in IEC units.

screen shot 2017-10-02 at 2 12 11 pm

These should be made consistent, which is as simple as replacing the axis labels with the IEC unit names ("kibibytes", "mebibytes", etc.).  It's been two decades since these units have been standardized, we should be adopting them.

couchand:
Previously, byte axes were mislabeled with the SI unit names, even though they were actually showing IEC units.  The tooltip correctly shows that the units are IEC.  This change switches the axis labels to use the IEC units.  In the interest of avoiding the kinds of surprise expressed in #4735, this change uses the unit symbol instead of the full name.  To fill out the space along the axis, the measure name is listed first, with the unit symbol in parentheses (e.g., "capacity (GiB)").

Closes #18949.

screen shot 2017-11-07 at 11 43 15 am

### DIFF
--- a/pkg/ui/src/util/format.ts
+++ b/pkg/ui/src/util/format.ts
@@ -24,11 +24,25 @@ export function ComputePrefixExponent(value: number, prefixMultiple: number, pre
   return prefixScale;
 }
 
-export function BytesToUnitValue(bytes: number): UnitValue {
+/**
+ * ComputeByteScale calculates the appropriate scale factor and unit to use to
+ * display a given byte value, without actually converting the value.
+ *
+ * This is used to prescale byte values before passing them to a d3-axis.
+ */
+export function ComputeByteScale(bytes: number): UnitValue {
   const scale = ComputePrefixExponent(bytes, kibi, byteUnits);
   return {
-    value: bytes / Math.pow(kibi, scale),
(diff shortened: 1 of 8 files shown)


couchand:
If this seems like a reasonable strategy, we should probably go ahead and make the rest of the axis labels also show the measure name for consistency.

mrtracy:
I'm on board with this from a design perspective, and the code change is LGTM.

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19870#-:-KySD1WDDub-LQC320eS:bh50f7z)*

a-robinson:
Thanks again for fixing this! I just got annoyed by the difference between the Capacity graphs and the capacity in the summary bar on a v1.1 cluster and was very happy to see it resolved on 2.0 👍

PR #19868, opened 2017-11-07
cherrypick-1.1: sql: fix KV TRACE crash with COUNT(*)

RaduBerinde:
Also adding a test.

Fixes #19846.

This is not really a cherrypick since this is no longer a problem on master. I will open a PR to add just the test.
CC @cockroachdb/release

Issue #19846, opened 2017-11-06
sql: panic during SHOW TRACE for SELECT COUNT(*)


nvanbenschoten:
Seen in #19199.

A user is seeing the following panic whenever they run the query `show trace for select count(*) from public.account`:

```
E171102 18:25:36.157731 17922 sql/session.go:1370  [client=[::1]:43328,user=root,n1] a SQL panic has occurred!
E171102 18:25:36.157781 17922 util/log/crash_reporting.go:117  [n1] a panic has occurred!
panic: decoding unset EncDatum [recovered]
	panic: panic while executing "show trace for select count(*) from public.account": decoding unset EncDatum [recovered]
	panic: panic while executing "show trace for select count(*) from public.account": decoding unset EncDatum [recovered]
	panic: panic while executing "show trace for select count(*) from public.account": decoding unset EncDatum

goroutine 17922 [running]:
github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).Recover(0xc42067e120, 0x7fd9ea980140, 0xc42049f110)
	/go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:183 +0xb2
panic(0x1b44e00, 0xc42028c320)
	/usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/sql/pgwire.(*v3Conn).serve.func1(0xc42065d500, 0x7fd9ea980140, 0xc420e214d0)
	/go/src/github.com/cockroachdb/cockroach/pkg/sql/pgwire/v3.go:423 +0xa1
panic(0x1b44e00, 0xc42028c320)
	/usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/sql.(*Session).maybeRecover(0xc42198a000, 0x1d08e47, 0x9, 0xc42198c048, 0x32)
	/go/src/github.com/cockroachdb/cockroach/pkg/sql/session.go:1381 +0x28d
panic(0x1a890a0, 0xc420347e70)
	/usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/cockroachdb/cockroach/pkg/sql/sqlbase.(*EncDatum).EnsureDecoded(0xc4219a58f8, 0xc4219a59d8, 0x0, 0x0)
	/go/src/github.com/cockroachdb/cockroach/pkg/sql/sqlbase/encoded_datum.go:175 +0x533
[...]

PR #19871, opened 2017-11-07
sql: add test for KV TRACE with COUNT(*)

RaduBerinde:
Adding a test for #19846, from PR #19868.

### DIFF
--- a/pkg/sql/trace_test.go
+++ b/pkg/sql/trace_test.go
@@ -350,3 +350,18 @@ func TestBracketInTracetags(t *testing.T) {
 		t.Fatal("expected message not found in trace")
 	}
 }
+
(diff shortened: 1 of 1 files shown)


andreimatei:
I dunno bro, seems quite rando. I wouldn't add it, but if you want it then LGTM.
Perhaps I'd feel differently about a more targeted test checking that we can resolve keys in some situation or something (I didn't pay much attention to 19846)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19871#-:-KyNeHYRFQSBFynTztRV:by70mxf)*

RaduBerinde:
Feels wrong to have a test for a crash someone actually encountered in 1.1 but not in 1.2+

PR #19872, opened 2017-11-07
ui: remove dependency on PhantomJS

### LINKED ISSUE #19866: ui: completely banish PhantomJS
Even though #14240 got rid of a direct dependency on PhantomJS, we still have a transitive dependency in the form of:

```
ui@1.0.0 /Users/couch/go/src/github.com/cockroachdb/cockroach/pkg/ui
└─┬ favicons-webpack-plugin@0.0.7
  └─┬ favicons@4.8.3
    └─┬ svg2png@3.0.1
      └── phantomjs-prebuilt@2.1.14
```

PhantomJS isn't adding very much value, and it's giving us a lot of grief, so let's get rid of it entirely.

benesch:
PhantomJS is a very heavyweight dependency whose install script often
fails. This is all the more pressing now that we ask all contributors to
generate the UI.

We were only indirectly depending on PhantomJS via the
favicons-webpack-plugin, which used it to convert SVGs to PNGs. This is
silly. We aren't building a web app that we expect users to be saving to
their iPhone or Android home screen, so we're not getting the benefit of
the multiple favicons the plugin was generating. Just bundle one
favicon.ico that covers 90% of cases and avoid the dependency.

Additionally add a lint to make sure we don't reintroduce a dependency
on PhantomJS in the future.

Fixes #19866.


couchand:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/ui/Makefile, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19872#-KyM_qj7-X5bEU5uvs7N:-KyM_qj7-X5bEU5uvs7O:b-5jqlmr) ([raw file](https://github.com/cockroachdb/cockroach/blob/f0c84457a489e3dfe4f3f8e2ae3baf02376f8b9f/pkg/ui/Makefile#L45)):*
> ```Makefile
> 	@# TODO(benesch): Invoke tslint just once when palantir/tslint#2827 is fixed.
> 	$(TSLINT) -c tslint.json *.js
> 	@if $(YARN) list | grep phantomjs; then echo ^ forbidden UI dependency >&2; exit 1; fi
> ```

:1st_place_medal:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19872#-:-KyMb90dtPJnbku5cUSu:bnfp4nl)*

jordanlewis:
I like this aggressive dependency banning.

nvb (approved):

PR #19873, opened 2017-11-07
log: simplify crash report anonymization

tbg:
We had an optimization in place that allowed `panic(err)` to report the
type of `err` instead of reporting a wrapper error. This optimization
incurred too much complexity to be worth it and is removed in this commit.


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19873#-:-KyMbB5PA3g-v16xGsoe:bnfp4nl)*

PR #19875, opened 2017-11-07
build: Improve error logging in teamcity-post-failures

bdarnell:
This failed last night for jepsen tests with a `422 Unprocessable
Entity` error response from github (which should indicate a problem
like a missing field in the input, although the input structure is the
same between jepsen and other modes, which appear to be working).
Include more information to figure out what's going on.

### DIFF
--- a/build/teamcity-post-failures.py
+++ b/build/teamcity-post-failures.py
@@ -85,7 +85,14 @@ def post_issue(issue):
         'https://api.github.com/repos/cockroachdb/cockroach/issues',
         data=json.dumps(issue).encode('utf-8'),
         headers={'Authorization': 'token {0}'.format(os.environ['GITHUB_API_TOKEN'])})
-    opener.open(req).read()
+    try:
+        opener.open(req).read()
+    except urllib.error.HTTPError as e:
+        # Github's error response bodies include useful information, so print them.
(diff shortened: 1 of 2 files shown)


build/teamcity-post-failures.py
```diff
@@ -85,7 +85,14 @@ def post_issue(issue):
         'https://api.github.com/repos/cockroachdb/cockroach/issues',
         data=json.dumps(issue).encode('utf-8'),
         headers={'Authorization': 'token {0}'.format(os.environ['GITHUB_API_TOKEN'])})
-    opener.open(req).read()
+    try:
+        opener.open(req).read()
+    except urllib.error.HTTPError as e:
+        # Github's error response bodies include useful information, so print them.
```
tschottdorf: Hope this doesn't leak our API token.

tbg (approved):

build/teamcity-post-failures.py
bdarnell: No, that's in the header, not in the body. A sample error response looks like this according to github's docs:

```
HTTP/1.1 422 Unprocessable Entity
Content-Length: 149

{
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "Issue",
      "field": "title",
      "code": "missing_field"
    }
  ]
}
```

PR #19877, opened 2017-11-07
sql/parser: prepare extraction of bits into separate packages

knz:


justinj:
:lgtm: w/ a couple comments

---

Reviewed 23 of 24 files at r1, 1 of 2 files at r3.
Review status: 24 of 34 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/csv.go, line 548 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMjbciJ2CwBPQNQO28:-KyMjbciJ2CwBPQNQO29:b-y5aa40) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/ccl/sqlccl/csv.go#L548)):*
> ```Go
> 	}
> 
> 	var trans parser.ExprTransformContext
> ```

I don't feel super strongly but I think it might be helpful for the conventional name for this to have `Ctx` somewhere in it, like `evalCtx` does.

---

*[pkg/sql/parser/expr_transform.go, line 29 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMkMk2kAnwDvNiNKhp:-KyMkMk2kAnwDvNiNKhq:b-lewswu) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/sql/parser/expr_transform.go#L29)):*
> ```Go
> }
> 
> // NormalizeExpr is wrapper around ctx.NormalizeExpr which avoids allocation of
> ```

"is a wrapper", and `EvalContext.NormalizeExpr` maybe?

---

*[pkg/sql/parser/expr_transform.go, line 47 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMkpyxAOPepQCy0oIm:-KyMkpyxAOPepQCy0oIn:b-4ip6bw) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/sql/parser/expr_transform.go#L47)):*
> ```Go
> // AggregateInExpr determines if an Expr contains an aggregate function.
> func (t *ExprTransformContext) AggregateInExpr(expr Expr, searchPath SearchPath) bool {
> 	if expr != nil {
> ```

Is it valid for `expr` to be `nil`? If yes, I would say either: a) `WalkExprConst` should handle that case or b) this check should be commented as being an optimization(?)

---

*[pkg/sql/parser/expr_transform.go, line 58 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMlgR5LBOzGP6dRGn4:-KyMlgR6tCUlro0nkSzf:b-ezhgf) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/sql/parser/expr_transform.go#L58)):*
> ```Go
> }
> 
> // IsAggregate determines if SelectClause contains an aggregate function.
> ```

either `a SelectClause` or `n`, I think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-:-KyMncbMKNfl9fJHMNKr:b2wotfg)*

jordanlewis:
:lgtm:

---

Reviewed 24 of 24 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 8 of 8 files at r4.

---

*[pkg/ccl/sqlccl/csv.go, line 548 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMjbciJ2CwBPQNQO28:-KyMsLqICgex3TDicmEU:bzi8mxu) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/ccl/sqlccl/csv.go#L548)):*

I don't feel super strongly but I think it might be helpful for the conventional name for this to have `Ctx` somewhere in it, like `evalCtx` does.
</blockquote></details>

+1 `trans` isn't a particularly clear variable name. i think `transCtx` or `xformCtx` or something would be better.

---

*[pkg/sql/parser/encode.go, line 62 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMtC0jUtNEeTaUNvz_:-KyMtC0jUtNEeTaUNvza:b-2y1r7g) ([raw file](https://github.com/cockroachdb/cockroach/blob/b070d8906d37fd3157a162fada35fff9e2861cfa/pkg/sql/parser/encode.go#L62)):*
> ```Go
> // if f.bareStrings is true, the output string will not be wrapped in quotes
> // if the strings contains no special characters.
> func encodeSQLStringWithFlags(buf *bytes.Buffer, in string, f FmtFlags) {
> ```

Now this function name doesn't make sense anymore - there are no flags!

Hmm, how about adding an EncodeFlags interface that lives in this package with an IsBare() method that FmtFlags implements? Too janky? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-:-KyMuYTFhIHLZvVyCG3L:bnfp4nl)*

knz:

---

*[pkg/ccl/sqlccl/csv.go, line 548 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMjbciJ2CwBPQNQO28:-KyQWnNWxqvefU2SY6wh:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/ccl/sqlccl/csv.go#L548)):*

+1 `trans` isn't a particularly clear variable name. i think `transCtx` or `xformCtx` or something would be better.
</blockquote></details>

Done.

---

*[pkg/sql/parser/encode.go, line 62 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMtC0jUtNEeTaUNvz_:-KyQYAo9Mf4iqP5UGsLv:b-wm672j) ([raw file](https://github.com/cockroachdb/cockroach/blob/b070d8906d37fd3157a162fada35fff9e2861cfa/pkg/sql/parser/encode.go#L62)):*

Now this function name doesn't make sense anymore - there are no flags!

Hmm, how about adding an EncodeFlags interface that lives in this package with an IsBare() method that FmtFlags implements? Too janky? 
</blockquote></details>

Given how often this function is called, not worth the overhead of an interface allocation (!) and a dynamic dispatch (!!).

However I do agree that it is good to have a new standalone EncodeFlags type, and that's what I implemented. Thanks for the suggestion.

---

*[pkg/sql/parser/expr_transform.go, line 29 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMkMk2kAnwDvNiNKhp:-KyQXNHrK5QWhM8kIw9t:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/sql/parser/expr_transform.go#L29)):*

"is a wrapper", and `EvalContext.NormalizeExpr` maybe?
</blockquote></details>

Done.

---

*[pkg/sql/parser/expr_transform.go, line 47 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMkpyxAOPepQCy0oIm:-KyQX_heZRgYPVebntiP:b-ju200q) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/sql/parser/expr_transform.go#L47)):*

Is it valid for `expr` to be `nil`? If yes, I would say either: a) `WalkExprConst` should handle that case or b) this check should be commented as being an optimization(?)
</blockquote></details>

Yes it is valid for an expr to be nil -- this happens with filter predicates that are always-true.
WalkExprConst can't handle this case because it uses dynamic dispatch and that can't happen on a nil reference.
The check is not an optimization :)

I did pull up the test though to aid readability.

---

*[pkg/sql/parser/expr_transform.go, line 58 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877#-KyMlgR5LBOzGP6dRGn4:-KyQXofGpBWXio72d9ZY:bj3eoea) ([raw file](https://github.com/cockroachdb/cockroach/blob/8e18265fc5223f062174ebfa5862e671e4320c1b/pkg/sql/parser/expr_transform.go#L58)):*

either `a SelectClause` or `n`, I think?
</blockquote></details>

Fixed, thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19877)*

PR #19883, opened 2017-11-07
Makefile: allow `make test TESTS=-`

tbg:
This is useful as it compiles all tests and serves as a quick
sanity check during refactors.


jordanlewis (approved):
Sorry to break your workflow! This is a neat trick - maybe add it to the "useful patterns" section of `make help` :)

jordanlewis:
Replaced with #19886

PR #19884, opened 2017-11-07
sql : add functions to encode JSON paths as binary

m-schneider:
These functions will allow us to generate inverted index keys from a JSON interface.

### DIFF
--- a/pkg/util/encoding/encoding.go
+++ b/pkg/util/encoding/encoding.go
@@ -660,6 +660,21 @@ func EncodeNotNullAscending(b []byte) []byte {
 	return append(b, encodedNotNull)
 }
 
+// EncodeArrayAscending encodes a value used to signify membership of an array for JSON objects.
+func EncodeArrayAscending(b []byte) []byte {
+	return append(b, byte(Array))
+}
+
+// EncodeTrueAscending encodes the boolean value true for use with JSON inverted indexes.
+func EncodeTrueAscending(b []byte) []byte {
+	return append(b, byte(True))
+}
+
+// EncodeFalseAscending encodes the boolean value false for use with JSON inverted indexes.
+func EncodeFalseAscending(b []byte) []byte {
+	return append(b, byte(False))
+}
+
 // EncodeNotNullDescending is the descending equivalent of EncodeNotNullAscending.
 func EncodeNotNullDescending(b []byte) []byte {
 	return append(b, encodedNotNullDesc)
--- a/pkg/util/json/json.go
+++ b/pkg/util/json/json.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/cockroachdb/apd"
 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
+	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 )
 
 type jsonType int
@@ -53,6 +54,10 @@ type JSON interface {
 	Format(buf *bytes.Buffer)
 	// Size returns the size of the JSON document in bytes.
 	Size() uintptr
+
+	// EncodeInvertedIndexKeys takes in a key prefix and returns a slice of inverted index keys,
+	// one per path through the receiver.
+	EncodeInvertedIndexKeys(b []byte) [][]byte
 
 	// FetchValKey implements the `->` operator for strings, returning nil if the
 	// key is not found.
@@ -368,6 +373,51 @@ func ParseJSON(s string) (JSON, error) {
 		return nil, errTrailingCharacters
 	}
 	return MakeJSON(result)
+}
+
+func (j jsonNull) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	return [][]byte{encoding.EncodeNullAscending(b)}
+}
+func (jsonTrue) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	return [][]byte{encoding.EncodeTrueAscending(b)}
+}
+func (jsonFalse) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	return [][]byte{encoding.EncodeFalseAscending(b)}
+}
+func (j jsonString) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	return [][]byte{encoding.EncodeStringAscending(b, string(j))}
+}
+func (j jsonNumber) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	var dec = apd.Decimal(j)
+	return [][]byte{encoding.EncodeDecimalAscending(b, &dec)}
+}
+func (j jsonArray) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	var outKeys [][]byte
+
+	for i := range j {
+		for _, childBytes := range j[i].EncodeInvertedIndexKeys(nil) {
+			encodedKey := bytes.Join([][]byte{b, encoding.EncodeArrayAscending(nil), childBytes}, nil)
+			outKeys = append(outKeys, encodedKey)
+		}
+	}
+
+	return outKeys
+}
+
+func (j jsonObject) EncodeInvertedIndexKeys(b []byte) [][]byte {
+	var outKeys [][]byte
+	for i := range j {
+		for _, childBytes := range j[i].v.EncodeInvertedIndexKeys(nil) {
+			encodedKey := bytes.Join([][]byte{b,
+				encoding.EncodeNotNullAscending(nil),
+				encoding.EncodeStringAscending(nil, string(j[i].k)),
+				childBytes}, nil)
+
+			outKeys = append(outKeys, encodedKey)
+		}
+	}
+
+	return outKeys
 }
 
 // MakeJSON returns a JSON value given a Go-style representation of JSON.
--- a/pkg/util/json/json_test.go
+++ b/pkg/util/json/json_test.go
@@ -21,7 +21,10 @@ import (
 	"strings"
 	"testing"
 
+	"bytes"
+	"github.com/cockroachdb/apd"
 	"github.com/cockroachdb/cockroach/pkg/sql/pgwire/pgerror"
+	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 )
 
@@ -552,6 +555,82 @@ func TestJSONRemoveIndex(t *testing.T) {
 				}
 			})
 		}
+	}
+}
+
+func getApdEncoding(num float64) *apd.Decimal {
+	dec := &apd.Decimal{}
+	dec, _ = dec.SetFloat64(num)
+	return dec
+}
+
+func arraySeparator(b []byte) []byte {
+	return encoding.EncodeArrayAscending(b)
+}
+
+func keyPrefix(b []byte) []byte {
+	return encoding.EncodeNotNullAscending(b)
+}
+
+func TestEncodeDecodeJSONInvertedIndex(t *testing.T) {
+	bytePrefix := make([]byte, 1, 100)
+	bytePrefix[0] = 0x0f
+
+	testCases := []struct {
+		value  string
+		expEnc [][]byte
+	}{
+		{`{"a":"b"}`, [][]byte{bytes.Join([][]byte{keyPrefix(bytePrefix),
+			encoding.EncodeStringAscending(nil, "a"), encoding.EncodeStringAscending(nil, "b")}, nil)}},
+		{`["a", "b"]`, [][]byte{bytes.Join([][]byte{arraySeparator(bytePrefix),
+			encoding.EncodeStringAscending(nil, "a")}, nil),
+			bytes.Join([][]byte{arraySeparator(bytePrefix),
+				encoding.EncodeStringAscending(nil, "b")}, nil)}},
+		{`null`, [][]byte{encoding.EncodeNullAscending(bytePrefix)}},
+		{`false`, [][]byte{encoding.EncodeFalseAscending(bytePrefix)}},
+		{`true`, [][]byte{encoding.EncodeTrueAscending(bytePrefix)}},
+		{`1.23`, [][]byte{encoding.EncodeDecimalAscending(bytePrefix, getApdEncoding(1.23))}},
+		{`"a"`, [][]byte{encoding.EncodeStringAscending(bytePrefix, "a")}},
+		{`["c", {"a":"b"}]`, [][]byte{bytes.Join([][]byte{arraySeparator(bytePrefix),
+			encoding.EncodeStringAscending(nil, "c")}, nil),
+			bytes.Join([][]byte{arraySeparator(bytePrefix),
+				keyPrefix(nil),
+				encoding.EncodeStringAscending(nil, "a"),
+				encoding.EncodeStringAscending(nil, "b")}, nil)}},
+		{`["c", {"a":["c","d"]}]`, [][]byte{bytes.Join([][]byte{
+			arraySeparator(bytePrefix),
+			encoding.EncodeStringAscending(nil, "c")}, nil),
+
+			bytes.Join([][]byte{arraySeparator(bytePrefix),
+				keyPrefix(nil),
+				encoding.EncodeStringAscending(nil, "a"),
+				arraySeparator(nil),
+				encoding.EncodeStringAscending(nil, "c")}, nil),
+
+			bytes.Join([][]byte{arraySeparator(bytePrefix),
+				keyPrefix(nil),
+				encoding.EncodeStringAscending(nil, "a"),
+				arraySeparator(nil),
+				encoding.EncodeStringAscending(nil, "d")}, nil)}},
+
+		{`{"a":"b","e":"f"}`, [][]byte{
+			bytes.Join([][]byte{keyPrefix(bytePrefix),
+				encoding.EncodeStringAscending(nil, "a"), encoding.EncodeStringAscending(nil, "b")}, nil),
+
+			bytes.Join([][]byte{keyPrefix(bytePrefix),
+				encoding.EncodeStringAscending(nil, "e"), encoding.EncodeStringAscending(nil, "f")}, nil),
+		}},
+	}
+
+	for _, c := range testCases {
+		enc := jsonTestShorthand(c.value).EncodeInvertedIndexKeys(bytePrefix)
+		for j, path := range enc {
+			if !bytes.Equal(path, c.expEnc[j]) {
+				t.Errorf("unexpected encoding mismatch for %v. expected [%#v], got [%#v]",
+					c.value, c.expEnc[j], path)
+			}
+		}
+
 	}
 }
 


jordanlewis:
Accidentally added some binary files here it looks like, pprof and a callgraph.

---

Reviewed 6 of 6 files at r1.

---

*[pkg/util/encoding/encoding_test.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMpExmG6zF7TFG8yv_:-KyMpExmG6zF7TFG8yva:b-4rxjjl) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/encoding/encoding_test.go#L30)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/util/duration"
> 	"github.com/cockroachdb/cockroach/pkg/util/ipaddr"
> 	"github.com/cockroachdb/cockroach/pkg/util/json"
> ```

nit: extra import

---

*[pkg/util/json/json.go, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMqT0CBCZ9d7YrZ_By:-KyMqT0CBCZ9d7YrZ_Bz:bxfje6z) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L59)):*
> ```Go
> 	Size() uintptr
> 
> 	// Encode for index returns binary encoding needed for the inverted index key.
> ```

s/Encode for index/EncodeForIndex/, and add a `the` before binary encoding

Also don't you think these should take in a byte slice to append to? They'll need to be appending to a table key prefix eventually. 

---

*[pkg/util/json/json.go, line 358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrYu7NqRkGbmAvZvo:-KyMrYu7NqRkGbmAvZvp:b-45jqaa) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L358)):*
> ```Go
> 	for i := range j {
> 		for _, childBytes := range j[i].EncodeForIndex() {
> 			outBytes = append(outBytes, append(encoding.EncodeArrayAscending(nil), childBytes...))
> ```

instead of passing nil here you should pass in the byte slice that will be passed to this function.  This is going to be very underperformant, since it has to allocate and throw away so many new byte slices.

---

*[pkg/util/json/json.go, line 369 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMr54gBwoiY2Yo3Eej:-KyMr54gBwoiY2Yo3Eek:bwg6rx) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L369)):*
> ```Go
> 	for i := range j {
> 		for _, childBytes := range j[i].v.EncodeForIndex() {
> 			outBytes = append(outBytes, append(append(encoding.EncodeNotNullAscending(nil),
> ```

This is really hard to read. I think it would be better to append each byte sequence on a source line of its own.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-KyMrvyCzU0q3ewqGXAw:bej6bux)*

justinj:
nit: extra space before the `:` in the commit header

---

---

*[pkg/util/encoding/encoding.go, line 669 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMsTNktigMTgFPfgbZ:-KyMsTNktigMTgFPfgb_:bxfiftk) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/encoding/encoding.go#L669)):*
> ```Go
> func EncodeTrueAscending(b []byte) []byte {
> 	return append(b, byte(True))
> }
> ```

nit: missing extra line

---

*[pkg/util/json/json.go, line 335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMt8Z5jr1gX58bUZGZ:-KyMt8Z5jr1gX58bUZG_:b934zss) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L335)):*
> ```Go
> 	return MakeJSON(result)
> }
> ```

I might be off the mark here, but I think `EncodeForIndex` might be confusing some things - it says it encodes for an index **key**, but it has implementations for `null`, `true`, `false`, and `number`, which never(?) exist in keys - only in values. Am I confused? Could the comment for `EncodeForIndex` be extended to explain exactly what its return value is used for, and what the format is? To my relatively uninformed eyes it seems that this method should return a pair of `(key, value)`?.

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyMq4S64DQLMpnX-yml:b-8mjwas) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*
> ```Go
> }
> 
> func (j jsonNull) EncodeForIndex() [][]byte {
> ```

I think it would make more sense to come up with new sentinel values here, rather than re-using the value tags from `encoding`, unless there's a pressing reason not to? These seem notably distinct from those encodings.

---

*[pkg/util/json/json.go, line 337 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMqdsPfohaLn7VFene:-KyMqdsPfohaLn7VFenf:b-mydpjr) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L337)):*
> ```Go
> 
> func (j jsonNull) EncodeForIndex() [][]byte {
> 	return [][]byte{encoding.EncodeNullAscending(nil)}
> ```

I think the pattern we use elsewhere is the right one - these encoding functions as currently written will force a lot of additional allocations over the usual pattern of passing in a byte slice to directly append to and then returning the result.

---

*[pkg/util/json/json_test.go, line 524 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrmsfMsH7XXzGZBqz:-KyMrmsfMsH7XXzGZBr-:bykve6o) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L524)):*
> ```Go
> }
> 
> type testCaseJSON struct {
> ```

Personally I generally prefer these to be defined inline and locally to the tests to limit their scope and emphasize where they apply to

---

*[pkg/util/json/json_test.go, line 529 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrk0IhLyee3m3cZ8m:-KyMrk0IhLyee3m3cZ8n:buj5c7e) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L529)):*
> ```Go
> }
> 
> func makeJson(str string) JSON {
> ```

this exists already as `jsonTestShorthand` (I usually alias inside the tests themselves to `json`).

---

*[pkg/util/json/json_test.go, line 534 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrXZ8GzKWs9WUYyig:-KyMrXZ8GzKWs9WUYyih:b-831qa9) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L534)):*
> ```Go
> }
> 
> func getApdEnconding(num float64) *apd.Decimal {
> ```

nit: `Enconding`

---

*[pkg/util/json/json_test.go, line 536 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMss7SGwvGNcPvElOf:-KyMss7SGwvGNcPvElOg:b-8zigu4) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L536)):*
> ```Go
> func getApdEnconding(num float64) *apd.Decimal {
> 	dec := apd.Decimal{}
> 	dec.SetFloat64(num)
> ```

This might be a little sketchy in some cases because this uses Float64 whereas ParseJSON uses Decimal. It will probably be fine for the scale of numbers used here, though.

---

*[pkg/util/json/json_test.go, line 540 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMufVoWSHP803AusZq:-KyMufVoWSHP803AusZr:b-2uujlp) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L540)):*
> ```Go
> }
> 
> func TestEncodeDecodeJSON(t *testing.T) {
> ```

Could you modify the name here to emphasize this is specifically for the inverted index encoding?

---

*[pkg/util/json/json_test.go, line 542 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMs-XTV5mXJhRH-LfZ:-KyMs-XTV5mXJhRH-Lf_:b-k5bt55) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L542)):*
> ```Go
> func TestEncodeDecodeJSON(t *testing.T) {
> 	testCases := []testCaseJSON{
> 		{makeJson(`{"a":"b"}`), [][]byte{bytes.Join([][]byte{encoding.EncodeNotNullAscending(nil),
> ```

Since the things on the left are always JSON, it might make this a little simpler if you instead make the test case parameterized by a string and then parse it in the test body.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-KyMp7Zd2pIs4-AZACup:bnue34i)*

justinj:

---

*[pkg/util/json/json.go, line 335 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMt8Z5jr1gX58bUZGZ:-KyMwcrCg0PFWvKjlium:bt5alo6) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L335)):*

I might be off the mark here, but I think `EncodeForIndex` might be confusing some things - it says it encodes for an index **key**, but it has implementations for `null`, `true`, `false`, and `number`, which never(?) exist in keys - only in values. Am I confused? Could the comment for `EncodeForIndex` be extended to explain exactly what its return value is used for, and what the format is? To my relatively uninformed eyes it seems that this method should return a pair of `(key, value)`?.
</blockquote></details>

ignore! jordan confirmed that i am in fact the confused one here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
Removed both.

---

Review status: 0 of 3 files reviewed at latest revision, 14 unresolved discussions.

---

*[pkg/util/encoding/encoding.go, line 669 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMsTNktigMTgFPfgbZ:-KyMxCcaxEy0K5TmMtzR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/encoding/encoding.go#L669)):*

nit: missing extra line
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMqT0CBCZ9d7YrZ_By:-KyMxhO7Oj_hAon-kxjk:b-i7kzq7) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L59)):*
> EncodeForIndex
Sure I'll pass a byte slice.

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyN0PBIqB4z6BDyks68:b4fahco) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*

I think it would make more sense to come up with new sentinel values here, rather than re-using the value tags from `encoding`, unless there's a pressing reason not to? These seem notably distinct from those encodings.
</blockquote></details>

Why? There's really no conflict with the existing encoding and I don't see a reason not to use what's there.

---

*[pkg/util/json/json.go, line 337 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMqdsPfohaLn7VFene:-KyN2UxEdq5s4P4Xv6td:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L337)):*

I think the pattern we use elsewhere is the right one - these encoding functions as currently written will force a lot of additional allocations over the usual pattern of passing in a byte slice to directly append to and then returning the result.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrYu7NqRkGbmAvZvo:-KyN2UQY3dW5pLRMIQOU:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L358)):*

instead of passing nil here you should pass in the byte slice that will be passed to this function.  This is going to be very underperformant, since it has to allocate and throw away so many new byte slices.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 369 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMr54gBwoiY2Yo3Eej:-KyN3jT4dzJXXFc2-Tui:bu7i25p) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L369)):*

This is really hard to read. I think it would be better to append each byte sequence on a source line of its own.
</blockquote></details>

Make sense.

---

*[pkg/util/json/json_test.go, line 524 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrmsfMsH7XXzGZBqz:-KyN3-2B2fDdcgsxo-WQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L524)):*

Personally I generally prefer these to be defined inline and locally to the tests to limit their scope and emphasize where they apply to
</blockquote></details>

Done.

---

*[pkg/util/json/json_test.go, line 529 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrk0IhLyee3m3cZ8m:-KyN34emd1Lbt9WzuM21:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L529)):*

this exists already as `jsonTestShorthand` (I usually alias inside the tests themselves to `json`).
</blockquote></details>

Done.

---

*[pkg/util/json/json_test.go, line 534 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMrXZ8GzKWs9WUYyig:-KyN2cYOJMdclCS010pQ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L534)):*

nit: `Enconding`
</blockquote></details>

Done.

---

*[pkg/util/json/json_test.go, line 536 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMss7SGwvGNcPvElOf:-KyN376sgqNsctrLa5pa:b-ptel5h) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L536)):*

This might be a little sketchy in some cases because this uses Float64 whereas ParseJSON uses Decimal. It will probably be fine for the scale of numbers used here, though.
</blockquote></details>

It think it's ok because I'm testing the encoding of Decimal and initializing it with reasonable values.

---

*[pkg/util/json/json_test.go, line 540 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMufVoWSHP803AusZq:-KyN3Lga-B3fF3RXqJX4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L540)):*

Could you modify the name here to emphasize this is specifically for the inverted index encoding?
</blockquote></details>

Done.

---

*[pkg/util/json/json_test.go, line 542 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMs-XTV5mXJhRH-LfZ:-KyN3e-DlkzJSRBU4SQE:b-xq5nlh) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L542)):*

Since the things on the left are always JSON, it might make this a little simpler if you instead make the test case parameterized by a string and then parse it in the test body.
</blockquote></details>

Yep agreed looks cleaner.

---

*[pkg/util/encoding/encoding_test.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMpExmG6zF7TFG8yv_:-KyMxY8Jdd_iUkSQrok0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/encoding/encoding_test.go#L30)):*

nit: extra import
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-KyMwrRv0NaFDSitNkLY:b-ts4a2g)*

justinj:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyN6sScUiHUmpZ1-iS7:bt3j7x) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*

Why? There's really no conflict with the existing encoding and I don't see a reason not to use what's there.
</blockquote></details>

In my mind this is an entirely different situation than those. JSON `true` is completely distinct from SQL `true`and it's the same with the other values, and I feel like reusing the variables conflates them.

---

*[pkg/util/json/json.go, line 336 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN7cf2tmkhIkGivTTK:-KyN7cf2tmkhIkGivTTL:b-8gvib3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L336)):*
> ```Go
> }
> 
> func (j jsonNull) EncodeForIndex(b []byte) [][]byte {
> ```

Could you still extend the comment for `EncodeForIndex` to explain precisely what the interface it presents is? This seems a little weird to me still, the goal of passing in the byte slice is to avoid doing more allocations than necessary, but we're still allocating on every call here. Is there a way we can set this up so that we only ever append to the passed-in byte slice and don't allocate any new slices?

---

*[pkg/util/json/json.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8HyYZsgm_j-M5ke-:-KyN8HyYZsgm_j-M5ke0:b-isi606) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L371)):*
> ```Go
> 			outBytes = append(outBytes,
> 				bytes.Join([][]byte{
> 					encoding.EncodeNotNullAscending(b),
> ```

This seems wrong, passing `b` through like this multiple times in one statement. I think each append here should be done in a new statement, threading `b` through.

---

*[pkg/util/json/json_test.go, line 536 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMss7SGwvGNcPvElOf:-KyN8WqvU_LxxMSf9y6U:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json_test.go#L536)):*

It think it's ok because I'm testing the encoding of Decimal and initializing it with reasonable values.
</blockquote></details>

:+1:

---

*[pkg/util/json/json_test.go, line 544 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8ip-hxlF4-BEHgJ3:-KyN8ip-hxlF4-BEHgJ4:bv2m1py) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json_test.go#L544)):*
> ```Go
> 		{`1.23`, [][]byte{encoding.EncodeDecimalAscending(nil, getApdEncoding(1.23))}},
> 		{`"a"`, [][]byte{encoding.EncodeStringAscending(nil, "a")}},
> 		{`["c", {"a":"b"}]`, [][]byte{append(encoding.EncodeArrayAscending(nil),
> ```

Could you also add a test with more than one level of object and more than one key per level?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

jordanlewis:
Reviewed 6 of 6 files at r2.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyNIG-_tWWENVFnXMDl:b-8usq2j) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*
> ```Go
> 
> 	for i := range j {
> 		for _, childBytes := range j[i].EncodeForIndex(b) {
> ```

Hmm this is still confusing. It's not obvious what the order of operations here are at all.

I would do something like

```
for i, elt := range j {
  b = encoding.EncodeArrayAscending(b)
  b = elt.encodeForIndex(b)
}

return b
```

---

*[pkg/util/json/json.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8HyYZsgm_j-M5ke-:-KyNJMSd_duaEf4Dk0Sb:b-h5c4av) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L371)):*

This seems wrong, passing `b` through like this multiple times in one statement. I think each append here should be done in a new statement, threading `b` through.
</blockquote></details>

Something like the example I wrote above would be better here too I think.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
Review status: 2 of 3 files reviewed at latest revision, 10 unresolved discussions.

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyRIHom2MJ66-5M_Tpn:b-6kppad) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*

In my mind this is an entirely different situation than those. JSON `true` is completely distinct from SQL `true`and it's the same with the other values, and I feel like reusing the variables conflates them.
</blockquote></details>

We also no that these we chosen to no overlap with anything else in the encoding scheme. I don't want to accidentally introduce any nondeterminism into the key generation.

---

*[pkg/util/json/json.go, line 336 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN7cf2tmkhIkGivTTK:-KyRIadmXPV6t4NoKySS:b-rn0nov) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L336)):*

Could you still extend the comment for `EncodeForIndex` to explain precisely what the interface it presents is? This seems a little weird to me still, the goal of passing in the byte slice is to avoid doing more allocations than necessary, but we're still allocating on every call here. Is there a way we can set this up so that we only ever append to the passed-in byte slice and don't allocate any new slices?
</blockquote></details>

We can't do that for the arrays and objects because we need to generate multiple paths for the same prefix.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyRh3l8EPOmRIgcKrkT:b-ufc10s) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*
>   b = encoding.EncodeArrayAscending(b)
I'm a bit confused, elt.encodeForIndex(b) returns an array of byte arrays.

---

*[pkg/util/json/json.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8HyYZsgm_j-M5ke-:-KyRi9iaPN_eIgMRcRtC:b-kct16o) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L371)):*

Something like the example I wrote above would be better here too I think.
</blockquote></details>

Same comment as for above.

---

*[pkg/util/json/json_test.go, line 544 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8ip-hxlF4-BEHgJ3:-KyRwjGdjxYMZIXyVWfr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json_test.go#L544)):*

Could you also add a test with more than one level of object and more than one key per level?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

jordanlewis:
Reviewed 1 of 1 files at r3.

---

*[pkg/util/json/json.go, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMqT0CBCZ9d7YrZ_By:-KyRzukyb9MIHogThoLl:b-39kyyc) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L59)):*

> EncodeForIndex
Sure I'll pass a byte slice.
</blockquote></details>

The comment here needs some improvement I think. It should be clear that we're returning one key per path, which I had totally forgotten about when I was reading this =)

Maybe something like:

EncodeForIndex takes in a key prefix and returns a slice of inverted index keys, one per path through the receiver. 

To make it more clear to a code reader that this method returns many keys, I would recommend renaming this function to `EncodeInvertedIndexKeys` or something similar.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyRzkL0A-DQkMQYIWCu:b3cwph) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

>   b = encoding.EncodeArrayAscending(b)
I'm a bit confused, elt.encodeForIndex(b) returns an array of byte arrays.
</blockquote></details>

Oh yeah. Now I understand what's going on.  Added a comment to the interface regarding the function naming.

I would still prefer to see a simpler way of expressing this algorithm. Personally, I find a one liner with 2 appends and a `...` very hard to read and understand.

Maybe something like this? I might still be missing some complexity.

```
newKey := make([]int, len(b), <can you estimate the ouptut key size?>)
copy(newKey, b)
newKey = encoding.EncodeArrayAscending(newKey)
newKey = elt.encodeForIndex(newKey)
outBytes = append(outBytes, newKey)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KySn1SwwQFr-xjWDes4:bqvxpkl) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*
> newKey := make([]int, len(b), <can you estimate the ouptut key size?>)
> copy(newKey, b)
> newKey = encoding.EncodeArrayAscending(newKey)
> newKey = elt.encodeForIndex(newKey)
> outBytes = append(outBytes, newKey)

I think you're still missing a layer because elt.encodeForIndex(newKey) will also return an array of byte arrays.  I added another variable, not sure how much clearer it is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

justinj:
Review status: 1 of 3 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyXEya97U-pE1l1HDu1:b3bdvz5) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*

We also no that these we chosen to no overlap with anything else in the encoding scheme. I don't want to accidentally introduce any nondeterminism into the key generation.
</blockquote></details>

I'm not super convinced this solves that problem though? If you look at how these tags are used, they're not just appended as a byte as is done here, they use some more complex process via `EncodeValueTag`. Sorry, my understanding of this area of the codebase and its concerns is a bit lacking.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyX5Vo1qnpp7TI_1vJi:bk2mn8j) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

> newKey := make([]int, len(b), <can you estimate the ouptut key size?>)
> copy(newKey, b)
> newKey = encoding.EncodeArrayAscending(newKey)
> newKey = elt.encodeForIndex(newKey)
> outBytes = append(outBytes, newKey)

I think you're still missing a layer because elt.encodeForIndex(newKey) will also return an array of byte arrays.  I added another variable, not sure how much clearer it is.
</blockquote></details>

Ok, now that I think I understand what this function is supposed to do, I now don't understand what `b` being passed in here represents, could you help me understand that? Since we're not appending to a single byte slice it might not be helpful.

---

*[pkg/util/json/json.go, line 342 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXEO-680DEgFMPlJem:-KyXEO-680DEgFMPlJen:b-1ne926) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L342)):*
> ```Go
> 	return [][]byte{encoding.EncodeTrueAscending(b)}
> }
> func (j jsonFalse) EncodeForIndex(b []byte) [][]byte {
> ```

nit: here and in true and null you can omit the receiver name to make it clear that it's unused

---

*[pkg/util/json/json.go, line 357 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXIbpyAdiogWRgDVSl:-KyXIbpyAdiogWRgDVSm:bp9ydpv) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L357)):*
> ```Go
> 	for i := range j {
> 		for _, childBytes := range j[i].EncodeForIndex(b) {
> 			encodedValue := append(encoding.EncodeArrayAscending(b), childBytes...)
> ```

Does there need to be a separator after the array tag (as the RFC seems to describe)? Could there be a conflict with a string that happened to start with the same character? ditto with the other tags? Also I'm not super sure, but does this use of `...` cause another allocation?

---

*[pkg/util/json/json_test.go, line 544 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXDOpDQhRSox9PfReu:-KyXDOpDQhRSox9PfRev:bim6muz) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json_test.go#L544)):*
> ```Go
> 		{`true`, [][]byte{encoding.EncodeTrueAscending(nil)}},
> 		{`1.23`, [][]byte{encoding.EncodeDecimalAscending(nil, getApdEncoding(1.23))}},
> 		{`"a"`, [][]byte{encoding.EncodeStringAscending(nil, "a")}},
> ```

Just a thought - maybe creating some local shorter aliases to functions like `encoding.EncodeStringAscending` (that also partially apply `nil`) would make these tests more legible? 

---

*[pkg/util/json/json_test.go, line 570 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXBgEq3hgv2lkwmUbC:-KyXBgEq3hgv2lkwmUbD:bx23r5c) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json_test.go#L570)):*
> ```Go
> 	}
> 
> 	//t.Logf("[%#v]", jsonTestShorthand(`["c", {"a":["c","d"]}]`).EncodeForIndex(nil))
> ```

nit: should be deleted?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

jordanlewis:
Reviewed 2 of 2 files at r4.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyXKPXlP4QQSN3ecD5Y:b-7wkhbq) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

> newKey := make([]int, len(b), <can you estimate the ouptut key size?>)
> copy(newKey, b)
> newKey = encoding.EncodeArrayAscending(newKey)
> newKey = elt.encodeForIndex(newKey)
> outBytes = append(outBytes, newKey)

I think you're still missing a layer because elt.encodeForIndex(newKey) will also return an array of byte arrays.  I added another variable, not sure how much clearer it is.
</blockquote></details>

This is much clearer. Thank you. My last complaint is that `encodedValue` is a strange variable name for an encoded key, and `outBytes` is a strange variable name for a *collection* of out bytes. Maybe something like `outKeys`.

---

*[pkg/util/json/json.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8HyYZsgm_j-M5ke-:-KyXKb28O2qem2mXWzdn:b-ms9xtf) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L371)):*

Same comment as for above.
</blockquote></details>

Now that you've improved the above,  can you improve this one as well? 

---

*[pkg/util/json/json.go, line 353 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXJJARuSwDoZwOTh68:-KyXJJARuSwDoZwOTh69:b35mlm) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L353)):*
> ```Go
> }
> func (j jsonArray) EncodeForIndex(b []byte) [][]byte {
> 	var outBytes [][]byte
> ```

Presize this to the length of `j` like we discussed

---

*[pkg/util/json/json.go, line 366 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXJOdJIYSkabs-Peo7:-KyXJOdKiYf9BvAja7sP:b35mlm) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L366)):*
> ```Go
> 
> func (j jsonObject) EncodeForIndex(b []byte) [][]byte {
> 	var outBytes [][]byte
> ```

Presize this to the length of `j` like we discussed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

justinj:

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyXLhQq_XFh90msp2Dp:b8l81tx) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*

I'm not super convinced this solves that problem though? If you look at how these tags are used, they're not just appended as a byte as is done here, they use some more complex process via `EncodeValueTag`. Sorry, my understanding of this area of the codebase and its concerns is a bit lacking.
</blockquote></details>

I should clarify that I don't care too much, if you think this makes more sense then go with it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
Review status: 1 of 3 files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[pkg/util/json/json.go, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMqT0CBCZ9d7YrZ_By:-KyXgrFfRbnYndAtlDub:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L59)):*

The comment here needs some improvement I think. It should be clear that we're returning one key per path, which I had totally forgotten about when I was reading this =)

Maybe something like:

EncodeForIndex takes in a key prefix and returns a slice of inverted index keys, one per path through the receiver. 

To make it more clear to a code reader that this method returns many keys, I would recommend renaming this function to `EncodeInvertedIndexKeys` or something similar.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 336 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyMq4S5lOIJlBJ0XSqV:-KyqUYUrLvHEZ0Lr4NxF:bigpj19) ([raw file](https://github.com/cockroachdb/cockroach/blob/2616b6e104ff7a4d95d986b1a0d31bca26859dbe/pkg/util/json/json.go#L336)):*

I should clarify that I don't care too much, if you think this makes more sense then go with it.
</blockquote></details>

Ok, sounds good!

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyXfP7tIEeFf6UCMD5x:bpx9uzf) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

This is much clearer. Thank you. My last complaint is that `encodedValue` is a strange variable name for an encoded key, and `outBytes` is a strange variable name for a *collection* of out bytes. Maybe something like `outKeys`.
</blockquote></details>

Agreed, renaming everything to key.

---

*[pkg/util/json/json.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyN8HyYZsgm_j-M5ke-:-KyXfk7ToT7vMJ6pHvQf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L371)):*

Now that you've improved the above,  can you improve this one as well? 
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 342 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXEO-680DEgFMPlJem:-KyXeOXTZY8xqDJPM9jN:bdcv6kv) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L342)):*

nit: here and in true and null you can omit the receiver name to make it clear that it's unused
</blockquote></details>

Good point!

---

*[pkg/util/json/json.go, line 353 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXJJARuSwDoZwOTh68:-KyXerEBf3u1nWge0ZNb:bxjjsyr) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L353)):*

Presize this to the length of `j` like we discussed
</blockquote></details>

I thought about this a bit more and even though we know how many values we have in an array or an object, we still don't know how many byte arrays the EncodeForIndex functions are going to return for the values.

---

*[pkg/util/json/json.go, line 357 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXIbpyAdiogWRgDVSl:-KyXh5GgfHQ4RKF9NWTO:b-qhuo47) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L357)):*

Does there need to be a separator after the array tag (as the RFC seems to describe)? Could there be a conflict with a string that happened to start with the same character? ditto with the other tags? Also I'm not super sure, but does this use of `...` cause another allocation?
</blockquote></details>

The array separator is the separator. I don't think strings start with the same character, it's using on of the type tags from the encoding package which we described in the RFC. We do need to prepend the tag so I'm not sure if allocation is avoidable.

---

*[pkg/util/json/json.go, line 366 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXJOdJIYSkabs-Peo7:-KyXfXt-FhjY66g67Sx7:benpsw9) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L366)):*

Presize this to the length of `j` like we discussed
</blockquote></details>

Same comment as above.

---

*[pkg/util/json/json_test.go, line 544 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXDOpDQhRSox9PfReu:-KyqUNEQm9LULDwsRbj1:b-4iun0j) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json_test.go#L544)):*

Just a thought - maybe creating some local shorter aliases to functions like `encoding.EncodeStringAscending` (that also partially apply `nil`) would make these tests more legible? 
</blockquote></details>

Added some functions for separators.

---

*[pkg/util/json/json_test.go, line 570 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXBgEq3hgv2lkwmUbC:-KyXfwF8LtF1hv_6aPGI:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json_test.go#L570)):*

nit: should be deleted?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

justinj:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyqXXhSxEef_iBCPy0d:bfbg4ek) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

Agreed, renaming everything to key.
</blockquote></details>

My only point of confusion left is the byte-slice here - it's used sort of inconsistently (referenced in the `jsonString` impl but not the `jsonNumber` impl) and it's still not clear to me what it represents?

---

*[pkg/util/json/json.go, line 357 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyXIbpyAdiogWRgDVSl:-KyqW65KDdL_EOOrbNdI:b-rwc3yg) ([raw file](https://github.com/cockroachdb/cockroach/blob/4eb7ff76cfc5e4961afaae3343d7f88a8173db75/pkg/util/json/json.go#L357)):*

The array separator is the separator. I don't think strings start with the same character, it's using on of the type tags from the encoding package which we described in the RFC. We do need to prepend the tag so I'm not sure if allocation is avoidable.
</blockquote></details>

Oh I see, I missed that `EncodeStringAscending` includes a tag :thumbsup:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-Kyq_GHRCF4cRf57rFF5:b384z5b) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

My only point of confusion left is the byte-slice here - it's used sort of inconsistently (referenced in the `jsonString` impl but not the `jsonNumber` impl) and it's still not clear to me what it represents?
</blockquote></details>

Good catch, byte slice is added.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

jordanlewis:
:lgtm:

---

Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-KyrJxynsJaSE1HpfdJf:bnfp4nl)*

jordanlewis:
:lgtm_cancel: 

---

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyrLN9HWUrHObw-t6M_:b-817cqi) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

Good catch, byte slice is added.
</blockquote></details>

Wait a minute I'm confused. Isn't this wrong? `b` is our key prefix right? So we want to append it just once to all of our output keys.

It seems to me that since we're calling this function recursively, we're accidentally prepending our key prefix more than once. Isn' that true?

I was wondering why this wasn't caught by the tests. It's because you never pass a key prefix to the encode function. We definitely need to test that case, because that is always how this function will be used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-KyrLr3zBPQK4ERqfpRZ:b8b79qu)*

jordanlewis:

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyrLvzmz2CIkuhbnF47:biwxr3a) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

Wait a minute I'm confused. Isn't this wrong? `b` is our key prefix right? So we want to append it just once to all of our output keys.

It seems to me that since we're calling this function recursively, we're accidentally prepending our key prefix more than once. Isn' that true?

I was wondering why this wasn't caught by the tests. It's because you never pass a key prefix to the encode function. We definitely need to test that case, because that is always how this function will be used.
</blockquote></details>

s/append/prepend/ in the first paragraph there, sorry

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

jordanlewis:

---

*[pkg/util/json/json_test.go, line 587 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyrO_qtA7lAO1BwcPVi:-KyrO_qtA7lAO1BwcPVj:b-pjlktn) ([raw file](https://github.com/cockroachdb/cockroach/blob/a42704e9a46eeb7199d68dbeef71d60bd13af6c2/pkg/util/json/json_test.go#L587)):*
> ```Go
> 
> 	for _, c := range testCases {
> 		enc := jsonTestShorthand(c.value).EncodeInvertedIndexKeys(nil)
> ```

Specifically, I think these test cases should all be testing `EncodeInvertedIndexeKeys(b)` where `b` is some arbitrary common bytes prefix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyvNm1jFIa_CZQ6u1Ym:bf86iv1) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

s/append/prepend/ in the first paragraph there, sorry
</blockquote></details>

No we're not appending it recursively because we're using the output of the Encode methods, not the modified datum. We still need to prepend the prefix to every child.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

jordanlewis:

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyvbVm7oN0qCEuhdtJQ:b-1vp3td) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

No we're not appending it recursively because we're using the output of the Encode methods, not the modified datum. We still need to prepend the prefix to every child.
</blockquote></details>

I understand that the prefix needs to be appended to every child, but that needs to happen only once, right?

Let's say we call this on `[1]` with prefix `foo`. We recursively call encodeInvIndKey on each of the elements of the array - in this case there's just the one, let's call it `jsonNumber(1)`.

`jsonNumber(1).encodeInvIndKey(foo)` will return a single key with value `foo/dec/1`. 

So the first encoded key will be `append(encodeArray(foo)), 'foo/dec/1)`, or `foo/arr/foo/dec/`.

Maybe I'm still missing something! Either way we need to test this with an explicit prefix passed in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

justinj:

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyveoChqq74wPVpd4NW:b-bt5qiy) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

I understand that the prefix needs to be appended to every child, but that needs to happen only once, right?

Let's say we call this on `[1]` with prefix `foo`. We recursively call encodeInvIndKey on each of the elements of the array - in this case there's just the one, let's call it `jsonNumber(1)`.

`jsonNumber(1).encodeInvIndKey(foo)` will return a single key with value `foo/dec/1`. 

So the first encoded key will be `append(encodeArray(foo)), 'foo/dec/1)`, or `foo/arr/foo/dec/`.

Maybe I'm still missing something! Either way we need to test this with an explicit prefix passed in.
</blockquote></details>

Separate from what Jordan is describing, I think there's another problem with the use of `b` - it's being `append`ed to multiple times, which doesn't necessarily force a copy of the underlying array, so it's possible that we're going to be overwriting data that we've already assumed to be written. I think in the array case there needs to be an explicit copy of the bytes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-KyvsnHr5u2G8ApgGxK_:byx9t41) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

Separate from what Jordan is describing, I think there's another problem with the use of `b` - it's being `append`ed to multiple times, which doesn't necessarily force a copy of the underlying array, so it's possible that we're going to be overwriting data that we've already assumed to be written. I think in the array case there needs to be an explicit copy of the bytes.
</blockquote></details>

I've updated the code with another prefix in the tests, I think it shows that we're not overwriting anything.

---

*[pkg/util/json/json_test.go, line 587 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyrO_qtA7lAO1BwcPVi:-Kyvt-dPgX9gLEsvKzdf:bqmjwcm) ([raw file](https://github.com/cockroachdb/cockroach/blob/a42704e9a46eeb7199d68dbeef71d60bd13af6c2/pkg/util/json/json_test.go#L587)):*

Specifically, I think these test cases should all be testing `EncodeInvertedIndexeKeys(b)` where `b` is some arbitrary common bytes prefix.
</blockquote></details>

Added!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

justinj:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-Kyvui9BvsSJGfMFJjH6:b-8msdq7) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

I've updated the code with another prefix in the tests, I think it shows that we're not overwriting anything.
</blockquote></details>

I still think this is wrong - I think it's succeeding in this case because the byte slice has a fixed size and thus every `append` forces a reallocation (and thus a copy), if you specify the prefix as something like
```
bytePrefix := make([]byte, 1, 100)
bytePrefix[0] = 0x0f
```
I think that will reveal the issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

m-schneider:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-Kyw5lLFpmRTkyP9b9q0:bqfdaf3) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

I still think this is wrong - I think it's succeeding in this case because the byte slice has a fixed size and thus every `append` forces a reallocation (and thus a copy), if you specify the prefix as something like
```
bytePrefix := make([]byte, 1, 100)
bytePrefix[0] = 0x0f
```
I think that will reveal the issue.
</blockquote></details>

Great point, Justin. I fixed that and the tests!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884)*

justinj:
Looks great now! :lgtm:

---

Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/util/json/json.go, line 357 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-KyNIG-_tWWENVFnXMDk:-Kyw60ji3K4_xaPUT4W_:b-ucfknh) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2150629f637b7d166692df82f28292a31d5ae2a/pkg/util/json/json.go#L357)):*

Great point, Justin. I fixed that and the tests!
</blockquote></details>

Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-Kyw6P9TtqxjB9YtZxfI:b-aoeqyr)*

jordanlewis:
:lgtm:

---

Reviewed 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19884#-:-Kyw6r8AMVkEqGS0eOJU:bnfp4nl)*

PR #19886, opened 2017-11-07
Makefile: allow `make test TESTS=-`

jordanlewis:
This is useful as it compiles all tests and serves as a quick
sanity check during refactors. 

Also fix `make lint TESTS=foo`.


tbg (approved):
Thanks!

Issue #16262, opened 2017-06-01
sql: actively cancel queries when client connection drops


knz:
Reported in #16168: if a long query is issued and the client connection is dropped, the query remains running in the background. 

The problem with this is that some apps may issue a long query, timeout on the lack of result and close their connection, and open a new connection to try again, resulting in a mounting pile of queries that end up all slowing each other down and making the server less and less responsive.

There are three proximate causes that both need to be addressed to solve this:

1) the connection being closed must result in the context being visibly cancelled. This is not currently the case: *while a query is running*, pgwire does not interact with the connection (no reads/writes), and thus the net package has no reason to find that the connection has been closed . This may be partially alleviated once #7775 is addressed, but not completely: any query that does a lot of processing before a single row is returned (e.g. full table scan + filter) or when there are no rows (e.g. insert) would still be susceptible.
2) the local SQL engine must learn to regularly check the context to see if it's been cancelled. 
3) context cancellation at the gateway must propagate through the network of distsql processors.

cc @andreimatei for technical insight
cc @petermattis @cuongdo for triage and prioritization

petermattis:
To address 1, we'll need to have a separate goroutine that reads from the connection vs the one that executes queries. I think that is independent of #7775 (though I suppose it could be done at the same time). As long as we're deriving contexts correctly, we get 3 for free.

andreimatei:
This issue is part of @itsbilal's project, isn't it?

knz:
Not that I know, no. Bilal's project is about cancelling queries/sessions remotely (from another session), not reacting from a connection drop by a client.

cuongdo:
Reassigning unassigned 1.1 issues. Is this staying in 1.1?

petermattis:
The restructuring of how we read from the pgwire connection is relatively small. Assuming Bilal's work properly deals with watching the `Context` for cancellation, I think the remaining work here should be straightforward.

andreimatei:
> The restructuring of how we read from the pgwire connection is relatively small

Yeah... For a better woman, perhaps.
This has been addressed following the principle of  maximal intrusion in https://github.com/cockroachdb/cockroach/pull/22277.

PR #19889, opened 2017-11-07
sql,parser: Add SCRUB DATABASE and fix an issue with NULL values

lgo:
This adds the statement

    SCRUB DATABASE <database>

Which will run checks by default. In the future, specifying certain
check types may be possible.

Also, if a value in the secondary column was NULL this would cause
false positives due to the equivalence on `NULL = NULL`. This is fixed.

This also refactors things into a much cleaner interface.
Specific check logic has been separated into individual structs. These
structs implement an iterator-like interface that the main SCRUB
execution iterates through. This makes it far simpler and cleaner to add
new checks.

Ideally, when adding a new check, all that needs to be done is
implementing the interface for that check and adding code for
initialization.


lego:
The `indexCheckDetails` struct is not very ideal. I have plans on changing the way checks are "queued" up and iterated through to a much cleaner iterator-like `interface`. Each type of check could have a fairly simple set of functions, `Start`, `Next`, `Done` and a single queue is populated in the beginning. That change is incoming in the physical change list.

knz:
:lgtm:  with a test nit.

Also I'd like to see a bit more attention given to the fact these statements are going to run for a .... loooong .... time in even moderately sized databases:

- have you checked that a long-running scrub properly aborts when the client disconnects?
- have you checked that a long-running scrub can be canceled with CANCEL QUERY?
- can you remind me at which stage of your project approximately you're planning to introduce jobs?

---

Reviewed 8 of 8 files at r1.
Review status: 6 of 8 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/scrub, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-KySTcnGhcoLgQrAB3b7:-KySTcnGhcoLgQrAB3b8:be0dddt) ([raw file](https://github.com/cockroachdb/cockroach/blob/41f221c0c1b34a8746fe618acf18d8f6b6619546/pkg/sql/logictest/testdata/logic_test/scrub#L38)):*
> ```
> 
> statement error "v1" is not a table
> EXPERIMENTAL SCRUB TABLE v1
> ```

This test does not belong to this commit -- make a separate commit for it.

---

*[pkg/sql/parser/sql.y, line 2029 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-KySTKkZF3-n-OC2r6Rv:-KySTKkZF3-n-OC2r6Rw:b-dwro0m) ([raw file](https://github.com/cockroachdb/cockroach/blob/41f221c0c1b34a8746fe618acf18d8f6b6619546/pkg/sql/parser/sql.y#L2029)):*
> ```Yacc
>   {
>     $$.val = &Scrub{Typ: ScrubDatabase, Database: Name($4)}
>   }
> ```

Can you also follow-up with another PR where you add the complementary `| ... error // SHOW HELP: ...` non-terminals. Look elsewhere in this file for examples.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-:-KySWhFRxL69QJQGTA0M:bwxra2h)*

knz:
Reviewed 2 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889)*

lego:
TFTR! :)

> have you checked that a long-running scrub properly aborts when the client disconnects?

It does not abort when the client disconnects. How would I go about making sure that happens? (I tested this just by CTRL+C'ing the CLI. Is there a proper way to close a session?)

> have you checked that a long-running scrub can be canceled with CANCEL QUERY?

It respond immediately to CANCEL QUERY

> can you remind me at which stage of your project approximately you're planning to introduce jobs?

For now, I've been focusing on adding the various checks.

As for introducing jobs, adding only `CANCEL` support can be done in a simple, albeit hacky manner: it can just delegate the call to  `CANCEL QUERY`. In order to do the rest, the short version is we need to be able to provide a callback to `delegateQuery` and `distSQLPlanner.Run` that runs between units of work to provide responsiveness of `PAUSE`, but we are still left with resuming from the beginning of an individual check.

The crux is for the RESUME/PAUSE job controls. Pausing relies on the running job to call `Job.Progressed`. All of the checks rely on delegating queries, so we can only call `Progressed` between specific checks. The same problem applies for RESUME, where we will only be able to RESUME to a point at the beginning or end of an individual check.

It can be done, but I'll think about this more whether there is a better way. This serves no purpose for checks against a single index or constraint. Where there are benefits is where a `SCRUB DATABASE ...` call is made, and it might check more than 10 indexes, constraints, tables, etc.

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/scrub, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-KySTcnGhcoLgQrAB3b7:-KySYGK9dMFh-AHZWaiV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/41f221c0c1b34a8746fe618acf18d8f6b6619546/pkg/sql/logictest/testdata/logic_test/scrub#L38)):*

This test does not belong to this commit -- make a separate commit for it.
</blockquote></details>

Done.

---

*[pkg/sql/parser/sql.y, line 2029 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-KySTKkZF3-n-OC2r6Rv:-KySYNF40FLyfJKlfNaL:bhy9ot6) ([raw file](https://github.com/cockroachdb/cockroach/blob/41f221c0c1b34a8746fe618acf18d8f6b6619546/pkg/sql/parser/sql.y#L2029)):*

Can you also follow-up with another PR where you add the complementary `| ... error // SHOW HELP: ...` non-terminals. Look elsewhere in this file for examples.
</blockquote></details>

Will do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-:-KySZ_yOSD_RuSRYrqoJ:bqes8c)*

knz:
Regarding the problem with queries that keep running when clients disconnect -- I had thought this was addressed already, but it might have not been done properly. Could you check with another query that also doesn't produce any output until it's complete, like `select count(*) from generate_series(1, 10000000000)` and see if that properly aborts. If there's an open problem, please approach Andrei and ask him for advice -- is this a known issue? if so, please add a comment about SCRUB on the known issue. Otherwise, create one.

As to progress control on a job, that's a separate discussion. There are multiple approaches possible. I would be OK with a first solution that only calls Progressed in between checks (i.e. not *while* scanning a table/index). That should be easier I think?

---

Reviewed 6 of 8 files at r6, 1 of 2 files at r7, 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-:-KyVHG1wTKDzESUMM2So:b-llvbns)*

lego:
Seems like it's a general issue, https://github.com/cockroachdb/cockroach/issues/16262. When you say add notes somewhere, I'm not quite sure where you're referring to or a good place to add them (commit, help message, etc.)

Yea, I'll work on the jobs bit at some point then and make that first iteration :)

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19889#-:-KyWbKe_eDuv-IK5o6QO:brk0q4m)*

PR #19897, opened 2017-11-07
rpc: add a stats handler to grpc traffic to measure throughput

spencerkimball:
Also, augment the `NodeStatus` protobuf with maps for incoming and
outgoing throughput from a node to all other connected nodes,
similar to latencies.


petermattis:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/rpc/stats_handler.go, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyR5e8irLPmOCg5EEj2:-KyR5e8irLPmOCg5EEj3:bssdvff) ([raw file](https://github.com/cockroachdb/cockroach/blob/f873ed72c9321e7a95cdd6c938e63b73b238134c/pkg/rpc/stats_handler.go#L28)):*
> ```Go
> )
> 
> const remoteAddrKey = "remoteAddr"
> ```

Context keys are usually types. This should be something like:

```
type remoteAddrKey struct{}
```

And then you use it as `remoteAddrKey{}`. This is better than what is currently here because stuffing a string into an interface requires an allocation. While stuffing an empty struct into an interface does not. 

---

*[pkg/rpc/stats_handler.go, line 105 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyR6H2KnfYE0d_gQRtF:-KyR6H2KnfYE0d_gQRtG:b-8xkn9u) ([raw file](https://github.com/cockroachdb/cockroach/blob/f873ed72c9321e7a95cdd6c938e63b73b238134c/pkg/rpc/stats_handler.go#L105)):*
> ```Go
> // client connections from this node to remote nodes.
> type StatsHandler struct {
> 	// stats is a map from remote targets to Stats objects.
> ```

Should mention that we never remove items from this map. Doesn't seem like that will be a problem in practice, but something for a future reader to be aware of.

---

*[pkg/server/status/status.proto, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyR6dMlR6XYhKWL34BQ:-KyR6dMlR6XYhKWL34BR:bdvmnfw) ([raw file](https://github.com/cockroachdb/cockroach/blob/f873ed72c9321e7a95cdd6c938e63b73b238134c/pkg/server/status/status.proto#L50)):*
> ```protobuf
>     (gogoproto.nullable) = false,
>     (gogoproto.castkey) = "github.com/cockroachdb/cockroach/pkg/roachpb.NodeID"
>   ];
> ```

These maps will get fairly large in large clusters. I wonder if that will be problematic. 

Can we change this to a single map from NodeID to <latency,incoming,outgoing>? Somewhat more efficient on the wire to do that. Not sure if there are backwards compatibility issues with this proto. Is it only used by the admin UI, or is it ever persisted to disk?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897)*

petermattis:
:lgtm:

---

Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-:-KyR7O28c7XYf3fQzLvB:bnfp4nl)*

spencerkimball:
Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/rpc/stats_handler.go, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyR5e8irLPmOCg5EEj2:-KyRIeKSC6xx6VIlMcWJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f873ed72c9321e7a95cdd6c938e63b73b238134c/pkg/rpc/stats_handler.go#L28)):*

Context keys are usually types. This should be something like:

```
type remoteAddrKey struct{}
```

And then you use it as `remoteAddrKey{}`. This is better than what is currently here because stuffing a string into an interface requires an allocation. While stuffing an empty struct into an interface does not. 
</blockquote></details>

Done.

---

*[pkg/rpc/stats_handler.go, line 105 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyR6H2KnfYE0d_gQRtF:-KyRJ5kGesf31ocAAB-N:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f873ed72c9321e7a95cdd6c938e63b73b238134c/pkg/rpc/stats_handler.go#L105)):*

Should mention that we never remove items from this map. Doesn't seem like that will be a problem in practice, but something for a future reader to be aware of.
</blockquote></details>

Done.

---

*[pkg/server/status/status.proto, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyR6dMlR6XYhKWL34BQ:-KyRKF0T-KeXsBkhwX7w:b-j5n690) ([raw file](https://github.com/cockroachdb/cockroach/blob/f873ed72c9321e7a95cdd6c938e63b73b238134c/pkg/server/status/status.proto#L50)):*

These maps will get fairly large in large clusters. I wonder if that will be problematic. 

Can we change this to a single map from NodeID to <latency,incoming,outgoing>? Somewhat more efficient on the wire to do that. Not sure if there are backwards compatibility issues with this proto. Is it only used by the admin UI, or is it ever persisted to disk?
</blockquote></details>

Changed. The value is currently used in the debug section and the network component of the admin UI. The proto is written to the database every 10s by each node, FYI.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897)*

couchand:
:lgtm: modulo the lint issue

---

Reviewed 1 of 10 files at r1, 12 of 12 files at r2.

---

*[pkg/rpc/stats_handler_test.go, line 52 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWz1C20U1NsQy1ZaQr:-KyWz1C3LFwQL2wiDVRe:b-zci4qc) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/rpc/stats_handler_test.go#L52)):*
> ```Go
> 	sh.HandleRPC(ctx, &stats.InPayload{WireLength: 3})
> 	sh.HandleRPC(ctx, &stats.InTrailer{WireLength: 5})
> 	expResults["10.10.1.3:26257"].incoming += 15
> ```

It might be worth referencing the comment with your issue link here, just to make it clear why the math doesn't seem to work right.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 212 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWuQIXuADjwlVPz2qr:-KyWuQIXuADjwlVPz2qs:b-kjvmbs) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/ui/src/views/reports/containers/network/index.tsx#L212)):*
> ```XML
>       }
>       const a = nodesSummary.nodeStatusByID[nodeIDa].activity;
>       if (_.isNil(a) || _.isNil(a[nodeIDb])) {
> ```

Is this new check needed?  It looks like it's marked as not nullable in the proto...

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 215 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWuSlk-UOe1UYVPH-c:-KyWuSlk-UOe1UYVPH-d:bdqi60v) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/ui/src/views/reports/containers/network/index.tsx#L215)):*
> ```XML
> 	return <td key={key} className="network-table__cell network-table__cell--no-connection">
> 	  X
> 	</td>;
> ```

We use spaces, not tabs, in TypeScript code.  Strange that the linter didn't catch this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-:-KyX-JBVi1vbcOBerG0X:bm52a1y)*

spencerkimball:
Review status: 10 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/rpc/stats_handler_test.go, line 52 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWz1C20U1NsQy1ZaQr:-KyqiM0r3jLINK0-ZZ4A:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/rpc/stats_handler_test.go#L52)):*

It might be worth referencing the comment with your issue link here, just to make it clear why the math doesn't seem to work right.
</blockquote></details>

Done.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 212 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWuQIXuADjwlVPz2qr:-KyqiNmtj84KTfa8yBs9:borfl7u) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/ui/src/views/reports/containers/network/index.tsx#L212)):*

Is this new check needed?  It looks like it's marked as not nullable in the proto...
</blockquote></details>

The idea was to avoid an exception in the event that a new server showing this admin UI page wouldn't throw an exception if it encountered old data, where activity is nil.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 215 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWuSlk-UOe1UYVPH-c:-Kyqiu56uTRO1D3tUWHa:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/ui/src/views/reports/containers/network/index.tsx#L215)):*

We use spaces, not tabs, in TypeScript code.  Strange that the linter didn't catch this?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897)*

couchand:
Review status: 10 of 13 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/ui/src/views/reports/containers/network/index.tsx, line 212 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897#-KyWuQIXuADjwlVPz2qr:-KyqqJxqdfhZ8Aoa0mR8:bb6mt9b) ([raw file](https://github.com/cockroachdb/cockroach/blob/0f8baccbb9ba26a669fbf5229a95c2ec1d8fc771/pkg/ui/src/views/reports/containers/network/index.tsx#L212)):*

The idea was to avoid an exception in the event that a new server showing this admin UI page wouldn't throw an exception if it encountered old data, where activity is nil.
</blockquote></details>

:+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19897)*

PR #19900, opened 2017-11-08
storage: move many commands to `batcheval`

tbg:
This spares `endTransaction` which merits its own PR due to the commit
triggers.

It also does not yet move the span declaration methods, but nothing is
blocking that.

### DIFF
--- a/pkg/storage/batcheval/cmd_begin_transaction.go
+++ b/pkg/storage/batcheval/cmd_begin_transaction.go
@@ -1,0 +1,103 @@
+// Copyright 2014 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package batcheval
+
+import (
+	"fmt"
+
+	"golang.org/x/net/context"
+
+	"github.com/cockroachdb/cockroach/pkg/keys"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
+	"github.com/cockroachdb/cockroach/pkg/storage/batcheval/result"
+	"github.com/cockroachdb/cockroach/pkg/storage/engine"
+	"github.com/cockroachdb/cockroach/pkg/util/hlc"
+)
+
+// BeginTransaction writes the initial transaction record. Fails in
+// the event that a transaction record is already written. This may
+// occur if a transaction is started with a batch containing writes
+// to different ranges, and the range containing the txn record fails
+// to receive the write batch before a heartbeat or txn push is
+// performed first and aborts the transaction.
+func BeginTransaction(
+	ctx context.Context, batch engine.ReadWriter, cArgs CommandArgs, resp roachpb.Response,
+) (result.Result, error) {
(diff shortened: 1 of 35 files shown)


bdarnell:
:lgtm:, but if we're going to have a file per RPC I think the filenames should always match the RPC name (so I'd rather have `cmd_*_txn.go` instead of `cmd_txn_*.go`). Alternately, we could group "related" rpcs together into a handful of files (one file for the lease commands, one for txn stuff, one for basic KV ops...)

---

Reviewed 35 of 35 files at r1.

---

*[pkg/storage/batcheval/eval_txn_push.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900#-KyRc0Dfsd4qQ_g_aovw:-KyRc0Dfsd4qQ_g_aovx:bcy40l) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97d41114aab963daa8d0cc3e96803c8c3db9e75/pkg/storage/batcheval/eval_txn_push.go#L1)):*
> ```Go
> // Copyright 2014 The Cockroach Authors.
> ```

This file doesn't fit the pattern (`eval_` vs `cmd_`). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900#-:-KyRb_4SsfQe2Znt_DJw:besiwmm)*

nvanbenschoten:
Reviewed 15 of 35 files at r1.

---

*[pkg/storage/replica_command.go, line 63 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900#-KyRhda-GmORH2jOr2Zt:-KyRhda-GmORH2jOr2Zu:b-o22sck) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97d41114aab963daa8d0cc3e96803c8c3db9e75/pkg/storage/replica_command.go#L63)):*
> ```Go
> )
> 
> // A Command is the implementation of a single request within a BatchRequest.
> ```

Do you think the `Command` type and the `commands` map should be moved to the new `batcheval` package? That way, a new command doesn't even need to worry about this file. Conceptually, this means that a new command doesn't need to worry about how it will be evaluated by a `Replica`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900)*

tschottdorf:
TFTR!
I went for filename==RPC name for now.

---

Review status: 28 of 35 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/storage/replica_command.go, line 63 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900#-KyRhda-GmORH2jOr2Zt:-KyRxguQFnaRvICWpY76:b-58jni1) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97d41114aab963daa8d0cc3e96803c8c3db9e75/pkg/storage/replica_command.go#L63)):*

Do you think the `Command` type and the `commands` map should be moved to the new `batcheval` package? That way, a new command doesn't even need to worry about this file. Conceptually, this means that a new command doesn't need to worry about how it will be evaluated by a `Replica`.
</blockquote></details>

Yes, I definitely want to move these over, just haven't done so in this PR.

---

*[pkg/storage/batcheval/eval_txn_push.go, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900#-KyRc0Dfsd4qQ_g_aovw:-KyRyGjHn1lUereA9Uo9:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/b97d41114aab963daa8d0cc3e96803c8c3db9e75/pkg/storage/batcheval/eval_txn_push.go#L1)):*

This file doesn't fit the pattern (`eval_` vs `cmd_`). 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19900#-:-KyRyUC08rf7kGyHu5qN:b-fx7sr0)*

PR #19901, opened 2017-11-08
parser: move basic lexical functions to a new package 'lex'

knz:
### parser: move basic lexical functions to a new package 'lex' 

tl;dr: this patch moves the token and keyword definitions, the lexical
predicates (isIdentifier etc.) and the string / identifier escaping
functions into a new package `sql/lex`.

This is achieved as follows:

- the main functions from `encode.go` are exported, then `encode.go`
  is moved to `sql/lex`.

- the lexical predicates are extracted from `scan.go`, moved to
  `predicates.go`, exported, then moved to `sql/lex`.

- the keyword definitions are moved to `sql/lex` because they are
  needed by the lexical predicates and the encode functions. This is
  achieved by modifying `all_keywords.awk`, and
  `reserved_keywords.awk` to refer to the new `lex` package, and
  modifying `Makefile` accordingly. The generated files `keywords.go`
  and `reserved_keywords` are also moved accordingly.

- Also, since `keywords.go` uses the token constant definitions (the
  mapping of token name to numeric value used in the parser), these
  are moved to `lex` as well, by adding a new extraction rule in
  `parser/Makefile`. The generated `sql.go` is edited automatically to
  alias these token definitions from `lex`, instead of defining them
  in-place.

### sql/lex: simplify encodeEscapedSQLIdent

Moving this function to its own `lex` package reveals
it is not used elsewhere.

Its only two callers always call it with the bareIdentifiers flag set,
so there is really no need for the parameter.

Simplify the function accordingly.


petermattis:
Should the package be named `sqllex` instead of `lex`?

---

Review status: 0 of 29 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-:-KyRgGcgeTx7PQC19UKl:bg7oqob)*

jordanlewis:
It's already `sql/lex`, so I think that would be redundant? Of course you don't know that when you use stuff from the package in the code, though.

Raphael, there are a lot of other code movements in here, and potentially some functionality changes. I'm not exactly sure what I'm looking at. Can you summarize more clearly in the PR message? I was expecting purely a package move and I found some file rearrangement and other stuff as well.

---

Reviewed 32 of 32 files at r1.

---

*[pkg/sql/lex/encode.go, line 161 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-KyRn5c6X8fl3s-UfEVB:-KyRn5c6X8fl3s-UfEVC:b-usndsw) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7ae456fe2206e1ec3d56ca9964a46b52aff047d/pkg/sql/lex/encode.go#L161)):*
> ```Go
> // identifier is always quoted. Double quotes inside the identifier
> // are escaped.
> func encodeEscapedSQLIdent(buf *bytes.Buffer, s string) {
> ```

This seems like a code change?

---

*[pkg/sql/parser/Makefile, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-KyRnv7Ayl36SbvqQjRk:-KyRnv7Ayl36SbvqQjRl:b-8khqok) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7ae456fe2206e1ec3d56ca9964a46b52aff047d/pkg/sql/parser/Makefile#L28)):*
> ```Makefile
> 	 sed -E 's/^const ([A-Z][_A-Z0-9]*) =.*$$/const \1 = lex.\1/g') > $@
> 
> ../lex/tokens.go: gen/sql.go
> ```

What's going on here? This guy is getting complicated! Time to document these rules?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-:-KyRmicM0VFRN9dvyAIZ:bdga6z5)*

petermattis:
> It's already sql/lex, so I think that would be redundant? Of course you don't know that when you use stuff from the package in the code, though.

External references see `lex.Foo`. If we every need a lexer for a different language, what package does that go in? 

I don't feel strongly about this, just pointing out a limitation in Go package hierarchies. We can ignore for now and deal with a collision in the future if it ever occurs.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-:-KyRp-BlcU6LYcMJstfk:b-dgtn0v)*

knz:
I have edited the commits messages to be more descriptive. Also extracted the functionality change into its own commit. PTAL.

(To respond to the other concern: I have chosen `lex` instead of `sqllex` to reduce stutter, and reduce the amount of typing in the points where the package is used. We do not currently have any package name conflict here.)

---

Review status: 27 of 29 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/lex/encode.go, line 161 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-KyRn5c6X8fl3s-UfEVB:-KySOUf_PCwY8omv-kM6:b38nsxo) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7ae456fe2206e1ec3d56ca9964a46b52aff047d/pkg/sql/lex/encode.go#L161)):*

This seems like a code change?
</blockquote></details>

Reverted in this commit, added a separate commit to clarify.

---

*[pkg/sql/parser/Makefile, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-KyRnv7Ayl36SbvqQjRk:-KySOxw-S0SarN3d1Opp:b6wydsl) ([raw file](https://github.com/cockroachdb/cockroach/blob/d7ae456fe2206e1ec3d56ca9964a46b52aff047d/pkg/sql/parser/Makefile#L28)):*

What's going on here? This guy is getting complicated! Time to document these rules?
</blockquote></details>

Added comments to document.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19901#-:-KySR_Nw6hhuK67TIh5K:b-4hh25o)*

knz:
TFYR! 🎉

PR #19908, opened 2017-11-08
ui: flatten cluster summary sections

### LINKED ISSUE #19902: ui: cluster overview layout issue
The cluster overview uses `display: contents`, which only works on Chrome behind a flag.  Let's just get rid of that for now.

display-contents

couchand:
Because Chrome's support for display: contents is still behind a flag,
we can't rely on it to handle flattening the markup for the cluster
overview.  This is uglier than it would be otherwise, because we're not
on React 16 yet, so we can't just have each section's component return
the arrays of elements.

Fixes #19902.

### DIFF
--- a/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl
+++ b/pkg/ui/src/views/cluster/containers/clusterOverview/cluster.styl
@@ -22,65 +22,62 @@
     grid-template-rows repeat(3, auto)
     grid-template-areas "cap-t cap-t . live-t live-t live-t . rep-t rep-t rep-t ." "cap-m cap-c . live-a live-b live-c . rep-a rep-b rep-c ." "cap-a cap-a . live-1 live-2 live-3 . rep-1 rep-2 rep-3 ."
 
-    &__section
-      display contents
+    .capacity-usage
+      &.cluster-summary__title
+          grid-area cap-t
 
-      &.capacity-usage
-        .cluster-summary__title
-            grid-area cap-t
+      &.cluster-summary__chart
+          grid-area cap-c
 
-        .cluster-summary__chart
-            grid-area cap-c
+      &.cluster-summary__metric
+          grid-area cap-m
 
-        .cluster-summary__metric
-            grid-area cap-m
+      &.cluster-summary__aside
+          grid-area cap-a
 
-        .cluster-summary__aside
-            grid-area cap-a
+    .node-liveness
+      &.cluster-summary__title
+          grid-area live-t
 
-      &.node-liveness
-        .cluster-summary__title
-            grid-area live-t
+      &.cluster-summary__metric.live-nodes
+          grid-area live-a
 
-        .cluster-summary__metric.live-nodes
-            grid-area live-a
+      &.cluster-summary__metric.suspect-nodes
+          grid-area live-b
 
-        .cluster-summary__metric.suspect-nodes
-            grid-area live-b
+      &.cluster-summary__metric.dead-nodes
+          grid-area live-c
 
-        .cluster-summary__metric.dead-nodes
-            grid-area live-c
+      &.cluster-summary__label.live-nodes
+          grid-area live-1
 
-        .cluster-summary__label.live-nodes
-            grid-area live-1
+      &.cluster-summary__label.suspect-nodes
+          grid-area live-2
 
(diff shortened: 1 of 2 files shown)


tschottdorf:
Can't speak to the code changes, but this fixes it:

image

though I'm not sure I find that bar very intuitive.

couchand:
Glad to hear the layout issue is fixed.  Looking at it again, I think you're right, it's odd that the total capacity (which is the size of the grey bar) doesn't actually appear anywhere explicit.  I'll make a note over on #19512 about that.

vilterp:
:lgtm: with a small style comment

---

Reviewed 2 of 2 files at r1.

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19908#-KySoIFa0IzZFF4R-iaG:-KySoIFb95kS5DHpQ13q:boich68) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7dd6749462df8dec5a2b98488bc7a2ceba17863/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L139)):*
> ```XML
>     ];
>     return <section className="cluster-summary" children={children} />;
>   }
> ```

Passing children as a prop seems un-idiomatic. Would prefer

```js
<section className="cluster-summary">
  {renderCapacityUsage(this.props.capacityUsage)}
  {renderNodeLiveness(this.props.nodeLiveness)}
  {renderReplicationStatus(this.props.replicationStatus)}
</section>
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19908#-:-KySoegelwuwlqCiTLy6:bxarigd)*

couchand:
> Hah, now i'm really turned off by the idea of the grid if we have to flatten the DOM like this. 

I was worried you'd say that.  Yeah, it's ugly.  I agree it's annoying to have to have the DOM flat (that's why `display: contents` exists, after all, it's just still in WD stage).

> You can check this in to fix what i'm assuming is a total breakage, but it should really be revisited.

I'll go ahead and merge this to fix the total breakage for `supports(grid) - supports(contents)`, a set which includes Chrome (without the experimental features flag) and Safari, but you're right we shouldn't leave it there.

> For example, can you nest grids? Like, can you place the three sections using a grid, then within each section use a grid to align the elements in the section?

Not currently.  That's a possible extension of the spec they've done some work on, but not currently supported anywhere, and it doesn't look like it's coming any time soon.  As it is, if you nest grids they each get independent sizings, so there's no good way to make things line up across the subgrids.

I've been putting thought into how to make this work, and I think you're right that it involves something more like the first implementation with nested flexboxen.  Flex does the horizontal layout, and the vertical alignment is just done by assigning explicit pixel heights and offsets to everything.

The thing that makes the grid really compelling for something like this is media queries: we can redefine the named areas for different sizes of viewports to layout the same content in appropriate ways, and it only requires one rule for each media query (the `grid-template-areas` rule).  BTW, that `grid-template-areas` rule would look much nicer if stylus allowed the values to run over multiple lines (maybe you can?  I find the documentation there very lacking).

---

---

*[pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19908#-KySoIFa0IzZFF4R-iaG:-KyTBga68ohXOO_0HRxW:bl6k3lj) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7dd6749462df8dec5a2b98488bc7a2ceba17863/pkg/ui/src/views/cluster/containers/clusterOverview/index.tsx#L139)):*

Passing children as a prop seems un-idiomatic. Would prefer

```js
<section className="cluster-summary">
  {renderCapacityUsage(this.props.capacityUsage)}
  {renderNodeLiveness(this.props.nodeLiveness)}
  {renderReplicationStatus(this.props.replicationStatus)}
</section>
```
</blockquote></details>

The problem with that is React then wants you to add a key prop to ever child, which IMO is uglier.  Children are passed as a prop regardless of if you use the sugar for it.

That being said, this will go back to components per the discussion with @mrtracy.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19908#-:-KyTC3gOTxxj7oR223Qo:b22ai7o)*

PR #19913, opened 2017-11-08
WIP: Add CANCEL JOB support for schema changes

lgo:
This adds the ability to cancel schema changes, using:

    # Start schema change, e.g.
    CREATE INDEX ...
    CANCEL JOB <job_id>

This will cause the session running the schema change to have an error
returned. NB: schema changes following the canceled one will still run
in the background.


lego:
cc: @vivekmenezes I'd like a second opinion on this.

Currently, this is failing because when you run the statement `CANCEL JOB ...` it will go to cancel the schema change and correctly start a new rollback schema change job, but then the cancel routine will run a second time immediately after on the new rollback job.

I haven't been able to figure out why, but it may be due to the direct call to `reverseMutations` made in `cancelSchemaChange` doing something weird. As far as I can tell, this should not be causing that problem though.

Of course, because the cancel routine is only creating the schema change job there will need to be changes to the `SchemaChangeManager` call to `SchemaChanger.exec` to make sure to not rollback a rollback (sidenote: I believe this can happen right now if a rollback is happening on a node and the node dies. another node should pick it up and do exactly this)

mjibson:
I think it's starting the second job because you remove `jobs.WithoutCancel`, which means the registry will attempt to resume the job if the original coordinator node has a liveness problem (which is presumably something in one of the tests). But since the schema changer already does that, it happens twice.

I suggest you hold off (and definitely do not merge anything) on any changes to jobs until my refactor is done, and then we can work together to get this done.

lego:
Ah, yea. I assumed that onFail would need to only happen if the job was job controllable (i.e. has a lease) so I popped that check where it calls `onFail` and changed the `jobs.WithoutCancel` without thinking about the implications on that. Both of those changes are definitely not needed, so that might be the problem :).

(It's probably just trying to cancel it due to the liveliness, as it is unable to resume it as there's no resumeHook).

Sounds good to me, definitely keep me in the loop with those changes!

lego:
Closing, as this change now needs to be done in combination with pause/resume after the refactor. See https://github.com/cockroachdb/cockroach/issues/16018#issuecomment-351185206 for a description of what's needed to do this.

Issue #19882, opened 2017-11-07
distsql: node shutdown/failure results in query returning an internal error


asubiotto:
Example: On a 3-node cluster, running a distsql query through node 1 that schedules some processors on node 2 and 3 and then shutting down node 2 results in:
```
pq: query execution canceled: [n1] communication error: rpc error: code = Canceled desc = context canceled
```
For now, we should be attempting to retry the query from the gateway node, scheduling the flows on the gateway node, and then returning a retryable error to the client if all else fails. However, better solutions like dynamically rescheduling processors should be explored in the future.

asubiotto:
Created above PR (#19919) as a stop gap solution. Closing this issue as future work is outlined in #15637

Issue #15637, opened 2017-05-03
distsql: better handling of failures and dead nodes


RaduBerinde:
In 1.0, we avoid hosts that known to be dead, but we may fail the first attempt of a query if we don't know the host is dead.

This issue tracks implementing better handling of these situations. The idea is to support multiple flows for the same query on a single node, and to reschedule flows (which we failed to schedule) on the gateway node.

One important aspect of "error" in this discussion is version mismatch: if we can reschedule flows that can't run on a host because of its version, we can provide query availability (with decreased performance) during an upgrade even when there are incompatible changes in distsql.

cuongdo:
how is this looking for 1.1?

RaduBerinde:
We have a good idea of what we want to do here but nothing is implemented at this point.

arjunravinarayan:
This is being addressed by #17497, which should merge in the 1.1 timeframe.

andreimatei:
The version mismatched problem has been address some. The node failed recently case remains open.

vivekmenezes:
Probably obvious but this is a high priority issue for us to fix.

vivekmenezes:
radu and andrei are thinking about this problem and trying to figure out what to do in the next release.

nstewart:
This issue is preventing me from doing standard reporting. Is there an update? I'm getting `psycopg2.InternalError: [n3] communication error: rpc error: code = Canceled desc = context canceled` and I see in my 5 node cluster a couple nodes restarted. Happy to create a new ticket but I see #19882 was closed in favor of this one.

jordanlewis:
@nstewart, does the query complete if you try it again? Or are you never able to complete your query?

nstewart:
My script runs ~10 queries serially. Each query runs for about several minutes (but not hours). I've run the script several times over the weekend and it fails at queries places with `context canceled`. 

I'm betting if I added retry loops this thing would eventually complete (I'm going to do this in the short term so I'm not blocked), but since these aren't serializable retries and there is only one, single-threaded client for this database, I wonder if that would be masking a bigger problem.

jordanlewis:
Yes, surely doing that would be masking a problem. During your script's run, are nodes restarting a lot?

nstewart:
Yes, I am running a 5 node cluster and I saw 1 node restart once and another restart 4 times. I saved the debug zip here: https://github.com/cockroachdb/cockroach/issues/28278

PR #19919, opened 2017-11-08
[WIP] distsql: retry on communication errors

asubiotto:
Addresses #19882

This is a stop-gap solution to retry on certain errors seen when a node that
participates in a distsql query goes down or is shutdown to not have the
onus be on the client to retry, especially since these errors are not
returned as retriable errors. There is more work to be done here,
tracked in #15637.


asubiotto:
To be cherry-picked into 1.1 as customers are having issues with this

asubiotto:
Closing PR. Will have another PR that solves Rubrik's problem without retries.

PR #19928, opened 2017-11-08
log: report hostname when COCKROACH_REPORT_SENSITIVE_DETAILS=true

tbg:
This environment variable will be set on our internal clusters and nightly test clusters
to allow easier identification.

### DIFF
--- a/pkg/util/log/crash_reporting.go
+++ b/pkg/util/log/crash_reporting.go
@@ -66,6 +66,12 @@ var (
 	// startTime records when the process started so that crash reports can
 	// include the server's uptime as an extra tag.
 	startTime = timeutil.Now()
+
+	// ReportSensitiveDetails enables reporting of unanonymized data.
+	//
+	// This should not be used by anyone unwilling to share the whole cluster
+	// data with Cockroach Labs and various cloud services.
+	ReportSensitiveDetails = envutil.EnvOrDefaultBool("COCKROACH_REPORT_SENSITIVE_DETAILS", false)
 )
 
 // RecoverAndReportPanic can be invoked on goroutines that run with
@@ -341,10 +347,12 @@ func SendCrashReport(
 
 	ex := raven.NewException(err, raven.NewStacktrace(depth+1, contextLines, crdbPaths))
 	packet := raven.NewPacket(err.Error(), ex)
-	// Avoid leaking the machine's hostname by injecting the literal "<redacted>".
-	// Otherwise, raven.Client.Capture will see an empty ServerName field and
-	// automatically fill in the machine's hostname.
-	packet.ServerName = "<redacted>"
+	if !ReportSensitiveDetails {
+		// Avoid leaking the machine's hostname by injecting the literal "<redacted>".
+		// Otherwise, raven.Client.Capture will see an empty ServerName field and
+		// automatically fill in the machine's hostname.
+		packet.ServerName = "<redacted>"
+	}
 	tags := map[string]string{
 		"uptime": uptimeTag(timeutil.Now()),
 	}
--- a/pkg/util/log/crash_reporting_packet_test.go
+++ b/pkg/util/log/crash_reporting_packet_test.go
@@ -108,8 +108,11 @@ func TestCrashReportingPacket(t *testing.T) {
 		p := packets[0]
 		packets = packets[1:]
 		t.Run("", func(t *testing.T) {
-			if e, a := "<redacted>", p.ServerName; e != a {
-				t.Errorf("expected ServerName to be '<redacted>', but got '%s'", a)
+			if !log.ReportSensitiveDetails {
+				e, a := "<redacted>", p.ServerName
+				if e != a {
+					t.Errorf("expected ServerName to be '<redacted>', but got '%s'", a)
+				}
 			}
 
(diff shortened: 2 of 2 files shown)


pkg/util/log/crash_reporting.go
```diff
@@ -66,6 +66,12 @@ var (
 	// startTime records when the process started so that crash reports can
 	// include the server's uptime as an extra tag.
 	startTime = timeutil.Now()
+
+	// ReportUnanonymized enables reporting of unanonymized data.
+	//
+	// This should not be used by anyone unwilling to share the whole cluster
+	// data with Cockroach Labs and various cloud services.
+	ReportUnanonymized = envutil.EnvOrDefaultBool("COCKROACH_REPORT_UNANONYMIZED", false)
```
mberhault: unanymized.. nonymized... anon.. ugh! this is a mouthful. How about "uncensored", "unredacted", "..._SENSITIVE_DETAILS`, etc...

mberhault (approved):

pkg/util/log/crash_reporting.go
tschottdorf: The first isn't correct, the second one isn't clear enough, I could do the third.

mberhault: I think the third is clear and describes the danger pretty well. Bonus points for being easier to say.

dt:
random idea: as long as we're adding a env var, we could make it a string instead with just a default value `<redacted>`, and then we can explicitly set it to just what we want without needing to worry about the hostname?

---

---

*[pkg/util/log/crash_reporting.go, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-KySmB-wCytN8l_4-NOn-r1-74:-KyW7kSE84ItboipTIse:brfqkpc) ([raw file](https://github.com/cockroachdb/cockroach/blob/343a3b4adad0c1889c457726a7156213347b3b23/pkg/util/log/crash_reporting.go#L74)):*
<details><summary><i>Previously, mberhault (marc) wrote…</i></summary><blockquote>

I think the third is clear and describes the danger pretty well. Bonus points for being easier to say.
</blockquote></details>

even more bike shed: I think the wording is too broad -- the diagnostic reporting is certainly still anonymized of scheme strings and such. maybe `COCKROACH_CRASH_REPORTING_NAME` or something?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-:-KyW71BSV2kV1uxqmkP7:b-khr5lj)*

tschottdorf:
I considered that, but then it needs to be set to something useful and it's more likely than not that during copy-pasting we'll just have conflicting values. Hostname is usually clear and it can't be messed up during setup.

---

---

*[pkg/util/log/crash_reporting.go, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-KySmB-wCytN8l_4-NOn-r1-74:-KyW8Ro11QVyvP701tQu:bhwodii) ([raw file](https://github.com/cockroachdb/cockroach/blob/343a3b4adad0c1889c457726a7156213347b3b23/pkg/util/log/crash_reporting.go#L74)):*

even more bike shed: I think the wording is too broad -- the diagnostic reporting is certainly still anonymized of scheme strings and such. maybe `COCKROACH_CRASH_REPORTING_NAME` or something?
</blockquote></details>

My thinking here was that we may want to deanonymize more in the future, and that under this env var we would be continually free to do so. For example, if we want to report unredacted errors, we don't want to go through all of our infra and rename all vars. This is the "don't expect any privacy" flag. I think that's most useful given the intended usage.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-:-KyW8eXPUYVdgyqjBv6g:b-yp4hzh)*

dt:
:thumbsup: 

---

---

*[pkg/util/log/crash_reporting.go, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-KySmB-wCytN8l_4-NOn-r1-74:-KyW9-hbecFY0PhwXJJJ:b4c9lyr) ([raw file](https://github.com/cockroachdb/cockroach/blob/343a3b4adad0c1889c457726a7156213347b3b23/pkg/util/log/crash_reporting.go#L74)):*

My thinking here was that we may want to deanonymize more in the future, and that under this env var we would be continually free to do so. For example, if we want to report unredacted errors, we don't want to go through all of our infra and rename all vars. This is the "don't expect any privacy" flag. I think that's most useful given the intended usage.
</blockquote></details>

:thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-:-KyW91KOIGAjglFUlKqC:b4c9lyr)*

pkg/util/log/crash_reporting.go
tschottdorf: Yup, done.

tschottdorf:
TFTRs!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19928#-:-KyWAY1SHYlWRT9WJhXN:bz5mair)*

Issue #15849, opened 2017-05-10
stability: Support for large transactions


a-robinson:
Forked from @andreimatei's comment on #15713:

> Unfortunately the "transaction is too large to commit" is by design - currently an EndTransaction carries all the intents, and so we need to be reasonably sure that each replica will have enough memory to process the corresponding Raft command.
> 
> This is a topic that I've randomly discussed at the watercooler with the likes of @bdarnell and @tschottdorf at times. Dumping stuff here for institutional memory.
> There are various things we could do: one way or another, intents (or, at least, all intents) shouldn't have to be attached to an EndTransaction request. They could be stored in the db and the cleanup happening after EndTransaction has marked the txn record as committed could retrieve them from there. For example, they could be stored in the transaction record itself. The TxnCoordSender could keep appending intent spans when it heartbeats the record, for example. That might not be sufficient, since you'll end up with a humongous txn record. What'd be better is to store them adjacently to the txn rec, such that the range can be split, etc. But then you prolly need to be careful with the orchestration of the cleanup process and the cleanup of these intent records themselves.
> 
> Without, or besides, storing the intents into the db, the client (the client.Txn layer) could be more involved in the cleanup of such large txns; it could store intents in memory and not attach them to EndTransaction, but do cleanup afterwards. It could keep the intents below a memory budget by having a limited-size interval tree and merging spans when it needs to - trading pessimistic scanning at cleanup time for memory at execution time. This would probably work great for cases as the one here (I assume): copying a table.
> An approach like this would probably be made more feasible by the merging of the TxnCoordSender (which currently keeps track of intents) and client.Txn that I'm hoping to do.
> 
> @petermattis, I'm assigning back to you... But I will keep an eye on supporting work here when I do TxnCoordSender refactorings.
> 
> Incidentally - here's one idea about the codelab we were talking about: implement CREATE TABLE AS <query>, or a version of INSERT INTO... SELECT... that does some limited version of the above - doesn't accumulate intents and scans the whole table for cleanup.

andreimatei:
Note that it's exceedingly easy to get yourself into one of \`em "large transactions" by doing something like a `INSERT INTO... SELECT ...` (#15713) or a `TRUNCATE` (#15723) or through the GC queue (https://github.com/cockroachdb/cockroach/issues/15756#issuecomment-299924227) - but this last one may be FUD.

tschottdorf:
[...]

PR #19929, opened 2017-11-09
tscache: enforce minimum retention window policy in intervalSkl

nvb:
This change adds a retention policy to `intervalSkl` that it should never evict
pages that contain intervals within a certain window of recency. This will allow
the structure to exhibit the `MinRetentionWindow` semantics that `tscache.Cache`
implementations require.

We do this by replacing the `earlier` and `later` `sklPages` with a linked list
of `sklPages` that can dynamically grow when necessary.

See https://github.com/cockroachdb/cockroach/pull/19508#issuecomment-339337357 for more discussion.

### DIFF
--- a/pkg/storage/tscache/interval_skl.go
+++ b/pkg/storage/tscache/interval_skl.go
@@ -140,12 +140,12 @@ type intervalSkl struct {
 	clock  *hlc.Clock
 	minRet time.Duration
 
-	// Soft maximum size of the data structure in bytes. When the data structure
-	// fills, older entries are discarded. However, this maximum can be violated
-	// if the intervalSkl needs to grow larger to enforce a minimum retention
-	// policy. As such, the minimum retention policy takes priority over the
-	// maximum size policy (which is why we call this a "soft" maximum).
-	size uint32
+	// The size of each page in the data structure, in bytes. When a page fills,
+	// the pages will be rotated and older entries will be discarded. The entire
+	// data structure will usually have a size limit of pageSize*minPages.
+	// However, this limit can be violated if the intervalSkl needs to grow
+	// larger to enforce a minimum retention policy.
+	pageSize uint32
 
 	// The linked list maintains fixed-size skiplist pages, ordered by creation
 	// time such that the first page is the one most recently created. When the
@@ -168,11 +168,11 @@ type intervalSkl struct {
 
 // newIntervalSkl creates a new interval skiplist with the given minimum
 // retention duration and the maximum size.
-func newIntervalSkl(clock *hlc.Clock, minRet time.Duration, size uint32) *intervalSkl {
+func newIntervalSkl(clock *hlc.Clock, minRet time.Duration, pageSize uint32) *intervalSkl {
 	s := intervalSkl{
 		clock:    clock,
 		minRet:   minRet,
-		size:     size,
+		pageSize: pageSize,
 		minPages: defaultMinSklPages,
 	}
 	s.pushNewPage(0)
@@ -334,7 +334,7 @@ func (s *intervalSkl) frontPage() *sklPage {
 
 // pushNewPage prepends a new empty page to the front of the pages list.
 func (s *intervalSkl) pushNewPage(maxWallTime int64) {
-	p := newSklPage(s.size / uint32(s.minPages))
+	p := newSklPage(s.pageSize)
 	p.maxWallTime = maxWallTime
 	s.pages.PushFront(p)
 }
--- a/pkg/storage/tscache/interval_skl.go
+++ b/pkg/storage/tscache/interval_skl.go
@@ -16,9 +16,11 @@ package tscache
 
 import (
 	"bytes"
+	"container/list"
 	"encoding/binary"
 	"fmt"
 	"sync/atomic"
+	"time"
 	"unsafe"
 
 	"github.com/andy-kimball/arenaskl"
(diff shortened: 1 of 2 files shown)


petermattis:
:lgtm: though @tschottdorf and @andy-kimball should take a look as well.

What default are we going to use for the skip-list page size? 

---

---

*[pkg/storage/tscache/interval_skl.go, line 148 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-KyWGE64Ba-j_GJrqkcI:-KyWGE65LDb3EKIPFQ2t:bhfp860) ([raw file](https://github.com/cockroachdb/cockroach/blob/ed6d1e0414bb141752385de06370a9a5e0c9c01e/pkg/storage/tscache/interval_skl.go#L148)):*
> ```Go
> 	// policy. As such, the minimum retention policy takes priority over the
> 	// maximum size policy (which is why we call this a "soft" maximum).
> 	size uint32
> ```

Perhaps rename to `softMaxSize`.

---

*[pkg/storage/tscache/interval_skl.go, line 163 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-KyWGTKJhexKHgsZFyyZ:-KyWGTKJhexKHgsZFyy_:b-91241) ([raw file](https://github.com/cockroachdb/cockroach/blob/ed6d1e0414bb141752385de06370a9a5e0c9c01e/pkg/storage/tscache/interval_skl.go#L163)):*
> ```Go
> 	// In order to ensure that timestamps never decrease, intervalSkl maintains
> 	// a floor timestamp, which is the minimum timestamp that can be returned by
> 	// the lookup operations. When the earlier page is discarded, its current
> ```

`s/earlier/earliest/g`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-:-KyWHDIROXSvy_9e93H-:b-3ec3xu)*

bdarnell:
LGTM, but once we've tested and stabilized the new TS cache we should probably have a debate about whether this potentially-unbounded growth is even something we want. In the long run we want every component to be able to regulate its memory usage; I think it would be best for the ts cache to have a fixed size even if that may limit the lifetime of transactions (or if we do want the ts cache to grow dynamically we have to identify something else we can dynamically shrink that won't lead to cascading problems).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-:-KyWcPhnlmwcRulklOTS:brh1ejq)*

petermattis:
> I think it would be best for the ts cache to have a fixed size even if that may limit the lifetime of transactions (or if we do want the ts cache to grow dynamically we have to identify something else we can dynamically shrink that won't lead to cascading problems).

We can dynamically adjust the size of the RocksDB cache. With the large blocks of allocation in the new tscache impl it seems feasible to adjust the size of the RocksDB cache up and down in response to the tscache size.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-:-KyWel2Ash_ggrgLEQBq:bsopo5h)*

nvanbenschoten:
> What default are we going to use for the skip-list page size?

I was planning on making each page 64MB, which is the current size for the entire cache (but which seems small).

> or if we do want the ts cache to grow dynamically we have to identify something else we can dynamically shrink that won't lead to cascading problems

I'm not sure exactly how this works in other production systems, but at some point, I've always pictured us creating a system-wide memory policy for how memory bounds of different components grow, shrink, or stay constant in relation to each other as load increases. For lack of a better analogy, think of how flexbox bounds total size while allowing individual elements to dynamically adjust in size.

---

---

*[pkg/storage/tscache/interval_skl.go, line 148 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-KyWGE64Ba-j_GJrqkcI:-KyWk-2usJSdPVdElDtP:b-lofyqv) ([raw file](https://github.com/cockroachdb/cockroach/blob/ed6d1e0414bb141752385de06370a9a5e0c9c01e/pkg/storage/tscache/interval_skl.go#L148)):*

Perhaps rename to `softMaxSize`.
</blockquote></details>

I actually switched this to `pageSize` because it's easier to think about.

---

*[pkg/storage/tscache/interval_skl.go, line 163 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-KyWGTKJhexKHgsZFyyZ:-KyWjzqZw_6ExiDTE--s:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/ed6d1e0414bb141752385de06370a9a5e0c9c01e/pkg/storage/tscache/interval_skl.go#L163)):*

`s/earlier/earliest/g`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-:-KyWiYAqOnGzFFzkJ5PO:bvcf23e)*

tschottdorf:
@cardbot this is similar to what's discussed [https://github.com/cockroachdb/cockroach/issues/15849#issuecomment-309504798](here) in the context of large transactions. (Pointing that out mostly to link these two issues).

:lgtm: 

---

Reviewed 2 of 2 files at r3, 1 of 1 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19929#-:-KyYYddzY5Sd6SDZCP4u:b-ulwkav)*

nvanbenschoten:
Thanks for the reviews.

> I suspect that eventually some kind of "span coalescing" will be needed, in which long-running transactions lay claim to larger and larger spans which take less memory to represent (but which also can block other transactions that overlap with those spans).

Are there benefits to this kind of span coalescing strategy (and also the lock promotion policy) other than reducing a transaction's footprint over time as it performs more and more operations? I ask because I've traditionally thought of this kind of retention policy as being in place for long running transactions, not necessarily for large transactions. I guess those two will often go hand in hand. Regardless, I agree with @tschottdorf that this kind of approach has applications in multiple areas of our transaction system and foresee them being used in the future.

PR #19922, opened 2017-11-08
sql/parser: split into more separate packages

knz:
First commit from #19901 - will rebase when that merges.

tl;dr this patch separates items from the `sql/parser` package into
the following:

- new package `sem/builtins`: the built-in function *definitions*

- new package `sem/transforms`: `ContainsWindow`,
  `IsAggregateVisitor`, `ExprTransformContext` (not
  `NormalizeVisitor`, that's still entangled with `parser`)

- new package `sem/tests`: tests that need both the parser and the
  built-in function definitions, to break the dependency cycle.

Details follow.

The following `parser` definitions remain in the `parser` package for
now but are moved to their own files:

- `FunctionClass` and constants, `Builtin` and methods (`builtins.go` -> `builtin.go`)
- `AggregateFunc` (`aggregate_builtins.go` -> `aggregate_funs.go`)
- `ValueGenerator` (`generator_builtins.go` -> `generators.go`)
- `IndexedRow`, `WindowFrame`, `WindowFunc` (`window_builtins.go` -> `window_funs.go`)
- `intPow` (`builtins.go` -> `eval.go`)
- `pickFromTuple` (`builtins.go` -> `eval.go`)
- `mockNameTypes` (`type_check.go` -> `testutils.go`)
- `init`, (`builtins.go` -> `all_builtins.go`, then moved below)
- `collectBuiltins` (`window_builtins.go` -> `all_builtins.go`, then moved below)

- `TestCastToCollatedString` (`eval_test.go` -> `collatedstring_test.go`)
- `TestEvalComparisonExprCaching` (`eval_test.go` -> `compare_test.go`)
- `TestSimilarEscape`, `Benchmark*Like*` (`eval_test.go` -> `like_test.go`)
- `TestClusterTimestampConversion` (`eval_test.go` -> `timestamp_test.go`)

This patch then extracts code into the following separate modules:

- `canBeInArray` -> `IsValidArrayElementType` in `sql/sem/types`

- `ContainsWindowVisitor` -> `window.go` in `sql/sem/transform`
- `IsAggregateVisitor` -> `aggregates.go` in `sql/sem/transform`
- `expr_transform.go` -> `sql/sem/transform`

- `all_builtins.go`, `builtins.go` -> `sql/sem/builtins`
- `aggregate_builtins.go` -> `sql/sem/builtins`
- `aggregate_builtins_test.go` -> `sql/sem/builtins`
- `window_builtins.go` -> `sql/sem/builtins`
[...]

PR #19941, opened 2017-11-09
sql/parser: extract code in different packages

knz:
Supersedes #19922.

tl;dr:
- new package `sem/builtins` with built-in function definitions
- new package `sem/transform` with the expression visitors
- new package `sem/tests` for tests that depend both on `parser` and `builtins`

All split up in miniature commits, for your (re)viewing pleasure. :)


pkg/sql/parser/expr_test.go
```diff
@@ -151,7 +151,7 @@ func TestExprString(t *testing.T) {
 		`(a OR (g BETWEEN (h+i) AND (j+k))) AND b`,
 		`(1 >= 2) IS OF (BOOL)`,
 		`(1 >= 2) = (2 IS OF (BOOL))`,
-		`count(1) FILTER (WHERE true)`,
+		// `count(1) FILTER (WHERE true)`,
```
jordanlewis: Second to last commit does this for some reason, probably a mistake

jordanlewis (approved):

jordanlewis:
LGTM! :tada: (Reviewable and GitHub's review view are both unusably slow on this size of PR so I can't leave top level comments there.)

Overall I'm *very* happy with this new structure. Thank you for taking on this large burden! I admit to not having read every single line of code change, but I feel confident for the record that this is 98% restructuring.

I saw that you added new exported fields to `Builtin`, but left the accessor methods unexported. This seems backwards - if there's a good reason please chime in but if not I'd prefer to see the opposite done.

Also if we end up doing much more iteration on this PR please split this PR into one before the builtins change and one containing just the builtins change.

I would suggest adding a readme to the new `sem/tests` directory explaining what you did in the commit message.

I think our test structuring could still use some work - it's always been hard to figure out where a test lives or where a new one should live, and it's not clear that it is getting easier - but that's not really the responsibility of this PR to fix. Just putting it for the record.

knz:
TFYR!

The Builtin methods you have seen are only visible to TypeCheck, and thus need not be exported.

PR #19948, opened 2017-11-09
cli,cliccl: add tests for setting index/partition zone configs

benesch:
The underlying support for index/partition zone configs landed in #19854.
This commit merely adds tests that setting these zone configs from the
CLI works as expected.

### DIFF
--- a/pkg/ccl/cliccl/cli_test.go
+++ b/pkg/ccl/cliccl/cli_test.go
@@ -1,0 +1,199 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed as a CockroachDB Enterprise file under the Cockroach Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+//     https://github.com/cockroachdb/cockroach/blob/master/LICENSE
+
+package cliccl
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/cockroachdb/cockroach/pkg/base"
+	"github.com/cockroachdb/cockroach/pkg/cli"
+	"github.com/cockroachdb/cockroach/pkg/server"
+	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
+	"golang.org/x/net/context"
+)
+
+// cliTest is a stripped-down version of package cli's cliTest. It's currently
+// easier to reimplement the pieces we need than it is to export cli.cliTest, as
+// cli's tests frequently mutate internal package variables. We may need to
+// revisit this decision if this package's tests become more complex.
+type cliTest struct {
+	*server.TestServer
+	connArgs []string
+}
+
+func newCLITest() cliTest {
+	s, err := serverutils.StartServerRaw(base.TestServerArgs{Insecure: true})
+	if err != nil {
+		panic(err)
+	}
+	host, port, err := net.SplitHostPort(s.ServingAddr())
+	if err != nil {
+		panic(err)
+	}
+	return cliTest{
+		TestServer: s.(*server.TestServer),
+		connArgs:   []string{"--insecure", "--host=" + host, "--port=" + port},
+	}
+}
+
+func (c *cliTest) close() {
+	c.Stopper().Stop(context.Background())
+}
+
+func (c *cliTest) run(line string) {
+	c.runWithArgs(strings.Fields(line))
+}
+
+func (c *cliTest) runWithArgs(args []string) {
+	fmt.Println(strings.Join(args, " "))
+	cli.TestingReset()
+	if err := cli.Run(append(args, c.connArgs...)); err != nil {
+		fmt.Println(err)
+	}
+}
+
+func Example_cclzone() {
+	c := newCLITest()
+	defer c.close()
+
+	c.runWithArgs([]string{"sql", "-e", "CREATE DATABASE db"})
+	c.runWithArgs([]string{"sql", "-e", `CREATE TABLE db.t (
+  c1 STRING PRIMARY KEY,
+  c2 STRING
+) PARTITION BY LIST (c1) (
+  PARTITION p0 VALUES ('a'),
+  PARTITION p1 VALUES (DEFAULT)
+)`})
+	c.runWithArgs([]string{"sql", "-e", "CREATE INDEX ON db.t (c2)"})
+	c.run("zone set db.t@nonexistent --file=./../../cli/testdata/zone_attrs.yaml")
+	c.run("zone set db.t.nonexistent --file=./../../cli/testdata/zone_attrs.yaml")
+	c.run("zone set db.t.p0@t_c2_idx --file=./../../cli/testdata/zone_attrs.yaml")
+	c.run("zone set db.t@primary --file=./../../cli/testdata/zone_attrs.yaml")
+	c.run("zone get db.t.p0")
+	c.run("zone get db.t")
+	c.run("zone get db.t@t_c2_idx")
+	c.run("zone set db.t.p1 --file=./../../cli/testdata/zone_range_max_bytes.yaml")
+	c.run("zone get db.t.p1")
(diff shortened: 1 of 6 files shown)


benesch:
@knz requesting your review on the CLI interfaces I publicly exposed and renamed, plus the script. Obviously feel free to review the rest too :)

danhhz:
The last commit lgtm. I'll leave the first two to @knz

benesch:
Ugh, I had to rebase out the first commit. Exposing `cli.cliTest` proves to be challenging, as it links a bunch of forbidden test code into the main binary. It's easier at the moment to reimplement the pieces we need in `cliccl`, but we can revisit that decision down the road if the CCL CLI gets more complicated.

knz:
Reviewed 1 of 1 files at r1, 4 of 4 files at r2.

---

*[pkg/cli/cli.go, line 164 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_Ppxbx9cj5tGlD1Tq:-Ky_Ppxbx9cj5tGlD1Tr:bewuusy) ([raw file](https://github.com/cockroachdb/cockroach/blob/1397ec535f2b65f57ed4a9d00d1c88eb9959bac3/pkg/cli/cli.go#L164)):*
> ```Go
> // TestingReset resets global mutable state so that Run can be called multiple
> // times from the same test process. It is public for cliccl.
> func TestingReset() {
> ```

I'd move this to a new file `testutils.go`

Also you're missing this:

```
  baseCfg.InitDefaults()
  InitCLIDefaults()
```

---

*[scripts/diff-example.sh, line 3 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_NsIFviEe5PMCsNec:-Ky_NsIFviEe5PMCsNed:b-dzwc5f) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ddd3286c7ab946097022e88b78af851f46485d7/scripts/diff-example.sh#L3)):*
> ```Shell
> #!/usr/bin/env bash
> 
> # Diff a failed Go example [0] against its expected output.
> ```

:sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart: 

---

*[scripts/diff-example.sh, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_O-NMniuXSi1j9r-r:-Ky_O-NMniuXSi1j9r-s:b-rgemk8) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ddd3286c7ab946097022e88b78af851f46485d7/scripts/diff-example.sh#L24)):*
> ```Shell
> 
> do_diff() {
>   if [ -x "$(command -v colordiff)" ]; then
> ```

command: command not found

s/command/which ?

---

*[scripts/diff-example.sh, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_OE4mbbcG0ufdYueo:-Ky_OE4mbbcG0ufdYuep:b2ng0y0) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ddd3286c7ab946097022e88b78af851f46485d7/scripts/diff-example.sh#L37)):*
> ```Shell
> 
> # The actual output is everything after "got:" but before "want:".
> <<<"$in" sed -n  "
> ```

There's no need for this indirection style. Since you already have a temporary directory (see above) simply dump the file to it:

```
cat >input
```

then here:

```
<input sed -n ...
```

and below too.

Then you can (conceptually) drop the dependency on bash (you still need it for `set -o pipefail` but that's it -- it makes the reading easier!)

The other advantage is that if the diff is very large, poor bash won't have to explode its memory usage.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948)*

benesch:

---

*[scripts/diff-example.sh, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_O-NMniuXSi1j9r-r:-Kyc1CT00Ml0ZEU_T_9K:bj8a8bp) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ddd3286c7ab946097022e88b78af851f46485d7/scripts/diff-example.sh#L24)):*

command: command not found

s/command/which ?
</blockquote></details>

o_O what shell are you using?! `command` is supposedly POSIX: https://stackoverflow.com/a/677212/1122351

---

*[scripts/diff-example.sh, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_OE4mbbcG0ufdYueo:-Kyc1FUKqp-hrfyse--R:bypr4fe) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ddd3286c7ab946097022e88b78af851f46485d7/scripts/diff-example.sh#L37)):*

There's no need for this indirection style. Since you already have a temporary directory (see above) simply dump the file to it:

```
cat >input
```

then here:

```
<input sed -n ...
```

and below too.

Then you can (conceptually) drop the dependency on bash (you still need it for `set -o pipefail` but that's it -- it makes the reading easier!)

The other advantage is that if the diff is very large, poor bash won't have to explode its memory usage.
</blockquote></details>

Hah, forgot I added in the tmpdir. (I didn't have it in the first version.) Nice catch, thanks. Fixed.

---

*[pkg/cli/cli.go, line 164 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_Ppxbx9cj5tGlD1Tq:-Kyc1KcaDZ-vwV2CRs7o:b-ru7myh) ([raw file](https://github.com/cockroachdb/cockroach/blob/1397ec535f2b65f57ed4a9d00d1c88eb9959bac3/pkg/cli/cli.go#L164)):*

I'd move this to a new file `testutils.go`

Also you're missing this:

```
  baseCfg.InitDefaults()
  InitCLIDefaults()
```
</blockquote></details>

Previously we were only doing that once per test, but looks like it's safe (and perhaps more correct?) to do once per `Run` or `RunWithArgs`. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948)*

knz:
:lgtm:

---

Reviewed 2 of 5 files at r3, 5 of 5 files at r4.

---

*[pkg/cli/testutils.go, line 23 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-KyppEdz3vjMMXPChagO:-KyppEdz3vjMMXPChagP:bob85ku) ([raw file](https://github.com/cockroachdb/cockroach/blob/33dee233c93c7d5414cc8d3a4b8381922840566a/pkg/cli/testutils.go#L23)):*
> ```Go
> 	// pointers (because they are tied into the flags), but instead
> 	// overwrite the existing structs' values.
> 	baseCfg.InitDefaults()
> ```

:tada: 

This makes me happy.

---

*[scripts/diff-example.sh, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-Ky_O-NMniuXSi1j9r-r:-KyppOq-p0aJNFysNKM9:b-jhkb7z) ([raw file](https://github.com/cockroachdb/cockroach/blob/9ddd3286c7ab946097022e88b78af851f46485d7/scripts/diff-example.sh#L24)):*

o_O what shell are you using?! `command` is supposedly POSIX: https://stackoverflow.com/a/677212/1122351
</blockquote></details>

Aye. TIL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19948#-:-KyppWowryuYiqsCEF0F:bnfp4nl)*

PR #19952, opened 2017-11-09
sql,sql/parser: update partitioning syntax per RFC

benesch:
Adjust the partitioning syntax per the latest update to the RFC. In
summary, list partitioning syntax is changed from

    CREATE TABLE ... PARTITION BY LIST (<cols>) VALUES (<exprs>) [, (<exprs>)...]

to

    CREATE TABLE ... PARTITION BY LIST (<cols>) VALUES IN (<expr>...)

and range partitioning syntax is changed from

    CREATE TABLE ... PARTITION BY RANGE (<cols>) VALUES LESS THAN (<expr>...)

to

    CREATE TABLE ... PARTITION BY RANGE (<cols>) VALUES < <expr>.

In the new syntax, <expr> must be a tuple whenever <cols> has more than
one column. When <cols> has only column, single exprs are automatically
promoted to 1-tuples.

As discussed in the RFC, this requires making DEFAULT and MAXVALUE part
of a_expr. (One could theoretically create a parallel expression
hierarchy without DEFAULT and MAXVALUE, but that would cause a
combinatorial explosion of the number of expression types.) The change
to move DEFAULT into a_expr already landed, so this commit only moves
MAXVALUE into a_expr.

### DIFF
--- a/pkg/ccl/sqlccl/zone_test.go
+++ b/pkg/ccl/sqlccl/zone_test.go
@@ -40,8 +40,8 @@ func TestValidIndexPartitionSetShowZones(t *testing.T) {
 		CREATE DATABASE d;
 		USE d;
 		CREATE TABLE t (c STRING PRIMARY KEY) PARTITION BY LIST (c) (
-			PARTITION p0 VALUES ('a'),
-			PARTITION p1 VALUES (DEFAULT)
+			PARTITION p0 VALUES IN ('a'),
+			PARTITION p1 VALUES IN (DEFAULT)
 		)`)
 
 	yamlDefault := fmt.Sprintf("gc: {ttlseconds: %d}", config.DefaultZoneConfig().GC.TTLSeconds)
@@ -280,8 +280,8 @@ func TestGenerateSubzoneSpans(t *testing.T) {
 		{
 			name: `single col list partitioning`,
 			schema: `CREATE TABLE t (a INT PRIMARY KEY) PARTITION BY LIST (a) (
-				PARTITION p3 VALUES (3),
-				PARTITION p4 VALUES (4)
+				PARTITION p3 VALUES IN (3),
+				PARTITION p4 VALUES IN (4)
 			)`,
 			subzones: []string{`@primary`, `.p3`, `.p4`},
 			expected: []string{
@@ -294,9 +294,9 @@ func TestGenerateSubzoneSpans(t *testing.T) {
 		{
 			name: `single col list partitioning - DEFAULT`,
 			schema: `CREATE TABLE t (a INT PRIMARY KEY) PARTITION BY LIST (a) (
-				PARTITION p3 VALUES (3),
-				PARTITION p4 VALUES (4),
-				PARTITION pd VALUES (DEFAULT)
+				PARTITION p3 VALUES IN (3),
+				PARTITION p4 VALUES IN (4),
+				PARTITION pd VALUES IN (DEFAULT)
 			)`,
 			subzones: []string{`@primary`, `.p3`, `.p4`, `.pd`},
 			expected: []string{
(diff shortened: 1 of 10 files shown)


danhhz:
:lgtm:

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/create.go, line 1269 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952#-KyXe0_5b6WIwPgoXsyF:-KyXe0_5b6WIwPgoXsyG:b-u6ukoc) ([raw file](https://github.com/cockroachdb/cockroach/blob/f3dedde3bdb0f92a82303e37e81bfa44a32a0132/pkg/sql/create.go#L1269)):*
> ```Go
> 		// This ensures we see the MAXVALUE in VALUES < (MAXVALUE). We don't bother
> 		// walking the expression to discover crazier nestings, like
> 		// VALUES < (((MAXVALUE))).
> ```

I would expect the crazy nesting to work (for things other than DEFAULT and MAXVALUE) because of how Eval works. Add a test either way plz

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952#-:-KyXeZ3nz6gSTp4gLTSM:bnfp4nl)*

benesch:
Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/create.go, line 1269 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952#-KyXe0_5b6WIwPgoXsyF:-KyXeco4SrJrss_wtGiH:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f3dedde3bdb0f92a82303e37e81bfa44a32a0132/pkg/sql/create.go#L1269)):*

I would expect the crazy nesting to work (for things other than DEFAULT and MAXVALUE) because of how Eval works. Add a test either way plz
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952)*

knz:
Reviewed 11 of 12 files at r1, 1 of 1 files at r2.

---

*[pkg/sql/create.go, line 1261 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952#-Ky_SpKpC8IwEA3Jh8Le:-Ky_SpKpC8IwEA3Jh8Lf:b-riigqc) ([raw file](https://github.com/cockroachdb/cockroach/blob/362273fcdbb20f228a985ff25ca4ba0d6319b663/pkg/sql/create.go#L1261)):*
> ```Go
> ) ([]byte, error) {
> 	var tuple *parser.Tuple
> 	switch t := maybeTuple.(type) {
> ```

call `StripParens` before the `switch` then remove the `ParenExpr` case below, and you'll have handled everything -- even cases where the tuple you want is inside parentheses.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952)*

benesch:
Review status: 10 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/create.go, line 1261 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952#-Ky_SpKpC8IwEA3Jh8Le:-Kyc8lNGSvmER5I9Wrt7:b17zozu) ([raw file](https://github.com/cockroachdb/cockroach/blob/362273fcdbb20f228a985ff25ca4ba0d6319b663/pkg/sql/create.go#L1261)):*

call `StripParens` before the `switch` then remove the `ParenExpr` case below, and you'll have handled everything -- even cases where the tuple you want is inside parentheses.
</blockquote></details>

:raised_hands: thank you! Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952)*

knz:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19952#-:-Kyc9OsIpMWWJtUszRSk:bnfp4nl)*

PR #19956, opened 2017-11-09
sql: crib Go's JSON string encoder

justinj:
There were a few cases where we were not encoding non-printable
characters correctly, but I realized we can just defer to Go's string
encoder rather than modifying our existing SQL one. The public interface
to this method is a few levels deep, requires reflection, and claims it
can return an error (but there are no errors possible in the string
case), so I decided it was easier to just copy this code over and adapt
it to our purposes.


m-schneider:
Looks good, what modifications did you make?

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/util/json/json.go, line 233 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-KyXxp9ftky9YrT0JDAj:-KyXxnQ6PpoLbCZPkSGW:b-dm2dc) ([raw file](https://github.com/cockroachdb/cockroach/blob/06477f76be7e390275621ba5f8bf30c187362b03/pkg/util/json/json.go#L233)):*
> hex
Can you rename this to something like hex alphabet or valid hex characters or something to make it more obvious how your going to use it in the future?

---

*[pkg/util/json/json_test.go, line 146 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-KyXyJoxZMMynBESF0Ze:-KyXyJoxZMMynBESF0Zf:bio3u53) ([raw file](https://github.com/cockroachdb/cockroach/blob/06477f76be7e390275621ba5f8bf30c187362b03/pkg/util/json/json_test.go#L146)):*
> ```Go
> 		`"\u001e"`,
> 		`"\u001f"`,
> 		`"\uffff"`,
> ```

Can you add tests for the \" and \\ cases as well?

---

*[pkg/util/json/tables.go, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-KyXyfoK4FTJhZgFHxgz:-KyXyfoK4FTJhZgFHxh-:bceoe28) ([raw file](https://github.com/cockroachdb/cockroach/blob/06477f76be7e390275621ba5f8bf30c187362b03/pkg/util/json/tables.go#L17)):*
> ```Go
> // All values are true except for the ASCII control characters (0-31), the
> // double quote ("), and the backslash character ("\").
> var safeSet = [utf8.RuneSelf]bool{
> ```

Is that how go does it? It seems like most of this can be expressed by a few ranges.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-:-KyXyzGMeeoHepi8IMzT:b-w83jp0)*

justinj:
I just deleted a section concerned with embedding in HTML.

---

---

*[pkg/util/json/json.go, line 233 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-KyXxp9ftky9YrT0JDAj:-KyY17zjFzioaoah2UwX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/06477f76be7e390275621ba5f8bf30c187362b03/pkg/util/json/json.go#L233)):*

> hex
Can you rename this to something like hex alphabet or valid hex characters or something to make it more obvious how your going to use it in the future?
</blockquote></details>

Done.

---

*[pkg/util/json/json_test.go, line 146 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-KyXyJoxZMMynBESF0Ze:-KyY1PLKB1r0k5fdbCIT:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/06477f76be7e390275621ba5f8bf30c187362b03/pkg/util/json/json_test.go#L146)):*

Can you add tests for the \" and \\ cases as well?
</blockquote></details>

Done.

---

*[pkg/util/json/tables.go, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-KyXyfoK4FTJhZgFHxgz:-KyY1PjcrRkrrvxhJw6E:b-bscy6n) ([raw file](https://github.com/cockroachdb/cockroach/blob/06477f76be7e390275621ba5f8bf30c187362b03/pkg/util/json/tables.go#L17)):*

Is that how go does it? It seems like most of this can be expressed by a few ranges.
</blockquote></details>

Yeah, this file is taken from Go.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-:-KyXzG8j1pMzYvpmQsjq:b-jf3oxb)*

pkg/util/json/json.go
```diff
@@ -231,22 +230,64 @@ func (j jsonNumber) String() string { return asString(j) }
 func (j jsonArray) String() string  { return asString(j) }
 func (j jsonObject) String() string { return asString(j) }
 
+const hexAlphabet = "0123456789abcdef"
+
 // encodeJSONString writes a string literal to buf as a JSON string.
-// Very similar to encodeSQLStringInsideArray. Primary difference is that it is
-// legal to directly print out unicode characters.
-func encodeJSONString(buf *bytes.Buffer, in string) {
+// Cribbed from https://github.com/golang/go/blob/master/src/encoding/json/encode.go.
```
benesch: Can you make this a permalink? Here and below. (Pressing <kbd>y</kdb> on the GH page will permalinkize the URL.)

justinj: I was not, that's useful!

m-schneider:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19956#-:-KyqRuv3XekxBmqMPzIl:bnfp4nl)*

justinj:
TFTR!

PR #19960, opened 2017-11-09
sql,parser: Add SCRUB support for checking SQL CHECK and FOREIGN KEY

lgo:
This adds the stubbed scrub command to check constraints.

    SCRUB TABLE <table> WITH OPTIONS CONSTRAINT ALL
    SCRUB TABLE <table> WITH OPTIONS CONSTRAINT (<constraint_name>...)

Currently, the constraints supported are SQL CHECK and FOREIGN KEY.


knz:
@dt would you mind checking the FK-related logic in here?

knz:
Reviewed 6 of 6 files at r1, 3 of 3 files at r2, 2 of 2 files at r3.

---

*[pkg/sql/scrub.go, line 525 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-Ky_RV4hqag-lya-omrD:-Ky_RV4hqag-lya-omrE:bvrbzi5) ([raw file](https://github.com/cockroachdb/cockroach/blob/6506f50b8263d30194b82a0125a8b833976c6b8a/pkg/sql/scrub.go#L525)):*
> ```Go
> 		if i > 0 {
> 			buf.WriteByte(' ')
> 			buf.WriteString(conjunction)
> ```

I don't understand this. Can you explain? (in a comment)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960)*

lego:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/scrub.go, line 525 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-Ky_RV4hqag-lya-omrD:-Kyq56p5AZt1qmJ5smWn:b-bopvs3) ([raw file](https://github.com/cockroachdb/cockroach/blob/6506f50b8263d30194b82a0125a8b833976c6b8a/pkg/sql/scrub.go#L525)):*

I don't understand this. Can you explain? (in a comment)
</blockquote></details>

Done, added it to the function comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960)*

lego:
Bump @knz.

Also bumping @dt for FKey specific stuff. I also realized earlier with another feature that querying with equality on nullable columns (i.e. when `NULL = NULL` is false) may result in missing some matches. I briefly checked, and it looks like FKey constraints work on the assumption that `NULL = NULL` matches, e.g. if some of the FKey columns are null. Does this sound right?

(If that is true, then I'll also need to modify the query used here to be similar to #20177 where the join will match on `NULL`)

knz:
:lgtm: - my sincerest apologies for delaying my review.
Please consider splitting `scrub.go` in multiple files.

---

Reviewed 3 of 9 files at r4, 1 of 3 files at r5, 2 of 2 files at r6.

---

*[pkg/sql/scrub.go, line 798 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-KzzLOARKQm2-3IsrCzY:-KzzLOARKQm2-3IsrCzZ:byzgjwq) ([raw file](https://github.com/cockroachdb/cockroach/blob/c2dcc4b9ecb6739c6c8c9cb8e41ef8b9a04618bc/pkg/sql/scrub.go#L798)):*
> ```Go
> 	return rows, nil
> }
> ```

Please split the various checkers in separate source files (`scrub_index.go`, `scrub_constraint.go`, `scrub_fk.go`,  etc)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-:-KzzL_QfD6Vu8LKBaCcL:b4aj9ff)*

lego:
TYFR.

I split the new checks into their respective files but left the existing index check to prevent conflicts in other in-flight PRs. I'll come back to moving that after those PRs have merged.


To follow up on the last comment I made about FKey check correctness. The current check is correct when there are `NULL` values in FKeys but this PR also exhibits the poor performance that https://github.com/cockroachdb/cockroach/pull/20177 fixes. Whichever merges later will have to be modified, most likely the other PR.

---

Review status: 1 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-:-L-2pECTRxuzuHxOuRLM:b-yxntrt)*

lego:
Sorry for another heap of changes!

This time it was rebasing around some of the refactorings and matching that (e.g., run state). I also made the FKey performance change in the last comment as this PR ended up coming after :). The FKey check now uses the same special join modifier as index checks, where we force `NULL = NULL` to be true in order to use a far more efficient join.

---

Review status: 1 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-:-L-D9WcUBBiA_C6rpb6N:b-esl3tu)*

knz:
:lgtm: with nit

---

Reviewed 3 of 9 files at r4, 5 of 5 files at r7, 1 of 4 files at r8, 5 of 5 files at r9.

---

*[pkg/sql/logictest/testdata/logic_test/distsql_scrub, line 26 at r9](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-L-Hv1bTXeMKvWvVw8Q4:-L-Hv1bTXeMKvWvVw8Q5:b-dhciru) ([raw file](https://github.com/cockroachdb/cockroach/blob/9cb867dcaee7fb1b05cea58c0969f6409d3a9a63/pkg/sql/logictest/testdata/logic_test/distsql_scrub#L26)):*
> ```
> statement ok
> CREATE TABLE parent (
> 	id INT PRIMARY KEY,
> ```

spaces not tabs, throughout

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19960#-:-L-HvL9SQldNvwvYe2u3:b-xv64lu)*

PR #19961, opened 2017-11-09
sql,distsqlrun: Add SCRUB checks for physical table integrity

lgo:
This updates the  #RFC to match the changes made by this pull request
and #20177. Also mark the RFC as completed. This PR will fulfill the
physical check part. #19586 and #20177 fulfilled the index check part.

This plumbs the physical check flag from the scanNode to the TableReader
and RowFetcher. The flag will be used to have special conditions apply
during scanning when running a physical check.

The error strings for scrub are moved into sql/scrub to consolidate the
error messages. The package also provides a scrub.Error wrapper for use
in MultiRowFetcher to annotate all errors passing through with a scrub
error type.

Modifcations are made to MultiRowFetcher to annotate all errors that
pass through it during iteration. In this commit, all of these
annotations are ignored by consumers. All of these errors are
categorized into specific physical data errors, including:

- IndexKeyDecodingError

  If an error is caused by a primary or secondary index's key,
  the error is annotated with this category. Problems that may have
  occurred include family ID decoding or fetching errors, failing round-trip
  encoding checks, or failing key-datum order checks.

- IndexValueDecodingError

  If an error is caused by the primary or secondary index's value, the
This moves the error strings f
  error is annotated with this category. This includes errors
  encountered while attempting to decode the value tuple, or failing
  round-trip encoding checks.

- SecondaryIndexExtraValueDecodingError

  If an error happens while attempting to decode the extra columns from a
  secondary index's key, it is annotated with this category.

- UnexpectedNullValueError

  If a row has a `NULL` value in a non-nullable column then this error
  is returned.

- PhysicalError

  All other errors that are not in the above categories are annotated
  with this category.

This modifies the distSQL TableReader so that a physical check is being
run via. indicating TableReaderSpec.IsCheck. When executing with
IsCheck true, the TableReader iterates on MultiRowfetcher
using NextRowWithErrors instead of NextRow. This will cause extra checks
to be run:
- Round-trip encoding checks of datums.
- Verify the key ordering matches the decoded datum ordering.

When the TableReader receives any errors that are wrapped by a
ScrubError when running the physical check, it streams them as rows
instead of propagating an error and stopping execution.

Using the prior modifications, physical checks are implemented by
generating a distSQL plan and using the TableReader.IsCheck flag.
They can be run with the following:

    EXPERIMENTAL SCRUB TABLE <table>  WITH OPTIONS PHYSICAL


RaduBerinde:
Review status: 0 of 15 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/sql/distsql_physical_planner.go, line 680 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrPuIt71EAvc5jb2oo:-KyrPuIt71EAvc5jb2op:b3fi1e9) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L680)):*
> ```Go
> 
> 	post := distsqlrun.PostProcessSpec{
> 		Filter: distsqlplan.MakeExpression(n.filter, &n.p.evalCtx, nil),
> ```

I don't think this is a valid PostProcessSpec if IsCheck is on. If it is impossible to have a filter with IsCheck, that should be verified here.

---

*[pkg/sql/distsql_physical_planner.go, line 906 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrNQEVtml7Zc8mM6Ql:-KyrNQEVtml7Zc8mM6Qm:b-h6gur2) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L906)):*
> ```Go
> 	readAsOf hlc.Timestamp,
> ) (physicalPlan, error) {
> 	spec, post, err := initTableReaderSpec(n)
> ```

[nit] Add a high level comment describing the plan (basically just a bunch of table readers in IsCheck mode).

---

*[pkg/sql/distsql_physical_planner.go, line 941 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrQZrPrHP33cYDuYgA:-KyrQZrPrHP33cYDuYgB:b-to4u4x) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L941)):*
> ```Go
> 	}
> 
> 	// Modify the TableReader PostProcess spec to correspond to the scrub
> ```

This comment is a bit confusing - we aren't really modifying the postprocess spec because we know it doesn't contain anything interesting in this case (otherwise we wouldn't know for sure that the output types are ScrubTypes). It might be more clear to not use `post` at all, and just do `p.ResultTypes = distsqlrun.ScrubTypes`.

---

*[pkg/sql/distsql_physical_planner.go, line 946 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrNvHkAXmhOmFqAGY_:-KyrNvHkAXmhOmFqAGYa:b-wq5hjc) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L946)):*
> ```Go
> 
> 	// Add the a projection stage, notably with the scrub types.
> 	// FIXME(joey): Is this necessary? We already modify the
> ```

No, I don't think this does anything.

---

*[pkg/sql/distsql_physical_planner.go, line 955 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrQR2yWIh8PsX_i_ZX:-KyrQR2yWIh8PsX_i_ZY:b-uj1rzc) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L955)):*
> ```Go
> 
> 	// Make the stream column map reflect that it's a 1-1 mapping.
> 	planToStreamColMap := identityMapInPlace(make([]int, len(distsqlrun.ScrubTypes)))
> ```

why the temp var?

---

*[pkg/sql/distsqlrun/processors.proto, line 158 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrO_kTK5ZLiyKjhlEg:-KyrO_kUBKT88EzKS1K3:b-cbuzb7) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsqlrun/processors.proto#L158)):*
> ```protobuf
> // The "internal columns" of a TableReader (see ProcessorSpec) are all the
> // columns of the table. Internally, only the values for the columns needed by
> // the post-processing stage are be populated.
> ```

Need to add to this description: if is_check is set, the "internal columns" are
 - Error type (string)
 - Primary key as a string, if it was obtainable.
 - JSON of all decoded column values.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961)*

lego:
@RaduBerinde @jordanlewis PTAL, the PR is no longer WIP (well, it's mostly polished with some noted exceptions).

(lmk if you're the right reviewers. I thought it would be appropriate for this touches encoding + table reader surfaces)

---

Review status: 0 of 17 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsql_physical_planner.go, line 680 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrPuIt71EAvc5jb2oo:-KyvhNHntqwBpthxc_ZA:b-dw87jq) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L680)):*

I don't think this is a valid PostProcessSpec if IsCheck is on. If it is impossible to have a filter with IsCheck, that should be verified here.
</blockquote></details>

Done, this now exits early with an empty post processor spec.

---

*[pkg/sql/distsql_physical_planner.go, line 906 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrNQEVtml7Zc8mM6Ql:-Kyvh4yOtn1Ppu9BoQJe:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L906)):*

[nit] Add a high level comment describing the plan (basically just a bunch of table readers in IsCheck mode).
</blockquote></details>

Done.

---

*[pkg/sql/distsql_physical_planner.go, line 941 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrQZrPrHP33cYDuYgA:-KyviG503FXC4MqszAZr:b-yeh1oz) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L941)):*

This comment is a bit confusing - we aren't really modifying the postprocess spec because we know it doesn't contain anything interesting in this case (otherwise we wouldn't know for sure that the output types are ScrubTypes). It might be more clear to not use `post` at all, and just do `p.ResultTypes = distsqlrun.ScrubTypes`.
</blockquote></details>

Awesome :)! I had no idea, I was sort of using `createTableReaders` for reference so I had assumed a few of these artifacts were required

---

*[pkg/sql/distsql_physical_planner.go, line 946 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrNvHkAXmhOmFqAGY_:-KyviFXUouPenzXwq3mB:b-cz0282) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L946)):*

No, I don't think this does anything.
</blockquote></details>

Awesome. Removed.

---

*[pkg/sql/distsql_physical_planner.go, line 955 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrQR2yWIh8PsX_i_ZX:-KyvigiQgTspbmMOiNjZ:bss4a2b) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsql_physical_planner.go#L955)):*

why the temp var?
</blockquote></details>

Removed.

---

*[pkg/sql/scrub.go, line 626 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KzyfURJQabn5GR7Z9fy:-KzyfURJQabn5GR7Z9fz:b-lnorzg) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub.go#L626)):*
> ```Go
> 	// Intermediate values.
> 	rows     *sqlbase.RowContainer
> 	rowIndex int
> ```

Self-note: when the PR that splits run-state from plan-state this should also be split (or can be right now to match that).

---

*[pkg/sql/scrub_test.go, line 543 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-Kzyfh0w2XQ7wpb-n-rj:-Kzyfh0w2XQ7wpb-n-rk:b-2f6nch) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L543)):*
> ```Go
> func TestScrubPhysicalUnexpectedFamilyID(t *testing.T) {
> 	t.Skip("currently unexpected family IDs are picked up")
> ```

From my understanding of RowFetcher, I would expect an error to be thrown if a k/v is scanned that has an unknown family ID as the key suffix. This does not seem to get caught, or throw an error during scanning. Any thoughts why?

---

*[pkg/sql/scrub_test.go, line 664 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KzyfzqhbrnmrropbRTq:-KzyfzqhbrnmrropbRTr:b-lhcck) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L664)):*
> ```Go
> func TestScrubPhysicalIncorrectPrimaryIndexValueColumn(t *testing.T) {
> 	t.Skip("this currently does not fail as maybe expected")
> ```

Same as the above but a different error case. I would expect scanning to fail if there is an unknown column ID encoded in the primary index value, but again there is not error raised. Any thoughts why?

---

*[pkg/sql/distsqlrun/processors.proto, line 158 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KyrO_kTK5ZLiyKjhlEg:-Kyvj9O_vmUMBZxjl4Il:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/6f16243d9022e39dad56254c6e1f9c4a39cf1df7/pkg/sql/distsqlrun/processors.proto#L158)):*

Need to add to this description: if is_check is set, the "internal columns" are
 - Error type (string)
 - Primary key as a string, if it was obtainable.
 - JSON of all decoded column values.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/multirowfetcher.go, line 901 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KzygkaXOKMG6Bp-J82g:-KzygkaXOKMG6Bp-J82h:b6f3jwj) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/sqlbase/multirowfetcher.go#L901)):*
> ```Go
> 		err = mrf.checkPrimaryIndexDatumEncodings(ctx)
> 	} else {
> 		// err = mrf.checkSecondaryIndexDatumEncodings(ctx)
> ```

Temporarily commented out as this is currently failing with unexpected errors (although it may correspond to other errors, as the test data is quite broken).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-:-KzyenWLj9OZVRyySkBr:bnh6bfh)*

lego:
Review status: 0 of 17 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

*[pkg/sql/scrub_test.go, line 752 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KzyhkIKnRsPjDaYG_Jd:-KzyhkIKnRsPjDaYG_Je:b35hx2u) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L752)):*
> ```Go
> 	} else if !strings.Contains(result.details, `"b":"<unset>"`) {
> 		t.Fatalf("expected erorr details to contain `%s`, got %s", `"b":"<unset>"`, result.details)
> 	}
> ```

There are quite a few possible errors to test. I don't know if we need to test all of them and I also don't know how to cause many of them, like the above two tests. Those tests I expect would encounter errors while scanning but don't seem to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961)*

lego:
(sorry for all the pushing churn, ran into tests failing on CI that I didn't catch locally for a while. that's now fixed although the last test run did time out)

lego:
Ping.

knz:
Thanks for the heads up!

Any reason why you're not using `errors.Wrap` / `errors.GetCause` instead of baking your own solution?

The remainder of the change LGTM modulo a few nits. Well done, I like how readable the tests are.

Speaking of the tests however, of course I'm not completely paged in on all that can go wrong with data. If you have a moment scheduled where you're going to talk to the rest of the team about your work, be sure to remind everyone that if they can think of any creative ways to  corrupt the data, this test code could be extended.

---

Reviewed 21 of 21 files at r7, 13 of 13 files at r8, 6 of 6 files at r9, 5 of 5 files at r10, 4 of 4 files at r11, 6 of 6 files at r12.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/scan.go, line 75 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cAsLZT-kCWRHBSaPp:-L-cAsL_xGYKJVeq6dFy:bgyzoji) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0991c211606b2c4c31b5b84e6e5ed934d2d9e4c/pkg/sql/scan.go#L75)):*
> ```Go
> 	// Indicates if this scanNode will do a physical data check. This is
> 	// only true when running SCRUB commands.
> 	isCheck bool
> ```

This must become a member of the `run` struct.

---

*[pkg/sql/scrub_physical.go, line 128 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cDGwloB2K6qsYsxhu:-L-cDGwloB2K6qsYsxhv:b7bo86e) ([raw file](https://github.com/cockroachdb/cockroach/blob/e1859c38034f5ef44673395f97eece2f2bac7659/pkg/sql/scrub_physical.go#L128)):*
> ```Go
> 		return err
> 	}
> 	defer plan.Close(ctx)
> ```

nit: move the `defer` just before the `if err != nil` and you got all your bases covered this way.

---

*[pkg/sql/scrub_test.go, line 543 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-Kzyfh0w2XQ7wpb-n-rj:-L-cBKpQTEgbemvRwadQ:b-84co4y) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L543)):*

From my understanding of RowFetcher, I would expect an error to be thrown if a k/v is scanned that has an unknown family ID as the key suffix. This does not seem to get caught, or throw an error during scanning. Any thoughts why?
</blockquote></details>

File an issue. Ask Vivek for triage.

---

*[pkg/sql/scrub_test.go, line 664 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KzyfzqhbrnmrropbRTq:-L-cBOK7TNrHZZ45NOzp:b-89a4l7) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L664)):*

Same as the above but a different error case. I would expect scanning to fail if there is an unknown column ID encoded in the primary index value, but again there is not error raised. Any thoughts why?
</blockquote></details>

Ditto

---

*[pkg/sql/scrub_test.go, line 709 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cEt-feJ5YFJTfWO3T:-L-cEt-feJ5YFJTfWO3U:b-kw5r0a) ([raw file](https://github.com/cockroachdb/cockroach/blob/e1859c38034f5ef44673395f97eece2f2bac7659/pkg/sql/scrub_test.go#L709)):*
> ```Go
> 		t.Fatalf("unexpected error: %s", err)
> 	}
> 	defer rows.Close()
> ```

defer early. The Fatal calls will cause a leak otherwise.

(Here and everywhere you've used rows.)

---

*[pkg/sql/distsqlrun/indexbackfiller.go, line 172 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cBeY0dSVvvoQsiEXG:-L-cBeY0dSVvvoQsiEXH:b301vn1) ([raw file](https://github.com/cockroachdb/cockroach/blob/262dc639a00c12c82f7f2f145ac9c1f038a8091d/pkg/sql/distsqlrun/indexbackfiller.go#L172)):*
> ```Go
> 			if err != nil {
> 				if scrub.IsScrubError(err) {
> 					err = scrub.UnwrapScrubError(err)
> ```

Move the conditional into `UnwrapScrubError` and call it unconditionally here.

---

*[pkg/sql/distsqlrun/interleaved_reader_joiner.go, line 240 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cBn2PWE3HOORMz_EJ:-L-cBn2PWE3HOORMz_EK:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/262dc639a00c12c82f7f2f145ac9c1f038a8091d/pkg/sql/distsqlrun/interleaved_reader_joiner.go#L240)):*
> ```Go
> 			if err != nil {
> 				if scrub.IsScrubError(err) {
> 					err = scrub.UnwrapScrubError(err)
> ```

ditto

---

*[pkg/sql/distsqlrun/joinreader.go, line 181 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cBpAFZyb9JMuSRxj5:-L-cBpAFZyb9JMuSRxj6:b-7ropln) ([raw file](https://github.com/cockroachdb/cockroach/blob/262dc639a00c12c82f7f2f145ac9c1f038a8091d/pkg/sql/distsqlrun/joinreader.go#L181)):*
> ```Go
> 			row, _, _, err := jr.fetcher.NextRow(ctx)
> 			if err != nil {
> 				if scrub.IsScrubError(err) {
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-:-L-cBwLSaW4DBOXPwi-n:b-d8i9dw)*

lego:
The reason I rolled a separate one as there was no way to identify if an error was a scrub error well. The prefix of scrub wrapped `error.String()` wouldn't be a good way to identify them, because of the variety of error messages. We also can't just call `Cause` on each error because it may not be a scrub error and it may be wrapped with something else.

Totally agree with the bit about the tests.

---

Review status: 1 of 24 files reviewed at latest revision, 14 unresolved discussions.

---

*[pkg/sql/scan.go, line 75 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cAsLZT-kCWRHBSaPp:-L-hPBGPslHLiUcdB7Sy:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a0991c211606b2c4c31b5b84e6e5ed934d2d9e4c/pkg/sql/scan.go#L75)):*

This must become a member of the `run` struct.
</blockquote></details>

Done.

---

*[pkg/sql/scrub_physical.go, line 128 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cDGwloB2K6qsYsxhu:-L-hH92uZ6Wyr0TIC3xv:b-spyiyv) ([raw file](https://github.com/cockroachdb/cockroach/blob/e1859c38034f5ef44673395f97eece2f2bac7659/pkg/sql/scrub_physical.go#L128)):*

nit: move the `defer` just before the `if err != nil` and you got all your bases covered this way.
</blockquote></details>

The linter doesn't allow a defer before an err check, so go here.

---

*[pkg/sql/scrub_test.go, line 543 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-Kzyfh0w2XQ7wpb-n-rj:-L-h_15Lbv6kMHyM_MTG:b2eqc70) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L543)):*

File an issue. Ask Vivek for triage.
</blockquote></details>

Done #20538.

---

*[pkg/sql/scrub_test.go, line 664 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-KzyfzqhbrnmrropbRTq:-L-h_1xlUIO9uE35OYlj:bjg0taj) ([raw file](https://github.com/cockroachdb/cockroach/blob/42facfdb62f642c0c2b2bb60e0df8b0970d9f178/pkg/sql/scrub_test.go#L664)):*

Ditto
</blockquote></details>

Same filed issue as above.

---

*[pkg/sql/scrub_test.go, line 709 at r12](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cEt-feJ5YFJTfWO3T:-L-hG1k7WUZIOYoe-YQG:b-849pke) ([raw file](https://github.com/cockroachdb/cockroach/blob/e1859c38034f5ef44673395f97eece2f2bac7659/pkg/sql/scrub_test.go#L709)):*

defer early. The Fatal calls will cause a leak otherwise.

(Here and everywhere you've used rows.)
</blockquote></details>

Done. These are a lot cleaned up after I realized I was incorrectly using the DB interface here, but it should no longer leak as you mentioned.

---

*[pkg/sql/distsqlrun/indexbackfiller.go, line 172 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cBeY0dSVvvoQsiEXG:-L-hH6B5xhvKUdmhlDiX:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/262dc639a00c12c82f7f2f145ac9c1f038a8091d/pkg/sql/distsqlrun/indexbackfiller.go#L172)):*

Move the conditional into `UnwrapScrubError` and call it unconditionally here.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/interleaved_reader_joiner.go, line 240 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cBn2PWE3HOORMz_EJ:-L-hH6hqdGAz1RqyxNdL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/262dc639a00c12c82f7f2f145ac9c1f038a8091d/pkg/sql/distsqlrun/interleaved_reader_joiner.go#L240)):*

ditto
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/joinreader.go, line 181 at r10](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-L-cBpAFZyb9JMuSRxj5:-L-hH75xG3Pdv4JSzgW9:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/262dc639a00c12c82f7f2f145ac9c1f038a8091d/pkg/sql/distsqlrun/joinreader.go#L181)):*

ditto
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-:-L-hMLX6eYiJ1nsO8SZV:b-e1q1lz)*

knz:
:lgtm:

---

Reviewed 6 of 14 files at r14, 2 of 5 files at r15, 4 of 5 files at r16, 4 of 4 files at r17, 7 of 7 files at r18.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19961#-:-L-hcIzNTIET3mmns4Qb:bnfp4nl)*

PR #19964, opened 2017-11-10
sql/parser: remove generated files

knz:
Checking in sql.go and related files causes several major problems:

* Every change to the SQL grammar that shifts the contents by one line
  or more, even if the diff is only one line, rewrites most of
  sql.go. This adds a lot of weight to the Git repository.

* Every commit that touches the SQL grammar must be serialized, as
  diffs in sql.go and related files cannot be automatically merged.

* The size of sql.go causes general slowdown of Reviewable in every PR
  that touches the SQL grammar.

Auto-generating these files uses goyacc which is vendored. The
generation is also rather fast, so there is no real visible downside.

cc @cockroachdb/sql


benesch (commented):
Omg ❤️❤️❤️

Your comment about goyacc is surprising though. We mount a separate bin dir in the builder, so we should definitely be re-runningthe bootstrapping the first time you run Make in the builder.

knz:
I don't think so. The rule for bootstrapping depends just on `.bootstrap`, not on the resulting binaries in $GOPATH/bin. That `.bootstrap` gets mounted in the builder, but not `$GOPATH/bin` (for good reason: it might be a different arch).

knz:
... So the deps on `.bootstrap` assume the things have been installed, whereas they are not.

knz:
What surprised me, is that no other rule really depends on these vendored tool dependencies. But then if you think of it, indeed they don't. These are dev only tools, not used in tests. Except for goyacc now, that is.

benesch:
But the .bootstrap file is stored inside `$REPO_ROOT/bin`: https://github.com/cockroachdb/cockroach/blob/93b6827575a8cab4790106fc0851ec03d06ff190/build/common.mk#L167-L171

knz:
FWIW it's green 🍏 📗 💚 🥗

knz:
Re .bootstrap: I don't know what's going on. What I do know is that if I don't use this new `goyacc.installed` thing, *all* the TC tests fail with "goyacc missing".

benesch (approved):
LGTM!

We can figure out the goyacc problem later.

knz:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #19966, opened 2017-11-10
sql: add transaction_read_only variable

### LINKED ISSUE #19911: sql: add transaction_read_only variable
Some drivers read this variable. We don't support setting it in any way, but we should support getting it (and it should be hardcoded to always be false).

bsnyder788:
Closes #19911. @justinj this should address adding the read-only transaction_read_only variable.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/explain
+++ b/pkg/sql/logictest/testdata/logic_test/explain
@@ -145,7 +145,7 @@ EXPLAIN SHOW DATABASE
 0  render  ·     ·
 1  filter  ·     ·
 2  values  ·     ·
-2  ·       size  2 columns, 21 rows
+2  ·       size  2 columns, 22 rows
 
 query ITTT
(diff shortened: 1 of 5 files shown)


knz:
Very good. Solid change. LGTM.

Thanks for your contribution!

justinj:
It actually looks to me like when this is not true PG outputs "off" rather than "false":
```
justin=# show transaction_read_only;
 transaction_read_only
-----------------------
 off
(1 row)
```
I should have checked this before opening the original issue, but could we fix this to match that behaviour?

bsnyder788:
You are correct, I just tested on 9.6 as well and totally missed that. Let me update the PR.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19966#-:-KyaxoRW6PCO5UZQypCH:b-li4gzb)*

PR #19970, opened 2017-11-10
cloud: Add config for running cockroach commands in secure k8 clusters

jseldess:
- Add deployment config for cockroach client commands in secure k8 clusters
- Remove limitation about rescheduled pods needing to go through the CSR approval process
- Update instructions for accessing the sql shell from k8

### DIFF
--- a/cloud/kubernetes/README.md
+++ b/cloud/kubernetes/README.md
@@ -189,7 +189,7 @@ kubectl create -f cockroachdb-client-secure.yaml
 Check and approve the CSR for the pod as described above, and then get a shell to the pod and run:
 
 ```shell
-kubectl exec -it <pod name> -- ./cockroach sql --certs-dir=/cockroach-certs --host=cockroachdb-public
+kubectl exec -it cockroachdb-client-secure -- ./cockroach sql --certs-dir=/cockroach-certs --host=cockroachdb-public
 ```
 
 You can see example SQL statements for inserting and querying data in the
--- a/cloud/kubernetes/README.md
+++ b/cloud/kubernetes/README.md
@@ -172,14 +172,25 @@ Along with our StatefulSet configuration, we expose a standard Kubernetes servic
 that offers a load-balanced virtual IP for clients to access the database
 with. In our example, we've called this service `cockroachdb-public`.
 
-Start up a client pod and open up an interactive, (mostly) Postgres-flavor
-SQL shell using:
+In insecure mode, start up a client pod and open up an interactive, (mostly) Postgres-flavor
+SQL shell:
 
-```console
-$ kubectl run -it --rm cockroach-client --image=cockroachdb/cockroach --restart=Never --command -- ./cockroach sql --host cockroachdb-public
+```shell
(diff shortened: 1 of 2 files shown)


cloud/kubernetes/README.md
```diff
@@ -71,15 +71,6 @@ first node is special in that the administrator must manually prepopulate the
 parameter. If this is not done, the first node will bootstrap a new cluster,
 which will lead to a lot of trouble.
 
-### Secure mode
-
-Secure mode currently works by requesting node/client certificates from the kubernetes
```
mberhault: It might be good to keep this part, it's still true. The only change is that restarted pods don't need to get a new CSR, they will use the certificate and key stored in secrets.

cloud/kubernetes/cockroachdb-client-secure.yaml
```diff
@@ -0,0 +1,51 @@
...
+      containers:
+      - name: cockroachdb-client
+        image: cockroachdb/cockroach:v1.1.2
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: client-certs
+          mountPath: /cockroach-certs
+        # Keep a pod open indefinitely so kubectl exec can be used to get a shell to it
+        # and run cockroach client commands, such as cockroach sql, cockroach node status, etc.
+        command:
```
mberhault: @a-robinson: do you know if there's a way to do without the `sleep 999...` hack?

mberhault (commented):
LGTM.

cloud/kubernetes/README.md
jseldess: K, added back in. I didn't understand that our current approach overall is a "limitation".

cloud/kubernetes/README.md
```diff
@@ -177,14 +172,25 @@ Along with our StatefulSet configuration, we expose a standard Kubernetes servic
...
+In secure mode, use our `cockroachdb-client-secure.yaml` config to launch a pod that runs indefinitely with the `cockroach` binary inside it:
+
+```shell
+kubectl create -f cockroachdb-client-secure.yaml
+```
+
+Check and approve the CSR for the pod as described above, and then get a shell to the pod and run:
+
+```shell
+kubectl exec -it <pod name> -- ./cockroach sql --certs-dir=/cockroach-certs --host=cockroachdb-public
```
a-robinson: If we make this a raw `Pod` object rather than a `Deployment`, we'll know the name in advance and can hard-code it in the instructions. Assuming it's ok with you, I'm going to push an update to your branch since that'll be easier than giving you the entire yaml in a comment.

cloud/kubernetes/cockroachdb-client-secure.yaml
a-robinson: The container has to have some sort of command, and we don't want that command to waste machine resources. If you know of a less ugly command, we can switch to that, but we do need to tell it to do something AFAIK.

cloud/kubernetes/README.md
jseldess: Sounds good, @a-robinson. Thanks!

a-robinson: Done.

PR #19971, opened 2017-11-10
sql: transaction_read_only should return off

### LINKED ISSUE #19911: sql: add transaction_read_only variable

bsnyder788:
Closes #19911 . @justinj @knz Sorry about that, I should have checked PG more closely too! Follow up to #19966


knz:
Code LGTM. @justinj can you give this the green light, given how I didn't catch the error in the first round.

justinj:
LGTM as well!

knz:
Thank you!

PR #19974, opened 2017-11-10
tscache: remove lowWaterTxnIDMarker, GetMaxRead/GetMaxWrite returns no flag

nvb:
Refactoring that simplifies #19508.

This is subtle, but I'm pretty confident it's correct. I split the change into
two commits, each of which I found easier to convince myself about in isolation.

The first of two changes removes `lowWaterTxnIDMarker` and uses `noTxnID` in its
place. To do this, it changes the semantics of the flag returned from
`GetMaxRead/GetMaxWrite` from returning whether the interval found is "not the
low-water mark" to returning whether the interval found "has a non-empty txnID".
This is ok, because the only difference between an interval having an empty
txnID and an interval being part of the low-water mark is that an interval with
an empty txnID could have also been created by two requests with different
txnIDs conflicting at the same timestamp or a request with no txnID to begin
with. However, the only consumer of this flag was looking at `TransactionKeys`,
which should only ever be accessed by a single txnID. This means that in the
only case where the flag was used, the two semantics are equivalent.However, to
take advantage of this, the change needed to give the `TransactionKey` timestamp
interval its `TxnID`, something we had not done before for historical reasons
(`GetMaxRead/GetMaxWrite` used to take a `TxnID` that it should ignore).

The next change takes advantage of the previous simplification to remove the
flag returned from `GetMaxRead/GetMaxWrite` completely. This removes a major
subtlety in the interface and makes it easier to implement.

### DIFF
--- a/pkg/storage/client_replica_test.go
+++ b/pkg/storage/client_replica_test.go
@@ -638,12 +638,17 @@ func TestRangeTransferLease(t *testing.T) {
 
 		replica1Lease, _ := replica1.GetLease()
 
-		// Verify the timestamp cache low water. Because we executed a transfer lease
-		// request, the low water should be set to the new lease start time which is
-		// less than the previous lease's expiration time.
-		if lowWater := replica1.GetTimestampCacheLowWater(); lowWater != replica1Lease.Start {
-			t.Fatalf("expected timestamp cache low water %s, but found %s",
-				replica1Lease.Start, lowWater)
+		// We'd like to verify the timestamp cache's low water mark, but this is
+		// impossible to determine precisely in all cases because it may have
+		// been subsumed by future tscache accesses. However, we know that there
+		// were no tscache accesses since we executed the transfer lease
+		// request, which means that the low water mark will still be equal to
+		// the high water mark. So instead of checking the low water mark, we
+		// make sure that the high water mark is set to the new lease start
+		// time, which is less than the previous lease's expiration time.
+		if highWater := replica1.GetTSCacheHighWater(); highWater != replica1Lease.Start {
+			t.Fatalf("expected timestamp cache high water %s, but found %s",
+				replica1Lease.Start, highWater)
 		}
 	})
 
--- a/pkg/storage/helpers_test.go
+++ b/pkg/storage/helpers_test.go
@@ -37,6 +37,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/util"
 	"github.com/cockroachdb/cockroach/pkg/util/hlc"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/uuid"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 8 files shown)


tschottdorf:
:lgtm:

---

Reviewed 5 of 5 files at r1, 7 of 7 files at r2.

---

*[pkg/storage/helpers_test.go, line 322 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-Kyq6xFpelE1q0a-wTzZ:-Kyq6xFpelE1q0a-wTz_:b3z9t3n) ([raw file](https://github.com/cockroachdb/cockroach/blob/5098799181fbef8b68e04b87249bbc31a56877c2/pkg/storage/helpers_test.go#L322)):*
> ```Go
> 	end := roachpb.Key(r.Desc().EndKey)
> 	noID := uuid.UUID{}
> 	if r, txnID := r.store.tsCacheMu.cache.GetMaxRead(start, end); txnID == noID && t.Less(r) {
> ```

The semantics here are a little weird (but they were before, too). Could you add a comment what the desired result here is, preferably on the method comment?

---

*[pkg/storage/tscache/cache.go, line 70 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-Kyq7YRN9bmP26uI8-Fj:-Kyq7YRN9bmP26uI8-Fk:b97khaa) ([raw file](https://github.com/cockroachdb/cockroach/blob/5098799181fbef8b68e04b87249bbc31a56877c2/pkg/storage/tscache/cache.go#L70)):*
> ```Go
> 	// specified range is overlapped by timestamps from different transactions
> 	// in the cache, the low water timestamp is returned for the read
> 	// timestamps.
> ```

This looks like a good place to explain the returned `UUID` and that it is returned only if that `UUID` supplied the strict maximum for the returned timestamp.

---

*[pkg/storage/tscache/cache_test.go, line 376 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-Kyq7r5kQGbxozIGnFI1:-Kyq7r5kQGbxozIGnFI2:b-dpsbql) ([raw file](https://github.com/cockroachdb/cockroach/blob/5098799181fbef8b68e04b87249bbc31a56877c2/pkg/storage/tscache/cache_test.go#L376)):*
> ```Go
> 		// Fetching any keys should give current time.
> 		if rTS, rTxnID := tc.GetMaxRead(key, nil); rTxnID != noTxnID {
> 			t.Errorf("expected %s to have cleared timestamp", key)
> ```

`t.Errorf("%s unexpectedly associated to txn %s", key, rTxnID)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-:-Kyq8LN2DOVJOEnM5H0P:bnfp4nl)*

nvanbenschoten:

---

*[pkg/storage/helpers_test.go, line 322 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-Kyq6xFpelE1q0a-wTzZ:-KyqpGcLtkDicQaXvy9a:bc5fb1h) ([raw file](https://github.com/cockroachdb/cockroach/blob/5098799181fbef8b68e04b87249bbc31a56877c2/pkg/storage/helpers_test.go#L322)):*

The semantics here are a little weird (but they were before, too). Could you add a comment what the desired result here is, preferably on the method comment?
</blockquote></details>

I agree. This method has been giving me all kinds of issues, primarily because the question it's asking isn't possible to answer 100% precisely. Replaced in the third commit.

---

*[pkg/storage/tscache/cache.go, line 70 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-Kyq7YRN9bmP26uI8-Fj:-KyqqMFDjNJf5wIXs_ZF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5098799181fbef8b68e04b87249bbc31a56877c2/pkg/storage/tscache/cache.go#L70)):*

This looks like a good place to explain the returned `UUID` and that it is returned only if that `UUID` supplied the strict maximum for the returned timestamp.
</blockquote></details>

Done.

---

*[pkg/storage/tscache/cache_test.go, line 376 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974#-Kyq7r5kQGbxozIGnFI1:-KyqqeKH4zw8eUdow3h1:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5098799181fbef8b68e04b87249bbc31a56877c2/pkg/storage/tscache/cache_test.go#L376)):*

`t.Errorf("%s unexpectedly associated to txn %s", key, rTxnID)`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974)*

tschottdorf:
Reviewed 7 of 7 files at r3, 7 of 7 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19974)*

PR #19975, opened 2017-11-10
sqlccl: don’t include system tables in all BACKUPs

dt:
This removes the automatic inclusion of the system.users and system.descriptors table in all backups.

Prior to this change, these tables were implicitly included in all backups, with the hope that if RESTORE later learned to use them for something (like restoring users and grants), they’d be available. 

However including them in their entirety could surprise a user: if they make a backup of some table or db, they may assume that the backup only includes information about that table or DB and, if they were to share it with someone else, and could be unpleasantly surprised to find it included all their table schemas and all cluster users with their hashed passwords.

These tables _can_ be easily included in their entirety in a backup if desired, simply by including e.g. `system.users` in the BACKUP command, which should also make clearer what data ends up in the resulting backup.

Release Note: BACKUPs no longer automatically include the system.users and system.descriptor table.

### DIFF
--- a/pkg/ccl/sqlccl/backup.go
+++ b/pkg/ccl/sqlccl/backup.go
@@ -66,14 +66,6 @@ var backupOptionExpectValues = map[string]bool{
 // to durable storage.
 var BackupCheckpointInterval = time.Minute
 
-// BackupImplicitSQLDescriptors are descriptors for tables that are implicitly
-// included in every backup, plus their parent database descriptors.
-var BackupImplicitSQLDescriptors = []sqlbase.Descriptor{
-	*sqlbase.WrapDescriptor(&sqlbase.SystemDB),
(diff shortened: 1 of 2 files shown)


dt:
Speculative / future-proofing. By throwing everything in there, we thought maybe we'd find it useful later, e.g. when handling different FK situations or something.

madelynnblue (approved):

dt:
We potentially could use that backed up descriptors table _instead_ of keeping a list of descritors in the backup proto at all, but a) it's much more convenient to be able to run `SHOW BACKUP`  or validate a `RESTORE` spec without fetching and reading SSTs and b) if we wanted to avoid the "leak" of all the unrelated tables, we'd need to do something fancy to backup that table... so I think I like this direction better. /cc @danhhz

PR #19815, opened 2017-11-06
build: makefile cleanup part 1

benesch:
One of the most unmaintainable parts of our build system is the fact that it is split across four files (Makefile, build/common.mk, build/protobuf.mk, ui/Makefile) that reference each other. Merging them will keep me sane.

Take the first step towards this goal, namely, teaching protobuf.mk not to need a CWD one level above the repository root. This has the unfortunate side effect of requiring a path update in every `.proto`; the diff isn't actually that large outside of the automated rename.

PR #19976, opened 2017-11-10
build: merge the makefiles

benesch:
Followup to #19815. Merge all our Makefiles into one. This resolves literally 90% of the problems with our build system, because Make can properly track dependencies that span packages, thus avoiding unnecessary work (in the best case) and build races (in the worst case).

@jordanlewis since you have a good eye for when Make is recompiling things right and when it's not, mind taking this PR for a spin, even if you nope-nope-nope out of the review?


knz:
:lgtm:

(Instead of a big Makefile I still would prefer several smaller bits all `include`d in the main one. But you decide, since you're still spending the most time in here.)

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, 4 of 4 files at r4, 3 of 3 files at r5, 3 of 3 files at r6, 4 of 4 files at r7, 1 of 1 files at r8, 2 of 2 files at r9.

---

*[build/common.mk, line 455 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/19976#-KybR5cIPPWyKKSiSYJN:-KybR5cIPPWyKKSiSYJO:b-ysuced) ([raw file](https://github.com/cockroachdb/cockroach/blob/8fd551525786b6aa0424f44e32bb599bf32c4917/build/common.mk#L455)):*
> ```Makefile
> 	@$(MAKE) --no-print-directory -C $(ROCKSDB_DIR) rocksdb
> 
> # libroach should depend on the generated C++ protobufrs.
> ```

r4: protobufs

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19976#-:-KybRvLpuGah3nMrW93Z:b-3znrtl)*

benesch:
I too prefer smaller Makefiles! It's just much easier to merge them all together first so the prerequisites become clear.

knz:
ACK. Carry on.

benesch:
Inexplicably, I am once again unable to reproduce the macOS Make bug. I've rebased the fix out for now, and I'm digging in further.

benesch:
Thanks for the review, @knz! Onwards and upwards!

PR #19977, opened 2017-11-10
build: add a note that make lint is very slow

justinj:
Not the most visible thing in the world with the build messages, but
it's better than nothing.


benesch:
Feel free to add some color! `$(shell tput setaf 6 2>/dev/null)COLORIZED$(shell tput sgr0 2>/dev/null)`

justinj:
PTAL!

benesch (approved):

justinj:
Thanks!

PR #19988, opened 2017-11-11
sql: IndexedVars no longer need a container ptr

jordanlewis:
Previously, an `IndexedVar` stored a pointer to the
`IndexedVarContainer` that begat it. This is a troublesome violation of
the principle that expression types should be immutable. The trouble
manifests itself during projects that involve copying and modifying
plans, such as plan caching - changing plan nodes that implement the
`IndexedVarContainer` interface is impossible without a full expression
tree walk to reparent the `IndexedVars` onto the new container.

This patch:
- removes the container pointer from `IndexedVar`
- adds a `NodeFormatter` pointer to `IndexedVar`
- adds an `IndexedVarHelper` pointer to `SemaContext`
- adds an `IndexedVarHelper` pointer to `EvalContext`

Callers of `Eval` and `TypeCheck` now have the responsibility of passing
the correct `IndexedVarHelper` via the respective context.

Callers of `Format` don't have to pass an `IndexedVarHelper` because the
`IndexedVar` now carries an AST node from its creation time that can be
`Format`'d to produce its string representation. This was done because
there are too many callsites of `NodeFormatter.Format` to plumb
`IndexedVarHelper` instances to, mainly all of the places where errors
containing expressions are created.

Also, a couple of minor changes to distsql to facilitate this:

- don't bother trying to evaluate a join condition filter if it doesn't exist
- make copies of the `EvalContext` from `FlowCtx` instead of sharing one

cc @nvanbenschoten

### DIFF
--- a/pkg/ccl/sqlccl/csv.go
+++ b/pkg/ccl/sqlccl/csv.go
@@ -1212,7 +1212,7 @@ func newReadCSVProcessor(
 		output:     output,
 		settings:   flowCtx.Settings,
 	}
-	if err := cp.out.Init(&distsqlrun.PostProcessSpec{}, csvOutputTypes, &flowCtx.EvalCtx, output); err != nil {
+	if err := cp.out.Init(&distsqlrun.PostProcessSpec{}, csvOutputTypes, flowCtx.NewEvalCtx(), output); err != nil {
 		return nil, err
 	}
 	return cp, nil
@@ -1367,7 +1367,7 @@ func newSSTWriterProcessor(
 		tempStorage:   flowCtx.TempStorage,
 		settings:      flowCtx.Settings,
 	}
-	if err := sp.out.Init(&distsqlrun.PostProcessSpec{}, sstOutputTypes, &flowCtx.EvalCtx, output); err != nil {
+	if err := sp.out.Init(&distsqlrun.PostProcessSpec{}, sstOutputTypes, flowCtx.NewEvalCtx(), output); err != nil {
 		return nil, err
 	}
 	return sp, nil
--- a/pkg/sql/analyze.go
+++ b/pkg/sql/analyze.go
@@ -1715,12 +1715,14 @@ func (p *planner) analyzeExpr(
 
 	// Type check.
 	var typedExpr tree.TypedExpr
+	p.semaCtx.IVarHelper = &iVarHelper
 	if requireType {
 		typedExpr, err = tree.TypeCheckAndRequire(resolved, &p.semaCtx,
 			expectedType, typingContext)
 	} else {
 		typedExpr, err = tree.TypeCheck(resolved, &p.semaCtx, expectedType)
 	}
+	p.semaCtx.IVarHelper = nil
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/sql/analyze_test.go
+++ b/pkg/sql/analyze_test.go
@@ -59,7 +59,7 @@ func testTableDesc() *sqlbase.TableDescriptor {
 	}
 }
 
-func makeSelectNode(t *testing.T) *renderNode {
+func makeSelectNode(t *testing.T, evalCtx *tree.EvalContext) *renderNode {
 	desc := testTableDesc()
 	sel := testInitDummySelectNode(desc)
 	if err := desc.AllocateIDs(); err != nil {
@@ -67,6 +67,7 @@ func makeSelectNode(t *testing.T) *renderNode {
 	}
 	numColumns := len(sel.sourceInfo[0].sourceColumns)
 	sel.ivarHelper = tree.MakeIndexedVarHelper(sel, numColumns)
+	evalCtx.IVarHelper = &sel.ivarHelper
 	sel.curSourceRow = make(tree.Datums, numColumns)
 	return sel
 }
@@ -84,7 +85,8 @@ func parseAndNormalizeExpr(
 	if expr, _, _, err = sel.resolveNames(expr); err != nil {
 		t.Fatalf("%s: %v", sql, err)
 	}
-	typedExpr, err := tree.TypeCheck(expr, nil, types.Any)
+	semaCtx := tree.SemaContext{IVarHelper: evalCtx.IVarHelper}
+	typedExpr, err := tree.TypeCheck(expr, &semaCtx, types.Any)
 	if err != nil {
 		t.Fatalf("%s: %v", sql, err)
 	}
@@ -142,7 +144,7 @@ func TestSplitOrExpr(t *testing.T) {
 		t.Run(d.expr+"~"+d.expected, func(t *testing.T) {
 			evalCtx := tree.NewTestingEvalContext()
 			defer evalCtx.Stop(context.Background())
-			sel := makeSelectNode(t)
+			sel := makeSelectNode(t, evalCtx)
 			expr := parseAndNormalizeExpr(t, evalCtx, d.expr, sel)
 			exprs := splitOrExpr(evalCtx, expr, nil)
 			if s := exprs.String(); d.expected != s {
@@ -168,7 +170,7 @@ func TestSplitAndExpr(t *testing.T) {
 		t.Run(d.expr+"~"+d.expected, func(t *testing.T) {
 			evalCtx := tree.NewTestingEvalContext()
 			defer evalCtx.Stop(context.Background())
-			sel := makeSelectNode(t)
+			sel := makeSelectNode(t, evalCtx)
 			expr := parseAndNormalizeExpr(t, evalCtx, d.expr, sel)
 			exprs := splitAndExpr(evalCtx, expr, nil)
 			if s := exprs.String(); d.expected != s {
@@ -295,10 +297,11 @@ func TestSimplifyExpr(t *testing.T) {
 		t.Run(d.expr+"~"+d.expected, func(t *testing.T) {
 			evalCtx := tree.NewTestingEvalContext()
 			defer evalCtx.Stop(context.Background())
-			sel := makeSelectNode(t)
+			sel := makeSelectNode(t, evalCtx)
 			// We need to manually close this memory account because we're doing the
 			// evals ourselves here.
 			defer evalCtx.ActiveMemAcc.Close(context.Background())
+			evalCtx.IVarHelper = &sel.ivarHelper
 			expr := parseAndNormalizeExpr(t, evalCtx, d.expr, sel)
 			expr, equiv := simplifyExpr(evalCtx, expr)
 			if s := expr.String(); d.expected != s {
@@ -351,7 +354,7 @@ func TestSimplifyNotExpr(t *testing.T) {
 		t.Run(d.expr+"~"+d.expected, func(t *testing.T) {
 			evalCtx := tree.NewTestingEvalContext()
 			defer evalCtx.Stop(context.Background())
-			sel := makeSelectNode(t)
+			sel := makeSelectNode(t, evalCtx)
 			expr1 := parseAndNormalizeExpr(t, evalCtx, d.expr, sel)
 			expr2, equiv := simplifyExpr(evalCtx, expr1)
 			if s := expr2.String(); d.expected != s {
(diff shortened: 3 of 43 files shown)


knz:
LGTM but please include a before/after benchmark with alloc numbers so we know where we stand (you've quadrupled the size of IndexedVar but reduced the number of allocs, so the effect is not trivial).

---

Reviewed 1 of 1 files at r1, 14 of 14 files at r2, 2 of 29 files at r3.
Review status: 16 of 43 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-:-Kypo6kAJgK5W7-1zIwq:b1809kp)*

knz:
Reviewed 27 of 29 files at r3.

---

*[pkg/sql/parser/scan.go, line 614 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KypoodWVQANQNJAXUVs:-KypoodWVQANQNJAXUVt:b-71dben) ([raw file](https://github.com/cockroachdb/cockroach/blob/1b9046249cd3bbb270e66f1640b764f8794cbe42/pkg/sql/parser/scan.go#L614)):*
> ```Go
> 		lval.str = s.in[start:s.pos]
> 	} else if isASCII {
> 		// We know that the identifier we've Used so far is ASCII, so we don't need
> ```

s/Used/used/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
```
SQL/Cockroach/Update/count=1-8        468µs ± 3%     479µs ± 4%  +2.34%  (p=0.004 n=10+10)
SQL/Cockroach/Update/count=10-8       856µs ±11%     842µs ± 6%    ~     (p=0.931 n=9+9)
SQL/Cockroach/Update/count=100-8     3.89ms ±12%    3.83ms ± 4%    ~     (p=0.796 n=9+9)
SQL/Cockroach/Update/count=1000-8    34.3ms ±10%    33.8ms ± 5%    ~     (p=0.842 n=10+9)

name                               old alloc/op   new alloc/op   delta
SQL/Cockroach/Update/count=1-8       51.2kB ± 1%    52.7kB ± 1%  +2.80%  (p=0.000 n=10+10)
SQL/Cockroach/Update/count=10-8       121kB ± 1%     122kB ± 1%  +1.44%  (p=0.000 n=9+9)
SQL/Cockroach/Update/count=100-8      776kB ± 1%     779kB ± 1%    ~     (p=0.156 n=9+10)
SQL/Cockroach/Update/count=1000-8    6.65MB ± 1%    6.68MB ± 1%    ~     (p=0.079 n=9+10)

name                               old allocs/op  new allocs/op  delta
SQL/Cockroach/Update/count=1-8          574 ±12%       594 ± 8%    ~     (p=0.158 n=10+10)
SQL/Cockroach/Update/count=10-8         948 ± 7%      1016 ±15%  +7.25%  (p=0.026 n=9+10)
SQL/Cockroach/Update/count=100-8      4.42k ±10%     4.60k ±13%    ~     (p=0.434 n=9+10)
SQL/Cockroach/Update/count=1000-8     37.8k ±20%     40.2k ±15%    ~     (p=0.182 n=9+10)
```
Hmm..

knz:
More allocs? I'm surprised.

The overall time/memory overhead is still negligible. That's good enough for me.

jordanlewis:
Review status: 0 of 42 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/parser/scan.go, line 614 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KypoodWVQANQNJAXUVs:-KyqPQEi29CRXk5ck-kg:b-49p4kh) ([raw file](https://github.com/cockroachdb/cockroach/blob/1b9046249cd3bbb270e66f1640b764f8794cbe42/pkg/sql/parser/scan.go#L614)):*

s/Used/used/
</blockquote></details>

Whoops, not sure how that snuck in. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

nvanbenschoten:
Reviewed 1 of 43 files at r4, 14 of 14 files at r5, 26 of 29 files at r6.
Review status: 39 of 42 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-KyqsB59CAJkqwog7hNP:b-m281lo) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*
> ```Go
> 
> // MakeEvalCtx returns a modifiable copy of the FlowCtx's EvalContext.
> func (ctx *FlowCtx) MakeEvalCtx() *parser.EvalContext {
> ```

This should be called `NewEvalContext` to make it clear that it's allocating. It may be where all those new allocations are coming from, although you were running `SQL/Cockroach/Update` so it's probably not using distSQL.

Also, the commit message mentions that sharing a single `EvalContext` is undesirable, but do you mind pointing out where that actually made a difference?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
Review status: 39 of 42 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-KyquSf1PA59IASHunis:b-729k80) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

This should be called `NewEvalContext` to make it clear that it's allocating. It may be where all those new allocations are coming from, although you were running `SQL/Cockroach/Update` so it's probably not using distSQL.

Also, the commit message mentions that sharing a single `EvalContext` is undesirable, but do you mind pointing out where that actually made a difference?
</blockquote></details>

Done.

This will only get called once per processor, which I think is acceptable. The rationale for this change is that individual plan nodes can mutate their `EvalContext` now, in order to pass the correct indexed var container to `Eval`. So either the individual plan nodes need to be the ones to copy the `EvalContext`, or we just blindly copy the `EvalContext` once per processor at the top of the food chain. I chose this way because some processors need to call `Eval` once per row, and making a once per row `EvalContext` copy is not a good idea.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

RaduBerinde:
*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-Kyr8lTpOp14hdDt3WgB:b8uvwb9) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

Done.

This will only get called once per processor, which I think is acceptable. The rationale for this change is that individual plan nodes can mutate their `EvalContext` now, in order to pass the correct indexed var container to `Eval`. So either the individual plan nodes need to be the ones to copy the `EvalContext`, or we just blindly copy the `EvalContext` once per processor at the top of the food chain. I chose this way because some processors need to call `Eval` once per row, and making a once per row `EvalContext` copy is not a good idea.
</blockquote></details>

I think this reasoning for why we might need a mutable copy should be explained in the comment for NewEvalCtx.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
Review status: 0 of 42 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-Kyr9HNzD9X31mvV0onr:b-o27crn) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

I think this reasoning for why we might need a mutable copy should be explained in the comment for NewEvalCtx.
</blockquote></details>

Good idea. Done.

Another idea is replacing this entire patch with a single explicit copy in `ProcOutputHelper.Init`. My fear is that there are other places that might eventually store pointers to the `EvalContext` and accidentally break things.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

RaduBerinde:
*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-KyrLdiEZI3T2oD9p9gu:bfxs6mh) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

Good idea. Done.

Another idea is replacing this entire patch with a single explicit copy in `ProcOutputHelper.Init`. My fear is that there are other places that might eventually store pointers to the `EvalContext` and accidentally break things.
</blockquote></details>

I would put it in `processorBase`. And have a `processorBase.init` function that also calls `out.Init`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
Review status: 0 of 42 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-KyrOyebS27et5e3x3xD:bscipk3) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

I would put it in `processorBase`. And have a `processorBase.init` function that also calls `out.Init`.
</blockquote></details>

I don't quite follow. How would `processorBase.init` know which `ProcOutputHelper` to call `Init` on? Perhaps you are suggesting passing the `ProcOutputHelper` into `processorBase.init`.

That might work nicely. I will try it out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
Review status: 0 of 42 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-KyrQ8q7KQrc4j5fXBR-:bhx7thp) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

I don't quite follow. How would `processorBase.init` know which `ProcOutputHelper` to call `Init` on? Perhaps you are suggesting passing the `ProcOutputHelper` into `processorBase.init`.

That might work nicely. I will try it out.
</blockquote></details>

Oh I see `processorBase` already has the `ProcOutputHelper`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
Review status: 0 of 43 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/distsqlrun/flow.go, line 89 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988#-KyqsB59CAJkqwog7hNO:-KyrSm5gQu97F6HA4Q79:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7a8fb007d38d2bcd572872833d8a56c22e5642ff/pkg/sql/distsqlrun/flow.go#L89)):*

Oh I see `processorBase` already has the `ProcOutputHelper`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

knz:
Reviewed 15 of 43 files at r7, 28 of 28 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19988)*

jordanlewis:
TFTRs!

PR #19989, opened 2017-11-11
cloud: Improve kubernetes pod disruption budget

a-robinson:
Ensure that Kubernetes will try to make sure only one pod gets brought
down at a time regardless of how many pods are in the cluster. This
wasn't an option in the original release of PodDisruptionBudgets, but
has been an option for multiple releases now and more clearly reflects
our intent.

Change the label on the secure client pod to be different from the label
used by the cockroachdb nodes because otherwise it messes with the
enforcement of the disruption budget.

### DIFF
--- a/cloud/kubernetes/cockroachdb-client-secure.yaml
+++ b/cloud/kubernetes/cockroachdb-client-secure.yaml
@@ -3,7 +3,7 @@ kind: Pod
 metadata:
   name: cockroachdb-client-secure
   labels:
-    app: cockroachdb
+    app: cockroachdb-client
 spec:
   initContainers:
   # The init-certs container sends a certificate signing request to the
--- a/cloud/kubernetes/cockroachdb-statefulset-secure.yaml
+++ b/cloud/kubernetes/cockroachdb-statefulset-secure.yaml
@@ -64,7 +64,7 @@ spec:
   selector:
     matchLabels:
       app: cockroachdb
-  minAvailable: 67%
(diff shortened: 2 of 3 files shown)


tschottdorf:
:lgtm:, though if you catch CI, a comment on the disruption policy wouldn't hurt.

Unrelated question out of curiosity: Assuming someone wanted to use kubernetes to "scale down" a cluster automatically by decommissioning pods before they are removed, is that something we can put in the template? I assume there would have to be some hook that has to pass before a pod is actually removed. Looks like https://github.com/kubernetes/kubernetes/issues/25275?

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19989#-:-KydGw99uIQbh83mVSWj:b-bhv33n)*

mberhault (approved):

a-robinson:
> Unrelated question out of curiosity: Assuming someone wanted to use kubernetes to "scale down" a cluster automatically by decommissioning pods before they are removed, is that something we can put in the template? I assume there would have to be some hook that has to pass before a pod is actually removed. Looks like kubernetes/kubernetes#25275?

It'd require some custom scripting, since as you found there isn't a standard signal for that yet, but could certainly be done. For example, we could put a small program into the docker image that gets called as part of the pre-stop hook and examines the StatefulSet it's a part of to determine (using its hostname) whether it's being removed or just restarted.

PR #19996, opened 2017-11-11
Makefile: suppress errors when tput fails

### LINKED ISSUE #19993: Makefile: "@if [ -t 1 ]; then tput setaf ..." breaks make lint
```
% gmake lint TESTS=TestGolint
GOPATH set to /data/home/kena/src/go
Running make with -j16
Running make with -j16
gmake[1]: Nothing to be done for 'build_lib_static'.
[100%] Built target libprotobuf
[100%] Built target snappy
[  0%] Built target build_version
[100%] Built target rocksdb
[100%] Built target roach
[ 86%] Built target roach
[100%] Built target roachccl
GEN pkg/cli/zcgo_flags.go
GEN pkg/server/status/zcgo_flags.go
GEN pkg/storage/engine/zcgo_flags.go
GEN pkg/ccl/storageccl/engineccl/zcgo_flags.go
NOTE: `make lint` is very slow! Perhaps `make lintshort`?
gmake: *** [Makefile:341: lint] Error 1
```

The problem is (I believe) that the last command in the if ... fi block fails (`tput sgr0` is not valid on this system). The following patch fixes:

```
- @if [ -t 1 ]; then tput setaf 3 2>/dev/null; echo 'NOTE: `make lint` is very slow! Perhaps `make lintshort`?'; tput sgr0 2>/dev/null; fi
+ @if [ -t 1 ]; then tput setaf 3 2>/dev/null; echo 'NOTE: `make lint` is very slow! Perhaps `make lintshort`?'; tput sgr0 2>/dev/null; fi || true
```

benesch:
Some platforms only accept the termcap capability names, not the
terminfo ones. Rather than trying to find a compatible invocation, just
suppress errors when tput fails by using command substitution instead of
a command list.

Fix #19993.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,7 +338,7 @@ generate: ## Regenerate generated code.
 .PHONY: lint
 lint: override TAGS += lint
 lint: ## Run all style checkers and linters.
-	@if [ -t 1 ]; then tput setaf 3 2>/dev/null; echo 'NOTE: `make lint` is very slow! Perhaps `make lintshort`?'; tput sgr0 2>/dev/null; fi
+	@if [ -t 1 ]; then echo $$(tput setaf 3 2>/dev/null)'NOTE: `make lint` is very slow! Perhaps `make lintshort`?'$$(tput sgr0 2>/dev/null); fi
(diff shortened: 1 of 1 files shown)


knz (approved):

knz:
Thank you!

PR #19999, opened 2017-11-12
stats: histogram proto and equi-depth construction

RaduBerinde:
#### stats: add proto for histogram

From the SQL stats RFC (#18399).

#### stats: equi-depth histogram

Code to build an equi-depth histogram from a bunch of Datums.


petermattis:
:lgtm:

Is determining selectivity or cardinality coming soon? 

---

Reviewed 5 of 5 files at r1, 2 of 2 files at r2.
Review status: 5 of 7 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/sql/stats/histogram.go, line 40 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylGiE4PhahA8VSKSWw:-KylGiE4PhahA8VSKSWx:b-zfo7s7) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L40)):*
> ```Go
> 	numSamples := len(samples)
> 	if numRows < int64(numSamples) {
> 		return HistogramData{}, errors.Errorf("more samples than rows")
> ```

Add a test for this error case and the one below. 

---

*[pkg/sql/stats/histogram.go, line 48 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFKGmcRDbEw302bSz:-KylFKGmcRDbEw302bT-:b-d0z7ok) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L48)):*
> ```Go
> 	}
> 	sort.Slice(samples, func(i, j int) bool {
> 		return samples[i].Compare(evalCtx, samples[j]) < 0
> ```

I think `evalCtx` is only needed if the datums can contain placeholders. It looks kosher to pass `nil`. Cc @jordanlewis because I think he added this recently. Perhaps he would prefer to keep the `evalCtx` plumbed.

---

*[pkg/sql/stats/histogram.go, line 51 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylEOArpM6S8T_ySCkn:-KylEOArpM6S8T_ySCko:bx7eoxy) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L51)):*
> ```Go
> 	})
> 	numBuckets := maxBuckets
> 	if numSamples < maxBuckets {
> ```

Nit: `numBuckets > numSamples`. I think this makes the subsequent assignment clearer and it maintains consistency in placing `numSamples` on the right side of both statements.

---

*[pkg/sql/stats/histogram.go, line 59 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylGOKDtSmxsF9IzLdK:-KylGOKDtSmxsF9IzLdL:b-r1r438) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L59)):*
> ```Go
> 	// i keeps track of the current sample and advances as we form buckets.
> 	i := 0
> 	for b := 0; b < numBuckets && i < numSamples; b++ {
> ```

You can declare `i` in the for-loop:

```
for b, i := 0, 0; b < numBuckets && i < numSamples; b++ {
```

---

*[pkg/sql/stats/histogram.proto, line 27 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylDu694qajPS8uQrUl:-KylDu694qajPS8uQrUm:bzgjfus) ([raw file](https://github.com/cockroachdb/cockroach/blob/6212576988b10f76e36ec87ec75c0165f86284f8/pkg/sql/stats/histogram.proto#L27)):*
> ```protobuf
> message HistogramData {
>   message Bucket {
>     // The estimated number of rows that are equal to upper_bound.
> ```

Histograms are over column values, not rows, right? Should this be `s/rows/values/g`?

---

*[pkg/sql/stats/histogram_test.go, line 153 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylEoBjBvecFjm9g764:-KylEoBjBvecFjm9g765:b-41agdb) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram_test.go#L153)):*
> ```Go
> 				val := tc.samples[perm[i]]
> 
> 				samples[i] = tree.NewDInt(tree.DInt(val))
> ```

Probably worthwhile to test some of the other data types.

---

*[pkg/sql/stats/histogram_test.go, line 165 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFsLTfafhqbJwHkhZ:-KylFsLTfafhqbJwHkh_:b-hn03gs) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram_test.go#L165)):*
> ```Go
> 			}
> 			for i, b := range h.Buckets {
> 				_, val, err := encoding.DecodeVarintAscending(b.UpperBound)
> ```

Rather than decoding the bucket `UpperBound`, you could encode `exp.upper` and use `bytes.Compare` which would ease testing of other datum types. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-:-KylGEeZNZccj6al9TI0:bx2r5j0)*

RaduBerinde:
TFTR! Cardinality is coming soon. Focusing on the infrastructure to get stats calculated right now. Using the stats for selectivity will come after.

---

Review status: 3 of 7 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/sql/stats/histogram.go, line 40 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylGiE4PhahA8VSKSWw:-KylU_d_j_Q0ctEef6Zx:b-c41q33) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L40)):*

Add a test for this error case and the one below. 
</blockquote></details>

Added but they feel unnecessary to me. Serves me right for not making them panics :)

---

*[pkg/sql/stats/histogram.go, line 48 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFKGmcRDbEw302bSz:-KylUtxv0v92odh-7SsC:bw8s6ev) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L48)):*

I think `evalCtx` is only needed if the datums can contain placeholders. It looks kosher to pass `nil`. Cc @jordanlewis because I think he added this recently. Perhaps he would prefer to keep the `evalCtx` plumbed.
</blockquote></details>

Not sure, it might be needed for comparing certain types (thinking of time/date types where the local timezone might matter?). I'd rather keep it plumbed just in case

---

*[pkg/sql/stats/histogram.go, line 51 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylEOArpM6S8T_ySCkn:-KylXl1rwbpVIAb865qn:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L51)):*

Nit: `numBuckets > numSamples`. I think this makes the subsequent assignment clearer and it maintains consistency in placing `numSamples` on the right side of both statements.
</blockquote></details>

Done.

---

*[pkg/sql/stats/histogram.go, line 59 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylGOKDtSmxsF9IzLdK:-KylXkoGDrHdLI3XyQTz:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L59)):*

You can declare `i` in the for-loop:

```
for b, i := 0, 0; b < numBuckets && i < numSamples; b++ {
```
</blockquote></details>

Done.

---

*[pkg/sql/stats/histogram.proto, line 27 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylDu694qajPS8uQrUl:-KylVqHX187TdC4BrKE7:b-a2ujfl) ([raw file](https://github.com/cockroachdb/cockroach/blob/6212576988b10f76e36ec87ec75c0165f86284f8/pkg/sql/stats/histogram.proto#L27)):*

Histograms are over column values, not rows, right? Should this be `s/rows/values/g`?
</blockquote></details>

Changed.

---

*[pkg/sql/stats/histogram_test.go, line 153 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylEoBjBvecFjm9g764:-KylXLBweR586-SgYCUU:b-tqit3w) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram_test.go#L153)):*

Probably worthwhile to test some of the other data types.
</blockquote></details>

The only thing the code uses from Datums is CompareTo.. I think using one type is enough to test this piece of code for now; things will be different when we implement max-diff (which only works for certain types).

---

*[pkg/sql/stats/histogram_test.go, line 165 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFsLTfafhqbJwHkhZ:-KylXCmWyd0P2ugDBPGb:bihz8o3) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram_test.go#L165)):*

Rather than decoding the bucket `UpperBound`, you could encode `exp.upper` and use `bytes.Compare` which would ease testing of other datum types. 
</blockquote></details>

How does that help? We would need to have different "encode" calls per type instead of "decode".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-:-KylXuT6ptiVr02qNFq7:b28s1m1)*

petermattis:
Review status: 3 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/stats/histogram.go, line 48 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFKGmcRDbEw302bSz:-KylkCOsP_IYrpooowpJ:b-1l4un3) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L48)):*

Not sure, it might be needed for comparing certain types (thinking of time/date types where the local timezone might matter?). I'd rather keep it plumbed just in case
</blockquote></details>

Ah, good point.

---

*[pkg/sql/stats/histogram_test.go, line 165 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFsLTfafhqbJwHkhZ:-KyljxnagCgzZECSQZy0:b-9v3z8q) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram_test.go#L165)):*

How does that help? We would need to have different "encode" calls per type instead of "decode".
</blockquote></details>

Doesn't `EncodeTableKey` do the right thing if passed a `Datum`? I don't feel strongly about making a change here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999)*

jordanlewis:
Review status: 3 of 7 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/stats/histogram.go, line 48 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFKGmcRDbEw302bSz:-KylkL1BDUS8AY4D0GFt:b92cl8o) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram.go#L48)):*

Ah, good point.
</blockquote></details>

Better to keep it plumbed anyway I think. Soon IndexedVar evaluation will need the EvalContext as well - I don't have the full context of this right now but even if it can't receive anything but Datums, it's good habit to make sure every part of the code that touches a Datum has a plumbed EvalContext.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999)*

RaduBerinde:
*[pkg/sql/stats/histogram_test.go, line 165 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KylFsLTfafhqbJwHkhZ:-Kym_LgMkHHoFXLF77Mv:b6sluog) ([raw file](https://github.com/cockroachdb/cockroach/blob/eaf2a7803e4d4ee37400981dbf7fa7e10b4ecabb/pkg/sql/stats/histogram_test.go#L165)):*

Doesn't `EncodeTableKey` do the right thing if passed a `Datum`? I don't feel strongly about making a change here.
</blockquote></details>

Good point. I started doing it, but realized we still need to Decode if we're going to have a useful Errorf message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999)*

knz:
:lgtm:

---

Reviewed 5 of 5 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 2 of 4 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-:-KypllFULAV-XXVi5b2V:bnfp4nl)*

rytaft:
:lgtm:

---

---

*[pkg/sql/stats/histogram.go, line 59 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-Kypswo9qUN7Zw4ftoIk:-Kypswo9qUN7Zw4ftoIl:b-jg4iih) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram.go#L59)):*
> ```Go
> 	// i keeps track of the current sample and advances as we form buckets.
> 	for i, b := 0, 0; b < numBuckets && i < numSamples; b++ {
> 		num := (numSamples - i) / (numBuckets - b)
> ```

Minor thing - it would be helpful to add a comment here saying that `num` is the number of samples in this bucket

---

*[pkg/sql/stats/histogram.go, line 64 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-Kypt9RP0-dcCjUc2lER:-Kypt9RP0-dcCjUc2lES:bd779gv) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram.go#L64)):*
> ```Go
> 		}
> 		upper := samples[i+num-1]
> 		numLess := 0
> ```

Would also help to add a comment here to say `numLess` is the number of samples less than `upper`

---

*[pkg/sql/stats/histogram.proto, line 38 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KyptaGKny1TWzOPQ8CN:-KyptaGKny1TWzOPQ8CO:bytkmvs) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram.proto#L38)):*
> ```protobuf
>     // single value. This might be particularly advantageous if the
>     // histogram algorithm makes sure the top "heavy hitters" (most
>     // frequent elements) are bucket boundaries (similar of a
> ```

^ similar to a

---

*[pkg/sql/stats/histogram_test.go, line 103 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KypuNa-yvLNw8VYEUDd:-KypuNa-yvLNw8VYEUDe:b-6d9p0z) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram_test.go#L103)):*
> ```Go
> 			buckets: []expBucket{
> 				{
> 					// Bucket contains everything.
> ```

fix this comment

---

*[pkg/sql/stats/histogram_test.go, line 107 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KypuP7xcyaV1QCTCv6e:-KypuP7xcyaV1QCTCv6f:b-wg78tq) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram_test.go#L107)):*
> ```Go
> 				},
> 				{
> 					// Bucket contains everything.
> ```

and this one

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-:-KypvpBkv-Kof6wcy1Ay:bnfp4nl)*

rytaft:
Reviewed 5 of 5 files at r1, 2 of 4 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999)*

RaduBerinde:
TFTRs!

---

Review status: 3 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[pkg/sql/stats/histogram.go, line 59 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-Kypswo9qUN7Zw4ftoIk:-Kyq95DW_VbE6kXHbOQ4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram.go#L59)):*

Minor thing - it would be helpful to add a comment here saying that `num` is the number of samples in this bucket
</blockquote></details>

Done.

---

*[pkg/sql/stats/histogram.go, line 64 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-Kypt9RP0-dcCjUc2lER:-Kyq9E2xW6hbdsDMpNaJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram.go#L64)):*

Would also help to add a comment here to say `numLess` is the number of samples less than `upper`
</blockquote></details>

Done.

---

*[pkg/sql/stats/histogram.proto, line 38 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KyptaGKny1TWzOPQ8CN:-Kyq9HjIWZNdDgxgy7jo:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram.proto#L38)):*

^ similar to a
</blockquote></details>

Done.

---

*[pkg/sql/stats/histogram_test.go, line 103 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KypuNa-yvLNw8VYEUDd:-Kyq9S5dCqoqHU19ISYq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram_test.go#L103)):*

fix this comment
</blockquote></details>

Done.

---

*[pkg/sql/stats/histogram_test.go, line 107 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-KypuP7xcyaV1QCTCv6e:-Kyq9STkl6uIV4Sf0qMS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3dc1cdf5d825f494772ebbef7a6b598a2d473725/pkg/sql/stats/histogram_test.go#L107)):*

and this one
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999#-:-Kyq9ZH7aiEovuWq0xeV:bz5mair)*

rytaft:
Reviewed 2 of 4 files at r6, 2 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/19999)*

PR #20001, opened 2017-11-13
sqlbase: fix encoding of index value when using unspecified columns

### LINKED ISSUE #20000: Cockroach server crashes on certain UPSERT
**BUG REPORT**

1.
Cockroach version: 1.1.2
Issues happens with all variants: insecure single node, secure cluster

2. 
I have following table definition:
```sql
CREATE TABLE recipes (
  hash          STRING(32) PRIMARY KEY,
  source_url    STRING(512),
  publisher     STRING(32)  NOT NULL,
  name          STRING(128) NOT NULL,
  language      STRING(5),
  image_url     STRING,
  image_url_md5 STRING(32),
  source_parser STRING(255),
  body          STRING,
  properties    STRING,
  noindex       BOOLEAN,
  disabled      BOOLEAN,
  product_analysis STRING,
  product_analysis_coverage NUMERIC(2,2),
  product_analysis_updated_at TIMESTAMP,
  created_at    TIMESTAMP DEFAULT CURRENT_TIMESTAMP(),
  updated_at    TIMESTAMP DEFAULT CURRENT_TIMESTAMP(),
  INDEX idx_recipe_analysis_metadata_coverage(product_analysis_coverage),
  INDEX idx_recipe_publisher(publisher),
  INDEX idx_recipe_updated_at(updated_at DESC)
);
```
then when I run query like below, server crashes. In cluster scenario, it crashes single node, while remaining keep operating. I see nothing in the logs.
```
upsert into recipes(hash, source_url, publisher, name, language, image_url, image_url_md5, source_parser, body, properties, noindex, updated_at)
                     values(
                     '5bf587d3ac3b4714b8845ebdfcfb4f17',
                     'http://example.com/recipe',
                     'example.com',
                     'Name',
                     'en',
                     'http://image.com',
                     'f12e4da141d5fa6f5506ee0502ce6f5d',
                     'sp',
                     '{}',
                     '{}',
                     null,
                     '2017-11-12 23:42:52.101000'
                     );
```

vivekmenezes:
When looking up a map using a key not present in the map,
the map will return the zero value of the map value type.

Release note: fix encoding of column stored in index using STORING. This was broken! 

fixes #20000

I didn't find a place where I could test this in table_test. I could always add a logictest.


lego:
The fix LGTM but while the commit says why the missing map key is a problem, it doesn't mention why the key is missing in the first place. (My guess is when a value is explicitly not provided and the column is nullable, but I could only figure that out from the linked issue and the `EncodeSecondaryIndex` comment.)

Personally, I still have some FUD on the problem: is there anywhere else that assumes columns are provided (e.g. `StoreColumns` and `CompositeColumns`)? If so, should the fix instead be adding NULL datums earlier on? That seems appropriate if the issue is only caused by row insertion because if a secondary index update procedure mistakenly misses values then they will be silently dropped.

As for testing the index encoding directly, you can add them to https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/sqlbase/table_test.go#L129. You'll need to modify `makeTableDescForTest ` and `indexKeyTest` to add `StoreColumnIDs` and `CompositeColumnIDs` so you can specify and test them.

Unless I'm wrong, this problem can also happen if columns omitted includes ones stored in secondary indexes (with `STORING`). Could there also be any possible difference between upsert and insert too? (I'm not familiar with the upsert code)

I will sleep better if you add all of those tests (:

---

Reviewed 2 of 2 files at r1.

---

*[pkg/sql/logictest/testdata/logic_test/collatedstring_nullinindex, line 10 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-KyngFDW0DGSw2a9s-kh:-KyngFDW0DGSw2a9s-ki:b4wh36) ([raw file](https://github.com/cockroachdb/cockroach/blob/82419c24cb7aaa2556477ca2ecc2a3ffa7e6a3f3/pkg/sql/logictest/testdata/logic_test/collatedstring_nullinindex#L10)):*
> ```
> 
> statement ok
> INSERT INTO t VALUES (1, 'foo' COLLATE en), (2, NULL), (3, 'bar' COLLATE en)
> ```

Why remove this? Clearly, there are different mechanisms at play when a value is provided as NULL and when it is not provided at all I think this should still be tested.

---

*[pkg/sql/sqlbase/table.go, line 1511 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-KyngHzZjSQGE8qFJngD:-KyngHzZjSQGE8qFJngE:b-94kllp) ([raw file](https://github.com/cockroachdb/cockroach/blob/8775b25fbdf0c5f4a5307f86c0c08ab6461c5a67/pkg/sql/sqlbase/table.go#L1511)):*
> ```Go
> 	for _, col := range cols {
> 		valIdx, isSet := colMap[col.id]
> 		if !isSet {
> ```

I think there should be a comment here clarifying why there is a check for both if the key exists and if the value is null.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KyneIxGj_ayIlezBLqA:b-gj4ix3)*

knz:
I side with Joey's review, it's spot on.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KypgodGeyDaeBffZSfU:bp3ljz4)*

vivekmenezes:
Thank you for your review comments @lego ! I really appreciate them.

Turn out I've uncovered a very bad bug which is triggered when using unspecified values for storing columns. We were creating a bad encoding for the index value!

I've confined all the tests to logictests. I'm glad we found this.

jordanlewis:
Nice! What was the bug exactly? What bad data was written? It's not clear from the logic tests what was broken before. I ran some of the test cases on master and things seem to behave as expected.

```
root@:26257/test> create table z (a int primary key, b decimal);
CREATE TABLE

Time: 5.670246ms

root@:26257/test> create index zzz on z(b);
CREATE INDEX

Time: 56.671355ms

root@:26257/test> insert into z values(1, .2);
INSERT 1

Time: 1.688691ms

root@:26257/test> insert into z values (2);
INSERT 1

Time: 2.993558ms

root@:26257/test> INSERT INTO z VALUES (3, .400);
INSERT 1

Time: 3.204228ms

root@:26257/test> SHOW KV TRACE FOR SELECT * FROM z@zzz;
+----------------------------------+------------+----------------------------------------+---------------------------------------+------------------+-------+
|            timestamp             |    age     |                message                 |                context                |    operation     | span  |
+----------------------------------+------------+----------------------------------------+---------------------------------------+------------------+-------+
| 2017-11-13 22:34:24.632962+00:00 | 0s         | Scan /Table/54/{2-3}                   | [client=127.0.0.1:51349,user=root,n1] | sql txn implicit | (0,0) |
| 2017-11-13 22:34:24.63299+00:00  | 27µs815ns  | querying next range at /Table/54/2     | [client=127.0.0.1:51349,user=root,n1] | sql txn implicit | (0,0) |
| 2017-11-13 22:34:24.633021+00:00 | 58µs475ns  | r33: sending batch 1 Scan to (n1,s1):1 | [client=127.0.0.1:51349,user=root,n1] | sql txn implicit | (0,0) |
| 2017-11-13 22:34:24.633196+00:00 | 233µs915ns | fetched: /z/zzz/NULL/2 -> NULL         | [client=127.0.0.1:51349,user=root,n1] | sql txn implicit | (0,0) |
| 2017-11-13 22:34:24.633204+00:00 | 241µs474ns | output row: [2 NULL]                   | [client=127.0.0.1:51349,user=root,n1] | consuming rows   | (0,2) |
| 2017-11-13 22:34:24.63322+00:00  | 257µs426ns | fetched: /z/zzz/0.2/1 -> NULL          | [client=127.0.0.1:51349,user=root,n1] | sql txn implicit | (0,0) |
| 2017-11-13 22:34:24.633224+00:00 | 261µs754ns | output row: [1 0.2]                    | [client=127.0.0.1:51349,user=root,n1] | consuming rows   | (0,2) |
| 2017-11-13 22:34:24.633236+00:00 | 273µs951ns | fetched: /z/zzz/0.4/3/b -> /0.400      | [client=127.0.0.1:51349,user=root,n1] | sql txn implicit | (0,0) |
| 2017-11-13 22:34:24.63324+00:00  | 277µs413ns | output row: [3 0.400]                  | [client=127.0.0.1:51349,user=root,n1] | consuming rows   | (0,2) |
+----------------------------------+------------+----------------------------------------+---------------------------------------+------------------+-------+
(9 rows)
```

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/table_test.go, line 255 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-KyrUc55VvVM-P2VTqyG:-KyrUc55VvVM-P2VTqyH:b-mttsjm) ([raw file](https://github.com/cockroachdb/cockroach/blob/037307e584584ef96088d856d98d6ae9b5990a49/pkg/sql/sqlbase/table_test.go#L255)):*
> ```Go
> 
> ```

accidental new lines added

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KyrVCQb7MU6-F7GYzSz:bmisq94)*

jordanlewis:
Oh and the select itself seems to be okay as well:

```
root@:26257/test> select * from z;
+---+-------+
| a |   b   |
+---+-------+
| 1 |   0.2 |
| 2 | NULL  |
| 3 | 0.400 |
+---+-------+
(3 rows)

root@:26257/test> select * from z@zzz order by b;
+---+-------+
| a |   b   |
+---+-------+
| 2 | NULL  |
| 1 |   0.2 |
| 3 | 0.400 |
+---+-------+
(3 rows)
```

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KyrWZDMRODwF9Kz9zuF:b-molb5d)*

vivekmenezes:
Run the new test I added to storing on master

lego:
@jordanlewis the problem is when doing inserts that explicitly specify the columns but omit some. e.g. this is problematic by specifying only `a` and omitting `b`.

```sql
CREATE TABLE z (a INT PRIMARY KEY, b DECIMAL);
INSERT INTO t (a) VALUES (4);
```

But on the other hand, this works fine.
```sql
INSERT INTO t VALUES (4);
```

lego:
I'm happy with the fix as is for solving the bug report, and I think it should be cherry-picked for 1.1.x.

I'm still not confident this is the best solution, as I'm concerned other places may be using a colIdxMap with the assumption that all the columns they use are present. I do think that will be a more involved change though and should be left for another PR. The two ways I see it:
- We could change the colIdxMap to type alias of the map in order to strictly ensure this doesn't happen, where we get the zero value
- We ensure unspecified columns are added with NULLs. I believe this is somewhere in the insertNode but I wasn't quite sure where. (e.g. [processColumns](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/insert.go#L472-L508)

Thoughts?

---

Reviewed 1 of 5 files at r2, 3 of 4 files at r3.
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KyuzFYXA0d68hVhJcM_:b-uj4d8g)*

vivekmenezes:
TFTR!

I'd like to explore part 1 that you mention of using a different type for colIdxMap throughout the code.

Regarding part 2 I already considered it but the more I looked the scarier the change looked. In particular I was unsure
if I was going to break something. I think it's beyond the scope of this PR.

---

Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KyvGPQQoBJ23CXiTRtj:b-yu1538)*

vivekmenezes:
Hey I'm not keen on doing 1 too. I could create an interface that hides the map behind it, but then I'll have to create methods for populating the map, etc. Doesn't feel like the right thing to do.

---

Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-KyvfaGbUWEML34I8rk2:bi8w44p)*

vivekmenezes:
Any more comments on this PR? Thanks!

---

Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-Kz-9R_3m94dgSBk8XJ6:b-n7l5ba)*

knz:
LGTM

---

Reviewed 1 of 5 files at r2, 3 of 4 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-Kz-Eix4e433wturAQ1A:bb6ywmx)*

vivekmenezes:
TYFTR!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20001#-:-Kz-GLYCDmYI08tgwKJg:bzftgo7)*

PR #20002, opened 2017-11-13
makefile: gen before archive, prefix paths

dt:
Reported in gitter: The explicit attempt to archive embedded.go failed if it didn't exist yet (i.e. clean checkout).
Also it looked like the explicit paths added to the archive were missing the remapping done to the ls-paths? Pattern-matched off that one, not sure if that's the right thing to do though?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -374,11 +374,11 @@ $(ARCHIVE): $(ARCHIVE).tmp
 # instead of scripts/ls-files.sh once Git v2.11 is widely deployed.
 .INTERMEDIATE: $(ARCHIVE).tmp
 $(ARCHIVE).tmp: ARCHIVE_BASE = cockroach-$(shell cat .buildinfo/tag)
-$(ARCHIVE).tmp: .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
+$(ARCHIVE).tmp: $(GOBINDATA_TARGET) $(SQLPARSER_TARGETS) $(BUILDINFO)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -374,11 +374,11 @@ $(ARCHIVE): $(ARCHIVE).tmp
 # instead of scripts/ls-files.sh once Git v2.11 is widely deployed.
 .INTERMEDIATE: $(ARCHIVE).tmp
 $(ARCHIVE).tmp: ARCHIVE_BASE = cockroach-$(shell cat .buildinfo/tag)
-$(ARCHIVE).tmp: .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
+$(ARCHIVE).tmp: $(GOBINDATA_TARGET) $(SQLPARSER_TARGETS) .buildinfo/tag .buildinfo/rev .buildinfo/basebranch
```
knz: Use $(BUILDINFO) here while you're at it.

knz (approved):
LGTM with nit

benesch (approved):
Thanks!

More generally, our source archives are pretty broken and require NodeJS/Yarn to build despite packaging embedded.go (even with this bug fixed). I've been meaning to get verify-archive working again (https://github.com/cockroachdb/cockroach/pull/19506) so they're automatically verified but haven't had the time.

Makefile
dt: Done.

PR #20004, opened 2017-11-13
sqlccl: add partitioning integration test

benesch:
Add an integration test for partitioning. For both list and range
partitionings, the test creates a partitioned table in a three node
cluster, applies zone configs to two of the partitions that constrain
said partition to a particular node, then verifies that reads within one
partition read only from the configured node.

XXX TODO XXX: This currently fails the range partitioning test due to what seems to be a legitimate bug in `GenerateSubzoneSpans`. By my debugging it looks like we're generating overlapping spans for the two partitions in `data.range`:

```
INDEX data.range@primary PARTITION dc1 subzone span: /Table/1 to /Table/1/4
I171113 14:39:30.893051 674 storage/replica_command.go:1395  [split,n2,s2,r21/2:/{Table/51/1/21-Max}] initiating a split of this range at key /Table/52 [r31]
INDEX data.range@primary PARTITION dc2 subzone span: /Table/1 to /Table/1/8
```

### DIFF
--- a/pkg/ccl/sqlccl/partition_test.go
+++ b/pkg/ccl/sqlccl/partition_test.go
@@ -1,0 +1,172 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package sqlccl_test
+
+import (
+	"bytes"
+	gosql "database/sql"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/cockroach/pkg/config"
+	"github.com/cockroachdb/cockroach/pkg/roachpb"
+	"github.com/cockroachdb/cockroach/pkg/testutils"
+	"github.com/cockroachdb/cockroach/pkg/testutils/sqlutils"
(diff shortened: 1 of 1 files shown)


benesch:
@danhhz this is ready for your review!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20004#-:-Kz-HReaazVWSZiyLOES:b-u6rylc)*

danhhz:
:lgtm:

I have some ideas for more tests to add here, but I like the idea of getting _something_ in and letting it grow organically for a bit

---

---

*[pkg/ccl/sqlccl/partition_test.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20004#-Kz-L8q2EKY4ICnOCenL:-Kz-L8q2EKY4ICnOCenM:b-k958m) ([raw file](https://github.com/cockroachdb/cockroach/blob/e1d0f6ea9d0d932f04ac6d643961366079ecd507/pkg/ccl/sqlccl/partition_test.go#L133)):*
> ```Go
> 				PARTITION dc2 VALUES < 8
> 			)`,
> 			scans: []testScan{{"a = 3", "n2"}, {"a = 7", "n3"}},
> ```

a = 4 for the edge case

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20004#-:-Kz-LxT-S-O2yHUWK8nV:b-nlrk8k)*

benesch:
Thanks!

---

---

*[pkg/ccl/sqlccl/partition_test.go, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20004#-Kz-L8q2EKY4ICnOCenL:-Kz030Td983mkq_fRtj1:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e1d0f6ea9d0d932f04ac6d643961366079ecd507/pkg/ccl/sqlccl/partition_test.go#L133)):*

a = 4 for the edge case
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20004#-:-Kz031_F0TglqOXLyDfL:b-ucfknh)*

PR #20006, opened 2017-11-13
kv: fix local server optimization with --advertise-addr

### LINKED ISSUE #19991: kv: local server optimization is brittle
In https://github.com/cockroachdb/cockroach/issues/17611, I realized that the local server optimization path was *never* taken. Investigating, I found that the nodes were running without `--host` but with `--advertise-host`. As a result, all outgoing RPCs were addressed to `<hostname>.crdb.io` (from `--advertise-host`) but the node identified itself with its default host flag, `<hostname>`. This was "fixed" by using `--host <hostname>.crdb.io` instead of `--advertise-host`, but this mechanism shouldn't be so easy to accidentally break.

It may also not even be possible for everyone to set this up this way. For example, it's at the very least super annoying that with the change, the process doesn't listen on localhost any more.

tbg:
We were using `Addr` to match against our local addr, but should have been using AdvertiseAddr.

Fixes #19991.

### DIFF
--- a/pkg/rpc/context_test.go
+++ b/pkg/rpc/context_test.go
@@ -110,6 +110,34 @@ func (*internalServer) Batch(
 	return nil, nil
 }
 
+// TestInternalServerAddress verifies that RPCContext uses AdvertiseAddr, not Addr, to
+// determine whether to apply the local server optimization.
+//
+// Prevents regression of https://github.com/cockroachdb/cockroach/issues/19991.
+func TestInternalServerAddress(t *testing.T) {
+	defer leaktest.AfterTest(t)()
+
+	stopper := stop.NewStopper()
+	defer stopper.Stop(context.TODO())
+
+	// Can't be zero because that'd be an empty offset.
+	clock := hlc.NewClock(timeutil.Unix(0, 1).UnixNano, time.Nanosecond)
+
+	baseCtx := testutils.NewNodeTestBaseContext()
+
+	baseCtx.Addr = "127.0.0.1:9999"
+	baseCtx.AdvertiseAddr = "127.0.0.1:8888"
+
+	serverCtx := NewContext(log.AmbientContext{Tracer: tracing.NewTracer()}, baseCtx, clock, stopper)
+
+	internal := &internalServer{}
+	serverCtx.SetLocalInternalServer(internal)
+
+	if is := serverCtx.GetLocalInternalServerForAddr(baseCtx.AdvertiseAddr); is != internal {
+		t.Fatalf("expected %+v, got %+v", internal, is)
+	}
+}
+
 // TestHeartbeatHealth verifies that the health status changes after
 // heartbeats succeed or fail.
 func TestHeartbeatHealth(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/rpc/context_test.go
```diff
@@ -110,6 +110,38 @@ func (*internalServer) Batch(
 	return nil, nil
 }
 
+// TestInternalServerAddress verifies that RPCContext uses AdvertiseAddr, not Addr, to
+// determine whether the apply the local server optimization.
```
a-robinson: s/the apply/to apply/ if you have to burn another CI cycle

pkg/rpc/context_test.go
```diff
@@ -110,6 +110,38 @@ func (*internalServer) Batch(
...
+
+	baseCtx.Addr = "127.0.0.1:9999"
+	baseCtx.AdvertiseAddr = "127.0.0.1:8888"
+
+	serverCtx := NewContext(log.AmbientContext{Tracer: tracing.NewTracer()}, baseCtx, clock, stopper)
+
+	internal := &internalServer{}
+	serverCtx.SetLocalInternalServer(internal)
+
+	if is := serverCtx.GetLocalInternalServerForAddr(baseCtx.Addr); is != nil {
```
a-robinson: It strikes me that this test case is kind of the opposite of what we want. We don't actually expect `baseCtx.Addr` to be the target of any traffic, but if it was I don't know why we wouldn't want to use the local server optimization for it.

More of an observation than anything, though, given that we don't expect it to be the target of real traffic.

a-robinson (approved):
LGTM

pkg/rpc/context_test.go
```diff
@@ -215,7 +248,10 @@ func TestHeartbeatHealth(t *testing.T) {
 
 	clientCtx.SetLocalInternalServer(&internalServer{})
 
-	if err := clientCtx.ConnHealth(clientCtx.Addr); err != nil {
+	if err := clientCtx.ConnHealth(clientCtx.Addr); err != ErrNotConnected {
+		t.Error(err)
```
nvanbenschoten: Switch this to a `t.Errorf` and explain that the error should have been `ErrNotConnected`, otherwise, if it returns no error, the test failure will be very unclear.

nvb (approved):
Nice find!

tschottdorf:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/rpc/context_test.go, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20006#-Kyq4kDc2ainQ6SpSpxe-r1-114:-KyqOmWHi90z8zVt398D:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/cb1d6ab599085045ba6d481e581fffe0d1ecae6d/pkg/rpc/context_test.go#L114)):*

s/the apply/to apply/ if you have to burn another CI cycle
</blockquote></details>

Done.

---

*[pkg/rpc/context_test.go, line 136 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20006#-Kyq4kDc2ainQ6SpSpxe-r1-136:-KyqOq3Dc3OAl3Wc9y6T:bfcqqov) ([raw file](https://github.com/cockroachdb/cockroach/blob/cb1d6ab599085045ba6d481e581fffe0d1ecae6d/pkg/rpc/context_test.go#L136)):*

It strikes me that this test case is kind of the opposite of what we want. We don't actually expect `baseCtx.Addr` to be the target of any traffic, but if it was I don't know why we wouldn't want to use the local server optimization for it.

More of an observation than anything, though, given that we don't expect it to be the target of real traffic.
</blockquote></details>

Agreed, removed.

---

*[pkg/rpc/context_test.go, line 252 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20006#-Kyq4kDc2ainQ6SpSpxe-r1-252:-KyqbPvNGsyhihhasNjr:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/cb1d6ab599085045ba6d481e581fffe0d1ecae6d/pkg/rpc/context_test.go#L252)):*

Switch this to a `t.Errorf` and explain that the error should have been `ErrNotConnected`, otherwise, if it returns no error, the test failure will be very unclear.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20006)*

PR #19972, opened 2017-11-10
cloud: Update Kubernetes configs to use the init command

a-robinson:
This removes a (very rare) potential edge case where a cluster could
end up in split brain if all pods went down, cockroachdb-0 lost its
disk, and the other pods didn't.

The docs will need to be updated in lockstep with this change.

Fixes #19954

PR #20011, opened 2017-11-13
 cloud: Use a Job yaml file for k8s initialization in insecure mode

a-robinson:
To better match what we do in secure mode. As requested by @bdarnell on https://github.com/cockroachdb/cockroach/pull/19972#pullrequestreview-76166750

Also, while I'm here:

**cloud: consistently use hyphens instead of underscores in k8s files**

This is different from the rest of our code base, but going the other
way would break any existing links to the more popular files
(cockroachdb-statefulset.yaml and cockroachdb-statefulset-secure.yaml).
This also matches the file naming style of all the Kubernetes examples:
https://github.com/kubernetes/kubernetes/tree/master/examples

While I'm renaming things, also rename cockroachdb-client-secure.yaml to
just client-secure.yaml -- the "cockroachdb" doesn't really clarify
anything and muddles up tab completion.

cc @jseldess


mberhault (approved):
LGTM. Thanks for the cleanup.

PR #20015, opened 2017-11-13
sql,sqlccl: handle multiple subzone configs with no table config

benesch:
Previously, executing

    ALTER TABLE t PARTITION p0 EXPERIMENTAL CONFIGURE ZONE 'yaml'
    ALTER TABLE t PARTITION p1 EXPERIMENTAL CONFIGURE ZONE 'yaml'

when table t did not have its own zone config would result in a zone
config on only p1, instead of on both p0 and p1. This was because the
second statement would blow away the existing subzone config on p0.

### DIFF
--- a/pkg/ccl/sqlccl/zone_test.go
+++ b/pkg/ccl/sqlccl/zone_test.go
@@ -80,6 +80,11 @@ func TestValidIndexPartitionSetShowZones(t *testing.T) {
 		CLISpecifier: "d.t.p0",
 		Config:       zoneOverride,
 	}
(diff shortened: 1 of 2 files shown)


danhhz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20015#-:-KyqaxIR82bwwihu8MtL:bnfp4nl)*

benesch:
Thanks for the insanely fast turnaround!

PR #20023, opened 2017-11-13
sqlccl: only allow RESTORE * from BACKUP *

dt:
This records in the backup metadata the databases for which all tables
were selected for backup, e.g. via `db.*` or `DATABASE db`. Only those
databases are eligible to be used in all-table expansions like `db.*`
when restoring from that backup.

Note: Backups created before this is recorded will not have these IDs,
meaning it is not possible to determine which databases have all their
tables backed up. Initially, we'll skip checking that expansions are OK
-- thus for older backups, we'll still allow "restoring" a partial DB
via `RESTORE DATABASE` if that is all that is in the backup.

Release Note: Enterprise: Only allow `RESTORE DATABASE` when backup contains whole database.


mjibson:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[pkg/ccl/sqlccl/restore.go, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-KyrQTyoC3TPHQOl3qnk:-KyrQTyoC3TPHQOl3qnl:b2sqa5o) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d13f07f73719a6b1eeb956f858c6a6a86dde4fb/pkg/ccl/sqlccl/restore.go#L97)):*
> ```Go
> 	}
> 
> 	if lastBackupDesc.FormatVersion > BackupFormatInitialVersion {
> ```

Is `>= BackupFormatDescriptorTrackingVersion` clearer?

---

*[pkg/ccl/sqlccl/targets.go, line 49 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-Kyt01UNAPycqeNuN6-l:-Kyt01UNAPycqeNuN6-m:b-614yyz) ([raw file](https://github.com/cockroachdb/cockroach/blob/fc728d64302e49353f385b1380f3ea6075e8f2f1/pkg/ccl/sqlccl/targets.go#L49)):*
> ```Go
> // database descriptor is included in this set if it matches the targets (or the
> // session database) or if one of its tables matches the targets. Those
> // databases that are explicitly mentioned (e.g. as `DATABASE foo`, as opposed
> ```

This part of the comment is no longer accurate.

---

*[pkg/ccl/sqlccl/targets.go, line 144 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-Kyt0XEiOPJNUpelOtNO:-Kyt0XEiOPJNUpelOtNP:b-2alb83) ([raw file](https://github.com/cockroachdb/cockroach/blob/fc728d64302e49353f385b1380f3ea6075e8f2f1/pkg/ccl/sqlccl/targets.go#L144)):*
> ```Go
> 				tablesInDB[dbDesc.ID] = make(map[sqlbase.ID]bool)
> 			}
> 			tablesInDB[dbDesc.ID][tableDesc.ID] = false
> ```

I can't figure out what this does. If it's indeed supposed to be here can you add a comment describing why?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023)*

dt:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[pkg/ccl/sqlccl/restore.go, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-KyrQTyoC3TPHQOl3qnk:-KyvF9EQ285aUp1dNLVp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0d13f07f73719a6b1eeb956f858c6a6a86dde4fb/pkg/ccl/sqlccl/restore.go#L97)):*

Is `>= BackupFormatDescriptorTrackingVersion` clearer?
</blockquote></details>

Done.

---

*[pkg/ccl/sqlccl/targets.go, line 49 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-Kyt01UNAPycqeNuN6-l:-KyvGGmP5OIjtmcVB31L:bd1krq) ([raw file](https://github.com/cockroachdb/cockroach/blob/fc728d64302e49353f385b1380f3ea6075e8f2f1/pkg/ccl/sqlccl/targets.go#L49)):*

This part of the comment is no longer accurate.
</blockquote></details>

Reworded a bit -- also the switch the the struct return type to help document what all these lists are / mean.
Done.

---

*[pkg/ccl/sqlccl/targets.go, line 144 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-Kyt0XEiOPJNUpelOtNO:-KyvFA2jTkraCRkNYBWG:bt69q83) ([raw file](https://github.com/cockroachdb/cockroach/blob/fc728d64302e49353f385b1380f3ea6075e8f2f1/pkg/ccl/sqlccl/targets.go#L144)):*

I can't figure out what this does. If it's indeed supposed to be here can you add a comment describing why?
</blockquote></details>

thanks, was leftover from when I did the fancier thing we'd discussed and then dropped.
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023)*

mjibson:
:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20023#-:-KyvrbIzYyFgHLHe7KQE:bnfp4nl)*

PR #20025, opened 2017-11-13
distsql: implement SampleAggregator

RaduBerinde:
#### distsql: implement SampleAggregator

Initial implementation for SampleAggregator. Currently it doesn't put the
results anywhere (just dumps them to stdout).

```
=== RUN   TestSampleAggregator
colIDs: [100]  numRows: 11  numNulls: 2  cardinality: 2
histogram:
  1: less=0 eq=6
  2: less=0 eq=4
colIDs: [101]  numRows: 11  numNulls: 1  cardinality: 8
histogram:
  1: less=0 eq=2
  3: less=1 eq=2
  5: less=1 eq=1
  8: less=2 eq=1


rytaft:
:lgtm: but I think there are a few things you can do to make the code clearer (see inline comments).  

---

Reviewed 3 of 3 files at r2, 6 of 7 files at r3, 1 of 1 files at r4.

---

*[pkg/sql/distsqlrun/processors.proto, line 533 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrGs5Op4iRItyaapfH:-KyrGs5Op4iRItyaapfI:b-ibyd2d) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L533)):*
> ```protobuf
> 
> // SampleAggregatorSpec is the specification of a processor that aggregates the
> // results form multiple sampler processors and writes out the statistics to
> ```

^ results from

---

*[pkg/sql/distsqlrun/processors.proto, line 536 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrR6_CMsRGCDFu5M2u:-KyrR6_CMsRGCDFu5M2v:b3y2eo0) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L536)):*
> ```protobuf
> // system.table_statistics.
> //
> // The input schema it expects matches the output schema of a sampler spec:
> ```

Maybe add a more explicit reference to the comment above SamplerSpec - I spent a while trying to understand these values and ultimately realized that the detailed info was above.

---

*[pkg/sql/distsqlrun/processors.proto, line 553 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrUmyQMvBXsX26Wmj2:-KyrUmyQMvBXsX26Wmj3:bffexer) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L553)):*
> ```protobuf
>   // The i-th value indicates the column ID of the i-th sampled row
>   // column.
>   repeated uint32 sampled_column_ids = 3 [
> ```

Why does the SampleAggregatorSpec have a separate field for `sampled_column_ids`, while the SamplerSpec has only a `columns` field in the SketchSpec?

---

*[pkg/sql/distsqlrun/processors.proto, line 563 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrVLEyFPfm3iMd3yCM:-KyrVLEyFPfm3iMd3yCN:b-owmlhg) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L563)):*
> ```protobuf
>     // Columns for which to compute the sketch.
>     // TODO(radu): currently only one column is supported.
>     repeated uint32 column_ids = 2 [
> ```

Why does the SamplerSpec call this field `columns` instead of `column_ids`?

---

*[pkg/sql/distsqlrun/sample_aggregator.go, line 196 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyraSl6esqlMis_vuxV:-KyraSl6esqlMis_vuxW:b-w9jdc9) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator.go#L196)):*
> ```Go
> 		colIdx, ok := colIDMap[spec.ColumnIDs[0]]
> 		if !ok {
> 			return false, errors.Errorf("columnID %d not sampled", spec.ColumnIDs[0])
> ```

Do we need to throw an error if the histogram is not requested?  Seems like this check should go inside the `if` block below.

---

*[pkg/sql/distsqlrun/sample_aggregator.go, line 98 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrpjBxa2ba6jY0BN_M:-KyrpjBxa2ba6jY0BN_N:b-48rupu) ([raw file](https://github.com/cockroachdb/cockroach/blob/e8b9a4f5bc505fb6c611456d2847be0dc43b5865/pkg/sql/distsqlrun/sample_aggregator.go#L98)):*
> ```Go
> }
> 
> func (s *sampleAggregator) mainLoop(ctx context.Context) (earlyExit bool, _ error) {
> ```

Is there any way to split this function up a bit?  Seems unnecessarily large...

---

*[pkg/sql/distsqlrun/sample_aggregator_test.go, line 53 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrbSHgu40ilVZO_dO4:-KyrbSHgu40ilVZO_dO5:bw8rk0o) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator_test.go#L53)):*
> ```Go
> 	}
> 	//cardinalities := []int{2, 8}
> 	//numNulls := []int{2, 1}
> ```

why are these commented out?  Maybe remove them?

---

*[pkg/sql/distsqlrun/sample_aggregator_test.go, line 85 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrjQzheLdGoz8B2ckv:-KyrjQzheLdGoz8B2ckw:b-j47b39) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator_test.go#L85)):*
> ```Go
> 
> 		spec := &SamplerSpec{
> 			SampleSize: 100,
> ```

Is it a problem that you have more samples than inputRows?

---

*[pkg/sql/distsqlrun/sample_aggregator_test.go, line 124 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrfHxibjSinG-UegqW:-KyrfHxibjSinG-UegqX:b-50zwx3) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator_test.go#L124)):*
> ```Go
> 		SampleSize:       100,
> 		SampledColumnIDs: []sqlbase.ColumnID{100, 101},
> 		Sketches: []SampleAggregatorSpec_SketchSpec{
> ```

Maybe also test whether it works to not generate the histogram?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-:-Kyrq3hcoA-EFGNQ2Ck-:btl16yl)*

RaduBerinde:
TFTR, great comments!

---

Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlrun/processors.proto, line 533 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrGs5Op4iRItyaapfH:-KyuV2JbA-VUWlMeqo4j:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L533)):*

^ results from
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/processors.proto, line 536 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrR6_CMsRGCDFu5M2u:-KyuVAWqIOFnYDQ64n0O:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L536)):*

Maybe add a more explicit reference to the comment above SamplerSpec - I spent a while trying to understand these values and ultimately realized that the detailed info was above.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/processors.proto, line 553 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrUmyQMvBXsX26Wmj2:-KyuVCo7LvD4WKhw8jWK:b3tggo0) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/processors.proto#L553)):*

Why does the SampleAggregatorSpec have a separate field for `sampled_column_ids`, while the SamplerSpec has only a `columns` field in the SketchSpec?
</blockquote></details>

The SampleAggregator will write out the results in the table_statistics table. It needs to use ColumnIDs (which are "stable"). But you are right that the SketchSpec can just have column indexes. I consolidated the two SketchSpecs into one.

---

*[pkg/sql/distsqlrun/sample_aggregator.go, line 196 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyraSl6esqlMis_vuxV:-KyuXDwtd7xE7efIWT6-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator.go#L196)):*

Do we need to throw an error if the histogram is not requested?  Seems like this check should go inside the `if` block below.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/sample_aggregator.go, line 98 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrpjBxa2ba6jY0BN_M:-KyubwDzp7fmn33B2fGx:b-t0ceyy) ([raw file](https://github.com/cockroachdb/cockroach/blob/e8b9a4f5bc505fb6c611456d2847be0dc43b5865/pkg/sql/distsqlrun/sample_aggregator.go#L98)):*

Is there any way to split this function up a bit?  Seems unnecessarily large...
</blockquote></details>

I consolidated the per-sketch data in sketchInfo (for both sampler and sampleAggregator) which moved away a lot of the initialization code, and I moved the "write results" phase to a different function.

---

*[pkg/sql/distsqlrun/sample_aggregator_test.go, line 53 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrbSHgu40ilVZO_dO4:-KyuXF60v_fpr7ih6T3X:b2juuew) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator_test.go#L53)):*

why are these commented out?  Maybe remove them?
</blockquote></details>

This test is a WIP, it's not really checking anything yet, but it will need to check against those expected values.

---

*[pkg/sql/distsqlrun/sample_aggregator_test.go, line 85 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrjQzheLdGoz8B2ckv:-KyuXN9fA0lBTZ_re8ze:b-iwxqzw) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator_test.go#L85)):*

Is it a problem that you have more samples than inputRows?
</blockquote></details>

That's the only way to make the test deterministic. Note that we are not interested in testing the Sampler, the test is for SampleAggregator.

---

*[pkg/sql/distsqlrun/sample_aggregator_test.go, line 124 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-KyrfHxibjSinG-UegqW:-Kyubob1ELVAEDmzkqM3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/18f3447601c38f6c919d0322360bc1dd0de8a28e/pkg/sql/distsqlrun/sample_aggregator_test.go#L124)):*

Maybe also test whether it works to not generate the histogram?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-:-KyuUGlpRAgq-7-WqNIc:b-1c6pnz)*

rytaft:
Good changes!

---

Reviewed 2 of 3 files at r6, 8 of 8 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20025#-:-Kyv9Fur35OufaPGVw76:b-45muys)*

PR #20027, opened 2017-11-13
storageccl: support 201 and 204 return codes

### LINKED ISSUE #20017: sqlccl: http export storage is too restrictive in what codes mean success
There is an explicit check for `if resp.StatusCode != 200 {`, but 200 is not the only status code that means success. This request is from one of our customers.

madelynnblue:
Also correctly close response bodies when unused.

Fixes #20017

### DIFF
--- a/pkg/ccl/storageccl/export_storage.go
+++ b/pkg/ccl/storageccl/export_storage.go
@@ -328,17 +328,17 @@ func (h *httpStorage) ReadFile(_ context.Context, basename string) (io.ReadClose
 }
 
 func (h *httpStorage) WriteFile(_ context.Context, basename string, content io.ReadSeeker) error {
-	_, err := h.req("PUT", basename, content)
+	_, err := h.reqNoBody("PUT", basename, content)
 	return err
 }
 
 func (h *httpStorage) Delete(_ context.Context, basename string) error {
-	_, err := h.req("DELETE", basename, nil)
+	_, err := h.reqNoBody("DELETE", basename, nil)
 	return err
 }
 
 func (h *httpStorage) Size(_ context.Context, basename string) (int64, error) {
-	resp, err := h.req("HEAD", basename, nil)
+	resp, err := h.reqNoBody("HEAD", basename, nil)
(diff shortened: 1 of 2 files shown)


dt:
:lgtm:

---

---

*[pkg/ccl/storageccl/export_storage.go, line 387 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20027#-Kz-DkhzZAI-Zz2pVJ4G:-Kz-DkhzZAI-Zz2pVJ4H:b8rgfdo) ([raw file](https://github.com/cockroachdb/cockroach/blob/72b5b1ff59317d604452b63eda97d9112c105c93/pkg/ccl/storageccl/export_storage.go#L387)):*
> ```Go
> 	}
> 	switch resp.StatusCode {
> 	case 200, 201, 204:
> ```

`resp.StatusCode >= 200 && resp.StatusCode < 300`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20027#-:-Kz-DcyLtqmOvXaZIjfZ:bnfp4nl)*

mjibson:

---

*[pkg/ccl/storageccl/export_storage.go, line 387 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20027#-Kz-DkhzZAI-Zz2pVJ4G:-Kz-xia3BCJR1bCvIAbq:b-yxk27o) ([raw file](https://github.com/cockroachdb/cockroach/blob/72b5b1ff59317d604452b63eda97d9112c105c93/pkg/ccl/storageccl/export_storage.go#L387)):*

`resp.StatusCode >= 200 && resp.StatusCode < 300`?
</blockquote></details>

Take status 202 Accepted, which is: "The request has been accepted for processing, but the processing has not been completed." In that case we definitely don't want to proceed because the file isn't actually there yet. The others are similarly iffy so I'd rather stick with whack-a-mole for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20027)*

PR #20028, opened 2017-11-13
client: avoid filling results for "raw" Batch requests

nvb:
While I don't think we'll ever actually want to merge #19050, it revealed that
there were cases where `fillResults` was being called twice for `client.Batches`.
The times when this could happen were where a batch was being used initially as
a "raw" batch (such as in `TxnCoordSender.resendWithTxn`). This in turn revealed
that we were actually always performing a lot of unnecessary bookkeeping on "raw"
requests in both `Batch.initResult` and `Batch.fillResults`, including lots of
`[]KeyValue` allocations. This change avoids all of this extra work.

### DIFF
--- a/pkg/internal/client/batch.go
+++ b/pkg/internal/client/batch.go
@@ -97,7 +97,7 @@ func (b *Batch) initResult(calls, numRows int, raw bool, err error) {
 	}
 	// TODO(tschottdorf): assert that calls is 0 or 1?
 	r := Result{calls: calls, Err: err}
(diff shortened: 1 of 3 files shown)


bdarnell:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20028#-:-Kyr_r0O_MI_iHoX4Zlx:bnfp4nl)*

Issue #19723, opened 2017-11-01
sql: sequences (tracking issue)


vilterp:
DEPRECATED TRACKING ISSUE IN FAVOR OF PROJECT: https://github.com/cockroachdb/cockroach/projects/13

Original issue: #5811 
RFC: #19196 
Docs issue: https://github.com/cockroachdb/docs/issues/2014

The implementation will be split into multiple PRs; this issue tracks them.

- [x] Implement `CREATE SEQUENCE` and `nextval` (biggest; others depend on this) (#19664)
- [ ] implement `setval(seqname, val, iscalled)`
- [x] Don't evaluate `DEFAULT` expressions on table create (#20031)
- [ ] Respect the `MINVALUE`, `MAXVALUE`, and `CYCLE` settings (#20574)
- [x] `pg_class` fixes (#20282)
- [x] `information_schema.sequences` entries and `pg_sequence` (#20111)
- [x] Implement `DROP SEQUENCE` and `ALTER SEQUENCE` (#20089)
- [x] Implement additional builtin functions (`currval`, `setval`, `lastval`) (#20306)
- [ ] enforce permissions in builtins (e.g. can't call `nextval` without write permissions)
- [ ] implement `ALTER SEQUENCE ... RESTART WITH x`
- [ ] Add dependency relationship between sequences and the columns that use them (protecting a sequence from being dropped if it's in use); implement `OWNED BY`
  - [ ] `DROP TABLE CASCADE`, `DROP SEQUENCE CASCADE` (what does PG do here?)
- [ ] Make sequences work in load/dump and backup/restore (#20725)
- [ ] Ensure that sequences work with zone configs (so an operator can control locality of a sequence)
- [ ] teach UI about sequence event types
- [ ] use regclass instead of string as argument to sequence functions
- [ ] update the RFC to say "complete"
- [ ] add Hibernate test to https://github.com/cockroachdb/examples-orms

lego:
Found usage of `setval(string, int, bool)` that isn't postgres backups :'(.

It is called in ActiveRecord's [PostgreSQL adapter](https://github.com/rails/rails/blob/9f33a8f06e5bba343b0d108a41246ab8368d71d3/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L275-L301).

It does seem to be special case thing that I only encountered it because of running the ActiveRecord test suite against CockroachDB. So no need to take this as a reason to implement it.

[...]

PR #15345, opened 2017-04-25
sql: check DEFAULT expressions in schema changes earlier.

knz:
Prior to this patch, a new DEFAULT expression would only be checked
after the schema change has started, which is typically at the end of
the schema change transaction. An error, if any, would not be reported
immediately upon issuing the ALTER (or CREATE) statement.

This patch improves upon this situation by adding another check
upfront, when the mutation descriptor is initially created.
Note that this check is not sufficient to guarantee the DEFAULT
expression is valid however (i.e. it may have false negatives),
because impure functions can cause the expression to become invalid
non-deterministically (e.g. `1 / (10*(random() - 1))::int`).

PR #15344, opened 2017-04-25
sql: ensure that DEFAULT exprs are re-evaluated during backfill.

knz:
Fixes #14352.

cc @andreimatei

PR #20031, opened 2017-11-13
sql: don't evaluate DEFAULT expressions on table creation

vilterp:
Not sure what this would catch, and it throws off sequences (see #19664) -- if you create a table with a column that uses a sequence it its DEFAULT expression, it'll increment the sequence at table creation time, before a row has been inserted into the table.

tracking issue: #19723

### DIFF
--- a/pkg/sql/sqlbase/table.go
+++ b/pkg/sql/sqlbase/table.go
@@ -213,13 +213,6 @@ func MakeColumnDefDescs(
 		if typedExpr, err = t.NormalizeExpr(evalCtx, typedExpr); err != nil {
 			return nil, nil, err
 		}
-		// Try to evaluate once. If it is aimed to succeed during a
-		// backfill, it must succeed here too. This tries to ensure that
-		// we don't end up failing the evaluation during the schema change
-		// proper.
-		if _, err := typedExpr.Eval(evalCtx); err != nil {
-			return nil, nil, err
-		}
 		d.DefaultExpr.Expr = typedExpr
 
 		s := tree.Serialize(d.DefaultExpr.Expr)
--- a/pkg/sql/logictest/testdata/logic_test/alter_table
+++ b/pkg/sql/logictest/testdata/logic_test/alter_table
@@ -663,3 +663,22 @@ query I
 SELECT COUNT(*) FROM crdb_internal.jobs WHERE status = 'pending' OR status = 'started'
 ----
 0
+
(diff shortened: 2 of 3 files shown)


pkg/sql/sqlbase/table.go
```diff
@@ -196,13 +196,6 @@ func MakeColumnDefDescs(
 		if typedExpr, err = t.NormalizeExpr(evalCtx, typedExpr); err != nil {
 			return nil, nil, err
 		}
-		// Try to evaluate once. If it is aimed to succeed during a
-		// backfill, it must succeed here too. This tries to ensure that
-		// we don't end up failing the evaluation during the schema change
-		// proper.
-		if _, err := typedExpr.Eval(evalCtx); err != nil {
-			return nil, nil, err
-		}
```
vivekmenezes: this was introduced in https://github.com/cockroachdb/cockroach/pull/15345 . Do you think you can disable it for some types of expressions?

knz: I think that "check by evaluation" is conceptually flawed. 
1) there are expressions that just cannot be checked this way -- they will only cause an error in some evaluations, so even if we have this check it doesn't guarantee that the backfill will succeed
2) as Pete says it prevents us from having expressions with side effects. `nextval` is our first instance of this, but if/when we support user-defined functions/procedures with side effects, "evaluation for checking" will be simply unacceptable.

I think we should simply accept the expression as-is if it passes type checking, and instead be robust with erroring out on backfills and allowing the user to replace it with ALTER.

knz:
@vilterp the code change looks good, however please check whether the backfill properly fails if an expression passes type checking but fails during evaluation. (e.g. using `force_error`)

knz:
The "properly fails" in my previous comment means:
- the backfill stops, and doesn't retry in a loop
- an error is properly reported in the admin UI
- the user can change the default expression with ALTER and restart the backfill.

For any of these points, if there's a deviation please file an issue.

vivekmenezes:
LGTM

The column backfill code should return an error which will stop the backfill and rollback the schema change

https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/distsqlrun/columnbackfiller.go#L221

vilterp:
Does a test for the backfill-aborting behavior @vivekmenezes exist? (Can't seem to find it). If not, where should such a test be added?

vivekmenezes:
https://github.com/cockroachdb/cockroach/pull/15344 is the relevant PR

PR #20034, opened 2017-11-14
sql: never use Datum.Next when computing spans

benesch:
tl;dr avoid computing overwide spans in SQL that cause multiple partitions to be scanned when the SQL query is constrained to one region.

For example, this query should only hit one range

```sql
CREATE TABLE db.users (country STRING PRIMARY KEY, name STRING) PARTITION BY LIST (users) (
    PARTITION us VALUES IN ('us'),
    PARTITION ca VALUES IN ('ca')
);
ALTER TABLE db.users PARTITION us EXPERIMENTAL CONFIGURE ZONE '{gc: {ttlseconds: 60}}';
INSERT INTO db.users VALUES ('us', 'nikhil'), ('ca', 'justin')
SELECT * FROM db.users WHERE country = 'us';
```

but without this patch it hits two.

Full details in the commit messages:

```
keys: check if undoing PrefixEnd makes keys pretty printable
Calling PrefixEnd can result in semantically invalid keys. Suppose we
have the following key:

    b2          12             61    00 01
    /Table/42   bytes marker   "a"   term marker

Calling PrefixEnd will blindly increment the last byte, resulting in:

    b2 12 61 00 02

This key has no semantic meaning, as the byte sequence termination
marker has been destroyed. The pretty printer currently has no choice
but to display this key with a bunch of ???s:

    /Table/42/???

We can do better than this by checking whether undoing a PrefixEnd
operation results in a printable key. Now, the pretty printer can
display the above key as:

    /Table/42/"a"/PrefixEnd

Note that the pretty printer will never append "/PrefixEnd" to a
semantically valid key. For example, "/Table/42/1".PrefixEnd() will be
printed as "/Table/42/2" because, in this case, PrefixEnd results in a
semantically valid key.
```

```
sql: never use Datum.Next when computing spans
The SQL layer must translate inclusive upper bounds into exclusive upper
bounds, as KV scans always use exclusive upper bounds. The PrefixEnd key
operation is provided for exactly this purpose, but it can result in
semantically-invalid keys that the pretty printer could not handle
before 8fea154.

As a workaround, span computation would call Datum.Next instead of
Key.PrefixEnd whenever the datum was not already the maximum value. This
kept spans "nice and pretty-printable" at the cost of making them too
wide.

For example, the key /Table/42/"a" is encoded as:

        b2          12             61    00 01
        /Table/42   bytes marker   "a"   term marker

Calling PrefixEnd increments only the last byte

        b2          12             61    00 02
        /Table/42   bytes marker   "a"   term marker

whereas calling Datum.Next appends a null byte to the bytes sequence
within the key, which results in a much greater key:

        b2          12             61 ff     00 02
        /Table/42   bytes marker   "a\x00"   term marker

Up until now, using Datum.Next instead of PrefixEnd had largely no
effect, as the SQL layer prevents any "in-between" keys, like b2 12 61
01, from ever appearing. With partitioning, however, it's possible to
introduce a split at the PrefixEnd of any key (e.g., at b2 12 61 00 02),
which means that the span computed by `SELECT * FROM t WHERE pk = 'a'`,
which extends to b2 12 61 ff 00 02, would be too wide and bleed into the
next range.

Now that PrefixEnd keys can always be pretty-printed, just remove the
special casing from SQL span generation and always use PrefixEnd. This
guarantees that the tightest possible span is computed.

(We could have instead taught partitioning to call Datum.Next instead of
PrefixEnd in exactly the cases where SQL span computation did, but that
seemed like furthering a hack that existed only to work around a
limitation in the pretty printer.)
```

This is kind of a scary change; please suggest additional reviewers!

### DIFF
--- a/pkg/keys/printer_test.go
+++ b/pkg/keys/printer_test.go
@@ -150,7 +150,10 @@ func TestPrettyPrint(t *testing.T) {
 		{makeKey([]byte("")), "/Min"},
 		{Meta1KeyMax, "/Meta1/Max"},
 		{Meta2KeyMax, "/Meta2/Max"},
-		{makeKey(MakeTablePrefix(42), roachpb.RKey([]byte{0x12, 'a', 0x00, 0x02})), "/Table/42/???"},
+		{makeKey(MakeTablePrefix(42), roachpb.RKey([]byte{0xf6})), `/Table/42/109/PrefixEnd`},
+		{makeKey(MakeTablePrefix(42), roachpb.RKey([]byte{0xf7})), `/Table/42/???`},
+		{makeKey(MakeTablePrefix(42), roachpb.RKey([]byte{0x12, 'a', 0x00, 0x02})), `/Table/42/"a"/PrefixEnd`},
+		{makeKey(MakeTablePrefix(42), roachpb.RKey([]byte{0x12, 'a', 0x00, 0x03})), `/Table/42/???`},
 	}
 	for i, test := range testCases {
 		keyInfo := MassagePrettyPrintedSpanForTest(PrettyPrint(test.key), nil)
--- a/pkg/sql/index_selection.go
+++ b/pkg/sql/index_selection.go
@@ -1138,20 +1138,6 @@ func spanFromLogicalSpan(
 			break
 		}
 	}
-	// Since the end of a span is exclusive, if the last constraint is an
-	// inclusive one, we might need to make the key exclusive by applying a
-	// PrefixEnd(). We normally avoid doing this by transforming "a = x" to "a =
-	// x±1" for the last end constraint, depending on the encoding direction
-	// (since this keeps the key nice and pretty-printable). However, we might not
-	// be able to do the ±1.
-	if n := len(ls.end); n > 0 && ls.end[n-1].inclusive {
-		last := &ls.end[n-1]
-		nextVal, hasNext := nextInDirection(evalCtx, last.val, last.dir)
-		if hasNext {
-			last.val = nextVal
-			last.inclusive = false
-		}
-	}
 	for i := 0; ; i++ {
 		s.EndKey = append(s.EndKey, interstices[i]...)
 		if i >= len(ls.end) {
@@ -1188,21 +1174,6 @@ func encodeLogicalKeyPart(
 		return nil, err
 	}
 	return sqlbase.EncodeTableKey(b, d, part.dir)
-}
-
-func nextInDirection(
-	evalCtx *tree.EvalContext, val tree.Datum, dir encoding.Direction,
-) (tree.Datum, bool) {
-	if dir == encoding.Ascending {
-		if val.IsMax(evalCtx) {
-			return nil, false
-		}
-		return val.Next(evalCtx)
-	}
-	if val.IsMin(evalCtx) {
-		return nil, false
-	}
-	return val.Prev(evalCtx)
 }
 
 // mergeAndSortSpans is used to merge a set of potentially overlapping spans
--- a/pkg/sql/index_selection_test.go
+++ b/pkg/sql/index_selection_test.go
@@ -563,7 +563,7 @@ func TestMakeSpans(t *testing.T) {
 			`/7/3/0-/7/3/1 /7/2/0-/7/2/1 /7/1/0-/7/1/1`},
 		// Test different directions for te columns inside a tuple.
 		{`(a,b,j) IN ((1,2,3), (4,5,6))`, `a-,b,j-`, `/4/5/6-/4/5/5 /1/2/3-/1/2/2`},
-		{`k = b'\xff'`, `k`, `/"\xff"-/"\xff\x00"`},
+		{`k = b'\xff'`, `k`, `/"\xff"-/"\xff"/PrefixEnd`},
 		// Test that limits on bytes work correctly: when encoding a descending limit for bytes,
 		// we need to go outside the bytes encoding.
 		// "\xaa" is encoded as [bytesDescMarker, ^0xaa, <term escape sequence>]
--- a/pkg/sql/index_selection_test.go
+++ b/pkg/sql/index_selection_test.go
@@ -506,15 +506,12 @@ func TestMakeSpans(t *testing.T) {
 		{`(a = 5) OR (a, b) IN ((1, 1), (3, 3))`, `a,b`,
 			`/1/1-/1/2 /3/3-/3/4 /5-/6`, `/5-/4 /3/3-/3/2 /1/1-/1/0`},
 
-		// When encoding an end constraint for a maximal datum, we use
-		// bytes.PrefixEnd() to go beyond the normal encodings of that datatype.
-		// This is the reason for the "???" suffix of the pretty-printed spans.
 		{fmt.Sprintf(`a = %d`, math.MaxInt64), `a`,
-			`/9223372036854775807-/???`,
+			`/9223372036854775807-/9223372036854775807/PrefixEnd`,
 			`/9223372036854775807-/9223372036854775806`},
 		{fmt.Sprintf(`a = %d`, math.MinInt64), `a`,
 			`/-9223372036854775808-/-9223372036854775807`,
-			`/-9223372036854775808-/???`},
+			`/-9223372036854775808-/-9223372036854775808/PrefixEnd`},
 
 		{`(a, b) >= (1, 4)`, `a,b`, `/1/4-`, `-/1/3`},
 		{`(a, b) > (1, 4)`, `a,b`, `/1/5-`, `-/1/4`},
--- a/pkg/sql/logictest/testdata/logic_test/deep_interleaving
+++ b/pkg/sql/logictest/testdata/logic_test/deep_interleaving
@@ -102,12 +102,16 @@ SELECT * FROM level4
 3  30  200
 3  30  300
 
+# The span below ends at the end of the first index of table 53, and is not
+# constraining the value of k2 or k3. This is confusing on first glance because
+# the second interleave in the hierarchy doesn't contain any new primary key
+# columns on top of the first interleave.
 query ITTT
 EXPLAIN SELECT * FROM level4 WHERE k1 > 1 AND k1 < 3
 ----
 0  scan  ·      ·
 0  ·     table  level4@primary
-0  ·     spans  /2/#/52/1/#/53/1-/3
+0  ·     spans  /2/#/52/1/#/53/1-/2/#/52/1/#/53/2
 
 query III rowsort
 SELECT * FROM level4 WHERE k1 > 1 AND k1 < 3
@@ -141,7 +145,7 @@ EXPLAIN SELECT * FROM level4 WHERE k1 = 2 AND k2 = 20 AND k3 > 100 AND k3 < 300
 ----
 0  scan  ·      ·
 0  ·     table  level4@primary
-0  ·     spans  /2/#/52/1/#/53/1/20/101/#/54/1-/2/#/52/1/#/53/1/20/300
+0  ·     spans  /2/#/52/1/#/53/1/20/101/#/54/1-/2/#/52/1/#/53/1/20/299/#/54/2
 
 query III
 SELECT * FROM level4 WHERE k1 = 2 AND k2 = 20 AND k3 > 100 AND k3 < 300
@@ -153,7 +157,7 @@ EXPLAIN SELECT * FROM level4 WHERE k1 = 2 AND k2 = 20 AND k3 = 200
 ----
 0  scan  ·      ·
 0  ·     table  level4@primary
-0  ·     spans  /2/#/52/1/#/53/1/20/200/#/54/1-/2/#/52/1/#/53/1/20/201
+0  ·     spans  /2/#/52/1/#/53/1/20/200/#/54/1-/2/#/52/1/#/53/1/20/200/#/54/2
 
 query III
 SELECT * FROM level4 WHERE k1 = 2 AND k2 = 20 AND k3 = 200
--- a/pkg/sql/logictest/testdata/logic_test/float
+++ b/pkg/sql/logictest/testdata/logic_test/float
@@ -24,7 +24,7 @@ EXPLAIN SELECT * FROM p WHERE f IS NaN
 0  render  ·      ·
 1  scan    ·      ·
 1  ·       table  p@p_f_key
-1  ·       spans  /NaN-/-Inf
+1  ·       spans  /NaN-/NaN/PrefixEnd
 
 query ITTT
 EXPLAIN SELECT * FROM p WHERE f = 'NaN'
@@ -32,7 +32,7 @@ EXPLAIN SELECT * FROM p WHERE f = 'NaN'
 0  render  ·      ·
 1  scan    ·      ·
 1  ·       table  p@p_f_key
-1  ·       spans  /NaN-/-Inf
+1  ·       spans  /NaN-/NaN/PrefixEnd
 
 query ITTT
 EXPLAIN SELECT * FROM p WHERE isnan(f)
--- a/pkg/sql/logictest/testdata/logic_test/select_index
+++ b/pkg/sql/logictest/testdata/logic_test/select_index
@@ -619,7 +619,7 @@ EXPLAIN SELECT * FROM test2 WHERE k <= '100' ORDER BY k DESC LIMIT 20
 1  index-join  ·      ·
 2  revscan     ·      ·
 2  ·           table  test2@test2_k_key
-2  ·           spans  /#-/"100\x00"
+2  ·           spans  /#-/"100"/PrefixEnd
 2  ·           limit  20
 2  scan        ·      ·
 2  ·           table  test2@primary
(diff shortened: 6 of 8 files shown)


knz:
The main gist of the change here is excellent insight. Congrats on having it and following through.

Radu will be most insightful to analyze the ramifications of this change I think. Jordan I think this also touches on your work with interleaved spans.

knz:
Code&tests LGTM. I'd welcome other reviewers to check I'm not missing anything conceptually.

---

Reviewed 3 of 3 files at r1, 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KyuB4kJ_Gd0QU_aM1hx:boi1yy5)*

RaduBerinde:
Review status: 1 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 110 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyusD3Fa11L6DrB5DTH:-KyusD3G8PfIAEUmNHeL:b5ktoil) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L110)):*
> ```
> 0  scan  ·      ·
> 0  ·     table  level4@primary
> 0  ·     spans  /2/#/52/1/#/53/1-/2/#/52/1/#/53/2
> ```

Is this right? It seems like we are constraining k2=k3=1?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034)*

knz:
Okay thanks to Radu's comment I think indeed I had missed something in my first review. Nikhil please double check the behavior when you have a tuple to encode. 

---

Reviewed 1 of 6 files at r3, 2 of 5 files at r4.
Review status: 4 of 7 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KyutiiuXO4V8GjsG1VQ:bn9o60w)*

jordanlewis:
<3 this change

---

Reviewed 1 of 3 files at r1, 6 of 6 files at r3, 5 of 5 files at r4.

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 110 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyusD3Fa11L6DrB5DTH:-Kyv8tOtNyzEEAHGbZBv:b-z1mlx3) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L110)):*

Is this right? It seems like we are constraining k2=k3=1?
</blockquote></details>

I think this is correct. This is saying that the span ends at the end of the 1st index of table 53 - it isn't actually saying anything about the values of k2 or k3. The example is confusing because the second interleave doesn't contain any unique values on top of the first interleave. All `1` values in this span are index ids.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KyvCoXe8P_0YGsOo5A5:b-ot0w36)*

RaduBerinde:
*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 110 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyusD3Fa11L6DrB5DTH:-KyvDIj9fBfS0JLnBgSe:bic91w4) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L110)):*

I think this is correct. This is saying that the span ends at the end of the 1st index of table 53 - it isn't actually saying anything about the values of k2 or k3. The example is confusing because the second interleave doesn't contain any unique values on top of the first interleave. All `1` values in this span are index ids.
</blockquote></details>

Ah, interesting. @benesch maybe you can add a comment explaining this. Very strange interleaving where child and parent have the same key, I didn't even know that was possible

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034)*

danhhz:
:lgtm_strong: insane turnaround on this! i'm very happy in the reduction in complexity here and the tighter scan bounds

---

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 144 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvFcLxaAY0KiDM5s1q:-KyvFcLxaAY0KiDM5s1r:bfpglkk) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L144)):*
> ```
> 0  scan  ·      ·
> 0  ·     table  level4@primary
> 0  ·     spans  /2/#/52/1/#/53/1/20/101/#/54/1-/2/#/52/1/#/53/1/20/299/#/54/2
> ```

i'm really impressed by how tight the end bound is now. gg index selection code!

---

*[pkg/util/encoding/encoding.go, line 965 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvDLYAfUO3YrIlb2g3:-KyvDLYAfUO3YrIlb2g4:b-qukcdy) ([raw file](https://github.com/cockroachdb/cockroach/blob/9d3e8b554ff53ccd0904861b92d3d78831c2b347/pkg/util/encoding/encoding.go#L965)):*
> ```Go
> func PrettyPrintValue(b []byte, sep string) string {
> 	s1, allDecoded := prettyPrintValueImpl(b, sep)
> 	if !allDecoded {
> ```

nit: this is slightly easier to read (for me) if you use allDecoded for the early return:
```
if allDecoded {
  return s1
}
if s2, allDecoded := prettyPrintValueImpl(undoPrefixEnd(b), sep); allDecoded {
  return s2 + sep + "PrefixEnd"
}
return s1
```

---

*[pkg/util/encoding/encoding.go, line 1057 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvDjZ-u-qPn_8AC26N:-KyvDjZ-u-qPn_8AC26O:bjw3vka) ([raw file](https://github.com/cockroachdb/cockroach/blob/9d3e8b554ff53ccd0904861b92d3d78831c2b347/pkg/util/encoding/encoding.go#L1057)):*
> ```Go
> 
> func undoPrefixEnd(b []byte) []byte {
> 	out := append([]byte{}, b...)
> ```

I think our style here is `{}byte(nil)`. Can't remember if that has actual runtime impact or not, but that's definitely what I see everywhere

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KyvFqnn_EZslarGWtu_:b2hq4zv)*

danhhz:

---

*[pkg/util/encoding/encoding.go, line 1057 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvDjZ-u-qPn_8AC26N:-KyvHaWfRJxN9VP0d6Om:b-6qkvlb) ([raw file](https://github.com/cockroachdb/cockroach/blob/9d3e8b554ff53ccd0904861b92d3d78831c2b347/pkg/util/encoding/encoding.go#L1057)):*

I think our style here is `{}byte(nil)`. Can't remember if that has actual runtime impact or not, but that's definitely what I see everywhere
</blockquote></details>

[]byte(nil)

---

*[pkg/util/encoding/encoding.go, line 1056 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvHbVHqfsQHSHv9Q2B:-KyvHbVHqfsQHSHv9Q2C:b15ib17) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/util/encoding/encoding.go#L1056)):*
> ```Go
> }
> 
> func undoPrefixEnd(b []byte) []byte {
> ```

does this function roundtrip `[]byte(0xff, 0xff)` and/or `[]byte(0, 0)` correctly with PrefixEnd?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034)*

jordanlewis:
:lgtm: - I think this will also make fixing the parent interleave selection precision bug easier. Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KyvJmYk6MiLvbua-Jaa:b-db7qu8)*

benesch:
Review status: 1 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 110 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyusD3Fa11L6DrB5DTH:-Kyvm9NsH68bfKVUZxuj:b-keansr) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L110)):*

Ah, interesting. @benesch maybe you can add a comment explaining this. Very strange interleaving where child and parent have the same key, I didn't even know that was possible
</blockquote></details>

Sure thing! Done.

---

*[pkg/util/encoding/encoding.go, line 965 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvDLYAfUO3YrIlb2g3:-KyvlLAZXaXVXBszvYui:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9d3e8b554ff53ccd0904861b92d3d78831c2b347/pkg/util/encoding/encoding.go#L965)):*

nit: this is slightly easier to read (for me) if you use allDecoded for the early return:
```
if allDecoded {
  return s1
}
if s2, allDecoded := prettyPrintValueImpl(undoPrefixEnd(b), sep); allDecoded {
  return s2 + sep + "PrefixEnd"
}
return s1
```
</blockquote></details>

Done.

---

*[pkg/util/encoding/encoding.go, line 1057 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvDjZ-u-qPn_8AC26N:-KyvlLf1s1ob09rGDbCI:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9d3e8b554ff53ccd0904861b92d3d78831c2b347/pkg/util/encoding/encoding.go#L1057)):*

[]byte(nil)
</blockquote></details>

Done.

---

*[pkg/util/encoding/encoding.go, line 1056 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyvHbVHqfsQHSHv9Q2B:-Kyvkkw-4BpH4aLBh1w_:b-ibnj8u) ([raw file](https://github.com/cockroachdb/cockroach/blob/bf53afee8a015811d70f8533b62d6aa1e5b7a651/pkg/util/encoding/encoding.go#L1056)):*
> does this function roundtrip []byte(0xff, 0xff) and/or []byte(0, 0) correctly with PrefixEnd?

It's "correct" for some definition of correct. Added a big ol' comment and a test to explain the situation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034)*

danhhz:
i hate to add yet another teamcity run, so feel free to ignore these last two comments if you like

---

Review status: 1 of 8 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[pkg/util/encoding/encoding.go, line 1067 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KywEaqzxitJqSEF1F_G:-KywEar-pIHIDgO1wWfx:bbr15wi) ([raw file](https://github.com/cockroachdb/cockroach/blob/db6311739d272839a9cdfa8a068ff2defc15e2f8/pkg/util/encoding/encoding.go#L1067)):*
> ```Go
> // Formally:
> //
> //     PrefixEnd(UndoPrefixEnd(p)) = p for all prefixes p
> ```

"for all prefixes" is not true, right? this has the same loss of data issue as `UndoPrefixEnd(PrefixEnd(p))`

i think we can just remove this line and leave only the one below this (which is really the assertion we care about)

---

*[pkg/util/encoding/printer_test.go, line 49 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KywEsmnKNFUmR8P8zpH:-KywEsmnKNFUmR8P8zpI:b-lqxhjx) ([raw file](https://github.com/cockroachdb/cockroach/blob/db6311739d272839a9cdfa8a068ff2defc15e2f8/pkg/util/encoding/printer_test.go#L49)):*
> ```Go
> 		{0x00, 0x01},
> 		{0x00, 0xff, 0xff, 0xff},
> 		{0x01, 0x00, 0x00, 0x00},
> ```

comment that a `{0xff, 0xff}` test case wouldn't roundtrip, plz

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KywF4eTuI1fjnHCnukV:b-bjihgx)*

benesch:
Thanks for the detailed reviews, all!

---

Review status: 1 of 8 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/util/encoding/encoding.go, line 1067 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KywEaqzxitJqSEF1F_G:-KywTVRqArdGnhExdISS:brr099x) ([raw file](https://github.com/cockroachdb/cockroach/blob/db6311739d272839a9cdfa8a068ff2defc15e2f8/pkg/util/encoding/encoding.go#L1067)):*

"for all prefixes" is not true, right? this has the same loss of data issue as `UndoPrefixEnd(PrefixEnd(p))`

i think we can just remove this line and leave only the one below this (which is really the assertion we care about)
</blockquote></details>

Oops, good point. Fixed.

---

*[pkg/util/encoding/printer_test.go, line 49 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KywEsmnKNFUmR8P8zpH:-KywcNw_sMAYC-JZUWPv:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/db6311739d272839a9cdfa8a068ff2defc15e2f8/pkg/util/encoding/printer_test.go#L49)):*

comment that a `{0xff, 0xff}` test case wouldn't roundtrip, plz
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KywcS7M5I-4bCLWOsjZ:b7n55o7)*

benesch:
@RaduBerinde can you take one last look?

RaduBerinde:
:lgtm:

---

Review status: 1 of 8 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 106 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyweyLkuOn-Wq2hi0Zj:-KyweyLkuOn-Wq2hi0Zk:bd6eqhp) ([raw file](https://github.com/cockroachdb/cockroach/blob/d5c263f3d73f0540025acade6a8f66e742faf1f5/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L106)):*
> ```
> 
> # The span below ends at the end of the first index of table 53, and is not
> # constraining the the value of k2 or k3. This is confusing on first glance
> ```

the the

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 107 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-Kywf0SZlE64g6Mui1rK:-Kywf0SZlE64g6Mui1rL:b-7psvgz) ([raw file](https://github.com/cockroachdb/cockroach/blob/d5c263f3d73f0540025acade6a8f66e742faf1f5/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L107)):*
> ```
> # The span below ends at the end of the first index of table 53, and is not
> # constraining the the value of k2 or k3. This is confusing on first glance
> # because the second interleave in the hierarchy doesn't contain any unique
> ```

s/any unique values/any new PK/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-:-KywfYZwB_W3TqLufBqj:bnfp4nl)*

benesch:
Review status: 1 of 8 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 106 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-KyweyLkuOn-Wq2hi0Zj:-Kz-9LJVIuELq_ZSsyy2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d5c263f3d73f0540025acade6a8f66e742faf1f5/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L106)):*

the the
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/deep_interleaving, line 107 at r8](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034#-Kywf0SZlE64g6Mui1rK:-Kz-9LjbhTRWicRCFaMJ:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d5c263f3d73f0540025acade6a8f66e742faf1f5/pkg/sql/logictest/testdata/logic_test/deep_interleaving#L107)):*

s/any unique values/any new PK/
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20034)*

PR #20036, opened 2017-11-14
sql: fix span generation with IN

### LINKED ISSUE #20035: sql: incorrect span generation
Table schema:

```sql
CREATE TABLE favourites (
	resource_type STRING(30) NOT NULL,
	resource_key STRING(255) NOT NULL,
	device_group STRING(30) NOT NULL,
	customerid INT NOT NULL,
	jurisdiction STRING(2) NOT NULL,
	brand STRING(255) NOT NULL,
	created_ts TIMESTAMP NULL,
	guid_id STRING(100) NOT NULL,
	locale STRING(10) NOT NULL DEFAULT NULL,
	CONSTRAINT "primary" PRIMARY KEY (id ASC),
	UNIQUE INDEX favourites_idx (resource_type ASC, device_group ASC, resource_key ASC, customerid ASC),
	INDEX favourites_guid_idx (guid_id ASC),
	INDEX favourites_glob_fav_idx (resource_type ASC, device_group ASC, jurisdiction ASC, brand ASC, locale ASC, resource_key ASC),
	FAMILY "primary" (id, resource_type, resource_key, device_group, customerid, jurisdiction, brand, created_ts, guid_id, locale)
);
```

Query:

```sql
SELECT
  resource_key,
  count(resource_key) total
FROM favourites f1
WHERE f1.jurisdiction   = 'MT'
      AND   f1.brand          = 'xxx'
      AND   f1.resource_type  = 'GAME'
      AND   f1.device_group   = 'web'
      AND   f1.locale         = 'en_GB'
      AND   f1.resource_key IN ('ts', 'ts2', 'ts3')
GROUP BY resource_key
ORDER BY total DESC;
```

This query produces the following plan:

```
Level   Type    Field   Description
0       sort
0               order   -total
1       group
1               group by        @1-@1
2       render
3       scan
3               table   favourites@favourites_glob_fav_idx
3               spans   "/""GAME""/""web""/""MT""/""xxx""/""en_GB""/""ts3""-/""GAME""/""web""/""MT""/""xxx""/""en_GB""/""ts3\x00"""
```

This is incorrect, because the spans here restrict the results to only `ts3`. There should be 3 spans instead.

@RaduBerinde can you have a look?

RaduBerinde:
Issue #20035 exposes a bug in span generation. We have an IN with three values
and we only see one span.

The root cause is that we are appending to the same slices for each IN value,
which means that if the slice has some extra capacity, we overwrite the previous
value. The result is that we get multiple spans that are identical (which get
merged later into a single span).

Fixes #20035.

Release Note: Fixed a bug leading to incorrect results for queries with IN
constraints (in some cases).

### DIFF
--- a/pkg/sql/index_selection.go
+++ b/pkg/sql/index_selection.go
@@ -990,9 +990,17 @@ func applyInConstraint(
 		}
 		for _, s := range existingSpans {
 			if c.start != nil {
+				// Don't append directly to the existing slice, or we
+				// may introduce unwanted aliasing (see #20035).
+				old := s.start
+				s.start = make([]logicalKeyPart, 0, len(old)+len(parts))
+				s.start = append(s.start, old...)
 				s.start = append(s.start, parts...)
 			}
 			if c.end != nil {
+				old := s.end
+				s.end = make([]logicalKeyPart, 0, len(old)+len(parts))
+				s.end = append(s.end, old...)
 				s.end = append(s.end, parts...)
 			}
 			spans = append(spans, s)
--- a/pkg/sql/logictest/testdata/logic_test/select_index
+++ b/pkg/sql/logictest/testdata/logic_test/select_index
@@ -675,3 +675,73 @@ fetched: /test2/primary/...PK.../k/v -> /'014'/42
 fetched: /test2/primary/...PK.../k/v -> /'013'/42
 fetched: /test2/primary/...PK.../k/v -> /'012'/42
 fetched: /test2/primary/...PK.../k/v -> /'011'/42
+
+# Regression test for #20035.
+statement ok
+CREATE TABLE favourites (
+  id INT NOT NULL DEFAULT unique_rowid(),
+  resource_type STRING(30) NOT NULL,
+  resource_key STRING(255) NOT NULL,
+  device_group STRING(30) NOT NULL,
+  customerid INT NOT NULL,
+  jurisdiction STRING(2) NOT NULL,
+  brand STRING(255) NOT NULL,
+  created_ts TIMESTAMP NULL,
+  guid_id STRING(100) NOT NULL,
+  locale STRING(10) NOT NULL DEFAULT NULL,
+  CONSTRAINT "primary" PRIMARY KEY (id ASC),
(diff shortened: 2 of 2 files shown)


knz:
LGTM - please cherry pick on 1.1 too.

jordanlewis:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[pkg/sql/logictest/testdata/logic_test/select_index, line 678 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20036#-Kyv6Zze-xpJXyh8V-bd:-Kyv6Zze-xpJXyh8V-be:b-prb25a) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a56b35df105d307f7a9709eb4167b3f3a79c0b8/pkg/sql/logictest/testdata/logic_test/select_index#L678)):*
> ```
> fetched: /test2/primary/...PK.../k/v -> /'012'/42
> fetched: /test2/primary/...PK.../k/v -> /'011'/42
> ```

Please put a comment indicating this is a regression test for #20035 - lots of our logic tests are regression tests and it's nice to be able to figure out the purpose of a big complicated test by jumping to the issue that caused its creation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20036#-:-Kyv6nW1TTBfme32ybK9:bnfp4nl)*

RaduBerinde:
*[pkg/sql/logictest/testdata/logic_test/select_index, line 678 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20036#-Kyv6Zze-xpJXyh8V-bd:-KyvALpFrWlo7TzO1k_2:bk9zpac) ([raw file](https://github.com/cockroachdb/cockroach/blob/6a56b35df105d307f7a9709eb4167b3f3a79c0b8/pkg/sql/logictest/testdata/logic_test/select_index#L678)):*

Please put a comment indicating this is a regression test for #20035 - lots of our logic tests are regression tests and it's nice to be able to figure out the purpose of a big complicated test by jumping to the issue that caused its creation.
</blockquote></details>

Good call, forgot to do that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20036)*

Issue #16379, opened 2017-06-07
sqlbase: panic in RowFetcher.finalizeRow()


knz:
https://sentry.io/cockroach-labs/cockroachdb/issues/286804754/

```
*errors.errorString: *errors.errorString v3.go:353
  File "github.com/cockroachdb/cockroach/pkg/sql/pgwire/v3.go", line 353, in func1
  File "github.com/cockroachdb/cockroach/pkg/sql/executor.go", line 499, in func1
  File "github.com/cockroachdb/cockroach/pkg/sql/sqlbase/rowfetcher.go", line 595, in finalizeRow
  File "github.com/cockroachdb/cockroach/pkg/sql/sqlbase/rowfetcher.go", line 542, in NextRow
  File "github.com/cockroachdb/cockroach/pkg/sql/sqlbase/rowfetcher.go", line 552, in NextRowDecoded
...
(24 additional frame(s) were not displayed)

*errors.errorString v3.go:353
```

Stack trace points to:

```
        panic("Non-nullable column with no value!")
```

cc @danhhz

knz:
https://sentry.io/cockroach-labs/cockroachdb/issues/286804754/

vivekmenezes:
Was a schema change running while this was happening

knz:
The stack trace doesn't say

vivekmenezes:
No I meant was there an ongoing schema change in the background (add/drop column)? When we're adding/deleting a column that is NON Nullable we temporarily make it nullable during the rollout of the backfill. Just wondering if that is failing here.

vivekmenezes:
I think at a minimum we should dump the table descriptor and column descriptor in this panic to start with?

knz:
The Sentry reports are generated by auto-reporting from user clusters. There is no more information provided about the panic than the stack trace, version number and a couple of other build-related details. In particular there is (currently) no detail about other ongoing activities in the cluster.

knz:
cc @dt for suggested additions to the panic report

vivekmenezes:
assigning to @RaduBerinde because he is making lots of optimization here at the moment. Perhaps you an add more table descriptor information to the panic without compromising user privacy. Thanks!

RaduBerinde:
We should never be reading columns that are in the progress of backfilling, no? I thought those are not visible to queries until the backfill is complete.

vivekmenezes:
[...]

PR #20038, opened 2017-11-14
sql: Provide more info on panic for NULL value in non-nullable column

lgo:
Previously the panic was not very verbose. This will improve it so that
the user can pinpoint the row that was causing the issue. An example of
the panic is:

    panic: Non-nullable column "bad:id" with no value! Scanning index
    was "primary". The index key columns (id) had the values (1)

### DIFF
--- a/pkg/sql/sqlbase/multirowfetcher.go
+++ b/pkg/sql/sqlbase/multirowfetcher.go
@@ -17,6 +17,7 @@ package sqlbase
 import (
 	"bytes"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


justinj:
:lgtm: modulo merge conflicts

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20038#-:-KyvFy3NBqhmpz1frKRc:b-gcbo1u)*

lego:
Welp, I just stumbled on https://github.com/cockroachdb/cockroach/issues/16379 where an explicit decision was made to not provide additional info in the panic due to sensitive information. This problem did not occur to me. I think the best course of action here would be to revert this, and if we want 1) add a log statement to provide this information instead and 2) mention the choice a comment. cc: @vivekmenezes 

(fortunately, this didn't make it into an alpha release yet)

justinj:
huh, I thought all our messages sent to sentry were anonymized automatically? it's not immediately clear to me peeking through the slack channel, but perhaps that's only for `log.*`s?  perhaps this can just be replaced with a `log.Fatal`? @dt is the authoritative source I think.

vivekmenezes:
I'm pretty sure we should not panic with this sensitive information, so we will need to fix this to log the output. @dt knows more.

dt:
We redact panics to sentry unless they're log.Safe'ed, so you shouldn't need to worry about it

PR #20040, opened 2017-11-14
cherrypick-1.1: sql: fix span generation with IN

### LINKED ISSUE #20035: sql: incorrect span generation

RaduBerinde:
Issue #20035 exposes a bug in span generation. We have an IN with three values
and we only see one span.

The root cause is that we are appending to the same slices for each IN value,
which means that if the slice has some extra capacity, we overwrite the previous
value. The result is that we get multiple spans that are identical (which get
merged later into a single span).

Fixes #20035.

Release Note: Fixed a bug leading to incorrect results for queries with IN
constraints (in some cases).

Cherrypick of #20036. CC @cockroachdb/release

### DIFF
--- a/pkg/sql/index_selection.go
+++ b/pkg/sql/index_selection.go
@@ -984,9 +984,17 @@ func applyInConstraint(
 		}
 		for _, s := range existingSpans {
 			if c.start != nil {
+				// Don't append directly to the existing slice, or we
+				// may introduce unwanted aliasing (see #20035).
+				old := s.start
+				s.start = make([]logicalKeyPart, 0, len(old)+len(parts))
+				s.start = append(s.start, old...)
 				s.start = append(s.start, parts...)
 			}
 			if c.end != nil {
+				old := s.end
+				s.end = make([]logicalKeyPart, 0, len(old)+len(parts))
+				s.end = append(s.end, old...)
 				s.end = append(s.end, parts...)
 			}
 			spans = append(spans, s)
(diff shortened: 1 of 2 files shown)


knz (approved):
Thanks.

bdarnell (approved):

PR #20043, opened 2017-11-14
Makefile: various cleanup

benesch:


knz (commented):
Kudos on the idea!

jordanlewis:
:lgtm: except I don't think the last commit actually makes the Makefile easier to understand - IMO it makes it harder to understand because you have to go hunting for which declarations are relevant. Basically it feels more like spaghetti-code Make to me now. Open to being convinced the other way but for now I'd prefer that you submit this without the last patch.

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20043#-:-KyvasnaZcR24QEaIRl3:b-q4vb7p)*

benesch:
Ok, rebased the last commit out for now. TFTR!

PR #20044, opened 2017-11-14
sql: do not leak traces in lease renewal

tbg:
Found while investigating an unrelated problem via the /debug/requests endpoint.


lgo (approved):
Welp, thanks for catching this. LGTM

tschottdorf:
TFTR!

PR #20049, opened 2017-11-14
sql: support partitioning in SHOW CREATE TABLE

danhhz:


benesch:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 257 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20049#-KywJcpj2j8ZydyLO3Kv:-KywJcpj2j8ZydyLO3Kw:b-qqu8oe) ([raw file](https://github.com/cockroachdb/cockroach/blob/858145bd93e3362ce33fd07bbaf27892621909fd/pkg/sql/logictest/testdata/logic_test/partitioning#L257)):*
> ```
> ) PARTITION BY LIST (a) (
>    PARTITION p1 VALUES IN ((1)),
>    PARTITION p2 VALUES IN ((2))
> ```

I think I'd prefer if we dropped parens around elements that appear alone, but it's a very weak preference. (Same for range partitioning if you go that route.)

---

*[pkg/sql/sqlbase/structured.go, line 1358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20049#-KywIO4a5nPxO9gc6NgM:-KywIO4bCyDiTlHzdn8i:b9su25q) ([raw file](https://github.com/cockroachdb/cockroach/blob/858145bd93e3362ce33fd07bbaf27892621909fd/pkg/sql/sqlbase/structured.go#L1358)):*
> ```Go
> 
> // PrintPartitioningTuple prints the first `numColumns` datums in a partitioning
> // tuple to `buf`. If `numColumns` is greated than `len(buf)`, then s (which is
> ```

"greated" and also it's`len(datums)`, no? to be super pedantic it's technically when `numColumns` is greater than or equal to `len(datums)`

perhaps:  "If numColumns specifies more datums than are present in `datums`..."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20049#-:-KywJtltKrR84mEPPnrg:bnfp4nl)*

danhhz:
TFTRs!

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning, line 257 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20049#-KywJcpj2j8ZydyLO3Kv:-KywscdJl0N0_W9JfmaJ:b5brmba) ([raw file](https://github.com/cockroachdb/cockroach/blob/858145bd93e3362ce33fd07bbaf27892621909fd/pkg/sql/logictest/testdata/logic_test/partitioning#L257)):*

I think I'd prefer if we dropped parens around elements that appear alone, but it's a very weak preference. (Same for range partitioning if you go that route.)
</blockquote></details>

After discussing offline, we're both have differing weak preferences on this, so I'm going to leave it as is.

---

*[pkg/sql/sqlbase/structured.go, line 1358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20049#-KywIO4a5nPxO9gc6NgM:-Kywsl2m1ExNVgl9YfSX:bb6un1j) ([raw file](https://github.com/cockroachdb/cockroach/blob/858145bd93e3362ce33fd07bbaf27892621909fd/pkg/sql/sqlbase/structured.go#L1358)):*

"greated" and also it's`len(datums)`, no? to be super pedantic it's technically when `numColumns` is greater than or equal to `len(datums)`

perhaps:  "If numColumns specifies more datums than are present in `datums`..."
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20049#-:-Kywstc6xneML9TVNlnU:bz5mair)*

PR #20054, opened 2017-11-14
rfcs: Change "follow the sun" to "follow the user"

a-robinson:


tschottdorf:
LGTM, but I believe various mentions of the sun exist in the code. Mind changing those, too?

a-robinson:
Yeah, there was one comment. Fixed.

bdarnell:
What's the motivation for this change? I think "follow the sun" is more accurate (and more evocative), because this feature (leaseholder locality) is designed for cases where the data is not user-specific but may be accessed by different users at different times of day. 

"Follow the sun" also captures the reason why this feature remains relevant even as we introduce partitioning features that can restrict data to the user's location so no "following" is necessary.

a-robinson:
"Follow the sun" is certainly more evocative, but engineers have frequently been mislead by the name, thinking that it implies some sort of time-based mechanism. It also doesn't cover cases where load isn't segmented by time of day but by some aspect of the primary key distribution across localities. @robert-s-lee said that while he often used the term "follow the sun" to introduce the feature, he normally switched to calling it "follow the user" more frequently to explain what it's really doing.

cc @awoods187 in case I missed anything important.

benesch:
Drive by suggestion: "follow the load."

PR #19499, opened 2017-10-24
sql: propagate column properties to elements

justinj:
Release notes: Fix interaction with  modified column types such as
length-limited strings in arrays.

Previously we would throw away any column modifiers when a type was used
with an array. This change addresses this problem.

Fixes #19237.

PR #20055, opened 2017-11-14
pkg/sql/parser: remove accidentally checked-in goyacc.installed

benesch:
This was accidentally checked in by #19499 (f458af7). (It was previously
git-ignored, so easy to overlook after running git add.)


justinj:
Oops! LGTM, thanks for the catch!

PR #20057, opened 2017-11-14
server/status: Don't log every 10 seconds about nodes without an address

a-robinson:
On the PM cluster, this is being logged every 10 seconds on every node
because one of the nodes in the cluster has been dead for hours.


tschottdorf:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20057#-:-KywEhUXCa4fmfSJMo4u:bnfp4nl)*

PR #20062, opened 2017-11-14
docs: Update release note commit guidance

jseldess:
To make it easy to pull together release notes, the recommendation going forward is to include a release note description in every commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,27 +112,29 @@ following guidelines:
   to indicate their reach.
 
 - We publish detailed [release notes](https://www.cockroachlabs.com/docs/releases/)
-  describing most non-test changes. To facilitate this, at least one commit in every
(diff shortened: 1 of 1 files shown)


tschottdorf:
I thought we had decided that the guidance would be that "at least one" commit in the PR must contain one. I think it's overkill to request one for each commit, and since I think the guidance needs linting to be effective, it would introduce too much drag.

jseldess:
@tschottdorf, my thinking is that multiple commits in the same PR sometimes require different release note descriptions.

tschottdorf:
@jseldess that's true sometimes, but your change stipulates that we'll enforce having it on *each commit* where it should be (or so I thought) on at least one commit.

jseldess:
After an offline chat with @tschottdorf, I understand now that our current guidance is sufficient, so I'm closing this. Hopefully we can start linting to check for at least one commit message per PR with a `Release note` description.

PR #20063, opened 2017-11-14
sql: tests for incorrect attempt to reference mutation columns

### LINKED ISSUE #19890: Require more test coverage: incorrect references to mutation columns via scanNode
[A bug was discovered that caused a panic](https://github.com/cockroachdb/cockroach/pull/19752) when the assumption was made that the columns in `scanNode.valNeededForCol`, `scanNode.cols`, etc. exist in the corresponding index descriptors passed along to other objects. This is of course not the case with mutation columns that are not yet available in descriptor leases.

We should increase the test coverage on mutations as proposed by @vivekmenezes.

vivekmenezes:
The bug itself was inadvertently fixed in #19701. See the change
to isCoveringIndex() in that PR.

fixes #19890

### DIFF
--- a/pkg/sql/index_selection.go
+++ b/pkg/sql/index_selection.go
@@ -801,7 +801,7 @@ func (v *indexInfo) isCoveringIndex(scan *scanNode) bool {
 	for _, colIdx := range scan.valNeededForCol.Ordered() {
 		// This is possible during a schema change when we have
 		// additional mutation columns.
-		if colIdx >= len(v.desc.Columns) {
+		if colIdx >= len(v.desc.Columns) && len(v.desc.Mutations) > 0 {
 			return false
 		}
 		colID := v.desc.Columns[colIdx].ID
--- a/pkg/sql/descriptor_mutation_test.go
+++ b/pkg/sql/descriptor_mutation_test.go
@@ -166,9 +166,11 @@ func TestOperationsWithColumnMutation(t *testing.T) {
 
 	// Fix the column families so the key counts below don't change if the
 	// family heuristics are updated.
+	// Add an index so that we test adding a column when a table has an index.
 	if _, err := sqlDB.Exec(`
 CREATE DATABASE t;
 CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR, i CHAR DEFAULT 'i', FAMILY (k), FAMILY (v), FAMILY (i));
+CREATE INDEX allidx ON t.test (k, v);
 `); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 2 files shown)


richardwu:
Hrm, I just realized I forgot to fix https://github.com/cockroachdb/cockroach/pull/19701/files#diff-32a45104bdefeaeb4868cc4521d5c05bR805 to return `true` instead of false (since the index is still useful for all the columns up to the mutation column). Would appreciate it if you could change that to true!

:lgtm: 

---

---

*[pkg/sql/descriptor_mutation_test.go, line 173 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20063#-KywRJw2MOedq78GX4I0:-KywRJw2MOedq78GX4I1:blxpx52) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2d51f2b8bb2c36567aff81779c29fd5749747fe/pkg/sql/descriptor_mutation_test.go#L173)):*
> ```Go
> CREATE DATABASE t;
> CREATE TABLE t.test (k CHAR PRIMARY KEY, v CHAR, i CHAR DEFAULT 'i', FAMILY (k), FAMILY (v), FAMILY (i));
> CREATE INDEX allidx ON t.test (k, v);
> ```

Just out of curiosity, what would happen to this index if `i` was also included while `i` is being deleted?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20063#-:-KywQaNcvjq8wB9N_ouC:baumw05)*

richardwu (approved):

pkg/sql/index_selection.go
```diff
@@ -801,7 +801,7 @@ func (v *indexInfo) isCoveringIndex(scan *scanNode) bool {
 	for _, colIdx := range scan.valNeededForCol.Ordered() {
 		// This is possible during a schema change when we have
 		// additional mutation columns.
-		if colIdx >= len(v.desc.Columns) {
+		if colIdx >= len(v.desc.Columns) && len(v.desc.Mutations) > 0 {
```
richardwu: Good catch 👍

vivekmenezes:
I don't know what you're asking for.  If we do change something we should test it.

---

---

*[pkg/sql/descriptor_mutation_test.go, line 173 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20063#-KywRJw2MOedq78GX4I0:-KyxY0V_VcBde1Cf8nXW:b-y2jdhx) ([raw file](https://github.com/cockroachdb/cockroach/blob/b2d51f2b8bb2c36567aff81779c29fd5749747fe/pkg/sql/descriptor_mutation_test.go#L173)):*

Just out of curiosity, what would happen to this index if `i` was also included while `i` is being deleted?
</blockquote></details>

You cannot drop a column while it's part of an index. the column DROP command will fail

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20063#-:-KyxZETUQIKtPoXEiKPz:b-azrhjq)*

PR #20064, opened 2017-11-14
sql: Adds ON DELETE CASCADE for foreign key references

BramGruneir:
This is still a work in progress, but I wanted to put it out there to start to get some early feedback.
Major things left to do before this can be considered mergeable:
- Tighten up the TablesNeededForFKs to not just be a full transitive closure.
- Add a depth limit to the recursive calls to row deleter so we don't panic.
- More testing.
- Remove extra logging debug statements.

---

Commit message follows...
- Enables the addition of the ON DELETE CASCADE action for foreign keys references.
- Augments the TablesNeededForFKs function to fetch all tables that might be needed.
- Adds a cascade helper to the fkDeleteHelper.
- DeleteRow now calls cascadeAll before attempting to delete the row.
- Adds a new cascade logic test specifically for testing the convoluted cascade scenarios.

There are a ton of improvements and additions that should happen here. This is just a MVP implementation.

### DIFF
--- a/pkg/sql/planner.go
+++ b/pkg/sql/planner.go
@@ -343,19 +343,17 @@ func (p *planner) exec(ctx context.Context, sql string, args ...interface{}) (in
 	return countRowsAffected(params, plan)
 }
 
-func (p *planner) fillFKTableMap(ctx context.Context, m sqlbase.TableLookupsByID) error {
-	for tableID := range m {
-		table, err := p.session.tables.getTableVersionByID(ctx, p.txn, tableID)
+func (p *planner) lookupFKTable(
+	ctx context.Context, tableID sqlbase.ID,
+) (sqlbase.TableLookup, error) {
+	table, err := p.session.tables.getTableVersionByID(ctx, p.txn, tableID)
+	if err != nil {
 		if err == errTableAdding {
-			m[tableID] = sqlbase.TableLookup{IsAdding: true}
-			continue
+			return sqlbase.TableLookup{IsAdding: true}, nil
 		}
-		if err != nil {
-			return err
-		}
-		m[tableID] = sqlbase.TableLookup{Table: table}
+		return sqlbase.TableLookup{}, err
 	}
-	return nil
+	return sqlbase.TableLookup{Table: table}, nil
 }
 
 // isDatabaseVisible returns true if the given database is visible
--- a/pkg/sql/backfill.go
+++ b/pkg/sql/backfill.go
@@ -527,7 +527,9 @@ func (sc *SchemaChanger) distBackfill(
 			// backfiller processor.
 			var otherTableDescs []sqlbase.TableDescriptor
 			if backfillType == columnBackfill {
-				fkTables := sqlbase.TablesNeededForFKs(*tableDesc, sqlbase.CheckUpdates)
+				fkTables, _ := sqlbase.TablesNeededForFKs(
+					ctx, *tableDesc, sqlbase.CheckUpdates, sqlbase.NoLookup, sqlbase.NoCheckPrivilege,
+				)
 				for k := range fkTables {
 					table, err := tc.getTableVersionByID(ctx, txn, k)
 					if err != nil {
--- a/pkg/sql/create_table.go
+++ b/pkg/sql/create_table.go
@@ -441,7 +441,8 @@ func resolveFK(
 	}
 
 	if d.Actions.Delete != tree.NoAction &&
-		d.Actions.Delete != tree.Restrict {
+		d.Actions.Delete != tree.Restrict &&
+		d.Actions.Delete != tree.Cascade {
 		feature := fmt.Sprintf("unsupported: ON DELETE %s", d.Actions.Delete)
 		return pgerror.Unimplemented(feature, feature)
 	}
--- a/pkg/sql/delete.go
+++ b/pkg/sql/delete.go
@@ -70,8 +70,10 @@ func (p *planner) Delete(
 		requestedCols = en.tableDesc.Columns
 	}
 
-	fkTables := sqlbase.TablesNeededForFKs(*en.tableDesc, sqlbase.CheckDeletes)
-	if err := p.fillFKTableMap(ctx, fkTables); err != nil {
+	fkTables, err := sqlbase.TablesNeededForFKs(
+		ctx, *en.tableDesc, sqlbase.CheckDeletes, p.lookupFKTable, p.CheckPrivilege,
+	)
+	if err != nil {
 		return nil, err
 	}
 	rd, err := sqlbase.MakeRowDeleter(p.txn, en.tableDesc, fkTables, requestedCols,
--- a/pkg/sql/distsqlrun/columnbackfiller.go
+++ b/pkg/sql/distsqlrun/columnbackfiller.go
@@ -150,7 +150,9 @@ func (cb *columnBackfiller) runChunk(
 			defer cb.flowCtx.testingKnobs.RunAfterBackfillChunk()
 		}
 
-		fkTables := sqlbase.TablesNeededForFKs(tableDesc, sqlbase.CheckUpdates)
+		fkTables, _ := sqlbase.TablesNeededForFKs(
+			ctx, tableDesc, sqlbase.CheckUpdates, sqlbase.NoLookup, sqlbase.NoCheckPrivilege,
+		)
 		for _, fkTableDesc := range cb.spec.OtherTables {
 			found, ok := fkTables[fkTableDesc.ID]
 			if !ok {
--- a/pkg/sql/insert.go
+++ b/pkg/sql/insert.go
@@ -155,8 +155,10 @@ func (p *planner) Insert(
 		}
 	}
 
-	fkTables := sqlbase.TablesNeededForFKs(*en.tableDesc, sqlbase.CheckInserts)
-	if err := p.fillFKTableMap(ctx, fkTables); err != nil {
+	fkTables, err := sqlbase.TablesNeededForFKs(
+		ctx, *en.tableDesc, sqlbase.CheckInserts, p.lookupFKTable, p.CheckPrivilege,
+	)
+	if err != nil {
 		return nil, err
 	}
 	ri, err := sqlbase.MakeRowInserter(p.txn, en.tableDesc, fkTables, cols,
@@ -218,8 +220,10 @@ func (p *planner) Insert(
 				return nil, err
 			}
 
-			fkTables := sqlbase.TablesNeededForFKs(*en.tableDesc, sqlbase.CheckUpdates)
-			if err := p.fillFKTableMap(ctx, fkTables); err != nil {
+			fkTables, err := sqlbase.TablesNeededForFKs(
+				ctx, *en.tableDesc, sqlbase.CheckUpdates, p.lookupFKTable, p.CheckPrivilege,
+			)
+			if err != nil {
 				return nil, err
 			}
 			tu := tableUpserterPool.Get().(*tableUpserter)
--- a/pkg/sql/logictest/testdata/logic_test/cascade
+++ b/pkg/sql/logictest/testdata/logic_test/cascade
@@ -1,0 +1,823 @@
+# LogicTest: default parallel-stmts distsql
+
+subtest DeleteCascade_Basic
+### Basic Delete Cascade
+#     a
+#    / \
+#   b1 b2
+#  / \
+# c1  c2
+
+statement ok
+CREATE TABLE a (
+  id STRING PRIMARY KEY
+);
+
+statement ok
+CREATE TABLE b1 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES a ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE b2 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES a ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c1 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES b1 ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c2 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES b1 ON DELETE CASCADE
+);
+
+statement ok
+INSERT INTO a VALUES ('a-pk1');
+
+statement ok
+INSERT INTO b1 VALUES ('b1-pk1', 'a-pk1');
+
+statement ok
+INSERT INTO b1 VALUES ('b1-pk2', 'a-pk1');
+
+statement ok
+INSERT INTO b2 VALUES ('b2-pk1', 'a-pk1');
+
+statement ok
+INSERT INTO b2 VALUES ('b2-pk2', 'a-pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk1-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk2-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk3-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk4-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk1-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk2-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk3-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk4-b1-pk2', 'b1-pk2');
+
+# ON DELETE CASCADE
+statement ok
+DELETE FROM a WHERE id = 'a-pk1';
+
+# Clean up after the test.
+statement ok
+DROP TABLE c2;
+
+statement ok
+DROP TABLE c1;
+
+statement ok
+DROP TABLE b2;
+
+statement ok
+DROP TABLE b1;
+
+statement ok
+DROP TABLE a;
+
+subtest DeleteCascade_PrimaryKeys
+### Basic Delete Cascade using primary keys
+#     a
+#    / \
+#   b1 b2
+#  / \
+# c1  c2
+
+statement ok
+CREATE TABLE a (
+  id STRING PRIMARY KEY
+);
+
+statement ok
+CREATE TABLE b1 (
+  id STRING PRIMARY KEY REFERENCES a ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE b2 (
+  id STRING PRIMARY KEY REFERENCES a ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c1 (
+  id STRING PRIMARY KEY REFERENCES b1 ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c2 (
+  id STRING PRIMARY KEY REFERENCES b1 ON DELETE CASCADE
+);
+
+statement ok
+INSERT INTO a VALUES ('pk1');
+
+statement ok
+INSERT INTO b1 VALUES ('pk1');
+
+statement ok
+INSERT INTO b2 VALUES ('pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('pk1');
+
+statement ok
+INSERT INTO c2 VALUES ('pk1');
+
+# ON DELETE CASCADE
+statement ok
+DELETE FROM a WHERE id = 'pk1';
+
+# Clean up after the test.
+statement ok
+DROP TABLE c2;
+
+statement ok
+DROP TABLE c1;
+
+statement ok
+DROP TABLE b2;
+
+statement ok
+DROP TABLE b1;
+
+statement ok
+DROP TABLE a;
+
+subtest DeleteCascade_CompositeFKs
+### Basic Delete Cascade with composite FKs
+#     a
+#    / \
+#   b1 b2
+#  / \
+# c1  c2
+
+statement ok
+CREATE TABLE a (
+  id STRING PRIMARY KEY
+ ,x INT
+ ,UNIQUE (id, x)
+);
+
+statement ok
+CREATE TABLE b1 (
+  id STRING PRIMARY KEY
+ ,a_id STRING
+ ,x INT
+ ,y INT
+ ,INDEX (a_id, x, y)
+ ,FOREIGN KEY (a_id, x) REFERENCES a (id, x) ON DELETE CASCADE
+ ,UNIQUE (id, x)
+);
+
+statement ok
+CREATE TABLE b2 (
+  id STRING PRIMARY KEY
+ ,a_id STRING
+ ,x INT
+ ,y INT
+ ,INDEX (a_id, x, y)
+ ,FOREIGN KEY (a_id, x) REFERENCES a (id, x) ON DELETE CASCADE
+ ,UNIQUE (id, x)
+);
+
+statement ok
+CREATE TABLE c1 (
+  id STRING PRIMARY KEY
+ ,b_id STRING
+ ,x INT
+ ,FOREIGN KEY (b_id, x) REFERENCES b1 (id, x) ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c2 (
+  id STRING PRIMARY KEY
+ ,b_id STRING
+ ,x INT
+ ,FOREIGN KEY (b_id, x) REFERENCES b1 (id, x) ON DELETE CASCADE
+);
+
+statement ok
+INSERT INTO a VALUES ('a-pk1', 1);
+
+statement ok
+INSERT INTO b1 VALUES ('b1-pk1', 'a-pk1', 1, 1);
+
+statement ok
+INSERT INTO b1 VALUES ('b1-pk2', 'a-pk1', 1, 2);
+
+statement ok
+INSERT INTO b2 VALUES ('b2-pk1', 'a-pk1', 1, 1);
+
+statement ok
+INSERT INTO b2 VALUES ('b2-pk2', 'a-pk1', 1, 2);
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk1-b1-pk1', 'b1-pk1', 1);
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk2-b1-pk1', 'b1-pk1', 1);
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk3-b1-pk2', 'b1-pk2', 1);
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk4-b1-pk2', 'b1-pk2', 1);
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk1-b1-pk1', 'b1-pk1', 1);
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk2-b1-pk1', 'b1-pk1', 1);
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk3-b1-pk2', 'b1-pk2', 1);
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk4-b1-pk2', 'b1-pk2', 1);
+
+# ON DELETE CASCADE
+statement ok
+DELETE FROM a WHERE id = 'a-pk1';
+
+# Clean up after the test.
+statement ok
+DROP TABLE c2;
+
+statement ok
+DROP TABLE c1;
+
+statement ok
+DROP TABLE b2;
+
+statement ok
+DROP TABLE b1;
+
+statement ok
+DROP TABLE a;
+
+subtest DeleteCascade_Restrict
+### Basic Delete Cascade with Restrict
+#     a
+#    / \
+#   b1 b2
+#  / \
+# c1  c2
+#     |
+#     d
+
+statement ok
+CREATE TABLE a (
+  id STRING PRIMARY KEY
+);
+
+statement ok
+CREATE TABLE b1 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES a ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE b2 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES a ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c1 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES b1 ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE c2 (
+  id STRING PRIMARY KEY
+ ,delete_cascade STRING NOT NULL REFERENCES b1 ON DELETE CASCADE
+);
+
+statement ok
+CREATE TABLE d (
+  id STRING PRIMARY KEY
+ ,delete_restrict STRING NOT NULL REFERENCES c2 ON DELETE RESTRICT
+);
+
+statement ok
+INSERT INTO a VALUES ('a-pk1');
+
+statement ok
+INSERT INTO b1 VALUES ('b1-pk1', 'a-pk1');
+
+statement ok
+INSERT INTO b1 VALUES ('b1-pk2', 'a-pk1');
+
+statement ok
+INSERT INTO b2 VALUES ('b2-pk1', 'a-pk1');
+
+statement ok
+INSERT INTO b2 VALUES ('b2-pk2', 'a-pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk1-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk2-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk3-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO c1 VALUES ('c1-pk4-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk1-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk2-b1-pk1', 'b1-pk1');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk3-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO c2 VALUES ('c2-pk4-b1-pk2', 'b1-pk2');
+
+statement ok
+INSERT INTO d VALUES ('d-pk1-c2-pk4-b1-pk2', 'c2-pk4-b1-pk2');
+
+# ON DELETE CASCADE
+statement error pq: foreign key violation: values \['c2-pk4-b1-pk2'\] in columns \[id\] referenced in table "d"
+DELETE FROM a WHERE id = 'a-pk1';
+
+# Clean up after the test.
+statement ok
+DROP TABLE d;
+
+statement ok
+DROP TABLE c2;
+
+statement ok
+DROP TABLE c1;
+
+statement ok
+DROP TABLE b2;
+
+statement ok
+DROP TABLE b1;
+
+statement ok
+DROP TABLE a;
+
+subtest DeleteCascade_Interleaved
+### Basic Delete Cascade with Interleaved Tables
+#     a
+#    / \
+#   b1 b2
+#  / \   \
+# c1  c2  c3
+
+statement ok
+CREATE TABLE a (
+  id STRING PRIMARY KEY
+);
+
+statement ok
+CREATE TABLE b1 (
+  id STRING PRIMARY KEY REFERENCES a ON DELETE CASCADE
+) INTERLEAVE IN PARENT a (id);
+
+statement ok
+CREATE TABLE b2 (
+  id STRING PRIMARY KEY REFERENCES a ON DELETE CASCADE
+) INTERLEAVE IN PARENT a (id);
+
+statement ok
+CREATE TABLE c1 (
+  id STRING PRIMARY KEY REFERENCES b1 ON DELETE CASCADE
+) INTERLEAVE IN PARENT b1 (id);
+
+statement ok
+CREATE TABLE c2 (
+  id STRING PRIMARY KEY REFERENCES b1 ON DELETE CASCADE
+) INTERLEAVE IN PARENT b1 (id);
+
+statement ok
+CREATE TABLE c3 (
+  id STRING PRIMARY KEY REFERENCES b2 ON DELETE CASCADE
+) INTERLEAVE IN PARENT b2 (id);
+
+statement ok
+INSERT INTO a VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO b1 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO b2 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO c1 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO c2 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO c3 VALUES ('pk1'), ('pk2');
+
+# ON DELETE CASCADE from b1 downward
+statement ok
+DELETE FROM b1 WHERE id = 'pk2';
+
+# ON DELETE CASCADE
+statement ok
+DELETE FROM a WHERE id = 'pk1';
+
+# Clean up after the test.
+statement ok
+DROP TABLE c3;
+
+statement ok
+DROP TABLE c2;
+
+statement ok
+DROP TABLE c1;
+
+statement ok
+DROP TABLE b2;
+
+statement ok
+DROP TABLE b1;
+
+statement ok
+DROP TABLE a;
+
+subtest DeleteCascade_InterleavedRestrict
+### Basic Delete Cascade with Interleaved Tables To Restrict
+#     a
+#    / \
+#   b1 b2
+#  / \   \
+# c1  c2  c3
+#
+# In this test, c3 is restricted, so deleting from a should fail, but from b1
+# should be ok.
+
+statement ok
+CREATE TABLE a (
+  id STRING PRIMARY KEY
+);
+
+statement ok
+CREATE TABLE b1 (
+  id STRING PRIMARY KEY REFERENCES a ON DELETE CASCADE
+) INTERLEAVE IN PARENT a (id);
+
+statement ok
+CREATE TABLE b2 (
+  id STRING PRIMARY KEY REFERENCES a ON DELETE CASCADE
+) INTERLEAVE IN PARENT a (id);
+
+statement ok
+CREATE TABLE c1 (
+  id STRING PRIMARY KEY REFERENCES b1 ON DELETE CASCADE
+) INTERLEAVE IN PARENT b1 (id);
+
+statement ok
+CREATE TABLE c2 (
+  id STRING PRIMARY KEY REFERENCES b1 ON DELETE CASCADE
+) INTERLEAVE IN PARENT b1 (id);
+
+statement ok
+CREATE TABLE c3 (
+  id STRING PRIMARY KEY REFERENCES b2 ON DELETE RESTRICT
+) INTERLEAVE IN PARENT b2 (id);
+
+statement ok
+INSERT INTO a VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO b1 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO b2 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO c1 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO c2 VALUES ('pk1'), ('pk2');
+
+statement ok
+INSERT INTO c3 VALUES ('pk1'), ('pk2');
+
+# ON DELETE CASCADE from b1 downward
+statement ok
+DELETE FROM b1 WHERE id = 'pk2';
+
+# ON DELETE CASCADE
+statement error pq: foreign key violation: values \['pk1'\] in columns \[id\] referenced in table "c3"
+DELETE FROM a WHERE id = 'pk1';
+
+# Clean up after the test.
+statement ok
+DROP TABLE c3;
+
+statement ok
+DROP TABLE c2;
+
+statement ok
+DROP TABLE c1;
+
+statement ok
+DROP TABLE b2;
+
+statement ok
+DROP TABLE b1;
+
+statement ok
+DROP TABLE a;
+
+subtest DeleteCascade_SelfReference
+### Self Reference Delete Cascade
+# self <- self
+
+statement ok
+CREATE TABLE self (
+  id INT PRIMARY KEY
+ ,other_id INT REFERENCES self ON DELETE CASCADE
+);
+
+statement ok
+INSERT INTO self VALUES (1, NULL);
+
+statement ok
+INSERT INTO self VALUES (2, 1);
+
+statement ok
+INSERT INTO self VALUES (3, 2);
+
+statement ok
+INSERT INTO self VALUES (4, 3);
+
+statement ok
+DELETE FROM self WHERE id = 1;
+
+# Clean up after the test.
+statement ok
+DROP TABLE self CASCADE;
+
+subtest DeleteCascade_SelfReferenceCycle
+### Self Reference Delete Cascade Cycle
+# self <- self
+
+statement ok
+CREATE TABLE self (
+  id INT PRIMARY KEY
+ ,other_id INT REFERENCES self ON DELETE CASCADE
+);
+
+statement ok
+INSERT INTO self VALUES (1, NULL);
+
+statement ok
+INSERT INTO self VALUES (2, 1);
+
+statement ok
+INSERT INTO self VALUES (3, 2);
+
+statement ok
+INSERT INTO self VALUES (4, 3);
+
+statement ok
+UPDATE self SET other_id = 4 WHERE id = 1;
+
+statement ok
+DELETE FROM self WHERE id = 1;
+
+# Clean up after the test.
+statement ok
+DROP TABLE self CASCADE;
+
+subtest DeleteCascade_TwoTableLoop
+### Delete cascade loop between two tables
+# loop_a <- loop_b
+# loop_b <- loop_a
+
+statement ok
+CREATE TABLE loop_a (
+  id STRING PRIMARY KEY
+ ,cascade_delete STRING
+ ,INDEX(cascade_delete)
+);
+
+statement ok
+CREATE TABLE loop_b (
+  id STRING PRIMARY KEY
+ ,cascade_delete STRING REFERENCES loop_a ON DELETE CASCADE
+);
+
+statement ok
+ALTER TABLE loop_a ADD CONSTRAINT cascade_delete_constraint
+  FOREIGN KEY (cascade_delete) REFERENCES loop_b (id)
+  ON DELETE CASCADE;
+
+statement ok
+INSERT INTO loop_a (id, cascade_delete) VALUES ('loop_a-pk1', NULL);
+
+statement ok
+INSERT INTO loop_b (id, cascade_delete) VALUES ('loop_b-pk1', 'loop_a-pk1');
+
+statement ok
+INSERT INTO loop_a (id, cascade_delete) VALUES ('loop_a-pk2', 'loop_b-pk1');
+
+statement ok
+INSERT INTO loop_b (id, cascade_delete) VALUES ('loop_b-pk2', 'loop_a-pk2');
+
+statement ok
+INSERT INTO loop_a (id, cascade_delete) VALUES ('loop_a-pk3', 'loop_b-pk2');
+
+statement ok
+INSERT INTO loop_b (id, cascade_delete) VALUES ('loop_b-pk3', 'loop_a-pk3');
+
+statement ok
+UPDATE loop_a SET cascade_delete = 'loop_b-pk3' WHERE id = 'loop_a-pk1';
+
+statement ok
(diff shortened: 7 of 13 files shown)


justinj:
If you're looking for test cases, you might find Postgres' fk tests useful (obv gonna have to filter out the stuff that doesn't apply to us): https://github.com/postgres/postgres/blob/master/src/test/regress/sql/foreign_key.sql (the expected output is a directory up in `output`)

---

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1QJQRabw-OAvDXk3A:-Kz1QJQRabw-OAvDXk3B:b-2r61yy) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/logictest/testdata/logic_test/cascade#L1)):*
> ```
> # LogicTest: default parallel-stmts distsql
> ```

I want to make the perhaps bold claim that our logic tests are insufficient for testing the kinds of structures we want to be able to test for this feature. It feels to me like what would really make this easier to test would be some kind of DSL that would allow you to easily describe the kinds of directed graph structures that you're writing here? Obviously these tests are very valuable but it feels like more effort than it should be to parse the relationships between the various tables, and the logic tests don't have any kind of infrastructure for the kind of abstraction you'd want for that.

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 8 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1QwCAafQG7eV5UwME:-Kz1QwCAafQG7eV5UwMF:b-d2vxo0) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/logictest/testdata/logic_test/cascade#L8)):*
> ```
> #   b1 b2
> #  / \
> # c1  c2
> ```

Would it also be beneficial to have a test for the interaction between a cascaded delete that would be blocked by an separate `ON DELETE RESTRICT` (and any of the other relevant edge cases you listed in the RFC)?

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 207 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz-hrKWJgGtIldgD21J:-Kz-hrKWJgGtIldgD21K:b-23oau4) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/logictest/testdata/logic_test/cascade#L207)):*
> ```
> 
> statement ok
> DELETE FROM race_a WHERE id = 'a1';
> ```

nit: missing trailing newline

---

*[pkg/sql/sqlbase/fk.go, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1S5WrYdnI-YgExrsn:-Kz1S5WrYdnI-YgExrso:b-qte21v) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L77)):*
> ```Go
> 	}
> 	elem := (*q)[0]
> 	*q = (*q)[1:]
> ```

this is a cool Go pattern for a queue, I hadn't seen it

---

*[pkg/sql/sqlbase/fk.go, line 380 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1WDSkaEqIPPW_kLzK:-Kz1WDSkaEqIPPW_kLzL:bh7smuk) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L380)):*
> ```Go
> 				return h, err
> 			}
> 			idxDesc, _ := otherTables[ref.Table].Table.FindIndexByID(ref.Index)
> ```

any reason not to check this error?

---

*[pkg/sql/sqlbase/fk.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1Y3puOcY6AVVJpUpK:-Kz1Y3puOcY6AVVJpUpL:b2uxpxd) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L609)):*
> ```Go
> }
> 
> type cascadeHelper struct {
> ```

I know the other helpers don't, but I feel like these helper structs would benefit from a comment describing their relationship to the operations they're helping with - if I understand correctly, there's one `cascadeHelper` per index being cascaded upon?

---

*[pkg/sql/sqlbase/fk.go, line 612 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1YcWUJfMoJLYm7U_f:-Kz1YcWUJfMoJLYm7U_g:b-xskf83) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L612)):*
> ```Go
> 	txn          *client.Txn
> 	table        *TableDescriptor // the table being searched for rows to cascade
> 	index        *IndexDescriptor // the fk index that will contain the entires to cascade
> ```

s/entires/entries/

---

*[pkg/sql/sqlbase/fk.go, line 725 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1ZpsT0r8BJ6fDElsO:-Kz1ZpsT0r8BJ6fDElsP:beb1lvf) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L725)):*
> ```Go
> }
> 
> // spanForIndexValues creates a span against an index to extra the primary keys
> ```

`s/extra/extract/` ?

---

*[pkg/sql/sqlbase/fk.go, line 794 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1_H0jDhHYYgvIq1Ol:-Kz1_H0jDhHYYgvIq1Om:b-x08zad) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L794)):*
> ```Go
> 			primaryKey = append(tree.Datums(nil), primaryKey...)
> 			primaryKeysToDel = append(primaryKeysToDel, primaryKey)
> 			log.Warningf(ctx, "******** pks to delete: %+v", primaryKey)
> ```

should this logging stick around?

(oh, i see now you mentioned this in the commit!)

---

*[pkg/sql/sqlbase/fk.go, line 844 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1iwJc_e_1M-7Ve8XL:-Kz1iwJc_e_1M-7Ve8XM:b-h3pw70) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L844)):*
> ```Go
> 	}
> 
> 	// Finally delete the rows.
> ```

`Finally,`? unless the command is meant to sound exasperated :D

---

*[pkg/sql/sqlbase/fk.go, line 845 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1aDubs6VfXM2LxZr9:-Kz1aDubs6VfXM2LxZrA:bg2egh8) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L845)):*
> ```Go
> 
> 	// Finally delete the rows.
> 	// TODO(bram): If the row was already deleted, this should not be a failure.
> ```

would this happen if a row fk-referenced another row more than once?

---

*[pkg/sql/sqlbase/fk.go, line 848 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1a5hFbIQDBTQHNpic:-Kz1a5hFbIQDBTQHNpid:b-n6k75d) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L848)):*
> ```Go
> 	// TODO(bram): Don't run the fk checks at the start of the delete, but gather
> 	// the checks and add them back into the batch checker at the end.
> 	// TODO(bram): Perhaps we should limit this depth to avoid panics.
> ```

ooc what depth does this TODO refer to? my understanding is that the Go stack can't overflow (within practical limits)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-Kz1hQOfjH1Nc0U3sLCc:b-cs106e)*

BramGruneir:
*[pkg/sql/logictest/testdata/logic_test/cascade, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1QJQRabw-OAvDXk3A:-Kz5KYlhwfl6bc3kEI7n:b9hflde) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/logictest/testdata/logic_test/cascade#L1)):*

I want to make the perhaps bold claim that our logic tests are insufficient for testing the kinds of structures we want to be able to test for this feature. It feels to me like what would really make this easier to test would be some kind of DSL that would allow you to easily describe the kinds of directed graph structures that you're writing here? Obviously these tests are very valuable but it feels like more effort than it should be to parse the relationships between the various tables, and the logic tests don't have any kind of infrastructure for the kind of abstraction you'd want for that.
</blockquote></details>

I agree that it would be useful.  But I'm worried about spending too much time on it beforehand.  It's something that I think could be added in after getting most of the basic functionality in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

BramGruneir:
Oh, good find.  I'll mine that for some good test cases.

---

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 8 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1QwCAafQG7eV5UwME:-Kz58ls_xrXcbpMLETti:b-xarq17) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/logictest/testdata/logic_test/cascade#L8)):*

Would it also be beneficial to have a test for the interaction between a cascaded delete that would be blocked by an separate `ON DELETE RESTRICT` (and any of the other relevant edge cases you listed in the RFC)?
</blockquote></details>

Indeed.  I've added one to ensure it does fail correctly.  There will plenty more to add once updates are added as well.

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 207 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz-hrKWJgGtIldgD21J:-Kz590UVZ48AgAs0vgq3:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/logictest/testdata/logic_test/cascade#L207)):*

nit: missing trailing newline
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/fk.go, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1S5WrYdnI-YgExrsn:-Kz5925TvvUmqwiH58yF:b1vkrs) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L77)):*

this is a cool Go pattern for a queue, I hadn't seen it
</blockquote></details>

I tried to use channels at first... but without an unlimited size buffer, it's just not worth it.

---

*[pkg/sql/sqlbase/fk.go, line 380 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1WDSkaEqIPPW_kLzK:-Kz599PM8HphumVKTjdz:bnnxavr) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L380)):*

any reason not to check this error?
</blockquote></details>

laziness at the time, fixed

---

*[pkg/sql/sqlbase/fk.go, line 609 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1Y3puOcY6AVVJpUpK:-Kz5AG7p3nju6w6o14y6:b2buyml) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L609)):*

I know the other helpers don't, but I feel like these helper structs would benefit from a comment describing their relationship to the operations they're helping with - if I understand correctly, there's one `cascadeHelper` per index being cascaded upon?
</blockquote></details>

You got it exactly right.  One cascadeHelper for index being cascaded.  Added.

---

*[pkg/sql/sqlbase/fk.go, line 612 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1YcWUJfMoJLYm7U_f:-Kz5BqIgE6P3t058eWPN:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L612)):*

s/entires/entries/
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/fk.go, line 725 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1ZpsT0r8BJ6fDElsO:-Kz5BwC4MCi9M2Z0xOJg:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L725)):*

`s/extra/extract/` ?
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/fk.go, line 844 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1iwJc_e_1M-7Ve8XL:-Kz5CCQvmnDNFmHvGbO_:b-b47b70) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L844)):*

`Finally,`? unless the command is meant to sound exasperated :D
</blockquote></details>

Fixed. :)  And it was kind of exasperated.  

---

*[pkg/sql/sqlbase/fk.go, line 845 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1aDubs6VfXM2LxZr9:-Kz5CJrpjIe2dRntoaxo:bd7f8qc) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L845)):*

would this happen if a row fk-referenced another row more than once?
</blockquote></details>

I've been trying to concoct a situation where this could happen.  What I want is to get to a point where there are two rows to delete, and by cascading the delete of the first one, it deletes the 2nd row preemptively.  I can't seem to figure out how to do that, without us having deferrable.  And I just don't think it's possible.

---

*[pkg/sql/sqlbase/fk.go, line 848 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1a5hFbIQDBTQHNpic:-Kz5CeO11fQUJPOPjLFG:b-bli62c) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L848)):*

ooc what depth does this TODO refer to? my understanding is that the Go stack can't overflow (within practical limits)?
</blockquote></details>

I think we should limit it artificially.  I can create a 1M row table that self-references itself where each row references the next. That means 1M rows have to be cascaded for a single delete at the top.  This todo is to add some limit later on and to perform testing to see where that should be.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-Kz5EgF-M8iqRrRWM7Hf:b-hvxkan)*

BramGruneir:
PTAL!

Ok, this has been completely rewritten.  Instead of a recursive solution this is now iterative.

DeleteRow (and all cascaded deletes) will preemptively delete rows and then check those deleted rows at the end.  This will add more delete intents on failure, but the tradeoff is that this will now be able to handle very large chains of references.

This also now allows the deletions of dependency cycles and some tests were added around that.  

This was done by adding in a new struct, called a cascader, which coordinates all cascading actions and fk checks and removing the fkCascadeHelper.

---

Review status: 3 of 12 files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-KzUUdBoKZW5j7k8poPK:b9i9wc9)*

BramGruneir:
Things left to do before this is production ready:
- Update the table lookups to not be just a transitive closure.
- Remove debug log lines.

---

Review status: 3 of 12 files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-KzUWaCYFY7wGRZxdMuu:b3xhna3)*

justinj:
I'll do another more in-depth look when I get a chance, just wanted to mention the thing from our conversation today before I forgot!

---

Review status: 3 of 12 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/cascader.go, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-KzWaqnRqFBXQ2GS39dZ:-KzWaqnRqFBXQ2GS39d_:b-gjq8io) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L1)):*
> ```Go
> // Copyright 2015 The Cockroach Authors.
> ```

nit: `s/2015/2017/`

---

*[pkg/sql/sqlbase/cascader.go, line 28 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-KzWb-l2gNobg-hdgxg1:-KzWb-l2gNobg-hdgxg2:b-sc6og0) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L28)):*
> ```Go
> )
> 
> // cascader is used to handle all referential integrity cascading actions.
> ```

I think @nvanbenschoten had a really good idea in suggesting that something like `cascader` could operate on an interface (`cascadable` maybe?) that could abstract away the logic of messing with row fetchers and table descriptors and the like to allow for easily testing more perverse cases and would also make it clearer where exactly the parts of this code that are the algorithm live, I think that's definitely something to consider! I don't know...exactly what that abstraction would look like... but I think it has the potential to decouple a lot here (though I'm not sure if there are any perf considerations)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-KzWeJNcSZLtKtX0AlA7:b-mssux1)*

dt:
Reviewed 3 of 11 files at r1, 7 of 9 files at r2.
Review status: 10 of 12 files reviewed at latest revision, 17 unresolved discussions, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_7QJ_tiVzNlC1VH4u:-Kz_7QJ_tiVzNlC1VH4v:bd0q8c8) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/logictest/testdata/logic_test/cascade#L213)):*
> ```
> CREATE TABLE self (
>   id INT PRIMARY KEY
>  ,other_id INT REFERENCES self ON DELETE CASCADE
> ```

do we have coverage of an index that references itself, e.g. `id INT PRIMARY KEY, FOREIGN KEY id REFERENCES self (id)` ?

---

*[pkg/sql/sqlbase/cascader.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_71mlrjz6l0WXjAkb:-Kz_71mlrjz6l0WXjAkc:b-sphfru) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L62)):*
> ```Go
> 	// when creating the index span.
> 	indexColIDs := make(map[ColumnID]int, len(index.ColumnIDs))
> 	for i, colID := range index.ColumnIDs {
> ```

Probably want make sure that some of the test tables above have
- multi-column FKs
- indexes that contain a strict superset of the columns in the FK

---

*[pkg/sql/sqlbase/cascader.go, line 407 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_83TV8s4Lw3LSzJEe:-Kz_83TV8s4Lw3LSzJEf:b-z5kv6y) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L407)):*
> ```Go
> 				}
> 			}
> 		}
> ```

probably want to check the ctx's Done periodically in the loop somewhere (though every time might be excessive).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

vivekmenezes:
This PR is waiting on the RFC being merged. Let's get that reviewed and merged first. Thanks!

BramGruneir:
*[pkg/sql/logictest/testdata/logic_test/cascade, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_7QJ_tiVzNlC1VH4u:-L-7Z9nB_w5CY-mQZ7ys:belb4rw) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/logictest/testdata/logic_test/cascade#L213)):*

do we have coverage of an index that references itself, e.g. `id INT PRIMARY KEY, FOREIGN KEY id REFERENCES self (id)` ?
</blockquote></details>

Why would one ever do this?  Right now, this fails when you try to insert a value.

```sql

### Self Reference to Itself Delete Cascade
# self <- self

statement ok
CREATE TABLE self (
  id INT PRIMARY KEY
 ,FOREIGN KEY (id) REFERENCES self ON DELETE CASCADE
);

statement ok
INSERT INTO self VALUES (1);
```

With the error:
`pq: foreign key violation: value [1] not found in self@primary [id]`

But what would be the use case for this?  

However, this does however work on postgres.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

BramGruneir:
Ok, lots of good progress on this.  

Only a fews things left to add before this is finished:
* Fixing up the table lookup code to use the algorithm in RFC. (and add some tests for it)
* Adding the memory manager to ensure we will fail well if we start to run out of memory.

And then finally, benchmarking our updates and deletes to see the full impact of this change.

---

Review status: 4 of 12 files reviewed at latest revision, 17 unresolved discussions.

---

*[pkg/sql/sqlbase/cascader.go, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-KzWaqnRqFBXQ2GS39dZ:-L-7UszHfNBk3Kjt1Qy0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L1)):*

nit: `s/2015/2017/`
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/cascader.go, line 28 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-KzWb-l2gNobg-hdgxg1:-L-DdzKZPGX-sI7N3BB2:b-c9swzt) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L28)):*

I think @nvanbenschoten had a really good idea in suggesting that something like `cascader` could operate on an interface (`cascadable` maybe?) that could abstract away the logic of messing with row fetchers and table descriptors and the like to allow for easily testing more perverse cases and would also make it clearer where exactly the parts of this code that are the algorithm live, I think that's definitely something to consider! I don't know...exactly what that abstraction would look like... but I think it has the potential to decouple a lot here (though I'm not sure if there are any perf considerations)
</blockquote></details>

I will be doing just this, but not in the PR.  The next PR will be all about adding more in depth testing. 

---

*[pkg/sql/sqlbase/cascader.go, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_71mlrjz6l0WXjAkb:-L-DdqG0FI4kWG21j0X-:b6c97is) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L62)):*

Probably want make sure that some of the test tables above have
- multi-column FKs
- indexes that contain a strict superset of the columns in the FK
</blockquote></details>

Done.  That second one sent me on a fairly long chase to get this working right.  Thanks for the suggestion.

---

*[pkg/sql/sqlbase/cascader.go, line 407 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_83TV8s4Lw3LSzJEe:-L-7W97kDHdq_V8tWyUn:biek8e6) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/sqlbase/cascader.go#L407)):*

probably want to check the ctx's Done periodically in the loop somewhere (though every time might be excessive).
</blockquote></details>

Sure, I added the check at the start of each dequeue. 

---

*[pkg/sql/sqlbase/fk.go, line 794 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz1_H0jDhHYYgvIq1Ol:-L-DdwsD4T8lOOj_WzPX:b-ng5w3u) ([raw file](https://github.com/cockroachdb/cockroach/blob/1c6a6e632cb8c34e53d21ebb1e24b183d067c120/pkg/sql/sqlbase/fk.go#L794)):*

should this logging stick around?

(oh, i see now you mentioned this in the commit!)
</blockquote></details>

Logging has been removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-L-De8_6T35MxAl_iMK8:be2bwjy)*

BramGruneir:
This is ready to go.  I'm going to take the WIP off.

There are two follow up PRs that are in the works.  The first is going to add memory monitoring and the second will add more testing and benchmarking.

---

Review status: 2 of 13 files reviewed at latest revision, 17 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-L-gmgqVB0e1cY-kofOj:b-1lggu6)*

pkg/sql/planner.go
```diff
@@ -343,19 +343,17 @@ func (p *planner) exec(ctx context.Context, sql string, args ...interface{}) (in
 	return countRowsAffected(params, plan)
 }
 
-func (p *planner) fillFKTableMap(ctx context.Context, m sqlbase.TableLookupsByID) error {
-	for tableID := range m {
-		table, err := p.session.tables.getTableVersionByID(ctx, p.txn, tableID)
+func (p *planner) lookupFKTable(
+	ctx context.Context, tableID sqlbase.ID,
```
dt: somewhere (maybe here?) do we also need to check that the user has the appropriate privileges on the other tables? probably also want a logic test that switches users to observe that check?

BramGruneir:
Review status: 0 of 13 files reviewed at latest revision, 16 unresolved discussions.

---

*[pkg/sql/planner.go, line 347 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-KywUKWcOBe955pv7hUU-r3-347:-L-loBUQjxqmTQU8h9JX:b-jmvatd) ([raw file](https://github.com/cockroachdb/cockroach/blob/87122e63243cc7f65f0c9ff7b5314035413778c5/pkg/sql/planner.go#L347)):*

somewhere (maybe here?) do we also need to check that the user has the appropriate privileges on the other tables? probably also want a logic test that switches users to observe that check?
</blockquote></details>

Done, required a bit of reworking and didn't fit in right here. You can see the change in fk.go.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

dt:
:lgtm:

---

Reviewed 9 of 12 files at r3, 6 of 9 files at r4.
Review status: 10 of 13 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/cascade, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-Kz_7QJ_tiVzNlC1VH4u:-L-DbvSzNaqFb75YFXJg:bn1n3uh) ([raw file](https://github.com/cockroachdb/cockroach/blob/5417264e6eb3c0d827f9aee9d8624cb6c39e03bd/pkg/sql/logictest/testdata/logic_test/cascade#L213)):*

Why would one ever do this?  Right now, this fails when you try to insert a value.

```sql

### Self Reference to Itself Delete Cascade
# self <- self

statement ok
CREATE TABLE self (
  id INT PRIMARY KEY
 ,FOREIGN KEY (id) REFERENCES self ON DELETE CASCADE
);

statement ok
INSERT INTO self VALUES (1);
```

With the error:
`pq: foreign key violation: value [1] not found in self@primary [id]`

But what would be the use case for this?  

However, this does however work on postgres.
</blockquote></details>

Ah, yeah, good point. I was thinking it made sense at least without a NOT NULL, but I think I was actually picturing a different index on the same table, not the same index. Carry on.

---

*[pkg/sql/sqlbase/cascader.go, line 113 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L05MMo7r7OMTZ-ZmO5L:-L05MMo7r7OMTZ-ZmO5M:b-ep8hq6) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/cascader.go#L113)):*
> ```Go
> 
> // spanForPKValues creates a span against the primary index of a table and is
> // used to fetch rows for cascading. This requires that the row deleter has
> ```

nit: not sure what row deleter this is referring to? seems like this is just a generic key encoding helper?

---

*[pkg/sql/sqlbase/fk.go, line 96 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0AWpukjTOlofNh3dv2:-L0AWpukjTOlofNh3dv3:b4hzucw) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/fk.go#L96)):*
> ```Go
> 		return TableLookup{}, err
> 	}
> 	if !tableLookup.IsAdding && tableLookup.Table != nil {
> ```

So I'm trying to figure out if the existing FK code got this wrong as far as skipping tables in Adding state: the reason for the state is so we start checking _everywhere_ before we start writings _anywhere_, since cached table descs mean these changes aren't transactional.

So it seems, but I might have forgotten something here that @vivekmenezes remembers, that we _should_ be checking the adding tables?

Given that it is an existing issue if it is one, I'm also fine pushing this to a follow-up.

---

*[pkg/sql/sqlbase/fk.go, line 126 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0AWjb2NjpOk32IYTSQ:-L0AWjb2NjpOk32IYTSR:bp8qb5g) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/fk.go#L126)):*
> ```Go
> 	q.alreadyChecked[tableID][usage] = struct{}{}
> 	// If the table is being added, there's no need to check it.
> 	if tableLookup.IsAdding {
> ```

So I've 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-L0AWgLSy7SR84XXLuL6:bnfp4nl)*

jordanlewis:
Planning on taking a look at this today!

BramGruneir:
Review status: 7 of 13 files reviewed at latest revision, 18 unresolved discussions.

---

*[pkg/sql/sqlbase/cascader.go, line 113 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L05MMo7r7OMTZ-ZmO5L:-L0AteON7jZCSn4wh3DE:b-f3becg) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/cascader.go#L113)):*

nit: not sure what row deleter this is referring to? seems like this is just a generic key encoding helper?
</blockquote></details>

Oh yeah, that's an old rotted comment from back when the code was in the fk.go.  Removed.

---

*[pkg/sql/sqlbase/fk.go, line 96 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0AWpukjTOlofNh3dv2:-L0AtunOTUTayAXogMzf:bovyah2) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/fk.go#L96)):*

So I'm trying to figure out if the existing FK code got this wrong as far as skipping tables in Adding state: the reason for the state is so we start checking _everywhere_ before we start writings _anywhere_, since cached table descs mean these changes aren't transactional.

So it seems, but I might have forgotten something here that @vivekmenezes remembers, that we _should_ be checking the adding tables?

Given that it is an existing issue if it is one, I'm also fine pushing this to a follow-up.
</blockquote></details>

I'll discuss this with @vivekmenezes offline.  And if this is an issue, I'll send a fix.

---

*[pkg/sql/sqlbase/fk.go, line 126 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0AWjb2NjpOk32IYTSQ:-L0Au235tUDaFahYuZKH:b155j) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/fk.go#L126)):*

So I've 
</blockquote></details>

Same here. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

vivekmenezes:
Review status: 7 of 13 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/fk.go, line 96 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0AWpukjTOlofNh3dv2:-L0CphqxVEF_u8X3HcmA:bgadytx) ([raw file](https://github.com/cockroachdb/cockroach/blob/7db2acfe3d073714ba716bc7bae8a1e48a400bb4/pkg/sql/sqlbase/fk.go#L96)):*

I'll discuss this with @vivekmenezes offline.  And if this is an issue, I'll send a fix.
</blockquote></details>

So I checked and it looks like the table lease code will return errTableAdding when it encounters a table in the Adding state. I'm surprised that any DML state would not simply return that error while executing. The error can only be ignored if another schema change is being applied to the table (we allow schema changes on tables in the ADDING state

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

jordanlewis:
Nice work! I like the approach you've taken.

---

Reviewed 2 of 12 files at r3, 7 of 9 files at r4, 4 of 4 files at r5.
Review status: all files reviewed at latest revision, 26 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/cascader.go, line 69 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0B9sZ1xVs4kAoztFG2:-L0B9sZ1xVs4kAoztFG3:buoyqkb) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L69)):*
> ```Go
> 	}
> 	key := roachpb.Key(keyBytes)
> 	return roachpb.Span{Key: key, EndKey: key.PrefixEnd()}, nil
> ```

This is incorrect in the face of interleaved indexes. It will produce too-wide spans that include the child interleaves. You can append an interleave sentinel to the end key directly to fix this problem instead of calling PrefixEnd.

---

*[pkg/sql/sqlbase/cascader.go, line 129 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BCd_YDk74D3w_AsVh:-L0BCd_YDk74D3w_AsVi:b-ym3jp) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L129)):*
> ```Go
> 	}
> 	key := roachpb.Key(keyBytes)
> 	return roachpb.Span{Key: key, EndKey: key.PrefixEnd()}, nil
> ```

Same comment as above, this span is too wide unfortunately.
 
We should make a method that takes a key and returns the precise primary key span for that key. It is used in several places, here and fk.go as well. The implementation is just `func(key roachpb.Key) {keyCopy := key, return roachpb.Span{Key:key, EndKey: encoding.EncodeInterleaveSentinel(keyCopy)}`.

See the commit on this branch: https://github.com/jordanlewis/cockroach/tree/fix-interleave-fk

I was going to submit that after I got to fixing the interleave primary key scan issue, but @richardwu ended up working on that instead.

Now that I think about it, I would be okay with doing this work along with the corresponding work in `fk.go` after this PR is merged.

---

*[pkg/sql/sqlbase/cascader.go, line 266 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BG0QS2U_IPWE9zf-D:-L0BG0QS2U_IPWE9zf-E:b77daig) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L266)):*
> ```Go
> 	if traceKV {
> 		log.VEventf(ctx, 2,
> 			"cascading delete from redIndex:%s, into table:%s, using index:%s for values:%+v",
> ```

nit: s/redIndex/refIndex/

---

*[pkg/sql/sqlbase/cascader.go, line 337 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0F7fPrPL6y_GcwWLzt:-L0F7fPs_y_Npz74YBSh:b-4u9rco) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L337)):*
> ```Go
> 	// TODO(Bram): use a row container here too for rowsToDelete.
> 	var rowsToDelete []tree.Datums
> 	for _, resp := range pkResp.Responses {
> ```

This code seems very similar to the code that fetches the PKs for deletion. Is there a way to DRY these two stanzas?

---

*[pkg/sql/sqlbase/cascader.go, line 467 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BHdyv8QlaSUAawUB9:-L0BHdyv8QlaSUAawUBA:b-cj9msy) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L467)):*
> ```Go
> 		rowDeleter, exists := c.rowDeleters[tableID]
> 		if !exists {
> 			return errors.Errorf("Could not find row deleter for table %d", tableID)
> ```

Is this an expected condition, or an internal error? It seems like an internal error. Is there any more detail you could add here to improve debuggability?

Errors in SQL land are generally expected to include a pgerror code if they're user facing, FYI, but ones without a pgerror code will get automatically annotated with Internal Error.

---

*[pkg/sql/sqlbase/fk.go, line 175 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BLfStrUA8NJpYD3yo:-L0BLfStrUA8NJpYD3yp:b-qp9802) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L175)):*
> ```Go
> 		return nil, err
> 	}
> 	for {
> ```

take-it-or-leave-it: couldn't this be `for ; exists; tableLookup, curUsage, exists = queue.dequeue()` if you  predeclare the 3 loop vars?

---

*[pkg/sql/sqlbase/fk.go, line 199 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0Blw5oJOS2W3TjREf7:-L0Blw5oJOS2W3TjREf8:b-nis9qh) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L199)):*
> ```Go
> 						return nil, err
> 					}
> 					if referencedTableLookup.IsAdding || referencedTableLookup.Table == nil {
> ```

When would referencedTableLookup.Table be nil? Is this just defensive programming? 

---

*[pkg/sql/sqlbase/fk.go, line 217 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0F5kvoj42fzvK1i9E2:-L0F5kvoj42fzvK1i9E3:b-imkg8f) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L217)):*
> ```Go
> 					}
> 				}
> 			case CheckUpdates:
> ```

This code block almost exactly duplicates the code block above  (exception is the check mode enqueued for the referenced tables).

Could you please DRY? The domain logic here is tricky, and I'd like to make sure that future programmers won't accidentally change just one of the blocks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-L0F8BPwiNNhEr-n495y:b-obryyd)*

vivekmenezes:
Review status: all files reviewed at latest revision, 26 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/fk.go, line 199 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0Blw5oJOS2W3TjREf7:-L0FLsX5SlERJkJvFDRO:b-yho2tv) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L199)):*

When would referencedTableLookup.Table be nil? Is this just defensive programming? 
</blockquote></details>

This seems bad. As I mentioned elsewhere the table lease code will not return a table that isAdding and if a nil table is return an error is returned. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

richardwu:
Review status: all files reviewed at latest revision, 26 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/cascader.go, line 129 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BCd_YDk74D3w_AsVh:-L0FV2TGD21XrnKYevZ_:br4w10e) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L129)):*

Same comment as above, this span is too wide unfortunately.
 
We should make a method that takes a key and returns the precise primary key span for that key. It is used in several places, here and fk.go as well. The implementation is just `func(key roachpb.Key) {keyCopy := key, return roachpb.Span{Key:key, EndKey: encoding.EncodeInterleaveSentinel(keyCopy)}`.

See the commit on this branch: https://github.com/jordanlewis/cockroach/tree/fix-interleave-fk

I was going to submit that after I got to fixing the interleave primary key scan issue, but @richardwu ended up working on that instead.

Now that I think about it, I would be okay with doing this work along with the corresponding work in `fk.go` after this PR is merged.
</blockquote></details>

You could also dry this method up with
```
return spanForIndexValues(table, &table.PrimaryIndex, len(table.PrimaryIndex.ColumnIDs), fetchColIDtoRowIndex, values, MakeIndexKeyPrefix(table, table.PrimaryIndex.ID))
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

BramGruneir:
TFTR!  All comments addressed and I added some new interleaved table tests.  Please give this one quick final pass!

---

Review status: 10 of 13 files reviewed at latest revision, 26 unresolved discussions.

---

*[pkg/sql/sqlbase/cascader.go, line 69 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0B9sZ1xVs4kAoztFG2:-L0G5irXVHXsuhhlf4OG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L69)):*

This is incorrect in the face of interleaved indexes. It will produce too-wide spans that include the child interleaves. You can append an interleave sentinel to the end key directly to fix this problem instead of calling PrefixEnd.
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/cascader.go, line 129 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BCd_YDk74D3w_AsVh:-L0G7oRAc98owqfzi5Tx:b7uoblz) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L129)):*

You could also dry this method up with
```
return spanForIndexValues(table, &table.PrimaryIndex, len(table.PrimaryIndex.ColumnIDs), fetchColIDtoRowIndex, values, MakeIndexKeyPrefix(table, table.PrimaryIndex.ID))
```
</blockquote></details>

Worked with @richardwu on this.  I've also added new interleaved table tests.

---

*[pkg/sql/sqlbase/cascader.go, line 266 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BG0QS2U_IPWE9zf-D:-L0Fhsnz4H_5WKeJHNfG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L266)):*

nit: s/redIndex/refIndex/
</blockquote></details>

Done.

---

*[pkg/sql/sqlbase/cascader.go, line 337 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0F7fPrPL6y_GcwWLzt:-L0FhyObqLedMrko8VNc:bqr0nzh) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L337)):*

This code seems very similar to the code that fetches the PKs for deletion. Is there a way to DRY these two stanzas?
</blockquote></details>

I'm going to leave this as is.  I've got a follow up PR that re-writes all of this using rowContainers and they diverge more at that point.

---

*[pkg/sql/sqlbase/cascader.go, line 467 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BHdyv8QlaSUAawUB9:-L0FiC4SsAWjBBEOUu7t:bs27uxd) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L467)):*

Is this an expected condition, or an internal error? It seems like an internal error. Is there any more detail you could add here to improve debuggability?

Errors in SQL land are generally expected to include a pgerror code if they're user facing, FYI, but ones without a pgerror code will get automatically annotated with Internal Error.
</blockquote></details>

This is never an expected condition.  It means something went horribly wrong and is more just a safety check.  What do you suggest in this situation? 

---

*[pkg/sql/sqlbase/fk.go, line 175 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BLfStrUA8NJpYD3yo:-L0FXEaS5jyfyWKHjMJj:b-85gorf) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L175)):*

take-it-or-leave-it: couldn't this be `for ; exists; tableLookup, curUsage, exists = queue.dequeue()` if you  predeclare the 3 loop vars?
</blockquote></details>

So this is a little ugly.  I have to predeclare exists to true, and then I still need to return the result at the end.

```
exists := true
var tableLookup TableLookup
var curUsage FKCheck
for ; exists; tableLookup, curUsage, exists = queue.dequeue() {
```

Or I could also dequeue as the init in the for loop, this keeps the scope, but it looks even worse.

```
for tableLookup, curUsage, exists := queue.dequeue(); exists; tableLookup, curUsage, exists = queue.dequeue() {
```

I don't feel strongly about this, but I think I'll leave it as for readability.  If you prefer another style, let me know.

---

*[pkg/sql/sqlbase/fk.go, line 199 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0Blw5oJOS2W3TjREf7:-L0FXwQu58CzhELW-NiJ:bxojqyc) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L199)):*

This seems bad. As I mentioned elsewhere the table lease code will not return a table that isAdding and if a nil table is return an error is returned. 
</blockquote></details>

So referencedTableLookup can be nil because the lookup function doesn't always actually perform a lookup.  

In columnBackfiller's runChunk and in SchemaChanger'sdistBackfill, the lookups do not need to walk the full relationship graph and only require the immediate relationships.  So I've left those as is and they use the NoLookup function which returns a nil for the table descriptor.

In both those cases, they call fetch the table descriptors after the call to TablesNeededForFK.
I've added some comments about this.

That being said, when I remove this specific check, it does not cause any errors or panics in our tests.

---

*[pkg/sql/sqlbase/fk.go, line 217 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0F5kvoj42fzvK1i9E2:-L0FeyriqYRobgY1uC4f:b-4wtytt) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/fk.go#L217)):*

This code block almost exactly duplicates the code block above  (exception is the check mode enqueued for the referenced tables).

Could you please DRY? The domain logic here is tricky, and I'd like to make sure that future programmers won't accidentally change just one of the blocks.
</blockquote></details>

Done.  It's ugly no matter how I try to formulate it but I tried to keep it as clear as possible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-L0GRAcVh8ImHrIk_NEQ:b-1888n0)*

jordanlewis:
:lgtm: 

---

Reviewed 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/cascader.go, line 337 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0F7fPrPL6y_GcwWLzt:-L0GSOu2qh6rH4JsRj7d:b1z3zbe) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L337)):*

I'm going to leave this as is.  I've got a follow up PR that re-writes all of this using rowContainers and they diverge more at that point.
</blockquote></details>

OK, sounds good.

---

*[pkg/sql/sqlbase/cascader.go, line 467 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BHdyv8QlaSUAawUB9:-L0GSU7pmREZ-vKo3EsF:b-rwc8xp) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L467)):*

This is never an expected condition.  It means something went horribly wrong and is more just a safety check.  What do you suggest in this situation? 
</blockquote></details>

Perhaps prefix this error message with "programming error:" then?

We can do this in a separate PR if you want.

I'd really like to start a separate initiative that 1) consistently applies some kind of "programming error:" or "internal error:" prefix to unexpected errors, and 2) replaces all SQL panics with internal errors. Unlike the core code, if we encounter an unexpected situation it's okay not to panic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-:-L0GV1XRiX59MWxYgQEH:bg7igkv)*

BramGruneir:
Review status: 12 of 13 files reviewed at latest revision, 22 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/cascader.go, line 467 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BHdyv8QlaSUAawUB9:-L0GYT7MoZ04PrghRu1M:b-l7mr9x) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L467)):*

Perhaps prefix this error message with "programming error:" then?

We can do this in a separate PR if you want.

I'd really like to start a separate initiative that 1) consistently applies some kind of "programming error:" or "internal error:" prefix to unexpected errors, and 2) replaces all SQL panics with internal errors. Unlike the core code, if we encounter an unexpected situation it's okay not to panic.
</blockquote></details>

Done.  I'm 100% behind that proposal.  Perhaps we can proactively do that for all future SQL PRs and maybe spend some time during bug fixing for 2.0 to retroactively go back and fix the panics and other errors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

richardwu:
Review status: 12 of 13 files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[pkg/sql/sqlbase/cascader.go, line 129 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064#-L0BCd_YDk74D3w_AsVh:-L0GdCusdtSJCgEaNK2Y:b-rbm8m5) ([raw file](https://github.com/cockroachdb/cockroach/blob/01231052c274d826d9c933189789b63f267e5be8/pkg/sql/sqlbase/cascader.go#L129)):*

Worked with @richardwu on this.  I've also added new interleaved table tests.
</blockquote></details>

The `spanFor(Index|PK)Values` :lgtm: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20064)*

PR #20066, opened 2017-11-14
ui: Initial Commit of Cluster Visualization Work

mrtracy:
Two commits.

First Commit: Deletes old prototype cluster viz, was non functional.

Second commit:

Adds an initial version of cluster visualization, worked on by both
myself and Spencer.

This code cannot currently be accessed because it is under the CCL
license tree, but our build system does not yet link CCL code.

Currently the rewrite is not yet at parity with the original prototype.
It does not:
- Group nodes into localities
- Display numeric values for traffic between node pairs
- Show locality links for only the currently hovered node (all links are
shown).

The rewrite also uses simulated location data; nodes on the cluster are
assigned simulated long/lat pairs in order to place them on the map for
visual effect.

Note that it is likely that the rewrite will begin to diverge from the
prototype somewhat at this point, and complete parity should not be
considered a upcoming milestone.

Release Notes: none


pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx
```diff
@@ -0,0 +1,254 @@
...
+  [114.0683, 22.54554],
+  [72.88261, 19.07283],
+  [77.59369, 12.97194],
+  [77.22445, 28.63576],
+];
+
+// CurrentNodes places all current nodes on the map based on their long/lat
+// coordinates (currently simulated). Also generates a link between each node
+// pair, visualized on the map as a directed path.
+class CurrentNodes extends React.Component<CurrentNodesProps & CurrentNodesOwnProps, any> {
```
spencerkimball: Is the plan to replace this with `CurrentLocalities` in order to deal with the aggregations?

pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx
```diff
@@ -0,0 +1,254 @@
...
+      link.points.push(Vector.sub(linkPair[1].translate, Vector.mult(norm, skip)));
+      link.points.push(linkPair[1].translate);
+
+      // Ensure that points sort from left to right.
+      link.points.sort(function (a, b) { return a[0] - b[0]; });
+
+      return link;
+    });
+
+    // Maximum client activity rate, used to normalize network bars across all
```
spencerkimball: Mark this code as temporary and for deletion (probably).

pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+    // top of the default scale and translation.
+    const scale = this.state.zoomTransform.scale();
+    const translate = this.state.zoomTransform.translate();
+    const projection = d3.geo.mercator();
+    projection.scale(projection.scale() * scale);
+    projection.translate(Vector.add(Vector.mult(projection.translate(), scale), translate));
+
+    return (
+      <g>
+        <WorldMap projection={projection} />
```
spencerkimball: Did the US states go bye bye?

pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright 2017 The Cockroach Authors.
```
spencerkimball: I'm surprised the link bending code didn't end up here...?

pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts
```diff
@@ -0,0 +1,47 @@
+// Copyright 2017 The Cockroach Authors.
```
spencerkimball: Do you think it might be preferable to just import a vector math library?

pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <path
+              key={i}
+              className="geopath"
+              id={`world-${feature.id}`}
+              d={pathGen(feature)}
+            />,
+          )}
+        </g>
+        <g>
+          {usStateShapes.features.map((feature: any, i: number) =>
```
spencerkimball: Ah, I see you have this here. Did I miss the code which phases it in and out?

spencerkimball (approved):

cuongdo:
@vilterp would love your thoughts here, because @couchand is busy with support duties

mrtracy:
Review status: 0 of 34 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 60 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHV-r2-60:-Kz-sTujZuDRdASttwde:b-8c8x41) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L60)):*

Is the plan to replace this with `CurrentLocalities` in order to deal with the aggregations?
</blockquote></details>

Yes, this is where that logic would live.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 213 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHV-r2-213:-Kz-uDu4jcoMpt5i2d2G:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L213)):*

Mark this code as temporary and for deletion (probably).
</blockquote></details>

Done.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 109 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHW-r2-109:-Kz-uFoKHDe6MavkWpGR:bqm0ht4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L109)):*

Did the US states go bye bye?
</blockquote></details>

No, you found them, they're just part of the world map.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHZ-r2-1:-Kz-uPZUUUIA-f7pCXmR:b-fekr6j) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts#L1)):*

I'm surprised the link bending code didn't end up here...?
</blockquote></details>

It could go here, but it would have to be abstracted a bit in order to extract a function, and I decided to leave it intact for now in the interest of getting this checked in.

In retrospect, I don't think I should have even attempted the paths at all for the initial commit, this is a thousand line review.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHe-r2-1:-Kz-v4s5Mgi52JZbDp9l:ba82cwu) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts#L1)):*

Do you think it might be preferable to just import a vector math library?
</blockquote></details>

I looked into it, but unfortunately I couldn't find an obvious stand-alone candidate. I think I *will* be able to find a good match, but I'm going to have to dig a bit deeper; I just kept your library for the time being.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx, line 38 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHg-r2-38:-Kz-wlgOnHIPNtegT_Mo:brjep1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx#L38)):*

Ah, I see you have this here. Did I miss the code which phases it in and out?
</blockquote></details>

No, that is not implemented in this version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066)*

vilterp:
Looks good; a few minor suggestions. Cool that no new data fetching code was needed.

---

Reviewed 18 of 18 files at r1, 16 of 16 files at r2, 16 of 16 files at r4.

---

*[pkg/ui/package.json, line 13 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01a6Yiz4GL98mG3IY:-Kz01a6Yiz4GL98mG3IZ:bat4jca) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/package.json#L13)):*
> ```JSON
>     "combokeys": "^2.4.6",
>     "d3-path": "^1.0.5",
>     "d3-shape": "^1.2.0",
> ```

We probably have this code in the bundle twice (modulo v3 vs v4 differences) — these are standalone libraries that are part of D3 v4; we already have their predecessors as part of D3, which we have because it's a dependency of NVD3.

Would be nice to not duplicate that somehow. Maybe @couch's graph stuff can use v4 and then we can just import these little v4 packages as we need them.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 142 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02hUiEDX64o0LDo5i:-Kz02hUiEDX64o0LDo5j:b-e6966j) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L142)):*
> ```XML
>     // Adjust locality links paths by routing them around other localities.
>     const scaledRadius = Locality.outerRadius * localityScale;
>     const localityLinks = linkPairs.map(linkPair => {
> ```

would be good to factor out this big function you're passing to `map`.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 49 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01ezFQZh3FPqUajLe:-Kz01ezFQZh3FPqUajLf:b-oija45) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L49)):*
> ```XML
>   }
> 
>   onZoom = () => {
> ```

does defining these this way instead of normal method definitions (`onZoom() { ... }`) change the `this` binding behavior or something? Otherwise, would prefer a normal method definition for consistency.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-zS-F3QNhP2NNbWJl:-Kz-zS-F3QNhP2NNbWJm:bqd0qt6) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx#L1)):*
> ```XML
> // Copyright 2017 The Cockroach Authors.
> ```

nit: capitalize the second L in the file name: `localityLink.tsx`

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx, line 24 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01sMs3kRnheXKBbeR:-Kz01sMs3kRnheXKBbeS:b-lue1fs) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx#L24)):*
> ```XML
>     return (
>       <g className="locality-link-group">
>         <path className="locality-link" id={pathId} d={d3.svg.line().interpolate("cardinal-open").tension(0.5)(data.points)} />
> ```

nit: line length

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01r14mdXX6SpgTdjD:-Kz01r14mdXX6SpgTdjE:bptjv3r) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx#L32)):*
> ```XML
>           >
>             in
>               </textPath>
> ```

nit: outdent the closing tag to match opening tag

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHZ-r2-1:-Kz02RLnBS2LmAyXh5Pk:b1ckf8o) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts#L1)):*

It could go here, but it would have to be abstracted a bit in order to extract a function, and I decided to leave it intact for now in the interest of getting this checked in.

In retrospect, I don't think I should have even attempted the paths at all for the initial commit, this is a thousand line review.
</blockquote></details>

I don't think this review is ridiculously large. Would appreciate path bending being factored out to here.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/sim.css, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-4zzjXghylLXUtkd:-Kz0-5--DCJFbEQZpJwZ:b-mfeyfc) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/sim.css#L1)):*
> ```CSS
> /* Copyright 2017 The Cockroach Authors.
> ```

I guess we'll make this into a `.styl` file with the usual namespacing stuff a little later.

In the meantime, maybe rename to `clusterViz.css`

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 15 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02571SLHu_zUqGO2q:-Kz02571SLHu_zUqGO2r:b-frg2bj) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L15)):*
> ```XML
> 
> export default class NodeStatusHistory {
>     clientActivityRate: number;
> ```

Please add a comment explaining what the "activity rate" is. It looks like QPS over the last couple of updates (?)

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/typings.d.ts, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-wV_vnx8TJ23xFt7i:-Kz-wV_vnx8TJ23xFt7j:bylg2h2) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/typings.d.ts#L9)):*
> ```XML
> //     https://github.com/cockroachdb/cockroach/blob/master/LICENSE
> 
> declare module "*.json" {
> ```

I see that this is for the world and US States JSON files. You could declare them as JS files, like

```
const states = {
  // ...geojson...
}

export default states;
```

And avoid this funny business with typings.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-KywZWHhWd__7dFGWdHe-r2-1:-Kz-xDCbD-hQcumIEA_J:b86ak8u) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts#L1)):*

Do you think it might be preferable to just import a vector math library?
</blockquote></details>

https://github.com/aszecsei/ts-vector-math/blob/master/src/vec2.ts looks decent. Don't have a strong opinion.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx, line 26 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02KESFwaTzloafhVA:-Kz02KESFwaTzloafhVB:bva4jhl) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx#L26)):*
> ```XML
>     return (
>       <g>
>         <rect className="projection" />
> ```

A comment that this forms the lavender background would be nice (had to look at the CSS to figure out what it was doing). Or maybe `className="background"` or something

---

*[pkg/ui/src/index.tsx, line 47 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-tRDHonY3-jOntcLu:-Kz-tRDISfVehK0mmDlK:b-eccf0b) ([raw file](https://github.com/cockroachdb/cockroach/blob/8faf12e2990f20f55c7a2e992c8b30f084021abd/pkg/ui/src/index.tsx#L47)):*
> ```XML
> import RaftRanges from "src/views/devtools/containers/raftRanges";
> import RaftMessages from "src/views/devtools/containers/raftMessages";
> import ClusterViz from "src/views/devtools/containers/clusterViz";
> ```

looks like the file being imported here is still around.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-:-Kz02qjQniPiKigiEUqp:b-qovlda)*

couchand:
Got a break in the support so I started to take a look at this.  I left a variety of comments, but didn't get through everything.  I think this has a way to go personally, as is it doesn't seem like it's all that approachable.  Changes got pushed while I was looking, I'll let it bake for a bit and come back through later to give a more thorough review.

---

Reviewed 3 of 16 files at r2.
Review status: all files reviewed at latest revision, 37 unresolved discussions, some commit checks failed.

---

*[pkg/ui/package.json, line 13 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-x8k2coBdjTCPcSfZ:-Kz-x8k2coBdjTCPcSf_:b1e36p4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/package.json#L13)):*
> ```JSON
>     "combokeys": "^2.4.6",
>     "d3-path": "^1.0.5",
>     "d3-shape": "^1.2.0",
> ```

It's a bit awkward to depend on d3 v3 and v4 at the same time, but it's probably fine.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 7 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-xRqwCz4G42cqpx1j:-Kz-xRqxQyYy9TONAjTO:b-vdacnw) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L7)):*
> ```XML
> // You may obtain a copy of the License at
> //
> //     https://github.com/cockroachdb/cockroach/blob/master/LICENSE
> ```

maybe update these to point to https://github.com/cockroachdb/cockroach/blob/master/licenses/CCL.txt ?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-Znnm-ED7GShBscy:-Kz0-ZnoalOdRr_RuXTp:b-ze74wv) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L90)):*
> ```XML
> 
>   componentWillReceiveProps(props: CurrentNodesProps & CurrentNodesOwnProps) {
>     props.refreshNodes();
> ```

Regardless of the props changes?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 95 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-euMXKSveB3mP9pC:-Kz0-euMXKSveB3mP9pD:b-12a999) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L95)):*
> ```XML
> 
>   render() {
>     this.accumulateHistory();
> ```

This seems like an odd time to do that...

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 164 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz044eVWnLP0IknFmYM:-Kz044eVWnLP0IknFmYN:b2w4fbu) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L164)):*
> ```XML
>       // the line drawn to represent this locality link. This inner
>       // loop just adds additional points to the cardinal curve.
>       for (let j = 0; j < localities.length; j++) {
> ```

The math in here is awfully hairy.  Can we extract it so that it can be sufficiently tested?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 221 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz04To2AdOeb_jlLskm:-Kz04To2AdOeb_jlLskn:bnug67i) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L221)):*
> ```XML
>       <g>
>         {localities.map((locality) =>
>           <g
> ```

It might be nice to push this `g` into the `Locality` for readability.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 27 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01xiRB0snzzO_jJ5X:-Kz01xiRB0snzzO_jJ5Y:bb76mu4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L27)):*
> ```XML
>   graphEl: SVGElement;
>   zoom: d3.behavior.Zoom<any>;
>   maxLatitude = 63;
> ```

Why?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 72 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01odsaHNMRl6x1jmu:-Kz01odsaHNMRl6x1jmv:b3zhvoq) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L72)):*
> ```XML
> 
>     // Compute zoomable area bounds based on the default mercator projection.
>     const projection = d3.geo.mercator();
> ```

I'll respectfully disagree with this choice of projection.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 105 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz03W-Oi9SOkg1M6oUW:-Kz03W-P6Q1b3X3nQ1NU:bgijeml) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L105)):*
> ```XML
>     const projection = d3.geo.mercator();
>     projection.scale(projection.scale() * scale);
>     projection.translate(Vector.add(Vector.mult(projection.translate(), scale), translate));
> ```

These seem really awkward and make me suspect that we're not using them correctly.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz04mZwjyctfek8FBNP:-Kz04mZwjyctfek8FBNQ:bay1nla) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L24)):*
> ```XML
> export class Locality extends React.Component<LocalityProps, any> {
>   static radius = 42;
>   static arcWidth = Locality.radius * 0.11111;
> ```

Why on both of these magic constants?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 33 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz056YksxjnXrBhVQ0k:-Kz056YksxjnXrBhVQ0l:b7gbu0) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L33)):*
> ```XML
>         className="capacity-background"
>         d={PathMath.createArcPath(
>           Locality.radius, Locality.outerRadius, PathMath.arcAngleFromPct(0), PathMath.arcAngleFromPct(1),
> ```

I think this would be simpler as just a circle with a stroke of the appropriate width?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 89 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz06xd6fq1ooK9Kx4vJ:-Kz06xd6fq1ooK9Kx4vK:biso5z5) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L89)):*
> ```XML
> 
>   renderNetworkActivity() {
>     const barsX = Locality.radius * Math.cos(PathMath.angleFromPct(0));
> ```

This seems excessively complicated for the purpose...

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz073XN93YyoAYslKw5:-Kz073XN93YyoAYslKw6:b-n4mjc9) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L101)):*
> ```XML
>           x1={Locality.outerRadius - 2}
>           y1={-labelH}
>           x2={Math.round(Locality.outerRadius - barsWidth * nodeHistory.clientActivityRate / maxClientActivityRate)}
> ```

Seems like using `d3-scale` would simplify this.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts, line 14 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz05wH3vREI50lQOLMe:-Kz05wH3vREI50lQOLMf:b81bwft) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts#L14)):*
> ```XML
> // createArcPath returns an svg arc object. startAngle and endAngle are
> // expressed in radians.
> export function createArcPath(innerR: number, outerR: number, startAngle: number, endAngle: number) {
> ```

I have to say I don't think this abstraction helps, I think it gets in the way.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz06_TFVFo7xT4CCa1U:-Kz06_TFVFo7xT4CCa1V:b-cj48p7) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts#L62)):*
> ```XML
>         " L" + c + "," + h + " A" + c + "," + c + " 0 0 1 0," + (h - c) +
>         " L0," + c + " A" + c + "," + c + " 0 0 1 " + c + ",0 Z";
> }
> ```

Seems like everything in this file would benefit from some testing.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 31 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-z14SDsc56kjVS1tS:-Kz-z14T9uo9IqUGYP7d:b8m5spc) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L31)):*
> ```XML
>             }
> 
>             this.computeClientActivityRate();
> ```

It doesn't seem like it adds anything but complexity to keep the old one around after this point.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 49 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-zyLxDX1z8nCR8V7R:-Kz-zyLxDX1z8nCR8V7S:b-g91v4p) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L49)):*
> ```XML
>             const seconds = NanoToMilli(latest.updated_at.subtract(prev.updated_at).toNumber()) / 1000;
>             const totalQPS = (latest.metrics["sql.select.count"] - prev.metrics["sql.select.count"]) +
>                 (latest.metrics["sql.distsql.select.count"] - prev.metrics["sql.distsql.select.count"]) +
> ```

I *think* this will double-count DistSQL selects, since `sql.distsql.select.count` is already included in `sql.select.count`.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-BCfZFX98C3I7tQd:-Kz0-BCg419VwHl5KrSk:b-psfqoi) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L53)):*
> ```XML
>                 (latest.metrics["sql.insert.count"] - prev.metrics["sql.insert.count"]) +
>                 (latest.metrics["sql.delete.count"] - prev.metrics["sql.delete.count"]);
>             this.clientActivityRate = totalQPS / seconds;
> ```

This looks odd, I guess it's just that `totalQPS` isn't actually per second?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts, line 10 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01HoUfvd1X9DOA3V-:-Kz01HoUfvd1X9DOA3V0:bjog47s) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts#L10)):*
> ```XML
> 
> export function distance(n1: [number, number], n2: [number, number]) {
>     return Math.sqrt((n1[0] - n2[0]) * (n1[0] - n2[0]) + (n1[1] - n2[1]) * (n1[1] - n2[1]));
> ```

maybe DRY as `return length(sub(n1, n2))`?

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/zoom.ts, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02WEnIQqvUdz39Rca:-Kz02WEnIQqvUdz39Rcb:b-8xj8ua) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/zoom.ts#L1)):*
> ```XML
> // Copyright 2017 The Cockroach Authors.
> ```

Do we really need to implement all this ourselves?  What do we need here that isn't offered directly by `d3-zoom`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-:-Kz07YGE_Vi3IiH6WJXm:bb7vzv9)*

mrtracy:
I've addressed many of the issues, and added TODOs for some suggested refactorings that are more extensive than necessary for this initial commit.

---

Review status: 17 of 34 files reviewed at latest revision, 37 unresolved discussions.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 7 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-xRqwCz4G42cqpx1j:-Kz0diEs0XaJoLaTKmA0:b-qa9vld) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L7)):*

maybe update these to point to https://github.com/cockroachdb/cockroach/blob/master/licenses/CCL.txt ?
</blockquote></details>

This is copy/pasted from our existing CCL code in the main cockroachDB project. I think keeping it identical to existing headers is the best move; if this needs to be updated, it can be updated across all files with a mechanical change later.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 90 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-Znnm-ED7GShBscy:-Kz09SsL9gog8-KI6YSV:btnxswq) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L90)):*

Regardless of the props changes?
</blockquote></details>

It's already a no-op if you don't need to refresh.  This is the pattern currently used by every other component.

Incidentally, this pattern will go away when my network refactoring actually lands.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 95 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-euMXKSveB3mP9pC:-Kz0q79ZGqm8b19-Y1Ff:b-jjhs4g) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L95)):*

This seems like an odd time to do that...
</blockquote></details>

It should be moved to lifecycle methods, but i was having trouble getting it to work correctly.  I've added a TODO, I will fix this once I can actually see the result again (after CCL build stuff is merged).

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 164 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz044eVWnLP0IknFmYM:-Kz0qfRdFQ8TJoJaRplA:b-nkjh15) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L164)):*

The math in here is awfully hairy.  Can we extract it so that it can be sufficiently tested?
</blockquote></details>

I've added a TODO to extract this to PathMath.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx, line 221 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz04To2AdOeb_jlLskm:-Kz09ldUCdEfNpJoIMQw:b-ajvxp9) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/currentNodes.tsx#L221)):*

It might be nice to push this `g` into the `Locality` for readability.
</blockquote></details>

Maybe, but it's not the right abstraction; the lower level component should just draw itself in a predictable fashion, the higher level component should position and scale it.

What's missing here is a component that makes the transforms more readable, rather than always setting these transform strings manually like this.  I promise you, such a component will be created in short order.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 27 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01xiRB0snzzO_jJ5X:-Kz0A8RbeQ7pITeuK4cC:b7tliu7) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L27)):*

Why?
</blockquote></details>

This limit was present in the original prototype (although I think it was a higher latitude); basically, in addition to political problems the mercator projection looks bizarre and takes up a lot of space at extreme latitudes (you an fill the whole screen with just the arctic).

This prevents the user from scrolling up to the useless parts of the map, which are unlikely to have a data center.  That's a pure judgement call, but we can easily remove this if desired.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 72 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01odsaHNMRl6x1jmu:-Kz0ezNgWy1TDPc6t1tt:b-mkq7o4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L72)):*

I'll respectfully disagree with this choice of projection.
</blockquote></details>

Noted, but every widely used map program on the web appears to use this projection.  I think the political aversion to this projection does not outweigh the fact that the average person is so familiar with this projection.

Perhaps we can be the change we want to see, but we can make that decision later; luckily, it's a very small change.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 105 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz03W-Oi9SOkg1M6oUW:-Kz0AUhrKfsVigJII2Im:b-q9rk74) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L105)):*

These seem really awkward and make me suspect that we're not using them correctly.
</blockquote></details>

The awkward part is that the default value of projection.scale() is 150, and the default value of projection.translate() is some weird number that I don't even know off the top of my head that's based on the assumption that the map is being centered in a 950/500 area.

An alternative to scaling the projection is to just take the zoom scale and apply it directly to a g element, but for some reason the performance of that seems to be *worse* than just rescaling the projection.

I think this is going to go through a few iterations before we get it right.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx, line 49 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01ezFQZh3FPqUajLe:-Kz0gdzrxOOxo01BA9VY:bqua414) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/index.tsx#L49)):*

does defining these this way instead of normal method definitions (`onZoom() { ... }`) change the `this` binding behavior or something? Otherwise, would prefer a normal method definition for consistency.
</blockquote></details>

Yes, this is the ES6 method of creating a "bound" method which is bound to its object.  This is necessary because onZoom and onResize are both used as callbacks, so they must be bound; this is the most efficient place to do that.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz04mZwjyctfek8FBNP:-Kz0C4xWK1keDWP5OhGb:b-yfa6m8) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L24)):*

Why on both of these magic constants?
</blockquote></details>

That's the base radius of the element to be drawn in space, just a pixel number that spencer came up with through experimentation.  This looks like the "right size" on the screen when drawn with a scale of 1.

This number is therefore art, not science, and I think that means it's allowed to be magic.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 33 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz056YksxjnXrBhVQ0k:-Kz0ChM6mg70_h4iqEGu:b-fsu0li) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L33)):*

I think this would be simpler as just a circle with a stroke of the appropriate width?
</blockquote></details>

Yes, I agree.  There are likely numerous optimizations to make here; I just took this almost directly from Spencer's code, which also allowed me to re-use his stylesheet.

However, the specific visualizations are likely to change dramatically - they don't resemble current design mocks much at all. So, while you are definitely correct, this specific component does not need to be optimized any further.  What's important is that it *is* a component, and thus gives us a blueprint for how to swap out other visual components.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 89 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz06xd6fq1ooK9Kx4vJ:-Kz0CvfVApvFv2Mm9pDU:bhtyt1r) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L89)):*

This seems excessively complicated for the purpose...
</blockquote></details>

Agreed, but see comment above.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz073XN93YyoAYslKw5:-Kz0hsI2uP5uvNqw4INq:bhtyt1r) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/locality.tsx#L101)):*

Seems like using `d3-scale` would simplify this.
</blockquote></details>

Agreed, but see comment above.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-zS-F3QNhP2NNbWJl:-Kz0iP-1mwYzydApMRtS:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx#L1)):*

nit: capitalize the second L in the file name: `localityLink.tsx`
</blockquote></details>

Done.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx, line 24 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01sMs3kRnheXKBbeR:-Kz0iW8dSSd3uT1CcyWF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx#L24)):*

nit: line length
</blockquote></details>

Done.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01r14mdXX6SpgTdjD:-Kz0ia2lm77rfw7C-PgT:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/localitylink.tsx#L32)):*

nit: outdent the closing tag to match opening tag
</blockquote></details>

Done.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts, line 14 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz05wH3vREI50lQOLMe:-Kz0ikB-oMlwRmI7r5bS:b-l11na3) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts#L14)):*

I have to say I don't think this abstraction helps, I think it gets in the way.
</blockquote></details>

I also don't like it much, but this is part of that "subject to be removed completely when we reskin" group.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts, line 62 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz06_TFVFo7xT4CCa1U:-Kz0idXBjPKtGfPSdXko:bllgaqy) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/pathmath.ts#L62)):*

Seems like everything in this file would benefit from some testing.
</blockquote></details>

Agreed, but I think a lot of this is going to change based on our actual visualizations.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 31 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-z14SDsc56kjVS1tS:-Kz0kwkZotc_33KLxvZ_:b-jqim2t) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L31)):*

It doesn't seem like it adds anything but complexity to keep the old one around after this point.
</blockquote></details>

It could be simplified in that way, but I think this component is going to go through some dramatic changes in the very near future.  Most importantly, the responsibility of handling old statuses should move to the reducer.

Another possibility is that we just query time series data when we need rates; i'm not sure the best way to approach it, the tradeoffs between node statuses and time series are subtle and it's tough to get an intuition about their relative merits.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 49 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-zyLxDX1z8nCR8V7R:-Kz0jNsuMib5LItWD7M-:bss4a2b) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L49)):*

I *think* this will double-count DistSQL selects, since `sql.distsql.select.count` is already included in `sql.select.count`.
</blockquote></details>

Removed.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz0-BCfZFX98C3I7tQd:-Kz0khGxgTaemmiE6q4V:bgghr06) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L53)):*

This looks odd, I guess it's just that `totalQPS` isn't actually per second?
</blockquote></details>

Renamed to "totalOps".

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts, line 15 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02571SLHu_zUqGO2q:-Kz0lzWZsWQMD5dPtSD7:b-f8sf1p) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/statusHistory.ts#L15)):*

Please add a comment explaining what the "activity rate" is. It looks like QPS over the last couple of updates (?)
</blockquote></details>

Added a comment.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/typings.d.ts, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-wV_vnx8TJ23xFt7i:-Kz0mTtANfOwtptRdRgd:b-7yd1q4) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/typings.d.ts#L9)):*

I see that this is for the world and US States JSON files. You could declare them as JS files, like

```
const states = {
  // ...geojson...
}

export default states;
```

And avoid this funny business with typings.
</blockquote></details>

Added a TODO.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts, line 10 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01HoUfvd1X9DOA3V-:-Kz0n3SacVNgT11GZ69D:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/vector.ts#L10)):*

maybe DRY as `return length(sub(n1, n2))`?
</blockquote></details>

Done.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx, line 26 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02KESFwaTzloafhVA:-Kz0nYqIV_w7HL33865y:beoudl0) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/ccl/src/views/clusterviz/containers/map/worldmap.tsx#L26)):*

A comment that this forms the lavender background would be nice (had to look at the CSS to figure out what it was doing). Or maybe `className="background"` or something
</blockquote></details>

Changed to class to `map-background`.

---

*[pkg/ui/ccl/src/views/clusterviz/containers/map/zoom.ts, line 1 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz02WEnIQqvUdz39Rca:-Kz0nbtqkjjwli4UqfgR:b1lhte2) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/ccl/src/views/clusterviz/containers/map/zoom.ts#L1)):*

Do we really need to implement all this ourselves?  What do we need here that isn't offered directly by `d3-zoom`?
</blockquote></details>

It seems that a lot of this is offered by the v4 version d3-zoom.  Specifically, it has both TranslateExtent and an extendable Constraint function; v3 doesn't have this, and that's what this code is mostly doing.

Note that we still might need *some* custom code in order to recompute constraints based on the viewport size (because in our design, that changes if the window resizes), but it might not be enough code to justify this companion class.

Most of this code here (Box, Point, etc) is also created for lack of a nice library for 2D vector math. I will figure out what to do on that front as this project continues developing.

---

*[pkg/ui/src/index.tsx, line 47 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-tRDHonY3-jOntcLu:-Kz0pRmc1Hsa-k55OZP-:bs9sv5z) ([raw file](https://github.com/cockroachdb/cockroach/blob/8faf12e2990f20f55c7a2e992c8b30f084021abd/pkg/ui/src/index.tsx#L47)):*

looks like the file being imported here is still around.
</blockquote></details>

Yup, compiler caught that; deleted.

---

*[pkg/ui/package.json, line 13 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz-x8k2coBdjTCPcSfZ:-Kz0c4AIilMttzbx3X-v:bzwiy2) ([raw file](https://github.com/cockroachdb/cockroach/blob/c88baf86655795291af896fd3e293192abd79d06/pkg/ui/package.json#L13)):*

It's a bit awkward to depend on d3 v3 and v4 at the same time, but it's probably fine.
</blockquote></details>

I think this is the result of some confusion while porting Spencer's code over to our codebase; that port was using a strange mixture of v3 and v4, but the code i'm checking in uses only v3.  Removed these.

---

*[pkg/ui/package.json, line 13 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-Kz01a6Yiz4GL98mG3IY:-Kz0a8-jJKztORHN0KTw:br0ghld) ([raw file](https://github.com/cockroachdb/cockroach/blob/a41ecad5ae48fb9e76991e80229215b62d3f6135/pkg/ui/package.json#L13)):*

We probably have this code in the bundle twice (modulo v3 vs v4 differences) — these are standalone libraries that are part of D3 v4; we already have their predecessors as part of D3, which we have because it's a dependency of NVD3.

Would be nice to not duplicate that somehow. Maybe @couch's graph stuff can use v4 and then we can just import these little v4 packages as we need them.
</blockquote></details>

 See Andrew's comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20066#-:-Kz0qmIk2tE6KG9HT18I:b-hn67i)*

couchand:
@mrtracy thank you for your thoughtful responses to the above comments.  I think you've got a good handle on the tradeoffs of how much it's worth it to clean things up given the state this is in and the direction it's headed.

I'd like to one last time detail my concerns, and after that we can shoot this pony and just come back to take care of it a few weeks from now.

My concerns about this don't appear in a vacuum.  The first commit on this PR is the removal of the non-functional code that was put in master the last time we tried this technique ("just stick it in so we can see what it's like").  A recent PR of mine was a similar situation: removing broken, dead code that was just sitting around because we half-baked a feature and then never got around to finishing it.

If this were a feature on any other part of the system I seriously doubt we would take this approach with it.  We wouldn't merge a SQL change that has syntax that doesn't match the RFC and an implementation we're expecting to change right away just to "make progress".  We wouldn't merge a RESTORE implementation that, not matter the backup file provided, always restores the TPC-H data set, just to "get a feel for it".

But that's exactly what we're doing here: we're going to merge something that relies (in part) on bogus data, then immediately starting work on reskinning, refactoring, and finishing it.  Reskinning: the implementation doesn't match the latest designs.  Refactoring: we already have plans to rewrite large swaths of the code.  Finishing it, because implementing it with all the real data would make this already-large PR that much larger.

The solution to a too-large PR isn't to just merge it as-is and then finish it on master.  The solution is to bite off a smaller chunk -- don't try to do the whole thing halfway, do half the thing completely.

I'm not ranting at you, Matt, I don't want to give you any stress.  I think you and I agree on the right way forward, and I understand that you're under pressure here.  No, I'm not ranting at you, I'm ranting at those who are putting the pressure on you to merge something incomplete.

But here we are, let's shoot this pony.  :shipit:

PR #20069, opened 2017-11-15
sql: make UPDATE and DELETE symmetric with INSERT/UPSERT

knz:
### sql: make DELETE support ORDER BY

This is for symmetry with INSERT/UPSERT that also supports it,
and is complementary to LIMIT (the effects of LIMIT are
non-deterministic without ORDER BY).

Both DELETE ... LIMIT (introduced previously) and
DELETE ... ORDER BY (introduced in this patch) are MySQL
extensions.

Release note (sql change): the DELETE statement now also supports
an ORDER BY clause, which constrains the deletion order,
the output of its LIMIT clause (if any) and the result order
of its RETURNING clause (if any).

### sql: support ORDER BY and LIMIT in UPDATE

This is a MySQL extension.

This patch introduces this feature:

1) for symmetry with INSERT/UPSERT/DELETE that also support these
   clauses.

2) to enable renumbering the PK of a table reliably without duplicate
   key violations (ORDER BY).

3) to limit the size of write transactions (LIMIT).

Regarding point 2. For example, consider a table contains PK values 0,
1, 2, 3, ...  Without this feature, one cannot renumber to 1, 2, 3
... (starting from 1 instead of 0), because the change from 0 to 1
would violate the existing row with PK 1. With an ORDER BY clause this
becomes possible: `UPDATE ... SET k = k + 1 ORDER BY k DESC` (starting
from the end).

Release note (sql change): the UPDATE statement now supports an ORDER
BY and LIMIT clause. This is a MySQL extension that can help with
updating the PK of a table (ORDER BY) and control the maximum size of
write transactions (LIMIT).

cc @bdarnell

### DIFF
--- a/pkg/sql/delete.go
+++ b/pkg/sql/delete.go
@@ -96,7 +96,7 @@ func (p *planner) Delete(
 		Exprs: sqlbase.ColumnsSelectors(rd.FetchCols),
 		From:  &tree.From{Tables: []tree.TableExpr{n.Table}},
 		Where: n.Where,
-	}, nil, n.Limit, nil, publicAndNonPublicColumns)
+	}, n.OrderBy, n.Limit, nil, publicAndNonPublicColumns)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/sql/logictest/testdata/logic_test/delete
+++ b/pkg/sql/logictest/testdata/logic_test/delete
@@ -174,6 +174,34 @@ CREATE TABLE indexed (id int primary key, value int, other int, index (value))
 statement ok
 DELETE FROM indexed WHERE value = 5
 
+# Check DELETE with ORDER BY clause (MySQL extension)
+
+statement ok
+INSERT INTO unindexed VALUES (1, 9), (8, 2), (3, 7), (6, 4)
+
+query ITTT
+EXPLAIN DELETE FROM unindexed WHERE v = 7 ORDER BY v
+----
+0  delete  ·      ·
+0  ·       from   unindexed
+1  nosort  ·      ·
+1  ·       order  +v
+2  scan    ·      ·
+2  ·       table  unindexed@primary
+2  ·       spans  ALL
+
+query II
+DELETE FROM unindexed WHERE k > 1 AND v < 7 ORDER BY v DESC RETURNING v,k
+----
+4  6
+2  8
+
+query II
+DELETE FROM unindexed ORDER BY v RETURNING k,v
+----
+3  7
+1  9
+
 # Check DELETE with LIMIT clause (MySQL extension)
 
 statement ok
--- a/pkg/sql/logictest/testdata/logic_test/insert
+++ b/pkg/sql/logictest/testdata/logic_test/insert
@@ -578,27 +578,66 @@ INSERT INTO kv (k.*) VALUES ('hello')
 statement error compound types not supported yet: "k.v"
 INSERT INTO kv (k.v) VALUES ('hello')
 
-# Test INSERT INTO ... SELECT FROM ... LIMIT 1
-statement ok
-CREATE TABLE insert_t (x INT,v INT,PRIMARY KEY (x))
+
 
 statement ok
-CREATE TABLE select_t (x INT,v INT,PRIMARY KEY (x))
+CREATE TABLE insert_t (x INT, v INT)
 
 statement ok
-INSERT INTO select_t VALUES(1, 2),(2,3),(3,4)
+CREATE TABLE select_t (x INT, v INT)
+
+statement ok
+INSERT INTO select_t VALUES (1, 9), (8, 2), (3, 7), (6, 4)
+
+# Check that INSERT supports ORDER BY (MySQL extension)
+
+query ITTT
+EXPLAIN (PLAN,EXPRS) INSERT INTO insert_t TABLE select_t ORDER BY v DESC
+----
+0  insert  ·          ·
+0  ·       into       insert_t(x, v, rowid)
+0  ·       default 0  NULL
+0  ·       default 1  NULL
+0  ·       default 2  unique_rowid()
+1  sort    ·          ·
+1  ·       order      -v
+2  render  ·          ·
+2  ·       render 0   x
+2  ·       render 1   v
+3  scan    ·          ·
+3  ·       table      select_t@primary
+3  ·       spans      ALL
+
+query II
+INSERT INTO insert_t TABLE select_t ORDER BY v DESC RETURNING x, v
+----
+1  9
+3  7
+6  4
+8  2
+
+# Check that INSERT supports LIMIT (MySQL extension)
+
+statement ok
+TRUNCATE TABLE insert_t
 
 query ITTT
 EXPLAIN (PLAN,EXPRS) INSERT INTO insert_t SELECT * FROM select_t LIMIT 1
 ----
-0  insert  ·      ·
-0  ·       into   insert_t(x, v)
-1  limit   ·      ·
-1  ·       count  1
-2  scan    ·      ·
-2  ·       table  select_t@primary
-2  ·       spans  ALL
-2  ·       limit  1
+0  insert  ·          ·
+0  ·       into       insert_t(x, v, rowid)
+0  ·       default 0  NULL
+0  ·       default 1  NULL
+0  ·       default 2  unique_rowid()
+1  limit   ·          ·
+1  ·       count      1
+2  render  ·          ·
+2  ·       render 0   x
+2  ·       render 1   v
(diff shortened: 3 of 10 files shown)


justinj:
Are there any case where the insertion/deletion order is significant besides self-referential foreign keys and `...RETURNING`?

---

Reviewed 4 of 5 files at r1.
Review status: 2 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/render.go, line 360 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-DLTFVMyY6HZKrjL9:-Kz-DLTFVMyY6HZKrjLA:b-xtgh8k) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f4437e4cd3775c0b346b7258eb38fd7400be366/pkg/sql/render.go#L360)):*
> ```Go
> }
> 
> // insertRenderNode creates a new renderNode that renders exactly its
> ```

nit: name in the comment doesn't match the name of the method

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-C_wHIB723yhZ9QJX:-Kz-C_wHIB723yhZ9QJY:bc0av1j) ([raw file](https://github.com/cockroachdb/cockroach/blob/1731d8d754eeea88fa88d33e996fac9ee257a326/pkg/sql/logictest/testdata/logic_test/delete#L177)):*
> ```
> DELETE FROM indexed WHERE value = 5
> 
> # Check DELETE with ORDER BY clause (MySQL extension)
> ```

Should there also be a test involving a self-referential foreign key (and ditto for the other tests?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-:-Kz-EFDV3pi5xnK18NO9:b-a7fm8b)*

knz:
Yes, any built-in functions with side effects:

-`current_time()` is evaluated on every row and this guarantees (currently) that the updated timestamps will be monotonic.
- sequences (as soon as Pete's PR lands) and `lastval()`

---

Review status: 2 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/render.go, line 360 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-DLTFVMyY6HZKrjL9:-Kz-ZkCf3gSSCFjVBXUn:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/4f4437e4cd3775c0b346b7258eb38fd7400be366/pkg/sql/render.go#L360)):*

nit: name in the comment doesn't match the name of the method
</blockquote></details>

Done.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-C_wHIB723yhZ9QJX:-Kz-Zlq9D3lcP5ZAI34E:b-j1snod) ([raw file](https://github.com/cockroachdb/cockroach/blob/1731d8d754eeea88fa88d33e996fac9ee257a326/pkg/sql/logictest/testdata/logic_test/delete#L177)):*

Should there also be a test involving a self-referential foreign key (and ditto for the other tests?)
</blockquote></details>

I do not understand. How are self-referential foreign keys here any more relevant than for every other DELETE test that were previously present in this file?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-:-Kz-ZvBPPKFEFOP2oQGx:bkat3ly)*

knz:
I meant `nextval()` sorry.

---

Review status: 2 of 10 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-:-Kz-_ICfwxq5KSEoWnXv:bfljz65)*

justinj:
Review status: 2 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-C_wHIB723yhZ9QJX:-Kz-frv56ewFX4TqFHvj:bk4nq4m) ([raw file](https://github.com/cockroachdb/cockroach/blob/1731d8d754eeea88fa88d33e996fac9ee257a326/pkg/sql/logictest/testdata/logic_test/delete#L177)):*

I do not understand. How are self-referential foreign keys here any more relevant than for every other DELETE test that were previously present in this file?
</blockquote></details>

IIRC, the MySQL doc for this feature explicitly mentions the use case of: If `a` references `b`, deleting in an order that deletes `b` before `a` will fail, but deleting in an order that deletes `a` before `b` will succeed. Is this not the case for us?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069)*

knz:
Review status: 2 of 10 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-C_wHIB723yhZ9QJX:-Kz3f1IMfduEHiSTiDU7:b-rxhy95) ([raw file](https://github.com/cockroachdb/cockroach/blob/1731d8d754eeea88fa88d33e996fac9ee257a326/pkg/sql/logictest/testdata/logic_test/delete#L177)):*

IIRC, the MySQL doc for this feature explicitly mentions the use case of: If `a` references `b`, deleting in an order that deletes `b` before `a` will fail, but deleting in an order that deletes `a` before `b` will succeed. Is this not the case for us?
</blockquote></details>

Yes, that is true, but IMHO does not warrant a separate test: once we have a mechanism that influences/restricts the order of writes, and a test establishes this to be working, we know we have the property.

Otherwise, not only I would need to add FK tests here, but also check that sequences are indeed updated in the right order, and timestamps are monotonic, and all these other things. 

I'm afraid I'm not seeing logic tests as the place to provide example uses of CockroachDB.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069)*

justinj:
:lgtm:

---

Review status: 2 of 10 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-C_wHIB723yhZ9QJX:-Kz4Ha8CFAYko-xqOVeg:b-5v1nc2) ([raw file](https://github.com/cockroachdb/cockroach/blob/1731d8d754eeea88fa88d33e996fac9ee257a326/pkg/sql/logictest/testdata/logic_test/delete#L177)):*

Yes, that is true, but IMHO does not warrant a separate test: once we have a mechanism that influences/restricts the order of writes, and a test establishes this to be working, we know we have the property.

Otherwise, not only I would need to add FK tests here, but also check that sequences are indeed updated in the right order, and timestamps are monotonic, and all these other things. 

I'm afraid I'm not seeing logic tests as the place to provide example uses of CockroachDB.
</blockquote></details>

That's fair! My thought were along the line of that this is something that could be affected by some kind of future or present fk check batching, but if you don't think there's any concern then that seems fine to me!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-:-Kz4HZQLCZhQS0MNBDc0:bnfp4nl)*

knz:
TFYR!

---

Review status: 2 of 10 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/delete, line 177 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-Kz-C_wHIB723yhZ9QJX:-Kz4LFxBx5haNWCZa9G5:b5igcy) ([raw file](https://github.com/cockroachdb/cockroach/blob/1731d8d754eeea88fa88d33e996fac9ee257a326/pkg/sql/logictest/testdata/logic_test/delete#L177)):*

That's fair! My thought were along the line of that this is something that could be affected by some kind of future or present fk check batching, but if you don't think there's any concern then that seems fine to me!
</blockquote></details>

Hah now I understand your concern. 
Note however how I was careful in the commit message to not say anything about FK checks. Indeed I do not know for sure whether we will want to preserve that use case from MySQL. I think we should keep this silent until then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20069#-:-Kz4LYEQtyK1X7iPY7jL:b-814cj3)*

PR #20053, opened 2017-11-14
RFCS: minor update to statistics rfc

RaduBerinde:
Changing to "cardinality" from "density" and removing the density vector stuff. I realized it's more flexible if each statistic has a single cardinality; we can have multiple entries as necessary (and we're not constrained on which groups to include).

PR #20072, opened 2017-11-15
sql: Add table system.table_statistics

rytaft:
Add a new system table for holding table statistics. This table
contains a row per attribute or group of attributes that we maintain
stats for. The schema is:

```
CREATE TABLE system.table_statistics (
  tableID INT NOT NULL,
  statisticID INT NOT NULL,
  columnIDs INT[] NOT NULL,
  createdAt TIMESTAMP NOT NULL DEFAULT now(),
  rowCount INT NOT NULL,
  cardinality INT NOT NULL,
  nullValues INT NOT NULL,
  histogram BYTES,
  PRIMARY KEY (tableID, statisticID)
)
```
Most of the files changed in this commit are test cases that need to
be updated whenever a new system table is added.

Release Note (sql change): Added a new table system.table_statistics
for maintaining statistics about columns or groups of columns in the
database.


petermattis:
Excited to see this happening.

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz-YaPg9kiR82PMn9yN:b-kkwqwf) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*
> ```Go
> 	"createdAt"   TIMESTAMP  NOT NULL DEFAULT now(),
> 	"rowCount"    INT        NOT NULL,
> 	cardinality   INT        NOT NULL,
> ```

Isn't `cardinality` the same as `rowCount`? The schema here doesn't match the one in the RFC. For example, `density` is missing and this contains `nullValues` instead of `null_fraction`. If that was intentional, we should update the RFC. 

---

*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz-YQjh8GXUbZEawpSn:b-vg9b7s) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*
> ```Go
> 			{Name: "fam_7_nullValues", ID: 7, ColumnNames: []string{"nullValues"}, ColumnIDs: []ColumnID{7}, DefaultColumnID: 7},
> 			{Name: "fam_8_histogram", ID: 8, ColumnNames: []string{"histogram"}, ColumnIDs: []ColumnID{8}, DefaultColumnID: 8},
> 		},
> ```

I would think we'd want one column family for this table. I realize some of the other system tables have multiple column families, but that is due to historical reasons. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz-Z91ng_9LDJKGJbG0:b-du1v13)*

danhhz:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz-jXsGwBsX9HY3WkL1:b-4n6jyi) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*

I would think we'd want one column family for this table. I realize some of the other system tables have multiple column families, but that is due to historical reasons. 
</blockquote></details>

 =peter. you can manually specify the family in the CREATE TABLE (see jobs or web_sessions above)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072)*

rytaft:
Thanks for the review!

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz-rqY9dTbRunZb_gj5:b-oa49vl) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

Isn't `cardinality` the same as `rowCount`? The schema here doesn't match the one in the RFC. For example, `density` is missing and this contains `nullValues` instead of `null_fraction`. If that was intentional, we should update the RFC. 
</blockquote></details>

Here `cardinality` refers to the number of unique values in the column group (i.e. set cardinality).  `density` was removed because `density = 1/cardinality`. The RFC is in the process of being updated to match this schema as part of PR #20053.

---

*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz-ro_QM60t45RUlnou:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*

 =peter. you can manually specify the family in the CREATE TABLE (see jobs or web_sessions above)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz-sZA3r3SYdwsUHmLY:b-9qdoz3)*

RaduBerinde:
:lgtm:

Thanks, that was fast!

Looking at other systems like PG, SQL Server, they have a CREATE STATISTICS command which takes a statistic name. Probably useful for users to keep track of statistics they add manually, wondering if we should include a "name" column.

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz-xmtOOjP_w9wuKxJD:byors35) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*

Done.
</blockquote></details>

Hm, I would put the histogram in a separate family (it can be on the order of a KB). Would be good to skip decoding it when we only need to get a list of available statistics for a table.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz-y4VFAgNFnbHFpEKw:bhmbjoa)*

danhhz:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz-zvvgNXjS85MnQHYR:b-fxcxmj) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*

Hm, I would put the histogram in a separate family (it can be on the order of a KB). Would be good to skip decoding it when we only need to get a list of available statistics for a table.
</blockquote></details>

You could be right, but I would suggest you only do this only if you can measure it. When I did column family performance testing, it was surprising how much better 1 family was, even for workload that you'd expect to trigger worst case behavior. It's also possible that this has changed since I did it, since our per-kv overhead has certainly gone down

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072)*

RaduBerinde:
*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz00DS7wBsJ_dHOhPGh:bnlygh7) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*

You could be right, but I would suggest you only do this only if you can measure it. When I did column family performance testing, it was surprising how much better 1 family was, even for workload that you'd expect to trigger worst case behavior. It's also possible that this has changed since I did it, since our per-kv overhead has certainly gone down
</blockquote></details>

Thanks! We'll go with your intuition since you are most knowledgeable about this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072)*

petermattis:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz021swd3t7rnos7xxz:birc4p6) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

Here `cardinality` refers to the number of unique values in the column group (i.e. set cardinality).  `density` was removed because `density = 1/cardinality`. The RFC is in the process of being updated to match this schema as part of PR #20053.
</blockquote></details>

`cardinality` and `rowCount` are very similar terms. I worry about that confusion. I've seen `cardinality` named `distinctValueCount` in some other database.

Apologies for the bike shedding. 

---

*[pkg/sql/sqlbase/system.go, line 664 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YQjh8GXUbZEawpSm:-Kz01veqVwZbIdAb6fs8:b-rntdsx) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L664)):*

Thanks! We'll go with your intuition since you are most knowledgeable about this.
</blockquote></details>

My intuition is one column family will be best here as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072)*

rytaft:
Thanks! Sure, I can add a "name" column - presumably nullable? Are there other columns you think we should add now (e.g., "filter_predicate", "rows_sampled", "filtered_rows"....)?  I guess we'll also need to update the RFC again in that case...

I'm not sure how easy it is to add columns to system tables after they've been created - I guess we'd just need to add an `ALTER TABLE` equivalent command to `migrations.go`?

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz03olFRs4P4vXPjwoV:bzavcc5) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

`cardinality` and `rowCount` are very similar terms. I worry about that confusion. I've seen `cardinality` named `distinctValueCount` in some other database.

Apologies for the bike shedding. 
</blockquote></details>

No worries I can make that change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz-zILoSG_NXcbkfpIR:btxbi21)*

RaduBerinde:
*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz04HFSJ9kHcyU3uvJL:b1r6111) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

No worries I can make that change.
</blockquote></details>

Maybe `distinctCount` so it's not too long?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072)*

RaduBerinde:
I was thinking of filtered statistics too, we may want to support those later. If migration is easy then we can worry about that later. I do see another "ALTER TABLE" in migrations.go so perhaps it's that easy. CC @knz

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz04VtPByYnAjqBHEb-:bao2r4t)*

rytaft:
It does look to be pretty straightforward.  Assuming that's the case, we can probably hold off on the columns related to filtered statistics until we're ready to support them.  But it does seem worthwhile adding something like "rows_sampled" or "sampled_fraction" now to indicate what fraction of the total rows were sampled.  What do you think?

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz08xedydRXJUtiQzpP:b-bo5wck) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

Maybe `distinctCount` so it's not too long?
</blockquote></details>

works for me...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz08GVTSpAOlO_ABvIL:bqdepgq)*

RaduBerinde:
I'm hesitant to add info about sampled rows because sampling only applies to the histogram. The cardinality and other stats are obtained from the full table.

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz0JZ5NRuPcyc_zB62z:b-ipuzxi)*

petermattis:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz0OP760rLUhexIdLrD:b-lnu8y6) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

works for me...
</blockquote></details>

`distinctCount` is good by me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072)*

rytaft:
Ah ok I was confused - thanks for clarifying, @RaduBerinde.  I guess we can also include any histogram metadata we want in the `histogram` field.

I've added the "name" column, and changed "cardinality" to "distinctCount".

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/sqlbase/system.go, line 157 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz-YaPg9kiR82PMn9yM:-Kz0YbTFhmaB2Drza26S:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/857c8b82515ca8f1738abee187fa35a2079b56a1/pkg/sql/sqlbase/system.go#L157)):*

`distinctCount` is good by me.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz0Q0jQMwF794aRkof4:b-sb3k8d)*

petermattis:
:lgtm:

---

Review status: 0 of 12 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/sql/sqlbase/system.go, line 159 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz0ZT-xGsoqcJHdw96O:-Kz0ZT-xGsoqcJHdw96P:b-c6amiw) ([raw file](https://github.com/cockroachdb/cockroach/blob/df8293ce0505d7cd2ea64bb0e2b2a5f7566eac6a/pkg/sql/sqlbase/system.go#L159)):*
> ```Go
> 	"rowCount"      INT        NOT NULL,
> 	"distinctCount" INT        NOT NULL,
> 	"nullValues"    INT        NOT NULL,
> ```

One more bikeshed: should this be `nullCount` (for consistency)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz0ZZ8DVHTQnSPzueQK:bnfp4nl)*

rytaft:
Thanks!

---

Review status: 0 of 12 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/sql/sqlbase/system.go, line 159 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz0ZT-xGsoqcJHdw96O:-Kz0cCPVb-ft5DX-oYlp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/df8293ce0505d7cd2ea64bb0e2b2a5f7566eac6a/pkg/sql/sqlbase/system.go#L159)):*

One more bikeshed: should this be `nullCount` (for consistency)?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz0cFitnau2O6fLXnvd:b-ucfknh)*

RaduBerinde:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz0kXc-yerNesMbFDNt:bnfp4nl)*

knz:
:lgtm:

---

Reviewed 12 of 12 files at r1, 2 of 4 files at r2, 2 of 2 files at r3.

---

*[pkg/sql/sqlbase/system.go, line 162 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz3Qbht2JE5sqrMzewc:-Kz3Qbht2JE5sqrMzewd:b-bk7max) ([raw file](https://github.com/cockroachdb/cockroach/blob/d62ee57df38152c9f4c858ab6cfba8809aaa2ce3/pkg/sql/sqlbase/system.go#L162)):*
> ```Go
> 	histogram       BYTES,
> 	PRIMARY KEY ("tableID", "statisticID"),
> 	FAMILY("tableID", "statisticID", name, "columnIDs", "createdAt", "rowCount", "distinctCount", "nullCount", histogram)
> ```

nit: space between `FAMILY` and `(`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz3QgFHX9ajeZwL0kBP:bnfp4nl)*

rytaft:
TFTRs!

---

Review status: 11 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/sqlbase/system.go, line 162 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-Kz3Qbht2JE5sqrMzewc:-Kz4v82P_YRkbR9oBOr0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/d62ee57df38152c9f4c858ab6cfba8809aaa2ce3/pkg/sql/sqlbase/system.go#L162)):*

nit: space between `FAMILY` and `(`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20072#-:-Kz4vDS519QbG9XdR68J:bz5mair)*

PR #20073, opened 2017-11-15
grpcutil: Treat all gRPC errors as ambiguous

bdarnell:
In gRPC 1.6, we could identify certain unambiguous failures with
string matching on the error. In 1.7, the same error message is used
for both ambiguous and unambiguous cases, so we must assume it is
ambiguous.

Release note (bug fix): Fixed a case in which ambiguous errors were
treated as unambiguous and led to inappropriate retries.

Updates #19708

### DIFF
--- a/pkg/util/grpcutil/grpc_util.go
+++ b/pkg/util/grpcutil/grpc_util.go
@@ -79,7 +79,11 @@ func RequestDidNotStart(err error) bool {
 		// This is a non-gRPC error; assume nothing.
 		return false
 	}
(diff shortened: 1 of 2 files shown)


a-robinson (approved):
:(

jordanlewis:
Hmm, this doesn't seem great. I might be missing some context, but have we already discussed trying to upstream a fix? As we know, ambiguous errors are a big pain point for our clients.

bdarnell:
Yeah, that's why #19708 is still open. This change is just to fix the test failure we're currently seeing every night.

PR #20074, opened 2017-11-15
cherrypick-1.1: sqlbase: fix encoding of index value when using unspecified columns

### LINKED ISSUE #20000: Cockroach server crashes on certain UPSERT

vivekmenezes:
prior to this fix we would treat an unspecified column as a NULL value
when encoding a column for an index key, but not when encoding it in the
index as the value part of the key:value index pair.

For the storing case, we would pick up the first specified value
for the unspecified value and encode it in the index creating a
completely broken index!

For the composite encoding case it would crash. See #20000

fixes #20000


lgo (approved):
LGTM

PR #20084, opened 2017-11-15
sql: support JSON columns

justinj:


m-schneider:
Review status: 0 of 18 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/cli/dump_test.go, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz57yioK_Im2xFqfQje:-Kz57yipx9HAXWyK67pC:b-ebgx4e) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/cli/dump_test.go#L88)):*
> ```Go
> 		'e9716c74-2638-443d-90ed-ffde7bea7d1d',
> 		'192.168.0.1',
> 		'{"a":"b"}',
> ```

Does it make sense to test more forms or JSON here?

---

*[pkg/sql/logictest/testdata/logic_test/json, line 86 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz58IE7KiOPdblyfGl4:-Kz58IE7KiOPdblyfGl5:b-wk4iuv) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/sql/logictest/testdata/logic_test/json#L86)):*
> ```
> 
> statement error arrays of JSONB not allowed
> CREATE TABLE x (y JSONB[])
> ```

Do you think it would be helpful to add tests with more nesting?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084)*

justinj:
Review status: 0 of 18 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/cli/dump_test.go, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz57yioK_Im2xFqfQje:-Kz5AcToEjKF6rW0i2Dj:b-i8kxw3) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/cli/dump_test.go#L88)):*

Does it make sense to test more forms or JSON here?
</blockquote></details>

I don't think so - this is mostly a smoke test that the type works properly, and `TestDumpRandom` inserts other forms as well.

---

*[pkg/sql/logictest/testdata/logic_test/json, line 86 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz58IE7KiOPdblyfGl4:-Kz5ApbGOk5ejIUOOGD2:b-immyqj) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/sql/logictest/testdata/logic_test/json#L86)):*

Do you think it would be helpful to add tests with more nesting?
</blockquote></details>

What nesting do you mean? If you mean multiple pairs of `[]` we don't support that syntax:

```
root@:26257/> create table x (y int[][]);
invalid syntax: statement ignored: syntax error at or near "["
DETAIL: source SQL:
create table x (y int[][]);
                       ^
HINT: try \h CREATE TABLE
```
(and in Postgres multiple `[]` is functionally exactly the same as a single `[]`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084)*

jordanlewis:
:lgtm: cool! This was pleasantly compact.

Should add some more language acceptance tests though, probably, although I suppose it might be somewhat less important with jsonb because it's transmitted via the text protocol.

Python seems to let you insert native Python json, for example https://stackoverflow.com/questions/46374367/how-to-insert-jsonb-into-postgresql-with-python

---

Reviewed 18 of 18 files at r1.

---

*[pkg/acceptance/testdata/java/src/main/java/JSONTest.java, line 29 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz4Un37x-po6pkDZwTx:-Kz4Un37x-po6pkDZwTy:b-r4k2ha) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/acceptance/testdata/java/src/main/java/JSONTest.java#L29)):*
> ```Java
>         ResultSet rs = stmt.executeQuery();
>         rs.next();
>         Assert.assertEquals("{\"a\":\"b\"}", rs.getString(1));
> ```

Yuck! Too bad there's no JDBC JSON support.

---

*[pkg/util/encoding/encoding.go, line 1406 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz5Dafqy7uNzgKKXB1T:-Kz5Dafqy7uNzgKKXB1U:b-a23scp) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/util/encoding/encoding.go#L1406)):*
> ```Go
> func EncodeJSONValue(appendTo []byte, colID uint32, data []byte) []byte {
> 	appendTo = EncodeValueTag(appendTo, colID, JSON)
> 	return EncodeUntaggedBytesValue(appendTo, data)
> ```

Probably just missing something, but why don't we ever call DecodeUntaggedBytesValue? Seems like we just reimplement what it does instead in DecodeJSON? It might be more clear to just call that unles s it does extra work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-:-Kz5LdZHeGZAnRLfePI2:b-6rt6w3)*

justinj:
I added a python acceptance test.

TFTRs! Planning to merge on green!

---

Review status: 15 of 23 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[pkg/util/encoding/encoding.go, line 1406 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-Kz5Dafqy7uNzgKKXB1T:-KzA_gJYfRlduRiBuKeo:bmfg19y) ([raw file](https://github.com/cockroachdb/cockroach/blob/54316604c030bdd4934bff3c4450df4b947ced5e/pkg/util/encoding/encoding.go#L1406)):*

Probably just missing something, but why don't we ever call DecodeUntaggedBytesValue? Seems like we just reimplement what it does instead in DecodeJSON? It might be more clear to just call that unles s it does extra work.
</blockquote></details>

Yes, good call!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20084#-:-KzA_ldXbLuDHSAmGzkC:b-352nm9)*

PR #20087, opened 2017-11-16
tscache: replace intervalSkl's useMaxVal with cantInit

nvb:
### `useMaxVal` Issue

When comparing `treeImpl` to the new `sklImpl`, some of my equivalent-behavior
tests occasionally failed. The problem was that transactions would occasionally
"lose ownership" over part of a range that they should have retained. After some
debugging, it became clear that this was due to interactions with the `useMaxVal`
`nodeOption`. Previously, this option would be set on nodes when it was impossible
to ratchet them precisely because of a full arena. Instead, we would ratchet
them all the way to the max value of the node's page by flipping a flag that
didn't require an arena allocation. The effect of this was that even if a
transaction had ownership over a range of nodes, it could lose that if the arena
filled up while it was ratcheting the nodes to reflect its ownership. This was
not ok because it violated the minimum retention policy and would result in
spurious transaction restarts even without key contention.

### `cantInit`  Fix

To address this, this change gets rid of the `useMaxVal` option, replacing it
with a `cantInit` option. This option indicates that the node should never be
allowed to initialize. It is set on nodes who were unable to ratchet their
values at some point because of a full arena. When the flag is set, the node's
values are never allowed to become final. Instead, any goroutines trying to
initialize the node will be forced to create it again in a new page.

To test this, the change adds a new `TestIntervalSklConcurrentVsSequential`,
which failed immediately when using `useMaxVal`. This test inserts rounds of
random values into two `intervalSkls`. A round of values is inserted into the
first `intervalSkl` sequentially and into the second `intervalSkl` concurrently.
After each round, the test makes sure that both `intervalSkls` give the same
answers for all questions asked.

### Unrelated Race Exposed and Removed

In doing this, the newly added `TestIntervalSklConcurrentVsSequential` revealed
a race condition in the previous implementation. The issue was that if two
goroutines raced in `addNode`, the losing goroutine would not wait for the new
node to be initialized. If this node was meant to be the end of a range then
this could allow the "losing" goroutine to insert and initialize the range's
starting node before the "winning" goroutine finished initializing the range's
ending node. Once ranges are allowed to have initialized starting nodes but
uninitialized ending nodes, it becomes possible for their "influence" to leak
out of the end of the range.

This was addressed by adding an option to `addNode` to ensure that the node
is initialized by the time the method returns. It wasn't caught before because
all of our concurrency testing was asserting inequality conditions ("timestamps
at this key are at-least this big"). `TestIntervalSklConcurrentVsSequential`
allows us to assert strict equality ("timestamps are exactly this big"), which
catches the "leaking influence issue".

**Note that this change doesn't have any effects on the benchmarks**

### DIFF
--- a/pkg/storage/tscache/interval_skl.go
+++ b/pkg/storage/tscache/interval_skl.go
@@ -73,21 +73,22 @@ const (
 	// initialized. Key and gap values are final, and can now be used.
 	initialized = 1 << iota
 
+	// cantInit indicates that the node should never be allowed to initialize.
+	// This is set on nodes which were unable to ratchet their values at some
+	// point because of a full arena. In this case, the node's values should
+	// never become final and any goroutines trying to initialize it it will be
+	// forced to create it again in a new page when they notice this flag.
+	cantInit
+
 	// hasKey indicates that the node has an associated key value. If this is
 	// not set, then the key timestamp is assumed to be zero and the key is
 	// assumed to not have a corresponding txnID.
 	hasKey
 
 	// hasGap indicates that the node has an associated gap value. If this is
-	// not set, then the gap timestamp is assumed to be zero and the gep is
+	// not set, then the gap timestamp is assumed to be zero and the gap is
 	// assumed to not have a corresponding txnID.
 	hasGap
-
-	// useMaxVal indicates that the intervalSkl's max value should be used in
-	// place of the node's gap and key timestamps. This is used when the
-	// structure has filled up and the value can no longer be set (but meta
-	// can).
-	useMaxVal
 )
 
 const (
@@ -265,13 +266,14 @@ func (s *intervalSkl) addRange(from, to []byte, opt rangeOptions, val cacheValue
 	// Start by ensuring that the ending node has been created (unless "to" is
 	// nil, in which case the range extends indefinitely). Do this before creating
 	// the start node, so that the range won't extend past the end point during
-	// the period between creating the two endpoints.
+	// the period between creating the two endpoints. Since we need the ending node
+	// to be initialized before creating the starting node, we pass mustInit = true.
 	var err error
 	if to != nil {
 		if (opt & excludeTo) == 0 {
-			err = fp.addNode(&it, to, val, hasKey)
+			err = fp.addNode(&it, to, val, hasKey, true /* mustInit */)
 		} else {
-			err = fp.addNode(&it, to, val, 0)
+			err = fp.addNode(&it, to, val, 0, true /* mustInit */)
 		}
 
 		if err == arenaskl.ErrArenaFull {
@@ -287,9 +289,9 @@ func (s *intervalSkl) addRange(from, to []byte, opt rangeOptions, val cacheValue
 
 	// Ensure that the starting node has been created.
 	if (opt&excludeFrom) == 0 || len(from) == 0 {
-		err = fp.addNode(&it, from, val, hasKey|hasGap)
+		err = fp.addNode(&it, from, val, hasKey|hasGap, false /* mustInit */)
 	} else {
-		err = fp.addNode(&it, from, val, hasGap)
+		err = fp.addNode(&it, from, val, hasGap, false /* mustInit */)
 	}
 
 	if err == arenaskl.ErrArenaFull {
@@ -488,11 +490,18 @@ func (p *sklPage) lookupTimestampRange(from, to []byte, opt rangeOptions) cacheV
 	it.Init(p.list)
 	it.SeekForPrev(from)
 
-	return p.scanForMaxValue(&it, from, to, opt, false /* initFrom */)
+	return p.maxInRange(&it, from, to, opt)
 }
 
+// addNode adds a new node at key with the provided value if one does not exist.
+// If one does exist, it ratchets the existing node's value instead.
+//
+// If the mustInit flag is set, the function will ensure that the node is
+// initialized by the time the method returns, even if a different goroutine
+// created the node. If the flag is not set and a different goroutine created
+// the node, the method won't try to help.
 func (p *sklPage) addNode(
-	it *arenaskl.Iterator, key []byte, val cacheValue, opt nodeOptions,
+	it *arenaskl.Iterator, key []byte, val cacheValue, opt nodeOptions, mustInit bool,
 ) error {
 	// Array with constant size will remain on the stack.
 	var arr [encodedValSize * 2]byte
@@ -507,39 +516,37 @@ func (p *sklPage) addNode(
 	}
 
 	if !it.SeekForPrev(key) {
-		util.RacePreempt()
+		// The key was not found. Scan for the previous gap value.
+		prevGapVal := p.incomingGapVal(it, key)
 
-		// The key was not found. If the previous node has a gap value that
-		// would not be updated with the new value, then there is no need to add
-		// another node, since its timestamp would be the same as the gap
-		// timestamp and its txnID would be the same as the gap txnID.
-		//
-		// NB: We pass excludeTo here because another thread may race and add a
-		// node at the key while we're scanning for the previous value. We don't
-		// want to consider this new key here because even if it has a larger
-		// key value that indicates that we can take this fast path, it may have
-		// a smaller gap value that we need to ratchet below. Now, if a node
-		// races and is added at key while we're scanning for the previous
-		// value, we'll get an empty previous value, which won't trigger the
-		// optimization and we'll ratchet the node's value below.
-		prevVal := p.scanForMaxValue(it, key, key, excludeTo, false /* initFrom */)
-		if _, update := ratchetValue(prevVal, val); !update {
-			return nil
+		var err error
+		if it.Valid() && bytes.Equal(it.Key(), key) {
+			// Another thread raced and added a node at key while we were
+			// scanning backwards. Ratchet the new node.
+			err = arenaskl.ErrRecordExists
+		} else {
+			// There is still no node at key. If the previous node has a gap
+			// value that would not be updated with the new value, then there is
+			// no need to add another node, since its timestamp would be the
+			// same as the gap timestamp and its txnID would be the same as the
+			// gap txnID.
+			if _, update := ratchetValue(prevGapVal, val); !update {
+				return nil
+			}
+
+			// Ratchet max timestamp before adding the node.
+			p.ratchetMaxTimestamp(val.ts)
+
+			// Ensure that a new node is created. It needs to stay in the
+			// initializing state until the gap value of its preceding node
+			// has been found and used to ratchet this node's value. During
+			// the search for the gap value, this node acts as a sentinel
+			// for other ongoing operations - when they see this node they're
+			// forced to stop and ratchet its value before they can continue.
+			b, meta := encodeValueSet(arr[:0], keyVal, gapVal)
+			err = it.Add(key, b, meta)
 		}
 
-		// Ratchet max timestamp before adding the node.
-		p.ratchetMaxTimestamp(val.ts)
-
-		// Ensure that a new node is created. It needs to stay in the
-		// initializing state until the gap value of its preceding node
-		// has been found and used to ratchet this node's value. During
-		// the search for the gap value, this node acts as a sentinel
-		// for other ongoing operations - when they see this node they're
-		// forced to stop and ratchet its value before they can continue.
-		b, meta := p.encodeValueSet(arr[:0], keyVal, gapVal)
-
-		util.RacePreempt()
-		err := it.Add(key, b, meta)
 		switch err {
 		case arenaskl.ErrArenaFull:
 			atomic.StoreInt32(&p.isFull, 1)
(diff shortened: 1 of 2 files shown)


nvanbenschoten:
@andy-kimball I'd be interesting in your input. I'm also pretty confident that [andy-kimball/tscache](https://github.com/andy-kimball/tscache/blob/master/cache.go#L341) suffers from the same race as the one described above.

andy-kimball:
When I created the cache, I wasn't worrying about values that were too high, since I reasoned they were very rare. And as long as values ratcheted upwards, there should be no incorrect behavior, only performance impact. But I see why aborted transactions are a real concern for Cockroach, even if they happen rarely, so we should definitely tighten up the guarantees. I'll take a look at your changes.

nvanbenschoten:
> And as long as values ratcheted upwards, there should be no incorrect behavior, only performance impact.

Yep exactly, the use of the `useMaxVal` flag would never result in correctness issues, it would just result in false-positive read-write conflicts. In fact, the race addressed here would also have the same effect, which is why the previous test suite never caught it. You're also correct that these issues would be very rare, so if they only materialized as minor performance hits then it would probably be fine to ignore them. Unfortunately, though, not all transaction aborts are retried transparently in Cockroach, some are forced to return all the way to the client. Because of this, the false-positives could actually result in user-visible anomalies, which we should be avoiding as diligently as possible.

andy-kimball (approved):
So before, we only allowed the "adding" goroutine to init. Now we also allow another concurrent "adding" goroutine to init as well, if it's adding the "to" node of a range. This should be safe, because we still have a guarantee that the node is added *before* we go find the gap timestamp. So we won't run into the same race issue that caused us to restrict init'ing to the adding node in the first place.

In terms of implementation, it appears that you restored something similar to the original ratchetValueSet logic to do this (i.e. restored the setInit flag), and also refactored how some of the other methods work to be more straightforward/reusable. I looked over the logic carefully, and couldn't see any holes. There are a lot of changes, however, and it's easy to get some tiny detail wrong, so just make sure to run heavy stress for a while. I also like your idea to run Jepsen as well.

tschottdorf:
Conceptually, I think I understand the change and how it was implemented. I wasn't able to give this as thorough a review as the first version of the code (there were enough changes so that I would have to spend a few hours piecing it all together again), but with the testing as strong as it is now I'm confident that you didn't regress (after all, all bugs were exposed by tests, which have only improved). :lgtm: 

---

Reviewed 2 of 2 files at r1.

---

*[pkg/storage/tscache/interval_skl.go, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4aEjJF9ojClJoo8_w:-Kz4aEjJF9ojClJoo8_x:b-khoafi) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L77)):*
> ```Go
> 
> 	// cantInit indicates that the node should never be allowed to initialize.
> 	// This is set on nodes who were unable to ratchet their values at some
> ```

`nodes which`?

---

*[pkg/storage/tscache/interval_skl.go, line 510 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4abMNe3-Y0JVD7rjp:-Kz4abMOBbD61PWFTrsI:bx84bfb) ([raw file](https://github.com/cockroachdb/cockroach/blob/4e89fd3b5ff0c8fd6f5b7d047005635cf478367f/pkg/storage/tscache/interval_skl.go#L510)):*
> ```Go
> 
> 	if !it.SeekForPrev(key) {
> 		util.RacePreempt()
> ```

Any method to removing the `RacePreempt` calls?

---

*[pkg/storage/tscache/interval_skl.go, line 633 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4cpZ2USHc4wB0jkNh:-Kz4cpZ2USHc4wB0jkNi:bslci3t) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L633)):*
> ```Go
> 
> 	// Make sure we're on the right key again.
> 	if !bytes.Equal(it.Key(), key) {
> ```

Want to gate this on `util.RaceEnabled`?

---

*[pkg/storage/tscache/interval_skl.go, line 637 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4ct0kMMAAzctLAcvG:-Kz4ct0kMMAAzctLAcvH:b-j3f5g4) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L637)):*
> ```Go
> 	}
> 
> 	// If the node isn't initialized, do so.
> ```

s/do so/initialize it/?

---

*[pkg/storage/tscache/interval_skl.go, line 641 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4l6e2knaZ1U6cwGJA:-Kz4l6e2knaZ1U6cwGJB:b-bht4gt) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L641)):*
> ```Go
> }
> 
> func (p *sklPage) ensureFloorValue(it *arenaskl.Iterator, to []byte, val cacheValue) bool {
> ```

Comment on the return value.

---

*[pkg/storage/tscache/interval_skl.go, line 716 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4d8nuR-kZXaYeiimC:-Kz4d8nuR-kZXaYeiimD:biqgtd9) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L716)):*
> ```Go
> 	// always is a policy to ratchet a node regardless of whether it is already
> 	// initialized or not.
> 	always ratchetPolicy = false
> ```

Pull this section to the top. It's used above already.

---

*[pkg/storage/tscache/interval_skl.go, line 736 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4kf3zUPnDdu7zOWiY:-Kz4kf3zUPnDdu7zOWiZ:bb0ky07) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L736)):*
> ```Go
> 	var arr [encodedValSize * 2]byte
> 
> 	for {
> ```

Are we sure this for loop is preemptible?

---

*[pkg/storage/tscache/interval_skl.go, line 846 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4dkgrPCmENHoPrEFT:-Kz4dkgrPCmENHoPrEFU:bjr5egx) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L846)):*
> ```Go
> // maxInRange scans the range of keys between from and to and returns the
> // maximum value found. When finished, the iterator will be positioned the same
> // as if it.Seek(to) had been called.
> ```

Should mention how uninitialized nodes count.

---

*[pkg/storage/tscache/interval_skl.go, line 907 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4bZ60y1rre0PXl59q:-Kz4bZ60y1rre0PXl59r:be2ttxg) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L907)):*
> ```Go
> // function then returns the maximum value seen along with the gap value at the
> // end of the scan. If the iterator is positioned at a key > "to", the function
> // will return empty values. The function takes an optional initial gap value
> ```

s/empty/zero/

---

*[pkg/storage/tscache/interval_skl.go, line 910 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4bQDJoYtjpXfJ3y9s:-Kz4bQDJoYtjpXfJ3y9t:b-i7zekq) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L910)):*
> ```Go
> // argument, which is used to initialize the running maximum and gap values.
> // When finished, the iterator will be positioned the same as if it.Seek(to) had
> // been called.
> ```

Do you want to mention that all uninitialized nodes are ratcheted and init'ed?

---

*[pkg/storage/tscache/interval_skl_test.go, line 214 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4dxPMn_J1M94cjeBd:-Kz4dxPMn_J1M94cjeBe:bygzq97) ([raw file](https://github.com/cockroachdb/cockroach/blob/4e89fd3b5ff0c8fd6f5b7d047005635cf478367f/pkg/storage/tscache/interval_skl_test.go#L214)):*
> ```Go
> 	// val3:  <----------k]
> 	s.AddRange([]byte(""), []byte("kiwi"), 0, val3)
> 	require.Equal(t, val3, s.LookupTimestamp(nil))
> ```

Why this movement?

---

*[pkg/storage/tscache/interval_skl_test.go, line 979 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4e8zFB7No6aZEtOKw:-Kz4e8zFB7No6aZEtOKx:bt61csf) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl_test.go#L979)):*
> ```Go
> 
> func TestIntervalSklConcurrentVsSequential(t *testing.T) {
> 	// Run one subtest using a real clock to generate timestampts and
> ```

`s/timestampts/timestamps/`

---

*[pkg/storage/tscache/interval_skl_test.go, line 1053 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4eULDDHHsKThc7Vsy:-Kz4eULDDHHsKThc7Vsz:bqmz9va) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl_test.go#L1053)):*
> ```Go
> 
> 			// Ask each intervalSkl the same questions. We should get the same
> 			// answers.
> ```

Nice test. :thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-:-Kz4ebgsSRckgWBFcr_r:b-mneino)*

nvanbenschoten:
Thanks for the reviews! The tests are pretty solid at this point and I stressed them overnight without issue, so I'm pretty confident I'm not introducing any regressions. Right now I'm running Jepsen on this so it will be interesting to see if that finds anything.

---

---

*[pkg/storage/tscache/interval_skl.go, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4aEjJF9ojClJoo8_w:-Kz5BHfIZchMzKPNaVzE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L77)):*

`nodes which`?
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl.go, line 510 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4abMNe3-Y0JVD7rjp:-Kz5AyhOk6Hi8lm3E1Zv:b35f20v) ([raw file](https://github.com/cockroachdb/cockroach/blob/4e89fd3b5ff0c8fd6f5b7d047005635cf478367f/pkg/storage/tscache/interval_skl.go#L510)):*

Any method to removing the `RacePreempt` calls?
</blockquote></details>

I tried to push them down as far as possible into the helper functions that iterated over nodes and are susceptible to races. Here, `incomingGapVal` will already call `RacePreempt`.

---

*[pkg/storage/tscache/interval_skl.go, line 633 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4cpZ2USHc4wB0jkNh:-Kz59xKRR9ziP6RcGJz0:bombmv2) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L633)):*

Want to gate this on `util.RaceEnabled`?
</blockquote></details>

Good call. Done.

---

*[pkg/storage/tscache/interval_skl.go, line 637 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4ct0kMMAAzctLAcvG:-Kz59wRl3XHWU3VpFWQb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L637)):*

s/do so/initialize it/?
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl.go, line 641 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4l6e2knaZ1U6cwGJA:-Kz5AraOgHf2XnSGDPK-:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L641)):*

Comment on the return value.
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl.go, line 716 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4d8nuR-kZXaYeiimC:-Kz5DxdmJlKwyNeigTzW:b-dshanw) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L716)):*

Pull this section to the top. It's used above already.
</blockquote></details>

I feel like this makes sense to live next to `ratchetValueSet` since it's the only place that deals with these (other than passing them as arguments). I added a better comment to `ratchetPolicy` though.

---

*[pkg/storage/tscache/interval_skl.go, line 736 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4kf3zUPnDdu7zOWiY:-Kz5EctfqIUHEp8sYqFz:bl8ruwu) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L736)):*

Are we sure this for loop is preemptible?
</blockquote></details>

I'm not sure it is when `RacePreempt` is a no-op. I don't think it needs to be, though, since it's not waiting on any external condition to make forward progress.

---

*[pkg/storage/tscache/interval_skl.go, line 846 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4dkgrPCmENHoPrEFT:-Kz5F0ujxym8jd6Q3eQi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L846)):*

Should mention how uninitialized nodes count.
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl.go, line 907 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4bZ60y1rre0PXl59q:-Kz5F6nt0esnxddHJLzb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L907)):*

s/empty/zero/
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl.go, line 910 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4bQDJoYtjpXfJ3y9s:-Kz5Fxm-Zbsb7D1Zz3cp:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl.go#L910)):*

Do you want to mention that all uninitialized nodes are ratcheted and init'ed?
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl_test.go, line 214 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4dxPMn_J1M94cjeBd:-Kz53e8djsuVas3XClWz:b-eroj13) ([raw file](https://github.com/cockroachdb/cockroach/blob/4e89fd3b5ff0c8fd6f5b7d047005635cf478367f/pkg/storage/tscache/interval_skl_test.go#L214)):*

Why this movement?
</blockquote></details>

Range boundaries were a little under-specified/inconsistently-specified before. You brought this up in the previous review, but I don't think I sufficiently address it. I added another commit to do that.

---

*[pkg/storage/tscache/interval_skl_test.go, line 979 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz4e8zFB7No6aZEtOKw:-Kz53cZk-whJTZoCRfPR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/3675dd2cf0fe3cf5df5b38cea3741f52ef7e01f7/pkg/storage/tscache/interval_skl_test.go#L979)):*

`s/timestampts/timestamps/`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-:-Kz5BJM2HL6vUb03fYs0:bt699cl)*

tschottdorf:
Reviewed 1 of 1 files at r2, 2 of 2 files at r3.

---

*[pkg/storage/tscache/interval_skl.go, line 604 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz5N_TZrLJIJsU_NyyE:-Kz5N_TZrLJIJsU_NyyF:bukiphe) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ea8699884ecb098999b934af0aaa7aa54feb9e3/pkg/storage/tscache/interval_skl.go#L604)):*
> ```Go
> //    values. Scanning goroutines encounter the initializing nodes and are
> //    forced to ratchet them before continuing. If they fail to ratchet them
> //    because an arena is full, the nodes must never be initialized so it is set
> ```

s/it is/they are/

---

*[pkg/storage/tscache/interval_skl.go, line 932 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz5O5hET_ySl_nTs1Fh:-Kz5O5hET_ySl_nTs1Fi:b-l3y0bo) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ea8699884ecb098999b934af0aaa7aa54feb9e3/pkg/storage/tscache/interval_skl.go#L932)):*
> ```Go
> 
> // scanTo scans from the current iterator position until the key "to". While
> // scanning, any uninitialized values are ratchetted with the current gap value,
> ```

`ratcheted`?

---

*[pkg/storage/tscache/interval_skl_test.go, line 220 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz5OgadiGU3SKn7p4uM:-Kz5OgadiGU3SKn7p4uN:byy3y1w) ([raw file](https://github.com/cockroachdb/cockroach/blob/888a1e20c83b537121cb04a307316c8b6bb3d0f7/pkg/storage/tscache/interval_skl_test.go#L220)):*
> ```Go
> 	// val3:       [b----------->
> 	s.AddRange([]byte("banana"), nil, 0, val3)
> 	require.Equal(t, val1, s.LookupTimestamp([]byte("")))
> ```

Is it illegal to call `LookupTimestamp(nil)`? It seems that it could make sense to do so. If not, worth a comment here (and obviously on LookupTimestamp, though I didn't check there).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087)*

nvanbenschoten:

---

*[pkg/storage/tscache/interval_skl.go, line 604 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz5N_TZrLJIJsU_NyyE:-Kz5RbFSxEhzyRAcpdfj:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ea8699884ecb098999b934af0aaa7aa54feb9e3/pkg/storage/tscache/interval_skl.go#L604)):*

s/it is/they are/
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl.go, line 932 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz5O5hET_ySl_nTs1Fh:-Kz5Rd8kPJ0YZu3HoFBk:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/5ea8699884ecb098999b934af0aaa7aa54feb9e3/pkg/storage/tscache/interval_skl.go#L932)):*

`ratcheted`?
</blockquote></details>

Done.

---

*[pkg/storage/tscache/interval_skl_test.go, line 220 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-Kz5OgadiGU3SKn7p4uM:-Kz5TH-bpsTvhMpg6zmM:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/888a1e20c83b537121cb04a307316c8b6bb3d0f7/pkg/storage/tscache/interval_skl_test.go#L220)):*

Is it illegal to call `LookupTimestamp(nil)`? It seems that it could make sense to do so. If not, worth a comment here (and obviously on LookupTimestamp, though I didn't check there).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087)*

tschottdorf:
Still good.

---

Reviewed 2 of 2 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20087#-:-Kz5YEMO79RNKUzGwzTz:b-8kb5bg)*

PR #20088, opened 2017-11-16
distsqlrun: close orderedSynchronizer sources that were neglected

### LINKED ISSUE #19951: distsql: query hung after cancellation
Ran a long-running read query (i.e. `SELECT * FROM tpch.lineitem ORDER BY l_extendedprice`) on a two-node cluster and canceled. The query has been hung and an initial glance seems to indicate that the [noop processor](https://cockroachdb.github.io/distsqlplan/decode.html?eJzUkUFLw0AQhe_-CpmrI3SStGpOc-2llepNcojZoSykmbC7AaX0v0uyh9rSxApePO7MfO899u2hUSOrcice8jcgQEigQGidVuK9un4cj5bmA_IZgm3aLvTjAqFSJ5DvIdhQC-TwWr7XspHSiAMEI6G09SDcOrsr3SfXthEbZAcI6y7kt0zICXKKnCHPkRfID8iPyE_INEMmQqYEmVJkypBpjkwLKA4I2oVjEB_KrUBOB7w-7Iu6cJ6TF3ej4smo-FFTnREn5pLohQQrvdf25HbMOz3xpv_Uwg9h_66FC-Ib8a02Xq7641lfkZitxEq9dq6SZ6fVYBOf64EbBkZ8iFuKj2UTV33A7zBNwskJTOdwMgnPp53TSTibhrNfxS4ON18BAAD__y6EXuo=) on the first node has been cleaned up while the second node's sorter is still trying to push rows.

What leads me to believe this is that the second node is stuck on:
```
goroutine 583 [chan send, 9 minutes]:
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*RowChannel).Push(0xc42146d8c0, 0xc427096000, 0x10, 0x10, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/base.go:337 +0x1ba
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*ProcOutputHelper).EmitRow(0xc42018ad80, 0x6845140, 0xc4207beb10, 0xc423996300, 0x10, 0x10, 0x0, 0x0, 0x0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/processors.go:302 +0x2be
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*sortAllStrategy).executeImpl(0xc4203742e0, 0x6845140, 0xc4207beb10, 0xc42018ad80, 0x6846080, 0xc4200a2d80, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/sorterstrategy.go:150 +0x2bc
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*sortAllStrategy).Execute(0xc4203742e0, 0x6845140, 0xc4207beb10, 0xc42018ad80, 0x0, 0x0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/sorterstrategy.go:105 +0x537
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*sorter).Run(0xc42018ad80, 0x6845140, 0xc4207beb10, 0xc4207529b0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/sorter.go:145 +0x5d8
created by github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*Flow).Start
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/flow.go:400 +0x3d7
```

And the first node is stuck processing this (trying to push to the noop processor):
```
goroutine 157923 [chan send, 9 minutes]:
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*RowChannel).Push(0xc4214fdc80, 0xc424814a00, 0x10, 0x10, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/base.go:337 +0x1ba
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.processInboundStreamHelper(0x6845140, 0xc4214fdec0, 0x6851020, 0xc42030f050, 0x0, 0x6832d00, 0xc4214fdc80, 0xc4205b6a00, 0xc42f45cb00, 0x0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/inbound.go:118 +0x2dc
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.ProcessInboundStream(0x6845140, 0xc4214fdec0, 0x6851020, 0xc42030f050, 0xc42fb2ec60, 0x6832d00, 0xc4214fdc80, 0xc4205b6a00, 0x2540be400, 0xc4205b6a00)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/inbound.go:39 +0x9b
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*ServerImpl).flowStreamInt(0xc4200ceb00, 0x6845140, 0xc4214fdec0, 0x6851020, 0xc42030f050, 0x0, 0x0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/server.go:401 +0x428
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*ServerImpl).FlowStream(0xc4200ceb00, 0x6851020, 0xc42030f050, 0x683e280, 0xc4200ceb00)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/server.go:407 +0x91
github.com/cockroachdb/cockroach/pkg/sql/distsqlrun._DistSQL_FlowStream_Handler(0x5a657c0, 0xc4200ceb00, 0x684e620, 0xc420220400, 0x4014aa8, 0x20)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/api.pb.go:367 +0xb2
github.com/cockroachdb/cockroach/pkg/rpc.NewServerWithInterceptor.func2(0x5a657c0, 0xc4200ceb00, 0x684e620, 0xc420220400, 0xc42123e960, 0x5b97f20, 0x4040441, 0x261d8)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/pkg/rpc/context.go:190 +0x103
github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc.(*Server).processStreamingRPC(0xc42009ec80, 0x68506c0, 0xc424a9c160, 0xc4297de400, 0xc4202e7620, 0x6d27fa0, 0x0, 0x0, 0x0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc/server.go:940 +0x2e8
github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc.(*Server).handleStream(0xc42009ec80, 0x68506c0, 0xc424a9c160, 0xc4297de400, 0x0)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc/server.go:1027 +0x14c1
github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc423a87080, 0xc42009ec80, 0x68506c0, 0xc424a9c160, 0xc4297de400)
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc/server.go:572 +0x9f
created by github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc.(*Server).serveStreams.func1
	/Users/asubiotto/go/src/github.com/cockroachdb/cockroach/vendor/google.golang.org/grpc/server.go:570 +0xa1
```

With no indication of any processors' presence on the first (gateway) node.

A user has run into this issue in #19797, but we have not yet established if the cause is the same.

asubiotto:
These sources were neglected because they were not added to the heap
when a heap initialization error happened. The ordered synchronizer
would then go into a draining phase, but it would only close sources
that were added to the heap.

Fixes #19951

This commit also improves the testing of distsql query cancellation. The
issue that this commit fixes is caught by stress testing the new test.

Another oddity was uncovered by this: an ORDER BY DESC run after a SPLIT AT results in an ordering issue. I'll file an issue tomorrow morning and verify the minimum steps to reproduce.

I will also add orderedSynchronizer unit tests probably in another PR. Should this be cherrypicked into 1.1? I think so.

Release note (bug fix): fix a race condition that would result in some queries hanging after cancellation

### DIFF
--- a/pkg/sql/distsqlrun/input_sync.go
+++ b/pkg/sql/distsqlrun/input_sync.go
@@ -131,17 +131,31 @@ func (s *orderedSynchronizer) Pop() interface{} {
 	return nil
 }
 
-// initHeap grabs a row from each source and initializes the heap.
+// initHeap grabs a row from each source and initializes the heap. Any given
+// source will be on the heap (even if an error was encountered while reading
+// from it) unless there are no more rows to read from it.
+// If an error is returned, heap.Init() has not been called, so s.heap is not
+// an actual heap. In this case, all members of the heap need to be drained.
 func (s *orderedSynchronizer) initHeap() error {
+	// consumeErr is the last error encountered while consuming metadata.
+	var consumeErr error
 	for i := range s.sources {
 		src := &s.sources[i]
-		if err := s.consumeMetadata(src, stopOnRowOrError); err != nil {
-			return err
+		err := s.consumeMetadata(src, stopOnRowOrError)
+		if err != nil {
+			consumeErr = err
 		}
-		if src.row != nil {
+		// We add the source to the heap either if we have received a row from
+		// it or there was an error reading from this source. We still add to
+		// the heap in case of error so that these sources can be drained in
+		// `drainSources`.
+		if src.row != nil || err != nil {
 			// Add to the heap array (it won't be a heap until we call heap.Init).
 			s.heap = append(s.heap, srcIdx(i))
 		}
+	}
+	if consumeErr != nil {
+		return consumeErr
 	}
 	heap.Init(s)
 	// heap operations might set s.err (see Less)
@@ -292,35 +306,36 @@ func (s *orderedSynchronizer) Next() (sqlbase.EncDatumRow, ProducerMetadata) {
 func (s *orderedSynchronizer) ConsumerDone() {
 	// We're entering draining mode. Only metadata will be forwarded from now on.
 	if s.state != draining {
-		s.state = draining
-		s.consumerStatusChanged(RowSource.ConsumerDone)
+		s.consumerStatusChanged(draining, RowSource.ConsumerDone)
 	}
 }
 
 // ConsumerClosed is part of the RowSource interface.
 func (s *orderedSynchronizer) ConsumerClosed() {
-	// The state should matter, as no further methods should be called, but we'll
-	// set it to something other than the default.
-	s.state = drainBuffered
-	s.consumerStatusChanged(RowSource.ConsumerClosed)
+	// The state shouldn't matter, as no further methods should be called, but
+	// we'll set it to something other than the default.
+	s.consumerStatusChanged(drainBuffered, RowSource.ConsumerClosed)
 }
 
 // consumerStatusChanged calls a RowSource method on all the non-exhausted
 // sources.
-func (s *orderedSynchronizer) consumerStatusChanged(f func(RowSource)) {
+func (s *orderedSynchronizer) consumerStatusChanged(
+	newState orderedSynchronizerState, f func(RowSource),
+) {
 	if s.state == notInitialized {
 		for i := range s.sources {
 			f(s.sources[i].src)
 		}
 	} else {
-		// The sources that are not in the heap have been consumed already. It would
-		// be ok to call ConsumerDone()/ConsumerClosed() on them too, but avoiding
-		// the call may be a bit faster (in most cases there should be no sources
-		// left).
+		// The sources that are not in the heap have been consumed already. It
+		// would be ok to call ConsumerDone()/ConsumerClosed() on them too, but
+		// avoiding the call may be a bit faster (in most cases there should be
+		// no sources left).
 		for _, sIdx := range s.heap {
 			f(s.sources[sIdx].src)
 		}
 	}
+	s.state = newState
 }
 
 func makeOrderedSync(
--- a/pkg/sql/run_control_test.go
+++ b/pkg/sql/run_control_test.go
@@ -17,6 +17,8 @@ package sql_test
 import (
 	gosql "database/sql"
 	"errors"
+	"fmt"
+	"math/rand"
 	"strings"
 	"testing"
 	"time"
@@ -26,9 +28,11 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/sql"
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/cockroachdb/cockroach/pkg/testutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
 	"github.com/cockroachdb/cockroach/pkg/testutils/sqlutils"
 	"github.com/cockroachdb/cockroach/pkg/util/leaktest"
+	"github.com/cockroachdb/cockroach/pkg/util/timeutil"
 )
 
 func TestCancelSelectQuery(t *testing.T) {
@@ -175,15 +179,20 @@ func TestCancelParallelQuery(t *testing.T) {
 	}
 }
 
-// Cancel a distSQL query pre-execution (before any streams have been established)
+// TestCancelDistSQLQuery runs a distsql query and cancels it randomly at
+// various points of execution.
 func TestCancelDistSQLQuery(t *testing.T) {
 	defer leaktest.AfterTest(t)()
-	const queryToCancel = "SELECT * FROM nums ORDER BY num DESC"
+	const queryToCancel = "SELECT * FROM nums ORDER BY num"
+	cancelQuery := fmt.Sprintf("CANCEL QUERY (SELECT query_id FROM [SHOW CLUSTER QUERIES] WHERE query = '%s')", queryToCancel)
 
 	// conn1 is used for the query above. conn2 is solely for the CANCEL statement.
 	var conn1 *gosql.DB
 	var conn2 *gosql.DB
 
+	var queryLatency *time.Duration
+	sem := make(chan struct{}, 1)
+	rng := rand.New(rand.NewSource(timeutil.Now().UnixNano()))
 	tc := serverutils.StartTestCluster(t, 2, /* numNodes */
 		base.TestClusterArgs{
 			ReplicationMode: base.ReplicationManual,
@@ -191,14 +200,15 @@ func TestCancelDistSQLQuery(t *testing.T) {
 				UseDatabase: "test",
 				Knobs: base.TestingKnobs{
 					SQLExecutor: &sql.ExecutorTestingKnobs{
-						BeforeExecute: func(ctx context.Context, stmt string, _ /* isParallel */ bool) {
-							// if queryToCancel
-							if strings.Contains(stmt, "ORDER BY num") {
-								// Cancel it
-								const cancelQuery = "CANCEL QUERY (SELECT query_id FROM [SHOW CLUSTER QUERIES] WHERE node_id = 1)"
-								if _, err := conn2.Exec(cancelQuery); err != nil {
-									t.Fatal(err)
-								}
+						BeforeExecute: func(_ context.Context, stmt string, _ /* isParallel */ bool) {
+							if strings.HasPrefix(stmt, queryToCancel) {
+								// Wait for the race to start.
+								<-sem
+							} else if strings.HasPrefix(stmt, cancelQuery) {
+								// Signal to start the race.
+								sleepTime := time.Duration(rng.Int63n(int64(*queryLatency)))
+								sem <- struct{}{}
+								time.Sleep(sleepTime)
 							}
 						},
 					},
@@ -211,15 +221,47 @@ func TestCancelDistSQLQuery(t *testing.T) {
 	conn2 = tc.ServerConn(1)
 
 	sqlutils.CreateTable(t, conn1, "nums", "num INT", 0, nil)
-	_, err := conn1.Exec("INSERT INTO nums SELECT generate_series(1,10)")
-	if err != nil {
+	if _, err := conn1.Exec("INSERT INTO nums SELECT generate_series(1,100)"); err != nil {
 		t.Fatal(err)
 	}
 
-	_, err = conn1.Exec(queryToCancel)
-	if err != nil && !sqlbase.IsQueryCanceledError(err) {
+	if _, err := conn1.Exec("ALTER TABLE nums SPLIT AT VALUES (50)"); err != nil {
 		t.Fatal(err)
-	} else if err == nil {
-		t.Fatal("didn't get an error from query that should have been canceled")
+	}
+	// Make the second node the leaseholder for the first range to distribute
+	// the query.
+	if _, err := conn1.Exec(fmt.Sprintf(
+		"ALTER TABLE nums TESTING_RELOCATE VALUES (ARRAY[%d], 1)",
+		tc.Server(1).GetFirstStoreID(),
+	)); err != nil {
+		t.Fatal(err)
+	}
+
+	// Run queryToCancel to be able to get an estimate of how long it should
+	// take. The goroutine in charge of cancellation will sleep a random
+	// amount of time within this bound. Signal sem so that it can run
+	// unhindered.
+	sem <- struct{}{}
+	start := timeutil.Now()
+	if _, err := conn1.Exec(queryToCancel); err != nil {
+		t.Fatal(err)
+	}
+	execTime := timeutil.Since(start)
+	queryLatency = &execTime
+
+	errChan := make(chan error)
+	go func() {
+		_, err := conn1.Exec(queryToCancel)
+		errChan <- err
+	}()
+	_, err := conn2.Exec(cancelQuery)
+	if err != nil && !testutils.IsError(err, "query ID") {
+		t.Fatal(err)
+	}
+	// Successful cancellation.
+	// Note the err != nil check. It exists because a successful cancellation
+	// does not imply that the query was canceled.
+	if err := <-errChan; err != nil && !sqlbase.IsQueryCanceledError(err) {
+		t.Fatal(err)
 	}
 }


andreimatei:
don't forget release notes

I've don't have much of an idea what's going on here. If there's an error while initializing the heap, doesn't the state stay `notInitialized`, in which case we drain all the sources? Sorry, I tried reading your discussion on the bug but it's long and I lost the point :)
Let's talk in person.

---

---

*[pkg/sql/distsqlrun/inbound.go, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-Kz4rqkQZGTV_LKc0fik:-Kz4rqkQZGTV_LKc0fil:b2i7m3b) ([raw file](https://github.com/cockroachdb/cockroach/blob/9dba327104472330893f8fdcd61accf0056f0655/pkg/sql/distsqlrun/inbound.go#L40)):*
> ```Go
> 	// err, if set, will also be propagated to the producer
> 	// as the last record that the producer gets.
> 	if err := processInboundStreamHelper(ctx, stream, firstMsg, dst, f); err != nil {
> ```

nit: spurious change :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-:-Kz4rhZry-7ZXA2Ba4Nm:b-fhtx3y)*

RaduBerinde:

---

*[pkg/sql/distsqlrun/input_sync.go, line 270 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-Kz5we3MV-WGst1OjMu1:-Kz5we3MV-WGst1OjMu2:bsfjxak) ([raw file](https://github.com/cockroachdb/cockroach/blob/9dba327104472330893f8fdcd61accf0056f0655/pkg/sql/distsqlrun/input_sync.go#L270)):*
> ```Go
> 		// Next() calls will return metadata records.
> 		s.drainSources()
> 		// Call ConsumerClosed to primarily take care of sources that were not
> ```

This suggests that some sources may get closed some other way. Is it guaranteed that we won't call ConsumerClosed on them twice?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

RaduBerinde:
*[pkg/sql/distsqlrun/input_sync.go, line 270 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-Kz5we3MV-WGst1OjMu1:-Kz5xv_9FLpV5IqTZfHf:byi7a4e) ([raw file](https://github.com/cockroachdb/cockroach/blob/9dba327104472330893f8fdcd61accf0056f0655/pkg/sql/distsqlrun/input_sync.go#L270)):*

This suggests that some sources may get closed some other way. Is it guaranteed that we won't call ConsumerClosed on them twice?
</blockquote></details>

In the bug you say "add a call to ConsumerClosed after the usual draining code. This will be a noop on streams that have already had ConsumerClosed called on them and will correctly close uninitialized streams" but the RowChannel doesn't tolerate that: https://github.com/cockroachdb/cockroach/blob/fc8a0ce3e4f4d716e915218a5c81c1a2f48a37b0/pkg/sql/distsqlrun/base.go#L381

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

asubiotto:
TFTR, added release notes.

The state does not stay `notInitialized`, because the `ConsumerDone` call in the if statement handling a heap initialization error switches the state to `draining`.

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[pkg/sql/distsqlrun/inbound.go, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-Kz4rqkQZGTV_LKc0fik:-KzyBUhYOhuVmmn5_fh4:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/9dba327104472330893f8fdcd61accf0056f0655/pkg/sql/distsqlrun/inbound.go#L40)):*

nit: spurious change :)
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/input_sync.go, line 270 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-Kz5we3MV-WGst1OjMu1:-KzyEX3WrnQoKE5NfhsY:be8a75e) ([raw file](https://github.com/cockroachdb/cockroach/blob/9dba327104472330893f8fdcd61accf0056f0655/pkg/sql/distsqlrun/input_sync.go#L270)):*

In the bug you say "add a call to ConsumerClosed after the usual draining code. This will be a noop on streams that have already had ConsumerClosed called on them and will correctly close uninitialized streams" but the RowChannel doesn't tolerate that: https://github.com/cockroachdb/cockroach/blob/fc8a0ce3e4f4d716e915218a5c81c1a2f48a37b0/pkg/sql/distsqlrun/base.go#L381
</blockquote></details>

Sorry, I'm not quite sure what I meant with what I wrote in the bug. You're right that calling `ConsumerClosed` twice will result in a panic, however it is guaranteed that we will not call `ConsumerClosed` on a stream twice. This is because once `ConsumerClosed` is called on the ordered synchronizer, it enters the `drainBuffered` state (I removed the state change below the `ConsumerClosed` call since it was unnecessary). Since we never transition from `drainBuffered` to `draining`, it's impossible to execute this code once `ConsumerClosed` has been called. This assumes that an ordered synchronizer's input streams do not get closed out from under it. In the case where `ConsumerClosed` is called after `Next`, it is specified in the `RowSource` interface that calling `ConsumerClosed` once `Next` returns an empty row is a noop. I added a state check in `ConsumerClosed` for this case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-:-KzyBDbcSa5EoM_0_klY:b5rnk9t)*

asubiotto:
PTAL when you have a chance.

asubiotto:
Updated the PR with an alternative approach suggested by @andreimatei: Add sources to the heap even on error in `initHeap`, this will result in proper draining of these sources.

andreimatei:
:lgtm:

---

---

*[pkg/sql/run_control_test.go, line 245 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-bZPHG8P5gpLufc0Qu:-L-bZPHG8P5gpLufc0Qv:b95kooc) ([raw file](https://github.com/cockroachdb/cockroach/blob/e074e232a91587a491cc1e5c5ebc40cb5459669b/pkg/sql/run_control_test.go#L245)):*
> ```Go
> 	// unhindered.
> 	go func() {
> 		sem <- struct{}{}
> ```

I think if `sem` was buffered (as semaphores implemented using this channel pattern usually are), you wouldn't need to do this in a separate goroutine (which is confusing).

---

*[pkg/sql/distsqlrun/input_sync.go, line 136 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-bX9lksOkzx9iLR5u6:-L-bX9llGqgByN-wbesf:b-xihwz9) ([raw file](https://github.com/cockroachdb/cockroach/blob/e074e232a91587a491cc1e5c5ebc40cb5459669b/pkg/sql/distsqlrun/input_sync.go#L136)):*
> ```Go
> // initHeap grabs a row from each source and initializes the heap. Any given
> // source will be on the heap (even if an error was encountered while reading
> // from it) unless there are no more rows to read from it.
> ```

please add a comment saying that if an error is returned, `heap.Init()` has not been called, so `heap` is not actually a heap. And also spell out that if an error is returned, all the members of the heap need to be drained.

---

*[pkg/sql/distsqlrun/input_sync.go, line 309 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-bY3Gy61kEEf2oMkfx:-L-bY3Gy61kEEf2oMkfy:be29p0v) ([raw file](https://github.com/cockroachdb/cockroach/blob/e074e232a91587a491cc1e5c5ebc40cb5459669b/pkg/sql/distsqlrun/input_sync.go#L309)):*
> ```Go
> // ConsumerClosed is part of the RowSource interface.
> func (s *orderedSynchronizer) ConsumerClosed() {
> 	// The state should matter, as no further methods should be called, but we'll
> ```

not your code, but I think this comment wants to say "_shouldn't_ matter"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-:-L-bZfdR60jLtVgjuHS2:bnfp4nl)*

asubiotto:

---

*[pkg/sql/run_control_test.go, line 245 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-bZPHG8P5gpLufc0Qu:-L-bmdDBP5mOCHX_oZK4:b-yh16l6) ([raw file](https://github.com/cockroachdb/cockroach/blob/e074e232a91587a491cc1e5c5ebc40cb5459669b/pkg/sql/run_control_test.go#L245)):*

I think if `sem` was buffered (as semaphores implemented using this channel pattern usually are), you wouldn't need to do this in a separate goroutine (which is confusing).
</blockquote></details>

Done. Good call.

---

*[pkg/sql/distsqlrun/input_sync.go, line 136 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-bX9lksOkzx9iLR5u6:-L-bmecyvJqKtAiM_-2p:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e074e232a91587a491cc1e5c5ebc40cb5459669b/pkg/sql/distsqlrun/input_sync.go#L136)):*

please add a comment saying that if an error is returned, `heap.Init()` has not been called, so `heap` is not actually a heap. And also spell out that if an error is returned, all the members of the heap need to be drained.
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/input_sync.go, line 309 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-bY3Gy61kEEf2oMkfx:-L-bmfX28zkreTZVnsNb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e074e232a91587a491cc1e5c5ebc40cb5459669b/pkg/sql/distsqlrun/input_sync.go#L309)):*

not your code, but I think this comment wants to say "_shouldn't_ matter"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

RaduBerinde:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[pkg/sql/distsqlrun/input_sync.go, line 148 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-cEYc_5wsPv7lF11-S:-L-cEYc_5wsPv7lF11-T:b-rlm7ob) ([raw file](https://github.com/cockroachdb/cockroach/blob/381b57cae3a832bf4f3b85cb779cfdf8877eb61a/pkg/sql/distsqlrun/input_sync.go#L148)):*
> ```Go
> 			consumeErr = err
> 		}
> 		if src.row != nil || err != nil {
> ```

was this supposed to be &&? (in which case it could be an `else if src.row != nil`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

asubiotto:

---

*[pkg/sql/distsqlrun/input_sync.go, line 148 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-cEYc_5wsPv7lF11-S:-L-cPbuNA-3WW2VoNbMX:blk7tze) ([raw file](https://github.com/cockroachdb/cockroach/blob/381b57cae3a832bf4f3b85cb779cfdf8877eb61a/pkg/sql/distsqlrun/input_sync.go#L148)):*

was this supposed to be &&? (in which case it could be an `else if src.row != nil`
</blockquote></details>

This is supposed to be `||` as I want to add to the heap either if the src has a row available or there was en error while consuming its metadata.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

RaduBerinde:

---

*[pkg/sql/distsqlrun/input_sync.go, line 148 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-cEYc_5wsPv7lF11-S:-L-cQVgI74qRr-QHpUa2:bb33esj) ([raw file](https://github.com/cockroachdb/cockroach/blob/381b57cae3a832bf4f3b85cb779cfdf8877eb61a/pkg/sql/distsqlrun/input_sync.go#L148)):*

This is supposed to be `||` as I want to add to the heap either if the src has a row available or there was en error while consuming its metadata.
</blockquote></details>

Oh. What's going on here is pretty subtle and needs more comments. Explain that we're still adding the source just so it's in `s.heap` for cleanup purposes (see `consumerStatusChanged`), we won't actually initialize the heap with a nil row.

In fact, I'm still missing something.. `initHeap` is only called in `notInitialized` state and if it returns an error, we will remain in that state and `consumerStatusChanged` won't look at the heap.


Instead of this, maybe we should add a `needsClosing util.FastIntSet` with the set of sources that weren't closed. `consumerStatusChanged` would just consult this set instead of behaving differently based on `state`. What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

asubiotto:

---

*[pkg/sql/distsqlrun/input_sync.go, line 148 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-L-cEYc_5wsPv7lF11-S:-L-h0H0Wu8K0XiWTHXmz:b-7s99gw) ([raw file](https://github.com/cockroachdb/cockroach/blob/381b57cae3a832bf4f3b85cb779cfdf8877eb61a/pkg/sql/distsqlrun/input_sync.go#L148)):*

Oh. What's going on here is pretty subtle and needs more comments. Explain that we're still adding the source just so it's in `s.heap` for cleanup purposes (see `consumerStatusChanged`), we won't actually initialize the heap with a nil row.

In fact, I'm still missing something.. `initHeap` is only called in `notInitialized` state and if it returns an error, we will remain in that state and `consumerStatusChanged` won't look at the heap.


Instead of this, maybe we should add a `needsClosing util.FastIntSet` with the set of sources that weren't closed. `consumerStatusChanged` would just consult this set instead of behaving differently based on `state`. What do you think?
</blockquote></details>

There are two `Next()` calls to consider:

1) The first calls `initHeap` and gets back an error, running `ConsumerDone()`, as you rightly pointed out, on all the sources. This call to `consumerStatusChanged` also changes the synchronizer's state to `draining`. The error is returned through metadata, so the downstream component will call Next() again.
2) The second call will then proceed to the `draining` conditional in `Next()`, this is the work that actually drains the sources in `drainSources`. The `drainSources` method only touches the heap though, so does not consume any data from sources not in the heap.

The `needsClosing` idea is good but I think is effectively implemented with this change. I don't want to add any additional state in this PR.

Added a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088)*

RaduBerinde:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20088#-:-L-h6HKZSUO4RIl_iJwe:bnfp4nl)*

PR #17214, opened 2017-07-25
distsql: basic planning for merge joins

RaduBerinde:
We detect when the two sides of the join have matching orderings on all equality
columns and use the merge joiner.

Note that as things stand, the join node does not guarantee any output ordering.
This means that we may have an unnecessary sorter after a merge join, or - much
more importantly - the output of a merge join won't be used as the input of
another merge join.

Updates #16580.

I have not looked at the running time of a merge join vs hash join on a large dataset (TPC-H). I was hoping @arjunravinarayan could help with that. Even a single-node deployment would give us a good idea of the speed difference.

Issue #10432, opened 2016-11-03
distsql: Set operations (UNION, INTERSECT, EXCEPT)


irfansharif:
Implementation of a processor "core" type from #7587. 
Set operations would include `UNION`, `UNION ALL`, `INTERSECT`, `INTERSECT ALL`, `EXCEPT`, `EXCEPT ALL`.

This processor, like join processors, would have two input streams and one output stream. Some of existing processor core type implementations under `pkg/sql/distsql` and spec definitions in `pkg/sql/distsql/processors.proto` can be used for reference.

a6802739:
@irfansharif , I could have a look on it.

irfansharif:
@a6802739: go for it! let me know if you need any pointers.

a6802739:
@irfansharif , okay, thanks a lot. At first, I'll have a see on it.

a6802739:
@irfansharif,  should I add a `Message SetSpec` in `pkg/sql/distsql/processors.proto`, and generate a new struct `Seter` under `pkg/sql/distsql`?

how could I associate struct `Seter` to the `input stream` and `out stream`, And how could I specify the input `colum` we are going to process?
should I just implement the interface `processor`?

irfansharif:
hey @a6802739,

for the spec definition going with `SetSpec` would be best, this would internally specify which operation to apply (`UNION, INTERSECT, EXCEPT`). As for the struct itself going with a simple `type set struct { ... }` should suffice (under `pkg/sql/distsql/set.go`). 

> And should I just translate `sql/union.go` to `distsql/union.go`?

I would the implementations would function similarly but keep in mind under `distsql` we're operating under the abstraction that we simply operate on _streams_. There's also a lot of additional baggage in `sql/union.go` given our current structure constructing a `planNode` to evaluate queries which we don't need to satisfy here. 

As far as your questions concerning `aggregator.go`, feel free to ping me on [gitter](https://gitter.im/cockroachdb/cockroach) @irfansharif and will be happy to walk you through it there just for the sake of not polluting this thread. for reference I would advise you to look at `distinct.go` as opposed to `aggregator.go`, it's more pertinent to `Set` as compared to `aggregator` responsible for grouping aggregators (`SUM`, `AVG`, etc).

irfansharif:
cc @arjunravinarayan, given you expressed interest in taking this on as well.

RaduBerinde:
This issue has two parts:
 - physical processor implementation for set operations
[...]

PR #20090, opened 2017-11-16
distsqlrun: Fix MergeJoiner correctness issue with NULL values

lgo:
Previously, when a distSQL merge join was being executed the comparison
check was using Datum.Compare while processing the incoming streams we
are joining on.

This is problematic when both sides of the stream have a NULL value,
because this comparison is considered equal which is a violation of the
semantics of equals (=) with SQL NULLs.

Unfortunately, this bug was present before distSQL merge joins were
being planned (#17214) and so it exists in 1.1.0.

### DIFF
--- a/pkg/sql/distsqlrun/stream_merger.go
+++ b/pkg/sql/distsqlrun/stream_merger.go
@@ -106,6 +106,13 @@ func CompareEncDatumRowForMerge(
 	for i, ord := range leftOrdering {
 		lIdx := ord.ColIdx
 		rIdx := rightOrdering[i].ColIdx
+		// If both datums are NULL, we need to follow SQL semantics where
+		// they are not equal. This differs from our datum semantics where
+		// they are equal.
+		if lhs[lIdx].IsNull() && rhs[rIdx].IsNull() {
+			// We can return either -1 or 1, it does not change the behavior.
+			return -1, nil
+		}
 		cmp, err := lhs[lIdx].Compare(&lhsTypes[lIdx], da, evalCtx, &rhs[rIdx])
 		if err != nil {
 			return 0, err
--- a/pkg/sql/logictest/testdata/logic_test/distsql_join
+++ b/pkg/sql/logictest/testdata/logic_test/distsql_join
@@ -194,3 +194,48 @@ query T
 SELECT "URL" FROM [EXPLAIN (DISTSQL) (SELECT * FROM (SELECT a,b from data AS data3 NATURAL JOIN ((SELECT a,b FROM data AS data1) JOIN (SELECT c,d FROM data AS data2 ORDER BY c,d) ON a=c AND b=d)))]
 ----
 https://cockroachdb.github.io/distsqlplan/decode.html?eJzsmE1v4kgQhu_7K6I67Sq9Et02H0FaydesNMkoM7cRBwf3ECRCo7aRJory30dAZhi7Sb1U2kECcQT8uKor7cdv-pnmrrA3-aMtafiNNCkypCghRSkp6tJI0cK7sS1L51eXbIDr4gcNO4qm88WyWn09UjR23tLwmappNbM0pK_5_cze2bywnhQVtsqns3WRhZ8-5v4pK_IqJ0W3y2p4kWmVGZUlKktp9KLILavXO29veP908ZCXD_Wb1cGRorLKJ5aG-kV9ZKvvaLLWnvm49nbMcFs3kdT94nzVLJnpS5WZy9j1p2_2sb2V84X1tniz_t5X7ljaJ-sn9n83nTfXN7Pfq79f0X_-89PJw_bjh23WbuQ0VJZcqiwVTOUP4r3T-XWLxpR-f809L9ul9_ZY-nK-ayk7O79x_7pF47Ldhfu1wvp4XNZ6q-26rNX2BC4DdQ_mMn12WdQ0TsZl5niU0nqr7Sql1fYESgF1D6YUc1ZK1DRORinJ8Sil9VbbVUqr7QmUAuoeTCnJWSlR0zgZpaTHo5TWW21XKa22J1AKqHswpaRnpURN42SUAk4V72y5cPPS7nVK01mtzBYTu5lj6ZZ-bD97N16X2Xy8XXPr_zULW1abX7ubD9fzzU-rBveHtYmiBzG0SWPopMPTmqVBaR7WOorux9AmiaKveNo06U5t5DW404QTwcSNDG5MXEr3Y-jGxKX0FU-ngmdbCDeebSk9iKEN-HPzdOPZDuguu017_B7v8Xtc85u8H-NiHkYuBjRwMU8jF_M0cvEgxsU8jFwMaOBinkYuBjRw8RW7T3WH36eaf3uCxxPQSMcIBz4GOBIywoGRNf8KBUoGNHIywoGUAY6sDHCkZc3HB-Blzb9IgVoBjdyKcCBXgCO7AhxGXf5tioqDHIDCLsBR2gU5AsVdgAPHaj5J6B6QbJAlRJLlaShZgCPJ8jiULMCRZCU5SkpDyYqSlBSHkhVlqRAPUoVIskGqEEmWp6FkAY4ky-NQsjyOJGskgUpKI8kiHEgW4EiyCEeHCkGqqO1YY3jJmiBVSCQLaCRZhAPJAhxJFuFAskaSqKQ0kizCgWQBjiQLcCRZE8QKiWRNkCokkgU0kizCgWQBjiQLcChZSaCS0lCyokAlxaFkRYEqxINUUZfsAEhWckQTPi6iMxoxjiQrOqUR40iykkQlpaFkRYlKikPJihJVeHAexApWsqOXv34GAAD___sgIII=
+
+
+# Test that the distSQL MergeJoiner follows SQL NULL semantics for ON predicate equivilance.
+# The use of sorts here force
+
+statement ok
+CREATE TABLE distsql_mj_test (k INT, v INT)
+
+statement ok
+INSERT INTO distsql_mj_test VALUES (0, NULL), (0, 1), (2, 4), (NULL, 4)
+
+# If SQL NULL semantics are not followed, NULL = NULL is truthy. This makes the rows with NULL also appear in the inner join.
+
+query IIII rowsort
+SELECT l.k, l.v, r.k, r.v FROM (SELECT * FROM distsql_mj_test ORDER BY k, v) l INNER JOIN (SELECT * FROM distsql_mj_test ORDER BY k, v) r ON l.k = r.k AND l.v = r.v
+----
+0  1  0  1
+2  4  2  4
+
+statement ok
+DELETE FROM distsql_mj_test WHERE TRUE;
+
+statement ok
+INSERT INTO distsql_mj_test VALUES (0, NULL), (1, NULL), (2, NULL)
+
+# We should not have any results for values with NULLs
+query IIII rowsort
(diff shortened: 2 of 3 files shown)


lego:
On a side note: this behavior we might like for an anti join, and in particular the scrub checks I've been working on ;). In the (near) future I expect I'll plumb a flag into here in order to make NULL comparisons truth-y.

pkg/sql/distsqlrun/stream_merger.go
```diff
@@ -107,6 +107,13 @@ func CompareEncDatumRowForMerge(
 		lIdx := ord.ColIdx
 		rIdx := rightOrdering[i].ColIdx
 		cmp, err := lhs[lIdx].Compare(&lhsTypes[lIdx], da, evalCtx, &rhs[rIdx])
+		// If the datums are both DNull, we need to follow SQL semantics
+		// where they are not equal. This differs from our datum semantics
+		// where they are equal.
+		if lhs[lIdx].Datum == tree.DNull && rhs[rIdx].Datum == tree.DNull {
+			// We can return either -1 or 1, it does not change the behavior.
+			return -1, nil
+		}
```
vivekmenezes: perhaps move this to before we call Compare()

lego: I should have mentioned this in the comment, but Compare is on two EncDatums so we don't yet know if the Datums are decoded (I'm assuming), so this happens specifically after.

lego:
Agh, the tests failed. The failure doesn't seem to be deterministic as they were working earlier, which is a bit worrisome. I'll look at it in the morning.

RaduBerinde:
Please file an issue to push down `NOT NULL` constraints on equality columns during planning (depending on join type). This can reduce the scan ranges.

---

---

*[pkg/sql/distsqlrun/stream_merger.go, line 116 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-Kz1aEywOoiXpLpPFJdr-r2-116:-Kz3dbhoKbrtGpINCsdP:b-dzgolh) ([raw file](https://github.com/cockroachdb/cockroach/blob/263c1edaa3169fa57eb10253c9fcaaf07c4fee64/pkg/sql/distsqlrun/stream_merger.go#L116)):*

I should have mentioned this in the comment, but Compare is on two EncDatums so we don't yet know if the Datums are decoded (I'm assuming), so this happens specifically after.
</blockquote></details>

Compare doesn't necessarily decode them (it could compare key encodings directly). You can use `EncDatum.IsNull`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-:-Kz3dyQUqejqcyXym3Re:b-uei7dy)*

lego:
Ah, I wasn't sure what you were referring to previously. The issue is referenced above this comment.

---

---

*[pkg/sql/distsqlrun/stream_merger.go, line 116 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-Kz1aEywOoiXpLpPFJdr-r2-116:-Kz4awoPxeN2x4szcXUN:b-emfdxo) ([raw file](https://github.com/cockroachdb/cockroach/blob/263c1edaa3169fa57eb10253c9fcaaf07c4fee64/pkg/sql/distsqlrun/stream_merger.go#L116)):*

Compare doesn't necessarily decode them (it could compare key encodings directly). You can use `EncDatum.IsNull`.
</blockquote></details>

Awesome, also moved this above the `Compare()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-:-Kz4ar3n6NxquhLYDt6z:bquxx6a)*

lego:
It seems like `algebraicSetOp` also uses the compare function that is being changed, but the code path that reaches it is never used (`algebraicSetOp.Run()` -> `algebraicSetOp.exceptAll()` -> `CompareEncDatumRowForMerge`). I'm guesisng this isn't an issue as it's not used, @RaduBerinde?

RaduBerinde:
I'm not sure how algebraicSetOp is supposed to work with NULLs, CC @arjunravinarayan

lego:

---

*[pkg/sql/distsqlrun/mergejoiner_test.go, line 359 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-Kz5CupqYWLD3kulyWZC:-Kz5CuprVRTrZFx7XEEG:b-cj7qpb) ([raw file](https://github.com/cockroachdb/cockroach/blob/69d9a4091a30702aaf4147c2ef5e0e04918d1692/pkg/sql/distsqlrun/mergejoiner_test.go#L359)):*
> ```Go
> 			},
> 		},
> 		{
> ```

I added these tests which should test this problem, but they seem to pass prior to the fix.

As far as I know, this should still be using the same logic in from the streamMerger, so I'm not sure why it is still working with the bug. Could it maybe be how these `null` are interpreted? I dug a bit but didn't find an answer. @RaduBerinde would you happen to know why?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090)*

RaduBerinde:
*[pkg/sql/distsqlrun/mergejoiner_test.go, line 359 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-Kz5CupqYWLD3kulyWZC:-Kz5FakaQPjgteR11ArA:b-q1me9h) ([raw file](https://github.com/cockroachdb/cockroach/blob/69d9a4091a30702aaf4147c2ef5e0e04918d1692/pkg/sql/distsqlrun/mergejoiner_test.go#L359)):*

I added these tests which should test this problem, but they seem to pass prior to the fix.

As far as I know, this should still be using the same logic in from the streamMerger, so I'm not sure why it is still working with the bug. Could it maybe be how these `null` are interpreted? I dug a bit but didn't find an answer. @RaduBerinde would you happen to know why?
</blockquote></details>

Hm, I don't know why that would be the case.. I would dig deeper, try to print out the result of every compare that happens in the stream merger

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090)*

lego:

---

*[pkg/sql/distsqlrun/mergejoiner_test.go, line 359 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-Kz5CupqYWLD3kulyWZC:-KzP4dFrsb_3v8XUnc27:btbdoge) ([raw file](https://github.com/cockroachdb/cockroach/blob/69d9a4091a30702aaf4147c2ef5e0e04918d1692/pkg/sql/distsqlrun/mergejoiner_test.go#L359)):*

Hm, I don't know why that would be the case.. I would dig deeper, try to print out the result of every compare that happens in the stream merger
</blockquote></details>

Pinned it down. It was because the tests had an `OnExpr` checking equality while in the code the expression was implicitly removed when the equalities are parsed. This caused the test to filter out the bad results. The tests now correctly test the bug :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090)*

pkg/sql/logictest/testdata/logic_test/distsql_join
```diff
@@ -194,3 +194,48 @@ query T
...
+
+query II rowsort
+SELECT l.k, r.k FROM (SELECT * FROM distsql_mj_test ORDER BY k) l INNER JOIN (SELECT * FROM distsql_mj_test ORDER BY k) r ON l.k = r.k
+----
+
+# If NULL = NULL is joined, we should expect a row of (NULL, NULL)
+query T
+SELECT "URL" FROM [EXPLAIN (DISTSQL) (SELECT l.k, r.k FROM (SELECT * FROM distsql_mj_test ORDER BY k) l INNER JOIN (SELECT * FROM distsql_mj_test ORDER BY k) r ON l.k = r.k)]
+----
+https://cockroachdb.github.io/distsqlplan/decode.html?eJzEkkFr4zAQhe_7K5Y57RIVIifOwVDwNYUmJe2thOBaU1fF8bijMbSE_Pdi65DE1EpzKL1Jo_nemydmBxUZXGRbdJA8ggYFMawV1Ew5Okfcln3T3LxDMlZgq7qRtrxWkBMjJDsQKyVCAgu6ohoUGJTMll3TXgE1ckCcZAVCMturI1kdln3InkpcYWaQT8ShZrvN-CM11ol7Kzfb142gE1CwbCT5m2oY8teX-N8TS9861aNB8eiXw01-Mtx0UPyg2VTEBhlNfxfOt3wx4S1ygTdkq_6YJT7Lv1SP_l-zLV788fA9Ko0GQ8QnIc5s9QpdTZXDby32uE2ApkD_I44azvGOKe9s_HXZcV3BoBP_OvOXeeWf2gGPYR2EozAcBeH4BNZ9eBKEp2Hn6QXOUR-Og_C457ze__kMAAD__4m9mGY=
```
vivekmenezes: Why do you need the SELECT "URL" ?

lego: A pretty unconventional way to assert the plan is using a merge join.

pkg/sql/logictest/testdata/logic_test/distsql_join
```diff
@@ -194,3 +194,48 @@ query T
...
+DELETE FROM distsql_mj_test WHERE TRUE;
+
+statement ok
+INSERT INTO distsql_mj_test VALUES (NULL)
+
+query II rowsort
+SELECT l.k, r.k FROM (SELECT * FROM distsql_mj_test ORDER BY k) l INNER JOIN (SELECT * FROM distsql_mj_test ORDER BY k) r ON l.k = r.k
+----
+
+# If NULL = NULL is joined, we should expect a row of (NULL, NULL)
```
vivekmenezes: Can you place this comment above the previous query. Should it say, "we shouldn't expect a row of (NULL, NULL)"

lego: Oof, thanks for catching that mistake.

vivekmenezes:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-:-KzPMN3w6Jou1vBoWcDw:bnfp4nl)*

vivekmenezes:
radu is on vacation so please feel free to merge

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-:-KzPMP294fX01MyZREZh:b-issy10)*

lego:
TYFR! :) (the last change was just a release note update, btw)

RaduBerinde:
LGTM, thanks.

arjunravinarayan:
Sorry for the delay, I was out of office.

@lego: the code path is not used because we currently don't plan set operations. However we still want the semantics to be right.

I did not think seriously about the treatment of NULLs in algebraic set op, and cannot find documentation about how Postgres handles them either. I'm flagging this for followup when we do algebraic set op planning in #10432. Thank you for pointing that out @raduberinde.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20090#-:-KzzMO7pWol8eoIO60QH:bmoqozh)*

arjunravinarayan:
I did some spelunking, and [Markus Winand](http://modern-sql.com/concept/null#equals-vs-distinct) points out that NULLs in group operations are supposed to sort into a single group, and that the SQL:2016 standard has a longer section on this, that we should ensure we are in compliance with (SQL:2016-2: §4.1.5 and SQL:2016-2: §9.12).

arjunravinarayan:
I apologize, I posted in the wrong issue. Deleting here and reposting there so that your email client registers my apology (then deleting this apology as well).

PR #20091, opened 2017-11-16
sql: test the exact encoding of JSON values

justinj:
Previously we had round-tripping tests, which are great because they're
very easy to read. However, there will come a point for the JSON
encoding where round-tripping isn't enough, since we must be certain
that the encoding is the same (this point in time is after the first
major release containing JSON columns). This commit adds a test along
with fixture data to verify that test data has not changed, along with a
way to easily update it in the case that it does.

@benesch and I discussed this a bit today, but after doing a bit of
research I came across this pattern that I like a lot. Note that both
halves of this test must exist, since if the value doesn't round-trip
then using it as the source of truth is useless.


justinj:
Following an offline discussion with @benesch I removed the constant and replaced it with the `-rewrite-results-in-testfiles` flag used for logic tests, which I pulled out of the logic test file and into the `testutils` package.

benesch (approved):
The CLI tests also import testutils, so they pull in the flag and get confused. The fact that Go's default flag API uses mutable global state is so frustrating. Anyway, looks like you'll need to extract that flag into its own testutils package, or just copy/paste it. Either is fine with me!

justinj:
TFTR!

PR #20092, opened 2017-11-16
sqlccl: correctly mutate table descriptors during resumed RESTORE

madelynnblue:
If a RESTORE was resumed after a pause or node failure, it would
attempt to regenerate the sql descriptors it was using. However it did
this in a way that was different than the first invocation that didn't
use the same resume codepath. This mutation was to remove the FK on
indexes if the user requested it. Move the mutation to a downstream
place so both invocation paths will mutate index descriptors in the
same way.

Also improve some test infra to return errors more quickly and with
useful messages.

### DIFF
--- a/pkg/ccl/sqlccl/backup_test.go
+++ b/pkg/ccl/sqlccl/backup_test.go
@@ -619,18 +619,32 @@ func TestBackupRestoreCheckpointing(t *testing.T) {
 }
 
 func waitForJob(db *gosql.DB, jobID int64) error {
-	return retry.ForDuration(testutils.DefaultSucceedsSoonDuration, func() error {
+	var jobFailedErr error
+	err := retry.ForDuration(testutils.DefaultSucceedsSoonDuration, func() error {
 		var status string
+		var payloadBytes []byte
 		if err := db.QueryRow(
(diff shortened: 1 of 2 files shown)


mjibson:
Ping @dt.

pkg/ccl/sqlccl/backup_test.go
```diff
@@ -619,18 +619,32 @@ func TestBackupRestoreCheckpointing(t *testing.T) {
...
+		}
 		if e, a := jobs.StatusSucceeded, jobs.Status(status); e != a {
 			return errors.Errorf("expected backup status %s, but got %s", e, a)
 		}
 		return nil
 	})
+	if jobFailedErr != nil {
+		return jobFailedErr
+	}
+	return err
```
benesch: Ah, much better!

benesch (approved):
Nice, LGTM!

dt:
:lgtm: nice catch!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20092#-:-KzPqpZ9DjMvYjVgZzcK:b-pe67a)*

PR #20093, opened 2017-11-16
sql: SHOW TRACE: misc improvements

### LINKED ISSUE #19453: sql: SHOW SESSION TRACE does not extract file-prefixed tags properly
Found in https://teamcity.cockroachdb.com/viewLog.html?buildId=384117&buildTypeId=Cockroach_UnitTests:

if a log statement logs a message of the form "filename:lineno [tags / context] message", as opposed to just "[tags / context] message", the tags and context information are not properly split into the separate context column.

Fixing this requires an extension to the definition of `traceClause` in `(*planner).ShowTrace()` (`sql/show.go`).

### LINKED ISSUE #20085: log messages with file:line prefix are not parsed correctly by SHOW TRACE FOR
When a log message is produced with `log.Infof()` (as opposed to `log.VEventf()`) - or, technically, I think it is when we want to print it to the console/file besides putting it in the trace - we include a `file:line` prefix to it. This prefix breaks the regex that `SHOW TRACE FOR` uses for extracting the "context" from the message (i.e. the log tags - e.g. `[n1,r1]`).

The regex is here:
https://github.com/cockroachdb/cockroach/blob/f86e6b83838c837b7de6033f2a8605dcc08a53f6/pkg/sql/show.go#L292

I can "fix it" with something like the following. There's a regex with 3 groups in there; the middle one is the one the code is currently using. The first one is the filename prefix. I'd extract the first and 3rd group.
@knz are you with me? At a later point we should make the trace messages more structured somehow...

```
root@:26257/> SELECT regexp_replace('sql/xxx:123 [client=[::1]:49731,user=root,n1] !!! plan completed execution', E'^(.+/.*:\\d+ )?(\\[(?:[^][]|\\[[^]]*\\])*\\])(.*)', '\1\3');
+-------------------------------------------+
| regexp_replace('sql/xxx:123 [client=[::1]:49731,user=root,n1] !!! plan completed execution', e'^(.+/.*:\\d+ )?(\\[(?:[^][]|\\[[^]]*\\])*\\])(.*)', e'\\1\\3') |
+-------------------------------------------+
| sql/xxx:123  !!! plan completed execution |
+-------------------------------------------+
(1 row)
```

knz:
Fixes #20085.
Fixes #19453.

###  sql: split the tag and filename/lineno information in tracing output

Prior to this patch, the filename/lineno information in tracing output
wasn't properly recognized, in turn causing the tag not to be split
off properly. Also, the extraction was performed using the SQL
function `regexp_replace`, causing some RE lookup overhead.

This patch fixes the issues by extracing the filename/lineno in a
separate column.  This adds new columns `tag` and `loc` in
`crdb_internal.session_trace`.  The extraction is done in Go, with a
well-identified, pre-compiled RE.

It also exposes the `loc` column in the output of SHOW TRACE.  The
`context` column is renamed to `tag`, since this name more closely
mirrors the terminology used in the source code, and users looking at
the output of SHOW TRACE will usually also look at the source code
too.

###  sql: new variant SHOW COMPACT TRACE for common cases

Requested by @andreimatei: in most cases SHOW TRACE is used to
investigate the duration of each operation and the message. Also the
file/lineno location as prefix of the message is desired in this case.

To respond to this common case, this patch introduces the variant
`SHOW COMPACT [KV] TRACE` which only renders the age, message, tag and
operation columns. The file/lineno information is prepended to the
message.

For example:

```sql
> SHOW COMPACT TRACE FOR SESSION;
+----------------+-----------------------------------------------------------------------------------------+---------------------------------------+------------------+
|      age       |                                                 message                                 |                         tag           |     operation    |
+----------------+-----------------------------------------------------------------------------------------+---------------------------------------+------------------+
| 0s             | === SPAN START: sql txn implicit ===                                                    |                                       | sql txn implicit |
| 174µs525ns     | === SPAN START: sql txn ===                                                             |                                       | sql txn          |
| 190µs114ns     | sql/executor.go:1184 executing 1/6: BEGIN TRANSACTION                                   | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| 225µs320ns     | sql/executor.go:1184 executing 2/6: SELECT 1                                            | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| 269µs362ns     | sql/executor.go:2099 query not supported for distSQL: unsupported node *sql.unaryNode   | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| 298µs987ns     | sql/executor.go:1184 executing 3/6: COMMIT TRANSACTION                                  | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
...
```


knz:
Will also cherry pick to 1.1.

andreimatei:
I think I'd prefer it if the location wasn't a dedicated column; instead it'd remain as a prefix of the message. I find it that many times that file name is the most important part of the log message :). I like reading it together with the msg.
I sometimes feel the same about the log tags, but there I'm more ambivalent because it is also sometimes useful to do processing on the tags (e.g. query for all messages from a node). I think that's less likely to be useful with file names.
What do you think?

---

---

*[pkg/sql/show.go, line 286 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20093#-Kz4mAIMF7Ef-fUm0Ybm:-Kz4mAIMF7Ef-fUm0Ybn:b-5iuf1o) ([raw file](https://github.com/cockroachdb/cockroach/blob/251bb10459c7302a35cd5f0b8adf09252cc8ad7f/pkg/sql/show.go#L286)):*
> ```Go
> 	// A regular expression for log messages.
> 	// It has three parts:
> 	// - the (optional) code location: ((?:[^ :]+:[0-9]+ )?)
> ```

don't we want at least a slash in here somewhere?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20093#-:-Kz4mDQPCHxE9Cq160JQ:b-ecpg21)*

andreimatei:
for testing, I'd extract the regex into a constant and test that in isolation on a string. And btw I'd do the same instead of `TestBracketInTracetags` - I had a "da hell is this" reaction when I saw that guy the other day :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20093#-:-Kz4qtvLqC5xIE5GETG8:b3uvcnt)*

knz:
I have renamed `TestBracketInTracetags` to `TestTraceFieldDecomposition`
and added a comment to clarify.

I have added the test for the filename to that
TestTraceFieldDecomposition test too, using log.Infof instead of
log.Event like you suggested. PTAL.

Regarding your remaining questions:

On 16-11-17 18:07, Andrei Matei wrote:
> I think I'd prefer it if the location wasn't a dedicated column; instead
> it'd remain as a prefix of the message. I find it that many times that
> file name is the most important part of the log message :). I like
> reading it together with the msg.

Then you can use `loc || ' ' || message` in your `show trace`
invocation. We could perhaps consider adding a view for this.

It's easier (when invoking) to group the separate strings than to
separate the grouped string. So I think CockroachDB should be in charge
of separating.

Also the match logic for SHOW KV TRACE needs them to be separated. It
would be silly to separate them in SHOW KV TRACE, then group them again,
for the user to potentially separate them again later.

> I sometimes feel the same about the log tags, but there I'm more
> ambivalent because it is also sometimes useful to do processing on the
> tags (e.g. query for all messages from a node). I think that's less
> likely to be useful with file names.
> What do you think?

`SELECT foo || bar`. That's what I think.

-- 
Raphael 'kena' Poss

andreimatei:
LGTM

I still think adding a new column is not a good idea, but I don't care too much; you can do it if you really think it's good.
If it is a column, I don't think recognizing it at the level of this SQL query is the way to go; we should put the structure in recorded trace and from there in the session_trace table. But again I don't care too much, you can do this if you want.


> Then you can use loc || ' ' || message in your show trace invocation. We could perhaps consider adding a view for this.

Well, but STF is a view! The raw data is crdb_internal.session_trace. STF is supposed to make common usage easier (which, btw, I was never fully on board with because I think we have already gotten its schema wrong with the loc column).

> It's easier (when invoking) to group the separate strings than to separate the grouped string. So I think CockroachDB should be in charge of separating.

This statement on its own is true, but I still maintain that you never want to separate the strings, so it doesn't matter how hard it is :). You say that SHOW KV TRACE needs to separate them. I question whether it really does - can't we just slap a `.*` in front of whatever they match now? Also, it doesn't really matter whether the KV TRACE regexes match lines with filenames or not because those lines are not printed at the Info level. I think it would matter when vmodule is enabled on one of the files that produces those lines, but, as far as I'm concerned, whatever. I think letting KV TRACE influence anything is, as a living classic recently put it, perpetuating (or was it "expanding"?) garbage. Btw, these "kv trace" messages should also be recognized at a different level than this SQL query. 

---

---

*[pkg/sql/show.go, line 286 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20093#-Kz4mAIMF7Ef-fUm0Ybm:-KzUOGf5_DE7Jox6gQMa:b-ahwtwa) ([raw file](https://github.com/cockroachdb/cockroach/blob/251bb10459c7302a35cd5f0b8adf09252cc8ad7f/pkg/sql/show.go#L286)):*

don't we want at least a slash in here somewhere?
</blockquote></details>

I don't really understand this regex. 
`[^]` - this can just be `^`, beginning of line, right?
`[ :]+` - one or more spaces or colons? Why accept spaces and more than one column (and also there's another column coming up)

And again I'm surprised that there's no slashes. Isn't it a good idea to mandate at least a slash somehow? Cause it's already easy for this expression to gobble up stuff that's not really a file name. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20093#-:-KzUM8VI7ucXYHl3tywQ:b-vredrn)*

knz:
Oh the idea of having the `crdb_internal.session_trace` do the split is actually not half bad. This way the regular expression can be properly pre-compiled. I'll have a look. Thanks for the hint!

knz:
Concerns addressed; PTAL.

andreimatei:
:lgtm:

I personally don't `COMPACT TRACE` very much; too few people are going to know about it. If I haven't convinced you to leave the location in the message in `SHOW TRACE FOR`, then I'd rather do the concatenations myself.
But as you think it's best.

---

Review status: 0 of 11 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20093#-:-KzzZAq1asP8x8DayROb:btg0kf3)*

knz:
cc @tschottdorf

PR #20096, opened 2017-11-16
sql: code reorganization to facilitate refactorings

knz:
**tl;dr: This patch series organizes the source in the `sql` package to follow the overall architecture of the SQL layer. This is a transitionary situation.**

In an ideal world, we would make the SQL components live in separate Go packages. However, because Go does not allow circular dependencies, this separation is technically difficult. 

Multiple refactorings will be needed over time to achieve this separation. **However, the process to perform these refactorings needs often to iterate over "all files corresponding to this or that component".** Because the `sql` package so far was a flat bag of files, these iterations were difficult, as they requires "warm cache" knowledge of which file belongs to which commit.

**To facilitate further refactorings,** and perhaps facilitate the onboarding of team members who wish to approach and/or learn about the SQL package while these refactorings are going on, this PR starts the separation of concerns by using separate file name prefixes.  **This is a transitionary situation.**

The file name prefixes automatically group the files corresponding to similar concerns next to each other in file listings, and also make iterations over "all files corresponding to this or that component" trivial.

The following name prefixes are adopted:

| SQL middle-end component              | File name prefix |
|---------------------------------------|------------------|
| Session, Executor                     | `x_`             |
| Prep phase (initial logical planning) | `p_`             |
| Logical plan optimization             | `o_`             |
| Schema access, including lease mgt.   | `sa_`            |
| Schema changes                        | `sc_`            |
| Logical plan constructors + local execution | `n_`             |

The latter set of files, currently prefixed with `n_`, will, over time, be split up between code pertaining to logical planning (planNode hierarchy and constructors), and code pertaining to local (i.e. non-distributed) execution.

cc @vivekmenezes


knz:
cc @cockroachdb/sql

vivekmenezes:
@knz while I reeally appreciate you making this change, I think it's not solving a problem. I think you've nailed the problem well--"In an ideal world, we would make the SQL components live in separate Go packages. However, because Go does not allow circular dependencies, this separation is technically difficult."

I'd much prefer we continue to place stuff in separate Go packages. Even if there are circular dependencies they can be fixed by introducing interfaces. I much prefer using the directory structure to organize this

knz:
On 16-11-17 17:18, Jordan Lewis wrote:
> Can you separate this PR into at least 2: name-prefix additions and
> everything else? The "everything else" bits are /much/ less
> controversial. My strategy for merging something like this needs to have
> a phase for getting the easy stuff in first.

Done: #20101

> Making strong interfaces and
> packages that provide them is really the only way forward, in my
> opinion. 

Yes I agree, and this will happen too. *Later.*

I wasn't clear about the intent. Please re-read the (modified) PR
description. I have highlighted how this is a transitionary state.

> I'm afraid that it also gives us a false
> sense of organization > in reality, the |sql| package is just as messy
> and intertwined with itself after this change. 

But that is not true on its face! Until today, we have been rather well
disciplined to separate concern into different files.

Yes, the dependencies between components are not exposed clearly in the
code by means of public interfaces, but no the code is not intertwined.
The concerns are, IMHO, actually rather well separated. For every
function or sub-calltree, we can cleanly categorize where it belongs in
the architecture of the SQL layer.

There is *some* cleanup needed (some individual tests / functions are
not exactly well placed in the file that corresponds most to the main
concern addressed in the rest of the file), but consider this:

**in order to realize that the separation of concern is not well
achieved, we need first to identify which files belong to which concerns.**

This PR facilitates this cleanup work as a first, intermediate step.


> I fear that such a change
> will sap us of motivation to try to improve the separation of concerns
> in |sql|.

Hah, are you crazy :)
I am well beyond this PR in terms of actually splitting the code.
However to facilitate the review of the later patches, it would be much
easier on the reviewer to be confident that the code movements pertain
to isolated concerns in the code. Seeing code movements restricted to a
group of files with the same filename prefix will enhance that confidence!


> I propose we work together to try to disentangle some the bits from the
> edges. I'd like to see (if at all possible) a reasonable list of
> "confounding dependencies" - dependency edges that prevent us from doing
> a useful package split, for a single "logical package". I would try to
> pick a simple one first - say, virtual tables.

I agree we should do that (and let's do it!) but that is a separate,
orthogonal concern.

-- 
Raphael 'kena' Poss

knz:
Another use case where this is already helping me. For every identifier I'm interested to perhaps migrate somewhere else, I use go-guru to list all the call/use sites. With the file name identifying the concern, I know the scope of the usage from the go-guru output (1 line per match) without having to open each file in turn to understand the context.

knz:
> I'm -1 on the name prefixes. They're ugly :)

Let's be disciplined here:

If change X helps me, and does not prevent you from doing your job as
effectively as previously, then change X is good, regardless of how
"ugly" you find it. I have made arguments, above already and now below
in this reply, to support this change. You can discuss those arguments.

If you want to make an argument that the name prefixes are going to be
effectively *detrimental* to anyone in an objective way (e.g. increases
the marginal time you need to spend on things you do often, or you have
a compelling argument that it will get in the way of learning), please
do that.

Just saying "they're ugly" sounds like "it's different and therefore I
don't like it" to me and I can't let that get in the way of making
things better.

> and I don't think people will know what they mean.

I added a README.md to explain this.

> Like everybody (including you I think), I think separating packages is
> the way to go

See my response to Jordan: I agree on this, and we'll work over time to
make this happen.

> I'm not convinced that this file rename step helps with
> anything. 

This is already false on its face: it helps *me* because I can more
quickly iterate over "all files in a concern" and identify dependencies
between abstract packages more clearly.

Also, Joey chimed in already and said it helped him perceive the
high-level structure.

> Btw, if we do any en-masse renames in a single commit, please make sure
> that git and github are able to maintain history for the files [...]

1) The thing to be mindful about is that git, unlike subversion and
other VCSs, does not have a primitive "rename file" operation. So it
gets confused when you both rename *and* modify a file in a major way in
the same commit. To avoid this confusion we must separate renames and
edits into separate commits, then the rename becomes visible. I've been
careful to do that in #19941 and here too.

2) tip: When using git blame, use -M -C -C -C (note: 3x -C!) to properly
track changes across renames/copies.

-- 
Raphael 'kena' Poss

RaduBerinde:
:lgtm:

I don't see the downside to the rename. If someone doesn't know what the `n_` stands for in `n_index_join.go`, it's still no worse than `index_join.go`. Plus, it reminds me of the DOOM and Quake 1 source code.

knz:
@jordanlewis moved the tests in sub-packages as you suggested - separate PR #20117

jordanlewis:
Like we discussed offline, I still have objections around this. Primarily, I'm concerned that the naming scheme is intimidating to outsiders and insiders alike, and will communicate disorder and arcane-ness despite the good intentions.

Since you're the primary consumer of this rename, I'd like to see a serious try at some tooling to help with the organization problem before we propagate the naming to everyone. The best idea I can think of so far (again like we discussed, just writing down for posterity) is to maintain a parallel source tree that has symlinks for all files, with the files in the `pkg/sql` tree symlinked to their prefixed names as you have suggested.

I know I'm being a PITA about this, so thank you for taking my suggestions seriously!

andreimatei:
> If you want to make an argument that the name prefixes are going to be
effectively *detrimental* to anyone in an objective way (e.g. increases
the marginal time you need to spend on things you do often, or you have
a compelling argument that it will get in the way of learning), please
do that.

That's what I mean; I was hoping it's implied. Adding (what I see as) random characters to names does hurt - e.g. takes space in my tabs and pushes out the real filename outside of view.
Anyway, this is not a big deal and if you insist, you can go ahead as far as I'm concerned.
I'd definitely rather have the renames than some symlinks besides the files. (But you can have the symlinks on your machine if that helps you do some refactoring :trollface: )

> The thing to be mindful about is that git, unlike subversion and
other VCSs, does not have a primitive "rename file" operation.

I'm well aware. But my understanding about what happened with the `pkg` move was that the commit was so large that all the git tools (including github) are abandoning the rename detection when looking at history, and so my conclusion is that we need to only move so much code in any single commit.

jordanlewis:
> I'd definitely rather have the renames than some symlinks besides the files. (But you can have the symlinks on your machine if that helps you do some refactoring :trollface: )

To be totally clear, I was suggesting local symlinks (and possibly a script to manage them), not checked-in symlinks. Agree that I'd prefer the renames to *checked-in* symlinks.

knz:
I have found the symlink solution workable for the time being. Let's close this for now, if there's a reason to re-open we can re-do this later.

PR #20098, opened 2017-11-16
tests: pass testing.T in individual calls

dt:
With our increased use of subtests, we often have multiple `testing.T`s
floating around in tests. When we close over one of them in a struct
like SQLRunner that then gets used elsewhere, we often end up calling
methods on the wrong `testing.TB`, leading to confusing and
uninformative errors rather than actual failue message / location. Since
these are usually only called when something breaks, this may be long
after the change introducing the subtest, and thus harder to recognize
what's wrong.

Instead, we should treat testing.T more like context -- always passed in
at individual callsites and not stored in stucts that may be passed
between scopes or goroutines.

There are probably a few places that, after this change, no longer
benefit as much from SQLRunner or could be cleaned up,  however I tried
to keep this change as simple and mechanical as possible given its
breadth.

### DIFF
--- a/pkg/ccl/acceptanceccl/backup_test.go
+++ b/pkg/ccl/acceptanceccl/backup_test.go
@@ -138,22 +138,22 @@ func (bt *benchmarkTest) Start(ctx context.Context) {
 		bt.b.Fatal(err)
 	}
 	defer sqlDBRaw.Close()
-	sqlDB := sqlutils.MakeSQLRunner(bt.b, sqlDBRaw)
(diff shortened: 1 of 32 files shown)


RaduBerinde:
:lgtm:  I agree that storing the wrong T can lead to very confusing errors, better to avoid this possibility altogether.

---

Review status: 0 of 32 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20098#-:-Kz4X0BYrF62m54lvxgZ:b7x75ol)*

PR #20099, opened 2017-11-16
sql: add benchmark for inserting into a table with a secondary key

m-schneider:


pkg/sql/bench_test.go
```diff
@@ -343,6 +376,7 @@ func BenchmarkSQL(b *testing.B) {
 			runBenchmarkTrackChoices,
 			runBenchmarkUpdate,
 			runBenchmarkUpsert,
+			runBenchmarkInsertSecondaryIndex,
```
jordanlewis: But now the nice sorted list isn't sorted anymore!

pkg/sql/bench_test.go
```diff
@@ -332,6 +332,39 @@ func runBenchmarkInsertFK(b *testing.B, db *gosql.DB, count int) {
...
+// runBenchmarkInsertSecondaryIndex benchmarks inserting count rows into a table with a
+// secondary index.
+func runBenchmarkInsertSecondaryIndex(b *testing.B, db *gosql.DB, count int) {
+	defer func() {
+		if _, err := db.Exec(`DROP TABLE IF EXISTS bench.insert`); err != nil {
+			b.Fatal(err)
+		}
+	}()
+
+	if _, err := db.Exec(`CREATE TABLE bench.insert (k INT PRIMARY KEY, v INT , INDEX(v))`); err != nil {
```
jordanlewis: I think you should make several indexes on this column, so that the benchmark is affected as much as possible by the index entry creation code. You can do that with a `CREATE INDEX` statement, which can be run on the same column as many times as you want.

justinj:
:lgtm: but :microscope: :dog: and could you explain in the commit message what prompted the addition of this benchmark?

---

---

*[pkg/sql/bench_test.go, line 351 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4bBrsFBcUJoYjCEWa:-Kz4bBrsFBcUJoYjCEWb:bacnpeo) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L351)):*
> ```Go
> 	var buf bytes.Buffer
> 	val := 0
> 	for i := 0; i < b.N; i++ {
> ```

I'm not really up on our best practices re: these tests, but would it be good to stop the timer before constructing the query string and start it again after?

Something else seems odd to me - I had assumed before that most benchmarks were predicated on every run of them being the ~same modulo random variation, but this benchmark and all the other ones in this file constantly change the state of the database throughout their execution? Is that weird? Won't the benchmarking library try to make statistical inferences that aren't reasonable?

---

*[pkg/sql/bench_test.go, line 358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4bQtMjulWhSwo59Fx:-Kz4bQtMjulWhSwo59Fy:bipm22v) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L358)):*
> ```Go
> 				buf.WriteString(", ")
> 			}
> 			fmt.Fprintf(&buf, "(%d, %d)", val, val)
> ```

Is the fact that all the inserts are sequential going to impact the performance of this in any significant way (I don't know the answer)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-:-Kz4c7rKm64e_q2MGJfE:bdya3ln)*

m-schneider:

---

*[pkg/sql/bench_test.go, line 344 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4_uxTvFZcADioxbWf-r1-344:-Kz5413F43Yrj3O5IGoo:bayfjx7) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L344)):*

I think you should make several indexes on this column, so that the benchmark is affected as much as possible by the index entry creation code. You can do that with a `CREATE INDEX` statement, which can be run on the same column as many times as you want.
</blockquote></details>

Do you think 10 is enough?

---

*[pkg/sql/bench_test.go, line 351 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4bBrsFBcUJoYjCEWa:-Kz5237NcCM9e033xF_u:btk7tfo) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L351)):*

I'm not really up on our best practices re: these tests, but would it be good to stop the timer before constructing the query string and start it again after?

Something else seems odd to me - I had assumed before that most benchmarks were predicated on every run of them being the ~same modulo random variation, but this benchmark and all the other ones in this file constantly change the state of the database throughout their execution? Is that weird? Won't the benchmarking library try to make statistical inferences that aren't reasonable?
</blockquote></details>

I'm not sure about our best practices either, I modified another test. 

Jordan, do you have any thoughts?

---

*[pkg/sql/bench_test.go, line 358 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4bQtMjulWhSwo59Fx:-Kz54AuLqp4nBLjPBmTr:b-fyx39o) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L358)):*

Is the fact that all the inserts are sequential going to impact the performance of this in any significant way (I don't know the answer)?
</blockquote></details>

I added a [][]byte instead of []byte as a return type for the function that creates the secondary index and we're testing to see if this change has a negative impact on memory.

---

*[pkg/sql/bench_test.go, line 379 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4_uxTvFZcADioxbWf-r1-379:-Kz54Rc05a_fZR3QKisb:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L379)):*

But now the nice sorted list isn't sorted anymore!
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099)*

jordanlewis:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

*[pkg/sql/bench_test.go, line 344 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4_uxTvFZcADioxbWf-r1-344:-Kz55rS1ZIbZTmFCqWiW:b-c7wi02) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L344)):*

Do you think 10 is enough?
</blockquote></details>

Yes that sounds good

---

*[pkg/sql/bench_test.go, line 351 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-Kz4bBrsFBcUJoYjCEWa:-Kz55sinQogqtUbuV3sP:b-xftsrc) ([raw file](https://github.com/cockroachdb/cockroach/blob/16d990677fe43d9be59e0bed62785b0485fdb9ce/pkg/sql/bench_test.go#L351)):*

I'm not sure about our best practices either, I modified another test. 

Jordan, do you have any thoughts?
</blockquote></details>

Agreed that this is funky but since all of our benchmark tests do this I'd say now is not the time to answer this question.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-:-Kz5APqqXrMW5N_m5yR5:bnfp4nl)*

m-schneider:
TFTR!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20099#-:-Kz5YG_S5gjmzJBCfGB-:b-814g8k)*

PR #20101, opened 2017-11-16
sql: minor refactors

knz:
Suggested by @jordanlewis here: https://github.com/cockroachdb/cockroach/pull/20096#issuecomment-344974610


jordanlewis:
First 2 commits :lgtm:, last one I have a better idea for.

---

Reviewed 4 of 4 files at r1, 5 of 5 files at r2, 9 of 9 files at r3.

---

*[pkg/sql/benchpg_test.go, line 31 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20101#-Kz4gtD6JEIANQH1a2-H:-Kz4gtD6JEIANQH1a2-I:b-3sfyzc) ([raw file](https://github.com/cockroachdb/cockroach/blob/3db94973c5ffa15129cd3c05b5740eed57c348b8/pkg/sql/benchpg_test.go#L31)):*
> ```Go
> 	"github.com/cockroachdb/cockroach/pkg/base"
> 	"github.com/cockroachdb/cockroach/pkg/security"
> 	"github.com/cockroachdb/cockroach/pkg/sql/benchpg"
> ```

Why does this file have to live in `sql` at all? I think it should be placed in the `pgbench` package.

Actually, I no longer agree with this patch. This package is called `pgbench` because there's a tool called `pgbench` that it uses.

What about moving this whole package into a new top level package called `bench` or something? This code is almost dead anyway. We never use or look at the results of these benchmarks.

---

*[pkg/sql/benchpg/setup.go, line 106 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20101#-Kz4gZ9o0oTh_IXwF7zd:-Kz4gZ9paZtWCBpIlP-z:b-olvsxr) ([raw file](https://github.com/cockroachdb/cockroach/blob/3db94973c5ffa15129cd3c05b5740eed57c348b8/pkg/sql/benchpg/setup.go#L106)):*
> ```Go
> }
> 
> // SetupBenchDB sets up a db with the schema and initial data used by `benchpg`.
> ```

Hmm I think these comments shouldn't have changed. `pgbench` is the name of a program outside of cockroachdb.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20101#-:-Kz4hY8Q8cr6ksYpId77:bzea40x)*

knz:
Okay I removed the last commit about the pgbench package from this PR. Will submit as separate PR. 
Will merge (The remaining 2 commits) when green.

jordanlewis (approved):

PR #20103, opened 2017-11-16
kv: simplify next() and truncate()

### LINKED ISSUE #18395: kv: check that next() and truncate() are sound
This is a reminder to self to double-check these methods in the wake of #18351.

tbg:
The old code was correct, but this is slightly more efficient and
also gives the reader less to think about.

Release notes: none.

Closes #18395.

### DIFF
--- a/pkg/kv/batch.go
+++ b/pkg/kv/batch.go
@@ -60,7 +60,7 @@ func truncate(ba roachpb.BatchRequest, rs roachpb.RSpan) (roachpb.BatchRequest, 
 		if err != nil {
 			return false, emptySpan, err
 		}
-		if l, r := !keyAddr.Equal(header.Key), !endKeyAddr.Equal(header.EndKey); l || r {
+		if l, r := keys.IsLocal(header.Key), keys.IsLocal(header.EndKey); l || r {
 			if !l || !r {
 				return false, emptySpan, errors.Errorf("local key mixed with global key in range")
 			}
@@ -219,6 +219,10 @@ func next(ba roachpb.BatchRequest, k roachpb.RKey) (roachpb.RKey, error) {
 			return nil, err
 		}
 		if addr.Less(k) {
+			if len(h.EndKey) == 0 {
+				// `h` affects only `[KeyMin,k)`, all of which is less than `k`.
+				continue
+			}
 			eAddr, err := keys.AddrUpperBound(h.EndKey)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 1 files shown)


nvanbenschoten:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/kv/batch.go, line 63 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20103#-Kz5MNytRavZdPkPvOC-:-Kz5MNytRavZdPkPvOC0:b351q58) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7193ca80e06d162bb9f0ad0c68785e1f0bdf04e/pkg/kv/batch.go#L63)):*
> ```Go
> 			return false, emptySpan, err
> 		}
> 		if l, r := keys.IsLocal(header.Key), keys.IsLocal(header.EndKey); l || r {
> ```

Nice :+1: I've seen the previous pattern used a few times in different areas and it always stops me in my tracks for a minute or two.

---

*[pkg/kv/batch.go, line 223 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20103#-Kz5OC_o2uVEt0G0hqJC:-Kz5OC_o2uVEt0G0hqJD:bnal94b) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7193ca80e06d162bb9f0ad0c68785e1f0bdf04e/pkg/kv/batch.go#L223)):*
> ```Go
> 		if addr.Less(k) {
> 			if len(h.EndKey) == 0 {
> 				// Affects only [KeyMin,k).
> ```

Could you change here and below to `// h affects only [KeyMin,k), which is less than k.`? Like this, it's easier to read wrong (e.g. "this branch affects only ...").

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20103)*

tschottdorf:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/kv/batch.go, line 223 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20103#-Kz5OC_o2uVEt0G0hqJC:-Kz5RktdvLKKhwaU_xQu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7193ca80e06d162bb9f0ad0c68785e1f0bdf04e/pkg/kv/batch.go#L223)):*

Could you change here and below to `// h affects only [KeyMin,k), which is less than k.`? Like this, it's easier to read wrong (e.g. "this branch affects only ...").
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20103)*

nvanbenschoten:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20103#-:-Kz5UWwcVtkkNJJZnuhV:bnfp4nl)*

tschottdorf:
TFTR!

PR #20104, opened 2017-11-16
storage: report stringified RocksDB error code

tbg:
We are now seeing first error reports that have a RocksDB-originating error at their root,
so next it would be nice to see what classes of errors we see. We hope to see "IO error"
a lot, as that usually means out of disk.

See https://github.com/facebook/rocksdb/blob/master/include/rocksdb/status.h for a list
of status code messages that we can expect to see.

Release note: none

### DIFF
--- a/pkg/storage/engine/rocksdb.go
+++ b/pkg/storage/engine/rocksdb.go
@@ -24,6 +24,7 @@ import (
 	"runtime"
 	"runtime/debug"
 	"sort"
+	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


bdarnell:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20104#-:-Kz5acHVFWOPUFQP2zgv:bnfp4nl)*

Issue #5811, opened 2016-04-01
sql: support CREATE SEQUENCE


petermattis:
...and add support for the `nextval()`, `currval()` and `setval()` functions. Note that these functions are atomic, but not transactional. We can implement them using the KV `{Inc,Get,Set}` operations. We'll need to figure out where to store the sequence key. Postgres says that a sequence creates a special single-row table, which could work for us as well. Something like:

```
CREATE TABLE sequence (
  val INT64
);
```

Note that Postgres supports tying the lifetime of a sequence to the lifetime of a specific column in a table. That would require adding a field to `ColumnDescriptor` that would allowing cascading the deletion of a column (either by dropping the column or the entire table) to deletion of the associated sequences.

JackKrupansky:
Sort of related to the issue I filed on SERIAL:
sql: support SERIAL PK for appending rows to a table #4491

mjibson:
One reason to do this as not just a wrappr around `unique_rowid()` is for sorting. Since a sequence usually results in things being inserted with increasing IDs, you can use those IDs as a sort column (otherwise you'd have to use a column with default `NOW()` or something). Sure, you get some things out of order, but in general it'll be correct.

petermattis:
`unique_rowid` is only slightly less sorted than a sequence backed by KV `Inc`. Am I missing something?

mjibson:
`unique_rowid` is only sorted per node since the node ID is MSB of the output. A sequence would be sorted per cluster.

petermattis:
We never made that change. `unique_rowid()` has the timestamp in the upper bits and the node ID in the lower bits.

mjibson:
I thought it was the opposite. Strike my comment then.

petermattis:
@paperstreet Should we close this issue?

paperstreet:
I think leave it open and stick it in Later until someone needs it. SEQUENCE can be used for other things than implementing SERIAL

bdarnell:
A need for this was reported in #12481

dianasaur323:
A user also mentioned an issue with this when trying to load in pg dumped file. @nstewart

vilterp:
@petermattis Is there a reason why `lastval` was left out of this issue? It doesn't seem necessary (since you can always use `currval`, and probably should to avoid accidentally getting a value from the wrong sequence), but was just wondering what your reasoning was.

petermattis:
No reason, just an oversight.

[...]

PR #20111, opened 2017-11-16
sql: information_schema.sequences and pg_catalog.pg_sequence

vilterp:
- `information_schema.sequences` was there previously as a placeholder. Now that we actually have sequences, return them.
- also add `pg_catalog.pg_sequence`, since it's practically the same thing.

tracking issue for SQL sequences: #19723

### DIFF
--- a/pkg/sql/information_schema.go
+++ b/pkg/sql/information_schema.go
@@ -20,6 +20,8 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 
+	"fmt"
+
 	"github.com/cockroachdb/cockroach/pkg/security"
 	"github.com/cockroachdb/cockroach/pkg/sql/privilege"
 	"github.com/cockroachdb/cockroach/pkg/sql/sem/tree"
@@ -305,10 +307,25 @@ CREATE TABLE information_schema.sequences (
     CYCLE_OPTION STRING NOT NULL DEFAULT 'NO'
 );`,
 	populate: func(ctx context.Context, p *planner, prefix string, addRow func(...tree.Datum) error) error {
-		// Sequences are not yet supported: #5811
-		// However, we support an empty information_schema table to enable
-		// clients to observe that there are no sequences.
-		return nil
+		return forEachTableDesc(ctx, p, prefix, func(db *sqlbase.DatabaseDescriptor, table *sqlbase.TableDescriptor) error {
+			if !table.IsSequence() {
+				return nil
+			}
+			return addRow(
+				defString,                        // catalog
+				tree.NewDString(db.GetName()),    // schema
+				tree.NewDString(table.GetName()), // name
+				tree.NewDString("INT"),           // type
+				tree.NewDInt(64),                 // numeric precision
+				tree.NewDInt(2),                  // numeric precision radix
+				tree.NewDInt(0),                  // numeric scale
+				tree.NewDString(fmt.Sprintf("%d", table.SequenceOpts.Start)),     // start value
+				tree.NewDString(fmt.Sprintf("%d", table.SequenceOpts.MinValue)),  // min value
+				tree.NewDString(fmt.Sprintf("%d", table.SequenceOpts.MaxValue)),  // max value
+				tree.NewDString(fmt.Sprintf("%d", table.SequenceOpts.Increment)), // increment
+				yesOrNoDatum(table.SequenceOpts.Cycle),                           // cycle
+			)
+		})
 	},
 }
 
@@ -525,6 +542,9 @@ CREATE TABLE information_schema.tables (
 );`,
 	populate: func(ctx context.Context, p *planner, prefix string, addRow func(...tree.Datum) error) error {
 		return forEachTableDesc(ctx, p, prefix, func(db *sqlbase.DatabaseDescriptor, table *sqlbase.TableDescriptor) error {
+			if table.IsSequence() {
+				return nil
+			}
 			tableType := tableTypeBaseTable
 			if isVirtualDescriptor(table) {
 				tableType = tableTypeSystemView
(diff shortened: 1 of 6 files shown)


lego:

---

*[pkg/sql/logictest/testdata/logic_test/information_schema, line 936 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111#-Kz695QMkqqYh9TVXYLI:-Kz695QMkqqYh9TVXYLJ:bi1f1ff) ([raw file](https://github.com/cockroachdb/cockroach/blob/753c657975ca62e14303dfa726da6249334e0c6f/pkg/sql/logictest/testdata/logic_test/information_schema#L936)):*
> ```
> # information_schema.sequences
> 
> statement ok
> ```

You should also test
- before any sequences are made
- what you see when you switch databases

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111)*

justinj:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[pkg/sql/information_schema.go, line 310 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111#-Kz6ceA04kdoLK0LlU8n:-Kz6ceA04kdoLK0LlU8o:by9f17c) ([raw file](https://github.com/cockroachdb/cockroach/blob/753c657975ca62e14303dfa726da6249334e0c6f/pkg/sql/information_schema.go#L310)):*
> ```Go
> );`,
> 	populate: func(ctx context.Context, p *planner, prefix string, addRow func(...tree.Datum) error) error {
> 		// Sequences are not yet supported: #5811
> ```

you can get rid of this comment now!

---

*[pkg/sql/information_schema.go, line 314 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111#-Kz6gr0Bcg4XsLP3dVAE:-Kz6gr0CWuJ2a44cU6Qt:bp6bmof) ([raw file](https://github.com/cockroachdb/cockroach/blob/753c657975ca62e14303dfa726da6249334e0c6f/pkg/sql/information_schema.go#L314)):*
> ```Go
> 		// clients to observe that there are no sequences.
> 		forEachTableDesc(ctx, p, prefix, func(dbDesc *sqlbase.DatabaseDescriptor, tableDesc *sqlbase.TableDescriptor) error {
> 			if tableDesc.SequenceOpts == nil {
> ```

I notice that Postgres's `information_schema.tables` doesn't include sequence while ours does, was that intentional?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111)*

vilterp:
Reviewed 1 of 3 files at r1, 2 of 2 files at r2.

---

*[pkg/sql/information_schema.go, line 310 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111#-Kz6ceA04kdoLK0LlU8n:-KzypabBbC8tjdpzfp8T:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/753c657975ca62e14303dfa726da6249334e0c6f/pkg/sql/information_schema.go#L310)):*

you can get rid of this comment now!
</blockquote></details>

Done.

---

*[pkg/sql/information_schema.go, line 314 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111#-Kz6gr0Bcg4XsLP3dVAE:-KzypbNkE5DezLQHcmOA:b-s3oksw) ([raw file](https://github.com/cockroachdb/cockroach/blob/753c657975ca62e14303dfa726da6249334e0c6f/pkg/sql/information_schema.go#L314)):*

I notice that Postgres's `information_schema.tables` doesn't include sequence while ours does, was that intentional?
</blockquote></details>

Done. Not intentional; fixed.

---

*[pkg/sql/logictest/testdata/logic_test/information_schema, line 936 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111#-Kz695QMkqqYh9TVXYLI:-KzypdzOhmvNBGgzQYeL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/753c657975ca62e14303dfa726da6249334e0c6f/pkg/sql/logictest/testdata/logic_test/information_schema#L936)):*

You should also test
- before any sequences are made
- what you see when you switch databases
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20111)*

pkg/sql/information_schema.go
```diff
@@ -525,6 +542,9 @@ CREATE TABLE information_schema.tables (
 );`,
 	populate: func(ctx context.Context, p *planner, prefix string, addRow func(...tree.Datum) error) error {
 		return forEachTableDesc(ctx, p, prefix, func(db *sqlbase.DatabaseDescriptor, table *sqlbase.TableDescriptor) error {
+			if table.IsSequence() {
```
justinj: No super strong feelings but it feels to me like this might merit inverting and extracting the logic somehow, like `IsRegularTable`? Or just something so that this table code doesn't have to consider the special case of sequences explicitly (or maybe since sequences are such a special case it's fine, I'm not sure).

justinj (approved):
LGTM

pkg/sql/information_schema.go
vilterp: I see your point — `IsTable` does that, but in `information_schema.tables` we actually don't want that, since it also returns views.

If we had a few more cases like this I'd do a refactor, but it doesn't seem that bad now. As refactors go, I think creating an enum for descriptor types and replacing if/else trees with switch statements would get more mileage.

vilterp:
@justinj I snuck in another commit to add `pg_catalog.pg_sequence`, since both the implementation and tests for it are practically the same.

justinj:
New commit LGTM as well!

PR #20112, opened 2017-11-17
sql: enforce an order to deflake alter_table logic test

benesch:


justinj:
LGTM (you can also fix something like this by adding `rowsort` after the typestring, which will pre-sort both the actual and expected results in cases where you'd rather not change the query)

benesch:
Is `rowsort` preferable? I went back and forth on that for a while and couldn't uncover any conclusive evidence on when to ORDER BY and when to `rowsort`.

justinj:
I don't know if anyone really cares! I usually opt for it because it seems more reasonable to change the behaviour of the test runner than the query for something like this, but I don't have any strong feelings.

benesch:
Ok! 👍 

If this CI cycle is happy I'm going to merge as in, but keeping your advice in mind for the future.

benesch:
And thank you for the quick review!

tschottdorf:
Did this close any issues?

benesch:
Don't believe so. It only ever failed an unrelated PR build of mine AFAICT.

PR #20116, opened 2017-11-17
cloud: add RBAC config for RBAC-enabled Secure Kubernetes Cluster

### LINKED ISSUE #20065: cloud: Secure Kubernetes certificate manager doesn't work on clusters with RBAC enabled
Our Kubernetes certificate manager (https://github.com/cockroachdb/k8s/tree/master/request-cert) relies on being able to read and write secrets via the Kubernetes API, which isn't allowed by default when [RBAC](https://kubernetes.io/docs/admin/authorization/rbac/) is enabled. This leads to the init container failing with the error:

```
2017/11/14 21:55:38 failed to read from secrets: secrets "default.node.cockroachdb-0" is forbidden: User "system:se
rviceaccount:default:default" cannot get secrets in the namespace "default"
```

We don't have this problem on GKE because for backwards compatibility reasons they let use cases like ours keep working, but [their docs](https://cloud.google.com/kubernetes-engine/docs/role-based-access-control) make it sound like on 1.8+ clusters (which will become the default very soon) we'll have the same problem, so this is something we do need to get working.

acoshift:
Hello, I try to run CockroachDB on RBAC-enabled Kubernetes Cluster. Hope this config can help other who have problem like me too.
This might fixed #20065 too ?

### DIFF
--- a/cloud/kubernetes/README.md
+++ b/cloud/kubernetes/README.md
@@ -333,7 +333,7 @@ Because all of the resources in this example have been tagged with the label `ap
 we can clean up everything that we created in one quick command using a selector on that label:
 
 ```shell
-kubectl delete statefulsets,pods,persistentvolumes,persistentvolumeclaims,services,poddisruptionbudget -l app=cockroachdb
+kubectl delete statefulsets,pods,persistentvolumes,persistentvolumeclaims,services,poddisruptionbudget,jobs,rolebinding,clusterrolebinding,role,clusterrole,serviceaccount -l app=cockroachdb
 ```
 
 If running in secure mode, you'll want to cleanup old certificate requests:
--- a/cloud/kubernetes/README.md
+++ b/cloud/kubernetes/README.md
@@ -86,7 +86,7 @@ Set up your cluster following the
 You can either set up your cluster following the
 [instructions provided in the Kubernetes docs](https://kubernetes.io/docs/getting-started-guides/gce/)
 or by using the hosted
-[Container Engine](https://cloud.google.com/container-engine/docs) service:
+[Kubernetes Engine](https://cloud.google.com/container-engine/docs) service:
 
 ```shell
 gcloud container clusters create NAME
--- a/cloud/kubernetes/README.md
+++ b/cloud/kubernetes/README.md
@@ -86,11 +86,27 @@ Set up your cluster following the
 You can either set up your cluster following the
 [instructions provided in the Kubernetes docs](https://kubernetes.io/docs/getting-started-guides/gce/)
 or by using the hosted
-[Kubernetes Engine](https://cloud.google.com/container-engine/docs) service:
+[Google Kubernetes Engine (GKE)](https://cloud.google.com/container-engine/docs) service:
 
 ```shell
 gcloud container clusters create NAME
 ```
+
+If you're using GKE and want to run a secure cluster, one extra step is
+required. A limitation in GKE's Role-Based Access Control (RBAC) integration
+necessitates running a special command in order to let you create the RBAC
+roles that CockroachDB needs to manage certificates.
+
+1. Get the email address associated with your Google Cloud account by running:
+   ```shell
+   `gcloud info | grep Account`
+   ```
+2. Run [the following command from the GKE
+   documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#prerequisites_for_using_role-based_access_control)
+   with that email address:
+   ```shell
+   kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=<your.google.cloud.email@example.org>
+   ```
 
 ### On Azure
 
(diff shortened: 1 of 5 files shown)


a-robinson:
Thanks for the contribution, @acoshift! I'll take this for a spin soon. And @a6802739, if you have comments or thoughts on this based on what you've done for #20065, I'd appreciate them.

a6802739:
@a-robinson, when I want to tackle with #20065, I just want to learn more about Kubernetes. 
this is great, Thanks!

cloud/kubernetes/rbac.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cockroachdb
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
a-robinson: GKE unfortunately requires an extra step before creating this `Role` and the `ClusterRole` is permitted (https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#prerequisites_for_using_role-based_access_control). I can update the README/docs for this if you don't want to.

a-robinson (commented):
Thanks a ton for the contribution, @acoshift! I've gone through, learned about, and tried some of this for myself, and it looks like exactly what we need.

I'd prefer to just always use RBAC in secure mode, though. Is there a reason not to? If not, would you mind modifying your commit to just add the `serviceAccountName` to all the existing `-secure.yaml` files and add the contents of `rbac.yaml` to `cockroachdb-statefulset-secure.yaml`? If you're too busy just let me know and I can take care of it.

acoshift:
@a-robinson Hello, I've update the yaml. Can you add comments, and prerequisites for GKE ?
My English is very bad 😅

a-robinson (approved):
Thanks @acoshift! I've squashed your commits together and added documentation about how to use this on GKE.

I'm going to update our docs site at the same time as this so I won't merge this quite yet. Also cc @mberhault so that I'm not effectively approving my own docs change.

cloud/kubernetes/README.md
```diff
@@ -116,13 +116,37 @@ kubectl create -f cluster-init.yaml
...
+#### Prerequisites
+
 **REQUIRED**: the kubernetes cluster must run with the certificate controller enabled.
 This is done by passing the `--cluster-signing-cert-file` and `--cluster-signing-key-file` flags.
 On minikube, you can tell it to use the minikube-generated CA by specifying:
 ```shell
 minikube start --extra-config=controller-manager.ClusterSigningCertFile="/var/lib/localkube/ca.crt" --extra-config=controller-manager.ClusterSigningKeyFile="/var/lib/localkube/ca.key"
 ```
 
+**REQUIRED**: because secure mode creates certificate signing requests and
```
mberhault: This paragraph is a bit rambly.

cloud/kubernetes/README.md
```diff
@@ -116,13 +116,37 @@ kubectl create -f cluster-init.yaml
...
+pods to use. The caller of the below `kubectl create` command must have
+sufficient privileges to create such `Role`s. If you get an error that says
+"attempt to grant extra privileges" when attempting to create the CockroachDB
+cluster the below command, your user not have the necessary privileges. [On
+Google Kubernetes Engine (GKE), you must specifically give yourself the ability
+to create new RBAC
+roles](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#prerequisites_for_using_role-based_access_control)
+by running:
+```shell
+kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=<your.google.cloud.email@example.org>
```
mberhault: This is a GKE-specific thing, right? It probably belongs in the `On GCE` section.

a-robinson: Yes, moved.

cloud/kubernetes/README.md
a-robinson: Copied the text from my docs change, which was a bit less rambly. Look better?

a-robinson (commented):
TFTR!

cloud/kubernetes/README.md
```diff
@@ -86,12 +86,28 @@ Set up your cluster following the
...
-[Container Engine](https://cloud.google.com/container-engine/docs) service:
+[Google Kubernetes Engine (GKE)](https://cloud.google.com/container-engine/docs) service:
 
 ```shell
 gcloud container clusters create NAME
 ```
 
+If you're using GKE and want to run a secure cluster, one extra step is
+required. A limitation in GKE's Role-Based Access Control (RBAC) integration
+necessatitates running a special command in order to let you create the RBAC
```
mberhault: s/necessaahhh-I-can't-stop-typing/necessitates/

mberhault (commented):
LGTM

cloud/kubernetes/README.md
a-robinson: :)

a-robinson:
Docs updates done in https://github.com/cockroachdb/docs/pull/2281

PR #20121, opened 2017-11-17
acceptance: handle macOS group IDs

benesch:
On macOS, users typically have a GID in the reserved ID space, which
means the acceptance test image complains when it can't automatically
add a group with that ID. Check for this case, handle it when we can,
and complain earlier with a better message when we can't.


benesch:
ping @tschottdorf!

tbg (approved):
🔬 🐶

PR #20124, opened 2017-11-17
core: add preamble format

mberhault:
Encryption at rest: part 1 of N.

Add preamble format for on-disk rocksdb instances:
* new field in the `--store` flag: `format=preamble` (empty defaults to `classic`)
* new on-disk version and added format field
* proto definition of preamble
* plaintext (no-op) block cipher stream

The on-disk version gets set to 2 iff we use the preamble format. This allows:
* old binaries to keep reading classic format files created using new binaries
* old binaries to fail with version check on preamble formats
* new binaries to handle preamble/classic properly

Eventually, we may want to always write version 2. We may need to give it some time to give enough people to upgrade their binaries.

A few todos to resolve before merging this:
* testing at the C++ level
* is my message length correct? protos use little-endian, it's a little silly to write the length in network byte order
* better error codes would be good. No built-in `errorf` in `rocksdb::Status`


c-deps/libroach/preamble.h
```diff
@@ -0,0 +1,89 @@
...
+  rocksdb::EncryptionProvider* passthrough_provider_;
+};
+
+// Blocksize for the plaintext cipher stream.
+// TODO(mberhault): we can pick anything we like. What's good?
+const static size_t plaintextBlockSize = 4096;
+
+// PlaintextCipherStream implements BlockAccessCipherStream with
+// no-op encrypt/decrypt operations.
+class PlaintextCipherStream final : public rocksdb::BlockAccessCipherStream {
```
mberhault: I forgot how this works. How do I get a concrete class without overriding the pure protected methods?

pkg/cli/flags.go
```diff
@@ -197,8 +197,6 @@ func init() {
 	// special severity value DEFAULT instead.
 	pf.Lookup(logflags.LogToStderrName).NoOptDefVal = log.Severity_DEFAULT.String()
 
-	// Security flags.
```
mberhault: Just saw this in passing, that comment have nothing to do here.

c-deps/libroach/preamble.h
benesch: Er, isn't the point that you _have_ to override the pure protected methods to get a concrete class?

mberhault: yeah, I know. I suppose my complaint is about those protected methods being required.

benesch: Yeah, the encryption support in RocksDB that landed seems pretty specialized for the use case it was extracted from.

bdarnell:
LGTM, aside from doubts about whether we should be using this EncryptedEnv vs building our own.

---

Reviewed 8 of 10 files at r1, 6 of 7 files at r2, 2 of 4 files at r3, 1 of 1 files at r4, 3 of 3 files at r5.

---

*[c-deps/libroach/db.cc, line 1705 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzErICY4ZPxgGNMu-uH:-KzErICY4ZPxgGNMu-uI:b-jz8gg1) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/db.cc#L1705)):*
> ```cpp
>   PreambleHandler* preamble = nullptr;
>   if (db_opts.use_preamble) {
>     // The caller makes sure we're not an in-memory DB.
> ```

What's wrong with using the preamble format in memory?

---

*[c-deps/libroach/preamble.cc, line 40 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEs3vwhCnc_M2HoOEP:-KzEs3vwhCnc_M2HoOEQ:b-59q4vn) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.cc#L40)):*
> ```cpp
> 
>   // Determine the serialized length and size of the length prefix.
>   // TODO(mberhault): protobuf encoding is little-endian, so this is a little weird.
> ```

Weird in that you're writing this big-endian while protobuf uses little-endian? I don't think that's too bad; I'd rather use big-endian here and not even know what protobuf uses internally. 

---

*[c-deps/libroach/preamble.cc, line 42 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEs-UOo4fGShEMwTJp:-KzEs-UOo4fGShEMwTJq:b-c2asht) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.cc#L42)):*
> ```cpp
>   // TODO(mberhault): protobuf encoding is little-endian, so this is a little weird.
>   assert(byte_size < UINT16_MAX);
>   uint16_t message_size = htons(byte_size);
> ```

s/message_size/encoded_size/g

---

*[c-deps/libroach/preamble.cc, line 46 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEsVJ86gxOUG1HANlt:-KzEsVJ86gxOUG1HANlu:b-lqwhzz) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.cc#L46)):*
> ```cpp
> 
>   if ((byte_size + num_length_bytes) > prefixLength ) {
>     // TODO(mberhault): is NoSpace the right thing? Nothing really fits.
> ```

I'd go with Corruption, same as below.

---

*[c-deps/libroach/preamble.h, line 24 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEnjhb2XxOxOsk__PI:-KzEnjhb2XxOxOsk__PJ:b-xzyb0c) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.h#L24)):*
> ```objc
> // WARNING: changing this will result in incompatible on-disk format.
> // The preamble length must fit in a uint16_t.
> const static size_t defaultPreambleLength = 4096;
> ```

s/default/k/

This is also required to be a multiple of the page size, right? Mention that in the comment.

Unless you need them for tests, both constants in this header probably belong in the .cc file instead.

---

*[c-deps/libroach/preamble.h, line 48 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEq35bNhU-rBvWNIIV:-KzEq35bNhU-rBvWNIIW:b-tccfbq) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.h#L48)):*
> ```objc
> // Blocksize for the plaintext cipher stream.
> // TODO(mberhault): we can pick anything we like. What's good?
> const static size_t plaintextBlockSize = 4096;
> ```

It looks like every write gets padded to a multiple of the block size, so setting it too high would waste space (but setting it too low wastes CPU to iterate over the data to do nothing in small chunks). 

I'd set this to 16 for now to match the block size of the AES cipher version. In the long run, we should probably refactor the EncryptionProvider interface so it can return something other than a BlockAccessCipherStream that can be an efficient no-op. 

---

*[c-deps/libroach/preamble.h, line 76 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEp4CQ68mW6azIBWnv:-KzEp4CQ68mW6azIBWnw:blvktvn) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.h#L76)):*
> ```objc
>   // Length of data is equal to BlockSize();
>   virtual rocksdb::Status EncryptBlock(uint64_t blockIndex, char *data, char* scratch) override {
> 		return rocksdb::Status::OK();
> ```

Don't mix tabs and spaces. We use two-space indents in our c++ code.

---

*[pkg/cli/cliflags/flags.go, line 448 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEn9b6xicn4V0P6DBB:-KzEn9b6xicn4V0P6DBC:b1ni2mv) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/pkg/cli/cliflags/flags.go#L448)):*
> ```Go
> </PRE>
> The "format" field specifies the underlying file format for the store.
> Allowed values are "classic" (default if omitted), or "preamble". Preamble
> ```

One day the preamble format will be the default and then the "classic" behind whatever comes next. Let's just give these numbers instead of names. 

---

*[pkg/storage/engine/version.go, line 26 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEsxfukpKZfuEdjhts:-KzEsxfukpKZfuEdjhtt:b-ocakn0) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/pkg/storage/engine/version.go#L26)):*
> ```Go
> 
> type storageVersion int
> type formatVersion int
> ```

This two-level versioning scheme is confusing and needs more documentation. Or maybe two version numbers is the wrong approach - is the "format" a sequence of versions or is it a set of features that may or may not be present?

It's really unfortunate that EncryptedEnv doesn't give us any feasible migration path so we'll have to keep supporting the old format forever. It ought to be possible to make an Env that lets us encode the format somehow (maybe in the filename?) instead of an all-or-nothing cutover. Then we could treat this like a normal version upgrade and eventually get all files written with the preamble format. 

---

*[pkg/storage/engine/version_test.go, line 84 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEw5hUYZMU05zuzCjz:-KzEw5hUYZMU05zuzCk-:b-q8xbia) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/pkg/storage/engine/version_test.go#L84)):*
> ```Go
> 
> // TestOldVersionFormat verifies that new version format (not numbers, but actual json fields)
> // parses properly when using the older version data structures.
> ```

What do you mean? Just that the new Format field is ignored by old code?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-:-KzEwdyfM0Mp3dl66fmQ:bi9ao36)*

mberhault:

---

*[c-deps/libroach/db.cc, line 1705 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzErICY4ZPxgGNMu-uH:-KzF-8c40zPeCr6rHzLS:big881e) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/db.cc#L1705)):*

What's wrong with using the preamble format in memory?
</blockquote></details>

I haven't actually checked that it works properly, but there's no reason to right now.
If we decide to make the preamble the default format, then it may be simpler. I don't think we'll do encryption on it though

---

*[c-deps/libroach/preamble.cc, line 40 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEs3vwhCnc_M2HoOEP:-KzF-K9ceBuehQrMbenC:bp1pgwf) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.cc#L40)):*

Weird in that you're writing this big-endian while protobuf uses little-endian? I don't think that's too bad; I'd rather use big-endian here and not even know what protobuf uses internally. 
</blockquote></details>

Ok, removing the TODO.

---

*[c-deps/libroach/preamble.cc, line 42 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEs-UOo4fGShEMwTJp:-KzF-cGo3xpox_Ic_YJm:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.cc#L42)):*

s/message_size/encoded_size/g
</blockquote></details>

Done.

---

*[c-deps/libroach/preamble.cc, line 46 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEsVJ86gxOUG1HANlt:-KzF-dKBLpToXEWb86Bt:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.cc#L46)):*

I'd go with Corruption, same as below.
</blockquote></details>

Done.

---

*[c-deps/libroach/preamble.h, line 24 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEnjhb2XxOxOsk__PI:-KzF-gsIQdYxSGLYRYCP:bvuzjft) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.h#L24)):*

s/default/k/

This is also required to be a multiple of the page size, right? Mention that in the comment.

Unless you need them for tests, both constants in this header probably belong in the .cc file instead.
</blockquote></details>

Renamed, added "page size" command (not actually a requirement, just preferred).
Moved both constants to .cc and moved entire PlaintextCipherStream.

---

*[c-deps/libroach/preamble.h, line 48 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEq35bNhU-rBvWNIIV:-KzF3S2jG9t5Yjikrbk6:bowlbzs) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.h#L48)):*

It looks like every write gets padded to a multiple of the block size, so setting it too high would waste space (but setting it too low wastes CPU to iterate over the data to do nothing in small chunks). 

I'd set this to 16 for now to match the block size of the AES cipher version. In the long run, we should probably refactor the EncryptionProvider interface so it can return something other than a BlockAccessCipherStream that can be an efficient no-op. 
</blockquote></details>

Done.
By implementing as a `BlockAccessCipherStream`, we already avoid a lot of overhead. We'll need to benchmark for 1) good default size, and 2) whether to strip out more of the env_encryption wrapper.

---

*[c-deps/libroach/preamble.h, line 76 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEp4CQ68mW6azIBWnv:-KzF493wD5ALgbvatpbP:btnn5y5) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/c-deps/libroach/preamble.h#L76)):*

Don't mix tabs and spaces. We use two-space indents in our c++ code.
</blockquote></details>

Done. I was copy/pasting. I miss `gofmt`

---

*[pkg/cli/cliflags/flags.go, line 448 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEn9b6xicn4V0P6DBB:-KzF4H-BJSdrhMELJ7WX:b-guqlbd) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/pkg/cli/cliflags/flags.go#L448)):*

One day the preamble format will be the default and then the "classic" behind whatever comes next. Let's just give these numbers instead of names. 
</blockquote></details>

We can certainly use numbers instead, but this feels more user-friendly. The docs will have to mention the need for the preamble format, skipping a level of indirection ("which number is preamble again?") seems convenient.

---

*[pkg/storage/engine/version.go, line 26 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEsxfukpKZfuEdjhts:-KzF4_z5AmlyrxRs9zso:bm48s1h) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/pkg/storage/engine/version.go#L26)):*

This two-level versioning scheme is confusing and needs more documentation. Or maybe two version numbers is the wrong approach - is the "format" a sequence of versions or is it a set of features that may or may not be present?

It's really unfortunate that EncryptedEnv doesn't give us any feasible migration path so we'll have to keep supporting the old format forever. It ought to be possible to make an Env that lets us encode the format somehow (maybe in the filename?) instead of an all-or-nothing cutover. Then we could treat this like a normal version upgrade and eventually get all files written with the preamble format. 
</blockquote></details>

The format is a "feature" of the storage. We still want to be able to have "version 2 with preamble" and "version 2 without preamble'.
The part that gets tricky here is that we don't want to wreck old binaries running against a "version 2 without preamble".

As for a different env, I mentioned this is the RFC. We could perform our own mapping of `filename -> format`, but that seems a bit fragile. The one major advantage of that solution is that we could use the classic format for plaintext.
However, that still would not solve the version issue, you would need to make sure you can handle the preamble format, requiring a version bump.

---

*[pkg/storage/engine/version_test.go, line 84 at r5](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-KzEw5hUYZMU05zuzCjz:-KzF5REPuPmnXLnLRuYk:b-psitjo) ([raw file](https://github.com/cockroachdb/cockroach/blob/c8bb7cf02e2f2af0d5f88f73e14a8842eeea46d0/pkg/storage/engine/version_test.go#L84)):*

What do you mean? Just that the new Format field is ignored by old code?
</blockquote></details>

Yup. Technically the json parsing code does not fail on unknown fields, but that's about to become an option. If someone turns that on, I want the test to fail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124)*

mberhault:
I've added a "custom env" section in the [rfc](https://github.com/cockroachdb/cockroach/pull/19785) under "Rationale and alternatives".

---

Review status: 18 of 20 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20124#-:-KzFCG_3itON6ok_0vtg:b-gbf7pa)*

mberhault:
Closing in favor of the switching env. See updated RFC.

PR #20127, opened 2017-11-17
build: update README on deploying Docker image

nvb:
Running `./build/builder.sh make build TYPE=release-linux-gnu` will
create a binary called `cockroach-linux-2.6.32-gnu-amd64`, not
`cockroach`.

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -37,7 +37,7 @@ usual fashion. To be more specific, the steps to do this are:
 
 ```
 go/src/github.com/cockroachdb/cockroach $ ./build/builder.sh make build TYPE=release-linux-gnu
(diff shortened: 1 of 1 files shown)


a-robinson (approved):

nvanbenschoten:
Just to confirm, running `./build/builder.sh make build TYPE=release-linux-gnu` on any system, even linux, will create a binary called `cockroach-linux-2.6.32-gnu-amd64`, right?

a-robinson:
Judging by the Makefile, yes. It'll be easy for this documentation to fall out of date, though, if the suffix changes.

nvanbenschoten:
Yeah that's a good point. It looks like there are a number of other places where the suffix is hardcoded too, though, so a global find-replace will be required then anyway.

PR #20132, opened 2017-11-17
build: split style_test into a proper lint package

benesch:
build/style_test.go is rather substantial and deserves its own package.
Move it to pkg/testutils/lint. This has the nice side effect of ensuring the
linter lints itself.


benesch:
@arjunravinarayan, can I toss this one your way? Who will lint the linter 'n all.

knz (approved):
cc @nvanbenschoten

knz:
ugh we probably want to ensure that `make test` doesn't run the linter. how do you do that?

mberhault:
Why `pkg/lint`? this seems a bit high level for what it is. How about `pkg/testutils/lint` instead?

benesch:
I have additional concerns about how util and testutils are also dumping
grounds (and don’t even use consistent pluralization), but that’s a gripe
for another day. testutils/lint works for me.

@knz, the linter tests are hidden behind the "lint" build tag to prevent
being picked up by `go test ./pkg/...`. I know we're typically opposed to
build tags, but the linter has always hidden itself this way so seems fine
enough.

benesch:
Ok, moved to `pkg/testutils/lint`. TFTRs!

mberhault:
LGTM. Thanks. I agree that `testutils` is getting a bit "random", but it still feels better than `pkg/`

knz:
LGTM thanks

benesch:
You bet—thanks for the reviews!

For posterity, I should say: I have fewer qualms about `testutils`, which is a nice place to put symbols that have to be public to avoid circular dependencies but [should never be linked into the main binary](https://github.com/cockroachdb/cockroach/blob/master/pkg/cli/flags_test.go#L54). `util`, on the other hand, is the wild west. It's not clear there's any rhyme or reason to how things are divided between `util` and its subpackages.

knz:
I'd say the default good direction is to separate in sub-package. Everything in `util` that's not yet separate in packages is where the mess is.

arjunravinarayan:
LGTM. Thanks for doing this. In general I'm happy to help with this sort of thing, but I was away on vacation whilst this was under review.

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20132#-:-KzzGYFe7X9WUxL5aV-S:bra7nqi)*

PR #20133, opened 2017-11-17
Makefile: make TESTS without PKG check more targeted

benesch:
It's valid to specify TESTS without also specifying PKG when running
`make acceptance` and `make lint`. Due to an oddity of how `make
acceptance` recursively invokes Make, it ends up calling `make build
TESTS=...`, which trips the check as-written.

Adjust the check to use a blacklist of affected targets rather than a
whitelist of unaffected targets. Now the check only triggers when
running Make against a test, bench, or stress target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ endif
 
 # Prevent invoking make with a specific test name without a constraining
 # package.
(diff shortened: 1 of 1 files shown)


justinj:
LGTM!

Issue #20137, opened 2017-11-17
perf: slight performance regression in kv workload


jordanlewis:
Seems to have been introduced by the MultiRowFetcher change in db2aedc884b391dd04c34516e896b6d783118fe6.

The below measurements were taken on my local laptop with a fresh `cockroach-data` directory and `kv --read-percent=100`:

Before: (3844bbd30)
```
_elapsed___errors__ops/sec(inst)___ops/sec(cum)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
      1s        0        19658.8        19658.0      0.7      2.0      4.5     14.2
      2s        0        19369.4        19514.4      0.7      2.2      4.7     18.9
      3s        0        19354.1        19461.0      0.7      2.2      4.5     15.2
      4s        0        18888.8        19318.1      0.7      2.4      4.7     15.2
      5s        0        18500.7        19154.5      0.7      2.5      4.7     16.8
      6s        0        18582.2        19059.6      0.7      2.5      4.7     26.2
      7s        0        18282.6        18948.6      0.7      2.5      4.7     30.4
      8s        0        18595.5        18904.4      0.7      2.5      4.7     13.6
      9s        0        18990.6        18913.9      0.7      2.4      4.5     17.8
     10s        0        18701.9        18892.7      0.7      2.4      4.5     21.0
^C
_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   10.4s        0         195247        18847.2      0.8      0.7      2.4      4.7     30.4

BenchmarkLoadgenKV/readPercent=100/splits=0/concurrency=16/duration=0s    195247             53058 ns/op
```

After (db2aedc884b391dd04c34516e896b6d783118fe6):
```
_elapsed___errors__ops/sec(inst)___ops/sec(cum)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
      1s        0        18932.3        18931.7      0.7      2.2      4.5     11.0
      2s        0        18837.4        18884.8      0.7      2.4      4.7     17.8
      3s        0        19054.6        18941.3      0.7      2.2      4.5     16.3
      4s        0        18532.5        18839.3      0.7      2.4      4.7     25.2
      5s        0        18305.5        18732.5      0.7      2.5      4.5     21.0
      6s        0        18487.3        18691.6      0.7      2.4      4.7     16.8
[...]

PR #20138, opened 2017-11-17
sql: MultiRowFetcher does less work in common case

jordanlewis:
When the MultiRowFetcher only gets 1 table as input, it doesn't need to
compute any table signatures during initialization.

Updates #20137


richardwu:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20138#-:-KzBVBD5mNY1li5Q7F4Q:bnfp4nl)*

PR #20145, opened 2017-11-18
sql: use binary search to fetch key in JSON

justinj:
I finally sat down to benchmark various operators and realized that this
was desperately in need of this.

Benchstat result for always binary searching vs. linear scan:

    name                                   old time/op    new time/op    delta
    FetchKey/object_size_1/fetch_key-8       51.6ns ± 2%    59.9ns ± 3%  +16.07%  (p=0.000 n=9+10)
    FetchKey/object_size_10/fetch_key-8      94.6ns ± 2%   101.3ns ± 2%   +7.04%  (p=0.000 n=10+10)
    FetchKey/object_size_100/fetch_key-8      346ns ± 2%     142ns ± 1%  -58.93%  (p=0.000 n=10+10)
    FetchKey/object_size_1000/fetch_key-8    2.91µs ± 1%    0.20µs ± 2%  -93.14%  (p=0.000 n=9+10)

Benchstat result for binary searching for objects > 20 elements and
linear searching otherwise vs. pure linear scan:

    name                                   old time/op    new time/op    delta
    FetchKey/object_size_1/fetch_key-8       51.6ns ± 2%    50.9ns ± 2%     ~     (p=0.092 n=10+10)
    FetchKey/object_size_10/fetch_key-8      93.1ns ± 1%    94.0ns ± 3%     ~     (p=0.078 n=10+10)
    FetchKey/object_size_100/fetch_key-8      349ns ± 2%     147ns ± 4%  -57.99%  (p=0.000 n=10+10)
    FetchKey/object_size_1000/fetch_key-8    2.94µs ± 2%    0.20µs ± 2%  -93.24%  (p=0.000 n=10+9)

I did a quick search to figure out a good cutoff, 20 seemed reasonable
on my laptop as the point where the two were more or less equal.

### DIFF
--- a/pkg/util/json/json.go
+++ b/pkg/util/json/json.go
@@ -514,14 +514,28 @@ func MakeJSON(d interface{}) (JSON, error) {
 	return nil, pgerror.NewError("invalid value %s passed to MakeJSON", d.(fmt.Stringer).String())
 }
 
+// This value was determined through some rough experimental results as a good
+// place to start doing binary search over a linear scan.
+const bsearchCutoff = 20
+
 func (j jsonObject) FetchValKey(key string) JSON {
-	for i := range j {
-		if string(j[i].k) == key {
-			return j[i].v
+	// For small objects, the overhead of binary search is significant and so
+	// it's faster to just do a linear scan.
+	if len(j) < bsearchCutoff {
+		for i := range j {
+			if string(j[i].k) == key {
+				return j[i].v
+			}
+			if string(j[i].k) > key {
+				break
+			}
 		}
-		if string(j[i].k) > key {
-			break
-		}
+		return nil
+	}
+
+	i := sort.Search(len(j), func(i int) bool { return string(j[i].k) >= key })
+	if i < len(j) && string(j[i].k) == key {
+		return j[i].v
 	}
 	return nil
 }
--- a/pkg/util/json/json_test.go
+++ b/pkg/util/json/json_test.go
@@ -328,6 +328,25 @@ func TestJSONFetch(t *testing.T) {
 	}
 }
 
+func TestJSONRandomFetch(t *testing.T) {
+	rng := rand.New(rand.NewSource(timeutil.Now().Unix()))
+	for i := 0; i < 1000; i++ {
+		// We want a big object to trigger the binary search behaviour in FetchValKey.
+		j, err := Random(1000, rng)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if obj, ok := j.(jsonObject); ok {
+			// Pick a key:
+			idx := rand.Intn(len(obj))
+			result := obj.FetchValKey(string(obj[idx].k))
+			if result.Compare(obj[idx].v) != 0 {
+				t.Fatalf("%s: expected fetching %s to give %s got %s", obj, obj[idx].k, obj[idx].v, result)
+			}
+		}
+	}
+}
+
 func TestJSONFetchIdx(t *testing.T) {
 	json := jsonTestShorthand
 	cases := map[string][]struct {
@@ -825,5 +844,31 @@ func TestNegativeRandomJSONContains(t *testing.T) {
 		if realResult != slowResult {
 			t.Fatal("mismatch for document " + j1.String() + " @> " + j2.String())
 		}
+	}
+}
+
+func BenchmarkFetchKey(b *testing.B) {
+	for _, objectSize := range []int{1, 10, 100, 1000} {
+		b.Run(fmt.Sprintf("object size %d", objectSize), func(b *testing.B) {
+			keys := make([]string, objectSize)
+
+			obj := make(map[string]interface{})
+			for i := 0; i < objectSize; i++ {
+				key := fmt.Sprintf("key%d", i)
+				keys = append(keys, key)
+				obj[key] = i
+			}
+			j, err := MakeJSON(obj)
+			if err != nil {
+				b.Fatal(err)
+			}
+
+			// TODO(justin): add benchmarks for fetching from still-encoded objects.
+			b.Run("fetch key", func(b *testing.B) {
+				for i := 0; i < b.N; i++ {
+					j.FetchValKey(keys[rand.Intn(len(keys))])
+				}
+			})
+		})
 	}
 }


vivekmenezes:
Awesome!

jordanlewis:
Reviewed 2 of 2 files at r1.

---

*[pkg/util/json/json_test.go, line 841 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzOz4uqtKXpmD1XT5Xm:-KzOz4uqtKXpmD1XT5Xn:b-d294c) ([raw file](https://github.com/cockroachdb/cockroach/blob/66d317a6c9a6156c6602d794d5475fc355976518/pkg/util/json/json_test.go#L841)):*
> ```Go
> 				obj[key] = i
> 				if lastKey == "" || key > lastKey {
> 					lastKey = key
> ```

This benchmark seems to always test retrieval of the last key in the object. Is that true or am I missing something? That could definitely have biased your finding that binary search is always faster than linear search in this algorithm.

---

*[pkg/util/json/json_test.go, line 850 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzOyZq5DDzmbrWH8YxK:-KzOyZq5DDzmbrWH8YxL:b-wy0khs) ([raw file](https://github.com/cockroachdb/cockroach/blob/66d317a6c9a6156c6602d794d5475fc355976518/pkg/util/json/json_test.go#L850)):*
> ```Go
> 
> 			// This will eventually also include benchmarks for fetching from a
> 			// still-encoded JSON object.
> ```

nit: this kind of comment should be phrased as a TODO - it's not really useful to have this without an assignee. You could move this comment into the commit message instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145)*

jordanlewis:
Also you should use `benchstat` to compare the benchmarks. There's a neat script that automates this for you in `./scripts/bench`. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-:-KzOzUNIcnkgeK5eAKpu:bsnz2vz)*

justinj:
Ok, included a `benchstat`! I also ended up adding a cutoff for doing a linear scan for small objects which might be a premature optimization, but the logic is pretty straightforward, though I added a test for it as well. I think this is a worthwhile optimization because I suspect objects in the <20 key range are a somewhat common use-case.

---

---

*[pkg/util/json/json_test.go, line 841 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzOz4uqtKXpmD1XT5Xm:-KzP-Gki6SFXqmaz-yiX:b-hzv1pv) ([raw file](https://github.com/cockroachdb/cockroach/blob/66d317a6c9a6156c6602d794d5475fc355976518/pkg/util/json/json_test.go#L841)):*

This benchmark seems to always test retrieval of the last key in the object. Is that true or am I missing something? That could definitely have biased your finding that binary search is always faster than linear search in this algorithm.
</blockquote></details>

Yes that's true, I was testing the worst case. Is there a good methodology to make this more fair? Maybe testing for keys at various percentages through the object? Is it fair in a benchmark to just pick a random key? I updated the benchmark to pick random keys and the results were far less extreme, as one would expect.

---

*[pkg/util/json/json_test.go, line 850 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzOyZq5DDzmbrWH8YxK:-KzP-go8OKfJOfmw228j:b-w2iqy4) ([raw file](https://github.com/cockroachdb/cockroach/blob/66d317a6c9a6156c6602d794d5475fc355976518/pkg/util/json/json_test.go#L850)):*

nit: this kind of comment should be phrased as a TODO - it's not really useful to have this without an assignee. You could move this comment into the commit message instead.
</blockquote></details>

Good point, I switched it to a TODO because without the comment in the source I think the fact that this uses a subbenchmark is odd.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-:-KzPL6QpC0tvjrei_LrE:bsbk0oq)*

jordanlewis:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---

*[pkg/util/json/json.go, line 518 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzPPbzXiGoYt9J-s6Y1:-KzPPbzXiGoYt9J-s6Y2:b-75d7o8) ([raw file](https://github.com/cockroachdb/cockroach/blob/36a5ed28250a78f07bdc6bcdbb811a26ae9d8428/pkg/util/json/json.go#L518)):*
> ```Go
> 
> func (j jsonObject) FetchValKey(key string) JSON {
> 	// Do linear scan in this case!
> ```

This comment can be improved. I suggest explaining the reasoning of this decision as well. Something like "Binary search has better time complexity for searching the key list, but if the number of keys is small, the overhead of binary search dominates the runtime. Experimental testing shows that 16 is a good cutoff."

---

*[pkg/util/json/json_test.go, line 841 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzOz4uqtKXpmD1XT5Xm:-KzPQUk9vtg-h00Xjie7:bgucabl) ([raw file](https://github.com/cockroachdb/cockroach/blob/66d317a6c9a6156c6602d794d5475fc355976518/pkg/util/json/json_test.go#L841)):*

Yes that's true, I was testing the worst case. Is there a good methodology to make this more fair? Maybe testing for keys at various percentages through the object? Is it fair in a benchmark to just pick a random key? I updated the benchmark to pick random keys and the results were far less extreme, as one would expect.
</blockquote></details>

Dunno about fairness but random works for me!

---

*[pkg/util/json/json_test.go, line 850 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzOyZq5DDzmbrWH8YxK:-KzPQYiodWLKO6jxhdfT:b23ezzv) ([raw file](https://github.com/cockroachdb/cockroach/blob/66d317a6c9a6156c6602d794d5475fc355976518/pkg/util/json/json_test.go#L850)):*

Good point, I switched it to a TODO because without the comment in the source I think the fact that this uses a subbenchmark is odd.
</blockquote></details>

super-silly-nit (but good practice for TODOs in files that aren't tests and aren't fully in your working set I think): this TODO looks odd to my eyes because it's not an action? Like I was expecting `TODO(justin): add benchmarks for fetching from encoded json objects when they're ready` or whatever. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-:-KzPR8QfhFW1gew0cCvA:bnfp4nl)*

jordanlewis:
Oh please update the PR description with the updated commit message as well!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-:-KzPRAilzYjNCCtL8MKc:b-yrfy47)*

justinj:

---

*[pkg/util/json/json.go, line 518 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145#-KzPPbzXiGoYt9J-s6Y1:-KzPZb5gs6wQLHS7wgJW:b-6iz0g) ([raw file](https://github.com/cockroachdb/cockroach/blob/36a5ed28250a78f07bdc6bcdbb811a26ae9d8428/pkg/util/json/json.go#L518)):*

This comment can be improved. I suggest explaining the reasoning of this decision as well. Something like "Binary search has better time complexity for searching the key list, but if the number of keys is small, the overhead of binary search dominates the runtime. Experimental testing shows that 16 is a good cutoff."
</blockquote></details>

:/ I actually had a better comment here but it got lost in my benchmarking-branch switching, fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20145)*

justinj:
Ok, fixed all those, TFTR!

PR #20146, opened 2017-11-18
jepsen: Add disk fault injection nemesis

bdarnell:
Another temporary PR to get something i can test on teamcity

### DIFF
--- a/build/jepsen-common.sh
+++ b/build/jepsen-common.sh
@@ -40,6 +40,7 @@ function progress {
 }
 
 nemeses=(
+    "--nemesis disk"
     "--nemesis majority-ring"
     "--nemesis split"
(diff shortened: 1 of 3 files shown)


bdarnell:
OK, this is ready to go with jepsen-io/jepsen#235. The branch reference in main.tf will be reverted once I've updated the branch in our fork.

benesch (approved):
🎉 🎉 🎉

tschottdorf:
@bdarnell, what's the plan with this PR?

bdarnell:
This PR is obsolete. I want to push the disk nemesis upstream but Aphyr wants it in the main jepsen pacakge instead of a separate one and that means I need to update our tests to work with jepsen master, so I've got some yaks to shave....

PR #20151, opened 2017-11-20
sql: fix PARTITION BY multiple columns with window functions 

### LINKED ISSUE #20143: PARTITION BY doesn't support multiple columns
Is this a question, feature request, or bug report?

This is bug report

**BUG REPORT**

     grep -F '[config]' cockroach-data/logs/cockroach.log

   When log files are not available, supply the output of `cockroach version`
   and all flags/environment variables passed to `cockroach start` instead.

```
Build Tag:    v1.1.2
Build Time:   2017/11/02 20:53:00
Distribution: CCL
Platform:     darwin amd64
Go Version:   go1.9.2
C Compiler:   4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.38)
Build SHA-1:  c726df22152a771cdd1882679c772d4e7799440d
Build Type:   development
```

```
cockroach start --host=localhost --store=$HOME/cockroach-data/node1 --background
cockroach start --host=localhost --store=$HOME/cockroach-data/node2 --background --port=26258 --http-port=8081 --join=localhost:26257
cockroach start --host=localhost --store=$HOME/cockroach-data/node2 --background --port=26259 --http-port=8082 --join=localhost:26257
```

- What did you do?

I created a test table and values like the following:

```
create table data (
    id serial primary key,
    a VARCHAR,
    b VARCHAR
);

insert into data (a, b) values
('foo', 'a'),
('foo', 'b'),
('foo', 'b'),
('bar', 'b'),
('bar', 'a'),
('foo', 'b');
```
 
And ran sql which display rank like the folling:

```
SELECT *, rank() OVER (PARTITION BY a, b ORDER BY id DESC) FROM data;
```

- What did you expect to see?

I expect rank is partitioned by multiple columns (`a` and `b`).

- What did you see instead?

Rank is partitioned only by the last column (`b`).

```
+--------------------+-----+---+--------------------------------+
|         id         |  a  | b | rank() OVER (PARTITION BY a, b ORDER BY id DESC) |
+--------------------+-----+---+--------------------------------+
| 297992315749990401 | foo | a |                              2 |
| 297992315750023169 | foo | b |                              4 |
| 297992315750055937 | foo | b |                              3 |
| 297992315750088705 | bar | b |                              2 |
| 297992315750121473 | bar | a |                              1 |
| 297992315750154241 | foo | b |                              1 |
+--------------------+-----+---+--------------------------------+
```

I tested the same query on PosrgreSQL, and the result is expected one (partitioned by multiple columns).

nvb:
Fixes #20143.

This has been broken since 278e2e7. Instead of adding to the
`partitionIdxs` slice, we were replacing it for each new partitioning
column.

### DIFF
--- a/pkg/sql/logictest/testdata/logic_test/show_trace
+++ b/pkg/sql/logictest/testdata/logic_test/show_trace
@@ -13,10 +13,10 @@ query TTT
 SELECT span, message, operation FROM [SHOW TRACE FOR SESSION] WHERE message NOT LIKE e'%\n%' AND message NOT LIKE '%canceling context'
 ----
 (0,0)  === SPAN START: sql txn implicit ===  sql txn implicit
-(1,0)  === SPAN START: sql txn ===           sql txn implicit
-(1,0)  executing 1/5: BEGIN TRANSACTION      sql txn implicit
-(1,0)  executing 2/5: SELECT 1               sql txn implicit
-(1,0)  executing 3/5: COMMIT TRANSACTION     sql txn implicit
+(1,0)  === SPAN START: sql txn ===           sql txn
+(1,0)  executing 1/5: BEGIN TRANSACTION      sql txn
+(1,0)  executing 2/5: SELECT 1               sql txn
+(1,0)  executing 3/5: COMMIT TRANSACTION     sql txn
 (2,0)  === SPAN START: sql txn implicit ===  sql txn implicit
 (2,0)  executing 1/1: SELECT 2               sql txn implicit
 (3,0)  === SPAN START: sql txn implicit ===  sql txn implicit
--- a/pkg/sql/logictest/testdata/logic_test/window
+++ b/pkg/sql/logictest/testdata/logic_test/window
@@ -68,16 +68,15 @@ SELECT avg(k) OVER (PARTITION BY v) FROM kv ORDER BY 1
 5.5
 5.5
 
-# Not working until #12482 is fixed.
-#query R
-#SELECT avg(k) OVER (PARTITION BY kv.*) FROM kv ORDER BY 1
-#----
-#1
-#3
-#5
-#6
-#7
-#8
+query R
+SELECT avg(k) OVER (PARTITION BY kv.*) FROM kv ORDER BY 1
+----
+1
+3
+5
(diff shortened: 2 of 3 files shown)


knz:
:lgtm:

---

Reviewed 1 of 1 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20151#-:-KzO-TPG_EiP2TZmFMpo:bnfp4nl)*

knz:
please consider merging onto 1.1 too.

pkg/sql/logictest/testdata/logic_test/show_trace
```diff
@@ -13,10 +13,10 @@ query TTT
 SELECT span, message, operation FROM [SHOW TRACE FOR SESSION] WHERE message NOT LIKE e'%\n%' AND message NOT LIKE '%canceling context'
 ----
 (0,0)  === SPAN START: sql txn implicit ===  sql txn implicit
-(1,0)  === SPAN START: sql txn ===           sql txn implicit
-(1,0)  executing 1/5: BEGIN TRANSACTION      sql txn implicit
-(1,0)  executing 2/5: SELECT 1               sql txn implicit
-(1,0)  executing 3/5: COMMIT TRANSACTION     sql txn implicit
+(1,0)  === SPAN START: sql txn ===           sql txn
```
knz: I hadn't seen this part of the patch. When I reviewed the PR I saw a TC failure and I thought this was a flaky test. I don't understand why this needs to change.

nvanbenschoten: I didn't realize when I first made the change that `SHOW TRACE` [uses window functions](https://github.com/cockroachdb/cockroach/blob/6dc8f306719a0be1a0627fe3419315f73643b740/pkg/sql/show.go#L294) under the hood. This trace was actually incorrect before because of the bug this change fixes.

knz: Hah! Amazing. Figure, I wrote that code, I simply had forgotten. :) Carry on.

PR #20148, opened 2017-11-18
 cli/sql: bump the go-libedit dependency to fix signal handling on darwin (again)

knz:
First commit from  #20147.

Details can be found in the commit message at
knz/go-libedit@ae72256

Fixes #19132.

Gosh I hate macOS.

PR #20153, opened 2017-11-20
cli/sql: bump the go-libedit dependency to fix signal handling on darwin (cont.)

### LINKED ISSUE #19132: cli: cli crashes when I resize my terminal

knz:
Completes #20148.

Fixes  #19132. (for real)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -486,8 +486,8 @@
 [[projects]]
   name = "github.com/knz/go-libedit"
   packages = [".","common","other","unix","unix/sigtramp"]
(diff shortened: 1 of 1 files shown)


lgo (approved):
Just checked out the branch and it does fix the bug. Thank you @knz! :)

PR #20154, opened 2017-11-20
sql: remove the planner ref from planNodes

knz:
The first commit from #20169.

Visible changes: the `*planner` reference is removed from planNodes.

Release notes: none.

### DIFF
--- a/pkg/sql/group.go
+++ b/pkg/sql/group.go
@@ -27,6 +27,7 @@ import (
 	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
 	"github.com/cockroachdb/cockroach/pkg/util/encoding"
 	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/mon"
 	"github.com/pkg/errors"
 )
 
@@ -162,8 +163,7 @@ func (p *planner) groupBy(
 	}
 
 	group := &groupNode{
-		planner: p,
-		plan:    r,
+		plan: r,
 	}
 
 	// We replace the columns in the underlying renderNode with what the
@@ -299,7 +299,7 @@ func (p *planner) groupBy(
 		log.Infof(ctx, "Group: %s", strings.Join(strs, ", "))
 	}
 
-	group.desiredOrdering = group.desiredAggregateOrdering()
+	group.desiredOrdering = group.desiredAggregateOrdering(&p.evalCtx)
 
 	return plan, group, nil
 }
@@ -307,8 +307,6 @@ func (p *planner) groupBy(
 // A groupNode implements the planNode interface and handles the grouping logic.
 // It "wraps" a planNode which is used to retrieve the ungrouped results.
 type groupNode struct {
-	planner *planner
-
 	// The source node (which returns values that feed into the aggregation).
 	plan planNode
 
@@ -397,7 +395,7 @@ func (n *groupNode) Next(params runParams) (bool, error) {
 				value = values[f.argRenderIdx]
 			}
 
-			if err := f.add(params.ctx, n.planner.session, bucket, value); err != nil {
+			if err := f.add(params.ctx, &params.p.evalCtx, bucket, value); err != nil {
 				return false, err
 			}
 		}
@@ -421,7 +419,7 @@ func (n *groupNode) Next(params runParams) (bool, error) {
 			// No input for this bucket (possible if f has a FILTER).
 			// In most cases the result is NULL but there are exceptions
 			// (like COUNT).
-			aggregateFunc = f.create(&n.planner.evalCtx)
+			aggregateFunc = f.create(&params.p.evalCtx)
 		}
 		var err error
 		n.values[i], err = aggregateFunc.Result()
@@ -444,7 +442,7 @@ func (n *groupNode) setupOutput() {
 func (n *groupNode) Close(ctx context.Context) {
 	n.plan.Close(ctx)
 	for _, f := range n.funcs {
-		f.close(ctx, n.planner.session)
+		f.close(ctx)
 	}
 	n.buckets = nil
 }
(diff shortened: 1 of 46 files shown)


knz:
cc @jordanlewis

knz:
@justinj I was advised by Vivek that you could help @jordanlewis with reviews around this area.

jordanlewis:
Haven't reviewed carefully, but blanket +1 on removing references to `planner` from plan nodes. I suggest limiting the scope of this PR to exactly that if you have not already.

knz:
ok they're all gone.

knz:
Rebased on top of #20169 so no file renames, for your reviewing pleasure.

pkg/sql/group.go
```diff
@@ -484,7 +481,12 @@ func (n *groupNode) addIsNotNullFilter(where *filterNode, render *renderNode) {
 //
 // We only have a desired ordering if we have a single MIN or MAX aggregation
 // with a simple column argument and there is no GROUP BY.
-func (n *groupNode) desiredAggregateOrdering() sqlbase.ColumnOrdering {
+//
+// TODO(knz/radu): it's expensive to instantiate the aggregate
+// function here just o determine whether it's a min or max. We could
```
justinj: `o -> to`

pkg/sql/render.go
```diff
@@ -295,14 +293,14 @@ func (p *planner) SelectClause(
 }
 
 // initFrom initializes the table node, given the parsed select expression
-func (r *renderNode) initFrom(
-	ctx context.Context, parsed *tree.SelectClause, scanVisibility scanVisibility,
+func (p *planner) initFrom(
```
justinj: any particular reason this and other methods moved from the renderNode to the planner?

justinj (approved):
General idea LGTM, I have some other questions that I messaged you about, but definitely in favour of this decoupling!

knz:
There's an error in the last commit. Fixing now.

---

Review status: 0 of 35 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[pkg/sql/group.go, line 486 at r16](https://reviewable.io:443/reviews/cockroachdb/cockroach/20154#-KzOcM0jrwLhpQeRxcPO-r16-486:-KzUD5aqN7FKLwCGKyui:bt2vr1b) ([raw file](https://github.com/cockroachdb/cockroach/blob/41e503b20401e26ae118f64b6f1514c34c1539ba/pkg/sql/group.go#L486)):*

`o -> to`
</blockquote></details>

Fixed.

---

*[pkg/sql/render.go, line 296 at r16](https://reviewable.io:443/reviews/cockroachdb/cockroach/20154#-KzOcM0kxygUzEUfMhdT-r16-296:-KzUD9_wjJtfMoNCJoJM:b-4jwlc) ([raw file](https://github.com/cockroachdb/cockroach/blob/41e503b20401e26ae118f64b6f1514c34c1539ba/pkg/sql/render.go#L296)):*

any particular reason this and other methods moved from the renderNode to the planner?
</blockquote></details>

This function calls `(*planner).getSources()` to construct the logical plan corresponding to the FROM clause. The `planner` reference has to come from somewhere. Si it's either making it a method of `*planner` or passing `p*planner`  as explicit argument. 

Now consider the name "planner". The planner object has too many responsibilities right now, but its name is clear: its purpose, the only purpose it *eventually* will fulfill, is to take an AST and construct a logical plan. So a method that takes a FROM AST and returns a logical plan (or, in this case, annotates it in a pre-initialized logical plan node) is the proper design.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20154#-:-KzUDnpRCsYnVUPgLGmq:b-1j6x2s)*

knz:
I had forgotten a step because of some oddity in the distsql physical planner. Added an intermediate commit that addresses the problem. The PR should be ready now. PTAL. 

---

Review status: 0 of 45 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20154#-:-KzWExBV8K2hvAxtKb3p:bca5bfa)*

pkg/sql/trace.go
```diff
@@ -43,9 +42,9 @@ type traceNode struct {
 	// around the interaction of SQL with KV. Some of the messages are per-row.
 	kvTracingEnabled bool
 
-	// recording is set if this node started tracing on the session. If it did,
-	// then Close() needs to stop the recording.
-	recording bool
+	// stopTracing is set is set if this node started tracing on the
```
RaduBerinde: is set is set. Also "Close() must close it" is confusing

Why is this a function? It seems that the only purpose is so we can implicitly store `params.p.session` in it (so we're still storing runtime state, we just "hide" it a bit).

RaduBerinde:
LGTM

knz:
Review status: 0 of 46 files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/sql/trace.go, line 45 at r23](https://reviewable.io:443/reviews/cockroachdb/cockroach/20154#-KzOcM0mxRZAT9iAmvW7-r23-45:-KztgonGcXLv3MWgVzCS:b-aig6ka) ([raw file](https://github.com/cockroachdb/cockroach/blob/2ab209691caa92dd2dc4fdbdff6c6c5552a26fc5/pkg/sql/trace.go#L45)):*

is set is set. Also "Close() must close it" is confusing

Why is this a function? It seems that the only purpose is so we can implicitly store `params.p.session` in it (so we're still storing runtime state, we just "hide" it a bit).
</blockquote></details>

It reduces the range of operations that can be performed on the stored session. This way, the only thing the planNode can do is stopping the tracing. Keeping the planner/session around would leave the door open to arbitrary other things.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20154)*

knz:
TFYRs

PR #20158, opened 2017-11-20
rfcs: Rename scrub index implementation RFC to add merge-date prefix

lgo:
Unfortunately, when merging the RFC, I ran into rebase problems (welp,
rebased a month in there...). Those were un-done, only to find out that the
change to finalize the RFC before merging were also removed.

cc: @richardwu. Thanks for catching this!


richardwu (approved):
💯 👍 :shipit: 🐶 🍍

benesch:
Whoops, looks like this got renamed into the repository root!

PR #20161, opened 2017-11-20
rfcs: Fix the accidental misplacement of an RFC

lgo:
Welp, take 2. The RFC was accidentally moved into the root when renaming it :(

cc: @benesch


knz (approved):
thanks!

benesch (approved):
Thanks for the quick turnaround!

Issue #20157, opened 2017-11-20
tracing: some query errors are returned as a result of SHOW TRACE FOR <stmt>


andreimatei:
Query errors are not supposed to cause `SHOW TRACE FOR <stmt>` to fail. However, it would appear that planning errors do:

```
root@:26257/t> show trace for select * from ttt;
pq: relation "ttt" does not exist
```

jordanlewis:
This error is from planning. Do we expect `SHOW TRACE` to include results from planning?

andreimatei:
Yes, `SHOW TRACE` traces the planning too. I think it's useful for `SHOW TRACE` to never error.

jordanlewis:
SHOW TRACE FOR <stmt> is dead, having been replaced by some CLI magic that accomplishes the same thing. Closing unless you think this is reproduceible with normal SHOW TRACE FOR SESSION.

PR #20162, opened 2017-11-20
sql: support SHOW TRACE FOR SELECT ... AS OF SYSTEM TIME 

andreimatei:
Release note: SHOW TRACE FOR SELECT ... AS OF SYSTEM TIME is now
supported.

Before this patch, the Executor wasn't recognizing the AS OF SYSTEM TIME
and so the SELECT was failing with a planning error.

I'm going to cherry-pick the first commit into 1.1 because a client wants it.

### DIFF
--- a/pkg/sql/as_of_test.go
+++ b/pkg/sql/as_of_test.go
@@ -106,13 +106,25 @@ func TestAsOfTime(t *testing.T) {
 	} else if i != val2 {
 		t.Fatalf("expected %v, got %v", val2, i)
 	}
-	if err := db.QueryRow(fmt.Sprintf("SELECT a, c FROM d.t, d.j AS OF SYSTEM TIME %s", tsVal1)).Scan(&i, &j); err != nil {
-		t.Fatal(err)
-	} else if i != val1 {
-		t.Fatalf("expected %v, got %v", val1, i)
-	} else if j != val2 {
-		t.Fatalf("expected %v, got %v", val2, j)
-	}
+
+	// Test a simple query, and do it with and without wrapping parens to check
+	// that parens don't matter.
+	testutils.RunTrueAndFalse(t, "parens", func(t *testing.T, useParens bool) {
+		openParens := ""
+		closeParens := ""
+		if useParens {
+			openParens = "(("
+			closeParens = "))"
+		}
+		query := fmt.Sprintf("%sSELECT a, c FROM d.t, d.j AS OF SYSTEM TIME %s%s", openParens, tsVal1, closeParens)
+		if err := db.QueryRow(query).Scan(&i, &j); err != nil {
+			t.Fatal(err)
+		} else if i != val1 {
+			t.Fatalf("expected %v, got %v", val1, i)
+		} else if j != val2 {
+			t.Fatalf("expected %v, got %v", val2, j)
+		}
+	})
 
 	// Future queries shouldn't work.
 	if err := db.QueryRow("SELECT a FROM d.t AS OF SYSTEM TIME '2200-01-01'").Scan(&i); !testutils.IsError(err, "pq: AS OF SYSTEM TIME: cannot specify timestamp in the future") {
--- a/pkg/sql/as_of_test.go
+++ b/pkg/sql/as_of_test.go
@@ -22,6 +22,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/cockroachdb/apd"
+	"github.com/cockroachdb/cockroach/pkg/base"
 	"github.com/cockroachdb/cockroach/pkg/internal/client"
 	"github.com/cockroachdb/cockroach/pkg/roachpb"
 	"github.com/cockroachdb/cockroach/pkg/sql"
@@ -168,8 +169,10 @@ func TestAsOfTime(t *testing.T) {
 	}
 
 	// Subqueries shouldn't work.
-	if _, err := db.Query(fmt.Sprintf("SELECT (SELECT a FROM d.t AS OF SYSTEM TIME %s)", tsVal1)); !testutils.IsError(err, "pq: unexpected AS OF SYSTEM TIME") {
-		t.Fatal(err)
+	_, err := db.Query(
+		fmt.Sprintf("SELECT (SELECT a FROM d.t AS OF SYSTEM TIME %s)", tsVal1))
+	if !testutils.IsError(err, "pq: AS OF SYSTEM TIME not supported in this context") {
+		t.Fatalf("expected not supported, got: %v", err)
 	}
 
 	// Verify that we can read columns in the past that are dropped in the future.
@@ -183,8 +186,10 @@ func TestAsOfTime(t *testing.T) {
 	}
 
 	// Can't use in a transaction.
-	if _, err := db.Query(fmt.Sprintf("BEGIN; SELECT a FROM d.t AS OF SYSTEM TIME %s; COMMIT;", tsVal1)); !testutils.IsError(err, "pq: unexpected AS OF SYSTEM TIME") {
-		t.Fatal(err)
+	_, err = db.Query(
+		fmt.Sprintf("BEGIN; SELECT a FROM d.t AS OF SYSTEM TIME %s; COMMIT;", tsVal1))
+	if !testutils.IsError(err, "pq: AS OF SYSTEM TIME not supported in this context") {
+		t.Fatalf("expected not supported, got: %v", err)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


knz:
Ugh. Why do it this way? Does this not work:
`SELECT * FROM [SHOW TRACE FOR SELECT ....] AS OF SYSTEM TIME...`?

would not need extra code.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-:-KzPlHeA1KMWwwGrLk5s:bg48xfw)*

andreimatei:
`SELECT * FROM [SHOW TRACE FOR SELECT ....] AS OF SYSTEM TIME...` "works", but that's not a query I would write (and it's also not the query that @tschottdorf and friends wrote).

andreimatei:
Like, `SHOW TRACE FOR` is special; you should be able to put it in front of anything; I don't think of what comes after it as a subquery (for which we indeed don't support AOST).

knz:
But that's the problem, right, I can't write `SELECT message from [SHOW TRACE FOR SELECT ... AS OF SYSTEM TIME ...` with your change, it wouldn't work.

So if your PR really wants to address the use case "stick SHOW TRACE FOR in front of something and expect it to work", there is more work to do (make the isAsOf function recurse, or perhaps use `walkPlan()`)

mjibson:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/show_trace, line 722 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-KzPsxnNJYMfJ4dGDCnc:-KzPsGLAEpCuZGDZpuMY:bcp6m7r) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7013cab0f05286952df29f4e87dcbaf5b522a29/pkg/sql/logictest/testdata/logic_test/show_trace#L722)):*
> ```
> # not return the error, see #20157).
> statement error pq: database "t" does not exist
> SHOW TRACE FOR SELECT * FROM t.abc AS OF SYSTEM TIME '2017-01-01'
> ```

Can there be a test where it doesn't return an error, to guarantee it works?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162)*

andreimatei:
Indeed; I think ideally `AS OF SYSTEM TIME` should be allowed in any select, including the ones that our code considers subqueries. But doing that is more involved; we need to make some of, but not all of, the reads in a query non-transactional. Just exploring the tree and fishing out AS OF SYSTEM TIME would be, in my opinion, the wrong way to go; for example if you have a union of two things or a join and one side has it and the other doesn't, it'd be incorrect to just fish it out and set a txn timestamp.
That's why I think what I have here is simple and worth doing; it'll be the last time we hear of this problem because, let's face it, nobody knows the `FROM [...]` syntax.

knz:
Be sure to also test explicit PREPARE/EXECUTE statements here.

knz:
@jordanlewis I'd like you to have a second look as your insight wrt plan caching may influence the evaluation of this approach.

andreimatei:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/show_trace, line 722 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-KzPsxnNJYMfJ4dGDCnc:-KzPt6cSge_e-Vi1PXei:bmlqv6q) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7013cab0f05286952df29f4e87dcbaf5b522a29/pkg/sql/logictest/testdata/logic_test/show_trace#L722)):*

Can there be a test where it doesn't return an error, to guarantee it works?
</blockquote></details>

well you tell me. As far as I can see, I can't use any expression with AS OF SYSTEM TIME, so I can't run a query at a present time. And so without that I need to write a program to get the current timestamp so I can't do it in a logic test. I think that'd be overkill for testing this.
Also, this test does show that something works - we get to planning the statement and we pretty convincingly read descriptors at an old timestamp.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162)*

knz:
Regarding the design I still believe it's unfortunate to create a situation we have statements that are valid in one syntactic context (e.g. at the top level) but not at another (e.g. in a subquery). 
The problem here is one of assymmetry -- we're trying very hard in other places to ensure that SQL is algebraic (the same expression has the same meaning in every situation) and here you're just casually breaking the rule simple because you don't want to call `walkPlan` with a suitable plan visitor and instead bake a half-functional predicate that's just a glorified `_, ok := v.(cast); return ok`.

mjibson:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pkg/sql/logictest/testdata/logic_test/show_trace, line 722 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-KzPsxnNJYMfJ4dGDCnc:-KzPyMBp5w9cPPCClJtY:b-t7b03v) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7013cab0f05286952df29f4e87dcbaf5b522a29/pkg/sql/logictest/testdata/logic_test/show_trace#L722)):*

well you tell me. As far as I can see, I can't use any expression with AS OF SYSTEM TIME, so I can't run a query at a present time. And so without that I need to write a program to get the current timestamp so I can't do it in a logic test. I think that'd be overkill for testing this.
Also, this test does show that something works - we get to planning the statement and we pretty convincingly read descriptors at an old timestamp.
</blockquote></details>

How about adding something in as_of_test.go? It already does all of the work of getting a timestamp at which something is present.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162)*

andreimatei:
> and here you're just casually breaking the rule simple because you don't want to call walkPlan with a suitable plan visitor 

Please be more generous with what I want and what I don't want; I want everything to be clean and great.
I didn't implement AS OF SYSTEM TIME; I see myself as an innocent bystander. This change makes AOST work in more contexts, not fewer. It's making our language more "symmetric".
I'm trying to address an issue with `SHOW TRACE FOR` that someone is complaining about. You're complaining about issues with AOST; `select * from (SELECT * from t.t AS OF SYSTEM TIME '2017-11-20')` does not work.

Now, what is your suggestion exactly? We write a visitor that recognizes `AOST` - cool; but it needs to reject the general query in which some sources have AOST and others don't. How would we do that exactly?

jordanlewis:
This approach seems fine to me. Agree with @andreimatei that it is an incremental improvement on AOST. A more complete restructuring seems out of scope for this change.

---

Reviewed 7 of 7 files at r1.
Review status: 5 of 7 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/render.go, line 304 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-KzPzMbByHwvwcvqPjxG:-KzPzMbByHwvwcvqPjxH:byp3qaf) ([raw file](https://github.com/cockroachdb/cockroach/blob/00e508fb381f89639da7285e22e4aa85673f4690/pkg/sql/render.go#L304)):*
> ```Go
> 	// definition), or they should have been handled by the executor.
> 	if parsed.From.AsOf.Expr != nil && !r.planner.avoidCachedDescriptors {
> 		return fmt.Errorf("AS OF SYSTEM TIME not supported")
> ```

Since I was pinged on this review - Do we know the statement context here? I would prefer this message be extended with specifics if possible - for example, something like "AS OF SYSTEM TIME not supported in CREATE TABLE AS". If we don't know the context, or it's hard to find out, I'd prefer it if you appended "in this context" to the error message. Otherwise I believe it will be misleading to users.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-:-KzQ1Jhc4-yN7qDzDUjr:bbw7rso)*

knz:
All right, I'm all for small step improvements. Carry on.

LGTM

andreimatei:
@knz, if you do have an idea about how that visitor should look like, I would like to hear about it because, independently of this change, the way in which the executor has special code for recognizing AOST does bother me. If possible, I think it'd be much nicer to push checking of valid uses to planning and actual handling of to the `ScanNode` or such. That's because the current code doesn't go well with a world in which session handling code is supposed to be different from query execution code. And also because the current implementation is a bit broken: see #20182

---

Review status: 1 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/render.go, line 304 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-KzPzMbByHwvwcvqPjxG:-KzQ9s6afSyKIT4fZdDT:b34pypl) ([raw file](https://github.com/cockroachdb/cockroach/blob/00e508fb381f89639da7285e22e4aa85673f4690/pkg/sql/render.go#L304)):*

Since I was pinged on this review - Do we know the statement context here? I would prefer this message be extended with specifics if possible - for example, something like "AS OF SYSTEM TIME not supported in CREATE TABLE AS". If we don't know the context, or it's hard to find out, I'd prefer it if you appended "in this context" to the error message. Otherwise I believe it will be misleading to users.
</blockquote></details>

I've added "not supported in this context", good suggestion. The reason why I changed this error in the first place was that before it looked too much like a parse error.
I don't know what kind of "context" is or could be available - like, are you thinking of a part of the statement, or what exactly? I don't quite see what we could add here other than perhaps the whole statement (which would be useful), but if you do, I'm listening.

---

*[pkg/sql/logictest/testdata/logic_test/show_trace, line 722 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-KzPsxnNJYMfJ4dGDCnc:-KzQASf_GnNbjDSh1jjw:bgghkmu) ([raw file](https://github.com/cockroachdb/cockroach/blob/c7013cab0f05286952df29f4e87dcbaf5b522a29/pkg/sql/logictest/testdata/logic_test/show_trace#L722)):*

How about adding something in as_of_test.go? It already does all of the work of getting a timestamp at which something is present.
</blockquote></details>

I didn't do that originally because deciding on what to test here exactly is not easy. But I added something now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20162#-:-KzQQSq5TR83AczW3Thb:b-xyfhjc)*

madelynnblue (approved):

PR #20167, opened 2017-11-20
release-1.1: distsqlrun: Fix MergeJoiner correctness issue with NULL values

lgo:
This cherrypicks the fix from #20090.

---

Previously, when a distSQL merge join was being executed the comparison
check was using Datum.Compare while processing the incoming streams we
are joining on.

This is problematic when both sides of the stream have a NULL value,
because this comparison is considered equal which is a violation of the
semantics of equals (=) with SQL NULLs.

Unfortunately, this bug was present before distSQL merge joins were
being planned (#17214) and so it exists in 1.1.0.

### DIFF
--- a/pkg/sql/distsqlrun/mergejoiner_test.go
+++ b/pkg/sql/distsqlrun/mergejoiner_test.go
@@ -345,6 +345,44 @@ func TestMergeJoiner(t *testing.T) {
 				{null, v[5], v[1]},
 			},
 		},
+		{
+			spec: MergeJoinerSpec{
+				LeftOrdering: convertToSpecOrdering(
+					sqlbase.ColumnOrdering{
+						{ColIdx: 0, Direction: encoding.Ascending},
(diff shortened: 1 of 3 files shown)


lego:
There were a few cherrypick conflicts due to the refactor changing `EncDatum` function signatures to need `ColumnType` arguments and the tests, but it's a relatively simple change and conflict.

lego:
Fixed one test failure (the 1.1 code generates different distSQL plans then master -- table readers seem to include an `NULL` in the projection values in 1.1).

There was a second test failure that is quite unexpected though, from `alter_table`  running in distSQL which is a little unexpected. https://teamcity.cockroachdb.com/viewLog.html?buildId=415591&buildTypeId=Cockroach_UnitTests

Going to let the tests run to see what happens.

lego:
Not going to merge this into 1.x after speaking with @vivekmenezes

PR #20170, opened 2017-11-20
build: rename files to not have backslash

justinj:
Apparently these were messing with the build system in some way, so
easier to just rename them.


benesch (approved):
:shipit: thank you!

Issue #20100, opened 2017-11-16
sql/distsql: push down NOT NULL constraints on equality columns during planning


lgo:
We can push down a `NOT NULL` filter depending on the join type. This can reduce the data that is streamed to the join. For example, if we have:

```sql
SELECT * FROM t1 AS l INNER JOIN t1 AS r ON l.id = r.id
```

If either `l.id` or `r.id` are NULL they will never be in the join result due to `NULL = <any>` is false-y. This applies to joins that are composed of only equalities, in particular:
- All inner joins
- When doing a left or right outer join, we can push the constraint down to the other side of the join (e.g. when doing a `LEFT OUTER JOIN`, we can push it down the right side of the join)

Mentioned from https://github.com/cockroachdb/cockroach/pull/20090 and https://github.com/cockroachdb/cockroach/issues/17135#issuecomment-344744926.

cc: @RaduBerinde 

Jira issue: CRDB-5942

sydnever:
Hello, I am a new bee here. I would like to know whether you will code this issue yourself (I noted you fix #20090 ). If you will, I will look for another issue which maybe looks simple; if not, I want to have a try.

knz:
@sydnever hello! Thanks for your interest in this issue. We haven't looked into this yet, so the issue is still "free". Do you know already what the solution looks like?

sydnever:
@knz I have some ideas in my mind and I am trying to find where and how I will code. When I make it clear, I will share my plan here

knz:
Thanks.

sydnever:
@knz Should I make the final plan like this:
```sql
root@localhost:26257/test> EXPLAIN(EXPRS,METADATA) 
select * from 
(select * from t1 where t1.id is not null) as r 
inner join 
(select * from t1 where t1.id is not null) as j 
on r.id = j.id;
+-------+--------+----------+----------------+----------------------------------+-----------------------------------+
| Level |  Type  |  Field   |  Description   |             Columns              |             Ordering              |
+-------+--------+----------+----------------+----------------------------------+-----------------------------------+
|     0 | join   |          |                | (id, num, id, num)               |                                   |
|     0 |        | type     | inner          |                                  |                                   |
|     0 |        | equality | (id) = (id)    |                                  |                                   |
|     1 | render |          |                | (id, num)                        | id!=NULL                          |
|     1 |        | render 0 | id             |                                  |                                   |
|     1 |        | render 1 | num            |                                  |                                   |
[...]

PR #20177, opened 2017-11-20
sql: Improve scrub index check performance

lgo:
This significantly improves the performance of the check after #19889.
After adding the condition to filter false positives, the check was
running an expensive and slow hash join.

After this patch, the scan and check uses a sort and merge join and is
much faster. In order to do this, we needed to leverage outer joins in
the distSQL merge joiner and the capability to have `NULL = NULL` be
truth-y.

Roughly estimating, prior to the patch a check of 5000 rows took >10s.
After the patch this is well under 1s.

Also after changing the scrub index check to use a sorted merge join, it
failed to handle tables with an implicit primary key (i.e. a hidden
column `rowid`).

This fixes the bug and adds a test to ensure scrub checks in general
will not fail when executing on such a table.


lego:
Right now there's a commit in there that now plans sorted merge joins in the distSQL engine for outer joins. I expect this commit to be pulled into another PR and tested more, but was required to make this happen (and is also extremely small).

I'd like to get some thoughts on 1) the "hack" for nullEquality done here and 2) what it will take to get distSQL outer merge joins to production, which this depends on.

lego:
@RaduBerinde PTAL.

This will conflict with the in-progress work for #20100. Once that change lands, the push-down filter will need to be removed (or not added) specifically when running the index check, and that does seem like a bad workaround.

RaduBerinde:
:lgtm:

---

Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[pkg/sql/scrub.go, line 391 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3WqiLDb3OoL9Vj1yK:-L-3WqiLDb3OoL9Vj1yL:b-fk4uco) ([raw file](https://github.com/cockroachdb/cockroach/blob/30249d3a52b74c36d21a7145d5f1f3d9d599863c/pkg/sql/scrub.go#L391)):*
> ```Go
> 
> 	// Set the MergeJoiner NullEquality to true. This changes the
> 	// behaviour of the query's equality semantics in the ON predicate.
> ```

[nit] behavior

---

*[pkg/sql/scrub.go, line 395 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3X8vXyXJv7g6T36O3:-L-3X8vXyXJv7g6T36O4:b-hptbn4) ([raw file](https://github.com/cockroachdb/cockroach/blob/30249d3a52b74c36d21a7145d5f1f3d9d599863c/pkg/sql/scrub.go#L395)):*
> ```Go
> 	// we desire when testing the equivilance of two index entries. We do
> 	// not know which processor is the merge joiner, but we do know that
> 	// there will only be one.
> ```

Why is there always just one? We could have multiple merge-joiners (with hash-routing to them). In any case, we don't need to worry about this here, we could just set the flag in all MergeJoiners we find

By the way, we should have some tests with `EXPLAIN (DISTSQL) EXPERIMENTAL SCRUB ..` to see how the plans look. See e.g. `distsql_agg` for examples.

---

*[pkg/sql/scrub.go, line 629 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3YEZ1VRqp6OLz-SsC:-L-3YEZ1VRqp6OLz-SsD:bci42lm) ([raw file](https://github.com/cockroachdb/cockroach/blob/fea62aa5b65f43a23412d3208060728357249ea4/pkg/sql/scrub.go#L629)):*
> ```Go
> 				SELECT %[1]s, %[2]s
> 				FROM
> 					(SELECT %[9]s FROM %[3]s@{FORCE_INDEX=[1],NO_INDEX_JOIN} ORDER BY %[5]s) AS leftside
> ```

TIL! We should add `// 1`, `// 2` and so on to the lines below so it's easy to see what a `%[x]s` is

---

*[pkg/sql/distsqlrun/stream_merger.go, line 117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3WXCSq4wBJI992kzo:-L-3WXCSq4wBJI992kzp:bxf1gs6) ([raw file](https://github.com/cockroachdb/cockroach/blob/0c11cadfb872b02fc43307468a0878a184c7b9b6/pkg/sql/distsqlrun/stream_merger.go#L117)):*
> ```Go
> 		// they are not equal. This differs from our datum semantics where
> 		// they are equal.
> 		if !nullEquality && lhs[lIdx].IsNull() && rhs[rIdx].IsNull() {
> ```

I think we should `continue;` if they're both Null and nullEquality is set. It probably does the same thing, but it's more clear.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-:-L-3YVAKNHgx3KRqEuEl:bnfp4nl)*

RaduBerinde:
Reviewed 1 of 1 files at r1, 5 of 5 files at r2, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177)*

lego:
I duplicated some logictests in `join` and modified them so that they will also execute with a merge join, in order to add some more testing to the distSQL merge joiner.

---

Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/sql/scrub.go, line 391 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3WqiLDb3OoL9Vj1yK:-L-7N7Z9K9s8M74SLqG_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/30249d3a52b74c36d21a7145d5f1f3d9d599863c/pkg/sql/scrub.go#L391)):*

[nit] behavior
</blockquote></details>

Done.

---

*[pkg/sql/scrub.go, line 395 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3X8vXyXJv7g6T36O3:-L-7NAR9rJe0sUR1VqyN:b-tlg6u0) ([raw file](https://github.com/cockroachdb/cockroach/blob/30249d3a52b74c36d21a7145d5f1f3d9d599863c/pkg/sql/scrub.go#L395)):*

Why is there always just one? We could have multiple merge-joiners (with hash-routing to them). In any case, we don't need to worry about this here, we could just set the flag in all MergeJoiners we find

By the way, we should have some tests with `EXPLAIN (DISTSQL) EXPERIMENTAL SCRUB ..` to see how the plans look. See e.g. `distsql_agg` for examples.
</blockquote></details>

Ah, I didn't know about that. Changed it so it'll modify all of them, and the comment to reflect that.

Re the explain: scrub is run in local execution. Instead, I added a `distsql_scrub` file which tests the plan for the query that is generated and run internally.

---

*[pkg/sql/scrub.go, line 629 at r4](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3YEZ1VRqp6OLz-SsC:-L-7WTdNKxWOeUKTF44t:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/fea62aa5b65f43a23412d3208060728357249ea4/pkg/sql/scrub.go#L629)):*

TIL! We should add `// 1`, `// 2` and so on to the lines below so it's easy to see what a `%[x]s` is
</blockquote></details>

Done.

---

*[pkg/sql/distsqlrun/stream_merger.go, line 117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3WXCSq4wBJI992kzo:-L-7XVuXykrQfbSYqVK5:b-9f087b) ([raw file](https://github.com/cockroachdb/cockroach/blob/0c11cadfb872b02fc43307468a0878a184c7b9b6/pkg/sql/distsqlrun/stream_merger.go#L117)):*

I think we should `continue;` if they're both Null and nullEquality is set. It probably does the same thing, but it's more clear.
</blockquote></details>

Yep, done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-:-L-7hn48YTMsQdZd3raE:b-tajzo0)*

RaduBerinde:
Review status: 0 of 11 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/sql/scrub.go, line 395 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3X8vXyXJv7g6T36O3:-L-7zBj_fz4uTLynNQzb:b-5k7037) ([raw file](https://github.com/cockroachdb/cockroach/blob/30249d3a52b74c36d21a7145d5f1f3d9d599863c/pkg/sql/scrub.go#L395)):*

Ah, I didn't know about that. Changed it so it'll modify all of them, and the comment to reflect that.

Re the explain: scrub is run in local execution. Instead, I added a `distsql_scrub` file which tests the plan for the query that is generated and run internally.
</blockquote></details>

Ah, I see, scrub runs locally and then generates a distsql plan so we can't get that internal plan through EXPLAIN

---

*[pkg/sql/distsqlrun/stream_merger.go, line 117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3WXCSq4wBJI992kzo:-L-7yOhBoj95NWs1SBWh:bmt1dfu) ([raw file](https://github.com/cockroachdb/cockroach/blob/0c11cadfb872b02fc43307468a0878a184c7b9b6/pkg/sql/distsqlrun/stream_merger.go#L117)):*

Yep, done.
</blockquote></details>

We should avoid running `IsNull` multiple times unnecessarily, this could be:
```go
if lhs[lIdx].IsNull() && rhs[rIdx].IsNull() {
    if !nullEquality {
      return -1, nil
    }
    continue
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177)*

lego:
Review status: 0 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/sql/distsqlrun/stream_merger.go, line 117 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-L-3WXCSq4wBJI992kzo:-L-8Gj6hW0016CVE5OOi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0c11cadfb872b02fc43307468a0878a184c7b9b6/pkg/sql/distsqlrun/stream_merger.go#L117)):*

We should avoid running `IsNull` multiple times unnecessarily, this could be:
```go
if lhs[lIdx].IsNull() && rhs[rIdx].IsNull() {
    if !nullEquality {
      return -1, nil
    }
    continue
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177)*

RaduBerinde:
:lgtm:

---

Review status: 0 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20177#-:-L-8TCLDFJ4TuuM4dq2c:bnfp4nl)*

Issue #20180, opened 2017-11-20
storage: SysBytes incorrect


tschottdorf:
Found investigating https://github.com/cockroachdb/cockroach/issues/20159.

It looks that our SysBytes are usually incorrect on each and every range I've seen (and I've pointed this at a few data directories from various clusters):

```
$ ./cockroach start --insecure --store freshdir
^CTRL+C
$ ./cockroach debug check-store freshdir

checking MVCC stats

r1:/{Min-System/} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 175238 != 512
SysCount: 403 != 9

r2:/System/{-NodeLiveness} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 459
SysCount: 0 != 8

r3:/System/NodeLiveness{-Max} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 490
SysCount: 0 != 8

r4:/System/{NodeLivenessMax-tsd} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 483
SysCount: 0 != 8

r5:/System/ts{d-e} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 461
SysCount: 0 != 8

r6:/{System/tse-Table/SystemConfigSpan/Start} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 658
SysCount: 0 != 11

r7:/Table/{SystemConfigSpan/Start-11} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 622
SysCount: 0 != 10

r8:/Table/1{1-2} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 515
SysCount: 0 != 9

r9:/Table/1{2-3} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 515
SysCount: 0 != 9

r10:/Table/1{3-4} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 515
SysCount: 0 != 9

r11:/Table/1{4-5} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 515
SysCount: 0 != 9

r12:/Table/1{5-6} [(n1,s1):1, next=2]: diff(actual, claimed): SysBytes: 0 != 582
SysCount: 0 != 10

[...]

PR #20181, opened 2017-11-20
cli: check MVCCStats in `debug check-store`

tbg:
Unfortunately, this discovered #20180: The stats essentially never match (or this PR has a bug).


petermattis:
:lgtm: 

I don't see anything obviously wrong with the code, but I'm very suspicious there is something. Don't we check on disk stats vs computed stats elsewhere (e.g. during splits)?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20181#-:-KzU3hEyaOVd6QY4WFbF:b39ynjl)*

tschottdorf:
@petermattis no. I hoped that this would be fixed by #21070 but it's still happening. I'm working on making the replica GC queue check this (which I need anyway to trigger repair now that the stats are at least somewhat fixed, and more fixes are planned).

PR #20184, opened 2017-11-20
cli: pretty-print intents in `debug keys`

tbg:
Can improve this by pretty-printing the inlined value, but may not be
necessary for a while.

### DIFF
--- a/pkg/cli/debug.go
+++ b/pkg/cli/debug.go
@@ -92,7 +92,7 @@ func openStore(cmd *cobra.Command, dir string, stopper *stop.Stopper) (*engine.R
 }
 
 func printKey(kv engine.MVCCKeyValue) (bool, error) {
-	fmt.Printf("%s", kv.Key)
+	fmt.Printf("%s %s: ", kv.Key.Timestamp, kv.Key.Key)
 	if debugCtx.sizes {
 		fmt.Printf(" %d %d", len(kv.Key.Key), len(kv.Value))
 	}
@@ -101,11 +101,8 @@ func printKey(kv engine.MVCCKeyValue) (bool, error) {
 }
 
 func printKeyValue(kv engine.MVCCKeyValue) (bool, error) {
-	if kv.Key.Timestamp != (hlc.Timestamp{}) {
-		fmt.Printf("%s %s: ", kv.Key.Timestamp, kv.Key.Key)
-	} else {
-		fmt.Printf("%s: ", kv.Key.Key)
-	}
+	fmt.Printf("%s %s: ", kv.Key.Timestamp, kv.Key.Key)
+
 	if debugCtx.sizes {
 		fmt.Printf("%d %d: ", len(kv.Key.Key), len(kv.Value))
 	}
--- a/pkg/cli/debug.go
+++ b/pkg/cli/debug.go
@@ -112,6 +112,7 @@ func printKeyValue(kv engine.MVCCKeyValue) (bool, error) {
 		tryMeta,
 		tryTxn,
 		tryRangeIDKey,
+		tryIntent,
 	}
 	for _, decoder := range decoders {
 		out, err := decoder(kv)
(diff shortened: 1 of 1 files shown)


pkg/cli/debug.go
```diff
@@ -358,6 +356,21 @@ func printRangeDescriptor(kv engine.MVCCKeyValue) (bool, error) {
...
+	}
+	var meta enginepb.MVCCMetadata
+	if err := protoutil.Unmarshal(kv.Value, &meta); err != nil {
+		return "", err
+	}
+	s := fmt.Sprintf("%+v", meta)
+	if meta.Txn != nil {
+		s = meta.Txn.Timestamp.String() + " " + s
+	}
+	return s, nil
```
knz:
```go
var buf bytes.Buffer
if meta.Txn != nil {
   proto.CompactText(&buf, meta.Txn.Timestamp)
}
proto.MarshalText(&buf, meta)
return buf.String()
```

petermattis:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/cli/debug.go, line 104 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184#-KzU0fSjeUET4l0zkhZb:-KzU0fSjeUET4l0zkhZc:bl1htl2) ([raw file](https://github.com/cockroachdb/cockroach/blob/50992d73f4f1acad7ca6871d810ba215c30a9df4/pkg/cli/debug.go#L104)):*
> ```Go
> 
> func printKeyValue(kv engine.MVCCKeyValue) (bool, error) {
> 	if kv.Key.Timestamp != (hlc.Timestamp{}) {
> ```

Curious why you've removed this. Printing the zero timestamp isn't terribly useful.

---

*[pkg/cli/debug.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184#-KzQWB0cCIn4LCBe4T2m-r2-371:-KzU0xa_pcmb7-PfEZRL:b-nqn1tg) ([raw file](https://github.com/cockroachdb/cockroach/blob/aa9833fe8a820c8d3d5772c8c9150ad182c0d8f4/pkg/cli/debug.go#L371)):*

```go
var buf bytes.Buffer
if meta.Txn != nil {
   proto.CompactText(&buf, meta.Txn.Timestamp)
}
proto.MarshalText(&buf, meta)
return buf.String()
```
</blockquote></details>

The format of `proto.CompactText` is different from `Timestamp.String()`. I think it is desirable to use the `Timestamp.String()` format here for easier grep-ability of the output.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184#-:-KzU1L09A8c56NK4VHxl:bnfp4nl)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[pkg/cli/debug.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184#-KzQWB0cCIn4LCBe4T2m-r2-371:-KzU6xn1XwOkLjpDEYsH:bcuekk) ([raw file](https://github.com/cockroachdb/cockroach/blob/aa9833fe8a820c8d3d5772c8c9150ad182c0d8f4/pkg/cli/debug.go#L371)):*

The format of `proto.CompactText` is different from `Timestamp.String()`. I think it is desirable to use the `Timestamp.String()` format here for easier grep-ability of the output.
</blockquote></details>

I made the mistake to look at a different protobuf auto-generated source and saw this:
```go
func (m *TxnMeta) String() string            { return proto.CompactTextString(m) }
```

I didn't see we were overloading the String method explicitly.

Nevertheless, I would still 1) add a Format() method to `hlc.Timestamp` that takes a bytes.Buffer or io.Writer argument 2) use that here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184)*

tschottdorf:
Reviewed 1 of 1 files at r2.

---

*[pkg/cli/debug.go, line 104 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184#-KzU0fSjeUET4l0zkhZb:-L-gJVzgLrG3LtFB0gpv:bycuyor) ([raw file](https://github.com/cockroachdb/cockroach/blob/50992d73f4f1acad7ca6871d810ba215c30a9df4/pkg/cli/debug.go#L104)):*

Curious why you've removed this. Printing the zero timestamp isn't terribly useful.
</blockquote></details>

I've found the consistent output format useful. Also, as less experienced folks are poking around in these commands, it's useful for them too to realize that "everything has a timestamp".

---

*[pkg/cli/debug.go, line 371 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184#-KzQWB0cCIn4LCBe4T2m-r2-371:-L-gJoCJbIwh0l7wKc_3:b-bcluq8) ([raw file](https://github.com/cockroachdb/cockroach/blob/aa9833fe8a820c8d3d5772c8c9150ad182c0d8f4/pkg/cli/debug.go#L371)):*

I made the mistake to look at a different protobuf auto-generated source and saw this:
```go
func (m *TxnMeta) String() string            { return proto.CompactTextString(m) }
```

I didn't see we were overloading the String method explicitly.

Nevertheless, I would still 1) add a Format() method to `hlc.Timestamp` that takes a bytes.Buffer or io.Writer argument 2) use that here.
</blockquote></details>

I'll leave as-is for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20184)*

PR #20185, opened 2017-11-20
sql/jobs: refactor job registry to control resumable jobs

madelynnblue:
Reduce the jobs API by moving some functions (pause, resume, cancel)
to the registry. This prevents jobs that do not use the registry from
pausing or canceling jobs.

Refactor the registry to do the actual starting of jobs, instead
of only resuming them when a node outage occurs. This change means
that a function calling into the registry must not both register
a job and begin executing it (as backup and restore did). Instead,
all execution now flows through the resume path. This change had a
side effect of fixing some race condition bugs in the registry.

Since txns are now better handled in the registry, hook up
PAUSE/RESUME/CANCEL to session transactions, which allows them to be
used in COMMIT/ROLLBACK.

This fleshed out two bugs in backup and restore where the initial
path was different from the resume path. The restore bug has already
been fixed (#20092) because it needed a backport to 1.1.

The backup bug only applied to 1.2, and caused backup resume with a
bad backup descriptor, forgetting things like MVCC All and complete
DBs. Add a field to the backup job that stores the backup descriptor
so it can be retained. Refuse to work of the field is not set (which
prevents a 1.2 node from resuming a 1.1 backup).

Release note: Fix BACKUPs to correctly resume in all conditions.

Release note: Fix various race conditions with jobs.

Release note: PAUSE/RESUME/CANCEL JOB statements are now usable within
SQL transactions.


lego:
> This change means
that a function calling into the registry must not both register
a job and begin executing it

So, does this mean that the backup and restore commands immediately start a job which in turn begins execution? How does the caller run backup/restore in a synchronous manner -- assuming if you call it it's still run in the foreground in a SQL CLI session.

(I haven't taken a look at the changes, if there's a particular file feel free to point me to it, that or an approach is great)

mjibson:
registry.StartJob takes a results chan that is hooked up to the console. It also returns an error chan that receives a value once the current execution is done (either because of a pause/cancel or it's done).

benesch:
Still need to review the test changes you made, but wow, not much to say! Looks so much better. :tada: 

---

Reviewed 14 of 15 files at r1.
Review status: 14 of 15 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/csv.go, line 1076 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjK_kTugGsYgY1mIxb:-KzjK_kTugGsYgY1mIxc:bn5elun) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/csv.go#L1076)):*
> ```Go
> 			if err := job.Succeeded(ctx, nil); err != nil {
> 				return err
> 			}
> ```

These inline `nil`s could use inline comments. (See similar comment in schema_changer.go.)

---

*[pkg/ccl/sqlccl/restore.go, line 716 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjUs436S3ALJsSdA3b:-KzjUs436S3ALJsSdA3c:bldtx7r) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/restore.go#L716)):*
> ```Go
> 	tables []*sqlbase.TableDescriptor,
> 	user string,
> ) error {
> ```

This function is only called from the restore success handler. I'd either inline it there or at least drop the `func() error` closure. Between the closure and this function's call site, we're double-`errors.Wrap`ing all errors, which seems unnecessary.

---

*[pkg/ccl/sqlccl/restore.go, line 1185 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzkFTt_dAzg9QAM2UZb:-KzkFTt_dAzg9QAM2UZc:bs28pzh) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/restore.go#L1185)):*
> ```Go
> 			tables = append(tables, tableDesc)
> 		}
> 	}
> ```

Seems like we should make `r.tables` valid during the `OnFailOrCancel` callback too, so that we don't have to load the descriptors from the export store twice. Actually, we probably shouldn't be relying on the export store at all when cleaning up a failed restore. Ugh.

Don't worry about it in this PR, but seems worth filing as followup work.

---

*[pkg/sql/schema_changer.go, line 468 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjJsaHkd4TA2p3R-Zs:-KzjJsaHkd4TA2p3R-Zt:b-rpdgz0) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/schema_changer.go#L468)):*
> ```Go
> ) error {
> 	log.Warningf(ctx, "reversing schema change %d due to irrecoverable error: %s", *sc.job.ID(), err)
> 	if err := sc.job.Failed(ctx, err, nil); err != nil {
> ```

Per Andrei's style guide additions, can you add an inline comment `/* onFailOrCancel */` next to the `nil` argument?

---

*[pkg/sql/schema_changer.go, line 615 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjK3QzMET-izsR19sR:-KzjK3QzMET-izsR19sS:b-7esevx) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/schema_changer.go#L615)):*
> ```Go
> 		return nil
> 	}, func(txn *client.Txn) error {
> 		if err := sc.job.WithTxn(txn).Succeeded(ctx, nil); err != nil {
> ```

Ditto here.

---

*[pkg/sql/jobs/registry.go, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjBEPumHfXrhDaYFRt:-KzjBEPumHfXrhDaYFRu:b7tpbj6) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L68)):*
> ```Go
> 		// propogated to jobs via the .Progressed call. This function should not be
> 		// used to cancel a job in that way.
> 		jobs map[int64]func()
> ```

You could make this more self-documenting by using `context.CancelFunc` instead of `func()` directly.

---

*[pkg/sql/jobs/registry.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjCn4EaiUNYs8DoqP6:-KzjCn4EaiUNYs8DoqP7:b-tmtpoo) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L116)):*
> ```Go
> ) (*Job, <-chan error, error) {
> 	j := r.NewJob(record)
> 	id := r.makeJobID()
> ```

Wouldn't hurt to add a comment to the effect of "We create the job with a known ID, rather than relying on the column's `DEFAULT unique_rowid()`, to avoid a race condition where the job exists in the jobs table but is not yet present in our registry."

---

*[pkg/sql/jobs/registry.go, line 127 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjEBPBdkY7WFeIabl0:-KzjEBPBdkY7WFeIabl1:b-6r65hk) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L127)):*
> ```Go
> 		return nil, nil, err
> 	}
> 	errCh, err := r.resume(resumeCtx, resultsCh, j)
> ```

Seems to me like `resume` should take a `resumer`. Then it won't ever fail and won't need to return an `error`. Means calling `resume` will be a bit more annoying, as you'll have to call `getResumeHook` first, but I think that's actually desirable here. If you call `getResumeHook` before you register or create the job, then you can return early and avoid this complicated cleanup. You'll need to add a `Type() Type` method to `Details`, but that seems worthwhile.

---

*[pkg/sql/jobs/registry.go, line 209 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjFbXWBXAg-RNm41GK:-KzjFbXWBXAg-RNm41GL:b-2qfrqq) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L209)):*
> ```Go
> 			case <-time.After(adoptInterval):
> 				if err := r.maybeAdoptJob(ctx, nl); err != nil {
> 					log.Errorf(ctx, "%d: error while adopting jobs: %+v", r.nodeID.Get(), err)
> ```

Intentional? The log context tags should take care of this, no?

---

*[pkg/sql/jobs/registry.go, line 348 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzkH74lAXiP3KWq6TMJ:-KzkH74lAXiP3KWq6TMK:bo7q57v) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L348)):*
> ```Go
> 		return nil, err
> 	}
> 	errorCh := make(chan error, 1)
> ```

nit: you call this an `errCh` everywhere else

---

*[pkg/sql/jobs/registry.go, line 547 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjApRdfQ2R8Xp7Bg8e:-KzjApRdfQ2R8Xp7Bg8f:b-uld242) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L547)):*
> ```Go
> 	r.mu.Lock()
> 	cancel, ok := r.mu.jobs[jobID]
> 	// It is possible for a job to be double unregistired. unregister is always
> ```

"unregistired"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-:-KzkHRweBA3Zm206cMdw:b-mmdxvk)*

benesch:
Reviewed 1 of 15 files at r1.
Review status: all files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---

*[pkg/sql/jobs/jobs_test.go, line 181 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzo085Zzg962P2jJsX3:-Kzo085Zzg962P2jJsX4:biv0yyp) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L181)):*
> ```Go
> }
> 
> var _ jobs.Resumer = dummyResumer{}
> ```

If you aren't opposed, I'd stick `dummyResumer` in helpers_test.go and rename it to `FakeResumer` for consistency with the other stub objects.

---

*[pkg/sql/jobs/jobs_test.go, line 379 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzoiv-DnuIpElWVu0xc:-Kzoiv-DnuIpElWVu0xd:b-j3o3st) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L379)):*
> ```Go
> 			err := job.Progressed(ctx, 0.42, jobs.Noop)
> 			if err := job.FinishedWith(ctx, err); !testutils.IsError(err, "job canceled") {
> 				t.Fatalf("expected 'job paused' error, but got %v", err)
> ```

Would be ideal if the relevant registry lifecycle tests would verify that `job paused` and `job canceled` were indeed the error messages returned when a job is paused and resumed, respectively.

---

*[pkg/sql/jobs/jobs_test.go, line 382 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzocov71zy1zuJjBJk_:-Kzocov71zy1zuJjBJka:bt60e1o) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L382)):*
> ```Go
> 				t.Fatal(err)
> 			}
> 			progressCh <- struct{}{}
> ```

Sending over the `progressCh` races with `check`, right? What if you sent on it twice to verify that you stay in the loop at least once?

---

*[pkg/sql/jobs/jobs_test.go, line 394 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzod5tgRl6U3qeqiVkV:-KzobcCa8ISs3qgg-pGB:b-m8m3fa) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L394)):*
> ```Go
> 				t.Fatal(err)
> 			}
> 			// Not committed yet, so state shouldn't have changed.
> ```

Ditto here. Don't you want to send on the `progressCh` to actually trigger a call to `Progressed`?

---

*[pkg/sql/jobs/jobs_test.go, line 402 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzodgVUmC_wuWrCP2fY:-KzodgVVvrPJ-F6iNCWT:bao3w5j) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L402)):*
> ```Go
> 		// There's a race condition where progressCh isn't used anymore, so just
> 		// attempt to send on it.
> 		select {
> ```

Wait, I don't follow. How would the resume loop _not_ be blocked on a receive from `progressCh`?

---

*[pkg/sql/jobs/jobs_test.go, line 436 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzoefPzDaXP-789T5Ph:-KzoefPzDaXP-789T5Pi:b-k94zse) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L436)):*
> ```Go
> 				t.Fatal(err)
> 			}
> 			check(t)
> ```

This is asserting the unresumed state, isn't it? Isn't that racy given that you've just committed the resume statement?

---

*[pkg/sql/jobs/jobs_test.go, line 537 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogQSej0wmEp9Ysr2J:-KzogQSej0wmEp9Ysr2K:b-46pvlu) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L537)):*
> ```Go
> 		check(t)
> 	})
> }
> ```

Love this test. (There _is_ quite a bit of global mutable state, but I don't have a good alternative.)

---

*[pkg/sql/jobs/jobs_test.go, line 556 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzohKmc7JvjFBWP2z6R:-KzohKmc7JvjFBWP2z6S:b7helr1) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L556)):*
> ```Go
> 			t.Fatal(err)
> 		}
> 		payload := job.Payload()
> ```

optional nit: just inline `job.Payload` below

---

*[pkg/sql/jobs/jobs_test.go, line 597 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzohO3Dnf_dWIHtq4bQ:-KzohO3Dnf_dWIHtq4bR:b7helr1) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L597)):*
> ```Go
> 			t.Fatal(err)
> 		}
> 		payload := job.Payload()
> ```

optional nit: just inline `job.Payload` below

---

*[pkg/sql/jobs/jobs_test.go, line 659 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzogq2TQChnqYh8HpOf:-Kzogq2TQChnqYh8HpOg:b-8z58s3) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L659)):*
> ```Go
> 		}
> 
> 		if err := woodyJob.Succeeded(ctx, nil); err != nil {
> ```

nit: inline comment on the `nil`

---

*[pkg/sql/jobs/jobs_test.go, line 706 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogtmDsiSaPZHfoGEV:-KzogtmDsiSaPZHfoGEW:bs4q9op) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L706)):*
> ```Go
> 		}
> 
> 		if err := buzzJob.Failed(ctx, errors.New("Buzz Lightyear can't fly"), nil); err != nil {
> ```

Ditto.

---

*[pkg/sql/jobs/jobs_test.go, line 733 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogvuEzYIqJirLZqpl:-KzogvuEzYIqJirLZqpm:bs4q9op) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L733)):*
> ```Go
> 		}
> 
> 		if err := sidJob.Failed(ctx, errors.New("Sid is a total failure"), nil); err != nil {
> ```

Ditto.

---

*[pkg/sql/jobs/jobs_test.go, line 757 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzogwm5JY9UlBvEPDT9:-Kzogwm5JY9UlBvEPDTA:bs4q9op) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L757)):*
> ```Go
> 				t.Fatal(err)
> 			}
> 			if err := job.Succeeded(ctx, nil); err != nil {
> ```

Ditto.

---

*[pkg/sql/jobs/jobs_test.go, line 768 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogxWSovpPR99MfqaV:-KzogxWSovpPR99MfqaW:bs4q9op) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L768)):*
> ```Go
> 			job, exp := createDefaultJob()
> 			exp.Error = "boom"
> 			if err := job.Failed(ctx, errors.New(exp.Error), nil); err != nil {
> ```

Ditto.

---

*[pkg/sql/jobs/jobs_test.go, line 769 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzokDtr9H91QFB0DKQ3:-KzokDtr9H91QFB0DKQ4:b-dcsj10) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L769)):*
> ```Go
> 				}
> 				job.Failed(ctx, errors.New("in-job error"))
> 				return errors.New("lease transferred")
> ```

These two tests are, I think, the most important tests for `RunAndWaitForTerminalState`. Can you preserve them by installing a dummy resumer that fails on the first time through but not the second?

---

*[pkg/sql/jobs/registry_external_test.go, line 85 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzomfnsAveS4ws6RBEY:-KzomfnsAveS4ws6RBEZ:b-kb36jh) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L85)):*
> ```Go
> 	clock := hlc.NewClock(hlc.UnixNano, time.Nanosecond)
> 	nodeLiveness := jobs.NewFakeNodeLiveness(clock, 4)
> 	getRegistry := func(id roachpb.NodeID) *jobs.Registry {
> ```

nit: ideally call this `newRegistry` to make clear that it's producing a new registry out of thin air rather than handing you a ref to an already-created registry

---

*[pkg/sql/jobs/registry_external_test.go, line 117 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzkq6OooavhmvEXU_Kf:-Kzkq6OooavhmvEXU_Kg:b-xd1k2u) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L117)):*
> ```Go
> 	jobMap := make(map[roachpb.NodeID]int64)
> 	hookCallCount := 0
> 	// resomeCounts maps jobs IDs to number of start/resumes.
> ```

"resomeCounts"

---

*[pkg/sql/jobs/registry_external_test.go, line 124 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzkqFEyJEnI5dR3UI9g:-KzkqFEyJEnI5dR3UI9h:b-ktin6c) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L124)):*
> ```Go
> 	// receive on it will block until a job is running.
> 	resumeCalled := make(chan struct{})
> 	defer close(done)
> ```

very optional nit: seems like this line might as well live directly beneath the declaration of `done`.

---

*[pkg/sql/jobs/registry_external_test.go, line 145 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzonGI86HWc6xEUApZm:-KzonGI86HWc6xEUApZn:b-czm9n) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L145)):*
> ```Go
> 	})
> 
> 	nodeID := &base.NodeIDContainer{}
> ```

This should just be a `roachpb.NodeID`. Actually, I think you can drop it entirely, and just construct the nodeID on the fly from `i`. See below.

---

*[pkg/sql/jobs/registry_external_test.go, line 160 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzoofNYTu5FUBpAz_eM:-KzoofNYTu5FUBpAz_eN:b810o3d) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L160)):*
> ```Go
> 
> 	nodeID.Reset(jobCount + 1)
> 	getRegistry(nodeID.Get())
> ```

I think we can drop this registry entirely now that we have one registry per "node."

---

*[pkg/sql/jobs/registry_external_test.go, line 204 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzooZoEyhiOrBvXYqgA:-KzooZoFYIQbFTFeX3CS:b-dvhsl2) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L204)):*
> ```Go
> 			return errors.Errorf("expected total jobs to be %d, but got %d", e, a)
> 		}
> 		lock.Unlock()
> ```

I think we can drop `newJobs` entirely (as well as the registry on the fourth "node"). Instead, you should be able to iterate resumeCounts to derive a total resumeCounts of (hopefully) 5.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

benesch:
Whew, ok, made it through everything! I've tried to use Reviewable's "dispositions" appropriately. Comments I've marked as "Blocking" are the only ones I feel strongly about; comments I've marked as "Discussing" are places where you should feel free to leave the code as-is.

---

Review status: all files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/restore.go, line 1185 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzkFTt_dAzg9QAM2UZb:-KzouQu5_glGkZAoA3AU:b-d3np73) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/restore.go#L1185)):*

Seems like we should make `r.tables` valid during the `OnFailOrCancel` callback too, so that we don't have to load the descriptors from the export store twice. Actually, we probably shouldn't be relying on the export store at all when cleaning up a failed restore. Ugh.

Don't worry about it in this PR, but seems worth filing as followup work.
</blockquote></details>

Filed #20261 to track.

---

*[pkg/sql/jobs/jobs_test.go, line 19 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzouVF5gTP1A9GNLZ6H:-KzouVF5gTP1A9GNLZ6I:bnnhvdy) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L19)):*
> ```Go
> import (
> 	gosql "database/sql"
> 	fmt "fmt"
> ```

Unnecessary alias. (`goimports` has been doing this to me too.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-:-Kzoq_IrJDBGnQs1ryT_:b-lza7b0)*

mjibson:
Review status: 7 of 16 files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/csv.go, line 1076 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjK_kTugGsYgY1mIxb:-Kzyvc0t3hcaRQ5z0tgD:bhs3klr) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/csv.go#L1076)):*

These inline `nil`s could use inline comments. (See similar comment in schema_changer.go.)
</blockquote></details>

Added jobs.NoopFn instead.

---

*[pkg/ccl/sqlccl/restore.go, line 716 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjUs436S3ALJsSdA3b:-KzywsSjOIjbzT0L_uOe:b-vu8m5p) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/restore.go#L716)):*

This function is only called from the restore success handler. I'd either inline it there or at least drop the `func() error` closure. Between the closure and this function's call site, we're double-`errors.Wrap`ing all errors, which seems unnecessary.
</blockquote></details>

I've removed the closure. I think this function should remain instead of being inlined because it nicely encapsulates a purpose.

---

*[pkg/sql/schema_changer.go, line 468 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjJsaHkd4TA2p3R-Zs:-KzyyGrns4HlyRPlUWo7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/schema_changer.go#L468)):*

Per Andrei's style guide additions, can you add an inline comment `/* onFailOrCancel */` next to the `nil` argument?
</blockquote></details>

Done.

---

*[pkg/sql/schema_changer.go, line 615 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjK3QzMET-izsR19sR:-KzyyHQMWdO5YwYmkWw0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/schema_changer.go#L615)):*

Ditto here.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 19 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzouVF5gTP1A9GNLZ6H:-KzyyYVeDQFXLQImzoow:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L19)):*

Unnecessary alias. (`goimports` has been doing this to me too.)
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 181 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzo085Zzg962P2jJsX3:-Kzz-QaRXdMisdkc50bW:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L181)):*

If you aren't opposed, I'd stick `dummyResumer` in helpers_test.go and rename it to `FakeResumer` for consistency with the other stub objects.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 379 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzoiv-DnuIpElWVu0xc:-L-05yH5CzRBWZgPghmc:bax5v9s) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L379)):*

Would be ideal if the relevant registry lifecycle tests would verify that `job paused` and `job canceled` were indeed the error messages returned when a job is paused and resumed, respectively.
</blockquote></details>

I've moved these error message tests to TestRegistryLifecycle.

---

*[pkg/sql/jobs/jobs_test.go, line 382 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzocov71zy1zuJjBJk_:-KzzCGG9YoJZKN0mVTmt:b9j1i2x) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L382)):*

Sending over the `progressCh` races with `check`, right? What if you sent on it twice to verify that you stay in the loop at least once?
</blockquote></details>

Can you explain the race to me? I can't see it. I'm not sure how two processCh sends would be any different than one. I did lots of testing with stressrace until it didn't find any problems, at least.

---

*[pkg/sql/jobs/jobs_test.go, line 402 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzodgVUmC_wuWrCP2fY:-KzzLAEdTuoVam35teUK:b-rj4jsh) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L402)):*

Wait, I don't follow. How would the resume loop _not_ be blocked on a receive from `progressCh`?
</blockquote></details>

Did some digging. This was happening when the OnResume func was receiving on progressCh, check runs, the job is PAUSED and commited, and THEN then progressCh err runs jobs.Progressed and finds out the job is paused. I've added an additional synchronization chan to prevent this.

---

*[pkg/sql/jobs/jobs_test.go, line 436 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzoefPzDaXP-789T5Ph:-L--Pf1QfMTOxGhgbEbv:b-av023g) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L436)):*

This is asserting the unresumed state, isn't it? Isn't that racy given that you've just committed the resume statement?
</blockquote></details>

Yes, removed.

---

*[pkg/sql/jobs/jobs_test.go, line 556 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzohKmc7JvjFBWP2z6R:-L--PqxPwdYrSUYVg6tc:b-yb5q85) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L556)):*

optional nit: just inline `job.Payload` below
</blockquote></details>

You can't do that: `cannot call pointer method on job.Payload()`. Weird right?

---

*[pkg/sql/jobs/jobs_test.go, line 659 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzogq2TQChnqYh8HpOf:-L--PzozwYbq88iI9Qt2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L659)):*

nit: inline comment on the `nil`
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 706 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogtmDsiSaPZHfoGEV:-L--QU_RnH7AhwP3ITPx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L706)):*

Ditto.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 733 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogvuEzYIqJirLZqpl:-L--QV2M9smUkVyF81iF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L733)):*

Ditto.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 757 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzogwm5JY9UlBvEPDT9:-L--QVYj4Z8eb_4S-EDx:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L757)):*

Ditto.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 768 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzogxWSovpPR99MfqaV:-L--QW-l61COM4JYq-vG:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L768)):*

Ditto.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/jobs_test.go, line 769 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzokDtr9H91QFB0DKQ3:-L-0CyKz7uTcJbp1FtwW:b-71mkm6) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L769)):*

These two tests are, I think, the most important tests for `RunAndWaitForTerminalState`. Can you preserve them by installing a dummy resumer that fails on the first time through but not the second?
</blockquote></details>

Why do we care about this function at all? It's only used in a benchmark in backup_test.go. Also I'm not exactly sure what this test is trying to do, so I'm not sure how to correctly port it to the new API.

---

*[pkg/sql/jobs/registry.go, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjBEPumHfXrhDaYFRt:-L-0DgB8zCijUa2k0G3d:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L68)):*

You could make this more self-documenting by using `context.CancelFunc` instead of `func()` directly.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjCn4EaiUNYs8DoqP6:-L-0DrQfRSwCNeShkUsA:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L116)):*

Wouldn't hurt to add a comment to the effect of "We create the job with a known ID, rather than relying on the column's `DEFAULT unique_rowid()`, to avoid a race condition where the job exists in the jobs table but is not yet present in our registry."
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry.go, line 127 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjEBPBdkY7WFeIabl0:-L-0HIzI_ueIneBpa0b7:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L127)):*

Seems to me like `resume` should take a `resumer`. Then it won't ever fail and won't need to return an `error`. Means calling `resume` will be a bit more annoying, as you'll have to call `getResumeHook` first, but I think that's actually desirable here. If you call `getResumeHook` before you register or create the job, then you can return early and avoid this complicated cleanup. You'll need to add a `Type() Type` method to `Details`, but that seems worthwhile.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry.go, line 209 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjFbXWBXAg-RNm41GK:-L-0HPV8rwK6nOtd50Ad:bonxdxl) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L209)):*

Intentional? The log context tags should take care of this, no?
</blockquote></details>

Debugging. Removed.

---

*[pkg/sql/jobs/registry.go, line 348 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzkH74lAXiP3KWq6TMJ:-L-0HWGdWIPSogOhM7LF:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L348)):*

nit: you call this an `errCh` everywhere else
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry.go, line 547 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjApRdfQ2R8Xp7Bg8e:-L-0ITeyAaaUcHsvxFBV:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry.go#L547)):*

"unregistired"
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry_external_test.go, line 85 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzomfnsAveS4ws6RBEY:-L-0IiO8SEHBwU3WPBHE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L85)):*

nit: ideally call this `newRegistry` to make clear that it's producing a new registry out of thin air rather than handing you a ref to an already-created registry
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry_external_test.go, line 117 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzkq6OooavhmvEXU_Kf:-L-0IlP8mS9xKCZI2FdR:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L117)):*

"resomeCounts"
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry_external_test.go, line 124 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzkqFEyJEnI5dR3UI9g:-L-0Ir82wdoBakSThAr2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L124)):*

very optional nit: seems like this line might as well live directly beneath the declaration of `done`.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry_external_test.go, line 145 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzonGI86HWc6xEUApZm:-L-0JjozhxJvek-f7Ng2:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L145)):*

This should just be a `roachpb.NodeID`. Actually, I think you can drop it entirely, and just construct the nodeID on the fly from `i`. See below.
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry_external_test.go, line 160 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzoofNYTu5FUBpAz_eM:-L-0JkEcgMocOZZt5s9_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L160)):*

I think we can drop this registry entirely now that we have one registry per "node."
</blockquote></details>

Done.

---

*[pkg/sql/jobs/registry_external_test.go, line 204 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzooZoEyhiOrBvXYqgA:-L-0KKqoUBaUARmYXreu:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L204)):*

I think we can drop `newJobs` entirely (as well as the registry on the fourth "node"). Instead, you should be able to iterate resumeCounts to derive a total resumeCounts of (hopefully) 5.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

benesch:
Reviewed 9 of 9 files at r2.
Review status: all files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[pkg/ccl/sqlccl/restore.go, line 716 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzjUs436S3ALJsSdA3b:-L-0O47cWMXu1yIZRqvb:b-81nlyp) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/ccl/sqlccl/restore.go#L716)):*

I've removed the closure. I think this function should remain instead of being inlined because it nicely encapsulates a purpose.
</blockquote></details>

SGTM!

---

*[pkg/sql/jobs/jobs_test.go, line 556 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzohKmc7JvjFBWP2z6R:-L-0OXNa-yvISY01ppRI:bwq70wo) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L556)):*

You can't do that: `cannot call pointer method on job.Payload()`. Weird right?
</blockquote></details>

Sigh. Go.

---

*[pkg/sql/jobs/jobs_test.go, line 769 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzokDtr9H91QFB0DKQ3:-L-0OsUZwrOokcgFDnw2:b8xrsiy) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L769)):*

Why do we care about this function at all? It's only used in a benchmark in backup_test.go. Also I'm not exactly sure what this test is trying to do, so I'm not sure how to correctly port it to the new API.
</blockquote></details>

Those are the nightly 2tb backup/restore acceptance tests, so they're more important than your average test. (We use benchmarks instead of tests so we get timing information or something.) 

The tests you removed were checking that RunAndWaitForTerminalState reported success when the job was transferred to a new coordinator and succeeded there, rather than reporting "ambiguous result (job lease canceled)." The original test code managed to simulate this circumstance by returning similar errors in the right places rather than actually causing lease transfers.

There's probably an argument to be made that we don't need to write tests for our tests, but check with Dan if you want to remove these particular tests. (He's RunAndWaitForTerminalState's original author.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

danhhz:
Review status: 11 of 16 files reviewed at latest revision, 22 unresolved discussions, all commit checks successful.

---

*[pkg/sql/jobs/jobs_test.go, line 769 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzokDtr9H91QFB0DKQ3:-L-28BTHz9fNF1oAxzSo:bc3a29b) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L769)):*

Those are the nightly 2tb backup/restore acceptance tests, so they're more important than your average test. (We use benchmarks instead of tests so we get timing information or something.) 

The tests you removed were checking that RunAndWaitForTerminalState reported success when the job was transferred to a new coordinator and succeeded there, rather than reporting "ambiguous result (job lease canceled)." The original test code managed to simulate this circumstance by returning similar errors in the right places rather than actually causing lease transfers.

There's probably an argument to be made that we don't need to write tests for our tests, but check with Dan if you want to remove these particular tests. (He's RunAndWaitForTerminalState's original author.)
</blockquote></details>

I haven't really looked at the new jobs api, but if it's possible to keep these tests, we should. This was all very annoying to get right the first time since the iteration cycle is O(hour), so I'd hate to see RunAndWaitForTerminalState rot.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

mjibson:
Review status: 10 of 16 files reviewed at latest revision, 22 unresolved discussions, some commit checks pending.

---

*[pkg/sql/jobs/jobs_test.go, line 769 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzokDtr9H91QFB0DKQ3:-L-2UnEw0DQkV5aPZQh3:b-6idhf8) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L769)):*

I haven't really looked at the new jobs api, but if it's possible to keep these tests, we should. This was all very annoying to get right the first time since the iteration cycle is O(hour), so I'd hate to see RunAndWaitForTerminalState rot.
</blockquote></details>

Ok I've adapted these tests to just not use the registry, which I think achieves the same actual testing that was done before, but without having to really simulate a lease transfer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

benesch:
:lgtm_strong: modulo one last sync issue noted below

---

Reviewed 6 of 6 files at r3.

---

*[pkg/sql/jobs/jobs_test.go, line 382 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzocov71zy1zuJjBJk_:-L-2ptLhgCBUrKhm_xD2:brmy16y) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L382)):*

Can you explain the race to me? I can't see it. I'm not sure how two processCh sends would be any different than one. I did lots of testing with stressrace until it didn't find any problems, at least.
</blockquote></details>

Before you added the `resumeCheckCh`, there was no guarantee that sending over `progressCh` would trigger a call to `Progressed` before you called `check`. I think. So sometimes you were potentially asserting state before the `Rollback` had been processed. Looks like `resumeCheckCh` fixes that nicely though!

---

*[pkg/sql/jobs/jobs_test.go, line 394 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzod5tgRl6U3qeqiVkV:-L-2qrXuoBQkVwWInYq6:b-m38kvc) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L394)):*

Ditto here. Don't you want to send on the `progressCh` to actually trigger a call to `Progressed`?
</blockquote></details>

I think you want to send over `progressCh` and `resumeCheckCh` here for the reasons described above.

---

*[pkg/sql/jobs/jobs_test.go, line 402 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzodgVUmC_wuWrCP2fY:-L-2qNrSCrmHE7qvwMR3:b-83sjoh) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L402)):*

Did some digging. This was happening when the OnResume func was receiving on progressCh, check runs, the job is PAUSED and commited, and THEN then progressCh err runs jobs.Progressed and finds out the job is paused. I've added an additional synchronization chan to prevent this.
</blockquote></details>

Nice!

---

*[pkg/sql/jobs/jobs_test.go, line 769 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzokDtr9H91QFB0DKQ3:-L-2oGGXVB9MRJRzaUs1:b-83sjoh) ([raw file](https://github.com/cockroachdb/cockroach/blob/b6fdb185b3aa1003932655faba67555608dcf7da/pkg/sql/jobs/jobs_test.go#L769)):*

Ok I've adapted these tests to just not use the registry, which I think achieves the same actual testing that was done before, but without having to really simulate a lease transfer.
</blockquote></details>

Nice!

---

*[pkg/sql/jobs/registry.go, line 500 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-L-2otLBQ_zCmtnAkqPB:-L-2otLCJab1gtPhFc0J:bnlcufh) ([raw file](https://github.com/cockroachdb/cockroach/blob/cd2973b6be5986621675c6d4a33e1656f356f5dc/pkg/sql/jobs/registry.go#L500)):*
> ```Go
> 		if err != nil {
> 			if log.V(2) {
> 				log.Infof(ctx, "skipping job %d: %s", *id, err)
> ```

This logging is probably still useful when debugging the registry, but the logging in the jobs package is generally worthless/too spammy, so not worth worrying about now I suppose.

---

*[pkg/sql/jobs/registry_external_test.go, line 160 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzoofNYTu5FUBpAz_eM:-L-2pF7FmJTJfr-r0zbY:b-tsyuad) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L160)):*

Done.
</blockquote></details>

Looks like you're still creating a registry on a fourth node, but I don't think that's actually necessary. Is there something I'm missing?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-:-L-2r8QwKnVRjZquBDi2:bcc2nv4)*

mjibson:

---

*[pkg/sql/jobs/jobs_test.go, line 394 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzod5tgRl6U3qeqiVkV:-L-2tCRJ4TRlpT_HPqxT:b-ovtcz9) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L394)):*

I think you want to send over `progressCh` and `resumeCheckCh` here for the reasons described above.
</blockquote></details>

I just tried this and it blocks until the test timesout, I think because of the txn intent already on the job. So when .Progressed goes and attempts to read the job, it is prevented from doing so. So unless you object I'm going to leave this test as is.

---

*[pkg/sql/jobs/registry_external_test.go, line 160 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-KzoofNYTu5FUBpAz_eM:-L-2tnhwyMtnLGSdC0mm:b-pc25er) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/registry_external_test.go#L160)):*

Looks like you're still creating a registry on a fourth node, but I don't think that's actually necessary. Is there something I'm missing?
</blockquote></details>

No, you're right. I misunderstood the previous comment. Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

benesch:
:shipit: 

---

Reviewed 1 of 1 files at r4.

---

*[pkg/sql/jobs/jobs_test.go, line 394 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-Kzod5tgRl6U3qeqiVkV:-L-3_GVC0pIDK76VpN06:b-s6xfj0) ([raw file](https://github.com/cockroachdb/cockroach/blob/e5a005bab45c10394fa0e421af81c70bcb15027c/pkg/sql/jobs/jobs_test.go#L394)):*

I just tried this and it blocks until the test timesout, I think because of the txn intent already on the job. So when .Progressed goes and attempts to read the job, it is prevented from doing so. So unless you object I'm going to leave this test as is.
</blockquote></details>

Ugh, makes sense. Fine with this as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-:-L-3_P8A-o1h2wJPSFtq:bizlbj6)*

benesch:
If you weren't already planning to, let's optimistically close all the related stress failure issues once this lands.

mjibson:
Yes, planning on doing that.

lego:
Didn't do a thorough review, but the new interface looks great :). Thanks so much @mjibson!

---

Review status: 15 of 16 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185#-:-L-2rgvb5bU6BWX4_5Bh:b-slrx37)*

benesch:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20185)*

mjibson:
Ugh I mangled the branches. Will fix when I get into work tomorrow.

mjibson:
Thanks for the reviews.

Issue #20120, opened 2017-11-17
sql: add JSON builtins


justinj:
Tracking issue for various JSON builtins. Most of these have a `json` version and a corresponding `jsonb` version. Details [here](https://www.postgresql.org/docs/9.5/static/functions-json.html).

I think basically any of these would be a good first PR:

* [x] `to_json`
* [x] `array_to_json`
* [x] `json_build_array`
* [x] `json_build_object`
* [x] `json_object`
* [x] `json_array_length`
* [x] `json_each`
* [x] `json_each_text`
* [x] `json_extract_path`
* [x] `json_extract_path_text`
* [x] `json_object_keys`
* [x] `json_array_elements`
* [x] `json_array_elements_text`
* [x] `json_typeof`
* [x] `json_strip_nulls`
* [x] `jsonb_set`
* [x] `jsonb_pretty`
* [x] `jsonb_insert`

The below exist in Postgres, but don't apply to CockroachDB right now since we don't support row or record types:
* [ ] ~`row_to_json`~
* [ ] ~`json_populate_record`~
* [ ] ~`json_populate_recordset`~
* [ ] ~`json_to_record`~
* [ ] ~`json_to_recordset`~

joe-el-khoury:
Are the ones with the ~strikethrough~ complete or not required?

justinj:
Hey, @joe-el-khoury, I should have been more clear in the description - the strikethough'd functions don't apply to us right now since we don't have row or record types.

joe-el-khoury:
Currently trying to build cockroachdb, will try to tackle one once I'm able to build it. Should be a fun Go learning exercise.

justinj:
Sweet, feel free to hit me up here or on gitter if you have any questions. If you're looking for ones that would be particularly straightforward given that you're new to Go, `json_array_length` and `json_typeof` should be not bad (look around `builtins.go` to get a feel for how builtins are implemented), `json_extract_path` and `json_extract_path_text` already have implementations to support the `#>` and `#>>` operators. Ones like `json_each` will be interesting I think, but a bit more involved, since they require returning a set of results. You can check out the implementation of `generate_series` and others in `generator_builtins.go` for some guidance on those.

joe-el-khoury:
[...]

PR #20190, opened 2017-11-21
sql: add JSON builtins json_typeof

heyihong:
Hi there, I implement the json_typeof function of JSON builtins #20120 in the sql module. Please take a look.

### DIFF
--- a/pkg/sql/sem/tree/datum.go
+++ b/pkg/sql/sem/tree/datum.go
@@ -2074,6 +2074,30 @@ func MakeDJSON(d interface{}) (Datum, error) {
 	return &DJSON{j}, nil
 }
 
+// AsDJSON attempts to retrieve a *DJSON from an Expr, returning a *DJSON and
+// a flag signifying whether the assertion was successful. The function should
+// be used instead of direct type assertions wherever a *DJSON wrapped by a
+// *DJSON is possible.
+func AsDJSON(e Expr) (*DJSON, bool) {
+	switch t := e.(type) {
+	case *DJSON:
+		return t, true
+	case *DOidWrapper:
+		return AsDJSON(t.Wrapped)
+	}
+	return nil, false
+}
+
+// MustBeDJSON attempts to retrieve a DJSON from an Expr, panicking if the
+// assertion fails.
+func MustBeDJSON(e Expr) DJSON {
+	i, ok := AsDJSON(e)
+	if !ok {
+		panic(pgerror.NewErrorf(pgerror.CodeInternalError, "expected *DJSON, found %T", e))
+	}
+	return *i
+}
+
 // ResolvedType implements the TypedExpr interface.
 func (*DJSON) ResolvedType() types.T {
 	return types.JSON
--- a/docs/generated/sql/functions.md
+++ b/docs/generated/sql/functions.md
@@ -355,6 +355,17 @@ Compatible elements: hour, minute, second, millisecond, microsecond.</p>
 </span></td></tr></tbody>
 </table>
 
+### JSONB Functions
+
+<table>
+<thead><tr><th>Function &rarr; Returns</th><th>Description</th></tr></thead>
+<tbody>
+<tr><td><code>json_typeof(val: jsonb) &rarr; <a href="string.html">string</a></code></td><td><span class="funcdesc"><p>Returns the type of the outermost JSON value as a text string.</p>
+</span></td></tr>
+<tr><td><code>jsonb_typeof(val: jsonb) &rarr; <a href="string.html">string</a></code></td><td><span class="funcdesc"><p>Returns the type of the outermost JSON value as a text string.</p>
+</span></td></tr></tbody>
+</table>
+
 ### Math and Numeric Functions
 
 <table>
--- a/pkg/sql/logictest/testdata/logic_test/json_builtins
+++ b/pkg/sql/logictest/testdata/logic_test/json_builtins
@@ -1,0 +1,73 @@
+# LogicTest: default distsql parallel-stmts
+
+## json_typeof and jsonb_typeof
+
+query T
+SELECT json_typeof('-123.4'::JSON)
+----
+number
+
+query T
+SELECT jsonb_typeof('-123.4'::JSON)
+----
+number
+
+query T
+SELECT json_typeof('"-123.4"'::JSON)
+----
+string
+
+query T
+SELECT jsonb_typeof('"-123.4"'::JSON)
+----
+string
+
+query T
+SELECT json_typeof('{"1": {"2": 3}}'::JSON)
+----
+object
+
+query T
+SELECT jsonb_typeof('{"1": {"2": 3}}'::JSON)
+----
+object
+
+query T
+SELECT json_typeof('[1, 2, [3]]'::JSON)
+----
+array
+
+query T
+SELECT jsonb_typeof('[1, 2, [3]]'::JSON)
+----
+array
+
+query T
+SELECT json_typeof('true'::JSON)
+----
+true
+
+query T
+SELECT jsonb_typeof('true'::JSON)
+----
+true
+
+query T
+SELECT json_typeof('false'::JSON)
+----
+false
+
+query T
+SELECT jsonb_typeof('false'::JSON)
+----
+false
+
+query T
+SELECT json_typeof('null'::JSON)
+----
+null
+
+query T
+SELECT jsonb_typeof('null'::JSON)
+----
+null
--- a/pkg/sql/sem/builtins/builtins.go
+++ b/pkg/sql/sem/builtins/builtins.go
@@ -1495,6 +1495,10 @@ CockroachDB supports the following flags:
 	// implemented.
 	},
 
+	"json_typeof": {jsonTypeOfImpl},
+
+	"jsonb_typeof": {jsonTypeOfImpl},
+
 	"ln": {
 		floatBuiltin1(func(x float64) (tree.Datum, error) {
 			return tree.NewDFloat(tree.DFloat(math.Log(x))), nil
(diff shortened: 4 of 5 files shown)


justinj:
Hi @heyihong! Thanks for much for the contribution! This looks great - just a couple comments.

---

---

*[pkg/sql/logictest/testdata/logic_test/json_builtins, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRuHUkIgeThfVpL0E5:-KzRuHUljMXe9FWb4OoJ:b-y4tfoa) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/logictest/testdata/logic_test/json_builtins#L1)):*
> ```
> # LogicTest: default distsql parallel-stmts
> ```

Good call to make a new file for this, I would have stuffed it in the `json` test file but I think this is cleaner!

---

*[pkg/sql/sem/builtins/builtins.go, line 1499 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRrrrkraD3efB1MNhM:-KzRrrrkraD3efB1MNhN:b89s5v8) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/sem/builtins/builtins.go#L1499)):*
> ```Go
> 	},
> 
> 	"json_typeof": {
> ```

We should actually support both `json_typeof` and `jsonb_typeof`, though they should be the same. You can accomplish this by pulling the `tree.Builtin` struct below up into a value and then reusing it across both cases. (CockroachDB only implements what Postgres calls JSONB, but then aliases `JSON` to it).

---

*[pkg/sql/sem/builtins/builtins.go, line 1501 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRsQT1lfPjuHVdfVK9:-KzRsQT1lfPjuHVdfVKA:b8rnqwb) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/sem/builtins/builtins.go#L1501)):*
> ```Go
> 	"json_typeof": {
> 		tree.Builtin{
> 			Types:      tree.ArgTypes{{"val", types.String}},
> ```

This should operate on `types.JSON`, rather than `types.String`. This appears to work as it does in Postgres because the string literal passed in Postgres is automatically parsed and converted to a JSON datum, but the function itself operates on `JSON`:
```
justin=# select json_typeof('1'::json);
 json_typeof
-------------
 number
(1 row)

justin=# select json_typeof('1'::text);
ERROR:  function json_typeof(text) does not exist
LINE 1: select json_typeof('1'::text);
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

---

*[pkg/sql/sem/builtins/builtins.go, line 1504 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRsqbgCtp7kZynKFYm:-KzRsqbgCtp7kZynKFYn:b-oc38v) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/sem/builtins/builtins.go#L1504)):*
> ```Go
> 			ReturnType: tree.FixedReturnType(types.String),
> 			Fn: func(_ *tree.EvalContext, args tree.Datums) (tree.Datum, error) {
> 				str := string(tree.MustBeDString(args[0]))
> ```

That means that the value coming in here will already be a `DJSON`, so there's no need to parse it.

---

*[pkg/util/json/json.go, line 82 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRsyXr2aFcJKlFVjnz:-KzRsyXr2aFcJKlFVjo-:b-413dm3) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/util/json/json.go#L82)):*
> ```Go
> 
> 	// TypeAsText returns the type of JSON document as a string.
> 	TypeAsText() *string
> ```

This can return a `string`, rather than a `*string`. The reason `AsText` returns `*string` is that it has the possibility of returning `nil` to signal a sql `NULL`, but `TypeAsText` doesn't have that problem.

---

*[pkg/util/json/json.go, line 673 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRtCWYVa3u4TnJgSZF:-KzRtCWYVa3u4TnJgSZG:bzgw93h) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/util/json/json.go#L673)):*
> ```Go
> func (j jsonString) TypeAsText() *string {
> 	s := j.jsonType().String()
> 	return &s
> ```

I would just inline the return of the value here, no need to delegate to the type, so
```
func (jsonString) TypeAsText() string { return "string" }
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-:-KzRrhfPFKkOmLLsR8tg:b-74xh97)*

heyihong:
Hi @justinj, Thanks for your careful review. Please take an another look.

---

Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-:-KzSKcAHw8a1F-R4g7m2:bjswb0m)*

pkg/sql/sem/tree/datum.go
```diff
@@ -2074,6 +2074,26 @@ func MakeDJSON(d interface{}) (Datum, error) {
 	return &DJSON{j}, nil
 }
 
+func AsDJSON(e Expr) (*DJSON, bool) {
```
justinj: Our CI/linter is complaining because this is an exported function without a comment.

justinj:
All the changes look great! Three final things:

1. Our linter caught a small style nit that needs to be fixed before CI will pass, after fixing that
2. our docs generator needs to be re-run manually on any new builtins, so can you run `make generate PKG=./docs/...`, commit the result and then
3. squash all your commits into a single commit?

Thanks!

heyihong:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[pkg/sql/logictest/testdata/logic_test/json_builtins, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRuHUkIgeThfVpL0E5:-KzURM2pVuSji09m4MCB:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/logictest/testdata/logic_test/json_builtins#L1)):*

Good call to make a new file for this, I would have stuffed it in the `json` test file but I think this is cleaner!
</blockquote></details>

Done.

---

*[pkg/sql/sem/builtins/builtins.go, line 1499 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRrrrkraD3efB1MNhM:-KzURJW_NthMDwD5ar4T:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/sem/builtins/builtins.go#L1499)):*

We should actually support both `json_typeof` and `jsonb_typeof`, though they should be the same. You can accomplish this by pulling the `tree.Builtin` struct below up into a value and then reusing it across both cases. (CockroachDB only implements what Postgres calls JSONB, but then aliases `JSON` to it).
</blockquote></details>

Done.

---

*[pkg/sql/sem/builtins/builtins.go, line 1501 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRsQT1lfPjuHVdfVK9:-KzURKLtR8H_pDI0fF5v:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/sem/builtins/builtins.go#L1501)):*

This should operate on `types.JSON`, rather than `types.String`. This appears to work as it does in Postgres because the string literal passed in Postgres is automatically parsed and converted to a JSON datum, but the function itself operates on `JSON`:
```
justin=# select json_typeof('1'::json);
 json_typeof
-------------
 number
(1 row)

justin=# select json_typeof('1'::text);
ERROR:  function json_typeof(text) does not exist
LINE 1: select json_typeof('1'::text);
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```
</blockquote></details>

Done.

---

*[pkg/sql/sem/builtins/builtins.go, line 1504 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRsqbgCtp7kZynKFYm:-KzURKbieDHCsDHiGjFL:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/sql/sem/builtins/builtins.go#L1504)):*

That means that the value coming in here will already be a `DJSON`, so there's no need to parse it.
</blockquote></details>

Done.

---

*[pkg/sql/sem/tree/datum.go, line 2077 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzSJYmLDgZNvSxmsftO-r2-2077:-KzURHW4-tbqddQTtX9I:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/7570d8e7491177d733bb7223c58e7e86af0c935d/pkg/sql/sem/tree/datum.go#L2077)):*

Our CI/linter is complaining because this is an exported function without a comment.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 82 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRsyXr2aFcJKlFVjnz:-KzURI4nrB-oKAaKJTAa:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/util/json/json.go#L82)):*

This can return a `string`, rather than a `*string`. The reason `AsText` returns `*string` is that it has the possibility of returning `nil` to signal a sql `NULL`, but `TypeAsText` doesn't have that problem.
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 673 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190#-KzRtCWYVa3u4TnJgSZF:-KzURIPNTmUdI4KFD_Ty:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/0b2a98aaf4aac5c41d9298ad16e22a92f9562661/pkg/util/json/json.go#L673)):*

I would just inline the return of the value here, no need to delegate to the type, so
```
func (jsonString) TypeAsText() string { return "string" }
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20190)*

justinj:
Wonderful, thanks for the help. Hope we see you again!

PR #20191, opened 2017-11-21
sql: ensure all data belongs to a partition in partitioned tables

benesch:
Primary commit message:

```
Unlike range partitions, which must cover all possible values, list
partitions are explicitly permitted to omit the catch-all DEFAULT
partition. In this case, we want to prevent the user from writing data
that does not belong to any partition.

To this end, this commit teaches `CREATE TABLE ... PARTITION BY` to
install a CHECK constraint that rejects rows that would not belong to a
partition. This CHECK constraint is marked as "derived" so that users
cannot drop it except by dropping the table's partitioning scheme.*

*Partitioning schemes cannot yet be modified or dropped.
```

Additional prep work in the first two commits.


danhhz:
sorry for the radio silence!

general approach looks good, but i haven't done a detailed review because i've been mulling over whether we should revisit requiring MAXVALUE in range partitions. i'm starting to think no (which would mean we have the same issues in range partitioning as list). it seems like an artificial constraint and other partitioning impls allow it. wdyt?

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning_constraints, line 56 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzUJD03iy5lrOEtrsJ3:-KzUJD03iy5lrOEtrsJ4:bykpuom) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8067d4b37d050055e1c666d84c26de94bd0c3f4/pkg/sql/logictest/testdata/logic_test/partitioning_constraints#L56)):*
> ```
> statement ok
> ----
> CREATE TABLE ok4 (a INT, b INT, c INT, PRIMARY KEY (a, b)) PARTITION BY LIST (a, b) (
> ```

what's the difference between the ok4 tests and the ok3 tests?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-KzVWZ2nRaUmbkC3dQsO:b-6uon17)*

benesch:
So I think the arbitrary constraint is really requiring contiguous range partitions starting from the beginning of the keyspace. I think we should either allow arbitrary non-overlapping ranges a la

```
PARTITION foo VALUES BETWEEN lower_bound AND upper_bound
```

or require that you partition the entire keyspace. Otherwise the fact that you can omit exactly one partition—the partition from the last declared boundary point and the maximum possible value—seems like a weird accident. If partitions were declared in the reverse order (i.e., if you named each partition's lower bound instead of its upper bound) I suspect we'd be discussing whether to allow omitting the `MINVALUE` partition. It seems weSeems like we should either add support for non-contiguous range partitions in a principled way or remove the max-partition exception. Thoughts?

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning_constraints, line 56 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzUJD03iy5lrOEtrsJ3:-KzVcQWA37m-CXwI_C6f:b-swi8gq) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8067d4b37d050055e1c666d84c26de94bd0c3f4/pkg/sql/logictest/testdata/logic_test/partitioning_constraints#L56)):*

what's the difference between the ok4 tests and the ok3 tests?
</blockquote></details>

Whoops, I probably copy/pasted it and never got around to making it different. Will fix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-KzV_Hcf0CPTUCRxjxDa:bgd1l8o)*

benesch:
s/It seems we//

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-KzVdSXWRasAmiKE81jj:b-byynpz)*

danhhz:
Okay, seems like we should step back and discuss. (Sorry this has all only come up after you finished this PR!) I haven't made up my mind, but I'm definitely a lean toward a non-required MAXVALUE partition. My thoughts:

I predict that requiring MAXVALUE is going to be confusing/onerous to users (though given our q/a buddy experience, I could certainly be wrong on that). Oracle and MySQL both do not require the entire space to be covered and will reject anything that cannot be mapped to a partition, exactly like list with no DEFAULT.

SQL Server has a syntax more like `PARTITION foo VALUES BETWEEN lower_bound AND upper_bound`, which I am definitely open to. Maybe that is indeed our sweet spot; our q/a buddy was unhappy that the range partitions had to be sorted by value (and I think it's waaay too confusing to let them stay unsorted).

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-KzYyUBVocSTyfKDjs9a:bz8sxbh)*

knz:
Argh I am just looking at this now and I feel very paged out from the topic. What are we looking at here?

I have  separate concerns, one related to my ignorance and one structural about the commit messages:

1) I don't understand why this work needs to be done -- why do we want all the data in a partition in the first place? I don't recall this from the RFC. If you have just recently developed a reason, that's all right, but then please mention so in the PR description and one or more of the commit messages.

2) the commit message for the first commit says "there's a problem, this patch averts it" but it does not say *how* the problem is averted. The commit message should include an example that shows what this patch makes possible.

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-KzZNVKTmut21SH0pJn4:b2okeqm)*

jordanlewis:
I agree that it seems onerous and not necessarily useful to enforce that all data be included in a partition. Seems like it would make initial partitioning setup a pain. I can imagine a use case where all data for a single row must stay somewhere but everything else should be evenly distributed (or just "whatever" distributed).

Perhaps if we are super divided on this we could add an optional constraint (PARTITION ALL BY... vs PARTITION BY? MUST PARTITION BY? just spitballing) that is a switch for this behavior.

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-Kz_-kOCMxsWUGaMPBZ_:b-dfdlu4)*

benesch:
> I agree that it seems onerous and not necessarily useful to enforce that all data be included in a partition. Seems like it would make initial partitioning setup a pain. I can imagine a use case where all data for a single row must stay somewhere but everything else should be evenly distributed (or just "whatever" distributed).

That's what `DEFAULT` partitions are for!

```
CREATE TABLE users (
  country STRING,
  email STRING ...,
  PRIMARY KEY (country, email)
) PARTITION BY LIST (country) (
  PARTITION us VALUES IN ('us'),
  PARTITION everything_else VALUES IN (DEFAULT)
)
```

When partitioning for regulatory compliance, it'll be very important to have a mode where data that does not belong to any partition is rejected. I think we want that restrictive mode to be the default to prevent accidents. Specifying a catch-all `DEFAULT` partition when that behavior is desired doesn't seem too onerous to me.

Our story for range partitioning isn't nearly so nice at the moment. Suppose you want articles from 2013 and 2015 to live on slower hardware. (Contrived, I know.) We currently force you to declare dummy interstitial range partitions:

```
CREATE TABLE articles (
  year_published INT,
  id INT,
  PRIMARY KEY (year_published, id)
) PARTITION BY RANGE (year) (
  PARTITION before_2013 VALUES < 2013,
  PARTITION during_2013 VALUES < 2014,
  PARTITION between_2013_and_2015 VALUES < 2015,
  PARTITION during_2015 VALUES < 2016
)
```

You're not, however, required to specify a dummy `after_2015` partition. That is, you can omit:

```
PARTITION after_2015 VALUES < MAXVALUE
```

This is weird and arbitrary, and just a result of how the syntax anchors the start of each range to the end of the last specified range. I think the best option is to eventually provide support for a syntax like this:

```
CREATE TABLE articles (
  year_published INT,
  id INT,
  PRIMARY KEY (year_published, id)
) PARTITION BY RANGE (year) (
  PARTITION during_2013 VALUES BETWEEN 2013 AND 2014,
  PARTITION during_2015 VALUES BETWEEN 2015 and 2016
)
```

This makes the example above way cleaner, but there's a ton of details to work out. Do we support `MAXVALUE` _and_ `MINVALUE`? Maybe even an additional catch-all partition for the remaining gaps? If no catch-all partition is specified, do we install a `CHECK` constraint to reject values that belong to no partition?

For now, though, spelling out a bunch of extraneous, unused range partitions is not so bad. At worst it doubles the number of partitions you need to specify.

My preference is to land this PR for now while planning to eventually support the more flexible `BETWEEN` syntax. It really bugs me that you can create a gap between the last-declared partition and the maximum  possible value, but nowhere else.

Thoughts?

danhhz:
Did a full review of the last two commits and they look good. Given that we're now leaning toward a followup where range partitionings are allowed to have holes, I'm inclined to omit the first commit. It seems to me that it will only create unnecessary churn (for example, you're removing some tests here that we probably want back). Range can be left unpolished while we figure it out

---

Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[pkg/sql/create.go, line 1269 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyAG0wM-0E2c1SYH62:-KzyAG0wM-0E2c1SYH63:b9shmie) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/create.go#L1269)):*
> ```Go
> 	maybeTuple tree.Expr,
> 	cols []sqlbase.ColumnDescriptor,
> ) (tree.Datums, []byte, error) {
> ```

this deserves a callout in the method doc

---

*[pkg/sql/create.go, line 1389 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyCXosi6AgoLmT25-S:-KzyCXosi6AgoLmT25-T:b-samdhp) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/create.go#L1389)):*
> ```Go
> 			}
> 			p.Values = append(p.Values, encodedTuple)
> 			partCheckExpr = tree.NewTypedOrExpr(partCheckExpr, tree.NewTypedComparisonExpr(tree.EQ,
> ```

the way the check constraint construction here works is :ok_hand:! maybe add a tiny comment about how it works from a high level (really i'm mostly interested in DEFAULT)

---

*[pkg/sql/create.go, line 1451 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyArLzmKaOqM3q6-45:-KzyArLzmKaOqM3q6-46:b-7dj2l4) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/create.go#L1451)):*
> ```Go
> 		}
> 		// In order to typecheck during simplification and normalization, we used
> 		// dummy IndexVars. Swap them out for acutal column references.
> ```

nit: actual

---

*[pkg/sql/logictest/testdata/logic_test/partitioning_constraints, line 56 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzUJD03iy5lrOEtrsJ3:-KzyDD4cLRygOFxCB1Vb:b-2vb6a9) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8067d4b37d050055e1c666d84c26de94bd0c3f4/pkg/sql/logictest/testdata/logic_test/partitioning_constraints#L56)):*

Whoops, I probably copy/pasted it and never got around to making it different. Will fix.
</blockquote></details>

am i missing something or is this still the same?

---

*[pkg/sql/logictest/testdata/logic_test/partitioning_constraints, line 26 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyCyWe2RXxujKuBy__:-KzyCyWe2RXxujKuBy_a:b9fdwkd) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/logictest/testdata/logic_test/partitioning_constraints#L26)):*
> ```
> )
> 
> query T
> ```

comment with why this is expected to be empty, plz

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-KzyE3FOhcs5aOc0_82Q:b-362d4a)*

benesch:
Oooh, super happy to rebase out the first commit! SGTM.

benesch:
Review status: 0 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/create.go, line 1269 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyAG0wM-0E2c1SYH62:-L-3h7f4KiE7RgFE3_Hs:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/create.go#L1269)):*

this deserves a callout in the method doc
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1389 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyCXosi6AgoLmT25-S:-L-3hAkT6zPuPMoYqecc:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/create.go#L1389)):*

the way the check constraint construction here works is :ok_hand:! maybe add a tiny comment about how it works from a high level (really i'm mostly interested in DEFAULT)
</blockquote></details>

Done.

---

*[pkg/sql/create.go, line 1451 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyArLzmKaOqM3q6-45:-L-3hBOlv9UwxKiBeysj:bo3vl6y) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/create.go#L1451)):*

nit: actual
</blockquote></details>

Thanks, done.

---

*[pkg/sql/logictest/testdata/logic_test/partitioning_constraints, line 56 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzUJD03iy5lrOEtrsJ3:-L-3gsjblpVb9uihPnZz:bwpv780) ([raw file](https://github.com/cockroachdb/cockroach/blob/a8067d4b37d050055e1c666d84c26de94bd0c3f4/pkg/sql/logictest/testdata/logic_test/partitioning_constraints#L56)):*

am i missing something or is this still the same?
</blockquote></details>

Not missing anything; hadn't pushed yet. It's now fixed!

---

*[pkg/sql/logictest/testdata/logic_test/partitioning_constraints, line 26 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-KzyCyWe2RXxujKuBy__:-L-3gwBErT2ZmfE9ccqE:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/af8c22ed3f17f5857d88601057175702abca5457/pkg/sql/logictest/testdata/logic_test/partitioning_constraints#L26)):*

comment with why this is expected to be empty, plz
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191)*

danhhz:
Reviewed 8 of 8 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191)*

benesch:
Ok, thanks to @danhhz's advice I've much reduced the scope of this sucker. We simply no longer attempt to handle range partitions, nor do we attempt to install CHECK constraints for indexes added after partition time. We only prevent rows that do not belong to any list partition, if said partitioning scheme was specified at table creation scheme.

Hoping to land this before it rots due to the refactor because @danhhz is happy with the general approach. Definitely planning to come up with a more consistent story for gaps between partitions before 2.0.

@knz did you want to take a more thorough look?

---

Review status: 0 of 9 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-L-88rxLjobT2KSTRCpz:bn9r2xi)*

knz:
I did have a look

1) you probably want to extend `(*TableDescriptor).ValidateTable()` in `sqlbase/structured.go` to check that mutations don't delete derived constraints

2) please remind me, are we bumping the format version of descriptors with partitioning? Otherwise an old node could delete the derived constraints because they don't know better.

---

Reviewed 10 of 10 files at r9, 9 of 9 files at r10.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20191#-:-L-Hwsj-yXTWc8Ayx3Z1:b-p2er2)*

benesch:
Ok, @knz, to address your comments:

1. Turns out mutations can't actually cause a derived constraint to be deleted. This is a bug! Filed #20630 to track.
2. Not bumping the format version (apparently format versions are on the way out), but we'll be sure to add a cluster migration before 2.0 and probably an acceptance tests as well.

Per @danhhz's advice, I'm going to merge this now to reduce some rebase load, but we'll be sure to address the versioning issue before feature freeze.

PR #20199, opened 2017-11-21
build: extract ui-maintainer-clean

couchand:
In the course of UI development I like to be able to run a clean against the UI files, almost always from within `pkg/ui`.  This seems like the lowest-touch way of enabling that workflow.


benesch (approved):
LGTM, thanks!

PR #20212, opened 2017-11-21
sql/stats: Add new cache for system.table_statistics

rytaft:
Add a new cache for the system.table_statistics table called
TableStatisticsCache. The cache is keyed by table ID, and each
entry stores all the statistics for a given table. The cache is
responsible for querying the underlying database in case of a
cache miss or requested Refresh. The eviction policy is LRU, and
entries are evicted when the cache exceeds a specified size.

For simplicity, this first implementation includes histogram data
in each entry. In the next iteration, we will split this cache into
two underlying caches: one keyed by table ID, and one keyed by
(tableID, statisticID). The former will not include histogram data,
but the latter will.

See docs/RFCS/20170908_sql_optimizer_statistics.md for additional
details.

Relsease note (sql): Added a cache for the system.table_statistics
table.

### DIFF
--- a/pkg/sql/stats/main_test.go
+++ b/pkg/sql/stats/main_test.go
@@ -1,0 +1,33 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package stats_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/cockroachdb/cockroach/pkg/security"
+	"github.com/cockroachdb/cockroach/pkg/security/securitytest"
+	"github.com/cockroachdb/cockroach/pkg/server"
+	"github.com/cockroachdb/cockroach/pkg/testutils/serverutils"
+	"github.com/cockroachdb/cockroach/pkg/util/randutil"
+)
+
+func TestMain(m *testing.M) {
+	security.SetAssetLoader(securitytest.EmbeddedAssets)
+	randutil.SeedForTests()
+	serverutils.InitTestServerFactory(server.TestServerFactory)
+	os.Exit(m.Run())
+}
--- a/pkg/sql/stats/stats_cache.go
+++ b/pkg/sql/stats/stats_cache.go
@@ -1,0 +1,270 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package stats
+
+import (
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/aws/aws-sdk-go/aws/awsutil"
+	"github.com/cockroachdb/cockroach/pkg/internal/client"
+	"github.com/cockroachdb/cockroach/pkg/sql/sem/tree"
+	"github.com/cockroachdb/cockroach/pkg/sql/sem/types"
+	"github.com/cockroachdb/cockroach/pkg/sql/sqlbase"
+	"github.com/cockroachdb/cockroach/pkg/sql/sqlutil"
+	"github.com/cockroachdb/cockroach/pkg/util/cache"
+	"github.com/cockroachdb/cockroach/pkg/util/log"
+	"github.com/cockroachdb/cockroach/pkg/util/protoutil"
+	"github.com/cockroachdb/cockroach/pkg/util/syncutil"
+	"github.com/pkg/errors"
+)
+
+// A TableStatistic object holds a statistic for a particular column or group
+// of columns. It mirrors the structure of the system.table_statistics table.
+type TableStatistic struct {
+	// The ID of the table.
+	TableID sqlbase.ID
+
+	// The ID for this statistic.  It need not be globally unique,
+	// but must be unique for this table.
+	StatisticID uint32
+
+	// Optional user-defined name for the statistic.
+	Name string
+
+	// The column ID(s) for which this statistic is generated.
+	ColumnIDs []sqlbase.ColumnID
+
+	// The time at which the statistic was created.
+	CreatedAt time.Time
+
+	// The total number of rows in the table.
+	RowCount uint64
+
+	// The estimated number of distinct values of the columns in ColumnIDs.
+	DistinctCount uint64
+
+	// The number of rows that have a NULL in any of the columns in ColumnIDs.
+	NullCount uint64
+
+	// Optional, and can only be set if there is a single column in ColumnIDs.
+	// Defines a histogram of the distribution of values in the column.
+	Histogram *HistogramData
+}
+
+func (s TableStatistic) String() string {
+	return awsutil.Prettify(s)
+}
+
+// A TableStatisticsCache is a cache of TableStatistic objects, keyed by
+// table ID.
+type TableStatisticsCache struct {
+	// NB: This can't be a RWMutex for lookup because UnorderedCache.Get
+	// manipulates an internal LRU list.
+	mu struct {
+		syncutil.Mutex
+		cache *cache.UnorderedCache
+	}
+	ClientDB    *client.DB
+	SQLExecutor sqlutil.InternalExecutor
+}
+
+// NewTableStatisticsCache creates a new TableStatisticsCache of the given size.
+// The underlying cache internally uses a hash map, so lookups are cheap.
+func NewTableStatisticsCache(
+	size int, db *client.DB, sqlExecutor sqlutil.InternalExecutor,
+) *TableStatisticsCache {
+	tableStatsCache := &TableStatisticsCache{
+		ClientDB:    db,
+		SQLExecutor: sqlExecutor,
+	}
+	tableStatsCache.mu.cache = cache.NewUnorderedCache(cache.Config{
+		Policy:      cache.CacheLRU,
+		ShouldEvict: func(s int, key, value interface{}) bool { return s > size },
+	})
+	return tableStatsCache
+}
+
+// LookupTableStats returns the cached statistics of the given table ID.
+// The second return value is true if the stats were found in the
+// cache, and false otherwise.
+func (sc *TableStatisticsCache) LookupTableStats(
+	ctx context.Context, tableID sqlbase.ID,
+) ([]*TableStatistic, bool) {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	if v, ok := sc.mu.cache.Get(tableID); ok {
+		if log.V(2) {
+			log.Infof(ctx, "r%d: lookup statistics for table: %s", tableID, v)
+		}
+		return v.([]*TableStatistic), true
+	}
+	if log.V(2) {
+		log.Infof(ctx, "r%d: lookup statistics for table: not found", tableID)
+	}
+	return nil, false
+}
+
+// Refresh updates the cached statistics for the given table ID
+// by issuing a query to system.table_statistics, and returns the statistics.
+func (sc *TableStatisticsCache) Refresh(
+	ctx context.Context, tableID sqlbase.ID,
+) ([]*TableStatistic, error) {
+	tableStatistics, err := sc.getTableStatistics(ctx, tableID)
+	if err != nil {
+		return nil, err
+	}
+
+	if log.V(2) {
+		log.Infof(ctx, "r%d: updating statistics for table: %s", tableID, tableStatistics)
+	}
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	sc.mu.cache.Add(tableID, tableStatistics)
+	return tableStatistics, nil
+}
+
+// GetTableStats is a convenience function to look up statistics for the
+// requested table ID in the cache using LookupTableStats, and if the stats
+// are not present in the cache, it looks them up in system.table_statistics
+// using Refresh.
+func (sc *TableStatisticsCache) GetTableStats(
+	ctx context.Context, tableID sqlbase.ID,
+) ([]*TableStatistic, error) {
+	if stats, ok := sc.LookupTableStats(ctx, tableID); ok {
+		return stats, nil
+	}
+	return sc.Refresh(ctx, tableID)
+}
+
+// Invalidate invalidates the cached statistics for the given table ID.
+func (sc *TableStatisticsCache) Invalidate(ctx context.Context, tableID sqlbase.ID) {
+	if log.V(2) {
+		log.Infof(ctx, "r%d: evicting statistics for table", tableID)
+	}
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	sc.mu.cache.Del(tableID)
+}
+
+const (
+	tableIDIndex = iota
+	statisticsIDIndex
+	nameIndex
+	columnIDsIndex
+	createdAtIndex
+	rowCountIndex
+	distinctCountIndex
+	nullCountIndex
+	histogramIndex
+	statsLen
+)
+
+// parseStats converts the given datums to a TableStatistic object. It only includes
+// a histogram in the output if the histogram in datums is non-null and
+// includeHistogram is true.
+func parseStats(datums tree.Datums, includeHistogram bool) (*TableStatistic, error) {
+	if datums == nil || datums.Len() == 0 {
+		return nil, nil
+	}
+
+	// Validate the input length.
+	if datums.Len() != statsLen {
+		return nil, errors.Errorf("%d values returned from table statistics lookup. Expected %d", datums.Len(), statsLen)
+	}
+
+	// Validate the input types.
+	expectedTypes := []struct {
+		fieldName    string
+		fieldIndex   int
+		expectedType types.T
+		nullable     bool
+	}{
+		{"tableID", tableIDIndex, types.Int, false},
+		{"statisticsID", statisticsIDIndex, types.Int, false},
+		{"name", nameIndex, types.String, true},
+		{"columnIDs", columnIDsIndex, types.TArray{Typ: types.Int}, false},
+		{"createdAt", createdAtIndex, types.Timestamp, false},
+		{"rowCount", rowCountIndex, types.Int, false},
+		{"distinctCount", distinctCountIndex, types.Int, false},
+		{"nullCount", nullCountIndex, types.Int, false},
+		{"histogram", histogramIndex, types.Bytes, true},
+	}
+	for _, v := range expectedTypes {
+		if datums[v.fieldIndex].ResolvedType() != v.expectedType &&
+			(!v.nullable || datums[v.fieldIndex].ResolvedType() != types.Null) {
+			return nil, errors.Errorf("%s returned from table statistics lookup has type %s. Expected %s",
+				v.fieldName, datums[v.fieldIndex].ResolvedType(), v.expectedType)
+		}
+	}
+
+	// Extract datum values.
+	tableStatistic := &TableStatistic{
+		TableID:       sqlbase.ID((int32)(*datums[tableIDIndex].(*tree.DInt))),
+		StatisticID:   (uint32)(*datums[statisticsIDIndex].(*tree.DInt)),
+		CreatedAt:     datums[createdAtIndex].(*tree.DTimestamp).Time,
+		RowCount:      (uint64)(*datums[rowCountIndex].(*tree.DInt)),
+		DistinctCount: (uint64)(*datums[distinctCountIndex].(*tree.DInt)),
+		NullCount:     (uint64)(*datums[nullCountIndex].(*tree.DInt)),
+	}
+	columnIDs := datums[columnIDsIndex].(*tree.DArray)
+	tableStatistic.ColumnIDs = make([]sqlbase.ColumnID, len(columnIDs.Array))
+	for i, d := range columnIDs.Array {
+		tableStatistic.ColumnIDs[i] = sqlbase.ColumnID((int32)(*d.(*tree.DInt)))
+	}
+	if datums[nameIndex].ResolvedType() == types.String {
+		tableStatistic.Name = string(*datums[nameIndex].(*tree.DString))
+	}
+	if includeHistogram && datums[histogramIndex].ResolvedType() == types.Bytes {
+		tableStatistic.Histogram = &HistogramData{}
+		if err := protoutil.Unmarshal([]byte(*datums[histogramIndex].(*tree.DBytes)), tableStatistic.Histogram); err != nil {
+			return nil, err
+		}
+	}
+
+	return tableStatistic, nil
+}
+
+// getTableStatistics retrieves the statistics in system.table_statistics
+// for the given table ID.
+func (sc *TableStatisticsCache) getTableStatistics(
+	ctx context.Context, tableID sqlbase.ID,
+) ([]*TableStatistic, error) {
+	const getTableStatisticsStmt = `
+SELECT "tableID", "statisticID", name, "columnIDs", "createdAt", "rowCount", "distinctCount", "nullCount", histogram
+FROM system.table_statistics
+WHERE "tableID" = $1
+`
+	var rows []tree.Datums
+	if err := sc.ClientDB.Txn(ctx, func(ctx context.Context, txn *client.Txn) error {
+		var err error
+		rows, err = sc.SQLExecutor.QueryRowsInTransaction(ctx, "get-table-statistics", txn, getTableStatisticsStmt, tableID)
+		return err
+	}); err != nil {
+		return nil, err
+	}
+
+	var statsList []*TableStatistic
+	for _, row := range rows {
+		stats, err := parseStats(row, true /* includeHistogram */)
+		if err != nil {
+			return nil, err
+		}
+		statsList = append(statsList, stats)
+	}
+
+	return statsList, nil
+}
--- a/pkg/sql/stats/stats_cache_test.go
+++ b/pkg/sql/stats/stats_cache_test.go
@@ -1,0 +1,240 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package stats_test
+
+import (
+	"testing"
+
+	"golang.org/x/net/context"
+
+	"time"
+
+	"reflect"
+
+	"sort"
+
+	"github.com/cockroachdb/cockroach/pkg/base"
+	"github.com/cockroachdb/cockroach/pkg/internal/client"
+	"github.com/cockroachdb/cockroach/pkg/sql"
+	"github.com/cockroachdb/cockroach/pkg/sql/sem/tree"
+	"github.com/cockroachdb/cockroach/pkg/sql/sem/types"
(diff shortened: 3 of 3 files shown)


rytaft:
@RaduBerinde I've added you as a reviewer, but no rush! No need to look at this until you return from vacation.

Thanks!

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-KzVZiT29OvuGrWNAnmJ:b-9l0zhx)*

knz:
I'm looking at this because Radu is away. Feel free to wait for additional review feedback from Radu. 

Overall the contribution is solid, I don't have anything to say about the gist of the change. My feedback is limited to small details. 

---

Reviewed 6 of 6 files at r1.

---

*[pkg/sql/stats/stats.proto, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_3WxxEilkYqE7t5Hq:-Kz_3WxxEilkYqE7t5Hr:b-mt71av) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats.proto#L26)):*
> ```protobuf
> message Stats {
>     // The ID of the table.
>   	uint32 table_id = 1 [
> ```

I think we do spaces and no tabs in these files too. (You can check.)

---

*[pkg/sql/stats/stats_cache.go, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_3esEYr1SOYYnG4yv:-Kz_3esFH5YlGeFutP5-:b-hco8t1) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L37)):*
> ```Go
> 	// NB: This can't be a RWMutex for lookup because UnorderedCache.Get
> 	// manipulates an internal LRU list.
> 	mu          syncutil.Mutex
> ```

The following is idiomatic Go:

```go
type TableStatisticsCache struct {
   mu struct {
        syncutil.Mutex
        cache *cache.UnorderedCache
   }
   SQLExecutor sqlutil.InternalExecutor
}
```

(I'm assuming the executor doesn't need to be locked)

Also this looks like the cache pointer needs to be locked, not the cache itself. Is there a way to put the mutex in the cache instead? For example:

```go
type LockedUnorderedCache {
   syncutil.Mutex
   cache.UnorderedCache
}

type TableStatisticsCache {
    muCache *LockedUnorderedCache
    ...
}
```

I'd then also re-define the UnorderedCache methods on LockedUnorderedCache to do the mutex lock in the method, instead of the caller.

(Mind you, I'm less confident about this one than the one above. Ask around in the team what they think.)

---

*[pkg/sql/stats/stats_cache.go, line 92 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_4Z_tlAeHHsEj-DbN:-Kz_4Z_uoV14URE2FIAI:b-wx3mml) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L92)):*
> ```Go
> 	defer sc.mu.Unlock()
> 	if log.V(2) {
> 		log.Infof(ctx, "r%d: updating statistics for table: %s", tableID, tableStatistics)
> ```

Maybe pull the logging above to not have it in the locked section.

---

*[pkg/sql/stats/stats_cache.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_4cNngRYTrTugvkoZ:-Kz_4cNo4O32IbZLUgPt:b-rqxvo7) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L116)):*
> ```Go
> 	defer sc.mu.Unlock()
> 	if log.V(2) {
> 		log.Infof(ctx, "r%d: evicting statistics for table", tableID)
> ```

ditto.

---

*[pkg/sql/stats/stats_cache.go, line 138 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_4nzJl39wVf5kekS7:-Kz_4nzK52CuGL_Mf53Y:bujl48n) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L138)):*
> ```Go
> // includeHistogram is true.
> func parseStats(datums tree.Datums, includeHistogram bool) (*Stats, error) {
> ```

The rest of our source code usually avoids an empty line between the function prototype and the first statement in the block.

---

*[pkg/sql/stats/stats_cache.go, line 174 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_52AN08F-IRmO_ius:-Kz_52AOaYvnxbaO0rL-:b-mlkvio) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L174)):*
> ```Go
> 
> 	// Extract datum values.
> 	tableStatistic := &Stats{}
> ```

I think it's more idiomatic to initialize all the fields that can be initialized this way, using this syntax:
```go
tableStatistic := &Stats{
    TableID: ...
    StatisticID: ....
    ...
}
```

I think it would also make the code (slightly) more compact.

---

*[pkg/sql/stats/stats_cache_test.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_6TZo4D5W__zcEvZR:-Kz_6TZo4D5W__zcEvZS:blj081d) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache_test.go#L54)):*
> ```Go
> INSERT INTO system.table_statistics ("tableID", "statisticID", name, "columnIDs", "createdAt",
> 	"rowCount", "distinctCount", "nullCount", histogram)
> VALUES($1, $2, $3, $4, $5, $6, $7, $8, $9)
> ```

nit: `VALUES (` with a space.

---

*[pkg/sql/stats/stats_cache_test.go, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_6WESV7MraL10s2HD:-Kz_6WESV7MraL10s2HE:bqitinm) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache_test.go#L66)):*
> ```Go
> 		stats.TableID,
> 		stats.StatisticID,
> 		nil, //name
> ```

nit: `// name` with a space

---

*[pkg/sql/stats/stats_cache_test.go, line 103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_6YFFCNqqXOm4A5WD:-Kz_6YFFCNqqXOm4A5WE:b-he8vg3) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache_test.go#L103)):*
> ```Go
> 	tableID sqlbase.ID,
> ) error {
> ```

nit: space (see above)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-Kz_6jHfYW_rwIPfsEhb:b-jn5b1x)*

rytaft:
Thanks for the review!

---

Review status: 3 of 6 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[pkg/sql/stats/stats.proto, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_3WxxEilkYqE7t5Hq:-Kz_NphFu-Za7eeu4a_u:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats.proto#L26)):*

I think we do spaces and no tabs in these files too. (You can check.)
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 37 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_3esEYr1SOYYnG4yv:-Kz_NtWF_AdYJFL5akki:b-bacgmf) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L37)):*

The following is idiomatic Go:

```go
type TableStatisticsCache struct {
   mu struct {
        syncutil.Mutex
        cache *cache.UnorderedCache
   }
   SQLExecutor sqlutil.InternalExecutor
}
```

(I'm assuming the executor doesn't need to be locked)

Also this looks like the cache pointer needs to be locked, not the cache itself. Is there a way to put the mutex in the cache instead? For example:

```go
type LockedUnorderedCache {
   syncutil.Mutex
   cache.UnorderedCache
}

type TableStatisticsCache {
    muCache *LockedUnorderedCache
    ...
}
```

I'd then also re-define the UnorderedCache methods on LockedUnorderedCache to do the mutex lock in the method, instead of the caller.

(Mind you, I'm less confident about this one than the one above. Ask around in the team what they think.)
</blockquote></details>

Thanks for the comment! I've implemented the first suggestion.  Holding off on the second based on our chat with @andreimatei.

---

*[pkg/sql/stats/stats_cache.go, line 92 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_4Z_tlAeHHsEj-DbN:-Kz_OQhnsLU8RSz8VhTi:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L92)):*

Maybe pull the logging above to not have it in the locked section.
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 116 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_4cNngRYTrTugvkoZ:-Kz_OSnVnfSEuAwliT4Z:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L116)):*

ditto.
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 138 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_4nzJl39wVf5kekS7:-Kz_OU0DSFJF31OjY6Gc:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L138)):*

The rest of our source code usually avoids an empty line between the function prototype and the first statement in the block.
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 174 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_52AN08F-IRmO_ius:-Kz_OVyEF4NXJ9kRorAq:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache.go#L174)):*

I think it's more idiomatic to initialize all the fields that can be initialized this way, using this syntax:
```go
tableStatistic := &Stats{
    TableID: ...
    StatisticID: ....
    ...
}
```

I think it would also make the code (slightly) more compact.
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache_test.go, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_6TZo4D5W__zcEvZR:-Kz_Oi-9sr15_vy2nf2U:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache_test.go#L54)):*

nit: `VALUES (` with a space.
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache_test.go, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_6WESV7MraL10s2HD:-Kz_Oins-WtEwUynpA-C:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache_test.go#L66)):*

nit: `// name` with a space
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache_test.go, line 103 at r1](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-Kz_6YFFCNqqXOm4A5WD:-Kz_OjobZEOOLejyezMO:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/94f4a05e76f7b922e544501654e000184c23a6b3/pkg/sql/stats/stats_cache_test.go#L103)):*

nit: space (see above)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-Kz_NkdUmkyA6yg7a4yz:b-9qdoz3)*

petermattis:
Review status: 3 of 6 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[pkg/sql/stats/stats.proto, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzdO6kYgY7-HjlV4QYe:-KzdO6kYgY7-HjlV4QYf:bnh4348) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats.proto#L24)):*
> ```protobuf
> 
> // Stats encodes the statistics stored in system.table_statistics
> message Stats {
> ```

Does this need to be a proto? Is it been marshaled and unmarshaled anywhere? If it isn't, there are some downsides to using a proto: it constrains the data structure that can be expressed. I don't have a concrete example of something I'd do differently, only that my expectation was that `Stats` was going to be a plain-Go-struct.

---

*[pkg/sql/stats/stats_cache.go, line 216 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzdPKnukbDQuMGOaEqI:-KzdPKnukbDQuMGOaEqJ:b357ctm) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L216)):*
> ```Go
> 	}
> 
> 	statsList := make([]*Stats, 0)
> ```

This can be `var statsList []*Stats`. A `nil` slice is almost equivalent to an empty slice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212)*

RaduBerinde:
Great stuff, that was quick. Some minor comments.

---

Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[pkg/sql/stats/stats.proto, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzdO6kYgY7-HjlV4QYe:-KziDNQQRl_vrnwCuWJe:b-rxehj5) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats.proto#L24)):*

Does this need to be a proto? Is it been marshaled and unmarshaled anywhere? If it isn't, there are some downsides to using a proto: it constrains the data structure that can be expressed. I don't have a concrete example of something I'd do differently, only that my expectation was that `Stats` was going to be a plain-Go-struct.
</blockquote></details>

+1, I don't think this needs to be a proto. The only reason I would see is if we wanted to store the stats in the cache in a marshaled format to save memory, but I don't think we would be saving all that much (and it would lead to unmarshaling+allocations every time we need them).

If we switch this to a go struct, then there is a question of whether we want it to store the histogram in a more convenient format. Namely, something that has `Datum`s or `EncDatum`s in the buckets instead of encoded `[]bytes`. We can worry about this later though, we can keep it in the HistogramData format for now.

---

*[pkg/sql/stats/stats.proto, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziEXEdWte1RwZLoZeW:-KziEXEdWte1RwZLoZeX:bm5nyet) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats.proto#L53)):*
> ```protobuf
> 
>     // The total number of rows in the table.
>     uint32 row_count = 6;
> ```

These should be 64-bit. 4 billion rows is not that insane these days :)

---

*[pkg/sql/stats/stats_cache.go, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziF89tjZF15_Tqp9R8:-KziF89tjZF15_Tqp9R9:b-ravaa4) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L85)):*
> ```Go
> 	}
> 	if tableStatistics == nil {
> 		if log.V(2) {
> ```

In what case does this happen? Seems like the slice is always allocated unless there's an error.

Maybe you wanted `len(tableStatistics) == 0`? If that is the case, I think we should still add it to the cache. If we run repeated queries against a table that has no stats, we don't want to keep querying the stats table every time.

---

*[pkg/sql/stats/stats_cache.go, line 216 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzdPKnukbDQuMGOaEqI:-KziF2FAsLB7tv00EXgG:b-spx22s) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L216)):*

This can be `var statsList []*Stats`. A `nil` slice is almost equivalent to an empty slice.
</blockquote></details>

I see that we are comparing the result of this function with `nil`, which is very subtle logic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-KziFerYW5IslTUHhDUo:b-2h0ku6)*

RaduBerinde:
Review status: 3 of 6 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[pkg/sql/stats/stats_cache.go, line 48 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziMQM6JLe_gRx23gp3:-KziMQM6JLe_gRx23gp4:b-o804tn) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L48)):*
> ```Go
> // are cheap.
> func NewTableStatisticsCache(
> 	size func() int64, sqlExecutor sqlutil.InternalExecutor,
> ```

What is the purpose of passing a function here? If we want to support changing the size dynamically, we should provide APIs (e.g. `TableStatisticsCache.Resize()`) and implement this function internally instead of the caller having to implement it. We can keep it simple for now and just pass in a static size though.

---

*[pkg/sql/stats/stats_cache.go, line 50 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziLKMXySdzpZ00QlzD:-KziLKMXySdzpZ00QlzE:butcz7y) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L50)):*
> ```Go
> 	size func() int64, sqlExecutor sqlutil.InternalExecutor,
> ) *TableStatisticsCache {
> 	tableStatsCache := &TableStatisticsCache{}
> ```

[nit] {SQLExecutor: sqlExecutor}

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212)*

RaduBerinde:
Review status: 3 of 6 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[pkg/sql/stats/stats_cache.go, line 104 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzsfmJn1o93rDeW9Abg:-KzsfmJn1o93rDeW9Abh:bs0evju) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L104)):*
> ```Go
> // and if the stats are not present in the cache, it looks them up in
> // system.table_statistics using Refresh.
> func (sc *TableStatisticsCache) LookupAndMaybeRefresh(
> ```

I don't think we should pass a `txn` to this function. If we need to read from the table, we should do it in a separate internal transaction (i.e. do `db.Txn(..)` ourselves; we may need to plumb a `*client.DB`). We want this small transaction to be restarted if necessary without affecting whatever larger transaction we are running.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212)*

rytaft:
TFTRs!

---

Review status: 1 of 3 files reviewed at latest revision, 16 unresolved discussions, all commit checks successful.

---

*[pkg/sql/stats/stats_cache.go, line 48 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziMQM6JLe_gRx23gp3:-KzzE1AQ6U_TM_g6Almm:b-v3l7go) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L48)):*

What is the purpose of passing a function here? If we want to support changing the size dynamically, we should provide APIs (e.g. `TableStatisticsCache.Resize()`) and implement this function internally instead of the caller having to implement it. We can keep it simple for now and just pass in a static size though.
</blockquote></details>

I just copied what was done for leaseholder_cache.go.  I've changed it to a static size now.

---

*[pkg/sql/stats/stats_cache.go, line 50 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziLKMXySdzpZ00QlzD:-KzzECCaYNfdRM5N_Nb_:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L50)):*

[nit] {SQLExecutor: sqlExecutor}
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziF89tjZF15_Tqp9R8:-KzzEG3jjFzVHXMvONDc:b-xhzxhr) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L85)):*

In what case does this happen? Seems like the slice is always allocated unless there's an error.

Maybe you wanted `len(tableStatistics) == 0`? If that is the case, I think we should still add it to the cache. If we run repeated queries against a table that has no stats, we don't want to keep querying the stats table every time.
</blockquote></details>

Good point! The cache now stores nil values, and I added a test case for this as well.

---

*[pkg/sql/stats/stats_cache.go, line 104 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzsfmJn1o93rDeW9Abg:-KzzER5Jtdpi6xAofd8g:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L104)):*

I don't think we should pass a `txn` to this function. If we need to read from the table, we should do it in a separate internal transaction (i.e. do `db.Txn(..)` ourselves; we may need to plumb a `*client.DB`). We want this small transaction to be restarted if necessary without affecting whatever larger transaction we are running.
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 216 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzdPKnukbDQuMGOaEqI:-KzzEbj2Y4k3y6kGhZbo:bmarvmz) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats_cache.go#L216)):*

I see that we are comparing the result of this function with `nil`, which is very subtle logic.
</blockquote></details>

I'm no longer comparing the result with `nil`, so I've made the change @petermattis suggested.

---

*[pkg/sql/stats/stats.proto, line 24 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzdO6kYgY7-HjlV4QYe:-KzzEtyZlfl3iIktm7tx:begpzu0) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats.proto#L24)):*

+1, I don't think this needs to be a proto. The only reason I would see is if we wanted to store the stats in the cache in a marshaled format to save memory, but I don't think we would be saving all that much (and it would lead to unmarshaling+allocations every time we need them).

If we switch this to a go struct, then there is a question of whether we want it to store the histogram in a more convenient format. Namely, something that has `Datum`s or `EncDatum`s in the buckets instead of encoded `[]bytes`. We can worry about this later though, we can keep it in the HistogramData format for now.
</blockquote></details>

Done - changed it to a Go struct.

---

*[pkg/sql/stats/stats.proto, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KziEXEdWte1RwZLoZeW:-KzzEuLYXh3E5RMK0mVf:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/f7c82527896fff886e13fe56374ee794f35dd6ff/pkg/sql/stats/stats.proto#L53)):*

These should be 64-bit. 4 billion rows is not that insane these days :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-KzzFQFXYBOtiGlJYej3:bz5mair)*

RaduBerinde:
:lgtm_strong: 

---

Review status: 1 of 3 files reviewed at latest revision, 13 unresolved discussions, all commit checks successful.

---

*[pkg/sql/stats/stats_cache.go, line 137 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzzJzQfb9tRthz73CBY:-KzzJzQfb9tRthz73CBZ:btyj3ep) ([raw file](https://github.com/cockroachdb/cockroach/blob/772fefca6e71bb1a093cf9e9070877f1c80d48bf/pkg/sql/stats/stats_cache.go#L137)):*
> ```Go
> // the requested table ID in the cache using Lookup, and if the stats are not
> // present in the cache, it looks them up in system.table_statistics using Refresh.
> func (sc *TableStatisticsCache) LookupAndMaybeRefresh(
> ```

[nit] I expect this to be the main API, and the name is kind of awkward. It's also not specific enough - in the future we will have separate APIs for retrieving histograms. Maybe rename to `GetTableStats` and change `Lookup` to `LookupTableStats` or `TryLookupTableStats`. 

---

*[pkg/sql/stats/stats_cache.go, line 138 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzzJsA5h68_N7HX3o5U:-KzzJsA5h68_N7HX3o5V:bh7re2g) ([raw file](https://github.com/cockroachdb/cockroach/blob/772fefca6e71bb1a093cf9e9070877f1c80d48bf/pkg/sql/stats/stats_cache.go#L138)):*
> ```Go
> // present in the cache, it looks them up in system.table_statistics using Refresh.
> func (sc *TableStatisticsCache) LookupAndMaybeRefresh(
> 	ctx context.Context, db *client.DB, tableID sqlbase.ID,
> ```

[nit] we should put the client.DB in `TableStatisticsCache`, it will be associated with a single server.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-KzzKuJOZq_07MtqnqQ4:b-uwkowd)*

rytaft:
Thanks! Made the updates.

---

Review status: 1 of 3 files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[pkg/sql/stats/stats_cache.go, line 137 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzzJzQfb9tRthz73CBY:-Kzzgo8uUlp1SpUKqQIY:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/772fefca6e71bb1a093cf9e9070877f1c80d48bf/pkg/sql/stats/stats_cache.go#L137)):*

[nit] I expect this to be the main API, and the name is kind of awkward. It's also not specific enough - in the future we will have separate APIs for retrieving histograms. Maybe rename to `GetTableStats` and change `Lookup` to `LookupTableStats` or `TryLookupTableStats`. 
</blockquote></details>

Done.

---

*[pkg/sql/stats/stats_cache.go, line 138 at r3](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-KzzJsA5h68_N7HX3o5U:-KzzgrSR9iFp1fhKjw3o:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/772fefca6e71bb1a093cf9e9070877f1c80d48bf/pkg/sql/stats/stats_cache.go#L138)):*

[nit] we should put the client.DB in `TableStatisticsCache`, it will be associated with a single server.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20212#-:-KzzgtMi366dia4eBAXD:b-8863sf)*

PR #20214, opened 2017-11-21
sql: support json_object_keys of JSON builtins

taku-k:
I implemented `json{b}_object_keys()` of JSON builtin functions.
However, I'm anxious that `json_object_keys()` returns sorted keys depending on the internal data structure. PostgreSQL's `json_object_keys()` returns keys in a sequential order.

Ref: #20120

### DIFF
--- a/pkg/util/json/iterator.go
+++ b/pkg/util/json/iterator.go
@@ -1,0 +1,31 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package json
+
+// ObjectKeyIterator is an iterator to access the keys of an object.
+type ObjectKeyIterator struct {
+	src jsonObject
+	idx int
+}
+
+// Next returns true and the next key in the iterator if one exists,
+// and false otherwise.
+func (it *ObjectKeyIterator) Next() (bool, string) {
+	it.idx++
+	if it.idx >= len(it.src) {
+		return false, ""
+	}
+	return true, string(it.src[it.idx].k)
+}
--- a/pkg/util/json/json.go
+++ b/pkg/util/json/json.go
@@ -89,6 +89,10 @@ type JSON interface {
 	// Exists implements the `?` operator.
 	Exists(string) bool
 
+	// IterObjectKey returns an ObjectKeyIterator, and it returns error if the obj
+	// is not an object.
+	IterObjectKey() (*ObjectKeyIterator, error)
+
 	// isScalar returns whether the JSON document is null, true, false, a string,
 	// or a number.
 	isScalar() bool
@@ -699,6 +703,34 @@ func (j jsonArray) Exists(s string) bool {
 }
 func (j jsonObject) Exists(s string) bool {
 	return j.FetchValKey(s) != nil
+}
+
+var errIterateKeysNonObject = pgerror.NewError(pgerror.CodeInvalidParameterValueError,
+	"cannot iterate keys of non-object")
+
+func (jsonNull) IterObjectKey() (*ObjectKeyIterator, error) {
+	return nil, errIterateKeysNonObject
+}
+func (jsonTrue) IterObjectKey() (*ObjectKeyIterator, error) {
+	return nil, errIterateKeysNonObject
+}
+func (jsonFalse) IterObjectKey() (*ObjectKeyIterator, error) {
+	return nil, errIterateKeysNonObject
+}
+func (jsonNumber) IterObjectKey() (*ObjectKeyIterator, error) {
+	return nil, errIterateKeysNonObject
+}
+func (jsonString) IterObjectKey() (*ObjectKeyIterator, error) {
+	return nil, errIterateKeysNonObject
+}
+func (jsonArray) IterObjectKey() (*ObjectKeyIterator, error) {
+	return nil, errIterateKeysNonObject
+}
+func (j jsonObject) IterObjectKey() (*ObjectKeyIterator, error) {
+	return &ObjectKeyIterator{
+		src: j,
+		idx: -1,
+	}, nil
 }
 
 func (jsonNull) isScalar() bool   { return true }
--- a/docs/generated/sql/functions.md
+++ b/docs/generated/sql/functions.md
@@ -752,9 +752,13 @@ Compatible elements: hour, minute, second, millisecond, microsecond.</p>
 </span></td></tr>
 <tr><td><code>json_array_elements_text(input: jsonb) &rarr; setof tuple{string}</code></td><td><span class="funcdesc"><p>Expands a JSON array to a set of text values.</p>
 </span></td></tr>
+<tr><td><code>json_object_keys(input: jsonb) &rarr; setof tuple{string}</code></td><td><span class="funcdesc"><p>Returns sorted set of keys in the outermost JSON object.</p>
+</span></td></tr>
 <tr><td><code>jsonb_array_elements(input: jsonb) &rarr; setof tuple{jsonb}</code></td><td><span class="funcdesc"><p>Expands a JSON array to a set of JSON values.</p>
 </span></td></tr>
 <tr><td><code>jsonb_array_elements_text(input: jsonb) &rarr; setof tuple{string}</code></td><td><span class="funcdesc"><p>Expands a JSON array to a set of text values.</p>
+</span></td></tr>
+<tr><td><code>jsonb_object_keys(input: jsonb) &rarr; setof tuple{string}</code></td><td><span class="funcdesc"><p>Returns sorted set of keys in the outermost JSON object.</p>
 </span></td></tr>
 <tr><td><code>oid(int: <a href="int.html">int</a>) &rarr; oid</code></td><td><span class="funcdesc"><p>Converts an integer to an OID.</p>
 </span></td></tr>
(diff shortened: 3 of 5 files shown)


justinj:
Hi @taku-k!

It's a fair point that we don't implement `JSON` having the same semantics as Postgres, we actually only implement `JSONB` and alias `JSON` to it. This decision is probably debatable but I'm not too bothered by not supporting the exact semantics of Postgres's `JSON`, since its use is generally discouraged with the introduction of `JSONB`.

While I do like the approach you've taken here, unfortunately it's not going to play super well with an upcoming change to JSON in Cockroach which will allow operating directly on JSON values which are still encoded, since we don't yet have particularly fast indexing of JSON keys in that case.

What do you think of an approach that would involve exposing some kind of iterator that would allow you to iterate over the keys of an object? That way we could just re-implement it for encoded JSON objects and everything will still work fine (I locally have such an implementation for still-encoded JSON objects). This has the added benefit of not prescribing a particular ordering of the keys from outside of the `json` package like `FetchKeyIdx` does, and I think it will be generally useful in many other cases as well!

taku-k:
Hi, @justinj 
Thank you! I understand the relationship between JSON and JSONB.

I also think that introducing new iterator over the keys of an object is better for this case, so I'll implement it! Thank you for your advice.

pkg/util/json/json.go
```diff
@@ -701,6 +705,23 @@ func (j jsonObject) Exists(s string) bool {
...
 }
 
+func (jsonNull) ObjectKeyIterator() <-chan JSON   { return nil }
+func (jsonTrue) ObjectKeyIterator() <-chan JSON   { return nil }
+func (jsonFalse) ObjectKeyIterator() <-chan JSON  { return nil }
+func (jsonNumber) ObjectKeyIterator() <-chan JSON { return nil }
+func (jsonString) ObjectKeyIterator() <-chan JSON { return nil }
+func (jsonArray) ObjectKeyIterator() <-chan JSON  { return nil }
+func (j jsonObject) ObjectKeyIterator() <-chan JSON {
+	ch := make(chan JSON)
```
justinj: This is cool, and very simple! But I have a couple anxieties about using channels and async code where sync code will suffice, besides a general fear of raceyness: firstly, I think they're pretty slow in general (though I'd have to benchmark to tell you exactly how slow), since every pull/push will have to go through the go runtime scheduler, but secondly and a little more concerning is that it's not completely clear to me if the generators will always be sure to completely exhaust their values, meaning if somehow someone only accesses a subset of the object values, we could end up with this channel never getting closed and this goroutine living forever (after writing this I see that the generators have a `Close` method which we could possibly use, but I'm still a little nervous).

It's a little more syntactically loud, but I much prefer a pattern like:
```
type ObjectKeyIterator struct {
  source jsonObject
  idx int
}
// Next returns true and the next key in the iterator if one exists, and false otherwise.
func (o *ObjectKeyIterator) Next() (more bool, key string) {
...
}
```

And that way we're safe because if we lose a handle onto the iterator then it will still just get garbage collected eventually.

What do you think?

pkg/util/json/json.go
```diff
@@ -89,6 +89,10 @@ type JSON interface {
 	// Exists implements the `?` operator.
 	Exists(string) bool
 
+	// ObjectKeyIterator is an iterator to access the keys of an object, and it returns nil if the obj
+	// is not an object.
+	ObjectKeyIterator() <-chan JSON
```
justinj: Would you mind naming this `IterObjectKeys`? Just because it will match up better with the iterators introduced in #20217.

Also, I think this might be more general and useful if it returned strings rather than JSON? That's fine because the keys are always strings.

pkg/util/json/json.go
taku-k: You're right. It is a bit dangerous to expose the channel outside the function.
I'll implement it with iterator pattern. Thank you!

pkg/util/json/json.go
taku-k: I think so too. I'll do that!

pkg/util/json/json.go
```diff
@@ -89,6 +89,10 @@ type JSON interface {
 	// Exists implements the `?` operator.
 	Exists(string) bool
 
+	// IterObjectKey returns an ObjectKeyIterator, and it returns nil if the obj
+	// is not an object.
+	IterObjectKey() *ObjectKeyIterator
```
justinj: Perhaps we could just push the error down into this function, then it doesn't have to return a pointer (so the non-object cases would return an error with the "cannot be called on non-object" message)? I think it's fine if the error message for this case lives in the `json` package. It seems sensible to me to consider this an error case.

pkg/util/json/iterator.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// Next returns true and the next key in the iterator if one exists,
+// and false otherwise.
+func (it *ObjectKeyIterator) Next() (bool, string) {
+	it.idx++
+	if it.idx >= len(it.src) {
+		return false, ""
+	}
+	return true, *it.src[it.idx].k.AsText()
```
justinj: Since `k` is a `jsonString` and not just `JSON`, I think this can just be `string(it.src[it.idx].k)` and we don't have to mess with pointers.

justinj (commented):
Hi @taku-k! Sorry for the delay, it's been a long weekend here. This looks really great - I just have two small comments, and then I think after a squash this will be good to merge!

taku-k:
@justinj Thank you for the review! Could you take another look?

justinj:
Looking good! Couple more small comments. Sorry for the delay in getting back to you - I've been traveling a bit.

---

Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/sem/builtins/generator_builtins.go, line 471 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-L-CaX9LAGZvrZAY1PGp:-L-CaX9LAGZvrZAY1PGq:b85bdcn) ([raw file](https://github.com/cockroachdb/cockroach/blob/36ac990facc0e9e992e7170be3601ea9b7aab8ad/pkg/sql/sem/builtins/generator_builtins.go#L471)):*
> ```Go
> ) (tree.ValueGenerator, error) {
> 	target := tree.MustBeDJSON(args[0])
> 	if target.Type() != json.ObjectJSONType {
> ```

Since we have `IterObjectKey` returning an error now I think this check is redundant? We can make the error returned from `IterObjectKey` a bit more general like `cannot iterate keys of non-object` or something and then just pass that up?

---

*[pkg/util/json/json.go, line 710 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-L-CapkCOlwosJEXacxk:-L-CapkCOlwosJEXacxl:b-wm4r20) ([raw file](https://github.com/cockroachdb/cockroach/blob/36ac990facc0e9e992e7170be3601ea9b7aab8ad/pkg/util/json/json.go#L710)):*
> ```Go
> }
> 
> var errCreatedForNonObject = errors.New("cannot be created for non-object")
> ```

Would you mind using `pgerror.NewError` for this? The reason is that we tag those with Postgres error codes so that clients can understand what went wrong:
```
var errCannotIterNonObjectKeys = pgerror.NewError(pgerror.CodeInvalidParameterValueError, "cannot iterate keys of non-object")
```

---

*[pkg/util/json/json.go, line 713 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-L-CakOyQMEgfY1MMjWz:-L-CakOzanzkXpdXTko7:bi8yvjo) ([raw file](https://github.com/cockroachdb/cockroach/blob/36ac990facc0e9e992e7170be3601ea9b7aab8ad/pkg/util/json/json.go#L713)):*
> ```Go
> 
> func (jsonNull) IterObjectKey() (*ObjectKeyIterator, error) {
> 	return &ObjectKeyIterator{}, errCreatedForNonObject
> ```

These can just return `nil` for the iterator instead of allocating one regardless.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-:-L-CbPG9WpCpZLnAms4j:batsxfm)*

taku-k:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/sql/sem/builtins/generator_builtins.go, line 471 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-L-CaX9LAGZvrZAY1PGp:-L-FRCgG-2o_gWUdC5ks:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/36ac990facc0e9e992e7170be3601ea9b7aab8ad/pkg/sql/sem/builtins/generator_builtins.go#L471)):*

Since we have `IterObjectKey` returning an error now I think this check is redundant? We can make the error returned from `IterObjectKey` a bit more general like `cannot iterate keys of non-object` or something and then just pass that up?
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 710 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-L-CapkCOlwosJEXacxk:-L-FRMGfQxgbZAo49Y_0:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/36ac990facc0e9e992e7170be3601ea9b7aab8ad/pkg/util/json/json.go#L710)):*

Would you mind using `pgerror.NewError` for this? The reason is that we tag those with Postgres error codes so that clients can understand what went wrong:
```
var errCannotIterNonObjectKeys = pgerror.NewError(pgerror.CodeInvalidParameterValueError, "cannot iterate keys of non-object")
```
</blockquote></details>

Done.

---

*[pkg/util/json/json.go, line 713 at r6](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214#-L-CakOyQMEgfY1MMjWz:-L-FROheHawLZwyBAL61:b-896fix) ([raw file](https://github.com/cockroachdb/cockroach/blob/36ac990facc0e9e992e7170be3601ea9b7aab8ad/pkg/util/json/json.go#L713)):*

These can just return `nil` for the iterator instead of allocating one regardless.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/cockroach/20214)*

justinj:
Looks great! Would you mind squashing?

taku-k:
@justinj I have done!

justinj:
Thanks!

taku-k:
Thank you for your review.

